PR #10, opened 2012-12-24
If pygments is installed, produce colorized json output by default.

aisipos:
Fall back to standard uncolorized output if pygments is not installed.


Kizilaslan:
MKizilaslan

JordonPhillips:
Closing this out since this isn't the way we want to go with this feature. Feature request located at #2123.

PR #523, opened 2013-12-03
Drop requirements for ListObjects in s3 cp command

### LINKED ISSUE #328: `s3 cp <object>` requires ListBucket permissions
The aws cli can't copy an object out of s3 unless the user has ListBucket permissions for the object's bucket.

Steps to reproduce:
1. Create an s3 bucket called `test-bucket`, or use an existing bucket.
2. Add an object to that bucket called `test-object`, or use an existing object.
3. Create a new user, and give it a policy like this:
   
   ```
   {
     "Version": "2012-10-17",
     "Statement": [
       {
         "Action": [
           "s3:GetObject"
         ],
         "Sid": "Stmt1378493159000",
         "Resource": [
           "arn:aws:s3:::test-bucket/*",
           "arn:aws:s3:::test-bucket"
         ],
         "Effect": "Allow"
       }
     ]
   }
   ```
4. Install the new user's credentials in your environment.
5. Run:
   
   ```
   $ aws --debug --region us-west-1 s3 cp s3://test-bucket/test-object /tmp/object
   ```
6. You should see the following output:
   
   ```
   2013-09-06 11:58:01,700 - botocore.auth - DEBUG - HTTP request method: GET
   2013-09-06 11:58:01,700 - botocore.auth - DEBUG - StringToSign:
   GET
   
   
   Fri, 06 Sep 2013 18:58:01 GMT
   /test-bucket/
   2013-09-06 11:58:01,709 - botocore.endpoint - DEBUG - Sending http request: <PreparedRequest [GET]>
   2013-09-06 11:58:01,789 - botocore.response - DEBUG - Response Body:
   <?xml version="1.0" encoding="UTF-8"?>
   <Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>132435C85AAD5EB0</RequestId><HostId>3t89YS52j6/GJQZXU/xQ4r8nvRqwzp7Qe8d6HYwF072oVpthXcTxFKH5ITYe8oML</HostId></Error>
   2013-09-06 11:58:01,790 - botocore.hooks - DEBUG - Event needs-retry.s3.ListObjects: calling handler <botocore.retryhandler.RetryHandler object at 0x10dca3990>
   2013-09-06 11:58:01,790 - botocore.retryhandler - DEBUG - No retry needed.
   2013-09-06 11:58:01,790 - botocore.hooks - DEBUG - Event after-call.s3.ListObjects: calling handler <awscli.errorhandler.ErrorHandler object at 0x10dc8ce90>
   2013-09-06 11:58:01,790 - awscli.errorhandler - DEBUG - HTTP Response Code: 403
   2013-09-06 11:58:01,790 - awscli.clidriver - DEBUG - Exception caught in main()
     File "/Library/Python/2.7/site-packages/awscli/clidriver.py", line 184, in main
     File "/Library/Python/2.7/site-packages/awscli/customizations/s3/s3.py", line 308, in __call__
       return self.op_table[parsed_args.operation](remaining, parsed_globals)
     File "/Library/Python/2.7/site-packages/awscli/customizations/s3/s3.py", line 397, in __call__
       cmd_params.add_paths(parsed_args.paths)
     File "/Library/Python/2.7/site-packages/awscli/customizations/s3/s3.py", line 614, in add_paths
       self.check_src_path(paths)
     File "/Library/Python/2.7/site-packages/awscli/customizations/s3/s3.py", line 677, in check_src_path
       delimiter='/')
     File "/Library/Python/2.7/site-packages/botocore/operation.py", line 82, in call
       parsed=response[1])
     File "/Library/Python/2.7/site-packages/botocore/session.py", line 550, in emit
       return self._events.emit(event_name, **kwargs)
     File "/Library/Python/2.7/site-packages/botocore/hooks.py", line 158, in emit
       response = handler(**kwargs)
     File "/Library/Python/2.7/site-packages/awscli/errorhandler.py", line 50, in __call__
       raise ClientError(msg)
   ClientError: A client error (AccessDenied) occurred: Access Denied
   2013-09-06 11:58:01,791 - awscli.clidriver - DEBUG - Exiting with rc 255
   A client error (AccessDenied) occurred: Access Denied
   ```

It appears that the solution is to use the `Verify=false` option in the appropriate `boto` call (which I haven't had time to dive in and locate). See: http://stackoverflow.com/questions/11478752/with-the-boto-library-can-i-avoid-granting-list-permissions-on-a-base-bucket-in.

Thanks,
Mark

jamesls:
This fixes #328.

The issue is that before running an `aws s3 cp` command we verify that the src and destination argument exist.  When either arg is an s3 location we verify the bucket exists.  This means that for uploads/downloads we require the `ListObjects` command (what we use to verify the bucket exists), which is problematic if you're trying to create an IAM policy that does not have the `ListObjects` command.  The policy in #328 would give a user the ability to download an object, but they wouldn't be able to see the contents in a bucket.

In order to fix this, a few changes were needed:
- First remove the src/dest pre-check.  Now we just try the operation without checking if the resources exist.
- Make the error messages more clear.  Now it's possible to have a number of errors when an operation is run.  Rather than immediately stopping, we keep going (it's not guaranteed that all operations will fail, the user could have access to only a specific prefix).  To accomodate this scenario, the error messages have been updated to print all on the same line, and the operation name is now included in the error message:

```
upload failed: ./file-0 to s3://test-bucket/file-0 A client error (AccessDenied) occurred when calling the PutObject operation: Access Denied
```

Also, the `filegenerator` module needed an update to not require a `ListObjects` call if we're only downloading a single file (https://github.com/jamesls/aws-cli/commit/2e7abb631f8a25241d419d3e1bb2486c4190a576).  One consequence of this change is that we no longer have information about the `last_update` time for the file, which we were previously using to set the mtime of the file (so it would play nicely with subsequent `sync` operations).  We're losing that information now, so the file's mtime will just be the time when the file is downloaded.

### DIFF
--- a/awscli/customizations/s3/tasks.py
+++ b/awscli/customizations/s3/tasks.py
@@ -70,6 +70,7 @@ class BasicTask(object):
             self._queue_print_message(self.filename, failed=True,
                                       dryrun=self.parameters['dryrun'],
                                       error_message=last_error)
+            return
         filename = self.filename
         try:
             if not self.parameters['dryrun']:
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -27,37 +27,8 @@ import botocore.session
 
 from tests.integration import aws
 from tests.unit.customizations.s3 import create_bucket as _create_bucket
+from tests.unit import FileCreator
 from awscli.customizations.s3 import constants
-
-
-class FileCreator(object):
-    def __init__(self):
-        self.rootdir = tempfile.mkdtemp()
-
-    def remove_all(self):
-        shutil.rmtree(self.rootdir)
-
-    def create_file(self, filename, contents):
-        """Creates a file in a tmpdir
-
-        ``filename`` should be a relative path, e.g. "foo/bar/baz.txt"
-        It will be translated into a full path in a tmp dir.
-
-        Returns the full path to the file.
-        """
-        full_path = os.path.join(self.rootdir, filename)
-        if not os.path.isdir(os.path.dirname(full_path)):
-            os.makedirs(os.path.dirname(full_path))
-        with open(full_path, 'w') as f:
-            f.write(contents)
-        return full_path
-
-    def full_path(self, filename):
-        """Translate relative path to full path in temp dir.
-
-        f.full_path('foo/bar.txt') -> /tmp/asdfasd/foo/bar.txt
-        """
-        return os.path.join(self.rootdir, filename)
 
 
 class BaseS3CLICommand(unittest.TestCase):
--- a/tests/unit/__init__.py
+++ b/tests/unit/__init__.py
@@ -14,6 +14,8 @@ from tests import unittest
 import os
 import copy
 import logging
+import tempfile
+import shutil
 
 import mock
 import six
@@ -45,6 +47,7 @@ class BaseAWSCommandParamsTest(unittest.TestCase):
         self.parsed_response = {}
         self.make_request_patch = mock.patch('botocore.endpoint.Endpoint.make_request')
         self.make_request_is_patched = False
+        self.operations_called = []
 
     def tearDown(self):
         # This clears all the previous registrations.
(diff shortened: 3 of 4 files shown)


dstromberg:
Did these changes make it into a production release?  I see above that it was merged into aws:develop on Dec 4th, 2013, but I still have problems uploading to a bucket without ListObjects access using aws-cli/1.11.40 and aws-cli/1.11.65 .

PR #554, opened 2013-12-13
Evaluate include/exclude filters from the source dir, not the CWD

### LINKED ISSUE #548: Define semantics of the root of include/exclude filters
Our docs currently don't specify _what_ an include/exclude filter is matched against.

Implementation-wise, we currently expand patterns into full paths.  The full paths are constructed based on our current working directory, which provides unintuitive behavior in certain cases, particularly:

1) when matching a prefix pattern (anything that doesn't start with a `*`)
2) when specifying a source directory that's not the same as the root of the source directory.

For example:

If I'm currently in `/a/b/c` and I run `aws s3 cp /z/y/x s3://bucket --recursive --exclude "*"` then nothing will be excluded this is because we expand the pattern `"*"` to an absolute path based on our CWD and try to match it against the abspath of each file.  So we'd have:

```
/z/y/x/file1  matched against /a/b/c/*   -> does not match
/z/y/x/file2  matched against /a/b/c/*   -> does not match
```

I believe most of these issues can be resolved by running the queries against the root of the source location so in the example above our source location is `/z/y/x` so our full exclude filter would be `/z/y/x/*` If a file is specified, we'd run it against the dirname of the file.

This should be a backwards compatible change.

jamesls:
This fixes #548, which addresses the semantics of the include/exclude filters.

The previous behavior was to expand patterns based on the current working directory, so if I'm in "/a/b/c" and I run an upload command with an exclude filter `cp /x/y/z --exclude "*"` then the exclude filter would be expanded into `/a/b/c/*`, which causes a number of issues.

This fix will expand the include/exclude filters based on the source directory so the example above we'd have an exclude filter of `/x/y/z/*` instead of `/a/b/c/*`.

I've tested this, and updated the unit tests and integration tests.

### DIFF
--- a/tests/unit/customizations/s3/test_filters.py
+++ b/tests/unit/customizations/s3/test_filters.py
@@ -12,9 +12,18 @@
 # language governing permissions and limitations under the License.
 import os
 import unittest
+import platform
 
 from awscli.customizations.s3.fileinfo import FileInfo
 from awscli.customizations.s3.filters import Filter
+
+
+def platform_path(filepath):
+    # Convert posix platforms to windows platforms.
+    if platform.system().lower() == 'windows':
+        filepath = filepath.replace('/', os.sep)
+        filepath = 'C:' + filepath
+    return filepath
 
 
 class FiltersTest(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


vikas027:
@jamesls I think this bug has re-appeared. The below command still tries to upload the directory
```bash
# aws s3 sync ${LOCAL_DIR} s3://${AWS_BUCKET} --exclude "My_Directory"  --delete
```

trinitronx:
still seeing this bug in `awscli` versions:

```
aws-cli/1.7.34 Python/2.6.6 Linux/2.6.32-279.5.2.el6.x86_64
aws-cli/1.8.6 Python/2.6.6 Linux/2.6.32-279.5.2.el6.x86_64
```

trying to exclude things still results in it trying to sync them.

Example command usage:

```
aws s3 sync  /path/to/things s3://bucket.name/path/to/things/    --exclude 'thing1/' --exclude '*thing2.cfg*' --exclude 'thing-3.[0123456]/*' --exclude '*/path/to/things/thing-3.6*' --exclude '*/path/to/things/thing4/*' --dryrun
```

thedukeness:
I'm still seeing this error as well
aws-cli/1.8.9 Python/2.6.6 Linux/2.6.32-573.7.1.el6.x86_64

mattbryson:
Still present in
aws-cli/1.8.13 Python/2.7.10 Darwin/14.5.0

cant --exclude ".DS_Store"

vikas027:
Hi Matt,

Try this --exclude "*.DS_Store*"

Regards,
Vikas

mattbryson:
@vikas027 awesome, worked a treat!

vikas027:
@mattbryson : No worries, I found this a hard way. Basically you need to enclose all files and directories in asterisks. I know that's odd, but that is how I could make it work. Additionally, if you've large number of files, you might want to have s3cmd.

PR #575, opened 2014-01-02
Added option to compare on ETag/MD5

adamsb6:
We have a use case for s3 sync in which we update a staging bucket with production data.  However, the current comparator options will fail to update keys in the staging bucket when they have been updated more recently than the production keys but do not differ in size.

This change adds a --compare-on-etag option that uses the ETag of the key or the md5 of the local file to determine whether or not the source and destination differ.

### DIFF
--- a/awscli/customizations/s3/comparator.py
+++ b/awscli/customizations/s3/comparator.py
@@ -33,6 +33,10 @@ class Comparator(object):
         self.delete = False
         if 'delete' in params:
             self.delete = params['delete']
+        
+        self.compare_on_etag = False
+        if 'compare_on_etag' in params:
+            self.compare_on_etag = params['compare_on_etag']
 
     def call(self, src_files, dest_files):
         """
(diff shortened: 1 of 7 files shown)


rmharrison:
@kyleknap Curious about the status of this PR. Blocked because:
- It hasn't been rebased in a while (just needs to be brought-up to-date)?
- Doesn't support multipart-uploads (@adamsb6 comment about documentation insufficient) 
- Another blocking reason?

simlu:
Would also really benefit from this. Same question as @rmharrison > What would be required to get this merged?

phene:
How has this not been added yet?

jdomingu:
+1. I would find this very helpful and also think that multipart uploads should not block this.

acejam:
Bump

acejam:
Any updates to this?

bes1002t:
any progress?

tim-finnigan:
Thanks for creating this PR and for your patience in waiting on a response. I recently brought this up for discussion as we are cleaning up the backlog. Now that S3 has extended [checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/), the CLI team is more inclined to explore using those for improvements to sync strategy rather than using Etags. Further investigation into this is still needed but unfortunately this PR cannot be considered at this time.

PR #994, opened 2014-11-10
fix for aws s3 sync intermittently setting mtime on a large multipart download to now

kirat-singh:
Hi folks,

I've noticed a lot of unnecessary downloads with aws s3 sync but only for large multi-part files.

Looking over the code I see that CompleteDownloadTask in awscli/customizations/s3/tasks.py sets the file mtime before it is closed.

I can't reliably reproduce it, but off-hand this seems wrong.  I suspect the file has unflushed data, and closing it updates the mtime to now.

So I've made a pretty straightforward fix where IOCloseRequest takes an optional last_updated.  If this is not None, then after the file is closed the IOWriterThread sets the mtime.  And I removed the os.utime call in CompleteDownloadTask since it is now redundant.

In my tests I could not reproduce the problem after the fix was applied.

Please let me know if this is a reasonable change and apologies in advance if I haven't followed protocol, this is the first time I've submitted a pull request here.

best,
-Kirat

### DIFF
--- a/awscli/customizations/s3/executor.py
+++ b/awscli/customizations/s3/executor.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 import logging
 from six.moves import queue
+import os
 import sys
 import threading
 
@@ -175,6 +176,8 @@ class IOWriterThread(threading.Thread):
                 if fileobj is not None:
                     fileobj.close()
                     del self.fd_descriptor_cache[task.filename]
+                    if task.last_modified is not None:
+                        os.utime(task.filename, (task.last_modified, task.last_modified))
 
     def _cleanup(self):
         for fileobj in self.fd_descriptor_cache.values():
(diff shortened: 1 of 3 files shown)


jamesls:
I see what's happening here.  This is definitely a race condition which is why this doesn't always happen.  The mtime is set immediately before enqueueing the IOCloseRequest.  However if the IO thread still has writes queued to that file, then it will continue writing to the file (which updates the mtime) after the original `os.utime` call. This works most of the time because the IO thread writes out IO tasks quickly.  However if you're writing out to a slow disk then this problem becomes much more likely.

The proposed fix here is definitely a better way.  The mtime of the file should be set in the IO thread at file close time rather than at enqueueing time.  In order to merge this though I'll need to add some tests to ensure this issue is fixed.

kirat-singh:
Thanks for merging and adding the test cases.

Issue #604, opened 2014-01-23
aws s3 ls inconsistent file names with/without --recursive


asieira:
When `aws s3 ls` is executed on awscli/1.2.11 without `--recursive`, the resulting files and prefixes are listed as relative to the given path:

```
| => aws s3 ls s3://testbucket/dir1/dir2/dir3/
                           PRE dir4/
2013-09-09 20:33:25    2865280 file1
```

However, when `--recursive` is added to the same path, the resulting files are listed as the entire prefix (full path sans `s3://<bucket name>`):

```
| => aws s3 ls s3://testbucket/dir1/dir2/dir3/ --recursive
2013-09-09 20:33:25    2865280 dir1/dir2/dir3/file1
2013-09-09 20:33:26    5117249 dir1/dir2/dir3/dir4/file2
```

See, if the behavior was consistent and relative paths to the listed prefix were always used, the second output should have been as follows:

```
| => aws s3 ls s3://testbucket/dir1/dir2/dir3/ --recursive
2013-09-09 20:33:25    2865280 file1
2013-09-09 20:33:26    5117249 dir4/file2
```

Perhaps new options `--relative` and `--absolute` could be added to `aws s3 ls` to choose which behavior we want, or at the very least both are made to function in a similar fashion. I vote for relative paths on both cases.

nkadel-skyhook:
Document this in the "aws s3 ls" command, please, or ideally make it optional. It very frustrating to sync, and review, a complex filesystem and have to reparse the output for comparison between local and remote content. And it's more complex because of the undocumented stripping of the bucket name from the output.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

[...]

PR #1009, opened 2014-11-18
Fix ls --recursive discrpency with prefixes

kyleknap:
Fixes https://github.com/aws/aws-cli/issues/604

As an example of what is changed, start off with a bucket ``mybucket`` with the key ``foo/bar``.

When we used to run ``aws s3 ls s3://mybucket/foo/ --recursive``, it would list the object as ``foo/bar``. This is inconsistent with ``aws s3 ls s3://mybucket/foo/`` since it would list the object as ``bar``.

This change makes the output consistent with the non-recusive ``ls``.

cc @jamesls @danielgtaylor


jamesls:
Closing for now.  If anyone has any feedback, please feel free to chime in.

PR #1062, opened 2014-12-15
Fix AssumeRole functionality on windows

jamesls:
On windows, the `:` char is not allowed in the filename, so we replace these with an `_` char.
Also, for POSIX systems, the file is created with `0600` permissions to restrict access to the cache files.

cc @kyleknap @danielgtaylor


onthemove31:
I use S3 Sync but everytime i use it, it error's out with the above error. I'm not able to download the files to my local desktop from s3 since the file name has ":" on it in S3. Can you tell me how to overcome this with your code fix?

PR #1081, opened 2015-01-05
fix minor typo, ACl -> ACL

tedder:
Simple tyop, I even double-checked that I'm not wrong- AWS documentation is consistent about it being ACL, and the L actually stands for something: http://docs.aws.amazon.com/AmazonS3/latest/dev/S3_ACLs_UsingACLs.html


kyleknap:
Thanks for the catch! Merging.

PR #1087, opened 2015-01-12
Fix S3 streaming download corruption

jamesls:
When a download is being piped to stdout `aws s3 cp s3://foo/ -`, it's possible that data can be written out twice.  The issue is that the response body is read in chunks smaller than the range get size.  When a retry occurs the whole request (for the entire range) is retried.  If this retry occurs at the Nth chunk, the chunks 1-N will be enqueued again.

The fix for now is to just enqueue writes at the same size as the chunk size.  I think longer term, there are definitely some optimizations we can do to improve this, but that shouldn't block getting a fix in now.

cc @kyleknap

### DIFF
--- a/awscli/customizations/s3/tasks.py
+++ b/awscli/customizations/s3/tasks.py
@@ -414,7 +414,7 @@ class DownloadPartTask(OrderableTask):
         # for only the remaining parts.  The other alternative, which is what
         # we do here, is to just request the entire chunk size write.
         self._context.wait_for_turn(self._part_number)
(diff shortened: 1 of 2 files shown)


jamesls:
Actually, hold up on the review.  I'm seeing some issues that still need to be fixed.

jamesls:
PR has been updated, @kyleknap feel free to review now.

kyleknap:
Agreed. Looks good. :ship:

PR #1088, opened 2015-01-12
Add integration test for --no-sign-request

kyleknap:
CLI integration test for this botocore PR: https://github.com/boto/botocore/pull/433.

Ensures that we are not signing requests when using ``--no-sign-requests``

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #1103, opened 2015-01-19
Add --humanize and --summarize options for s3 ls

chadhinton:


jamesls:
Thanks for the PR.  I think this would be a great addition to the S3 commands.  Before this is merged, we just need a few things:

1. Can you confirm that you are OK with us accepting this under the Apache 2.0 license?
2. Can you sign our [Contributor License Agreement](https://github.com/aws/aws-cla)?  You can send the signed CLA to the email address in my github profile.

Thanks again for the contribution.

jamesls:
Merged via https://github.com/aws/aws-cli/commit/23b0e69976f19bc8b97a840d5fae1469343a1076

chadhinton:
Great additions. Agree, binary makes more sense than decimal notation for compatibility. Thanks!

PR #1107, opened 2015-01-26
Add small modifications to #1103

jamesls:
This pulls in https://github.com/aws/aws-cli/pull/1103, and makes the following changes:

* Rename `--humanize` to `--human-readable` to be more consistent with other CLI tools that expose this option
* Use 1024 as the base for file sizes.  This is also consistent with what other tools show (ls, du, rsync, etc), but to be "correct" we use IEC suffixes.

cc @kyleknap @danielgtaylor


kyleknap:
Looks good :ship:

PR #1109, opened 2015-01-27
Use correct pathto args for referencing jquery

jamesls:
http://sphinx-doc.org/templating.html#pathto

pathto(document)
Return the path to a Sphinx document as a URL. Use this to refer to built documents.

pathto(file, 1)
Return the path to a file which is a filename relative to the root of the generated output. Use this to refer to static files.

Previously this was generating an href of `jquery-1.9.1.min.js.html` which is not what we want.

cc @kyleknap @danielgtaylor


kyleknap:
LGTM :ship:

PR #1113, opened 2015-01-27
Handle broken pipe errors

jamesls:
If stdout is being piped to another process and that process
closes stdin, then we get an IOError with a broken pipe message.
If the reading end of the pipe closes the pipe, we should just exit.
This is a normal (and common) case and we shouldn't be printing
an error message when this happens.

The error was being raised on the `.flush()` call to the stream.

Before:
```
$ aws ec2 describe-instances | head -n 1
{

[Errno 32] Broken pipe
```

After:

```
$ aws ec2 describe-instances | head -n 1
{
```

cc @kyleknap @danielgtaylor

### DIFF
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -44,6 +44,12 @@ class Formatter(object):
     def _get_default_stream(self):
         return compat.get_stdout_text_writer()
 
+    def _flush_stream(self, stream):
+        try:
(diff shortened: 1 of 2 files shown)


kyleknap:
Looks good. :ship: Even though coveralls shows a drop, you added a test for the case. So you are good.

Issue #1090, opened 2015-01-13
Add ability to limit bandwidth for S3 uploads/downloads


jamesls:
Original from #1078, this is a feature request to add the ability for the `aws s3` commands to limit the amount of bandwidth used for uploads and downloads.

In the referenced issue, it was specifically mentioned that some ISPs charge fees if you go above a specific mbps, so users need the ability to limit bandwidth.

I imagine this is something we'd only need to add to the `aws s3` commands.

austinsnow:
Hello jamesis,
Could you provide a timeframe when the bandwidth limit could become available?
Thanks
austinsnow

kjohnston:
:+1:

beauhoyt:
:+1:

bhegazy:
:+1:

seattledoug:
+1

dsclassen:
![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

godefroi:
![big thumb up](https://camo.githubusercontent.com/a2840914d539225aaf8d395397d8dbc488a0740d/68747470733a2f2f6173736574732d63646e2e6769746875622e636f6d2f696d616765732f69636f6e732f656d6f6a692f756e69636f64652f31663434642e706e67)

rayterrill:
:+1:

kazeburo:
:+1:

isaoshimizu:
:+1:

quiver:
Under Unix-flavor systems, [trickle](http://monkey.org/~marius/pages/?page=trickle) comes in handy for ad-hoc throttling. `trickle` hooks socket-APIs  using `LD_PRELOAD` and throttles bandwidth.

You can run commands something like 
```
$ trickle -s -u {UPLOAD_LIMIT(KB/s)} command
$ trickle -s -u {UPLOAD_LIMIT(KB/s)} -d {DOWNLOAD_LIMIT(KB/s)} command
```
Built-in feature will be really useful, but given cross-platform nature of AWS-CLI, it can cost a lot  to implement and maintain it.

binaryorganic:
Trickle is specifically mentioned in issue #1078 which is linked to in the first comment here. The two (trickle and AWS-CLI) just don't play nice together in my experience.

isp0000:
:+1:

l3rady:
:+1:

andrefelipe:
+1

ddehghan:
+1

[...]

Issue #907, opened 2014-09-03
Limit number of parallel s3 transfers


phemmer:
Can we get a way to limit the number of parallel s3 transfers? As it is, transfer jobs are consuming a lot of system resources (CPU, disk IO, bandwidth) because the `aws s3 sync` command is launching several parallel transfers.

The simplest way I can think of would be to pull constants from environment variables. This would let you override the `MAX_PARTS` from [constants.py](https://github.com/aws/aws-cli/blob/2bdb58ba8ae171fff6467fc9ee39936b5bc1940b/awscli/customizations/s3/constants.py#L17).
With this method, power users could override other constants as well. But for my use case, a `--max-parts` command line option would suffice.

jamesls:
This is now possible via https://github.com/aws/aws-cli/pull/1122, docs are here: https://github.com/aws/aws-cli/blob/develop/awscli/topics/s3-config.rst

Issue #699, opened 2014-03-11
Expose max queue size in S3 as a config/param


jamesls:
This will help users get more meaningful progress regarding their transfers.  See #488 for more info.  Also related: #489

kyleknap:
This has been implemented for a while. Closing.

PR #1122, opened 2015-02-03
Add config variables for changing S3 runtime values

jamesls:
This pull requests adds support for the following parameters in the `~/.aws/config` file:

* max_concurrent_requests: The maximum number of concurrent transfer requests to S3 at any given moment.
* max_queue_size: The maximum number of transfer tasks queued at any given point in time.
* multipart_threshold: The size at which we switch to multipart uploads/downlaods. The appropriate values are specified below.
* multipart_chunk_size: The chunk size to use when using multipart transfers.

A sample config looks like:

```
[profile development]
aws_access_key_id = foo
aws_secret_access_key = bar
region = us-west-2
s3 =
  max_concurrent_requests = 20
  max_queue_size = 10000
  multipart_threshold = 128MB
  multipart_chunk_size = 64MB
```

This addresses several needs we've heard including:

* The ability to limit the number of parallel s3 transfers (https://github.com/aws/aws-cli/issues/907, and indirectly https://github.com/aws/aws-cli/issues/1090)
* The ability to set the queue size for better progress info:  https://github.com/aws/aws-cli/issues/699
* Add the ability to set the threshold at which we switch to multipart operations
* Set the part size to use for multipart operations

This still needs documentation, but the code is ready for review.

## Code Overview

While working on this code, there were parts of the code I updated before making any functional changes.  These include things like pep8 formatting, removing code that's no longer called, improving some of the semantics of the `CommandParameters` class.  None of these changes have any functional changes:

fb51db1 - Cleanup code formatting
60f7377 - Remove unused method
ab7ba1a - Remove session dep from CommandParameters
e9fdaf7 - Remove trailing/leading whitespace
2e3477d - Move part size constants out to their respective modules
9ba4d38 - Move constants to transferconfig

Then the two commits that add in the change.  The first is the addition of a `RuntimeConfig` object that manages all of the merging/converting of config file values:

697662b - Add class for merging converting s3 runtime config

Then plumbing this newly created class into the S3 command codebase:

db24830 - Plumb through S3 runtime config

I'd suggest taking a look at the last two commits on their own and then reviewing the rest.

cc @kyleknap @danielgtaylor


jamesls:
Looking at the coveralls coverage drop...

jamesls:
Also, just noticed that there's an integ test failure for streaming.  Looks like I missed plumbing this into streaming s3handler as well, which I believe should not use the runtime config settings.

awscli/customizations/s3/s3handler.py
```diff
@@ -65,14 +68,16 @@ def __init__(self, session, params, result_queue=None,
...
                 self.params[key] = params[key]
-        self.multi_threshold = multi_threshold
-        self.chunksize = chunksize
+        self.multi_threshold = self._runtime_config['multipart_threshold']
+        self.chunksize = self._runtime_config['multipart_chunksize']
+        LOGGER.debug("Using a multipart threshold of %s and a part size of %s",
+                     self.multi_threshold, self.chunksize)
         self.executor = Executor(
-            num_threads=self.EXECUTOR_NUM_THREADS,
+            num_threads=self._runtime_config['max_concurrent_requests'],
```
kyleknap: Shouldn't this be
```
num_threads=self._runtime_config['max_concurrent_requests']-1
```

Because the main thread is acting as a producer by calling the ``ListObjects`` operation.

awscli/customizations/s3/transferconfig.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def build_config(self, **kwargs):
+        """Create and convert a runtime config dictionary.
+
+        This method will merge and convert S3 runtime configuration
+        data into a single dictionary that can then be passed to classes
+        that use this runtime config.
+
+        :param kwargs:  Any key in the ``DEFAULTS`` dict.
+        :return: A dictionar of the merged and converted values.
```
kyleknap: dictionar --> dictionary

awscli/customizations/s3/transferconfig.py
```diff
@@ -0,0 +1,76 @@
...
+            if value is not None and not isinstance(value, int):
+                runtime_config[attr] = human_readable_to_bytes(value)
+
+    def _validate_config(self, runtime_config):
+        for attr in self.POSITIVE_INTEGERS:
+            value = runtime_config.get(attr)
+            if value is not None:
+                try:
+                    runtime_config[attr] = int(value)
+                    if not runtime_config[attr] > 0:
```
kyleknap: If num_threads in the executor = max_concurrent_requests - 1, max_concurrent_requests must not be less than 2.

awscli/customizations/s3/utils.py
```diff
@@ -61,12 +71,34 @@ def human_readable_size(value):
...
+
+    """
+    suffix = value[-2:].lower()
+    has_size_identifier = (
+        len(value) >= 2 and suffix in SIZE_SUFFIX)
+    if not has_size_identifier:
+        try:
+            return int(value)
+        except ValueError:
+            raise ValueError("Invalid size value: %s" % value)
```
kyleknap: You might want to state what a valid size value qualifies as in the error message. I could see someone see Mib being used for ``ls --human-readable`` and try using that instead.

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -531,22 +529,13 @@ def test_check_src_path_pass(self):
         parameters = {}
         for filename in files:
             parameters['dir_op'] = filename[1]
-            cmd_parameter = CommandParameters(self.session, 'put',
-                                              parameters, '')
+            cmd_parameter = CommandParameters('put', parameters, '')
             cmd_parameter.add_region(mock.Mock())
             cmd_parameter.check_src_path(filename[0])
 
-    def test_check_force(self):
```
kyleknap: Hmm we were not actually asserting anything for this test...

kyleknap: It could be nice if this test was moved to the CommandArchitecture test suite and have something asserted.

kyleknap:
Looks good. I guess my main comment was related to the interpretation of ``max_concurrent_requests``. I just want to be on the same page with you before it gets merged.

awscli/customizations/s3/s3handler.py
jamesls: After initially implementing that, I ended up making this just match the number of threads.  The main reason is that if we do add ` max - 1`, then the minimum value needs to be `2`, which I think would confuse customers.  It's also not intuitive that you can't set this value to 1.  If a user wants to "turn off" parallelism, you'd think you'd set `max_concurrent_requests` to 1, but they would need to set max_concurrent_requests to `2`.  I could see us getting questions about that.

Another option is to rename this to be more explicit.  If we called it `max_transfer_requests` maybe that would clear up what this is doing.  Although I think `max_concurrent_requests` is more clear than `max_transfer_threads`.  What do you think?

awscli/customizations/s3/utils.py
jamesls: maybe we should also just support the *iB suffixes as well?

tests/unit/customizations/s3/test_subcommands.py
jamesls: Yeah I think that's probably the drop in coverage.  I'll add a test.

awscli/customizations/s3/s3handler.py
kyleknap: Yeah I like that last option the best. If I had no knowledge of the internals of the s3 commands, I would also find it confusing that the lowest that you can make it is 2. I do not like ``max_transfer_requests`` because it is ambiguous (it seems to imply that we will at most do that maximum amount of transfers for the one command).

awscli/customizations/s3/utils.py
kyleknap: +1 on that. I could see an issue come in on that.

jamesls:
@kyleknap I believe I've incorporated all the feedback.  Coveralls also looks good now.  Let me know if I've missed anything.

cc @danielgtaylor

kyleknap:
Thanks! :ship:

jamesls:
Closing, this was merged via: https://github.com/aws/aws-cli/commit/9911da327b39fbbff66e454fd2d765e20fc6a659

joelvanvelden:
The given sample config has an error:
"multipart_chunk_size = 64MB"
should instead read:
multipart_chunksize = 64MB

I found this out the hard way :(

Also, this (very useful) new feature is not documented anywhere apart from this pull request.

jamesls:
@joelvanvelden Thanks for pointing that out.  I wish I would have named it multipart_chunk_size to be consistent with max_queue_size but I suppose it's too late to change.

Regarding the documentation, there's an existing pull request out (https://github.com/aws/aws-cli/pull/1147) that adds topic guides to the CLI help system (as well as the online html docs).  This gives us a place to document advanced topics such as these s3 settings.  We'll be adding a topic guide on s3 configuration shortly.

TimLudwinski:
Unfortunately, setting `max_concurrent_requests` doesn't solve the issue of limiting the bandwidth for me.  With 1 connection (I verified that there was only one though netstat), our 100Mb pipe is clogged.  (Not that having this bandwidth is a bad thing necessarily).  We still need #1090 to be implemented.

PR #1123, opened 2015-02-03
Remove some references in iam examples

kyleknap:
These references caused the links to break in the command's reference page because the references are declared twice in the same document. Removed the reference in the examples to fix the broken links.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

kyleknap:
Yeah... coveralls does not make any sense. Merging.

PR #1126, opened 2015-02-04
Disable paramfile for sqs purge-queue

jamesls:
Valid values for this param will be https so we need to disable paramfile processing.

cc @kyleknap @danielgtaylor


jamesls:
Coveralls makes no sense here.  I think I should be ok test coverage, I added a test for this.

kyleknap:
Yeah its fine. :ship:

PR #1129, opened 2015-02-05
Fix failing test from recent Botocore changes

danielgtaylor:
This makes the CLI work with the changes in boto/botocore#448 that support
anonymous (unsigned) clients using the new `botocore.UNSIGNED` sentinel.

It requires the upstream Botocore change, so tests are likely to fail in this pull request.

cc @jamesls @kyleknap

### DIFF
--- a/awscli/customizations/globalargs.py
+++ b/awscli/customizations/globalargs.py
@@ -13,6 +13,7 @@
 import sys
 import os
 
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:

kyleknap:
Same :ship:

PR #1132, opened 2015-02-06
Tag parser

kyleknap:
This PR adds the class ``TopicTagDB`` to the CLI. It is a database like object will be used to help generate a related items sections from available topics in the CLI codebase across other topics and reference commands.

To add a topic to the CLI, the workflow is as follows:
   1) Add a topic to the awscli/topics directory with the appropriate tags.
   2) Run the make-topic-index script to generate a new index that can be loaded.

This reason an index is needed is to avoid having to scan through each topic file (from disk) each time you create a new ``TopicTagDB`` instance as this could get pretty time intensive if there were a lot of topics. 

The structure of the database can be viewed as a python dictionary:
```
    {'topic-name-1': {
        ':title:': ['My First Topic Title'],
        ':description:': ['This describes my first topic'],
        ':category:': ['General Topics', 'S3'],
        ':related command:': ['aws s3'],
        ':related topic:': ['topic-name-2']
     },
     'topic-name-2': { .....
    }
```
The keys of the dictionary are the CLI command names of the topics. These names are based off the name of the reStructed text file that corresponds to the topic. The value of these keys are dictionaries of tags, where the tags are keys and their value is a list of values for that tag. Note that all tag values for a specific tag of a specific topic are unique (no duplicates). All of the tags listed in the example is what is currently supported by the class.

The ``TopicTagDB`` class has the following public methods:
* ``load_json_index()``: Loads a JSON index into memory.
* ``save_json_index()``: Save the data in memory into a JSON file on disk.
* ``get_all_topic_names()``: Get the names of all topic names loaded into memory.
   Topic names are the name of their rst file
* ``get_all_topic_src_files()``: Get the file paths for all of the topics in the awscli/topics directory
* ``scan()``: Scan through all of the files provided and save the topic and tag into memory
* ``query()``: Get all topics related to a specific tag and/or tag value
* ``get_tag_value()``: Get the value of a specified tag and topic.

cc @jamesls @danielgtaylor


kyleknap:
Btw, the previous comment by coveralls sent me an email saying that the coverage dropped by -0.06%, when the comment says it went up by 0.08%. I added one file and it has 100% coverage...

tests/unit/test_topictags.py
```diff
@@ -0,0 +1,538 @@
...
+    def setUp(self):
+        self.topic_tag_db = TopicTagDB()
+        self.file_creator = FileCreator()
+
+        # In python 3, order matters when calling assertEqual to
+        # compare lists and dictionaries with lists. Therefore,
+        # assertItemsEqual needs to be used but it is renamed to
+        # assertCountEqual in python 3.
+        if six.PY3:
+            self.assertItemsEqual = self.assertCountEqual
```
kyleknap: Any ideas on a better way to handle this?

jamesls:
Before I get too far into this, my initial impression is that we should be leaning on docutils where possible to do the rst parsing.  I don't think we really get much from reimplementing this ourselves do we?

```
In [1]: contents = open('/tmp/stuff.rst').read()

In [2]: print contents
Topic Foo
=========

:category: foo
:description: foo
:title: bar

This is more stuff.


In [3]: import docutils.core

In [4]: parsed = docutils.core.publish_doctree(contents)

In [5]: print parsed.pformat()
<document ids="topic-foo" names="topic\ foo" source="<string>" title="Topic Foo">
    <title>
        Topic Foo
    <docinfo>
        <field>
            <field_name>
                category
            <field_body>
                <paragraph>
                    foo
        <field>
            <field_name>
                description
            <field_body>
                <paragraph>
                    foo
        <field>
            <field_name>
                title
            <field_body>
                <paragraph>
                    bar
    <paragraph>
        This is more stuff.
```

kyleknap:
Ok I switched out the parsing to rely on docutils as opposed to string parsing. Ready to be looked at.

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+
+    Tag                 Meaning                         Required?
+    ---                 -------                         ---------
+    :title:             The title of the topic          Yes
+    :description:       Sentence description of topic   Yes
+    :category:          Category topic falls under      Yes
+    :related topic:     A related topic                 No
+    :related command:   A related command               No
+
+    To see examples of how to specify tags looks in the directory
```
danielgtaylor: grammar: look

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+    awscli/topics. Note that tags can have multiple values by delimiting
+    values with commas. All tags must be on their own line in the file.
+
+    This class can load a JSON index represeting all of topics and their tags,
```
danielgtaylor: grammar: all ~~of~~ topics

scripts/make-topic-index
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+"""This is a script to generate s JSON index for the CLI Topic Guide
```
danielgtaylor: grammar: `s` should be `a`

tests/unit/test_topictags.py
danielgtaylor: I think this is probably fine. The comment explains it nicely. My only comment is that you might use the Python 3 way of doing things as the default, and fix Python 2.

danielgtaylor:
LGTM, but I'd say let @jamesls take another look before merging.

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+    scan all of the topics and store the values of their tags, retrieve the
+    tag value for a particular topic, query for all the topics with a specific
+    tag and/or value, and save the loaded data back out to a JSON index.
+
+    The structure of the database can be viewed as a python dictionary:
+
+    {'topic-name-1': {
+        'title': ['My First Topic Title'],
```
jamesls: I get why this is parsed as a list, but the semantics of a title and description being a list is confusing.  What does it mean if the title has multiple elements in the list?  Same thing for description.

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        index_filepath = self.JSON_INDEX
+        if index_file is not None:
+            index_filepath = index_file
+        with open(index_filepath, 'w') as f:
+            f.write(json.dumps(self._tag_dictionary, indent=4, sort_keys=True))
+
+    def get_all_topic_names(self):
+        """Retrieves all of the topic names of the loaded JSON index"""
+        return list(self._tag_dictionary.keys())
```
jamesls: `.keys()` is unnecessary.

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # The default JSON index to load.
+    JSON_INDEX = os.path.join(TOPIC_DIR, 'topic-tags.json')
+
+    def __init__(self, tag_dictionary=None):
+        self._tag_dictionary = tag_dictionary
+        if self._tag_dictionary is None:
+            self._tag_dictionary = {}
+
+    def load_json_index(self, index_file=None):
```
jamesls: is there a reason this can't just default to `JSON_INDEX`?

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+        :param index_file: The path to a specific JSON index to load.
+            If nothing is specified it will default to the default JSON
+            index at ``JSON_INDEX``.
+        """
+        index_filepath = self.JSON_INDEX
+        if index_file is not None:
+            index_filepath = index_file
+        with open(index_filepath, 'r') as f:
+            index = f.read()
+            self._tag_dictionary = json.loads(index)
```
jamesls: Not a huge deal, but this is also the same thing as:


```
with open(...) as f:
    self._tag_dictionary = json.load(f)
```

awscli/topictags.py
```diff
@@ -0,0 +1,254 @@
...
+
+        :param topic_files: A list of paths to topics to scan into memory.
+        """
+        for topic_file in topic_files:
+            with open(topic_file, 'r') as f:
+                # Parse out the name of the topic
+                topic_name = self._find_topic_name(topic_file)
+                # Add the topic to the dictionary if it does not exist
+                self._add_topic_name_to_dict(topic_name)
+                # Read the file
```
jamesls: FWIW, this comment seems unnecessary, it's pretty clear that `f.read()` reads the contents of a file.

tests/unit/test_topictags.py
jamesls: It doesn't actually seem like assertCountEqual is necessary.  The lists you're comparing don't contain dupes do they?  You can either just have a custom assertion method that just `set()`'s the values before `assertEqual` or `sorted()'d` the data before calling assertEqual.

Either way I'm not a fan of these version conditionals throughout the code and prefer to have them all collected to a single location wherever possible.  If you'd prefer to stick with assertCountEqual, I'd suggest moving these differences into `awscli.testutils` which is where all the various python version differences related to testing are encapsulated.

tests/unit/test_topictags.py
```diff
@@ -0,0 +1,538 @@
...
+        self.topic_tag_db = TopicTagDB(tag_dict)
+        self.assertItemsEqual(self.topic_tag_db.get_all_topic_names(),
+                              reference_topic_list)
+
+    def test_get_all_topic_source_files(self):
+        source_files = []
+        for i in range(5):
+            topic_name = 'topic-name-' + str(i)
+            source_files.append(self.file_creator.create_file(topic_name, ''))
+        topic_dir_patch = 'awscli.topictags.TopicTagDB.TOPIC_DIR'
```
jamesls: Why not let the `TopicTagDB` take a `topic_dir` arg?  This will avoid patching.

tests/unit/test_topictags.py
kyleknap: I really wanted to use ``set()`` or ``sorted()`` but I was not able to get away with it for some of the tests that test the equality of a nested dictionary i.e.
```
dictionary = {
    'foo': ['bar', 'baz']
...
}
```
I like the idea of using ``assertCountEqual`` and moving that to ``awscli.testutils``

awscli/topictags.py
kyleknap: Good point

awscli/topictags.py
kyleknap: True, I think that is remnants from when I was debugging.

awscli/topictags.py
kyleknap: The only reason I kept it as a list is that it simplifies the code since you can always expect a list, and I would not have to do any non-list handling. Having multiple elements in the tile or description list would be looked at as a bug or issue in the source file. Whenever, I try to grab the description or title (from outside the class), I will always try to grab the first element in that list, which should be the only element.

I agree the semantics is a bit weird. I just choose it to keep it that way to help simplify/shorten the codebase. @jamesls Let me know what you think.

tests/unit/test_topictags.py
kyleknap: Good point.

awscli/topictags.py
jamesls: If multiple elements in the title/description is considered a bug, is there going to be some code that will raise an error when this happens?

kyleknap:
@jamesls Response to this comment:
```
Should we be ignoring unknown tags? It either seems to me that the job of this class is to parse out 
fields and populate a JSON document without caring about the name of the fields, or it should validate 
that you've only specified the allowed fields and error out if it encounters an unknown field.
```
Out of both those options, I would prefer validation as it is very explicit if you formatted the source file incorrectly. I do not think I would like if we just add any tag to the dictionary because I would like it if my tag mistakes were brought up to me prior to actually generating the html/man page.

One thing I was afraid of was that the docutils parser may pick up a ``:field:`` directives that could exist in the body of the topic. However, I have not checked this yet. So that was one reason I just ignored it.

Without adding the validation ``TopicTagDB`` and the current direction of future PR's, the only place you can tell that you had an issue with a tag is when you render the topic to screen (the tag will show up on in its contents)

I am leaning toward validation at ``TopicTagDB`` now. Let me know what you think.

awscli/topictags.py
kyleknap: Yeah I was going to add something like that into the next pull request. I was going to put the validation at the command level. But I now that I think about it. @jamesls How would you feel if I added a ``get_tag_title()`` and ``get_tag_description()`` method to ``TopicTagDB`` that will raise an error if there are more than one values?

kyleknap: Actually I would make it generic with ``get_tag_single_value()`` method.

jamesls: Sounds good.

jamesls:
Yeah, I think that sounds fine to put it in the TopicTagDB class, especially because it has an explicit list of fields it's looking for.  I would also agree that I'd like to know about tag mistakes before generating the html/man page.

kyleknap:
Coverage look fine despite coveralls comment.

@jamesls @danielgtaylor 
I made changes as you guys suggested and we discussed. Ready to get looked at again.

awscli/topictags.py
kyleknap: I actually cannot default to ``JSON_INDEX``. There would be no way for me to overwrite the default if I was to subclass from the class. For example, if I do:
```
import awscli
from awscli.topictags import TopicTagDB

class Foo(TopicTagDB):

    JSON_INDEX = 'foo'

foo = Foo()

foo.load_json_index()
```

``load_json_index`` will use the original default as opposed to the expected new default for ``JSON_INDEX``.

kyleknap:
@jamesls @danielgtaylor I rebased off develop recently and force pushed. The commits with the changes is everything from "Made a few grammar fixes" https://github.com/kyleknap/aws-cli/commit/d6124fe795fae542576d6da3a4b3bc71f11796e5 to "Switch some defaults to not use globals" https://github.com/kyleknap/aws-cli/commit/6b741209bb653d98c4dcc2d5bf7be2a633f4558f

As to the last commit, I realized that I could not use a global as default based on future work I was doing. See the comment I made for the reasoning.

awscli/topictags.py
jamesls: Why do we need that feature though?  You can just pass the index_file name into the load_json_index right?

Otherwise, do we need to support both interfaces?  Can we remove the `index_file` param?  Just looking to avoid adding things we don't actually need.

kyleknap: @jamesls 
We need to support both interfaces, for these reason:

1) If you want to avoid patching the global argument ``JSON_INDEX``, we need to be able to pass it directly as an argument to ``load_json_index()``. The only other option is to make them arguments that you pass to the constructor.

2) This is more related to your first two questions: This is a problem that is run into later down the road. When you instantiate the ``TopicTableDB`` and the load the json, the ``HelpCommand`` (which is the current objects in my design that instantiate and load the ``TopicTableDB``) objects do not and should not need to know where the json index is.  Therefore, it will be making the ``load_json_index()`` call without that ``index_file`` argument. If I want to do any unit testing on the ``HelpCommand`` I need to be able to redirect the ``JSON_INDEX`` and ``TOPIC_DIR`` to a different value (like one that points to a temp file). However, if I make ``JSON_INDEX`` the default I can never patch that value out in the function since it is the default to the method.

The only work around I can think of is to pass these arguments to the constructor of the class. Thus, removing the need of them in some of the other methods.

jamesls: I wouldn't view the constructor of the class approach a work around.  It seems perfectly valid to me that the thing that creates a TopicTableDB should know where the index file is located.  The TopicTableDB should just be a generic object that can return restructured text with fields/values into a JSON index.  The thing that creates it should tell it where it look.

kyleknap: Cool I will do that then.

kyleknap:
Ok I moved the ``index_file`` and ``topic_dir`` parameters out of the class's methods to its constructor.

jamesls:
I hate to keep pushing on this point, but the whole discussion centered around just defaulting to JSON_INDEX instead of `None`.  Wasn't the whole point of moving these values to `__init__` so that we didn't have to default to `None`?

kyleknap:
@jamesls 
The reason that I had set it to ``None`` was the same reason as point 2 in this in this comment: https://github.com/aws/aws-cli/pull/1132#discussion_r24545155

Now since we have it in the constructor, I made it an attribute with getters and setters so I can avoid patching the global values. And I can either patch or set those attributes when I need to, if I constructed the ``TopicTagDB`` using the default values.

kyleknap:
Fixed the accidental move of globals in my latest commit.

jamesls:
:shipit:  Thanks for making all the edits.

kyleknap:
No problem. Merging.

Issue #1130, opened 2015-02-05
Unable to set website-redirect with `http://` or `https://` prefix


csexton:
Passing in a URL into the `--website-redirect` option is considered invalid, despite including a valid prefix.

```
% aws s3 cp --website-redirect 'http://otherserver.com' /var/folders/mr/hp5q732j74b25ht_d_2tn8bh0000gn/T/redirect.XXXXXX.2RcCGLGT s3://mys3bucket.server.com/path/to/index.html
upload failed: ../../../../var/folders/mr/hp5q732j74b25ht_d_2tn8bh0000gn/T/redirect.XXXXXX.2RcCGLGT to s3://mys3bucket.server.com/path/to/index.html 
A client error (InvalidRedirectLocation) occurred when calling the PutObject operation: 
  The website redirect location must have a prefix of 'http://' or 'https://' or '/'.
```
I have tried with both http and https (however if I use https on a site that doesn't have a valid cert I get a ssl error). Also tried a variety of escaping and quoting the url including, none of which worked :cry: 

- `aws s3 cp --website-redirect 'http://otherserver.com' file s3://mys3bucket/path/file`
- `aws s3 cp --website-redirect "http://otherserver.com" file s3://mys3bucket/path/file`
- `aws s3 cp --website-redirect http://otherserver.com file s3://mys3bucket/path/file`
- `aws s3 cp --website-redirect http:\/\/otherserver.com file s3://mys3bucket/path/file`
- `aws s3 cp --website-redirect https://otherserver.com file s3://mys3bucket/path/file`
- `aws s3 cp file s3://mys3bucket/path/file --website-redirect "http://otherserver.com" `

I was able to roll back to version 1.4 and get the command to work again.

kyleknap:
I can reproduce the issue as well. It makes sense that it works again in 1.4 because that was around the time we made the switch on some of the internals of the s3 commands.

The ``--website-redirect`` argument is trying to using the contents at that web address since it is a feature in the CLI that if you prepend an argument values with ``http://`` or ``https://``, it will try to retrieve the contents of the web address and use that as the value to the argument.

The solution is that we must black list that feature from this argument. Working on a fix for it.

PR #1137, opened 2015-02-09
Ignore paramfile for s3 --website-redirect

kyleknap:
Fixes https://github.com/aws/aws-cli/issues/1130

Since we switched the s3 commands to the ``BasicCommand`` we picked up the paramfile feature (i.e. ``file://``, ``http://``, etc). Thus we need to blacklist ``--website-redirect`` for a couple of the s3 commands.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -37,6 +37,10 @@ PARAMFILE_DISABLED = set([
     'cloudformation.update-stack.stack-policy-url',
     'cloudformation.set-stack-policy.stack-policy-url',
 
+    'custom.cp.website-redirect',
+    'custom.mv.website-redirect',
+    'custom.sync.website-redirect',
+
     'sqs.add-permission.queue-url',
     'sqs.change-message-visibility.queue-url',
     'sqs.change-message-visibility-batch.queue-url',
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -37,6 +37,8 @@ PARAMFILE_DISABLED = set([
     'cloudformation.update-stack.stack-policy-url',
     'cloudformation.set-stack-policy.stack-policy-url',
 
+    # We will want to change the event name to ``s3`` as opposed to
+    # custom in the near future along with ``s3`` to ``s3api``.
     'custom.cp.website-redirect',
     'custom.mv.website-redirect',
     'custom.sync.website-redirect',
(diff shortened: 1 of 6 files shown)


awscli/paramfile.py
```diff
@@ -37,6 +37,10 @@
     'cloudformation.update-stack.stack-policy-url',
     'cloudformation.set-stack-policy.stack-policy-url',
 
+    'custom.cp.website-redirect',
```
kyleknap: On a side note, I do not like that all custom commands have ``custom`` as its service in the event that is emitted. It is not good if there are ever custom commands that overlap in the cli namespace even though they may have a different command lineage.

However, I do not think any change can be made till after the command lineage PR is merged.

jamesls: Yeah, let's get the command lineage PR merged so we can update this PR to avoid `custom`.

kyleknap: Cool. This change may be a little larger than expected (a change in the clidriver and custom commands). If you look further down in the list of black listed arguments, the ``s3api`` commands that use ``--website-redirect`` have the event listed as ``s3``...

kyleknap: Correction: the change for the data driven commands need to come in ``arguments.py``

kyleknap: Sorry for all the spamming. Actually a change will need to take place at ``unpack_argument`` in ``argprocess.py``: https://github.com/aws/aws-cli/blob/develop/awscli/argprocess.py#L63

The reason being that the service_name/operation_name input limits all commands to the strict two command length structure.

For example, for the custom top layer ``s3`` command, a ``custom`` had to be added to fit the input because both ``service_name`` and ``operation_name`` are required. This is an issue because if I wanted to apply a handler to the event ``custom.s3.*.some-argument`` I would not be able to register to all of ``s3``'s operation commands because ``s3``'s operation commands must emit ``s3.operation.some-argument`` or ``custom.operation.some-argument`` to fit the inputs to that function.

This is also an issue for commands that have more than two commands like waiters.

I propose to leave the PR as is and make this change in a subsequent PR because it is outside the scope of the current bug fix as I will have to change the clidriver and some handlers as well.

@jamesls Let me know what you think.

jamesls: My preference is to just put in the long term fix rather than a temporary fix that will be removed once the changes are plumbed through clidriver/handlers.

However, I'm ok with putting this fix in for the time being.  Code and test themselves look good.  Feel free to merge if you want.

kyleknap: I added a reminder to fix it in my latest commit.

PR #1144, opened 2015-02-12
Switch over unit tests to assert_params_for_cmd2

jamesls:
This is a branch I've been working on every now and then which updates all tests such that nothing uses assert_params_for_cmd.  Instead anything that needs that type of assertion functionality instead uses assert_params_for_cmd2.  As a reminder:

* assert_params_for_cmd - This verified the serialized params sent to botocore's make_request.  This meant that we were also testing the serialization format of the particular protocol.
* assert_params_for_cmd2 - This tests at the `Operation.call` layer, i.e before it's serialized into the wire format.

### Why Switch?

This came about with the switch over to the normalized models, but the main reasons assert_params_for_cmd is going away is because:

* It shouldn't be the AWS CLI's concern to verify the wire protocol from a command.  We just need to ensure we're calling into botocore correctly and let botocore's test handle the verification of the wire protocol serialization.
* The assert_params_for_cmd could really only verify query services because anything that was rest-xml or json would need to be the request body as a string.  So some of the tests for s3 were verifying the built XML body.  These tests typically over-asserted their expectations such that changing the XML serialization slightly, despite being valid, would cause unnecessary unit test failures.

### Migration Plan

* Get this PR merged.
* Now that nothing depends on assert_params_for_cmd, we can remove the method entirely and rename assert_params_for_cmd2 back to assert_params_for_cmd
* Change to CamelCasing in the customizations where necessary.

As to that third point, some of the customizations (notably EC2) were written when `Operation.call` only accepted `snake_case` names.  Now that `CamelCasing` is supported, and in fact, is the only casing supported in botocore's clients, we need to update these customizations to be CamelCase.  This can wait until the switchover to clients, but I'd rather tackle this now to minimize the amount of work needed for the botocore client switchover.

cc @kyleknap @danielgtaylor

### DIFF
--- a/tests/unit/ec2/test_run_instances.py
+++ b/tests/unit/ec2/test_run_instances.py
@@ -26,20 +26,20 @@ class TestDescribeInstances(BaseAWSCommandParamsTest):
         args_list = (self.prefix + args).split()
         result = {
             'ImageId': 'ami-foobar',
-            'MaxCount': 1,
-            'MinCount': 1
+            'max_count': 1,
+            'min_count': 1
         }
-        self.assert_params_for_cmd(args_list, result)
+        self.assert_params_for_cmd2(args_list, result)
 
     def test_count_scalar(self):
         args = ' --image-id ami-foobar --count 2'
         args_list = (self.prefix + args).split()
         result = {
             'ImageId': 'ami-foobar',
-            'MaxCount': 2,
-            'MinCount': 2
+            'max_count': 2,
+            'min_count': 2
         }
-        self.assert_params_for_cmd(args_list, result)
+        self.assert_params_for_cmd2(args_list, result)
 
     def test_user_data(self):
         return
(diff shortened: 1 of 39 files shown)


tests/unit/ec2/test_run_instances.py
```diff
@@ -79,13 +79,15 @@ def test_block_device_mapping(self):
...
-            'BlockDeviceMapping.1.DeviceName': '/dev/sda1',
-            'BlockDeviceMapping.1.Ebs.VolumeSize': 20,
+            'BlockDeviceMappings': [
+                {'DeviceName': '/dev/sda1',
+                 'Ebs': {'VolumeSize': 20}},
+            ],
             'ImageId': 'ami-foobar',
-            'MaxCount': 1,
-            'MinCount': 1
+            'max_count': 1,
```
kyleknap: That's a weird format (consistency-wise). Botocore needs that argument as snake_casing?

jamesls: It accepts both in the Operation.call interface.  See the last paragraph in the PR description for more info.  The plan is to migrate these over eventually.

kyleknap: Cool

kyleknap:
Yeah looks good. :ship: Out of curiosity, how many more of these do we need to change?

jamesls:
You mean for switching over to assert_params_for_cmd2?  This should be all of them.

kyleknap:
Yeah, that is what I meant. And cool.

PR #1148, opened 2015-02-13
Add ``lineage_names`` to commands

kyleknap:
It is a convenience attribute to get the names of all of the commands in a lineage without having to iterate through the lineage each time. This saves us from a lot of duplicate code.

So for a ``lineage`` attribute, you get the elements in the list in terms of command objects. For the ``lineage_names`` attribute, you get the elements in the list in terms of the names of command objects. This attribute I could see really useful for at least two scenarios:

1) Generating event names. All you would have to do to get the command's representation in the event name is ``'.'.join(cmd.lineage_names)``

2) Looking for related command tags in the ``TagTopicDB``. All of the ``related command`` tag values follow the format of ``<service> [operation] ...``. To look for these values, all you would have to do is 
``' '.join(cmd.lineage_names)``. 

I also considered making it a utils function, but it would be pretty annoying to have to import it each time you need to use it. Let me know what you think.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -276,6 +276,14 @@ class CLICommand(object):
         # a list.
         return [self]
 
+    @property
+    def lineage_names(self):
+        # Represents the lineage of a command in terms of command ``name``
+        lineage_names = []
+        for cmd in self.lineage:
+            lineage_names.append(cmd.name)
+        return lineage_names
+
     def __call__(self, args, parsed_globals):
         """Invoke CLI operation.
 
@@ -473,6 +481,14 @@ class ServiceOperation(object):
     @lineage.setter
     def lineage(self, value):
         self._lineage = value
+
+    @property
+    def lineage_names(self):
+        # Represents the lineage of a command in terms of command ``name``
+        lineage_names = []
+        for cmd in self.lineage:
+            lineage_names.append(cmd.name)
+        return lineage_names
 
     @property
     def arg_table(self):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -279,10 +279,7 @@ class CLICommand(object):
     @property
     def lineage_names(self):
         # Represents the lineage of a command in terms of command ``name``
-        lineage_names = []
-        for cmd in self.lineage:
-            lineage_names.append(cmd.name)
-        return lineage_names
+        return [cmd.name for cmd in self.lineage]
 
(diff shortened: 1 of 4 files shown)


kyleknap:
Coverage is fine. All of the code that I added got covered.

awscli/clidriver.py
```diff
@@ -276,6 +276,14 @@ def lineage(self):
         # a list.
         return [self]
 
+    @property
+    def lineage_names(self):
+        # Represents the lineage of a command in terms of command ``name``
+        lineage_names = []
+        for cmd in self.lineage:
+            lineage_names.append(cmd.name)
+        return lineage_names
```
danielgtaylor: This is a good candidate for a list comprehension:

```python
@property
def lineage_names(self):
    # Represents the lineage of a command in terms of command ``name``
    return [cmd.name for cmd in self.lineage]
```

danielgtaylor:
This does look like it could be useful, but it would be nice to see how it simplifies the code before getting merged in. Right now it's just adding something that isn't yet used.

awscli/clidriver.py
kyleknap: Good point.

kyleknap:
Yeah I will be using this in future pull requests. I will note it when I do send the pull request. This just seemed something as a good candidate for a separate PR from what I am working on.

tests/unit/test_clidriver.py
```diff
@@ -773,6 +781,9 @@ def test_lineage(self):
         self.cmd.lineage = ['foo']
         self.assertEqual(self.cmd.lineage, ['foo'])
 
+    def test_lineage_names(self):
```
jamesls: This isn't technically wrong, but it's a little confusing given the test above has `self.assertEqual(self.cmd.lineage, ['foo'])`.  Do you think it's worth changing the `test_lineage` test to set the lineage to an object to be consistent with real world usage?

jamesls:
Given the link to #1151, I think this seems reasonable.  Just had one small question about the tests but otherwise, :shipit:

tests/unit/test_clidriver.py
kyleknap: Yeah I can update the test.

kyleknap:
Coveralls is not correct again. I switched those test to ``CLICommand`` objects as opposed to strings. Merging.

PR #1151, opened 2015-02-13
Rename doc event

kyleknap:
This PR switches the ``event_class`` for the the various operation objects. A big problem with the current ``event_class`` attributes was that the values varied widely and were not consistent across help commands.

For example, the values varied from ``Provider`` to ``Operation`` to an abreviation to a service and even just a straight ``command`` for the custom commands.

I switched it so that the event classes rely on a command's lineage for ``event_class`` with the names of the commands in the lineage delimited by periods here are some examples:

* ``aws ec2 help`` --> ``ec2``
* ``aws s3 cp`` --> ``s3.cp``
* ``aws s3api wait object-exists`` --> ``s3api.wait.object-exists``

The end event name that bcdoc fires looks a little weird in the end because bcdoc makes assumptions on how the ``event_class`` is formatted. I can make a subsequent pull request to fix that.

However by using this format in event_class it does several things for us:

1) All of the breadcrumbs are fixed. Here are some before and after pictures of some commands:

before (no breadcrumb):
screen shot 2015-02-13 at 3 45 56 pm
after (breadcrumb is there):
screen shot 2015-02-13 at 3 47 14 pm

before (custom commands had weird names):
screen shot 2015-02-13 at 3 48 02 pm
after (name is corrected):
screen shot 2015-02-13 at 3 48 39 pm

before (waiters missed some of the commands):
screen shot 2015-02-13 at 3 49 27 pm
after (all commands are included):
screen shot 2015-02-13 at 3 50 15 pm

2) Allow us to use ``:ref:`` syntax throughout the cli documentation. I added a reference label to the top of
each reference documentation title. The labels take the form of ``.. _aws [command] ...:``. This allows you to reference the command using the sphynx ``:ref:`` syntax which I switched the breadcrumbs to use.

Note that there are 4 commits in this PR. The first two are directly from this PR: https://github.com/aws/aws-cli/pull/1148. So you do not need to look at those changes for this PR, just for the other PR.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -79,13 +79,13 @@ class CLIDocumentEventHandler(object):
         if doc.target != 'man':
             cmd_names = help_command.event_class.split('.')
             doc.write('[ ')
-            doc.write(':ref:`aws <aws>`')
+            doc.write(':ref:`aws <cli:aws>`')
             full_cmd_list = ['aws']
             for cmd in cmd_names[:-1]:
                 doc.write(' . ')
                 full_cmd_list.append(cmd)
                 full_cmd_name = ' '.join(full_cmd_list)
-                doc.write(':ref:`%s <%s>`' % (cmd, full_cmd_name))
+                doc.write(':ref:`%s <cli:%s>`' % (cmd, full_cmd_name))
             doc.write(' ]')
 
     def doc_title(self, help_command, **kwargs):
@@ -94,7 +94,7 @@ class CLIDocumentEventHandler(object):
         reference = help_command.event_class.replace('.', ' ')
         if reference != 'aws':
             reference = 'aws ' + reference
-        doc.writeln('.. _%s:' % reference)
+        doc.writeln('.. _cli:%s:' % reference)
         doc.style.h1(help_command.name)
 
     def doc_description(self, help_command, **kwargs):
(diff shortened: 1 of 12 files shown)


kyleknap:
Coverage report looks fine despite comment

jamesls:
I think we should get the bcdoc event names updated before merging this.  I'd rather avoid having changes such as: https://github.com/aws/aws-cli/pull/1151/files#diff-10d9faf99f1b18c865fd1aebf80d163bR26 which should be unnecessary once we fix the bcdoc event names.

kyleknap:
So these tests will fail until this bcdoc PR is merged in: https://github.com/boto/bcdoc/pull/33

In addition, I made a change to how I do the bread crumbs and references at the top of section headers. I made the change such that they are prefixed by ``cli:``. This helped avoid collisions with references created from docs in the JSON model. There were a few ``aws <servicename>`` references in the models.

Otherwise it is ready to be looked at for both PR's.
@jamesls @danielgtaylor

jamesls:
:shipit:  Let's get a clean run through travis after the bcdoc merge.  Otherwise looks good.

kyleknap:
Test pass with bcdoc merge.

Coverage stayed the same or increased for files I had changed. Only the s3 ``tasks.py`` decreased, which has been happening for most cli pull request, and my changes should not have affected it at all.

Merging otherwise.

PR #1153, opened 2015-02-14
Clarify pronouns for --cli-input-json

markstos:
It was not clear what the "it" and "their" pronouns here referred, making it unclear whether the JSON values overrode the CLI values or vice-versa. To make it clear, explicit references to the CLI and JSON values were used instead.

Also it's clearer to express what *will* happen ("will override") instead of the old docs which expressed what *didn't* happen ("won't clobber").


kyleknap:
Thanks for clearing that up. Merging.

PR #1155, opened 2015-02-16
docs(readme): include AWS Container Service

gdi2290:
missing Amazon EC2 Container Service


jamesls:
Not opposed to adding this, but perhaps we should link to http://docs.aws.amazon.com/cli/latest/reference/#available-services instead.  The AWS CLI has parity with the other AWS SDKs now, so there shouldn't be any services it doesn't support as far as I know.

PR #1156, opened 2015-02-16
Switch s3 ls to use clients

kyleknap:
This is part of the switch of the ``s3`` commands to clients. I only switched the ``ls`` command to use clients. All other ``s3`` commands use operation objects.

cc @danielgtaylor @jamesls

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -266,12 +266,20 @@ def get_endpoint(service, region, endpoint_url, verify):
                                 verify=verify)
 
 
+def get_client(session, region, endpoint_url, verify):
+    return session.create_client('s3', region_name=region,
+                                 endpoint_url=endpoint_url, verify=verify)
+
+
 class S3Command(BasicCommand):
     def _run_main(self, parsed_args, parsed_globals):
         self.service = self._session.get_service('s3')
         self.endpoint = get_endpoint(self.service, parsed_globals.region,
                                      parsed_globals.endpoint_url,
                                      parsed_globals.verify_ssl)
+        self.client = get_client(self._session, parsed_globals.region,
+                                 parsed_globals.endpoint_url,
+                                 parsed_globals.verify_ssl)
 
 
 class ListCommand(S3Command):
@@ -321,11 +329,11 @@ class ListCommand(S3Command):
             return 0
 
     def _list_all_objects(self, bucket, key, page_size=None):
-        operation = self.service.get_operation('ListObjects')
-        iterator = operation.paginate(self.endpoint, bucket=bucket,
-                                      prefix=key, delimiter='/',
+        paginator = self.client.get_paginator('list_objects')
+        iterator = paginator.paginate(Bucket=bucket,
+                                      Prefix=key, Delimiter='/',
                                       page_size=page_size)
-        for _, response_data in iterator:
+        for response_data in iterator:
             self._display_page(response_data)
 
     def _display_page(self, response_data, use_basename=True):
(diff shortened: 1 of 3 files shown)


kyleknap:
Coverage in files that I actually changed improved.

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -76,50 +86,49 @@ def test_rb_command_with_force_deletes_objects_in_bucket(self):
...
         parsed_args = FakeArgs(dir_op=False, paths='s3://',
                                human_readable=False, summarize=False)
         ls_command._run_main(parsed_args, parsed_global)
         # We should only be a single call.
-        self.session.get_service.return_value.get_operation.assert_called_with(
-            'ListBuckets')
-        call = self.session.get_service.return_value.get_operation\
-                .return_value.call
+        call = self.session.create_client.return_value.list_buckets
+        self.assertTrue(call.called)
```
danielgtaylor: This will be asserted by the `call.called_count` check anyway, so you don't really need to assert `call.called`.

danielgtaylor:
Just the one minor comment, then LGTM :ship:-it!

awscli/customizations/s3/subcommands.py
```diff
@@ -266,12 +266,20 @@ def get_endpoint(service, region, endpoint_url, verify):
...
+                                 endpoint_url=endpoint_url, verify=verify)
+
+
 class S3Command(BasicCommand):
     def _run_main(self, parsed_args, parsed_globals):
         self.service = self._session.get_service('s3')
         self.endpoint = get_endpoint(self.service, parsed_globals.region,
                                      parsed_globals.endpoint_url,
                                      parsed_globals.verify_ssl)
+        self.client = get_client(self._session, parsed_globals.region,
```
jamesls: Why is this is the S3Command now?  Only the ListCommand uses it.  I think once we've moved more commands over it makes sense to move this over, but I'd rather keep these changes minimal.

awscli/customizations/s3/subcommands.py
```diff
@@ -266,12 +266,20 @@ def get_endpoint(service, region, endpoint_url, verify):
                                 verify=verify)
 
 
+def get_client(session, region, endpoint_url, verify):
```
jamesls: Why is this a module function?  Given the s3 commands will be using this, it seems to make the most sense as an internal method.

awscli/customizations/s3/subcommands.py
kyleknap: Sure, I can move it over, but the ``WebsiteCommand`` will use it as well when I switch it over. So this change will only be temporary when I make a pull request to convert ``s3 website`` to clients.

awscli/customizations/s3/subcommands.py
kyleknap: ``CommandArchitecture`` object will use it as well, and that does not inherit from any command class.

kyleknap:
@jamesls 

The changes that I have been making have a general pattern of wherever there is a method involving or is instantiating an endpoint, I make it in terms of clients. For most of these changes, I could optimize the code (in terms of location of method and naming) for that specific pull request, but in the end (once we are completely switched over to clients), the paralleling client related methods and attributes will need to replace where the endpoint methods and attributes are.


I am willing to make the optimizations specific to each pull request, but my migration plan was to mirror the endpoints and then once everything was running on clients. I was going to send in a final pull request that completely strips the s3 code of any traces of endpoints and operation objects. That seemed like a simple migration strategy. Let me know what you think?

jamesls:
Sure that's fine.  I'm not opposed to having specific commands not switched over, but for commands that aren't switched over we're unnecessarily creating a client, which seems less than ideal.  I'd either like to see us switch all of s3 over before getting this in a public release, or only create clients for the code paths  and commands that actually fully switched over to clients.

jamesls:
Sorry, that previous comment was a shipit.  To account for the fact that we may not have all the s3 code moved over to clients by the next release it might be worth creating a branch where we merge all the s3 conversions into (which can live in aws/aws-cli), and once everything's converted over, we can merge into develop.

kyleknap:
Closing this PR out I made a branch that started at the head of develop and merged this PR to it: https://github.com/aws/aws-cli/tree/clients-s3

PR #1159, opened 2015-02-16
Update s3 end to end integration test to clients

kyleknap:
This updates the ``test_plugin.py`` tests such that they rely on clients as opposed to operations objects.

cc @danielgtaylor @jamesls

### DIFF
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -113,8 +113,6 @@ class BaseS3CLICommand(unittest.TestCase):
         if extra_args is not None:
             call_args.update(extra_args)
         response = client.put_object(**call_args)
-        status_code = response['ResponseMetadata']['HTTPStatusCode']
-        self.assertEqual(status_code, 200)
         self.addCleanup(self.delete_key, bucket_name, key_name)
 
     def delete_bucket(self, bucket_name):
@@ -123,8 +121,6 @@ class BaseS3CLICommand(unittest.TestCase):
             's3', region_name=self.regions[bucket_name])
         response = client.delete_bucket(Bucket=bucket_name)
         del self.regions[bucket_name]
-        status_code = response['ResponseMetadata']['HTTPStatusCode']
-        self.assertEqual(status_code, 204, response)
 
     def remove_all_objects(self, bucket_name):
         client = self.session.create_client(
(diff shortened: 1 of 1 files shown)


danielgtaylor:
LGTM :ship:-it!

kyleknap:
@danielgtaylor I did not touch that file especially since this is an update on integration tests. The ``tasks.py`` coverage has been changing ever so often even though nothing has changed in it for other pull requests. James and I believe it has something to do with the threaded part of this code.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -104,74 +104,81 @@ def create_bucket(self, name=None, region=None):
...
+            'Bucket': bucket_name,
+            'Key': key_name, 'Body': contents
         }
         if extra_args is not None:
             call_args.update(extra_args)
-        http = operation.call(**call_args)[0]
-        self.assertEqual(http.status_code, 200)
+        response = client.put_object(**call_args)
+        status_code = response['ResponseMetadata']['HTTPStatusCode']
+        self.assertEqual(status_code, 200)
```
jamesls: This is not necessary.  Clients will raise exceptions on errors.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -104,74 +104,81 @@ def create_bucket(self, name=None, region=None):
...
-        operation = self.service.get_operation('DeleteBucket')
-        endpoint = self.service.get_endpoint(self.regions[bucket_name])
-        response = operation.call(endpoint, bucket=bucket_name)[0]
+        client = self.session.create_client(
+            's3', region_name=self.regions[bucket_name])
+        response = client.delete_bucket(Bucket=bucket_name)
         del self.regions[bucket_name]
-        self.assertEqual(response.status_code, 204, response.content)
+        status_code = response['ResponseMetadata']['HTTPStatusCode']
+        self.assertEqual(status_code, 204, response)
```
jamesls: Same thing, status code checks not required.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -104,74 +104,81 @@ def create_bucket(self, name=None, region=None):
...
-        endpoint = self.service.get_endpoint(self.regions[bucket_name])
-        http, parsed = operation.call(
-            endpoint, bucket=bucket_name, key=key_name)
-        return http.status_code == 200
+        client = self.session.create_client(
+            's3', region_name=self.regions[bucket_name])
+        try:
+            response = client.head_object(Bucket=bucket_name, Key=key_name)
+            status_code = response['ResponseMetadata']['HTTPStatusCode']
+            return status_code == 200
```
jamesls: status code check not needed.

jamesls:
Figured I'd just make a general comment, rather than mentioning every spot, but you should no longer need to do status checks for clients.  They will raise exceptions on errors.  Previously the `Operation.call` interface would never raise exceptions so we needed the status check.  Now we can just let the exception propogate and cause the test to error out.

kyleknap:
@jamesls 
I removed all of the status code checks from the test helper methods.

jamesls:
:shipit:

PR #1164, opened 2015-02-18
Switch to clients for S3 filegenerator

kyleknap:
This pull request switches the ``FileGenerator`` class for the s3 commands to rely on clients as opposed to operation objects. Ignore the first commit about the ``ls`` commands as that is just a commit from another pull request that I needed for this one: https://github.com/aws/aws-cli/pull/1156

I next plan to make the change to clients in the ``S3Handler`` and ``FileInfo`` classes. And then completely remove all functions that are related to operation objects in the s3 command code.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -314,7 +314,6 @@ class FileGenerator(object):
             if e.http_status_code == 404:
                 # The key does not exist so we'll raise a more specific
                 # error message here.
-                copy_fields['error_code'] = 'NoSuchKey'
                 copy_fields['error_message'] = 'Key "%s" does not exist' % key
             else:
                 reason = six.moves.http_client.responses[
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -114,10 +114,9 @@ class FileGenerator(object):
     under the same common prefix.  The generator yields corresponding
     ``FileInfo`` objects to send to a ``Comparator`` or ``S3Handler``.
     """
-    def __init__(self, service, endpoint, operation_name,
-                 follow_symlinks=True, page_size=None, result_queue=None):
-        self._service = service
-        self._endpoint = endpoint
+    def __init__(self, client, operation_name, follow_symlinks=True,
+                 page_size=None, result_queue=None):
+        self._client = client
         self.operation_name = operation_name
         self.follow_symlinks = follow_symlinks
         self.page_size = page_size
@@ -279,8 +278,7 @@ class FileGenerator(object):
         if not dir_op and prefix:
             yield self._list_single_object(s3_path)
         else:
-            operation = self._service.get_operation('ListObjects')
-            lister = BucketLister(operation, self._endpoint)
+            lister = BucketLister(self._client)
             for key in lister.list_objects(bucket=bucket, prefix=prefix,
                                            page_size=self.page_size):
                 source_path, size, last_update = key
@@ -304,10 +302,8 @@ class FileGenerator(object):
         # IAM policies with the smallest set of permissions needed) and
         # instead use a HeadObject request.
         bucket, key = find_bucket_key(s3_path)
-        operation = self._service.get_operation('HeadObject')
         try:
-            response = operation.call(
-                self._endpoint, bucket=bucket, key=key)[1]
+            response = self._client.head_object(Bucket=bucket, Key=key)
         except ClientError as e:
             # We want to try to give a more helpful error message.
             # This is what the customer is going to see so we want to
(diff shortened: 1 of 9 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -493,22 +493,23 @@ def _decode_keys(self, parsed, **kwargs):
...
         self._handler = handler
         self._unique_id = unique_id
         self._unique_id_uses_count = unique_id_uses_count
 
     def __enter__(self):
-        self._session.register(self._event_name, self._handler, self._unique_id,
-                               self._unique_id_uses_count)
+        self._event_emitter.register(self._event_name, self._handler,
+                                     self._unique_id,
+                                     self._unique_id_uses_count)
```
kyleknap: I thought by switching over to clients that I could remove ``unique_id_uses_count`` argument. I originally did the change but I saw test failures on the test that was added to catch that a handler is not deregistered too early when two ScopedEventHandlers are being used to listing objects. By switching to clients I thought that event emitters across clients were separate, so registering and deregistering on one client did not affect the other. We may need to look into it.

awscli/customizations/s3/filegenerator.py
```diff
@@ -304,10 +302,8 @@ def _list_single_object(self, s3_path):
...
         # instead use a HeadObject request.
         bucket, key = find_bucket_key(s3_path)
-        operation = self._service.get_operation('HeadObject')
         try:
-            response = operation.call(
-                self._endpoint, bucket=bucket, key=key)[1]
+            response = self._client.head_object(Bucket=bucket, Key=key)
         except ClientError as e:
             # We want to try to give a more helpful error message.
             # This is what the customer is going to see so we want to
```
jamesls: Please double check if we need all this still.  The client objects should handle these scenarios for us, so we shouldn't need to fix up error messages here.

awscli/customizations/s3/utils.py
jamesls: You shouldn't need to do this.  Is it possible the test is now invalid?

I'd rather get to the bottom of this rather than leaving in unique_id_uses_count.
If it is a bug in botocore, let's get that fixed, but botocore should have per client events as far as I know.

kyleknap: The test is not invalid because it was erroring on a ``test_plugin.py`` integration test...

I can dive a little bit deeper into this to see if it is an issue with botocore.

awscli/customizations/s3/filegenerator.py
kyleknap: Here is the error message using botocore clients:
```
In [1]: import botocore.session

In [2]: session = botocore.session.get_session()

In [3]: s3 = session.create_client('s3')

In [4]: s3.head_object(Bucket='mybucketfoo', Key='some_key')
---------------------------------------------------------------------------
ClientError                               Traceback (most recent call last)
<ipython-input-4-2bede845a716> in <module>()
----> 1 s3.head_object(Bucket='mybucketfoo', Key='some_key')

/Users/kyleknap/Documents/GitHub/botocore/botocore/client.pyc in _api_call(self, **kwargs)
    324 
    325             if http.status_code >= 300:
--> 326                 raise ClientError(parsed_response, operation_name)
    327             else:
    328                 return parsed_response

ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```
Here is the error from s3:
```
A client error (NoSuchKey) occurred when calling the HeadObject operation: Key "some_key" does not exist
```
I like the second one better. You could figure out that the object does not exist using the first exception message, but it is not very explicit. @jamesls What is your opinion?

awscli/customizations/s3/utils.py
kyleknap: @jamesls 
This looks like a bug in botocore: https://github.com/boto/botocore/issues/465

kyleknap:
With the botocore bug fixed: https://github.com/boto/botocore/pull/467 I was able to remove ``unique_id_uses_count``.

awscli/customizations/s3/filegenerator.py
jamesls: At the same time, the former is the same error message you get when you call `s3api head-object` so the consistency is nice.  I think the 404 vs. NoSuchKey is not that big of a deal, I'd think most people would understand a 404.  However, the latter message has the key name which I think is really helpful.  What about keeping the code the same, but just changing the error message to include the key name?

kyleknap: Yeah +1 on that.

kyleknap:
@jamesls I made the switch to use 404 instead of ``NoSuchKey``

Once this PR gets approved, I will merge this into the ``clients-s3`` branch **not** ``develop``, even though the PR is targeting ``develop``

jamesls:
:shipit:

kyleknap:
Closing PR merged it to the ``clients-s3`` branch

PR #1167, opened 2015-02-19
Use CloudTrail's regionalized policy templates

danielgtaylor:
This change switches away from always using an S3 bucket in `us-west-2`
to fetch CloudTrail policy information for S3 and SNS when setting up
new buckets/topics and instead uses the `awscloudtrail-policy-REGION`
regionalized buckets. It also updates the policy version to use the
most recent release.

Tests are updated and a couple new ones make sure the regionalized
buckets are used.

cc @jamesls @kyleknap 

Note: this code still needs to be updated to use clients, but that seems
like it should be a separate task.

### DIFF
--- a/awscli/customizations/cloudtrail.py
+++ b/awscli/customizations/cloudtrail.py
@@ -24,6 +24,10 @@ S3_POLICY_TEMPLATE = 'policy/S3/AWSCloudTrail-S3BucketPolicy-2014-12-17.json'
 SNS_POLICY_TEMPLATE = 'policy/SNS/AWSCloudTrail-SnsTopicPolicy-2014-12-17.json'
 
 
+class CloudTrailError(Exception):
+    pass
+
+
 def initialize(cli):
     """
     The entry point for CloudTrail high level commands.
@@ -191,13 +195,11 @@ class CloudTrailSubscribe(BasicCommand):
             data = self.s3.GetObject(
                 bucket='awscloudtrail-policy-' + self.region_name,
                 key=key_name)
-            policy = data['Body'].read().decode('utf-8')
-        except Exception:
-            LOG.error('Unable to get regional policy template for'
-                      ' region %s: %s', self.region_name, key_name)
-            raise
-
-        return policy
+            return data['Body'].read().decode('utf-8')
+        except Exception as e:
+            raise CloudTrailError(
+                'Unable to get regional policy template for'
+                ' region %s: %s. Error: %s', self.region_name, key_name, e)
 
     def setup_new_bucket(self, bucket, prefix, policy_url=None):
         """
--- a/awscli/customizations/cloudtrail.py
+++ b/awscli/customizations/cloudtrail.py
@@ -20,8 +20,8 @@ from botocore.vendored import requests
 
 
 LOG = logging.getLogger(__name__)
-S3_POLICY_TEMPLATE = 'policy/S3/AWSCloudTrail-S3BucketPolicy-2013-11-01.json'
-SNS_POLICY_TEMPLATE = 'policy/SNS/AWSCloudTrail-SnsTopicPolicy-2013-11-01.json'
+S3_POLICY_TEMPLATE = 'policy/S3/AWSCloudTrail-S3BucketPolicy-2014-12-17.json'
+SNS_POLICY_TEMPLATE = 'policy/SNS/AWSCloudTrail-SnsTopicPolicy-2014-12-17.json'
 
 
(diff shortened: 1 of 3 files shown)


kyleknap:
LGTM :ship: I agree it would be nice if you could update to clients. But, at this time it will not be feasible for you until ``region_name`` is exposed in clients for the reason you need the region name to determine the bucket to search in and setting the bucket configuration. It is the same problem that I am running into with the s3 commands.

awscli/customizations/cloudtrail.py
```diff
@@ -204,8 +206,9 @@ def setup_new_bucket(self, bucket, prefix, policy_url=None):
         if policy_url:
             policy = requests.get(policy_url).text
         else:
-            data = self.s3.GetObject(bucket='awscloudtrail',
-                                     key=S3_POLICY_TEMPLATE)
+            data = self.s3.GetObject(
```
jamesls: It might be nice to provide a better error message if a policy hasn't been copied over to a specific region now that this is a new possible failure mode.

awscli/customizations/cloudtrail.py
```diff
@@ -282,8 +284,9 @@ def setup_new_topic(self, topic, policy_url=None):
         if policy_url:
             policy = requests.get(policy_url).text
         else:
-            data = self.s3.GetObject(bucket='awscloudtrail',
-                                     key=SNS_POLICY_TEMPLATE)
+            data = self.s3.GetObject(
```
jamesls: Same thing here (better error messages).  Might be worth refactoring now too since this looks like dupe code from line 209.

jamesls:
Overall looks good.  Just a small comment on the new failure mode.

danielgtaylor:
@jamesls code has been refactored/updated to log an error message, then pass-through the underlying exception. If we want to expose a custom exception just for this case then I'm not opposed, but it's unclear to me which Botocore exceptions might happen during the `operation.call(...)` method.

awscli/customizations/cloudtrail.py
```diff
@@ -184,6 +186,18 @@ def _call(self, options, parsed_globals):
...
+        try:
+            data = self.s3.GetObject(
+                bucket='awscloudtrail-policy-' + self.region_name,
+                key=key_name)
+        except Exception:
+            LOG.error('Unable to get regional policy template for'
+                      ' region %s: %s', self.region_name, key_name)
+            raise
+
+        return data['Body'].read().decode('utf-8')
```
jamesls: This should probably be moved inside the try/except because we can still hit timeout/connection errors when we try to read the response body.

jamesls:
It would be good to add a test for this new code.

danielgtaylor:
@jamesls please take another look. The coverage decrease seems like a false alarm because I only added tests.

awscli/customizations/cloudtrail.py
```diff
@@ -184,6 +186,19 @@ def _call(self, options, parsed_globals):
                 'Logs will be delivered to {bucket}:{prefix}\n'.format(
                     bucket=bucket, prefix=options.s3_prefix or ''))
 
+    def _get_policy(self, key_name):
+        try:
+            data = self.s3.GetObject(
+                bucket='awscloudtrail-policy-' + self.region_name,
+                key=key_name)
+            policy = data['Body'].read().decode('utf-8')
```
jamesls: You can just `return data['Body'].read().decode('utf-8')` instead of needing the `return policy` on line 200.

tests/unit/customizations/test_cloudtrail.py
```diff
@@ -144,14 +158,41 @@ def test_s3_create_set_policy_fail(self):
...
-        orig = self.subscribe.s3.GetObject
+    def test_s3_get_policy_fail(self):
         self.subscribe.s3.GetObject = Mock(side_effect=Exception('Error!'))
 
         with self.assertRaises(Exception):
             self.subscribe.setup_new_bucket('test', 'logs')
 
-        self.subscribe.s3.GetObject = orig
+    def test_s3_get_policy_logs_messages(self):
+        cloudtrail.LOG = Mock()
```
jamesls: This is never restored.  A unit test shouldn't be modifying global state after its execution.

jamesls:
Why not raise a specific exception and just use `assertRaises()` in your test?  Simple code, simpler test, no `LOG` patching required.

jamesls:
:shipit:  Thanks!  Don't forget to update the changelog.

PR #1168, opened 2015-02-20
Add terminal paging to help for windows

nhumrich:
Add terminal paging for windows.  In linux, the "help" gets paginated with a terminal PAGER, but in windows it just prints to the screen. This fixes that problem. Note that pydoc is a python standard library and on windows it defaults to using "more" which is a windows utility much like less.

### DIFF
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -15,6 +15,7 @@ import logging
 import os
 import platform
 import shlex
+import pydoc
 from subprocess import Popen, PIPE
 
 from docutils.core import publish_string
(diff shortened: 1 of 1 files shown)


jamesls:
Seems like we should just pull up ` get_pager_cmdline` and other paging methods into the base help class and default to a PAGER of `more` for windows.  That way we'd have consistent behavior across platforms.

nhumrich:
Or you could just let pydoc.pager do all the work and remove some code. But i'm not sure how much that will change behavior.

jamesls:
Not opposed to that.  My only concern is that if we're going to make a change to support paging on windows, we should make it as consistent cross platform as possible.  If you could maintain these features, then feel free to switch over to pydoc:

1. Works "out of the box" for both powershell/cmd on windows.
2. Honors the MANPAGER/PAGER env vars that are standard in posix.

nhumrich:
Updated. Changed pydoc to be the pager for all platforms. PosixRender just does the man page formating. Works on powershell and cmd as well as python 2 and 3. Pydoc honors PAGER but not MANPAGER, added a line to handle MANPAGER.

awscli/help.py
```diff
@@ -59,12 +60,22 @@ class HelpRenderer(object):
...
+    
+    def format(self, contents):
         """
         Each implementation of HelpRenderer must implement this
-        render method.
+        format method unless it implements render.
         """
         pass
 
+    def get_pager_cmdline(self):
```
jamesls: It seems odd to have a method called `get_pager_cmdline` that doesn't actually return anything.

jamesls:
The Travis CI build is still failing for this PR and is missing tests.  I've sent an alternate PR that supports paging on windows and adds some tests for this new functionality.  I'm going to go ahead and close this out in favor of #1195.  Thanks again though for the PR and for bringing this to our attention.

PR #1173, opened 2015-02-23
Bumped CLI dependancy of bcdoc

kyleknap:
Bumped the CLI's dependency of bcdoc based on its latest version bump.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #1175, opened 2015-02-24
Add role credentials support to role_arn

kawaz:
Add role credentials support to role_arn.

My ~/.aws/credentials is here.

```ini
[default]
# default is role credentials
[user1]
aws_access_key_id = XXX
aws_secret_access_key = XXX
[profile1-by-default]
role_arn = arn:aws:iam::1111222233334444:role/profile1
source_profile = default
[profile1-by-user1]
role_arn = arn:aws:iam::1111222233334444:role/profile1
source_profile = user1
```

This was successful.
```bash
$ aws ec2 describe-instances --profile profile1-by-user1
{...}
```

This failed.

```
$ aws ec2 describe-instances --profile profile1-by-default
'aws_access_key_id'
```

I can't use role_arn, when source_profile is default and role credentials.
This patch fix this problem.

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -11,6 +11,7 @@ from dateutil.tz import tzlocal
 from botocore import credentials
 from botocore.compat import total_seconds
 from botocore.exceptions import PartialCredentialsError
+from botocore.utils import InstanceMetadataFetcher
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


kyleknap:
Thank you for the pull request!

Unfortunately, I do not think that we will be able to take this pull request in. The main reason being that the shared credentials file (``~/.aws/credentials``) is shared across all of the AWS SDKs and there is a rule across the SDK teams that the shared credential file only supports entries for ``aws_access_key_id``, ``aws_secret_access_key``, and ``aws_session_token``.  The entries for assume role such as ``role_arn`` and ``source_profile`` are specific to the CLI and isolated to the CLI's config file (``~/.aws/config``). In your example, the fact that the ``profile1-by-default`` assume role, using the shared credentials file, works is unintentional. Ultimately, we do not want to add any additional support for accepting the assume role entries in the shared credentials file because it may force other SDKs to have to be able to parse these entries even though they are only specific to the CLI.

I would suggest moving your assume role entries out of ``~/.aws/credentials`` and into ``~/.aws/config``. Here is a pull request to further help you set that up: https://github.com/aws/aws-cli/pull/1182. If you still run into the issue that you first ran into or any other issues please submit an issue to us.

PR #1182, opened 2015-02-25
Add config var topic guide

jamesls:
I added a full description for assume role functionality
in this topic guide.  I'm not opposed to doing a similar thing
like s3 config and just document at a high level the assume role
params, and refer the user to a separate assume role topic guide
for more information.

I know I'll need to update the topic index, but given there's a few topic guide reviews
out right now, I'll do that once I get other topics merged so I can avoid a merge
conflict.

cc @kyleknap @danielgtaylor

### DIFF
--- a/awscli/topics/config-vars.rst
+++ b/awscli/topics/config-vars.rst
@@ -1,0 +1,209 @@
+:title: AWS CLI Configuration Variables
+:description: Configuration Variables for the AWS CLI
+:category: General
+:related command: configure, configure get, configure set
+
+Configuration values for the AWS CLI can come from several sources:
+
+* As a command line option
+* As an environment variable
+* As a value in the AWS CLI config file
+* As a value in the AWS Shared Credential file
+
+Some options are only available in the AWS CLI config.  This topic guide covers
+all the configuration variables available in the AWS CLI.
+
+Note that if you are just looking to get the minimum required configuration to
+run the AWS CLI, we recommend running ``aws configure``, which will prompt you
+for the necessary configuration values.
+
+Config File Format
+==================
+
+The AWS CLI config file, which defaults to ``~/.aws/config`` has the following
+format::
+
+    [default]
+    aws_access_key_id=foo
+    aws_secret_access_key=bar
+    region=us-west-2
+
+The ``default`` section refers to the configuration values for the default
+profile.  You can create profiles, which represent logical groups of
+configuration.  Profiles that aren't the default profile are specified by
+creating a section titled "profile profilename"::
+
+    [profile testing]
+    aws_access_key_id=foo
+    aws_secret_access_key=bar
+    region=us-west-2
+
+Nested Values
+-------------
+
+Some service specific configuration, discussed in more detail below, has a
+single top level key, with nested sub values.  These sub values are denoted by
+indentation::
+
+    [profile testing]
+    aws_access_key_id = foo
+    aws_secret_access_key = bar
+    region = us-west-2
+    s3 =
+      max_concurrent_requests=10
+      max_queue_size=1000
+
+
+General Options
+===============
+
+The AWS CLI has a few general options:
+
+=========== ========= ===================== ===================== ============================
+Variable    Option    Config Entry          Environment Variable  Description
+=========== ========= ===================== ===================== ============================
+profile     --profile profile               AWS_DEFAULT_PROFILE   Default profile name
+----------- --------- --------------------- --------------------- ----------------------------
+region      --region  region                AWS_DEFAULT_REGION    Default AWS Region
+----------- --------- --------------------- --------------------- ----------------------------
+output      --output  output                AWS_DEFAULT_OUTPUT    Default output style
+=========== ========= ===================== ===================== ============================
+
+The third column, Config Entry, is the value you would specify in the AWS CLI
+config file.  By default, this location is ``~/.aws/config``.  If you need to
+change this value, you can specify the ``AWS_CONFIG_FILE`` environment variable
+to change this location.
+
+The valid values of the ``output`` configuration variable are:
+
+* json
+* table
+* text
+
+
+Credentials
+===========
+
+Credentials can be specified in several ways:
+
+* Environment variables
+* The AWS Shared Credential File
+* The AWS CLI config file
+
+=========== ===================== ===================== ============================
+Variable    Creds/Config Entry    Environment Variable  Description
+=========== ===================== ===================== ============================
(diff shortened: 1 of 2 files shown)


awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
+:title: AWS CLI Configuration Variables
+:description: Configuration Variables for the AWS CLI
+:category: General
```
kyleknap: The ``return-codes`` topic uses the category ``General Topics``. Do you prefer ``General`` or ``General Topics`` more? We should have some consistency.

awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
...
+:related command: configure, configure get, configure set
+
```
kyleknap: You can add ``:related topic: s3-config`` to get this topic to link to the s3 configuration topic when a related items section is added for topics.

awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
...
+General Options
+===============
+
+The AWS CLI has a few general options:
+
+=========== ========= ===================== ===================== ============================
+Variable    Option    Config Entry          Environment Variable  Description
+=========== ========= ===================== ===================== ============================
+profile     --profile profile               AWS_DEFAULT_PROFILE   Default profile name
```
kyleknap: Profile's entry in the config is different than region and output, which could be confusing. You may want to make it so that it shows up exactly as it would in the config file. So profile would be ``[profile <value>]`` and region would be ``region=<value>`` etc. Then you could also show how the option on the command line maps to the config file by appending ``<value>`` to each option so make it ``--profile <value>`` and do the same for environment variables possibly for consistency.

awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
...
+----------- --------- --------------------- --------------------- ----------------------------
+region      --region  region                AWS_DEFAULT_REGION    Default AWS Region
+----------- --------- --------------------- --------------------- ----------------------------
+output      --output  output                AWS_DEFAULT_OUTPUT    Default output style
+=========== ========= ===================== ===================== ============================
+
+The third column, Config Entry, is the value you would specify in the AWS CLI
+config file.  By default, this location is ``~/.aws/config``.  If you need to
+change this value, you can specify the ``AWS_CONFIG_FILE`` environment variable
```
kyleknap: you can specify the ``AWS_CONFIG_FILE`` environment variable to change this location --> you can set the ``AWS_CONFIG_FILE`` environment variable to this location.

awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
...
+``aws_secret_access_key``, ``aws_session_token``.  **These are the only
+supported values in the shared credential file.**
+
+Precedence
+----------
+
+Credentials from environment variables have precedence over credentials from
+the shared credentials and AWS CLI config file.  Credentials specified in the
+shared credentials file have precedence over credentials in the AWS CLI config
+file.
```
kyleknap: How about precedence for region and values specified on the command line? This could get real wordy. You may want to break this up to a nice diagram or table to make it really easy to tell precedence.

awscli/topics/config-vars.rst
```diff
@@ -0,0 +1,209 @@
...
+    [profile development]
+    aws_access_key_id=foo
+    aws_secret_access_key=bar
+    s3 =
+      max_concurrent_requests = 20
+      max_queue_size = 10000
+      multipart_threshold = 64MB
+      multipart_chunksize = 16MB
+
+.. I know I'll need to add a proper xref to the s3-config topic.
```
kyleknap: I am not sure if ``:ref:`` notation will work for man pages. The man page will generate but the topic will have something like ``:ref: S3 configuration <cli: aws help s3-config>`` in the body of the man page. Html will work fine because that is generates by sphinx but the man page docutil writer cannot handle the ``:ref:`` directive. As a result, you get the literal in the body.

kyleknap:
Looks good. I just had a few comments/suggestions to help make some of the concepts more clear like how configuration values are mapped across configuration location and how precedence works. Also can you generate the topic-tags.json for this as well?

awscli/topics/config-vars.rst
jamesls: I don't have a strong opinion either way.  If I had to pick, I'd go with "General", mainly because none of the other categories have "Topics" appended to them, and this is the topic guide list so it's implied these are all categories for topics.

awscli/topics/config-vars.rst
jamesls: Ok I'll update.  Out of curiosity, do you think these should be pluralized?  It does seem somewhat odd to call these "related command", "related topic", but have them actually accept a csv list.

awscli/topics/config-vars.rst
jamesls: I'm thinking I should just remove the `profile` from the `Config Entry` column, and just explain how profile maps after the table.

awscli/topics/config-vars.rst
kyleknap: We can make these pluralized. It makes more sense given that it is looking like most topics have multiple related commands and topics.

awscli/topics/config-vars.rst
kyleknap: General is fine with me. We just need to change the return-codes topic then.

danielgtaylor: +1 for `General`

awscli/topics/config-vars.rst
jamesls: I'll update this.

awscli/topics/config-vars.rst
kyleknap: Yeah I like that best.

awscli/topics/config-vars.rst
jamesls: One future option might be to have our own role for cross references where in the man page renderer it'll render as the "aws help <topic>" , and in the html docs it will render a link.

danielgtaylor:
Looks good. Generally agree with Kyle's comments.

awscli/topics/config-vars.rst
kyleknap: Yeah that is something that I would like to see in the future. It would be super useful for any custom docs that we do.

jamesls:
Ok I believe I've incorporated all the feedback.

kyleknap:
Thanks :ship:

PR #1183, opened 2015-02-25
Fix issue with parsing

kyleknap:
Issue was if there is a ] inside one of the values of a list, the end character would get removed. So something like ``foo=[bar, *[biz]*, bar]`` would get parsed to ``foo=bar,*[biz],bar``. To get that to work, you would have to use escape characters.

Note that we still run into the issue where the argument is ``foo=[bar,[biz],bar]``, it will be translated to ``foo=bar,[biz,bar`` and the escape character would need to be used to get the correct format. To fix that, that will take a much larger change in how we parse input.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -85,7 +85,7 @@ def _eat_items(value, iter_parts, part, end_char, replace_char=''):
         except StopIteration:
             raise ValueError(value)
         chunks.append(current.replace(replace_char, ''))
-        if end_char in current:
+        if end_char == current[-1]:
             break
     return ','.join(chunks)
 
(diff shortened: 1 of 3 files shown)


jamesls:
:shipit:

awscli/utils.py
```diff
@@ -85,7 +85,7 @@ def _eat_items(value, iter_parts, part, end_char, replace_char=''):
         except StopIteration:
             raise ValueError(value)
         chunks.append(current.replace(replace_char, ''))
-        if end_char in current:
+        if end_char == current[-1]:
```
danielgtaylor: Can't the line above replace some (and potentially all) characters with a blank? What happens when `current` is `''` and we do the `current[-1]` check? May want to check for that case.

kyleknap: @danielgtaylor How does using endswith() sound?

danielgtaylor:
Thanks. :ship:-it!

PR #1189, opened 2015-02-27
Convert cloudtrail customization over to clients.

jamesls:
Also added a bit to the test suite as there some components
I was changing that weren't tested.

The only weird thing I had to add was unregistering the error handler
the awscli uses.  Botocore clients raise exceptions now, but with this
customization in place the AWS CLI's client error was being raised
instead of botocores, which is what we want.

Also, I changed the RC to be 0 in the case of success, which is what we want.

cc @kyleknap @danielgtaylor

### DIFF
--- a/awscli/customizations/cloudtrail.py
+++ b/awscli/customizations/cloudtrail.py
@@ -15,8 +15,8 @@ import logging
 import sys
 
 from awscli.customizations.commands import BasicCommand
-from awscli.customizations.service import Service
 from botocore.vendored import requests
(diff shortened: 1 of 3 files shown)


jamesls:
Right, this depends on https://github.com/boto/botocore/pull/474 and will fail until that's merged.

danielgtaylor:
Thanks for cleaning up this module and its tests. :ship:-it!

kyleknap:
Looks good. :ship:

PR #1190, opened 2015-02-27
Rename assert_params_for_cmd2 to assert_params_for_cmd

jamesls:
assert_params_for_cmd, the only version, is now removed.
When we do these functional tests, we only check down to the
top most botocore layer, not at the wire protocol level.

cc @danielgtaylor @kyleknap


danielgtaylor:
LGTM :ship:-it!

PR #1191, opened 2015-03-02
Add retry logic for read timeouts for s3 commands

kyleknap:
It looks like when we upgraded requests, it upgraded urllib3 which better handles read timeouts. This allows us to retry their read timeouts.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #1193, opened 2015-03-02
Bump up ctrl c timeout to 60 seconds in test

jamesls:
The purpose of this test to just to ensure the process doesn't
hang, and it's dependent on the amount of time it takes to execute
a task.  Bumping this up to fix some sporadic failures we were
seeing.

cc @kyleknap @danielgtaylor


kyleknap:
:ship:

PR #1195, opened 2015-03-03
Support paging help through "more" command on windows

jamesls:
This is an alternate to  #1168 except it refactors the existing internals to support paging through "more" on windows rather than using pydoc.  I also confirmed that this works on powershell as well as cmd.

The existing PR had failing travis CI builds and was overwriting the PAGER env var.

Closes #1168.

cc @kyleknap @danielgtaylor


danielgtaylor:
LGTM :ship:-it.

kyleknap:
Same :ship:

PR #1196, opened 2015-03-03
Convert Rest of S3 Commands to Client

kyleknap:
So this is the final PR required to get the ``clients-s3`` branch off of endpoints/service/operation objects and onto clients. Along with switching the commands to use clients, I made some major changes to the s3 test suite. Here is a summary of changes that were made. Note that ``test_plugin.py`` was not really changed:

* Completely removed the use of ``FakeSession``. This whole concept relied on the fact that we were using operation and service object. We really could not reuse that for clients. Plus, it was not ideal because we had all of this extra logic when there was no for sure way of checking what we sent to botocore. Instead I switched the tests to rely on ``BaseAWSCommandParamsTest`` which allowed us to see what was being passed to botocore.

* Some of the unit and integration test relied on this ``make_loc_files`` and ``clean_loc_files`` methods which made and cleaned up a set of local files for testing. I instead made those use ``FileCreator`` which uses temporary files and ensures the files are not being shared across processes that may be running the tests because the temporary file paths are randomized.

* I removed tests that were not valid anymore (like having the s3 handler handle a ``list_objects`` task) and consolidated tests that should belong to the same class.

Also note that I removed all of the dead code that was not being used anymore like the ``get_endpoint``, ``operate``, and ``check_error`` methods.

Once this gets merged to the ``clients-s3`` branch, we will be able to merge that branch to ``develop``

cc @jamesls @danielgtaylor


kyleknap:
I am not sure what exactly went down. It did not help that I removed a lot of code that used to count to the covered totals.

I am going to try to bump up the coverage for ``tasks.py`` now that it is much easier to test items like the multipart transfers.

kyleknap:
Ok good to look at now. Added some unit test that were not in place before, and thanks to how I redesigned the tests it was pretty easy to add a test for it unlike before.

awscli/customizations/s3/fileinfobuilder.py
```diff
@@ -18,13 +20,12 @@ class FileInfoBuilder(object):
...
-    def __init__(self, service, endpoint, source_endpoint=None,
+    def __init__(self, client, source_client=None,
                  parameters = None, is_stream=False):
-        self._service = service
-        self._endpoint = endpoint
-        self._source_endpoint = endpoint
-        if source_endpoint:
-            self._source_endpoint = source_endpoint
+        self._client = client
+        self._source_client = copy.copy(client)
```
jamesls: Why is this necessary?  This just makes a shallow copy of the client's `__dict__` right?  Does `self._source_client = client` not work for this scenario?

awscli/customizations/s3/fileinfobuilder.py
```diff
@@ -18,13 +20,12 @@ class FileInfoBuilder(object):
...
+        if source_client:
```
jamesls: `if source_client is not None`

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -518,7 +518,7 @@ def test_copy_large_file_signature_v4(self):
             for i in range(num_mb):
                 f.write(b'a' * 1024 * 1024)
 
-        p = aws('s3 cp %s s3://%s/ --region eu-central-1' % (
+        p = aws('s3 cp %s s3://%s/ --region eu-central-1 --debug' % (
```
jamesls: Do we need the `--debug` here?

kyleknap:
I made changes to the PR based on the ``pr-check`` script. It runs cleanly now.

tests/unit/customizations/s3/__init__.py
```diff
@@ -11,130 +11,90 @@
...
+
+    def tearDown(self):
+        super(S3HandlerBaseTest, self).tearDown()
+        clean_loc_files(self.file_creator)
+
+    def run_s3_handler(self, s3_handler, tasks):
+        self.patch_make_request()
+        captured_stderr = six.StringIO()
+        captured_stdout = six.StringIO()
+        with mock.patch('sys.stderr', captured_stderr):
```
jamesls: This looks identical to the code in testutils.py.  What about extracting the captured stdout/stderr logic into a separate function in that module so we can share code?  I'd think something like would work:

```
with testutils.capture_output() as captured:
    <your code to run>
print(captured.stdout.getvalue())
print(captured.stderr.getvalue())
```

tests/unit/customizations/s3/test_cp_command.py
```diff
@@ -60,8 +60,45 @@ def test_trailing_slash_appended(self):
...
+                   'full=id=bar readacl=id=biz writeacl=id=baz' %
+                   (self.prefix, full_path))
+        self.parsed_responses = \
+            [{'ETag': '"c8afdb36c52cf4727836669019e69222"'}]
+        self.run_cmd(cmdline, expected_rc=0)
+        # The only operation we should have called is PutObject.
+        self.assertEqual(len(self.operations_called), 1,
+                         self.operations_called)
+        self.assertEqual(self.operations_called[0][0].name, 'PutObject')
+        self.assertEqual(self.operations_called[0][1]['Key'], 'key.txt')
```
jamesls: This might read a little better if you just grab the dict and use `assertDictEqual`, e.g.:

```
kwargs = self.operations_called[0][1]
self.assertDictEqual(kwargs, {'Key': 'key.txt', 'Bucket': '...', ...})
```

It would also catch the case where additional params showed up that weren't suppose to be there.

jamesls:
Overall I think this looks good, I just had some minor feedback.  Definitely like the move away from the `FakeSession`.  I'd like to get this all merged into the `clients-s3` branch soon so we can do some final perf testing before merging back to develop.

awscli/customizations/s3/fileinfobuilder.py
kyleknap: I think I was meaning to do deep copy here. But I do not think a deep copy is needed. We do not register any events in the s3handler so it should be fine to use the same client. It is only in the filegenerator that two seperate clients are required but this takes place after the filegenerator.

tests/integration/customizations/s3/test_plugin.py
kyleknap: That is just remnants from when I was trying to catch debug logs for a test that was failing. I think it was related to the sigv4 client signing.

kyleknap:
@jamesls I updated the code based on your feedback. Let me know what you think.

jamesls:
:shipit:  Looks good.

PR #1199, opened 2015-03-04
Use platform module to determine the operating system.

michaelgataws:
This is compatible across all supported Python versions and fixes a
problem with Python 3.3+ (where `os.platform` is 'linux', not 'linux2'.)


jamesls:
Thanks!

PR #1200, opened 2015-03-04
Update README with current info

### LINKED ISSUE #1197: "aws configure set" mixes up updating security token
Calling _aws configure set_ updates _$HOME/.aws/credentials_ when asked to update aws_access_key and aws_secret_key but aws_security_token is placed in _$HOME/.aws/config_.  This would be okay if subsequent command invocations still managed to read both the files.  However, subsequent commands fail with:
> A client error (InvalidClientTokenId) occurred when calling the ListAccessKeys operation: The security token included in the request is invalid.

If I move the security token manually from _$HOME/.aws/config_ to _$HOME/.aws/credentials_, commands work.  Attempted hinting at the file to use by adjusting AWS_CONFIG_FILE and AWS_CREDENTIAL_FILE variables but no luck.

What am I doing wrong?

jamesls:
* Add Amazon Elastic Container Service (closes #1155)
* Add ``aws configure`` as a quick way of getting started
* Change aws_security_token to aws_session_token (closes #1197)
* Add examples of manipulating output with ``--query``

cc @kyleknap @danielgtaylor

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,7 @@ Amazon Web Services.
 The currently supported services include:
 
 * AWS CloudFormation
-* AWS Data Pipeline (Preview)
+* AWS Data Pipeline
 * AWS Direct Connect
 * AWS Elastic Beanstalk
 * AWS Identity and Access Management
@@ -30,10 +30,10 @@ The currently supported services include:
 * Amazon CloudSearch
 * Amazon CloudWatch
 * Amazon DynamoDB
+* Amazon EC2 Container Service
 * Amazon ElastiCache
 * Amazon Elastic Compute Cloud
-* Amazon Elastic MapReduce (Preview)
-* Amazon Elastic Container Service (Preview)
+* Amazon Elastic MapReduce
 * Amazon Elastic Transcoder
 * Amazon Kinesis
 * Amazon Redshift
--- a/README.rst
+++ b/README.rst
@@ -33,6 +33,7 @@ The currently supported services include:
 * Amazon ElastiCache
 * Amazon Elastic Compute Cloud
 * Amazon Elastic MapReduce (Preview)
+* Amazon Elastic Container Service (Preview)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -33,6 +33,7 @@ The currently supported services include:
 * Amazon ElastiCache
 * Amazon Elastic Compute Cloud
 * Amazon Elastic MapReduce (Preview)
+* Amazon Elastic Container Service (Preview)
```
kyleknap: Official name is Amazon EC2 Container Service: http://aws.amazon.com/ecs/

README.rst
```diff
@@ -33,6 +33,7 @@ The currently supported services include:
 * Amazon ElastiCache
 * Amazon Elastic Compute Cloud
 * Amazon Elastic MapReduce (Preview)
```
kyleknap: What is the meaning of preview? Does it mean it is preview in the CLI or in preview for the service in general.

kyleknap:
Looks good. Just had a small comment about the preview label and the ECS official name. :ship:

jamesls:
Changed the name.  I actually believe these are suppose to be previews of the CLI, some of which are out of date.

I'm almost inclined to remove this now, since the CLI has supported all AWS services for a while now.  I believe this was added a while back when the CLI didn't have full service coverage.

kyleknap:
Thanks. +1 on removing preview for everything including cloudfront's, especially since I have fixed issues in cloudfront despite the preview tag placed on it. The command works fine.

jamesls:
@kyleknap I meant removing the list of supported services entirely from the README, since we support all the services now.  If we are going to leave them in, I'd prefer to leave the "preview" in cloudfront, as it is in fact still in preview.  What do you think about removing this list entirely?

kyleknap:
Oh yeah that's better. +1 on removing the list.

PR #1202, opened 2015-03-05
Fix htmlgen

jamesls:
HelpRenderer was removed with the paging changes on windows.
We now have to subclass from PagingHelpRenderer.

Verified I could build the html docs locally.

cc @kyleknap @danielgtaylor


kyleknap:
:ship:

PR #1208, opened 2015-03-09
Fix configure set nested vars

jamesls:
There was a bug where we were trying to set a nested value when it's the last item in the profile.

cc @kyleknap @danielgtaylor


kyleknap:
:ship:

PR #1210, opened 2015-03-09
Use choose-signer event for --no-sign-request

kyleknap:
This fixes the s3 integration test failure with no sign requests. I moved the CLI to rely on the same handler and event that botocore uses, ``choose-signer``. I also added an integration test for the data driven commands, and updated that test to use clients.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/customizations/globalargs.py
+++ b/awscli/customizations/globalargs.py
@@ -13,7 +13,7 @@
 import sys
 import os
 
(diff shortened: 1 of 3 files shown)


jamesls:
:shipit:

danielgtaylor:
:+1:

PR #1214, opened 2015-03-11
Update events to emit model/session objects

jamesls:
This commit updates the events emitted in cli driver
to emit service/operation object alternatives.  This change
allows handlers to be updated to not require service/operation
objects.

In many cases, some handlers took an `operation` object, but
did not actually use this, so these changes were minimal.

Note that this puts the CLI in an interim state.  The plan is
to not merge this into the develop branch until clidriver is
switched over, but this will allow a common base for all
customization switchovers to be based on.  This is why
this is a pull against the clidriver-no-clients branch for its
base, not develop.

cc @kyleknap @danielgtaylor

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -337,6 +337,7 @@ class ServiceCommand(CLICommand):
         else:
             self._service_name = service_name
         self._lineage = [self]
+        self._service_model = None
 
     @property
     def name(self):
@@ -368,6 +369,11 @@ class ServiceCommand(CLICommand):
             self._service_object = self.session.get_service(self._service_name)
         return self._service_object
 
+    def _get_service_model(self):
+        if self._service_model is None:
+            self._service_model = self.session.get_service_model(self._service_name)
+        return self._service_model
+
     def __call__(self, args, parsed_globals):
         # Once we know we're trying to call a service for this operation
         # we can go ahead and create the parser for it.  We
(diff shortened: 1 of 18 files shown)


jamesls:
This will fail until https://github.com/boto/botocore/pull/487 is merged.  It's been reviewed so I'll go ahead and merge and resubmit this build request.

tests/integration/test_cli.py
```diff
@@ -201,6 +201,9 @@ def test_no_paginate_arg(self):
         p = aws('s3api list-objects --bucket %s --no-paginate' % bucket_name)
         self.assertEqual(p.rc, 0, p.stdout + p.stderr)
 
+        p = aws('s3api list-objects --bucket %s' % bucket_name)
```
kyleknap: Is this just to make sure pagination works? Its under the no pagination test though.

awscli/clidriver.py
```diff
@@ -380,12 +386,16 @@ def __call__(self, args, parsed_globals):
...
+        service_model = self._get_service_model()
         for operation_object in service_object.operations:
             cli_name = xform_name(operation_object.name, '-')
             command_table[cli_name] = ServiceOperation(
                 name=cli_name,
                 parent_name=self._name,
+                session=self.session,
                 operation_object=operation_object,
+                operation_model=service_model.operation_model(
+                    operation_object.name),
```
kyleknap: Is there any other way we can pull the ``name`` out (other than from operation objects)? That way it would be deleting arguments from calls and constructors when we do the final purge of operation objects.

awscli/clidriver.py
```diff
@@ -598,6 +610,9 @@ def _create_argument_table(self):
         self._emit('building-argument-table.%s.%s' % (self._parent_name,
                                                       self._name),
                    operation=self._operation_object,
+                   operation_model=self._operation_model,
+                   session=self._session,
+                   command=self,
```
kyleknap: I think I am in favor of adding the command object to the event. But could we name it ``command_object`` to be consistent with ``building-command-table`` event of both the data driven and custom commands.

kyleknap:
Looks good. I had just a few some small comments.

danielgtaylor:
LGTM

jamesls:
Closing in favor of https://github.com/aws/aws-cli/pull/1220, which is more comprehensive and includes this PR.

PR #1216, opened 2015-03-11
Fix CloudTrail handling of custom policy URLs.

danielgtaylor:
This updates the CloudTrail customizations to stop making HTTP requests
to fetch custom policies, instead relying on the updated behavior of the
CLI to automatically handle file and URL parameters for custom commands.

Before this fix, the CLI would fetch the data into a string, then the
CloudTrail customization would try to fetch the string as if it were
a URL, fail, and throw an error.

cc @jamesls @kyleknap

### DIFF
--- a/tests/unit/customizations/test_cloudtrail.py
+++ b/tests/unit/customizations/test_cloudtrail.py
@@ -19,7 +19,7 @@ from tests.unit.test_clidriver import FakeSession
 from awscli.compat import six
 from awscli.customizations import cloudtrail
 from awscli.testutils import BaseAWSCommandParamsTest
-from awscli.testutils import unittest
+from awscli.testutils import unittest, temporary_file
 
 
 class TestCloudTrailPlumbing(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


kyleknap:
I like this change. Can you add a test that uses the ``BaseAWSCommandParamsTest`` to make sure that paramfile values like ``file://`` notation now works for the cloudtrail commands? I think there is already one test class that uses ``BaseAWSCommandParamsTest``: https://github.com/aws/aws-cli/blob/cloudtrail-policy/tests/unit/customizations/test_cloudtrail.py#L40

To create the custom policy in a local file, you can create a temp file with ``FileCreator`` in testutils.py

tests/unit/customizations/test_cloudtrail.py
```diff
@@ -46,6 +46,14 @@ def test_create_subscription_has_zero_rc(self):
         # sure it says log delivery is happening.
         self.assertIn('Logs will be delivered to foo', stdout)
 
+    def test_policy_from_paramfile(self):
+        with temporary_file('w') as f:
+            f.write('{"Statement": []}')
+            command = (
+                'cloudtrail create-subscription --s3-use-bucket foo '
+                '--name bar --s3-custom-policy file://{0}'.format(f.name))
```
jamesls: I agree with Kyle here.  Let's verify that the policy we pass into the botocore layer is the value we expect.

danielgtaylor:
@jamesls @kyleknap please take another look. I've updated the test to assert the policy contents are from the file.

kyleknap:
Thanks! :ship:

jamesls:
:shipit:

PR #1220, opened 2015-03-12
Switch CLI over to botocore clients

jamesls:
This removes dependencies on botocore's service/operation objects.  High level this is done in two ways:

1) Things that need to know about the structure of a service's operations/args then they now use the `botocore.model.*` classes.
2) Things that actually need to make service calls now use the client API.

cc @kyleknap @danielgtaylor


awscli/formatter.py
```diff
@@ -52,7 +52,8 @@ def _flush_stream(self, stream):
 
 
 class FullyBufferedFormatter(Formatter):
-    def __call__(self, operation, response, stream=None):
+    def __call__(self, command_name, response,
+                 is_response_paginated=False, stream=None):
```
jamesls: I think I can clean up this interface more.  Instead of having to pass a response object and a flag saying if it's a paginated response, I could just `isinstance` the response to check if it's paginated.  Either that or do a duck type check for `build_full_result` and follow that code path if response has that object.

Thoughts?

awscli/customizations/emr/emrutils.py
```diff
@@ -248,13 +240,13 @@ def call_and_display_response(session, operation_name, parameters,
                               parsed_globals):
         cli_operation_caller = CLIOperationCaller(session)
         cli_operation_caller.invoke(
-            session.get_service('emr').get_operation(operation_name),
+            'emr', xform_name(operation_name),
```
jamesls: I don't think we want the `xform_name` call here. `invoke` uses `xform_name` internally:

```
    def invoke(self, service_name, operation_name, parameters, parsed_globals):
        client = self._session.create_client(
            service_name, region_name=parsed_globals.region,
            endpoint_url=parsed_globals.endpoint_url,
            verify=parsed_globals.verify_ssl)
        py_operation_name = xform_name(operation_name)
```

awscli/arguments.py
```diff
@@ -440,11 +438,11 @@ def add_to_params(self, parameters, value):
             parameters[self._serialized_name] = unpacked
 
     def _unpack_argument(self, value):
-        service_name = self.operation_object.service.endpoint_prefix
-        operation_name = xform_name(self.operation_object.name, '-')
+        service_name = self._operation_model.service_model.endpoint_prefix
```
kyleknap: Should we be using endpoint_prefix?

I have been using ``operation_model.service_model.service_name``. I was not sure to use ``endpoint_prefix`` because they do not always line up with their cli name.

awscli/clidriver.py
```diff
@@ -379,15 +366,17 @@ def __call__(self, args, parsed_globals):
 
     def _create_command_table(self):
         command_table = OrderedDict()
-        service_object = self._get_service_object()
-        for operation_object in service_object.operations:
-            cli_name = xform_name(operation_object.name, '-')
+        service_model = self._get_service_model()
```
kyleknap: I added a self.service_model property after you wrote this. We can probably use that now.

awscli/clidriver.py
```diff
@@ -402,9 +391,8 @@ def _add_lineage(self, command_table):
 
     def create_help_command(self):
         command_table = self._get_command_table()
-        service_object = self._get_service_object()
         return ServiceHelpCommand(session=self.session,
-                                  obj=service_object,
+                                  obj=self._get_service_model(),
```
kyleknap: same thing here with ``self.service_model``

awscli/clidriver.py
```diff
@@ -570,44 +562,50 @@ def _build_call_parameters(self, args, arg_table):
     def _unpack_arg(self, cli_argument, value):
         # Unpacks a commandline argument into a Python value by firing the
         # load-cli-arg.service-name.operation-name event.
-        session = self._service_object.session
-        service_name = self._service_object.endpoint_prefix
-        operation_name = xform_name(self._operation_object.name, '-')
+        session = self._session
+        service_name = self._operation_model.service_model.endpoint_prefix
```
kyleknap: Another place  ``endpoint_prefix`` is used. I am not sure how many, if any customizations would get broken if we changed it to ``service_name``

awscli/customizations/ec2decryptpassword.py
```diff
@@ -79,11 +82,11 @@ def add_to_params(self, parameters, value):
             path = os.path.expanduser(path)
             if os.path.isfile(path):
                 self._key_path = path
-                service_name = self._operation.service.endpoint_prefix
+                service_name = \
+                    self._operation_model.service_model.service_name
```
kyleknap: Here is a place that I used ``service_name`` instead of ``endpoint_prefix``

awscli/customizations/emr/createdefaultroles.py
```diff
@@ -202,19 +202,12 @@ def _run_main(self, parsed_args, parsed_globals):
...
         emrutils.display_response(
             self._session,
-            self._session.get_service('iam').get_operation('CreateRole'),
+            'create_role',
             self._construct_result(ec2_result, emr_result),
             parsed_globals)
 
         return 0
 
-    def _check_for_iam_endpoint(self, region, iam_endpoint):
```
kyleknap: This method did not really any sense to me especially, since that botocore always gives the same iam endpoint (no matter the region or even if no region was provided) and UnknownEndpointError will never be thrown.
 
Do you think it is alright that I just removed the check.

awscli/customizations/emr/describecluster.py
```diff
@@ -65,20 +63,12 @@ def _find_master_public_dns(self, cluster_id, parsed_globals):
...
 
-    def _call(self, operation_object, parameters, parsed_globals):
-        # We could get an error from get_endpoint() about not having
-        # a region configured.  Before this happens we want to check
-        # for credentials so we can give a good error message.
-        result = []
-        if not self._session.get_credentials():
-            raise NoCredentialsError()
-        endpoint = operation_object.service.get_endpoint(
+    def _call(self, session, operation_name, parameters, parsed_globals):
```
kyleknap: I wanted to remove this function entirely and just use ``emrutils.call()``, but this internal method gets mocked out in tests and is used to provide mocked responses...
I can go back and fix the tests but I was trying to make the minimal amount of changes necessary especially for the tests.

awscli/customizations/emr/emrutils.py
```diff
@@ -153,20 +154,13 @@ def check_empty_string_list(name, value):
...
-        # for credentials so we can give a good error message.
-        if session.get_credentials() is None:
-            raise NoCredentialsError()
-
-        endpoint = operation_object.service.get_endpoint(
-            region_name=region_name, endpoint_url=endpoint_url,
+        client = session.create_client(
+            'emr', region_name=region_name, endpoint_url=endpoint_url,
             verify=verify)
-        LOG.debug('Calling ' + str(operation_object) + ' with endpoint: ' +
```
kyleknap: I guess that I can keep the log message since I have ``client.meta.endpoint_url``. I feel knowing the endpoint_url is not really important since it is logged in botocore.

awscli/customizations/emr/emrutils.py
kyleknap: Good catch. ``CLIOperationCaller`` might have changed from when I made this change.

awscli/customizations/ec2decryptpassword.py
kyleknap: I looked at botocore. I think for here it should be ``endpoint_prefix`` to match up with botocore.

awscli/formatter.py
kyleknap: Cleaning it up would be nice. I thought the ``is_response_paginated`` interface was a little awkward when I was updating the emr commands that was using it.

I do not have a preference on what check you make. I feel that checking for ``build_full_response`` would be fine since you call the method within the if statement checking if you can paginate. That way you can avoid an extra import as well for ``isinstance``

tests/unit/test_clidriver.py
```diff
@@ -383,100 +414,6 @@ def inject_command_schema(self, command_table, session, **kwargs):
 
         command_table['foo'] = command
 
-    def test_aws_with_endpoint_url(self):
```
kyleknap: Do we have anything to replace these tests? If not, we could just make these in terms of clients. Also some of the tests that were removed involved clients.

kyleknap: Nevermind this comment. Just saw the code block below it.

kyleknap:
I felt that the PR looked pretty clean. There were just a couple comments I had involving some inconsistencies across separate commits that we made such as whether to use ``endpoint_prefix`` or ``service_name`` for events.

awscli/formatter.py
jamesls: Pushed a change that updates this.

awscli/customizations/ec2decryptpassword.py
jamesls: Updated.  Don't think it matters in this specific scenario (`endpoint_prefix == service_name` for ec2), but we should be consistent.

awscli/clidriver.py
jamesls: I'm not sure I'm really a fan of adding a `service_model` property to this class. I'd like to move in a general direction towards unifying the custom commands and the clidriver command code paths.  In order to do this, all the command objects should have a consistent interface.  For many custom commands the concept of "service model" doesn't really apply.

From what I can tell, this is only used in waiters.  Maybe there's a way to implement waiters without having to expose service_model in the ServiceCommand objects.

kyleknap: Yeah that sounds good with me to remove. I believe I added it because there was a ``service_object`` property so I figured a parallel was going to be needed later down the road. The best option to avoid using ``service_model`` is to check whether the command is an instance of a ``ServiceCommand``.

awscli/arguments.py
```diff
@@ -362,10 +363,6 @@ def __init__(self, name, argument_model, operation_object,
         :type argument_model: ``botocore.model.Shape``
         :param argument_model: The shape object that models the argument.
 
-        :type operation_object: ``botocore.operation.Operation``
-        :param operation_object: The operation object associated with
-            this object.
-
```
danielgtaylor: Missing docs for `operation_model` and `event_emitter`?

awscli/clidriver.py
```diff
@@ -559,7 +551,7 @@ def _build_call_parameters(self, args, arg_table):
         # args is an argparse.Namespace object so we're using vars()
         # so we can iterate over the parsed key/values.
         parsed_args = vars(args)
-        for arg_name, arg_object in arg_table.items():
+        for _, arg_object in arg_table.items():
```
danielgtaylor: Why not just do `for arg_object in arg_table.values():`?

awscli/customizations/emr/emrutils.py
```diff
@@ -153,20 +154,13 @@ def check_empty_string_list(name, value):
         raise exceptions.EmptyListError(param=name)
 
 
-def call(session, operation_object, parameters, region_name=None,
+def call(session, operation, parameters, region_name=None,
```
danielgtaylor: In many places in this PR we've moved from `operation` to `operation_model`. Going from `operation_object` to `operation` feels backwards to me. Can we name this `operation_name` instead for consistency?

awscli/customizations/globalargs.py
```diff
@@ -69,7 +69,5 @@ def _resolve_endpoint_url(value):
 def no_sign_request(parsed_args, session, **kwargs):
     if not parsed_args.sign_request:
         # In order to make signing disabled for all requests
-        # we need to set the signature_version to None for
-        # any service created.  This ensures that get_endpoint()
-        # will not look for auth.
+        # we need to use botocore's ``disable_signing()`` hanlder.
```
danielgtaylor: Typo: `hanlder` -> `handler`

awscli/customizations/service.py
```diff
@@ -1,84 +0,0 @@
...
-
-        self._operation = service.get_operation(name)
-
-    def __call__(self, **kwargs):
-        res, data = self._operation.call(self.endpoint, **kwargs)
-
-        return data
-
-
-class Service(object):
```
danielgtaylor: Hooray! :raised_hands: Glad to see this one go away.

danielgtaylor:
Very minor comments from me. I am also curious about the `endpoint_prefix` issue @kyleknap brings up. Also, any idea why coverage is down slightly? Aside from those, LGTM :ship:-it!

awscli/clidriver.py
jamesls: After looking into this further, I'm not sure if we can remove the service_model entirely.  Even if we type check for a ServiceCommand, we need to know what the service name is in order to retrieve the correct service model.  However, because a command can be renamed, we can't rely on the command name being the same as what we use to `get_sevice_model`.

Have I overlooked anything?

jamesls:
@danielgtaylor I believe I've incorporated all your feedback.

jamesls:
Ok, I'd like to get a list of all the remaining work here.  I believe I've incorporated all the feedback except for a few outstanding issues:

- [ ] Figure out what (if anything) to do with `service_model`
- [ ] Figure out endpoint_prefix vs. service_name

Anything else I've missed? cc @kyleknap @danielgtaylor

awscli/arguments.py
jamesls: Actually, for this specific scenario, I'd be inclined to keep this as is for now.  The previous code used `endpoint_prefix` so to keep the changes to a minimum for the client migration, I'd be ok with keeping it the same for now.

I certainly think it would be worth standardizing on one vs. the other, but I don't think that necessarily has to happen with this PR.

awscli/clidriver.py
kyleknap: Yeah, that looks to be true. We *need* the service_name (like the one you use to create a client) in order to get the waiter model. We cannot rely on the ``name`` property. I cannot think of any way around it. Extracting the service name from the event name is unreliable as well.

awscli/arguments.py
kyleknap: That is fine with me. I agree minimalistic changes for this PR is ideal. We can fix it after it gets merged since it will require a botocore PR as well.

kyleknap:
@jamesls 

I made replies. For the first bullet point, I do not think there is much we can do about ``service_model``. So I say we leave it. For the second point, it is fine to leave the ``endpoint_prefix`` as is to keep changes minimal and we can fix it later down the road.

I cannot think of any other outstanding points other than the one where ``xform_name`` is used twice:
https://github.com/aws/aws-cli/pull/1220/files#r26405927

Otherwise, looks good :ship:

jamesls:
Hmm, it's a little concerning to me that even after changing the code I didn't get a failing test.  It sounds like there's nothing checking this part of the code.  I would have expected a test failure either before/after the change.  I'll look into possibly getting a test in place.

jamesls:
I retract my previous statement :).  @kyleknap points out that the `xform_name(name)` == `xform_name(xform_name(name))` so this is just an unnecessary call to `xform_name`, but it's doesn't actually break any functionality (hence why no test was failing).

kyleknap:
@jamesls 
One more item. We need to re-add this logic: https://github.com/aws/aws-cli/pull/1220/files#diff-3729be6e026ab99897911ab34e1c4dd3L211 back to one of the emr commands after talking to some of the contributors to the emr commands.

jamesls:
Ok, I've added the method back.  I believe that's all the feedback.  Anything else?

danielgtaylor:
LGTM :ship:-it

kyleknap:
I do not think there is anything more. Thanks for seeing it through! :ship:

PR #1223, opened 2015-03-17
Fix issue with configservice subscribe command

kyleknap:
When checking to see if a bucket exists, we assumed that an exception meant that the bucket did not exist when using HeadBucket. In reality, the error could be a 403, and the user wanted to use that bucket but did not have the proper permissions to run a HEAD object on it. Also expanded the output to inform users if a new bucket/topic is being used or an existing one. This helps notify users of what resources they created and/or are using when subscribing to AWS Config.

cc @jamesls @danielgtaylor

### DIFF
--- a/awscli/customizations/configservice/subscribe.py
+++ b/awscli/customizations/configservice/subscribe.py
@@ -13,6 +13,8 @@
 import json
 import sys
 
+from botocore.exceptions import ClientError
+
 from awscli.customizations.commands import BasicCommand
 from awscli.customizations.s3.utils import find_bucket_key
 
@@ -138,16 +140,25 @@ class S3BucketHelper(object):
         bucket_exists = self._check_bucket_exists(bucket)
         if not bucket_exists:
             self._create_bucket(bucket)
+            sys.stdout.write('Using new S3 bucket: %s\n' % bucket)
+        else:
+            sys.stdout.write('Using existing S3 bucket: %s\n' % bucket)
         return bucket, key
 
     def _check_bucket_exists(self, bucket):
         bucket_exists = True
+        self._s3_client.meta.events.unregister(
+            'after-call',
+            unique_id='awscli-error-handler')
         try:
             # See if the bucket exists by running a head bucket
             self._s3_client.head_bucket(Bucket=bucket)
-        except Exception:
-            # If a client error is thrown than the bucket does not exist.
-            bucket_exists = False
+        except ClientError as e:
+            # If a client error is thrown. Check that it was a 404 error.
+            # If it was a 404 error, than the bucket does not exist.
+            error_code = int(e.response['Error']['Code'])
+            if error_code == 404:
+                bucket_exists = False
         return bucket_exists
 
     def _create_bucket(self, bucket):
@@ -171,6 +182,9 @@ class SNSTopicHelper(object):
         if not self._check_is_arn(sns_topic):
             response = self._sns_client.create_topic(Name=sns_topic)
             sns_topic_arn = response['TopicArn']
+            sys.stdout.write('Using new SNS topic: %s\n' % sns_topic_arn)
+        else:
+            sys.stdout.write('Using existing SNS topic: %s\n' % sns_topic_arn)
         return sns_topic_arn
 
     def _check_is_arn(self, sns_topic):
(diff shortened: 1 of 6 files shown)


awscli/customizations/configservice/subscribe.py
```diff
@@ -138,16 +140,25 @@ def prepare_bucket(self, s3_path):
...
             self._s3_client.head_bucket(Bucket=bucket)
-        except Exception:
-            # If a client error is thrown than the bucket does not exist.
-            bucket_exists = False
+        except ClientError as e:
+            # If a client error is thrown. Check that it was a 404 error.
+            # If it was a 404 error, than the bucket does not exist.
+            error_code = int(e.response['Error']['Code'])
+            if error_code == 404:
+                bucket_exists = False
```
danielgtaylor: We have very similar code in our [CloudTrail customization](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudtrail.py#L237-L244). Maybe it's worth making a utility function that can be used across customizations?

danielgtaylor:
I think this is a good change, just wanted to mention the similarities between this and CloudTrail.

I'm also curious what the errors look like if I select a bucket I don't own and have zero rights to. The bucket won't be created again, but some other command will likely fail, right? The fix in this PR is really only for the case where I can write to the bucket but cannot run a HEAD operation against it. Is a rights check performed server-side?

In the CloudTrail customization we put the burden on the user by having them provide [one of two mutually exclusive parameters](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudtrail.py#L63-L66) that specify whether to create a new bucket or use an existing one. Now I'm not saying that's the best approach, but it makes the HEAD check defensive rather than part of the core logic of the command.

awscli/customizations/configservice/subscribe.py
jamesls: Are we sure this works?  In the cloudtrail customization I had to remove the error handler to get the client exceptions to propogate.  I would expect we'd need something here as well right?

kyleknap: @jamesls 
I tested it and I have had confirmation it works. I had to disable the error handler as well:
https://github.com/aws/aws-cli/pull/1223/files#diff-9acfa1e377b87026a3322468d0755a93R150

@danielgtaylor 
As to consolidating the logic, I could but it is not completely the same logic. I actually think it may be a potential bug in the cloudtrail commands as you handle 404 and 403 errors the same when they are not really the same relative to determining if a bucket exists. Is my concern correct?

Let me know.

jamesls: Oh wow, I completely missed L150-152 sorry about that.

jamesls:
:shipit:  Looks good to me.  However I'll let you and Daniel hash out the details about the cloudtrail commonality.  I'm neutral on it.

kyleknap:
@danielgtaylor 

Actually tested what happens if you select a bucket with zero rights. The command actually works cleanly through. It looks like AWS Config does not validate whether if you have rights to the bucket when you create the delivery channel. I have yet to figure out where errors propagate when you do not have permissions to a bucket, but I believe it happens when you deliver snapshots of your AWS resources. The status of the delivery will show a fail.

That is why I decided to add the print statement on whether a previous bucket is being used or a new one is being created. So users can catch themselves (the resources created/being used is as expected) before errors show up when you check to see if the snapshots were delivered successfully. Given that it may be better, if we used the same style of the cloudtrail commands, but it may be too late to change the API for the commands so that may be the next best option.

I will update the PR with what I find. Let me know if you have any other questions.

kyleknap:
UPDATE:

Actually I just tested it out again, and actually AWS Config does validate if you have permissions to the bucket upon creating a delivery channel. You will get the following error:
```
A client error (InsufficientDeliveryPolicyException) occurred when calling the PutDeliveryChannel operation: Insufficient delivery policy to s3 bucket: mybucket, unable to write to bucket, provided s3 key prefix is 'null'.
```
So I think everything is fine there.

kyleknap:
@danielgtaylor 

I updated the code to consolidate into the helper method ``s3_bucket_exists()`` in ``utils.py``. Should be good to look at again.

danielgtaylor:
LGTM :ship:-it!

PR #1228, opened 2015-03-19
Improved docs for ICMP port ranges

kyleknap:
Here is the api doc link to confirm: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AuthorizeSecurityGroupIngress.html

The reason I updated the docs was because it was not clear how to specify both ICMP code and type using ``--port``. Even the ``-1`` comment involving ICMP was a little ambiguous so I tried to clear that up as well.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

Issue #304, opened 2013-09-04
S3 prefix is not obvious


kaihendry:
The experience is again jarring. The UX doesn't guide you to use `s3://`

    $ aws s3 ls

           CreationTime Bucket
           ------------ ------
    2013-08-30 13:57:23 bbackup
    $ aws s3 ls bbackup
    usage: aws s3 ls <S3Path> or NONE
    Error: Invalid argument type
    sg:~/btsync$ aws s3 ls s3://bbackup

    Bucket: bbackup
    Prefix: 

          LastWriteTime     Length Name
          -------------     ------ ----

RichardBronosky:
@jamesls,
This seems to be resolved with `aws s3 ls` but is still an issue with `aws s3 cp`. For the former, the s3 [URI scheme] is optional. For the later it is not. I suggest putting `<S3Path>` in the usage statements where it is optional and `<S3Uri>` where the scheme is required.

Also, THANK YOU for https://www.youtube.com/watch?v=qiPt1NoyZm0 I go back and rewatch it every few weeks. There is SO MUCH to learn in there.

[URI scheme]: http://en.wikipedia.org/wiki/URI_scheme

RichardBronosky:
I don't know if https://github.com/aws/aws-cli/pull/1232 is too extreme of a change for you. Before grepping the source for s3path I didn't even know that `awscli/examples/s3/_concepts.rst` existed. If this change messes up a bunch of other documentation, let me know and I'll consider a different way to improve the UX.

The main thing you get from this is to go from:

    aws s3 cp mybucket/test.txt test2.txt

    usage: aws s3 cp <LocalPath> <S3Path> or <S3Path> <LocalPath> or <S3Path> <S3Path>
    Error: Invalid argument type

To:

    aws s3 cp mybucket/test.txt test2.txt

    usage: aws s3 cp <LocalPath> <S3Uri> or <S3Uri> <LocalPath> or <S3Uri> <S3Uri>
    Error: Invalid argument type

I think that better communicates what is invalid about the arguments.

mtdowling:
`aws s3 ls` now allows the `s3://` scheme to be optional. I've taken a look at your PR, @RichardBronosky. It looks great. I'm going to rebase it and merge it in.

SirR4T:
`aws s3 ls` now indeed allows the `s3://` scheme to be optional, but `aws s3 rm` doesn't yet. Should that be raised as a different issue?

PR #1232, opened 2015-03-21
change references to S3Path to S3Uri

RichardBronosky:
Addresses https://github.com/aws/aws-cli/issues/304

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -303,7 +303,7 @@ class ListCommand(S3Command):
     DESCRIPTION = ("List S3 objects and common prefixes under a prefix or "
                    "all S3 buckets. Note that the --output argument "
                    "is ignored for this command.")
-    USAGE = "<S3Path> or NONE"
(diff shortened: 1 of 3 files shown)


jamesls:
Not opposed to this, but I'm not seeing where we document what S3Path vs. S3Uri is.  I think somewhere in `_concepts.rst` we should explain what `S3Path` is.

Thanks for the PR.

RichardBronosky:
Actually, my PR contradicts my original https://github.com/aws/aws-cli/issues/304#issuecomment-84421513 that was made before looking at the code. Once I got in there and saw how rare the S3Path string was and how infrequently the scheme is optional, I changed course. I did add this note to `_concepts.rst` to explain the optionality in the case of `ls`. https://github.com/RichardBronosky/aws-cli/commit/86dd0400cae825683318e42c13c1bdf11692c482#diff-5216f19579f981938822963819dcc737R15

I thought it was "`_concept`ually" simpler to eliminate S3Path entirely. Do you disagree?

PR #1237, opened 2015-03-24
Add unit test for cloudfront command

jamesls:
This will prevent future regressions against basic
cloudfront commands.

cc @kyleknap @danielgtaylor

### DIFF
--- a/tests/unit/cloudfront/test_get_distribution_config.py
+++ b/tests/unit/cloudfront/test_get_distribution_config.py
@@ -1,0 +1,25 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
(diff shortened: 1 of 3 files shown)


tests/unit/cloudfront/test_get_distribution_config.py
```diff
@@ -0,0 +1,25 @@
...
+# language governing permissions and limitations under the License.
+from awscli.testutils import BaseAWSPreviewCommandParamsTest as \
+    BaseAWSCommandParamsTest
+
+
+class TestListDistributions(BaseAWSCommandParamsTest):
+
+    prefix = 'cloudfront get-distribution-config'
+
+    def test_list_distributions(self):
```
kyleknap: whoops. I messed up the naming for the class and the test method. Should be ``TestGetDistributionConfig`` and ``test_get_distribtuion_config``

kyleknap:
Had a small comment after reviewing the test I wrote. Otherwise, :ship:

tests/unit/cloudfront/test_get_distribution_config.py
jamesls: Also, what about the BaseAWSPreviewCOmmandParamsTest?

The test actually passes for me as is just using the BaseAWSCommandParamsTest.

jamesls: Nevermind completely misread the import statement.

kyleknap: Yeah if you do not configure cloudfront to preview in the config and use BaseAWSCommandParamsTest, it fails

PR #1239, opened 2015-03-24
Fix issue 1238 for unknown encoding on Mac OS X

freimer:


jamesls:
Hi, thanks for the pull request.  Can you give me a little more background on this issue?

* What was the error you were seeing before this fix?  Can you give a set of steps that I could follow that demonstrate the issue you were seeing?
* How does setting `do_setlocale` to `False` fix this issue?
* The docs for `getpreferredencoding` say that: *"On some systems, it is necessary to invoke setlocale() to obtain the user preferences"*.  By setting `do_setlocale` to `False`, does this have potential to affect users on non Mac OS-X platforms?

freimer:
On a Mac, encoding is set to a blank string:

```
DRFITSMAC11142:cft jmfgqlh$               python
Python 2.7.9 (v2.7.9:648dcafa7e5f, Dec 10 2014, 10:10:46) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
''
>>> locale.getpreferredencoding(do_setlocale=False)
'US-ASCII'
>>> quit()
You have new mail in /var/mail/jmfgqlh
DRFITSMAC11142:cft jmfgqlh$ echo $PYTHONIOENCODING
utf-8
DRFITSMAC11142:cft jmfgqlh$
```

So, I get things like:

```
DRFITSMAC11142:cft jmfgqlh$ aws cloudformation validate-template --template-body=file://tmds-manager.json

unknown encoding: 
DRFITSMAC11142:cft jmfgqlh$
```

Yes, this will effect other systems:

```
python
Python 2.7.5 (default, Feb 11 2014, 07:46:25) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-13)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
'UTF-8'
>>> locale.getpreferredencoding(do_setlocale=False)
'ANSI_X3.4-1968'
>>> quit()
```

So, I'll wrap it in a test to see if encoding is a blank string, and only re-call getpreferredencoding if it is blank...

jamesls:
Thanks for clarifying.  Out of curiosity, what's the output of running the `locale` command on your Mac?  Not opposed to the fix, just curious why you're getting an empty string for `getpreferredencoding()` in the first place.

freimer:
DRFITSMAC11142:_grains jmfgqlh$ locale
LANG="en_US"
LC_COLLATE="en_US"
LC_CTYPE="en_US"
LC_MESSAGES="en_US"
LC_MONETARY="en_US"
LC_NUMERIC="en_US"
LC_TIME="en_US"
LC_ALL=

kyleknap:
@freimer 

I was able to fix the issue by changing my locale settings:
```
export LC_ALL='en_US.UTF-8'
```
Does/would that work for you? We still have to do some digging into why explicitly setting ``do_setlocale=False`` causes a different output. Would like to know this before merging the PR.

jamesls:
Closing due to inactivity.  Let us know if @kyleknap's suggestion does not work for you and I'll reopen.

PR #1242, opened 2015-03-25
Tag slow integ tests

jamesls:
1. Tag slowest integ tests, similar to what we did for boto3.  This shaves about 10 min off of our integration tests
2. Sleep for 1 second between polling the process in the ctrl+c test.

cc @kyleknap @danielgtaylor


kyleknap:
Looks good. We should make a todo comment for removing the 1 second sleep so we don't forget about eventually trying to get it removed. :ship:

Issue #1178, opened 2015-02-24
aws s3 cp streaming download hangs when pipe breaks


tlunter:
When using `aws s3 cp s3://some-bucket/some-key -` and the downstream pipe dies due to one error or another, aws-cli hangs.

The exception is raised by the IOWriterThread, but is not checked by the Executor.  New jobs are still being submitted to the executor to download files, but they get held up: [here](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/tasks.py#L420) and [here](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/tasks.py#L416).

In the first, the queue has a max size of 20 ([here](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/s3handler.py#L43)).  In the second, nothing is being added to the queue, so the current stream part number is never going to be incremented.  The program then halts and nothing happens until it is forcefully killed.  A keyboard interrupt doesn't even kill the process.

The `MultipartDownloadContext` does have a cancel method to prevent the workers from continuing to wait for their turn, but workers stuck at `self._io_queue.put` will remain there.

kyleknap:
Thanks for looking into this!

I can confirm that if a downstream pipe breaks that the s3 streaming process does hang. However I have not been able to reproduce the keyboard interrupt issue. When I hit keyboard interrupt, it takes roughly 15 seconds to clean up (e.g. the length "Cleaning up. Please wait.." is being displayed).

For replication purposes here is the command, I was running:
```
aws s3 cp s3://mybucket/bigfile - | python some_process.py
```

Here is the ``some_process.py`` script that I wrote:
```
import sys

def main():
    total = 0
    kb = 1024
    mb = 1024 ** 2
    chunk = sys.stdin.read(kb)
    while chunk:
        total += len(chunk)
        if total > mb:
            raise Exception
        chunk = sys.stdin.read(kb)


if __name__ == "__main__":
    main()
```

This will cause the following to be displayed to screen:
```
Traceback (most recent call last):
  File "some_process.py", line 16, in <module>
    main()
  File "some_process.py", line 11, in main
    raise Exception
Exception
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 808, in __bootstrap_inner
    self.run()
[...]

PR #1245, opened 2015-03-25
IOWriterThread should continue even if a task fails

jamesls:
Otherwise this will cause a hang in the CLI because the
executor is waiting for the IO thread to process of all
its task items before exiting.

This actually fixes an integration test that would periodically
fail (the ctrl+c test).  I've also added an additional unit test
for this case.

This is actually how all the other threads (the task threads and the
result thread) work.  They catch exception and continue on to the
next task.

While this fixes the failing integration test, there are still additional
enhancements we can make.  We don't have a feedback mechanism
from within the IOthread to let other parts of the system know about
failed write tasks.  So while this technically "fixes" #1178 in the sense
that the code will no longer hang, it will still download all the content
from an S3 object.  We can improve this by failing fast as soon as we
have a streaming io task fail.  This will be a larger change and a
separate pull request.

cc @kyleknap @danielgtaylor

### DIFF
--- a/awscli/customizations/s3/executor.py
+++ b/awscli/customizations/s3/executor.py
@@ -39,7 +39,7 @@ class Executor(object):
     ``Executor``runs is a worker and a print thread.
     """
     STANDARD_PRIORITY = 11
-    IMMEDIATE_PRIORITY= 1
+    IMMEDIATE_PRIORITY = 1
 
     def __init__(self, num_threads, result_queue, quiet,
                  only_show_errors, max_queue_size, write_queue):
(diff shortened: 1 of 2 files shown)


kyleknap:
Looks good. :ship:

kyleknap:
Wait. Can you remove the second sleep on the cntrl-c test as well if that is fixed when you merge?

jamesls:
Do you mean this one (https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L335) or this one (https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L344).  The latter was needed to avoid excessive CPU on the system as well as it's basically a tight polling loop with no delay previously.

kyleknap:
Yeah I was referring to the one that you just added. Given the reason, I am good with keeping it.

PR #1262, opened 2015-04-02
Adding -t option to emr ssh command when using certificate for authentication

treff7es:
Cursor keys did not worked when running an interactive command with 'emr ssh' command like pig. This was because -t option was missing from the ssh command if authenticating with certificate although it was present when authenticating without certicate based on the code. 
I have added the -t option to ssh command when using certificate.

Here is the generated original command without my modification:
ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=10 -i certificate.pem -t hadoop@host.compute-1.amazonaws.com pig

The working comand with the modification:
ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=10 -i certificate.pem -t hadoop@host.compute-1.amazonaws.com -t pig

### DIFF
--- a/awscli/customizations/emr/ssh.py
+++ b/awscli/customizations/emr/ssh.py
@@ -91,7 +91,7 @@ class SSH(Command):
             command = ['ssh', '-o', 'StrictHostKeyChecking=no', '-o',
                        'ServerAliveInterval=10', '-i',
                        parsed_args.key_pair_file, constants.SSH_USER +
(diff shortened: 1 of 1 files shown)


treff7es:
Hi,
Do you have any issue with this small bugfix where I can help?
It's a small bugfix but it would help to step into the command history with cursor keys.

JordonPhillips:
Thanks for the PR! ⛵

PR #1277, opened 2015-04-10
bug fix: _min_version_required was not comparing version numbers correct...

santana:
converts version strings to vectors of integers, gets a comparison vector from them and then computes a single value to determine which one is higher.


danielgtaylor:
Closing duplicate of #1278.

santana:
Hi Daniel,

actually, #1278 is the good one (sorry, I didn't figure out how to remove #1277), because it adds some boilerplate to make it work on any version of Python. #1277 doesn't work in Python 3.

PR #1278, opened 2015-04-10
bug fix: _min_version_required was not comparing version numbers correctly

santana:
converts version strings to vectors of integers, gets a comparison vector from them and then computes a single value to determine which one is higher.

### DIFF
--- a/scripts/make-bundle
+++ b/scripts/make-bundle
@@ -12,6 +12,7 @@ interface for those not familiar with the python
 ecosystem.
 
 """
+import functools
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


scripts/make-bundle
```diff
@@ -155,13 +156,14 @@ def verify_preconditions():
...
-    min_split = min_version.split('.')
-    actual_split = actual_version.split('.')
-    for minimum, actual in zip(min_split, actual_split):
-        if int(actual) < int(minimum):
-            raise ValueError("%s requires at least version %s, but "
-                             "version %s was found." % (name, min_version,
-                                                        actual_version))
+    min_split = map(int, min_version.split('.'))
+    actual_split = map(int, actual_version.split('.'))
+    comp = map(cmp, actual_split, min_split)
```
jamesls: We don't need the cmp and reduce calls here.  The default __gt__ for lists does what we want:

```
>>> [1,7,8] < [1,7,9]
True
>>> [1,7,8] < [1,7,7]
False
>>> [1,10,8] < [1,9,7]
False
>>> [1,10,8] < [1,10,10]
True
```

jamesls:
Went ahead and remove the cmp/reduce calls and merged.  Thanks for the bug fix!

santana:
Cool :)

You might want you remove the import functools line too, since it was added for the reduce function only.

jamesls:
Good catch, I'll get that updated, thanks.

PR #1286, opened 2015-04-14
Black list --predict-endpoint for paramfiles

kyleknap:
Before, the CLI was trying to pull the contents from the url specified. In reality, we need to use the literal value when we make the request.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #1291, opened 2015-04-15
Fix kms docs for create grant

kyleknap:
The improper casing was being used as the docs were refering to actual api operation names not the CLI names. Here is a link to the docs: http://docs.aws.amazon.com/cli/latest/reference/kms/create-grant.html. It is especially notable at the ``--operations`` parameter.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #1292, opened 2015-04-16
Update wait command docs

kyleknap:
Docs were updated to include how timeouts happen. There has been some confusion from users on how waiter timeouts. Here is an example of how it would look for ``s3api wait bucket-exists``:

```
NAME
       bucket-exists -

DESCRIPTION
       Wait  until  200 response is received when polling with head-bucket. It
       will poll every 5 seconds until a successful state has been reached. An
       error is thrown after 20 failed checks.
```

cc @jamesls

### DIFF
--- a/awscli/customizations/waiters.py
+++ b/awscli/customizations/waiters.py
@@ -159,6 +159,8 @@ class WaiterStateDocBuilder(object):
             # Include what operation is being used.
             description += self._build_operation_description(
                 self._waiter_config.operation)
+        description += self._build_polling_description(
+            self._waiter_config.delay, self._waiter_config.max_attempts)
         return description
 
     def _build_success_description(self, acceptor):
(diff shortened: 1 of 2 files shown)


awscli/customizations/waiters.py
```diff
@@ -181,6 +183,13 @@ def _build_operation_description(self, operation):
         operation_name = xform_name(operation).replace('_', '-')
         return u'when polling with ``%s``.' % operation_name
 
+    def _build_polling_description(self, delay, max_attempts):
+        description = (
+            ' It will poll every %s seconds until a successful state '
+            'has been reached. An error is thrown after %s failed checks.'
```
jamesls: Can we reword this to be more cli-centric.  This will exit with an RC of <some rc> after xx many failed checks.  The CLI doesn't really throw exceptions.

jamesls:
:shipit: Just had one small comment.

awscli/customizations/waiters.py
kyleknap: sounds good

kyleknap:
Updated PR to mention the return code. Merging.

PR #1293, opened 2015-04-16
Update doc for route53 create-hosted-zone

kyleknap:
There has been some confusion where user's try to make a private hosted zone using the ``--hosted-zone-config`` parameter for ``create-hosted-zone`` command. If you do specify it in the command, you get the following error:

```
aws route53 create-hosted-zone --name cli123.com --caller-reference foo --hosted-zone-config Comment=bar,PrivateZone=true

A client error (InvalidInput) occurred when calling the CreateHostedZone operation: Invalid XML ; cvc-complex-type.2.4.d: Invalid content was found starting with element 'PrivateZone'. No child element is expected at this point. 
```

The member ``PrivateZone`` has the following doc string: 
```
"<p>A value that indicates whether this is a private hosted zone. The value is returned in the response; do not specify it in the request.</p>"
```
The issue is that we do not document the members of structures in the docs so users are not exposed to this. The long term solution would to be expose these descriptions, but that would be a lot more work. I also considered popping off the shape from the model, but I am really hesitant to be modifying model shapes. So I decided to just add it to the parameter description.

cc @jamesls

### DIFF
--- a/awscli/customizations/route53.py
+++ b/awscli/customizations/route53.py
@@ -13,8 +13,9 @@
 
 
 def register_create_hosted_zone_doc_fix(cli):
-    # Docs may actually refer to actual api name (not the CLI command).
-    # In that case we want to remove the translation map.
+    # We can remove this customization once we begin documenting
+    # members of complex parameters because the member's docstring
+    # has the necessary documentation.
     cli.register(
         'doc-option.route53.create-hosted-zone.hosted-zone-config',
         add_private_zone_note)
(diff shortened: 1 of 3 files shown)


awscli/customizations/route53.py
```diff
@@ -0,0 +1,29 @@
...
+        add_private_zone_note)
+
+
+def add_private_zone_note(help_command, **kwargs):
+    note = (
+        '<p>Note do <b>not</b> include <code>PrivateZone</code> in this '
+        'input structure. Its value is returned in the output to the command.'
+        '</p>'
+    )
+    help_command.doc.include_doc_string(note)
```
jamesls: Can we add a comment that we can remove this when we start documeting struct members?

jamesls:
:shipit:

awscli/customizations/route53.py
kyleknap: yeah that's a good idea.

PR #1296, opened 2015-04-17
Validate zip file param validation for AWS Lambda

jamesls:
This improves the error message when the customer does not provide a `fileb://` prefix for the zip file argument.  Now you get:

```
# Omitting any file/fileb prefix:
$ aws lambda update-function-code --function-name helloworld --zip-file tests.zip

--zip-file must be a file with the fileb:// prefix.
Example usage:  --zip-file fileb://path/to/file.zip

# Using file:// instead of fileb://

$ aws lambda update-function-code --function-name helloworld --zip-file file://tests.zip

Error parsing parameter '--zip-file': Unable to load paramfile (tests.zip), text contents could not be decoded.  If this is a binary file, please use the fileb:// prefix instead of the file:// prefix.

```

The same logic is applied to the `create-function` operation as well.

cc @kyleknap

### DIFF
--- a/awscli/customizations/awslambda.py
+++ b/awscli/customizations/awslambda.py
@@ -34,7 +34,7 @@ def _flatten_code_argument(argument_table, **kwargs):
 
 class ZipFileArgument(CustomArgument):
     ERROR_MSG = (
-        "--zip-file does not contain zip file content.\n"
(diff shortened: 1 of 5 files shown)


kyleknap:
Looks good. :ship:

PR #1310, opened 2015-04-27
API updates for botocore loader changes

jamesls:
All in the test code only.

Note that this will fail on travis until the botocore loader branch is merged to develop.
cc @kyleknap


kyleknap:
LGTM :ship:

Issue #1241, opened 2015-03-25
Examples section on S3 are rendering twice


jamesls:
Running "aws s3 cp help" will show the "Examples" section rendered twice.

I checked some of the data driven commands, and the examples section only appears once so this appears to only be for the "s3" commands.

kyleknap:
Fixed thanks to https://github.com/aws/aws-cli/pull/1312. Closing issue.

PR #1312, opened 2015-04-29
Remove duplicate example data input

carsonmcdonald:
I believe this will clear up issue #1241. The removed lines are the instances where the sub-commands are reading in the example docs after they have already been read in via this event hook in handlers.py:

```
event_handlers.register('doc-examples.*.*', add_examples)
```

### DIFF
--- a/awscli/customizations/emr/addsteps.py
+++ b/awscli/customizations/emr/addsteps.py
@@ -33,7 +33,6 @@ class AddSteps(Command):
          'help_text': helptext.STEPS
          }
     ]
-    EXAMPLES = BasicCommand.FROM_FILE('emr', 'add-steps.rst')
 
     def _run_main_command(self, parsed_args, parsed_globals):
         parsed_steps = parsed_args.steps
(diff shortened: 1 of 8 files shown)


kyleknap:
Thank you for the pull request. That is definitely what the issue was. I pulled your PR down as well and tried it out. Looks good.

However, typically we like to add a test to accompany the fix. Would it be possible to add tests for the commands that you fixed?

For me, it would be fine if you just add the emr commands and the one opsworks command you touched to this dictionary: https://github.com/aws/aws-cli/blob/develop/tests/unit/docs/test_examples.py#L30 to ensure that the example section is caught by the handler. If you want to go further, it would be awesome if you could add a test to make sure that the Examples section is not rendered twice in any of the commands you fixed.

carsonmcdonald:
Let me dig into what that test_examples test does. I assume it doesn't exercise everything because it should have caught this issue since it tests the s3 command examples and that was where someone noticed the duplication.

carsonmcdonald:
@kyleknap I added a count test to verify that it only shows up once and added the changed commands to the commands tested. Seem like enough?

kyleknap:
Thanks! Looks great. Merging.

Issue #969, opened 2014-10-25
pem and cer file names must end with '.pem' and '.cer' -- why?


tilusnet:
Hi,

For a number of times I had my pem files rejected while requesting ssh access until I realised that the code requires them to have a ".pem" extension.

Is this really necessary?

My team members have already registered their private keys with AWS and this artificially looking requirement would make the migration to the aws-cli client somewhat cumbersome.

kyleknap:
I believe this PR fixes this issue: https://github.com/aws/aws-cli/pull/1317

Resolving issue. If that is not the issue you are referring to, please reopen.

PR #1318, opened 2015-05-05
Remove key file format restrictions for SSH

kyleknap:
We restricted key file formats to pem and cer for use with ssh and
scp. This change removes it as customers have varied key file formats.


kyleknap:
I believe this fixes: https://github.com/aws/aws-cli/issues/969 as well

PR #1336, opened 2015-05-20
Fix switch of imports in botocore clients-only

kyleknap:
Fixed another test failure for clients-only branch. Note that this will not pass till clients-only branch is merged.

cc @jamesls


jamesls:
:shipit: once the botocore changes get merged.

PR #1345, opened 2015-05-27
Document preview services

jamesls:
This changes the behavior of documentation, help commands, and preview services.

* Always generate html docs for preview services.  Before, if a service was marked preview the html docs were never generated.
* Any preview service documentation now has a "this is a preview service" disclaimer added to the description.
* You don't need to opt into the preview in order to see docs.  Before something like `aws cloudfront help` would still require you to opt into the preview.  Now, only the actual service commands require you to opt in.

Because of the way our doc system works, it's easy to add to the *end* of the description, but not so easy to add to the beginning of the documentation.  Here's how this looks like with this change:

screen shot 2015-05-26 at 6 22 27 pm

cc @kyleknap


awscli/customizations/preview.py
```diff
@@ -70,15 +55,35 @@ def mark_as_preview(command_table, session, **kwargs):
...
-            preview_service, PREVIEW_SERVICES[preview_service])
+            is_enabled = True
+        original_command = command_table[preview_service]
+        preview_cls = type(
+            'PreviewCommand',
+            (PreviewModeCommandMixin, original_command.__class__), {})
+        command_table[preview_service] = preview_cls(
+            cli_name=original_command.name,
+            session=session,
+            service_name=original_command._service_name,
```
kyleknap: I think you would would be able to get this from ``original_command.service_model.service_name``

awscli/customizations/preview.py
```diff
@@ -70,15 +55,35 @@ def mark_as_preview(command_table, session, **kwargs):
...
+        # description in the docs to say that this is a preview service.
+        session.get_component('event_emitter').register_last(
+            'doc-description.%s' % preview_service,
+            update_description_with_preview)
+
+
+def update_description_with_preview(help_command, **kwargs):
+    style = help_command.doc.style
+    style.start_note()
+    style.bold("The service is currently in preview.")
```
kyleknap: Lets include instructions on how to enable a command out of a preview as well.

tests/unit/customizations/test_preview.py
```diff
@@ -40,14 +40,33 @@ def test_invoke_preview_mode_service(self):
...
+        rc = self.driver.main('cloudfront list-distributions'.split())
+        self.assertEqual(rc, 1)
+        self.assertIn(preview.PreviewModeCommandMixin.HELP_SNIPPET,
+                      self.stderr.getvalue())
+
+    def test_preview_service_enabled_makes_call(self):
         self.full_config['preview'] = {'cloudfront': 'true'}
+        # Registering this handler will prevent the actual service
+        # call to cloudfront.
+        self.driver.session.register('calling-command.cloudfront',
```
kyleknap: Seems a little hacky to me (requires knowledge of the event system and these are more prone to changes). I would prefer to mock out the endpoint or use ``BaseAWSCommandParamsTest``.

kyleknap:
Implementation looks good. Had a few minor comments. :ship:

jamesls:
@kyleknap feedback incorporated, please take another look.

kyleknap:
Thanks. :ship:

PR #1352, opened 2015-05-29
Make clidriver.main() accept arguments

nkalra-t:
Allows for arguments to be passed to a clidriver instance when
instantiated from inside a python script.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -43,9 +43,9 @@ LOG_FORMAT = (
     '%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s')
 
 
-def main():
+def main(args=None):
     driver = create_clidriver()
-    return driver.main()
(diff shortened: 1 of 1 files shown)


kyleknap:
Thanks for the pull request.

Quick question to get a better understanding of your use case. Would something like this work for you?:
```
import awscli.clidriver

driver = awscli.clidriver.create_clidriver()
driver.main(args)
```
Could you elaborate on how you plan to use it in your python script?

nkalra-t:
Yes, that would work for us.

We're currently using awscli in a shell script to sync a folder with build targets to s3. We're in the process of switching the build scripts over to python and since boto doesn't have a native sync command, we would like to use that of awscli without needing to call subprocess.Popen().

nkalra-t:
Sorry for the late response but yes, the above does work for us.

Thanks for your help!

Issue #647, opened 2014-02-13
Upload wheels to PyPI


Julian:
Hey.

It'd be lovely if there was a wheel built + uploaded to PyPI. There's some more information about what wheels are on [pythonwheels.com](http://pythonwheels.com/) if you'd like some info, the tl;dr is just that running:

    python setup.py bdist_wheel

when you're running your sdist/bdist uploads will upload a `whl` which will install slightly faster :) (and also not depend on setuptools and lots of other fun things).

PR #1374, opened 2015-06-08
Add support for wheels

jamesls:
Closes https://github.com/aws/aws-cli/issues/647

* The main change is moving the conditional dependencies over to extras_require so it'll work with .whl files.
* Also added a commit that did some cleanup to the setup.py to reflect the latest values.
* Updated travis test runner to build/install from a wheel.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1379, opened 2015-06-10
Add `ca_bundle` config option

lorengordon:
Previously, a ca_bundle could only be specified as an environment variable, 'AWS_CA_BUNDLE'. This extends support so `ca_bundle` can be defined in the config file or passed on the command line (`--ca-bundle`).

There's also an alias, 'System', to use the the system ca cert bundle. Logic is *very* simplistic right now. Would be nicer if the logic were implemented upstream in requests. The idea was based on the boto config option, but boto uses the ssl module, which natively implements support for reading the system ca cert bundle (or claims to).

### DIFF
--- a/awscli/__init__.py
+++ b/awscli/__init__.py
@@ -35,7 +35,7 @@ os.environ['AWS_DATA_PATH'] = os.pathsep.join(_awscli_data_path)
 
 
 EnvironmentVariables = {
-    'ca_bundle': ('ca_bundle', 'AWS_CA_BUNDLE', None),
+    'ca_bundle': ('ca_bundle', 'AWS_CA_BUNDLE', None, None),
     'output': ('output', 'AWS_DEFAULT_OUTPUT', 'json', None),
 }
 
--- a/awscli/__init__.py
+++ b/awscli/__init__.py
@@ -35,6 +35,7 @@ os.environ['AWS_DATA_PATH'] = os.pathsep.join(_awscli_data_path)
 
 
 EnvironmentVariables = {
+    'ca_bundle': ('ca_bundle', 'AWS_CA_BUNDLE', None),
     'output': ('output', 'AWS_DEFAULT_OUTPUT', 'json', None),
 }
 
--- a/awscli/customizations/configure.py
+++ b/awscli/customizations/configure.py
@@ -524,7 +524,6 @@ class ConfigureCommand(BasicCommand):
         '    AWS Secret Access Key [None]: secretkey\n'
         '    Default region name [None]: us-west-2\n'
         '    Default output format [None]:\n'
-        '    CA certificate bundle [None]: /etc/pki/tls/certs/ca-bundle.crt\n'
         '\n'
         'To update just the region name::\n'
         '\n'
@@ -533,7 +532,6 @@ class ConfigureCommand(BasicCommand):
         '    AWS Secret Access Key [****]:\n'
         '    Default region name [us-west-1]: us-west-2\n'
         '    Default output format [None]:\n'
-        '    CA certificate bundle [None]:\n'
     )
     SUBCOMMANDS = [
         {'name': 'list', 'command_class': ConfigureListCommand},
@@ -548,7 +546,6 @@ class ConfigureCommand(BasicCommand):
         ('aws_secret_access_key', "AWS Secret Access Key"),
         ('region', "Default region name"),
         ('output', "Default output format"),
-        ('ca_bundle', "CA certificate bundle"),
     ]
 
     def __init__(self, session, prompter=None, config_writer=None):
--- a/awscli/customizations/configure.py
+++ b/awscli/customizations/configure.py
@@ -275,6 +275,7 @@ class ConfigureListCommand(BasicCommand):
         '  access_key     ****************ABCD      config_file    ~/.aws/config\n'
         '  secret_key     ****************ABCD      config_file    ~/.aws/config\n'
         '      region                us-west-2              env    AWS_DEFAULT_REGION\n'
+        '   ca_bundle                <not set>             None    None\n'
         '\n'
     )
 
@@ -301,6 +302,9 @@ class ConfigureListCommand(BasicCommand):
 
         region = self._lookup_config('region')
         self._display_config_value(region, 'region')
+
+        ca_bundle = self._lookup_config('ca_bundle')
+        self._display_config_value(ca_bundle, 'ca_bundle')
 
     def _display_config_value(self, config_value, config_name):
         self._stream.write('%10s %24s %16s    %s\n' % (
@@ -520,6 +524,7 @@ class ConfigureCommand(BasicCommand):
         '    AWS Secret Access Key [None]: secretkey\n'
         '    Default region name [None]: us-west-2\n'
         '    Default output format [None]:\n'
+        '    CA certificate bundle [None]: /etc/pki/tls/certs/ca-bundle.crt\n'
         '\n'
         'To update just the region name::\n'
         '\n'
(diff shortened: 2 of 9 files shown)


lorengordon:
I'll dig into the test failures tomorrow. I hadn't looked at the test suite yet. Most of the failures appear to be because I added a new cli option and of course the tests aren't expecting it yet.

lorengordon:
I think I'm pretty much done with this one. Any thoughts?

kyleknap:
Thanks for the pull request. We will look into it. After an initial glance of the PR, I have a couple questions:

1. To confirm with this PR, a ca cert bundle can now be specified in two additional places: the ``~/.aws/config`` file and as a parameter on the command line? 

2. What is the order of precedence for these three different locations that you can specify a ca cert bundle? How does the order of precedence work if the user specifies both ``--no-verify-ssl`` and ``--ca-bundle`` for a command?

lorengordon:
1. Yes, that is exactly correct. I updated the parameter table in the README, as well as the help documentation, to try to convey that to users.
2. ``--no-verify-ssl`` takes precedence if both parameters are specified. That's how it behaved before, as well. I'll add a unit test for that scenario.

lorengordon:
Just realized I didn't answer the first half of your second question. The order of precedence is, from highest to lowest:

1. `--no-verify-ssl` via the command line
2. `--ca-bundle` via the command line
3. `ca_bundle` via the 'AWS_CA_BUNDLE' environment variable
4. `ca_bundle` via the config file
5. `ca_bundle` via the session var default defined in __init.py__ (`None`), which tells botocore to fall back on its default.

I believe this follows the precedence order of other variables, as well.

kyleknap:
Cool. Thanks for answering my questions. I think I have one last question for now:

What is your use case for passing it as a value to a parameter or specifying it in the config file as opposed to setting the environment variable?

lorengordon:
Greater flexibility, primarily. We support a lot of tenants, and they all have many ways of doing things. Environment variables aren't always great for cron jobs and some types of sessions, in particular. We try to make things easier for them, sometimes by managing the build (in which case we prefer file-based configurations to try to avoid polluting the user environment), and sometimes by providing guidance and tools. 

I'd eventually like to see a system-wide default config file (a la /etc/boto.cfg), with user settings merged with the system settings (conceptually, `systemconfig.update(userconfig)`). This was something of a first step to getting there. I'm not sure I have the cycles or expertise to add that bit myself, though, so I'll probably open an issue for the system config.

kyleknap:
Thanks for the explanation. We will look into the PR and get back to you.

lorengordon:
I went ahead and rebased against latest develop (00b361cd968cb634d63ae69168cd93aad77d00e9) and fixed the one merge conflict.

lorengordon:
Rebased again. No conflicts this time. @kyleknap, any feedback yet on the PR?

lorengordon:
Saw some merged PRs mentioning the argparser and the AWS_CA_BUNDLE. Rebased just to make sure there were no conflicts. Still clean.

kyleknap:
@lorengordon I will take a look at the PR sometime this week and get back to you.

lorengordon:
@kyleknap sounds good, thanks!

lorengordon:
Still rebases cleanly.

jamesls:
Thanks, taking a look.

awscli/customizations/configure.py
```diff
@@ -302,6 +303,9 @@ def _run_main(self, args, parsed_globals):
         region = self._lookup_config('region')
         self._display_config_value(region, 'region')
 
+        ca_bundle = self._lookup_config('ca_bundle')
```
jamesls: I'd prefer not to have this in the `aws configure` prompts.

We specifically ask a subset of config questions so that a new user can get up and running quickly by answering a small amount of questions.

I'd say configuring a custom ca_bundle is not in the common case for getting started.  Users can still run `aws configure set ca_bundle /foo` to set this programatically.

awscli/customizations/globalargs.py
```diff
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import six
```
jamesls: We generally import this from our compat module: `from awscli.compat import six`

awscli/customizations/globalargs.py
```diff
@@ -66,6 +58,23 @@ def _resolve_endpoint_url(value):
     return value
 
 
+def resolve_verify_ssl(parsed_args, session, **kwargs):
+    SYSTEM_CA_BUNDLE = '/etc/pki/tls/certs/ca-bundle.crt'
```
jamesls: I like this idea of using the system ca cert, but I'd like to explore a more generic solution that's more cross platform (windows, Mac, etc.).

I think this should be possible.  Would you be ok with removing this for now, and then have a separate PR for supporting a "system" value?

awscli/__init__.py
```diff
@@ -35,6 +35,7 @@
 
 
 EnvironmentVariables = {
+    'ca_bundle': ('ca_bundle', 'AWS_CA_BUNDLE', None, None),
```
jamesls: This is more a note for our team, but I think it would be worth moving this into botocore's `Session` so that python SDK users could also leverage setting a CA bundle in the `~/.aws/config` file and `AWS_CA_BUNDLE` env var.

awscli/customizations/globalargs.py
lorengordon: I can do that, yes. When I wrote the code, I did some research on cross-platform support for determining the system cert bundle, but I got waaaaay out of my depth really quickly. So, yeah, this was a shortcut for my immediate needs.

awscli/customizations/configure.py
lorengordon: Makes sense, no problem.

awscli/customizations/globalargs.py
mtdowling: FYI, when trying to add this in another PR, Guzzle does a decent job of finding a CA bundle on different platforms: https://github.com/guzzle/guzzle/blob/master/src/functions.php#L160-L174

lorengordon:
That should address the comments. New PR coming that contains the 'system' alias.

lorengordon:
Ok, I've got the new branch with only the commit introducing the 'system' alias, but it is dependent on the commits in this branch, so all of these commits would also show up in that PR. Should I wait to open the PR until this is merged, or go ahead and open it now and let it sort out after this is merged?

awscli/customizations/configure.py
lorengordon: Fyi, this comment didn't get outdated because this seemed like the wrong place to achieve the desired effect. Removing the lines here means that `aws configure list` won't print the line with the `ca_bundle` settings. I didn't think that was what you meant. Instead, I removed the lines [here](https://github.com/lorengordon/aws-cli/commit/b6d1dc2d06a3d1e78dfddd6ea64a8a8620fe1b11#diff-67fa1a65752518af4d764aa3e6a9cf31L551) so `aws configure` wouldn't prompt for `ca_bundle`.

jamesls: I'd prefer to remove this from the `list` command for now as well.

`aws configure/aws configure list` currently show the bare minimum configuration values that a customer needs to get started.

I'd like to approach this in a more general manner and have something like `aws configure list --all` to list all the known CLI options.  There's other config values that `aws configure list` is not currently showing so we probably need an option that shows everything.

jamesls:
I've rebased your PR against develop and fixed up a few small things.  This has been merged via: https://github.com/aws/aws-cli/commit/c700906c9f59d4699a677fdbf61d7f6d145af8ec

We can go ahead and start look at supporting some type of "system" alias.  Thanks again for the PR.

lorengordon:
Cool, thanks! I'll just open an issue for the "system" alias. My implementation was very naive, and I won't have time to work on it for several weeks (at least).

awscli/customizations/configure.py
lorengordon: I understand your point, but as it is, there is no way to verify where the `ca_bundle` option is getting its value. That kind of seems even less ideal to me. But oh well, I suppose I'll just open a new issue for the `aws configure list --all` CLI option to keep it on the radar.

PR #1381, opened 2015-06-12
Use the latest version of virtualenv/pip for bundled installer

jamesls:
We need a version of pip that supports the extras_require
conditional dependencies we now use for whl files.

This fixes a regression to the bundled installer for python2.6.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1382, opened 2015-06-12
Fix regression with old version of pip and cond deps

jamesls:
Verified with:

```
<mkvirtualenv with python2.6>
pip install pip==1.4.1
<create botocore and awscli sdists>
pip install botocore/dist/<sdist>.tar.gz
pip install awscli/dist/<sdist>.tar.gz
aws --version
```

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1387, opened 2015-06-15
Functional tests

kyleknap:
This PR moves many of the ``assert_params_for_cmd`` tests into a functional test section to help better categorize the various test that we have as the unit test section was getting cluttered and the definition of the types of test that go into it was a bit ambiguous.

cc @jamesls @mtdowling


mtdowling:
Looks fine to me

jamesls:
I think you can also move many of the tests/unit/customizations/emr tests over there as well.

Otherwise :shipit:

kyleknap:
@jamesls 
Did not do the emr tests because it was not a simple cut and paste. Instead did some of the s3 and lambda customizations. Those were cut and pastes. I also found a really nasty bug in the test suite where a patch was not stopped on datetime.

Let me know what you think of the additional tests added to the functional test suite.

jamesls:
:shipit:  Looks good.  Good catch on the datetime patch bug also.

PR #1388, opened 2015-06-15
Update env vars to sync with latest botocore changes

jamesls:
* Need to add another member that includes a conversion func
* Remove unnecessary var mappings.  These have since been moved
  down into botocore so this mapping no longer needs to be maintained
  at the CLI level (see: https://github.com/boto/botocore/blob/develop/botocore/session.py#L74).

Requires: https://github.com/boto/botocore/pull/559

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1391, opened 2015-06-16
Adding 'aws:kms' encryption support to S3 uploads

mckinnonderek:
These changes allow for users to specify 'aws:kms' as a server side encryption method for uploading files to S3.

Instead of `--sse` being a boolean flag to enable AES256, I made it a regular parameter switch that accepts either AES256 or aws:kms, the former being the default.

It's a bit messy as I was implementing this to solve a problem of mine. I will mark the code areas where I've done something probably hacky.

Also, I'm missing unit tests (which I can write when I get time to) and updating of the relevant documentation.

Thanks!

### DIFF
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -218,7 +218,7 @@ class FileInfo(TaskInfo):
         if self.parameters['sse']:
             sse = self.parameters['sse'][0]
             if sse == 'aws:kms':
-                self.client._client_config.signature_version = 's3v4'
+                self.client._request_signer._signature_version = 's3v4'
             params['ServerSideEncryption'] = sse
         if self.parameters['storage_class']:
             params['StorageClass'] = self.parameters['storage_class'][0]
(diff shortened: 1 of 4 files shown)


awscli/customizations/s3/fileinfo.py
```diff
@@ -216,7 +216,10 @@ def _handle_object_params(self, params):
                                      'permission=principal')
                 params[self._permission_to_param(permission)] = grantee
         if self.parameters['sse']:
-            params['ServerSideEncryption'] = 'AES256'
+            sse = self.parameters['sse'][0]
+            if sse == 'aws:kms':
+                self.client._request_signer._signature_version = 's3v4'
```
mckinnonderek: Directly changing the `RequestSigner`'s `_signature_version` is pretty hacky but at this point in the code, I'm not able to attach to the `event_emitter` (AFAIK) to mutate it.

awscli/customizations/s3/fileinfo.py
```diff
@@ -263,9 +266,10 @@ def _handle_upload(self, body):
         }
         self._handle_object_params(params)
         response_data = self.client.put_object(**params)
-        etag = response_data['ETag'][1:-1]
-        body.seek(0)
-        check_etag(etag, body)
+        if response_data['ServerSideEncryption'] != 'aws:kms':
```
mckinnonderek: Much like #1026, I had to disable the etag checking because of the difference in MD5ness between encrypted and non-encrypted.

PR #1396, opened 2015-06-17
Update dependencies and imports via bcdoc port

kyleknap:
Generated html docs look fine as well.

cc @jamesls  @mtdowling


jamesls:
:shipit:

PR #1402, opened 2015-06-19
Properly configure loader sharing in tests

jamesls:
The loader argument to the session has been a no op
for a while (being removed in
https://github.com/boto/botocore/pull/579).
The way to set a loader is to just use register_component().

This gives shaves about 4 seconds off of the total unit test time
for me.

cc @kyleknap @mtdowling


kyleknap:
Nice! :ship:

PR #1403, opened 2015-06-19
Read only profile

jamesls:
This is a possible way to handle the read only profile var from https://github.com/boto/botocore/pull/580.

This was branched off of the loader arg removal from the previous pull request so you'll see a diff of that as well until that's merged.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1405, opened 2015-06-22
Remove etag verification on uploads, not needed anymore

jamesls:
This has been pushed down to the botocore layer so other
projects such as boto3 can leverage this logic.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #1413, opened 2015-07-03
fix error in get-shipping-label example

mwunderl:


mtdowling:
How does this fix an error? I see you moved part of the command to another line. This actually breaks the formatting of the example.

mwunderl:
Error was in the --street-1 parameter (was misnamed). Line break was unintentional. Apologies.

PR #1418, opened 2015-07-09
Fix issue in windows test for codecommit

kyleknap:
There is special logic on windows machines to change the format to binary and that requires the file number. A StringIO object has no ``fileno()`` method so a mock ``fileno()`` method had to be added. Also did some pep8 cleanup.


mtdowling:
:ship:

PR #1419, opened 2015-07-09
Fix codecommit integration test

kyleknap:
Did not realize the integration test had the same issue as the unit tests that were fixed in this PR: https://github.com/aws/aws-cli/pull/1418

cc @mtdowling @jamesls @rayluo

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -34,6 +34,8 @@ import string
 from pprint import pformat
 from subprocess import Popen, PIPE
 
+from awscli.compat import StringIO
+
 
 try:
     import mock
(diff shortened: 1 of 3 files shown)


awscli/testutils.py
```diff
@@ -705,3 +707,8 @@ def assert_no_errors(self, p):
         self.assertNotIn("failed:", p.stderr)
         self.assertNotIn("client error", p.stderr)
         self.assertNotIn("server error", p.stderr)
+
+
+class StringIOWithFileNo(StringIO):
+    def fileno(self):
+        return 0
```
jamesls: Doesn't matter, but technically mocking out stdout would have a fileno() of 1.

jamesls:
:shipit:

awscli/testutils.py
kyleknap: True, I guess it would matter more if the class was specifically for standard output. It is just being used for standard output for both cases it is used.

PR #1433, opened 2015-07-27
Fix preview snippet for operation commands

kyleknap:
We had an incorrect example in the operation help page on how to set a general service out of preview. For example for ``aws cloudfront create-distribution``, the ``help`` command would say to
run ``aws configure set preview.create-distribution true``. Also look here for the example in the html: http://docs.aws.amazon.com/cli/latest/reference/cloudfront/create-distribution.html


jamesls:
:shipit:

PR #1436, opened 2015-07-31
Adding enum documentation

mtdowling:
This PR adds enum documentation to parameters.

screen shot 2015-07-30 at 10 11 58 pm

screen shot 2015-07-30 at 10 11 50 pm

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -14,6 +14,7 @@ import logging
 import os
 from botocore import xform_name
 from botocore.docs.bcdoc.docevents import DOC_EVENTS
+from botocore.model import StringShape
 
 from awscli import SCALAR_TYPES
 from awscli.argprocess import ParamShorthandDocGen
@@ -160,6 +161,7 @@ class CLIDocumentEventHandler(object):
         doc.write('%s (%s)\n' % (name, argument.cli_type_name))
         doc.style.indent()
         doc.include_doc_string(argument.documentation)
+        self._document_enums(argument, doc)
         doc.style.dedent()
         doc.style.new_paragraph()
 
@@ -176,6 +178,19 @@ class CLIDocumentEventHandler(object):
             text=related_item
         )
         doc.write('\n')
+
+    def _document_enums(self, argument, doc):
+        """Documents top-level parameter enums"""
+        if hasattr(argument, 'argument_model'):
+            model = argument.argument_model
+            if isinstance(model, StringShape):
+                if model.enum:
+                    doc.style.new_paragraph()
+                    doc.write('Possible values:')
+                    doc.style.start_ul()
+                    for enum in model.enum:
+                        doc.style.li('``%s``' % enum)
+                    doc.style.end_ul()
 
 
 class ProviderDocumentEventHandler(CLIDocumentEventHandler):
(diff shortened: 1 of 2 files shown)


awscli/clidocs.py
```diff
@@ -159,6 +159,7 @@ def doc_option(self, arg_name, help_command, **kwargs):
             name = '``%s``' % argument.cli_name
         doc.write('%s (%s)\n' % (name, argument.cli_type_name))
         doc.style.indent()
+        self._document_enums(argument, doc)
```
kyleknap: I feel that it may look better if the enum values are documented after the doc string. Thoughts?

awscli/clidocs.py
```diff
@@ -177,6 +178,19 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
         )
         doc.write('\n')
 
+    def _document_enums(self, argument, doc):
+        """Documents top-level parameter enums"""
+        model = getattr(argument, 'argument_model', {})
+        metadata = getattr(model, 'metadata', {})
+        enums = metadata.get('enum', [])
```
kyleknap: What is the reason in not using the work you did for this PR: https://github.com/boto/botocore/pull/608/files here? I also noticed the use of enum values in the JSON syntax related methods such as here: https://github.com/aws/aws-cli/blob/develop/awscli/clidocs.py#L43. It would be nice if we used what you added there as well.

tests/unit/test_clidocs.py
```diff
@@ -156,6 +161,25 @@ def test_breadcrumbs_wait_command_html(self):
...
+        help_command.arg_table = arg_table
+        operation_model = mock.Mock()
+        operation_model.service_model.operation_names = []
+        help_command.obj = operation_model
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option('arg-name', help_command)
+        self.assertTrue(
+            _rendered_docs_contains(help_command,
+                                    'Possible values:'))
+
```
kyleknap: Can we make sure that one of the enum values are included as well like ``Foo``?

kyleknap:
It looked fine. I had a couple of questions though that I would like answered first.

awscli/clidocs.py
jamesls: Yeah I agree with that.  Seems like the AWS API reference docs put the enums/metadata after the doc strings (e.g http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html)

tests/unit/test_clidocs.py
```diff
@@ -35,12 +42,6 @@ def setUp(self):
         self.operation_handler = OperationDocumentEventHandler(
             self.help_command)
 
-    def assert_rendered_docs_contain(self, expected):
```
jamesls: What about just extracting a base test class with this assertion method so the two test classes can share?

assertIn() will show the expected and the contents in the error message on test failure.  With `assertTrue(_rendered_docs_contains)` we'll just get "Assertion error: False is not True".

jamesls:
Also agree with Kyle's comments.  Otherwise looks good.

mtdowling:
I've updated the PR based on all of the feedback.

mtdowling:
Look good now, Kyle?

jamesls:
:shipit:

kyleknap:
Thanks! :ship:

PR #1439, opened 2015-08-06
Remove duplicate documentation of ec2 --dry-run

kyleknap:
Testing it was a bit tricky. I did not want to rely on the current documentation because that is subject to change and I do not want a documentation change to break the tests. So I added a test that makes sure that nothing gets registered with the event format the dryrun customization needed as that customization was the only one that registered to ``'doc-option-example.ec2.*.dry-run'``.


tests/unit/test_handlers.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestNoDryRunDocCustomization(unittest.TestCase):
+    def test_no_dry_run_doc_customizations(self):
+        event_emitter = mock.Mock()
+        awscli_initialize(event_emitter)
+        registered_handlers = event_emitter.register.call_args_list
+        # Make sure there is no doc customization for ec2's DryRun
+        self.assertNotIn(
+            mock.call('doc-option-example.ec2.*.dry-run', mock.ANY),
```
jamesls: This seems overspecified to me.  I wouldn't want this to fail if we eventually add some valid customization to this arg.  Is there a different way to test this?

kyleknap: @jamesls s I cannot think of a better idea. If you have suggestions that would awesome. The two other choices that I considered was:

* Add a functional/integration test to make sure the documentation was not repeated. However, this would rely on the actual wordings in the documentation and I really do not like relying on documentation values that comes from models because if you update the model, it may cause a false positive test failure if the documentation changes.
* Not add a test at all. I think this actually makes the most sense because it seems sort of backwards that a test is added to make sure that a customization was removed. Especially since the --dry-run parameter falls under the same logic as all model driven parameters/commands now. I did not choose this option on the premise of following the principle that we always add tests for PRs.

Let me know if you like either of these ideas.

jamesls: I would also be fine with not adding a negative test for this.  We are removing lines of code here and there's no drop in coverage, so I slightly prefer the second suggestion.

kyleknap: Cool. I will do that.

kyleknap:
@jamesls I updated. Ready to review again.

jamesls:
:shipit:

PR #1444, opened 2015-08-11
Add new shorthand parser for nested maps

jamesls:
This adds a new shorthand parser that can handle nested maps via a ``{foo=bar}`` syntax.  Note this is not plumbed into the argprocess module yet, it's entirely standalone in this PR.

The plan for this is:

1. Get this parser reviewed
2. Integrate this with the existing shorthand parser and add in back-compat code (separate PR).

There's one point in this PR not implemented that we need to decide.  Right now, the explicit list parser `[foo, bar baz]` assumes that all the values are strings.  If we want to support nested lists or lists of hashes, we will need to require string values that start with `[` and `{` to be quoted.  There's  a comment on line 148 that explains this.

### DIFF
--- a/awscli/shorthand.py
+++ b/awscli/shorthand.py
@@ -36,12 +36,17 @@ class ShorthandParseError(Exception):
         super(ShorthandParseError, self).__init__(msg)
 
     def _construct_msg(self):
+        if '\n' in self.value:
+            last_newline = self.value[:self.index].rindex('\n')
+            num_spaces = self.index - last_newline - 1
+        else:
+            num_spaces = self.index
         msg = (
             "Expected: '%s', received: '%s' for input:\n"
             "%s\n"
             "%s\n"
         ) % (self.expected, self.actual, self.value,
-             ' ' * self.index + '^')
+             ' ' * num_spaces + '^')
         return msg
 
 
--- a/awscli/shorthand.py
+++ b/awscli/shorthand.py
@@ -37,6 +37,12 @@ class ShorthandParseError(Exception):
 
     def _construct_msg(self):
         if '\n' in self.value:
+            # If there's newlines in the expression, we want
+            # to make sure we're only counting the spaces
+            # from the last newline:
+            # foo=bar,\n
+            # bar==baz
+            #     ^
             last_newline = self.value[:self.index].rindex('\n')
             num_spaces = self.index - last_newline - 1
         else:
@@ -107,12 +113,24 @@ class ShorthandParser(object):
             return self._csv_value()
 
     def _csv_value(self):
+        # Supports either:
+        # foo=bar     -> 'bar'
+        #     ^
+        # foo=bar,baz -> ['bar', 'baz']
+        #     ^
         first_value = self._first_value()
         self._consume_whitespace()
         if self._at_eof() or self._input_value[self._index] != ',':
             return first_value
         self._expect(',', consume_whitespace=True)
         csv_list = [first_value]
+        # Try to parse remaining list values.
+        # It's possible we don't parse anything:
+        # a=b,c=d
+        #     ^-here
+        # In the case above, we'll hit the ShorthandParser,
+        # backtrack to the comma, and return a single scalar
+        # value 'b'.
         while True:
             try:
                 current = self._second_value()
@@ -126,9 +144,19 @@ class ShorthandParser(object):
                 csv_list.append(current)
             except ShorthandParseError:
                 # Backtrack to the previous comma.
+                # This can happen when we reach this case:
+                # foo=a,b,c=d,e=f
+                #     ^-start
+                # foo=a,b,c=d,e=f
+                #          ^-error, "expected ',' received '='
+                # foo=a,b,c=d,e=f
+                #        ^-backtrack to here.
                 self._backtrack_to(',')
                 break
         if len(csv_list) == 1:
+            # Then this was a foo=bar case, so we expect
+            # this to parse to a scalar value 'bar', i.e
+            # {"foo": "bar"} instead of {"bar": ["bar"]}
             return first_value
         return csv_list
 
--- a/awscli/shorthand.py
+++ b/awscli/shorthand.py
@@ -102,15 +102,15 @@ class ShorthandParser(object):
     def _values(self):
         # values = csv-list / explicit-list / hash-literal
         if self._at_eof():
-            return None
+            return ''
         elif self._current() == '[':
             return self._explicit_list()
         elif self._current() == '{':
             return self._hash_literal()
         else:
-            return self._csv_list()
+            return self._csv_value()
 
-    def _csv_list(self):
+    def _csv_value(self):
         first_value = self._first_value()
         self._consume_whitespace()
         if self._at_eof() or self._input_value[self._index] != ',':
@@ -143,6 +143,7 @@ class ShorthandParser(object):
         result = self._FIRST_VALUE.match(self._input_value[self._index:])
         if result is not None:
             return self._consume_matched_regex(result)
+        return ''
 
     def _explicit_list(self):
         # explicit-list = "[" [value *(",' value)] "]"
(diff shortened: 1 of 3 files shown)


awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+        self.index = index
+        msg = self._construct_msg()
+        super(ShorthandParseError, self).__init__(msg)
+
+    def _construct_msg(self):
+        msg = (
+            "Expected: '%s', received: '%s' for input:\n"
+            "%s\n"
+            "%s\n"
+        ) % (self.expected, self.actual, self.value,
```
kyleknap: I really like the idea of the ``'^'`` pointer. It could be super useful for debugging. How well does it look if the self.value requires more than two lines on the terminal due to wrapping in the terminal? Will the ``'^'`` still match up with where the syntax error occurred? I am not sure if I can think of any better representation.

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+        key = self._key()
+        self._consume_whitespace()
+        self._expect('=')
+        self._consume_whitespace()
+        values = self._values()
+        return {key: values}
+
+    def _key(self):
+        # key = 1*(alpha / %x30-39)  ; [a-zA-Z0-9]
+        valid_chars = string.ascii_letters + string.digits
```
kyleknap: How does ``'.'`` get handled? We are still planning to add support for:
 ``foo.bar=baz`` --> ``{'foo': {'bar': 'baz'}}``

Right?

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+        while not self._at_eof():
+            if self._current() not in valid_chars:
+                break
+            self._index += 1
+        return self._input_value[start:self._index]
+
+    def _values(self):
+        # values = csv-list / explicit-list / hash-literal
+        if self._at_eof():
+            return None
```
kyleknap: When will this if statement be entered? 

Seems sort of strange that the value gets translated to None instead of an error or something because the value is completely missing making it ambiguous what was intended. Or does this get translated based on how the value is represented in the model?

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+            return self._hash_literal()
+        else:
+            return self._csv_list()
+
+    def _csv_list(self):
+        first_value = self._first_value()
+        self._consume_whitespace()
+        if self._at_eof() or self._input_value[self._index] != ',':
+            return first_value
+        self._consume_whitespace()
```
kyleknap: Is this second ``consume_whitespace()`` needed? It was already called once so all of the whitespace should have been consumed already?

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+                    break
+                self._expect(',')
+                self._consume_whitespace()
+                csv_list.append(current)
+            except ShorthandParseError:
+                # Backtrack to the previous comma.
+                self._backtrack_to(',')
+                break
+        if len(csv_list) == 1:
+            return first_value
```
kyleknap: So ``_csv_list()`` can return a list or scalar value? It is a little confusing as it would seem that a list should be returned. I am assuming that this is to just normalize single value lists and single values to help handle the ambiguous case of ``foo=bar`` where bar could be ``['bar']`` or ``'bar'`` depending on the model.

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+        if result is not None:
+            return self._consume_matched_regex(result)
+
+    def _explicit_list(self):
+        # explicit-list = "[" [value *(",' value)] "]"
+        self._expect('[')
+        self._consume_whitespace()
+        values = []
+        while self._current() != ']':
+            # TODO: We need to decide if we're ok with changing
```
kyleknap: I'm fine with changing this. Users should really be quoting these items when they are ambiguous.

kyleknap:
The changes look fine. I have just a couple of questions that probably would got have answered with the subsequent PR's that you are sending.

awscli/shorthand.py
jamesls: I don't believe it handles that case.  I can update that.

awscli/shorthand.py
jamesls: I don't believe we are at this point.  I think it was because it ended up being ambiguous:

```
foo=a.b=c,a.c=d,bar=baz
{"foo": {"a": {"b": "c", "c": "d"}}, "bar": "baz"}
{"foo": {"a": {"b": "c", "c": "d"}, "bar": "baz"}}
```

So in this example it's not clear if the last `bar=baz` belongs to the inner dict or the outer dict.  Using the `{}` makes this clear.

awscli/shorthand.py
jamesls: That's because the current parser allows for empty values.  So the question is what should this parse to:

```
foo=bar,baz=
```

I believe the current parser parses this as `{"foo": "bar", "baz": None}`.

The current parser also allows you to do things like: `foo=,bar=,baz=` so we still have to support that.

There's a few test for this case: https://github.com/aws/aws-cli/pull/1444/files#diff-1c705f01155c5e88adc98c5095118ff9R24

awscli/shorthand.py
jamesls: Good point, I don't think this is actually needed.

awscli/shorthand.py
jamesls: Yeah exactly.

So maybe it makes sense for csv list to always return a list, and then have the walker that compares against the model do the `[a] -> a` conversion instead of the walker doing a `a -> [a]` conversion.

I think the only reason I did it this way is that it's much more common to have key val pairs where the model actually has the value being a scalar value, but I think from an implementation standpoint it makes more sense to always have csv_list return a list.  That means we parse key=<anything> such that <anything> is always a list.


Thoughts?

awscli/shorthand.py
kyleknap: This makes sense. We do not need the ``'.'`` notation.

awscli/shorthand.py
kyleknap: If that is the current behavior, I think we really do not have the choice but to have it continue to behave that way. The case that I think would be weird is that if I was expecting it to be translated to an empty string. Probably would have to do something like ``foo=bar,baz=''``

awscli/shorthand.py
kyleknap: Yeah I would prefer a list to be consistent in return values for that method, if it is not anymore difficult for the model walker translation.

awscli/shorthand.py
```diff
@@ -0,0 +1,264 @@
...
+        if self._at_eof():
+            return None
+        elif self._current() == '[':
+            return self._explicit_list()
+        elif self._current() == '{':
+            return self._hash_literal()
+        else:
+            return self._csv_list()
+
+    def _csv_list(self):
```
mtdowling: This method does a lot, including backtracking the parser. It would be nice for maintainability reasons to have a comment here explaining how it works.

awscli/shorthand.py
jamesls: I double checked on this, and it looks like it actually parses to an empty string.  I'll go ahead and update the code.

awscli/shorthand.py
jamesls: I ended up keeping it so that `foo=bar -> {"foo": "bar"}` and `foo=bar,baz -> {"foo": ["bar", "baz"]}`.  I think that makes the most sense to me, and that the method name just wasn't named properly.  I've added some comments explaining this in more detail so hopefully it's more clear.

awscli/shorthand.py
jamesls: Updated.

jamesls:
I believe I've incorporated all the feedback here.  cc @kyleknap @mtdowling

Let me know if I missed anything.

kyleknap:
Looks good. Thanks for updating. :ship:

PR #1445, opened 2015-08-11
Update CLI docs about AWS_SHARED_CREDENTIALS_FILE

jamesls:
From https://github.com/boto/botocore/pull/623


rayluo:
Looks good to me.

In the future, we may also want to update the README.rst for the following sections:

* **Getting Started**: Before using aws-cli, you need to tell it about your AWS credentials. You can do this in several ways: ... (add description for new way here)
* **Other Configurable Variables**: ... (add description for new Configurable Variables here)

PR #1447, opened 2015-08-13
Handle case when create-virtual-mfa-device errors out

jamesls:
This was previously propogating a KeyError and would just
display the error message 'VirtualMFADevice'.  Now
we let the error propogate and print a full error message.


kyleknap:
Looks good. :ship:

PR #1457, opened 2015-08-19
sqs create-queue example with embedded JSON document

mwunderl:


rayluo:
Thanks! We will verify this.

PR #1459, opened 2015-08-19
A tentative cloudfront sign implementation, subject to refactoring

rayluo:
The sign feature is there. Two helper files and their test cases would better be moved into botocore, and then this PR will be changed accordingly.

And, the code structure is still subject to change when more CloudFront commands are coming.

### DIFF
--- a/awscli/customizations/cfsign.py
+++ b/awscli/customizations/cfsign.py
@@ -1,0 +1,77 @@
+# Should go into botocore/cloudfront.py
+
+# Copyright 2013 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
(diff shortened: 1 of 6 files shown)


jamesls:
Thanks.  Looks like the travis build is [failing](https://travis-ci.org/aws/aws-cli/builds/76377013).  Let's get travis passing as a next step.

mtdowling:
Perhaps we should add a new package instead of adding cf* files into the customizations package.

jamesls:
Hey, sorry for the delay here.  Can we just get the pull request updated to move the appropriate modules to botocore?  Now that we know we want to move the signing stuff to botocore, let's just go ahead and do that before reviewing this code.

Thanks!

rayluo:
Both of you are right. This tentative PR was intended to be moved to botocore since day one, so I foresaw those cf* files would be gone soon.

Now please go review its cousin in [this botocore PR](https://github.com/boto/botocore/pull/635). When it passes the review and gets merged, I will then refactor this CLI implementation to rely on that botocore PR.

jamesls:
Closing out for now.  We can reopen once we pick this work back up again.

PR #1461, opened 2015-08-20
Plumb in new shorthand parser into the CLI

jamesls:
This PR builds on https://github.com/aws/aws-cli/pull/1444 and plumbs the parser into the argprocess module.

This PR also includes a backwards compatibility layer that walks the parsed values from the new parser and makes modifications so that the parser produces the same output as the old parser.

All the tests are passing.  The one test customization I had to change was the lambda customization because it assumed that the shorthand parser would throw an error if you gave it a key that wasn't in the model.  Now the customization has to handle that logic itself.

And finally, I had to update the new shorthand parser to handle escaped commas in a CSV list, which the old parser supported, so you'll see some small modifications to the parser.

The only thing remaining is the documentation.  For now I've left the existing doc class as is, so it will generate the same docs as before, but it won't generate docs for the nested map syntax.  That will have to be a future addition.

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -15,13 +15,12 @@ import os
 import logging
 from awscli.compat import six
 
-from botocore import xform_name
 from botocore.compat import OrderedDict, json
 
-from awscli import utils
 from awscli import SCALAR_TYPES, COMPLEX_TYPES
 from awscli.paramfile import get_paramfile, ResourceLoadingError
 from awscli.paramfile import PARAMFILE_DISABLED
+from awscli import shorthand
 
 
 LOG = logging.getLogger('awscli.argprocess')
@@ -235,26 +234,9 @@ def unpack_scalar_cli_arg(argument_model, value, cli_name=''):
 
 class ParamShorthand(object):
 
-    # To add support for a new shape:
-    #
-    #  * Add it to SHORTHAND_SHAPES below, key is the shape structure
-    #    value is the name of the method to call.
-    #  * Implement parse method.
-    #  * Implement _doc_<parse_method_name>.  This is used to generate
-    #    the docs for this shorthand syntax.
-
-    SHORTHAND_SHAPES = {
-        'structure(scalars)': '_key_value_parse',
-        'structure(scalar)': '_special_key_value_parse',
-        'structure(list-scalar, scalar)': '_struct_scalar_list_parse',
-        'map-scalar': '_key_value_parse',
-        'list-structure(scalar)': '_list_scalar_parse',
-        'list-structure(scalars)': '_list_key_value_parse',
-        'list-structure(list-scalar, scalar)': '_list_scalar_list_parse',
-    }
-
     def __init__(self):
-        pass
+        self._parser = shorthand.ShorthandParser()
+        self._visitor = shorthand.BackCompatVisitor()
 
     def __call__(self, cli_argument, value, **kwargs):
         """Attempt to parse shorthand syntax for values.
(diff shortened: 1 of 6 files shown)


awscli/argprocess.py
```diff
@@ -316,171 +344,32 @@ def get_parse_method_for_param(self, cli_argument, value=None):
...
-    def _split_on_commas(self, value):
-        try:
-            return utils.split_on_commas(value)
-        except ValueError as e:
-            raise ParamSyntaxError(six.text_type(e))
+            return False
+        model = cli_argument.argument_model
+        # The second case is to make sure the argument is sufficiently
+        # complex, that is, it's base type is a complex type *and*
+        # if it's a list, then it can't be a list of scalar types.
```
kyleknap: Which I assume gets handled by this line https://github.com/aws/aws-cli/pull/1461/files#diff-5426f850c5250bfda7a8a2e923b6cee1R280?

awscli/shorthand.py
```diff
@@ -282,3 +332,71 @@ def _backtrack_to(self, char):
...
+
+
+class BackCompatVisitor(ModelVisitor):
+    def _visit_list(self, parent, shape, name, value):
+        if not isinstance(value, list):
+            # Convert a -> [a] because they specified
+            # "foo=bar", but "bar" should really be ["bar"].
+            if value is not None:
+                parent[name] = [value]
+        elif shape.member.type_name == 'structure' and \
```
kyleknap: Is this also handled here: https://github.com/aws/aws-cli/pull/1461/files#diff-5426f850c5250bfda7a8a2e923b6cee1R306? I see that this handles the case when the list is nested in another structure, but wouldn't your visitor handle the case for the line I linked as well?

awscli/shorthand.py
```diff
@@ -282,3 +332,71 @@ def _backtrack_to(self, char):
...
+                parent, shape, name, value)
+
+    def _visit_scalar(self, parent, shape, name, value):
+        if value is None:
+            return
+        type_name = shape.type_name
+        if type_name in ['integer', 'long']:
+            parent[name] = int(value)
+        elif type_name in ['double', 'float']:
+            parent[name] = decimal.Decimal(value)
```
kyleknap: This used to get parsed as a float. Is there any downsides in switching it to a Decimal? I know switching to a decimal improves precision.

kyleknap:
The changes look fine. I have a couple questions, but nothing big. The if statements for some of the cases were kind of long for my taste, but they needed to be specific in order to avoid any unwanted behavior.

awscli/argprocess.py
jamesls: Not sure if I understanding what you're asking, but the shorthand parser actually doesn't touch anything that's a list of scalar types.  Note that this is checking the *model* structure, not the input.  The line you've linked to is where the *model* is a list of structures, but the *value* provided by the customer happens to be a list of strings, which we convert to a list of structures.

awscli/shorthand.py
jamesls: I think this can actually be removed.  The visitor was written before plumbing it into the argprocess module.  Unfortunately the visitors are applied *after* parsing, so there's a requirement that the shorthand parser can interpret `i-123 i-124 i-125` and parse it into a list of strings: `['i-123', 'i-124', 'i-125']`.  This would complicate the parser to add additional special cased grammar rules.

So I opted to instead handle this special case in the argprocess module (via `_handle_special_cases`).  This means this processing can be removed from the visitor class because it's relying on an input format that the parser will never produce.

awscli/argprocess.py
kyleknap: That answers it for me. I was just making sure that you ignored the list of scalars, in terms of the shorthand parser, because it is handled in the line that I linked.

awscli/shorthand.py
jamesls: That is a good question.  I think there was actually a TODO somewhere in this module where we note that we're losing precision on float types.  I don't think there's any issues (all the unit/integ/functional tests pass), but I'm also ok with reverting this for now and addressing this in a future change to reduce the number of moving parts in this PR.

awscli/shorthand.py
kyleknap: Yeah that sounds good. I just saw repeated lines. So I figured some optimizing could be done.

awscli/shorthand.py
kyleknap: I think I would prefer to keep it a float for the sake of less moving parts.

jamesls:
@kyleknap I believe I've incorporated all the feedback you had.

kyleknap:
Thanks. Looks good. :ship:

mtdowling:
:ship:

PR #1464, opened 2015-08-24
Add count to output options

cstamm13:
This feature allows users to get a count for the number of elements in the output. This eliminates the need to pipe to wc (or other word count tool) by directly counting the number of values in the output.

This was motivated by the need to write a tool for alerting purposes, where I needed to count the number of machines using describe-instances.

### DIFF
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -27,7 +27,7 @@
                 "table",
                 "count"
             ],
-	        "help": "<p>The formatting style for command output. Note that \"count\" requires the use of \"--query\"</p>"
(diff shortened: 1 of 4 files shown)


kyleknap:
Thanks for the pull request. The only thing is that I am not seeing how much this differs from the ``length()`` function through the ``--query`` option in terms of functionality.  For example, I can reproduce this IAM ListUsers case using ``--query``and ``length()``:
```
$ aws iam list-users --query 'length(Users)'
```

Would this handle your use case? Thoughts?

cstamm13:
I wasn’t aware of such functionality. This pull request can be closed. Thank you!

PR #1467, opened 2015-08-25
Switch shorthand doc gen over to new approach.

jamesls:
This PR updates the shorthand documenter to work with the latest changes to the shorthand code (https://github.com/aws/aws-cli/pull/1461).  Specifically:

* It does not use a hardcoded list of "shapes" like the previous approach did.
* It documents the newly supported syntax (nested dictionaries)

It also changes some small things in the way docs are generated:

* The `--argname` is removed in the shorthand example.  This is to make the shorthand syntax consistent with the JSON example syntax.
* The type names are added instead of generic `values`.  Before we'd have `Key=value,Foo=value`.   Now we'll have `Key=string,Foo=float`.
* The single sentence description has been removed.  It's not possible to automatically generate these.

And finally, there's a few areas of improvement I think we'll need to make in the future:

* Right now, shorthand examples are only generated for code of a stack depth of 3 or less.  This is because sufficiently complicated structures are not readable in shorthand syntax.  We'd need to come up with a way to break these across multiple lines.
* Recursive shapes are documented as `(... recursive ...)`, which is how JSON params are documented.  This might be an area for improvement.

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -368,18 +368,8 @@ class ParamShorthand(object):
 class ParamShorthandDocGen(object):
     """Documentation generator for param shorthand syntax."""
 
-    SHORTHAND_SHAPES = {
-        'structure(scalars)': '_key_value_parse',
-        'structure(scalar)': '_special_key_value_parse',
-        'structure(list-scalar, scalar)': '_struct_scalar_list_parse',
-        'map-scalar': '_key_value_parse',
-        'list-structure(scalar)': '_list_scalar_parse',
-        'list-structure(scalars)': '_list_key_value_parse',
-        'list-structure(list-scalar, scalar)': '_list_scalar_list_parse',
-    }
     _DONT_DOC = object()
     _MAX_STACK = 3
-
 
     def supports_shorthand(self, argument_model):
         """Checks if a CLI argument supports shorthand syntax."""
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -454,7 +454,7 @@ class ParamShorthandDocGen(object):
             stack.pop()
         if not stack:
             # Top of the stack means we're a top level shorthand param.
-            return '%s ...' % (element_docs,)# element_docs)
+            return '%s ...' % element_docs
         elif list_member.type_name in COMPLEX_TYPES or len(stack) > 1:
             return '[%s,%s]' % (element_docs, element_docs)
         else:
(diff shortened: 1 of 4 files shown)


awscli/argprocess.py
```diff
@@ -369,118 +377,116 @@ class ParamShorthandDocGen(object):
         'list-structure(scalars)': '_list_key_value_parse',
```
kyleknap: Is ``SHORTHAND_SHAPES`` still used?

awscli/argprocess.py
```diff
@@ -369,118 +377,116 @@ class ParamShorthandDocGen(object):
...
+                model.member.type_name == 'structure' and \
+                len(model.member.members) == 1:
+            member_name = list(model.member.members)[0]
+            return '%s %s1 %s2 %s3' % (cli_name, member_name,
+                                       member_name, member_name)
+        elif model.type_name == 'structure' and \
+                len(model.members) == 1 and \
+                'Value' in model.members and \
+                model.members['Value'].type_name == 'string':
+            return self._DONT_DOC
```
kyleknap: Do we have a tests for both of these special cases?

awscli/argprocess.py
```diff
@@ -369,118 +377,116 @@ class ParamShorthandDocGen(object):
...
+    def _list_docs(self, argument_model, stack):
+        list_member = argument_model.member
+        stack.append(list_member.name)
+        try:
+            element_docs = self._shorthand_docs(argument_model.member, stack)
+        finally:
+            stack.pop()
+        if not stack:
+            # Top of the stack means we're a top level shorthand param.
+            return '%s ...' % (element_docs,)# element_docs)
```
kyleknap: strange in-line comment

kyleknap:
It looks fine. Had a few comments, but I think they would have been caught with pr-check. Did you run it? It would be good to get the tests to pass as well. Probably will need to use ``OrderedDict``

awscli/argprocess.py
jamesls: Debugging artifact, will remove.

awscli/argprocess.py
jamesls: Nope, forgot to remove.  Will update this.

awscli/argprocess.py
jamesls: Yes, this was actually a backwards compat thing so I didn't have to add any new tests for this.  This is covered by the pre-existing tests.

jamesls:
Feedback should be incorporated. Build is passing.

Looks like there's one test case I missed, I need to add a doc for the changed list type, `<value1> ...`.  I'll get a test added for that.

jamesls:
@kyleknap Ok, I believe everything's been incorporated, including updated tests.  Should be good to look at again.

kyleknap:
Looks good. :ship:

PR #1468, opened 2015-08-26
Fix a bug when creating emr role with AWS_CA_BUNDLE

rayluo:
It is reported that there is a bug causing failure to create emr role,
when customer has an AWS_CA_BUNDLE setting in the environment.

When the bug occurs, running::

    aws emr create-default-roles

fails with "SSLError" object has no attribute.

This commit should be able to fix that, and also fix the incorrect test cases.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,15 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+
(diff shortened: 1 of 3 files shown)


jamesls:
:shipit:  Please add an entry to the changelog as well.  Thanks!

kyleknap:
Looks good. :ship:

PR #1469, opened 2015-08-27
Add role_session_name in config for assumed roles

lbernail:
Role Session Names were automatically assigned using the following
format: role_session_name = 'AWS-CLI-session-%s' % (int(time.time()))
This led to logs difficult to interprete in cloudtrail:
arn:aws:sts::xxxxx:assumed-role/myrole/AWS-CLI-session-1440675385
This commit adds a new optional parameter to the configuration
(role_session_name) which is used if present (otherwise use the
current format)

Small difference in behavior: when using generated role names, the
initial timestamp is used when refreshing

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -60,8 +60,6 @@ def create_assume_role_provider(session, provider_cls):
 
 def create_refresher_function(client, params):
     def refresh():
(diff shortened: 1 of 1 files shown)


lbernail:
Hello

I just saw that there was another PR on the same topic. Sorry for the additional one

Laurent

lbernail:
I had a look at the other pull request which is much more complete than mine
I am closing this one

Issue #1471, opened 2015-08-28
Filter do not support spaces


andibadu:
I updated the aws client to 1.8.0. Now, the script for creating an instance does not work any more.

aws --profile default --region us-west-1 ec2 describe-subnets --filters 'Name=subnet-id,Values=Private A' Name=vpc-id,Values=vpc-MYVPC --query 'Subnets[*].SubnetId' --output=text

The error message is:
Error parsing parameter '--filters': Expected: ',', received: 'A' for input:
Name=subnet-id,Values=Private A
                              ^

So there is a space in the subnet-id and I cannot filter it any more. 
The command worked with 1.7.x

jamesls:
Thanks for reporting.  We'll get this fixed.

jamesls:
Just want to mention that while we will fix the current behavior, the workaround for now is to double quote the value with spaces:


```
 --filters 'Name=subnet-id,Values="Private A"' Name=vpc-id,Values=vpc-MYVPC 
```

Issue #1470, opened 2015-08-28
Unable to use dash in option value


kimh:
Version: 1.8.0

The same command used to work at least in 1.7.42 but not in 1.8.0.

```
$ aws s3api create-multipart-upload --metadata "uncompressed-length=111" --bucket my-bucket --key my-key
Error parsing parameter '--metadata': Expected: '=', received: '-' for input:
uncompressed-length=111
```

Looks it complains `-` in uncompressed-length. Is this an expected change?

jamesls:
Not intended, this is a regression.  We'll get this fixed.  Thanks for reporting.

henrjk:
I think I am seeing the same issue. I observed this in a build which used to work. The builds were off for 2 or 3 months. The error showed for me with this version:

```console
aws --version
aws-cli/1.8.2 Python/2.7.3 Linux/3.13.0-57-generic
```

The command that fails is this:
```console
aws s3api put-object --bucket builds-shippables-doggie.example.net --key henkr/doggie/doggie-228.1.tar.gz.gpg --content-md5 +A+JCM/D88l35LqcQQl5aw== --metadata SHIPPABLE_BUILD_ID=55eeb529b2ed760b00cfe732,BUILD_URL=https://app.shippable.com/builds/55eeb529b2ed760b00cfe732,BRANCH=master,COMPARE_URL=https://bitbucket.org/henkr/doggie/branches/compare/bdca26b53565fbc08e52e7a2915c192cd4002dae..e4e4a14374bb#diff --body shippable/shippable.tar.gz.gpg

Error parsing parameter '--metadata': Expected: '=', received: '_' for input:
SHIPPABLE_BUILD_ID=55eeb529b2ed760b00cfe732,BUILD_URL=https://app.shippable.com/builds/55eeb529b2ed760b00cfe732,BRANCH=master,COMPARE_URL=https://bitbucket.org/henkr/doggie/branches/compare/bdca26b53565fbc08e52e7a2915c192cd4002dae..e4e4a14374bb#diff
         ^
```

Can you comment on:

1. what is the most recent release which does not have the issue?  

2. when do you project that the fix might be released? 

Thank you!

henrjk:
Just tried it with `aws-cli/1.8.3 Python/2.7.3 Linux/3.13.0-57-generic`. It was still failing for me.

jamesls:
@henrjk Looks like there's a regression with the `_` char as well.  I'll get that updated.  Thanks for reporting.

[...]

PR #1473, opened 2015-08-28
Fix shorthand parser regressions in 1.8.0

jamesls:
Fixes https://github.com/aws/aws-cli/issues/1471  (unquoted values with spaces should be allowed).
Fixes https://github.com/aws/aws-cli/issues/1470 (dash in key name should be allowed).

Added unit tests for these so we don't regress in the future.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,17 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+
+* bugfix:Shorthand Parser: Fix regression where '-' character was not accepted
+  as a key name in a shorthand value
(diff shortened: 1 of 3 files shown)


kyleknap:
Change looks fine. Is there anything else that we need to take account for other than dashes like other symbol or unicode keys? Otherwise, :ship:

jamesls:
The key would correspond to valid names (not values) for AWS APIs.  The only thing that I can think of that might affect this would be map names.  I'll do some additional research on this.

jamesls:
Actually, DynamoDB's `--items` didn't previously qualify for shorthand because it is a nested map type, so it wasn't possible to use shorthand for `put-item`.  I think we're good there.

kyleknap:
Ok cool. Should be good to merge then.

PR #1474, opened 2015-08-28
Update rb help documentation

kyleknap:
Doc update specifically targets behavior with versioned objects.

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -63,7 +63,12 @@ QUIET = {'name': 'quiet', 'action': 'store_true',
 
 FORCE = {'name': 'force', 'action': 'store_true',
          'help_text': (
-             "Deletes all objects in the bucket including the bucket itself.")}
+             "Deletes all objects in the bucket including the bucket itself. "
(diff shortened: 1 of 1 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -593,7 +598,10 @@ class MbCommand(S3TransferCommand):
 
 class RbCommand(S3TransferCommand):
     NAME = 'rb'
-    DESCRIPTION = "Deletes an S3 bucket."
+    DESCRIPTION = (
+        "Deletes an empty S3 bucket. A bucket must be completely empty "
+        "of objects and versioned objects before it can be deleted."
```
jamesls: This makes is seem like you can't use `rb` if you have non-versioned objects in your bucket.  Should this be reworded?  Or maybe `--force` should be mentioned here.

kyleknap: Yeah I will mention ``--force``.

kyleknap:
@jamesls I added the note about ``--force`` in the ``rb`` help page. Should be good to be looked at again.

jamesls:
:shipit:

Issue #1432, opened 2015-07-21
aws s3 rb --force deletes buckets of contents even if explicitly forbidden to delete bucket


stephen-mw:
I'm hoping to learn if this is expected behavior or a bug. Even after an explicit bucket policy that prevents a user from deleting a bucket, they can still run `rb --force` and delete the contents of the bucket. The command is aware that the user doesn't have permissions to remove the bucket but still allows it to recursively delete the files.

It seems to me the expected behavior should be to check for delete permissions on the bucket first, and then proceed with a recursive to delete if the user has permission to delete the bucket.
## How to reproduce

Create a test bucket and apply the following bucket policy

```
{
    "Version": "2012-10-17",
    "Id": "foo",
    "Statement": [
        {
            "Sid": "Stmt1437437343806",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:DeleteBucket",
            "Resource": "arn:aws:s3:::mybucket"
        }
    ]
}
```

Fill the bucket with files and then run the `aws s3 rb --force s3://mybucket` command. Despite it knowing it's forbidden from deleting the bucket, it will clear out all of the files recursively.

```
$ for i in `seq 1 10`; do touch test${i}; aws s3 cp test${i} s3://zz-placeholder33/;done
$ aws s3 rb s3://zz-placeholder33/
remove_bucket failed: s3://zz-placeholder33/ A client error (AccessDenied) occurred when calling the DeleteBucket operation: Access Denied
$ dataeng@cron-gateway[2032]~: aws s3 rb --force s3://zz-placeholder33/
delete: s3://zz-placeholder33/test5
delete: s3://zz-placeholder33/test9
delete: s3://zz-placeholder33/test1
delete: s3://zz-placeholder33/test2
delete: s3://zz-placeholder33/test10
delete: s3://zz-placeholder33/test3
delete: s3://zz-placeholder33/test4
delete: s3://zz-placeholder33/test8
delete: s3://zz-placeholder33/test
delete: s3://zz-placeholder33/test6
delete: s3://zz-placeholder33/test7
remove_bucket failed: s3://zz-placeholder33/ A client error (AccessDenied) occurred when calling the DeleteBucket operation: Access Denied
```

stephen-mw:
In addition, is it expected behavior that the ```aws s3 rb --force``` recursively remove a bucket even if the argument is a key within the bucket?

[...]

PR #1477, opened 2015-08-31
Refuse rb if path contains keys.

stephen-mw:
This update will make ```rb --force``` refuse to delete an S3 bucket if the user also specified keys. This makes recursive deletes more strict and should lead to less confusion and accidental deletes by some users.

If a user runs ```rb --force``` command on an s3path that contains keys, they'll get a message:
```
$ aws s3 rb --force s3://mybucket/myfolder/mykey

Please specify a valid bucket name only. E.g. s3://mybucket
```

Reference: https://github.com/aws/aws-cli/issues/1432

I added a unit test as per the CONTRIBUTIONS doc. Please let me know if anything is missing.

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -527,8 +527,9 @@ class S3TransferCommand(S3Command):
 
     def _handle_rm_force(self, parsed_globals, parameters):
         """
-        This function recursive deletes objects in a bucket if the force
-        parameters was thrown when using the remove bucket command.
+        This function recursively deletes objects in a bucket if the force
+        parameter was thrown when using the remove bucket command. It will
+        refuse to delete if a key is specified in the s3path.
         """
         # XXX: This shouldn't really be here.  This was originally moved from
         # the CommandParameters class to here, but this is still not the ideal
@@ -540,7 +541,10 @@ class S3TransferCommand(S3Command):
         # to be kept simple.
         if 'force' in parameters:
             if parameters['force']:
-                bucket = find_bucket_key(parameters['src'][5:])[0]
+                bucket, key = find_bucket_key(parameters['src'][5:])
+                if key:
+                    raise ValueError('Please specify a valid bucket name only.'
+                                     ' E.g. s3://%s' % bucket)
                 path = 's3://' + bucket
                 del_objects = RmCommand(self._session)
                 del_objects([path, '--recursive'], parsed_globals)
@@ -605,7 +609,7 @@ class RbCommand(S3TransferCommand):
         "the non-versioned objects in the bucket before the bucket is "
         "deleted."
     )
-    USAGE = "<S3Path>"
+    USAGE = "<S3BucketPath>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 1, 'positional_arg': True,
                   'synopsis': USAGE}, FORCE]
 
--- a/tests/unit/customizations/s3/test_subcommands.py
+++ b/tests/unit/customizations/s3/test_subcommands.py
@@ -49,25 +49,32 @@ class TestGetClient(unittest.TestCase):
 
 
 class TestRbCommand(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -64,6 +64,10 @@ def test_rb_command_with_force_deletes_objects_in_bucket(self):
             with mock.patch(arch_name):
                 rb_command._run_main(parsed_args,
                                      parsed_globals=parsed_globals)
+                with self.assertRaises(ValueError):
```
kyleknap: It would be nice if we had the assertion in a separate test. We typically try to avoid testing different functionality in the same test. So I would recommend adding a new test for this specific assertion in this test class or adding a functional test in this [directory](https://github.com/aws/aws-cli/tree/develop/tests/functional/s3) (Note that you would have to create your own ``test_rb_command.py`` file for the ``rb`` command and sort of copy the what the other tests do in that directory if you go down this route)

kyleknap:
Thank you for taking the initiative on this. It looks good. I had one comment about the tests that would be great to have addressed before merged.

On a side note if you want to go for bonus points, it would be nice if we can get this validation applied to the ``rb`` command in general and not just when ``--force`` is specified. So feel free to try that as well.

tests/unit/customizations/s3/test_subcommands.py
stephen-mw: :+1:

kyleknap:
Thanks for updating. It looks good. One more thing. It looks like I merged a pull request that caused a merge conflict with yours. Can you rebase off of the HEAD of the develop branch and update the PR? Once you do I will merge it.

stephen-mw:
Is there an issue on the current develop branch? Looks like these tests are failing.
```
test_aws_with_cacert_env_var (tests.unit.test_clidriver.TestHowClientIsCreated)
test_aws_with_endpoint_url 
test_aws_with_region
test_aws_with_verify_false
```

kyleknap:
There was a change in botocore and the CLI tests were not updated. We will get these fixed. Once they get fixed, I will restart the test and merge your pull request.

awscli/customizations/s3/subcommands.py
```diff
@@ -540,8 +541,11 @@ def _handle_rm_force(self, parsed_globals, parameters):
         # to be kept simple.
         if 'force' in parameters:
             if parameters['force']:
-                bucket = find_bucket_key(parameters['src'][5:])[0]
-                path = 's3://' + bucket
+                bucket = find_bucket_key(parameters['src'][5:])
+                if bucket[1]:
```
stephen-mw: @kyleknap what do you think about line this line? The ```find_bucket_key``` function always returns an array with two elements, with the second element sometimes being empty. This looks like it may throw an ```IndexError``` by it's really just checking to see if the second element is an empty string.

kyleknap: This line is fine. The find_bucket_key returns a tuple of two elements, the bucket and the key. So ``bucket[1]`` results in returning the value for the key. So for example take a look at this snippet:
```py
In [1]: from awscli.customizations.s3.utils import find_bucket_key

In [2]: find_bucket_key('mybucket/')
Out[2]: ('mybucket', '')

In [3]: find_bucket_key('mybucket')
Out[3]: ('mybucket', '')

In [4]: find_bucket_key('mybucket/mykey')
Out[4]: ('mybucket', 'mykey')
```
If you further tried to index ``bucket[1]``, you may run into an IndexError, but you do not. So that is fine.

One thing that you could do to make it more clear is make the ``find_bucket_key`` line the following:
```py
bucket, key = find_bucket_key(parameters['src'][5:])
if key:
    ...
```
Keeping it the name bucket is a little misleading because in reality it is a tuple. Otherwise, I am fine with this PR. I can merge it unless you want to change the line. Let me know what you think.

stephen-mw:
Thanks @kyleknap that was just the readability change I was looking for.

kyleknap:
Cool. Looks good. Merging.

PR #1486, opened 2015-09-02
Add test for logs get-log-events command

kyleknap:
For making sure we do not add the GetLogEvents paginator back again until we actually add support for it. Related to https://github.com/boto/botocore/pull/641


mtdowling:
:ship:

PR #1488, opened 2015-09-02
Change test cases to accommodate botocore timeouts

rayluo:
This is needed after we merge [this botocore timeout PR](https://github.com/boto/botocore/pull/634).

The test cases are now organized in a way that it won't break if we add more
parameters in endpoint in future.

### DIFF
--- a/tests/unit/test_clidriver.py
+++ b/tests/unit/test_clidriver.py
@@ -620,17 +620,15 @@ class TestHowClientIsCreated(BaseAWSCommandParamsTest):
         self.assert_params_for_cmd(
             'ec2 describe-instances --endpoint-url https://foobar.com/',
             expected_rc=0)
-        self.create_endpoint.assert_called_with(
-            mock.ANY, 'us-east-1', verify=None, endpoint_url='https://foobar.com/',
-            is_secure=True, response_parser_factory=mock.ANY)
+        self.assertEqual(self.create_endpoint.call_args[1]['endpoint_url'],
+                         'https://foobar.com/')
 
     def test_aws_with_region(self):
         self.assert_params_for_cmd(
             'ec2 describe-instances --region us-west-2',
             expected_rc=0)
-        self.create_endpoint.assert_called_with(
-            mock.ANY, 'us-west-2', verify=None, endpoint_url=None,
-            is_secure=True, response_parser_factory=mock.ANY)
+        self.assertEqual(self.create_endpoint.call_args[0],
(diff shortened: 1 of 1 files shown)


tests/unit/test_clidriver.py
```diff
@@ -620,17 +621,23 @@ def test_aws_with_endpoint_url(self):
...
             expected_rc=0)
-        self.create_endpoint.assert_called_with(
-            mock.ANY, 'us-east-1', verify=None, endpoint_url='https://foobar.com/',
-            is_secure=True, response_parser_factory=mock.ANY)
+        call_args = self.create_endpoint.call_args
+        self.assertEqual(call_args[0], (mock.ANY, 'us-east-1'))
+        expected = dict(
+            timeout=(DEFAULT_TIMEOUT, DEFAULT_TIMEOUT),
+            verify=None, endpoint_url='https://foobar.com/', is_secure=True)
+        self.assertTrue(set(expected.items()) <= set(call_args[1].items()))
```
jamesls: Rather than do this, why not take the same approach you did with just verifying the specific kwarg the test is interested in?  So here we'd just check kwargs['endpoint_url']?

kyleknap:
I would prefer that we make the assertion specific as well, like it used to be:
```py
self.create_endpoint.assert_called_with(
    mock.ANY, 'us-east-1', verify=None, endpoint_url='https://foobar.com/',
    timeout=(DEFAULT_TIMEOUT, DEFAULT_TIMEOUT),
    is_secure=True, response_parser_factory=mock.ANY)
```

I know it is a pain that we have to update everytime something is added to ``create_endpoint`` but these tests are asserting specific settings were set based on parameters, and so they need to be checked

rayluo:
So I think the direction to go depends on our answers to the following several questions. See the pseudo code below.

    if we_want("the cli test cases to fail when we change sth in botocore")==True:
        do("original way, with one more parameter added each time")
    elif we_want("each of the current test cases is only testing one aspect")==True:
        do("we only check that one parameter in each test case, respectively")
    elif we_want("all known behaviors are checked for any given test case, yet future proof")==True:
        do("accept this PR")
    else:
        new_idea = raw_input("Your thoughts?")

jamesls:
Option 2.

Let's get this updated quickly because this is fixing failing builds.  It's starting to block other pull requests from being reviewed/merged.  We can change reassess our approaches later.

jamesls:
:shipit:  Thanks!

Issue #1489, opened 2015-09-03
cloud-formation validate-template example doesn't work


glb:
The example command for `cloud-formation validate-template` uses `--template-body` to reference a local file:// URL, which doesn't work:
```
aws cloudformation validate-template --template-body file:./sample.template

A client error (ValidationError) occurred when calling the ValidateTemplate operation: Template format error: JSON not well-formed. (line 1, column 5)
```

I'm using aws-cli/1.8.1. 

It looks from the rest of the docs that the example meant to be using `template-url`, which would also require using a URL pointing to an object in an AWS S3 bucket in the same region as the template.

mtdowling:
`--template-body` is a valid way to supply a template. When using a local file URI, you must include `file://` followed by your filename. What you have right now is not being picked up as a file URI because you have provided an invalid URI, which is being interpreted as a JSON document that has the contents `file:./sample.template`. Change your `--template-body` parameter to `file://./sample.template` and it will work.

glb:
OK, thanks for clarifying!

boompig:
This is unbelievably confusing and a very poor design choice.

While

```
aws cloudformation validate-template --template-body S3_Bucket.template
```

fails with the extremely cryptic error message

> A client error (ValidationError) occurred when calling the ValidateTemplate operation: Template format error: unsupported structure.

The following command

```
aws cloudformation validate-template --template-body file://S3_Bucket.template --region us-west-1
```

succeeds.

There should either be a more informative error message, or the `template-body` argument should be interpreted as a local file URI by default.

bamapookie:
@boompig That is because without the file URL syntax, the filename you entered is being evaluated as the template itself.  The file name is not a valid JSON or YAML file.  You could replace the filename in your first example with the actual JSON and it would evaluate it in place.

washad:
@boompig But I hope you would agree that it wouldn't be difficult to look at the length of the argument and recognize that not to be the case and supply a more useful error message. Even a hint such as (did you forget to include 'file:') would be nice. The existing error message not only does nothing to let the user know what the real problem is but misleads her or him into thinking they have an invalid template -- which is arguably worse.

boompig:
@bamapookie Thank you for your explanation.

>  the filename you entered is being evaluated as the template itself.

But *why* is that the case? This violates the logic assumption of the interface user.

Also, while your explanation helps explain the reason the observed behaviour occurs, it does not address:

[...]

PR #1490, opened 2015-09-03
Fix the example validate-template command

glb:
- The original example referenced a local file URL, which is not
  supported by the validate-template command. Switched to use an
  S3 URL.

Fixes aws/aws-cli#1489.


mtdowling:
The current example is valid and shows how to use a local file, so it should remain in the docs. As shown in the related issue, it is valid to provide a local file URI in order to provide a template on disk. You just need to ensure that your are using the `file` scheme followed by `//` in order for the URI parser to understand the URI.

PR #1491, opened 2015-09-03
Example of authorizing security group ingress from elb source

jraidan:
Added ec2 authorize ingress example for ELB source traffic. I've been asked how to do this quite a few times and thought it would be helpful to have an example.


mtdowling:
Thanks!

PR #1496, opened 2015-09-10
Don't print response only if an empty dict/list/str

jamesls:
The integer 0 should be printed as a result.  This can happen
from a JMESPath expression, e.g `length(something)`.

For example, currently:

```
$ aws ec2 describe-security-groups --query 'length(`[0,1]`)'
2
$ aws ec2 describe-security-groups --query 'length(`[0]`)'
1
$ aws ec2 describe-security-groups --query 'length(`[]`)'
<prints nothing>
$
```

Now you'll get:

```
$ aws ec2 describe-security-groups --query 'length(`[]`)'
0
```

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -23,6 +23,7 @@ advance_iterator = six.advance_iterator
 PY3 = six.PY3
 queue = six.moves.queue
 shlex_quote = six.moves.shlex_quote
+string_types = six.string_types
 StringIO = six.StringIO
 urlopen = six.moves.urllib.request.urlopen
 
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -23,7 +23,6 @@ advance_iterator = six.advance_iterator
 PY3 = six.PY3
 queue = six.moves.queue
 shlex_quote = six.moves.shlex_quote
-string_types = six.string_types
 StringIO = six.StringIO
 urlopen = six.moves.urllib.request.urlopen
 
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -92,11 +92,10 @@ class JSONFormatter(FullyBufferedFormatter):
         # the response will be an empty string.  We don't want to print
         # that out to the user but other "falsey" values like an empty
         # dictionary should be printed.
-        if isinstance(response, (dict, list) + string_types) and not response:
-            return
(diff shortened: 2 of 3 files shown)


awscli/formatter.py
```diff
@@ -91,10 +92,11 @@ def _format_response(self, command_name, response, stream):
         # the response will be an empty string.  We don't want to print
         # that out to the user but other "falsey" values like an empty
         # dictionary should be printed.
-        if response:
```
mtdowling: Would it be possible to just do `if response or response == 0:`, or are we worried about other types that are falsey but should show up in the output?

jamesls: After looking at this again, I think we can just change this to explicitly check for '{}'.

The original intent of this (years ago) was that any AWS operation that did not have an associated output shape should *not* generate any CLI output.  In older version of the CLI, some operations could generate either an empty dict or empty string.

However, with the switchover to the normalized model parser, anytime an operation does not have an output shape, botocore gives back a dictionary that only has the RequestMetadata key populated.  Once the formatter's here strip that key out, we're left with an empty dictionary.

So it should be safe to update this to only check for an empty dictionary.  This meets the constraint that anything without an output shape should not generate CLI output.

Does that work for you?

mtdowling: That makes sense. Works for me.

rayluo: Makes sense. By the way, will we also document this philosophy somewhere? Such as in the code as a comment paragraph?

awscli/compat.py
```diff
@@ -23,6 +23,7 @@
 PY3 = six.PY3
 queue = six.moves.queue
 shlex_quote = six.moves.shlex_quote
+string_types = six.string_types
```
mtdowling: This is no longer necessary

awscli/formatter.py
```diff
@@ -20,6 +20,7 @@
 from awscli.table import MultiTable, Styler, ColorizedStyler
 from awscli import text
 from awscli import compat
+from awscli.compat import string_types
```
mtdowling: I don't think this is necessary

jamesls:
@mtdowling Feedback incorporated.

mtdowling:
:ship:

PR #1511, opened 2015-09-17
Add STANDARD_IA as a valid enum choice to s3 commands

jamesls:
This is supported in the s3api commands, but we need to
add the valid value up to the s3 commands.


kyleknap:
:ship:

PR #1512, opened 2015-09-18
Point shared creds to bogus file

jamesls:
Otherwise you will get test failures if you
have a valid shared creds file configured.


kyleknap:
:ship:

Issue #1498, opened 2015-09-11
Downgrading rsa package


orome:
The awscli install or update (`pip install -U awscli `) is downgrading rsa (to 3.1.4).

rayluo:
Yes this is the real reason causing your issue in botocore. Again, we will need to test the new version rsa before we can upgrade.

orome:
Sorry to ask here but perhaps you can direct me to the right issues list for this (similar) issue:

The awsebcli package installation is again downgrading other packages:

    cement to 2.40
    docker-py to1.1.0
    pathspec to 0.3.3
    requests to 2.6.2

jamesls:
Diff looks fine, let's bump the version up to `<3.3`: https://bitbucket.org/sybren/python-rsa/branches/compare/version-3.2%0Dversion-3.1.4#diff

@orome For ebcli the best place to reach out would be the [AWS Elastic Beanstalk forums](https://forums.aws.amazon.com/forum.jspa?forumID=86).

jamesls:
This has now been fixed.

PR #1514, opened 2015-09-18
Bump RSA package to support 3.2.x

jamesls:
The diff didn't have any functional changes.
Given that, I think it makes sense to keep the
min version unchanged for now to give downstream
consumers time to upgrade.  We should eventually
just lock to 3.2.x only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ import awscli
 requires = ['botocore==1.2.4',
             'colorama>=0.2.5,<=0.3.3',
             'docutils>=0.10',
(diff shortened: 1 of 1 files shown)


jamesls:
Forgot to link the issue:

Fixes #1498 .  Diff between last and current release: https://bitbucket.org/sybren/python-rsa/branches/compare/version-3.2%0Dversion-3.1.4#diff

kyleknap:
Diff looks fine. :ship:

mtdowling:
:ship:

PR #1515, opened 2015-09-18
Catch more specific exceptions in assume role provider

jamesls:
To repro, call an aws command with an explicit ``--region`` and
a non-existent profile:

  aws ec2 describe-instances --region us-east-1 --profile does-not-exist

The issue was that the assume role provider was catching too general
of an exception so the profile does not exist error was not being
propogated.

There's also a bug in botocore with how we're handling failures
for lazily registered components.  This is why we're seeing the
"Unknown component" bug.  This will require a separate PR
to botocore.

Before:

```
$ aws ec2 describe-instances --profile does-not-exist --region us-east-1

Unknown component: credential_provider
```

After:

```
$ aws ec2 describe-instances --profile does-not-exist --region us-east-1

The config profile (does-not-exist) could not be found
```

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -11,6 +11,7 @@ from dateutil.tz import tzlocal
 from botocore import credentials
 from botocore.compat import total_seconds
 from botocore.exceptions import PartialCredentialsError
+from botocore.exceptions import UnknownCredentialError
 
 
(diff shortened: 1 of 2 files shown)


kyleknap:
Looks good. :ship:

mtdowling:
:ship:

PR #1516, opened 2015-09-18
Fix syncing across regions

JordonPhillips:
Performing a sync operation between buckets in different regions while under sig4 was raising errors when the delete flag was added in. A client associated with the origin bucket's region was being used to make the delete request rather than a client associated with the destination bucket's region. This simply switched the two and adds in some tests to verify.

### DIFF
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -298,7 +298,7 @@ class FileInfo(TaskInfo):
         Deletes the file from s3 or local.  The src file and type is used
         from the file info object.
         """
-        if (self.src_type == 's3'):
+        if self.src_type == 's3':
             bucket, key = find_bucket_key(self.src)
             params = {'Bucket': bucket, 'Key': key}
             self.source_client.delete_object(**params)
--- a/awscli/customizations/s3/fileinfobuilder.py
+++ b/awscli/customizations/s3/fileinfobuilder.py
@@ -42,8 +42,26 @@ class FileInfoBuilder(object):
         file_info_attr['src_type'] = file_base.src_type
         file_info_attr['dest_type'] = file_base.dest_type
         file_info_attr['operation_name'] = file_base.operation_name
-        file_info_attr['client'] = self._client
-        file_info_attr['source_client'] = self._source_client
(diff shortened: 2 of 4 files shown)


awscli/customizations/s3/fileinfo.py
```diff
@@ -301,7 +301,7 @@ def delete(self):
         if (self.src_type == 's3'):
             bucket, key = find_bucket_key(self.src)
             params = {'Bucket': bucket, 'Key': key}
-            self.source_client.delete_object(**params)
+            self.client.delete_object(**params)
```
kyleknap: Ahh now I remember why it is not simple as this. So this logic was put in place for the bucket to bucket ``mv`` command. When we move an object from one bucket to another, we first copy the object and then delete the other object. We must send the copy request to the destination region but the delete must happen from the source. If my assumption is correct, your PR will fail if you do a cross-region sigv4 enabled mv command. I think the change to get this fixed needs to happen in the [DeleteSync](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/syncstrategy/delete.py) strategy. Might be a little tricky though because a ``FileStat`` object does not know its client. I would also recommend looking in adding it in the ``FileInfoBuilder`` object as that is where the client gets injected.

kyleknap:
I would recommend you test out my suggestion to see if my prediction is correct about the mv command or even add an integration test for the future to ensure it never happens. Let me know what you find out.

JordonPhillips:
You are correct. I put in an integration test that fails for now, I'll look into the strategy.

JordonPhillips:
The delete sync strategy doesn't seem like the right place to put it. There's no manipulation I can do to the filestat object to correct the issue. I can make it try to send the request to the wrong region AND bucket, but that's obviously not desirable. FileInfoBuilder might work, though I'm worried about putting conditional checks in there.

JordonPhillips:
In the sync strategy I know it's a sync, but can't change anything useful. In the builder, I can change things but I don't know that it's a sync

JordonPhillips:
The `--delete` flag in `fileinfobuilder.py` will work for now as a distinguisher between syncs and other types of delete.

kyleknap:
Looks good. Thanks for looking into this. :ship:

PR #1526, opened 2015-09-23
Fixed include / exclude filters in bucket to bucket sync

### LINKED ISSUE #1481: aws s3 sync --delete behaves differently if source is <LocalPath> rather than <S3Path>
This issue was raised on: https://stackoverflow.com/questions/32304326/amazon-s3-sync-deleting-excluded-files

Basically, doing `aws s3 sync . s3://bucket2 --delete` behaves differently to `aws s3 sync s3://bucket1 s3://bucket2 --delete` even if the two source directories are identical.

It would appear that the logic for `--delete` behaves differently between Local and S3. An example is well-documented in the stackoverflow issue.

JordonPhillips:
closes #1481 

The logic for setting the root directory for filters only supported s3-local and local-s3 syncs, not s3-s3, causing the filters to fail when applied to the destination bucket in s3-s3 since it was assumed the destination was  a local directory.

### DIFF
--- a/tests/unit/customizations/s3/test_filters.py
+++ b/tests/unit/customizations/s3/test_filters.py
@@ -202,6 +202,26 @@ class FiltersTest(unittest.TestCase):
         filtered = list(s3_filter.call(s3_files))
         self.assertEqual(len(filtered), 0)
 
+    def test_create_filter_s3_to_s3(self):
+        source = 'bucket/'
+        destination = 'bucket-2/'
+        pattern = '*'
+        parameters = {'filters': [['--exclude', pattern],
+                                  ['--include', '*.jpg']],
+                      'dir_op': True,
+                      'src': 's3://' + source,
+                      'dest': 's3://' + destination}
+        s3_filter = self.create_filter(parameters=parameters)
+
+        source_pattern = s3_filter.patterns[0][1]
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/s3/test_filters.py
```diff
@@ -202,6 +202,21 @@ def test_create_root_s3_no_dir_op(self):
...
+        parameters = {'filters': [['--exclude', pattern]],
+                      'dir_op': True,
+                      'src': 's3://' + source,
+                      'dest': 's3://' + destination}
+        s3_filter = self.create_filter(parameters=parameters)
+        source_pattern = s3_filter.patterns[0][1]
+        destination_pattern = s3_filter.dst_patterns[0][1]
+        self.assertEquals(source_pattern, source + pattern)
+        self.assertEquals(destination_pattern, destination + pattern)
+
```
kyleknap: It would also be nice to test that the filter is applied correctly as well. So maybe pass some file stat objects that represent what the bug was and show that the issue is fixed.

kyleknap:
Implementation looks fine. Just had a comment about the test.

JordonPhillips:
:+1:

kyleknap:
Looks good. Thanks. :ship:

PR #1531, opened 2015-09-25
Update doc building documentation

christophermanning:
Building the docs seems to require Python 2.7 and the awscli to be installed. I've added those to the README as requirements. I used this docker command to test the build in an isolated environment:

```
docker run -v $(pwd):/src python:2.7.10 /bin/bash -c \
  "cd /src && pip install -r requirements.txt && pip install awscli && cd doc && make clean && make html"
```

Building the docs also puts build artifacts in `doc/source/topic` so I've ignored that directory in `.gitignore`.

Finally, I fixed the syntax of `print` to work with Python 3. Though, building with different Python versions gives different html output (e.g. urls in links are root relative when building with Python 3).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,5 @@ src/
 # CLI docs generation
 doc/source/aws_man_pages.json
 doc/source/reference
+doc/source/topic
 doc/source/tutorial/services.rst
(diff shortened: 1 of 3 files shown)


jamesls:
:shipit:

PR #1533, opened 2015-09-28
Override package deps when building wheels

jamesls:
Make sure that argparse has the env marker of
python2.6 only when creating a wheel.  Same
fix as https://github.com/boto/boto3/pull/282.


kyleknap:
:ship:

kyleknap:
botocore dependency now needs to be updated due to yesterday's release.

PR #1534, opened 2015-09-29
Add compression for pushed files on platforms that support zlib

### LINKED ISSUE #1371: [BUG] [aws deploy push] The Destination Bundle Type Does not Perform Compression (Zipping/Tarring)
When specifying a destination with a suffix of ".zip" or ".tar" or ".tar.gz" for use with the `aws deploy push` command ... which are all allowed revisions ... the resulting bundle is not compressed.

Either the extension of the bundle could automatically determine the desired compression or an alternate flag could be specified which determines that the actual file should be compressed using that appropriate mechanism.

I noticed this testing large deployments that  were filling up S3 very fast --- when compressed it would have been not noticeable.

Additionally, directories with no files in them do not appear in the resulting bundle.

JordonPhillips:
Closes #1371

The user's python must have access to the zlib module which is default on the vast majority of python installations. A small number of linux distributions may not have it by default. If they don't have it, we fall back to storage mode.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 import sys
 import os
+import zipfile
 
 from botocore.compat import six
 #import botocore.compat
@@ -25,6 +26,15 @@ queue = six.moves.queue
 shlex_quote = six.moves.shlex_quote
 StringIO = six.StringIO
 urlopen = six.moves.urllib.request.urlopen
+
+# Most, but not all, python installations will have zlib. This is required to
+# compress any files we send via a push. If we can't compress, we can still
+# package the files in a zip container.
+try:
+    import zlib
+    ZIP_COMPRESSION_MODE = zipfile.ZIP_DEFLATED
+except ImportError:
+    ZIP_COMPRESSION_MODE = zipfile.ZIP_STORED
 
 
 class BinaryStdout(object):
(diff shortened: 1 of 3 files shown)


awscli/customizations/codedeploy/push.py
```diff
@@ -22,7 +22,7 @@
 from awscli.customizations.codedeploy.utils import validate_s3_location
 from awscli.customizations.commands import BasicCommand
 from awscli.errorhandler import ServerError, ClientError
-
```
kyleknap: pep8: make sure there are two lines separating imports from where the code logic begins

awscli/compat.py
```diff
@@ -26,6 +27,15 @@
 StringIO = six.StringIO
 urlopen = six.moves.urllib.request.urlopen
 
+# Most, but not all, python installations will have zlib. This is required to
+# compress any files we send via a push. If we can't compress, we can still
+# package the files in a zip container.
+try:
+    import zlib
+    COMPRESSION_MODE = zipfile.ZIP_DEFLATED
```
kyleknap: Can we rename it to something specific to zipfile's like ``ZIP_COMPRESSION_MODE`` to be more explicit as it is located in compat.py module which is shared all across the CLI codebase?

kyleknap:
Looks good. A couple of very small comments.

JordonPhillips:
:+1:

kyleknap:
LGTM :ship:

PR #1536, opened 2015-09-30
Update Sphinx version to 1.3.1

dafyddcrosby:
Got errors building `make man` and `make text` on 1.1.3, fixed by updating Sphinx to 1.3.1


JordonPhillips:
We've updated our range to support 1.3, but it seems we forgot to close this out. Thanks for pointing out the issue!

PR #1541, opened 2015-10-02
Test fixes for Python 2.6 and 3.3

mtdowling:
Python 2.6 does not support opening gzip files in a context handler, so just opening and closing manually now.

Python 3.3 has issue with the "hex" codec (see http://bugs.python.org/issue7475). Now using binascii instead.


jamesls:
:shipit:

PR #1554, opened 2015-10-11
Also include botocore version in "aws --version"

rayluo:
Although it may not happen often in end user's daily scenario,
I did occasionally get bitten by my aws-cli and botocore version becoming out
of sync.  Until now the only way to find out is to append "--debug" at a dummy
aws command line (typically "aws s3 ls") and observe the first line of output.
This commit adds botocore version into the canonical "aws --version" command.

I test it manually to see this output.

    $ aws --version
    aws-cli/1.8.12 Python/2.7.10 Darwin/14.5.0 botocore/1.2.10

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -154,7 +154,8 @@ class CLIDriver(object):
         command_table['help'] = self.create_help_command()
         cli_data = self._get_cli_data()
         parser = MainArgParser(
-            command_table, self.session.user_agent(),
+            command_table,
+            "%s botocore/%s" % (self.session.user_agent(), botocore_version),
             cli_data.get('description', None),
             cli_data.get('synopsis', None),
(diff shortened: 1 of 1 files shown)


quiver:
I have one question : Why you didn't `change session.user_agent` itself?
Your PR only changes the output of `$ aws --version` and does not affect request header's `User-Agent`.

rayluo:
Actually the previous implementation tried purposely to limit the impact to the visual output of CLI only, so that nothing else would break.  But @quiver 's suggestion is good, and now it is implemented.

I have also manually tested "aws --version" to see botocore version.
I also tested "aws s3 ls --debug" to see no duplicated botocore version.

kyleknap:
:ship:

quiver:
Nice work, @rayluo

Issue #774, opened 2014-05-14
aws-cli s3 cp\sync 'ascii' codec can't decode byte 0xe0 in position 0: ordinal not in range(128)


kroniak:
Hello!

aws-cli/1.3.10 Python/2.7.5 Windows/8

Today I tried upload my files to s3 and I have this error on all my files:
'ascii' codec can't decode byte 0xe0 in position 0: ordinal not in range(128)

If --no-guess-mime-type enabled, is no errors.

Command: aws s3 cp . s3://...

Log with debug:

2014-05-14 13:55:05,302 - botocore.hooks - DEBUG - Event service-data-loaded.s3: calling handler <function signature_overrides at 0x00000000029D3E48>
2014-05-14 13:55:05,305 - botocore.hooks - DEBUG - Event service-created: calling handler <function register_retries_for_service at 0x00000000029D3C88>
2014-05-14 13:55:05,305 - botocore.handlers - DEBUG - Registering retry handlers for service: Service(s3)
2014-05-14 13:55:05,306 - awscli.customizations.s3.tasks - DEBUG - 'ascii' codec can't decode byte 0xe0 in position 0: ordinal not in range(128)
Traceback (most recent call last):
  File "awscli\customizations\s3\tasks.pyc", line 88, in _execute_task
  File "awscli\customizations\s3\fileinfo.pyc", line 219, in upload
  File "awscli\customizations\s3\fileinfo.pyc", line 191, in _handle_object_params
  File "awscli\customizations\s3\fileinfo.pyc", line 227, in _inject_content_type
  File "awscli\customizations\s3\utils.pyc", line 211, in guess_content_type
  File "mimetypes.pyc", line 294, in guess_type
  File "mimetypes.pyc", line 355, in init
  File "mimetypes.pyc", line 259, in read_windows_registry
  File "mimetypes.pyc", line 249, in enum_types
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe0 in position 0: ordinal not in range(128)
2014-05-14 13:55:05,309 - awscli.customizations.s3.executor - DEBUG - Submitting task: <awscli.customizations.s3.tasks.BasicTask object at 0x000000000332E550>
2014-05-14 13:55:05,312 - botocore.service - DEBUG - Creating service object for: s3
2014-05-14 13:55:05,312 - awscli.customizations.s3.executor - DEBUG - Worker thread invoking task: <awscli.customizations.s3.tasks.BasicTask object at 0x000000000332E550>
[...]

PR #1570, opened 2015-10-19
Catch a bug in mimetype library

kyleknap:
Issue was that on windows some MIME types in the Windows library could not be encoded using the system default encoding. This has been fixed in later versions of python, but we should not hard fail if the user is using a python version without the fix. Instead just catch the exception and not set the content type.

Fixes https://github.com/aws/aws-cli/issues/774

### DIFF
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -1,4 +1,5 @@
 import os
+import logging
 import sys
 import time
 from functools import partial
@@ -11,6 +12,9 @@ from dateutil.tz import tzlocal
 from botocore.compat import quote
 from awscli.customizations.s3.utils import find_bucket_key, \
     uni_print, guess_content_type, MD5Error, bytes_print
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class CreateDirectoryError(Exception):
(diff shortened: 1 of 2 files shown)


tests/functional/s3/test_cp_command.py
```diff
@@ -163,6 +163,15 @@ def test_no_metadata_directive_for_non_copy(self):
...
-if __name__ == "__main__":
-    unittest.main()
+    def test_cp_succeeds_with_mimetype_errors(self):
+        full_path = self.files.create_file('foo.txt', 'mycontent')
+        cmdline = '%s %s s3://bucket/key.txt' % (self.prefix, full_path)
+        self.parsed_responses = [
+            {'ETag': '"c8afdb36c52cf4727836669019e69222"'}]
+        with mock.patch('mimetypes.guess_type') as mock_guess_type:
+            # This should throw a UnicodeDecodeError.
+            mock_guess_type.side_effect = lambda x: b'\xe2'.decode('ascii')
```
kyleknap: I would be up for suggestions on a better way to throw the ``UnicodeDecodeError``. I was having troubles instantiate directly. Think its due to the fact that is a built-in C class.

awscli/customizations/s3/fileinfo.py
```diff
@@ -266,7 +270,22 @@ def _handle_upload(self, body):
...
+        # https://bugs.python.org/issue9291
+        #
+        # So instead of hard failing, just log the issue and fall back to the
+        # default guessed content type of None.
+        except UnicodeDecodeError:
+            LOGGER.debug(
+                'Unable to guess content type for %s due to '
+                'UnicodeDecodeError: ', filename, exc_info=True
+            )
         if guessed_type is not None:
```
mtdowling: I think this should be moved in the try block before the except block. We only need to check if `guessed_type` is not None when a unicode exception was not raised.

kyleknap: That would work to. I default it to ``None`` before the try except. I would be fine with either.

mtdowling:
Looks good. One small comment.

rayluo:
:ship:

mtdowling:
:ship:

Issue #1564, opened 2015-10-16
s3 sync from local source directory that doesn't exist behavior


prsullivan:
Well I think I totally borked myself. I was using a chron job to run s3 sync on a nightly basis to transfer files form a local directory on a mounted secondary hard drive to an s3 bucket. Last week this hard drive crashed and would not mount. I thought that since the local directory didn't exist the s3 sync command would error out. Noooooope. It synced an empty directory, and to complete perfect storm my command included the --delete option, which totally wiped out my backup. Is this intended behavior?

JordonPhillips:
I'm very sorry you had this issue, I know how frustrating it can be to lose your backups. I tested this out myself and can confirm. For posterity, here's an example command:

`aws s3 sync --delete ~/does/not/exist s3://test-bucket`

I did some brief investigating into other solutions in case something accidentally gets updated. Unfortunately it seems that there is no way to have an 'undelete' on s3 without mirroring. There is versioning available, but previous versions crop up when you delete the new ones, so in your case all your files would eventually get deleted anyway.

Off the top of my head, I could imagine a scenario where you replicate all your writes (possibly using Lambda) to a different s3 bucket whose life cycle policy immediately archives the files to Glacier. Glacier storage is vastly cheaper, so the cost wouldn't be too bad.

Still, the best solution is to fix the bug. We'll definitely be looking into it.

prsullivan:
Here is the output:

> sullivan@homeserv:~$ aws --version
> aws-cli/1.7.32 Python/2.7.6 Linux/3.16.0-50-generic

Thanks

kyleknap:
So the behavior is intended in terms of how delete works. There were no files identified in the local directory and there were files in the s3 bucket, and those were deleted because they existed in the s3 bucket and not at the local file system.

Erroring out on nonexistant directories and files has caused issues in the past: https://github.com/aws/aws-cli/issues/856

Also if you are syncing to a non existent folder, it is valid to do something like this:
```
$ aws s3 sync s3://mybucketfoo1/ nonexistentfolder
warning: Skipping file /Users/kyleknap/Documents/GitHub/aws-cli/foo/nonexistentfolder/. File does not exist.
download: s3://mybucketfoo1/bar.txt to nonexistentfolder/bar.txt
```

[...]

PR #1575, opened 2015-10-20
Add some path checks around user provided paths

kyleknap:
This mainly applies to the existence of user provided local paths. If the transfer is an upload and that local path does not exist, this will cause the command to fail instead of skipping. I realize this will cause a change in status code, but I think a hard fail is the correct behavior. Skipping on a specific path that the user specified should not be happening. 

Also, If the transfer is a syncing download and the local path does not existent, there should not be warnings.

All integration test pass as well.

Fixes https://github.com/aws/aws-cli/issues/1082, https://github.com/aws/aws-cli/issues/1564

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -130,24 +130,11 @@ class FileGenerator(object):
         ``dir_op`` and ``use_src_name`` flags affect which files are used and
         ensure the proper destination paths and compare keys are formed.
         """
+        function_table = {'s3': self.list_objects, 'local': self.list_files}
         source = files['src']['path']
         src_type = files['src']['type']
         dest_type = files['dest']['type']
-        if src_type == 'local':
-            direction_context = src_type + dest_type
-            # If the file generator has no associated operation, it
-            # is not the driving force behind an operation. That means
-            # it is being used as the other generator (the one being used
-            # for listing of files in the destination) in a sync command.
-            # More specifically this if statement will only be hit if it
-            # is the file generator listing files for a downloading sync.
-            if self.operation_name == '':
-                direction_context = dest_type + src_type
-            file_list = self.list_files(
-                source, files['dir_op'],
-                transfer_direction_context=direction_context)
-        else:
-            file_list = self.list_objects(source, files['dir_op'])
+        file_list = function_table[src_type](source, files['dir_op'])
         for src_path, size, last_update in file_list:
             dest_path, compare_key = find_dest_path_comp_key(files, src_path)
             yield FileStat(src=src_path, dest=dest_path,
@@ -156,7 +143,7 @@ class FileGenerator(object):
                            dest_type=dest_type,
                            operation_name=self.operation_name)
 
-    def list_files(self, path, dir_op, transfer_direction_context=None):
+    def list_files(self, path, dir_op):
         """
         This function yields the appropriate local file or local files
         under a directory depending on if the operation is on a directory.
@@ -167,7 +154,7 @@ class FileGenerator(object):
         """
         join, isdir, isfile = os.path.join, os.path.isdir, os.path.isfile
         error, listdir = os.error, os.listdir
-        if not self.should_ignore_file(path, transfer_direction_context):
+        if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
                 yield path, size, last_update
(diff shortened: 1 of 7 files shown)


awscli/customizations/s3/filegenerator.py
```diff
@@ -130,11 +130,24 @@ def call(self, files):
...
-        file_list = function_table[src_type](source, files['dir_op'])
+        if src_type == 'local':
+            direction_context = src_type + dest_type
+            # If the file generator has no associated operation, it
+            # is not the driving force behind an operation. That means
+            # it is being used as the other generator (the one being used
+            # for listing of files in the destination) in a sync command.
+            # More specifically this if statement will only be hit if it
+            # is the file generator listing files for a downloading sync.
+            if self.operation_name == '':
```
kyleknap: I feel that this was sort of hacky when I wrote it as I am relying on the fact in the code when I [instantiate](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/subcommands.py#L758) the FileGenerator for the reference files in the destination, the ``operation_name`` is set to ``''``. I want some feedback on this part.

The other option I was thinking of  was to introduce a new argument to the constructor that specifies the transfer direction. The reason I did not do it was because it would bloat the constructor and may require a bunch of updates to the tests. Note I am not too concerned about coupling to the command being run here because the FileGenerator already requires an operation name.

jamesls:
Doesn't it seem simpler to just, much higher up the stack before doing an upload, validate that the local source exists?  We should be able to do it before even getting into any of the command architecture code paths.

kyleknap:
I think it was because I was concerned with follow_symlinks check right [above it](https://github.com/aws/aws-cli/pull/1575/files#diff-2d6153b2f8310981a59f57a017f7a2a3R258). But now that I think about it, I do not think that should be a concern because the if statment is only hit if the ``--no-follow_symlinks`` flag is set. If this is set, it will still hit the previous warning about a file not existing because the file will not be a symlink as it does not exist.

The other issue that I was concerned is with regards to https://github.com/aws/aws-cli/issues/1082. If I add the logic higher up in the stack. The only way to fix this issue is to completely short circuit the reverse filegenerator, which I am not sure how easy it is for syncing. Maybe make a no op filegenerator but that seems messy to me.

I am fine with the first part, but with the second part, I do not have a great idea on how to get around it. Do you have any ideas on how to do that? I could hoist the logic to the ``call()`` method of the FileGenerator to prevent bloating the rest of the methods. Is that what you had in mind?

jamesls:
@kyleknap I'm not following how these is a problem.  Is your first concern we'd have to duplicate the symlink checking logic somewhere in `CommandParams`?

Also I'm not quite sure I understand the second point, maybe because I don't know the root cause.  Why do we need to possibly short circuit the reverse filegenerator?

kyleknap:
No the first paragraph is no longer an issue. I was just explaining (a little bit too my self as well) a possible issue, but it ended up not being an issue.

The second one is an issue because in order to avoid modifying the ``should_ignore`` method and not throw the warning, I need to avoid calling the ``list_files`` because currently it will always check the user provided path (which may not exist, but it does not matter because we will still download it and create the appropriate directories) for a downloading sync. So by short circuit, I mean I need to either provide a new reverse file generator that does a no op in the ``subcommands.CommandArchitecture`` class or immediately return before the ``list_files`` happens to avoid checking whether destination local directory exists. I think I am leaning toward the later option. Does that make sense?

jamesls:
I'm definitely missing something.  Can we not just create the directory before doing any transfers?  Let me give something more specific.  Why doesn't something like this work for your second case?

```
# in subcommands.py
if dir_op and path_type == 's3local':
  if not os.path.exists(params['dest']):
    os.makedirs(params['dest'])
```

kyleknap:
Yeah that would be simpler. The only thing is that it is kind of weird that we are making modifications to the file system there as opposed to in s3handler and fileinfo.

kyleknap:
Alright I updated the PR. Integration tests still pass.

jamesls:
[CommandParameters](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/subcommands.py#L852) already has has pre-existing validation of parameters.  Was there a reason not to add this logic to that class?

kyleknap:
No, I just had forgotten about that class. Had not looked at it for a long time, and it shown. There were methods in the class that we not even used and some tests were using internal methods...

I update the PR. Integration tests still pass.

jamesls:
:shipit:

PR #1579, opened 2015-10-21
Add pointer to help commands on error messages

### LINKED ISSUE #254: Make `aws -h` output suggest you also run `aws help`
Every time I use this tool I always run `aws -h` and get a not very helpful usage message. If it suggested I also try `aws help` it'd be much more useful.

jamesls:
These commands will now also include how you can look
at help text:

* aws
* aws <service>
* aws <service> <operation>
* aws <service> <operation> --unknown arg (including --help)

The plan is still to support --help eventually, but for now,
we can at least add pointers to how to get help with the
existing help commands.

```
$ aws
usage: aws [options] <command> <subcommand> [parameters]

  aws help
  aws <command> help
aws: error: too few arguments


$ aws ec2
usage: aws [options] <command> <subcommand> [parameters]

  aws help
  aws <command> help
aws: error: too few arguments

$ aws ec2 run-instances
usage: aws [options] <command> <subcommand> [parameters]

  aws help
  aws <command> help
aws: error: argument --image-id is required

$ aws ec2 describe-instances --help
usage: aws [options] <command> <subcommand> [parameters]

  aws help
  aws <command> help

Unknown options: --help
```

Closes #254.

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -24,7 +24,7 @@ HELP_BLURB = (
     "  aws <command> <subcommand> help\n"
 )
 USAGE = (
-    "aws [options] <command> <subcommand> [parameters]\n"
+    "aws [options] <command> <subcommand> [<subcommand> ...] [parameters]\n"
     "%s" % HELP_BLURB
 )
 
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -16,6 +16,19 @@ from awscli.compat import six
 from difflib import get_close_matches
 
 
+HELP_BLURB = (
+    "To see help text, you can run:\n"
+    "\n"
+    "  aws help\n"
+    "  aws <command> help\n"
+    "  aws <command> <subcommand> help\n"
+)
+USAGE = (
+    "aws [options] <command> <subcommand> [parameters]\n"
+    "%s" % HELP_BLURB
+)
+
+
 class CLIArgParser(argparse.ArgumentParser):
     Formatter = argparse.RawTextHelpFormatter
 
@@ -71,13 +84,13 @@ class MainArgParser(CLIArgParser):
     Formatter = argparse.RawTextHelpFormatter
 
     def __init__(self, command_table, version_string,
-                 description, usage, argument_table):
+                 description, argument_table):
         super(MainArgParser, self).__init__(
             formatter_class=self.Formatter,
             add_help=False,
             conflict_handler='resolve',
             description=description,
-            usage=usage)
+            usage=USAGE)
         self._build(command_table, version_string, argument_table)
 
     def _create_choice_help(self, choices):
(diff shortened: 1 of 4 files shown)


awscli/argparser.py
```diff
@@ -16,6 +16,19 @@
 from difflib import get_close_matches
 
 
+HELP_BLURB = (
+    "To see help text, you can run:\n"
+    "\n"
+    "  aws help\n"
+    "  aws <command> help\n"
+    "  aws <command> <subcommand> help\n"
```
kyleknap: You can also have more than one subcommand with the wait commands. Not sure if there is a better way to represent this. Maybe:
```py
"To see help text, you can run:\n"
"\n"
"  aws help\n"
"  aws <command> [<subcommand> ...] help\n"
```

Same comment applies about the usage

awscli/argparser.py
```diff
@@ -16,6 +16,19 @@
...
+)
+USAGE = (
+    "aws [options] <command> <subcommand> [parameters]\n"
```
kyleknap: Maybe:
```py
"aws [options] <command> <subcommand> [<subcommand> ...] [parameters]\n"
```

kyleknap:
I am not sure if my suggestions make the help text more confusing, but it is more correct than before. What do you think?

awscli/argparser.py
mtdowling: I'm not sure that's necessary -- couldn't subsequent subcommands be considered parameters of a parent command?

awscli/argparser.py
mtdowling: Can you show and example of what that looks like?

mtdowling:
LGTM

awscli/argparser.py
kyleknap: Yeah:
```
$ aws ec2 wait help
```
and
```
$ aws ec2 wait instance-running help
```

Have two completely different outputs.

awscli/argparser.py
kyleknap: Not really. It is actually already an incorrect assumption/help message with the current implementation. Here is how it is currently for wait commands:
```
$ aws ec2 wait

usage: aws [options] <command> <subcommand> [parameters]
aws: error: too few arguments

```
If it was a parameter, the error message would say it is missing a required argument such as:
```
$ aws s3api wait bucket-exists
usage: aws [options] <command> <subcommand> [parameters]
aws: error: argument --bucket is required
```

jamesls:
I don't have a strong preference.  On the one hand, I've simply copied what the exact usage messages are in the CLI today:

```
$ aws
usage: aws [options] <command> <subcommand> [parameters]
aws: error: too few arguments
```

On the other hand, what you've proposed with `aws <command> [<subcommand> ...] help` is technically more correct but maybe more confusing, especially because the amount of commands that are `aws command subcommand` vastly outnumber the number of waiter commands we have.

jamesls:
Ok ready for another look.

kyleknap:
:ship: I think it is better to keep the ``HELP_BLURB`` as is because the people who will really need it are those who are brand new to the CLI. Let's keep it simple even though it does not cover all possible scenarios.

rayluo:
:ship:

Issue #748, opened 2014-04-08
aws s3 sync does not ignore storage-class of glacier by default


e0d:
With:

awscli==1.3.6
botocore==0.40.0

While the default param for aws sync for storage-class is documented as STANDARD, files with storage-class of GLACIER are matched when the file list is generated, but throw errors when download is attempted.

```
A client error (InvalidObjectState) occurred when calling the GetObject operation: The operation is not valid for the object's storage class
```

gideononline:
This (InvalidObjectState) error also causes the sync process to be extremely slow.  We have millions of files in an S3 bucket and millions more in glacier.   So although the sync process successfully completes in the end, it takes many hours more than it should because of millions of exceptions.   Any ETA on the fix?

joehoyle:
+1 with this issue, I can't perform a `sync` while using Glacier

caedmonjudd:
+1 on this issue as well.

viyh:
+1, also being able to exclude or include Glacier object types for "ls" would be useful.

thomascate:
+1, has anyone figured out a work around for this?

cake-icing:
+1

andacata:
+1

yurasuka:
+1

Argoday:
+1

robbintt:
it would be so nice if the API provided an MD5 or any equivalent for files in the glacier. Is there any sort of unique ID like an object ID or date + object ID that could be used as a unique ID in order to allow s3 sync to work?

PR #1581, opened 2015-10-21
Warn and skip on glacier objects that may fail for s3 commands

kyleknap:
This pull request does the following:
* Skips operations that involve transferring a glacier object. This includes downloading, copying, and moving an object when the source is a glacier object. If the glacier object is skipped, a warning is displayed causing the return code to be ``2``. The most important part of this functionality is that we will not even try performing an operation, which could slow down a command especially if you have a lot of or large glacier objects.
* Add ``--ignore-glacier-warnings`` argument that allows you to hide the glacier warnings and not cause the return code to be 2 if a glacier object is encountered. Note that glacier objects are still always skipped if the command involves downloading, copying, or moving an object when the source is a glacier object.

All of the integration tests pass.

Fixes https://github.com/aws/aws-cli/issues/748

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -13,7 +13,6 @@
 import os
 import sys
 import stat
-from collections import namedtuple
 
 from dateutil.parser import parse
 from dateutil.tz import tzlocal
@@ -93,10 +92,6 @@ class FileDecodingError(Exception):
         super(FileDecodingError, self).__init__(self.error_message)
 
 
-LocalFileData = namedtuple(
-    'LocalFileData', ['size', 'last_update'])
-
-
 class FileStat(object):
     def __init__(self, src, dest=None, compare_key=None, size=None,
                  last_update=None, src_type=None, dest_type=None,
@@ -153,18 +148,12 @@ class FileGenerator(object):
 
     def _inject_extra_information(self, file_stat_kwargs, extra_information):
         src_type = file_stat_kwargs['src_type']
-        # For local sources, ``extra_information`` will be a ``LocalFileData``
-        # object.
-        if src_type == 'local':
-            file_stat_kwargs['size'] = extra_information.size
-            file_stat_kwargs['last_update'] = extra_information.last_update
+        file_stat_kwargs['size'] = extra_information['Size']
+        file_stat_kwargs['last_update'] = extra_information['LastModified']
 
-        # For s3 sources, ``extra_information`` will be a dictionary
-        # representing the response element the object represents from
-        # a ListObjects or HeadObject.
-        else:
-            file_stat_kwargs['size'] = extra_information['Size']
-            file_stat_kwargs['last_update'] = extra_information['LastModified']
+        # S3 objects require the response data retrieved from HeadObject
+        # and ListObject
+        if src_type == 's3':
             file_stat_kwargs['response_data'] = extra_information
 
     def list_files(self, path, dir_op):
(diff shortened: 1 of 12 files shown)


awscli/customizations/s3/filegenerator.py
```diff
@@ -324,4 +326,4 @@ def _list_single_object(self, s3_path):
         file_size = int(response['ContentLength'])
         last_update = parse(response['LastModified'])
         last_update = last_update.astimezone(tzlocal())
-        return s3_path, file_size, last_update
```
jamesls: This seems like a weird API.  We're returning parts of the response (such as LastModified) as well as just the entire response whole sale.  This can probably be cleaned up.

awscli/customizations/s3/utils.py
```diff
@@ -353,6 +353,44 @@ def relative_path(filename, start=os.path.curdir):
         return os.path.abspath(filename)
 
 
+def is_glacier_object(response_data):
```
jamesls: Is there a reason this needs to be extracted out?  Wouldn't it make more sense in fileinfo?

awscli/customizations/s3/filegenerator.py
kyleknap: I kept it like this to keep parity with the output for local file listings. There is no response associated with a local listing. I will probably break up the logic such that the listing of files and listing of s3 objects do not have the same interface.

awscli/customizations/s3/utils.py
kyleknap: I extracted it because I saw it as being potential useful if you were to write a glacier-specific sync strategy. I am fine with moving it to FileInfo though since we are not using it anywhere else.

mtdowling: Yeah, +1 on moving these to FileInfo.

awscli/customizations/s3/filegenerator.py
```diff
@@ -157,7 +159,7 @@ def list_files(self, path, dir_op):
         if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
-                yield path, size, last_update
+                yield path, size, last_update, None
```
mtdowling: Returning tuples with four elements, some of which are `None` seems like a code smell. Perhaps this method (and other similar methods) should return a dict or a specific object instead of a long tuple?

kyleknap: Yeah I was trying to touch the least amount of code paths as possible, but I agree it is getting too messy. I am going to just refactor the ``FileGenerator`` class to simplify some of the interfaces. Should not be hard.

kyleknap:
Alright, I think I incorporated the feedback. Refactored the ``FileGenerator`` a bit. Mainly it allowed the ``list_files`` and ``list_objects`` to yield data that may differ as opposed to forcing the two methods to return the same size tuples that may or may not be completely filled. Ready to get looked at again.

kyleknap:
Alright I cleaned up the code a bit more. Should be good to look at.

jamesls:
:shipit:

awscli/customizations/s3/subcommands.py
```diff
@@ -290,12 +290,23 @@
                  'Using a lower value may help if an operation times out.')}
 
 
+IGNORE_GLACIER_WARNINGS = {
+    'name': 'ignore-glacier-warnings', 'action': 'store_true',
+    'help_text': (
+        'Turns off glacier warnings. Warnings about operations that cannot '
+        'be performed because it involves copying, downloading, or moving '
+        'a glacier object will no longer be printed to standard error and '
```
mtdowling: Changing this to "*and will no longer* cause the return code of the command to be ``2``" might make it a bit more clear that using this option will prevent an error exit rather than cause one.

mtdowling:
:ship:

awscli/customizations/s3/subcommands.py
kyleknap: Updated

PR #1584, opened 2015-10-22
Adds more detail in sync debug log

rayluo:
This is an attempt to provide more information when debugging some cases.
Also refactor a subclass to not repeat a same function from its sibling.


rayluo:
The log will change from something like:

    ... awscli.customizations.s3.syncstrategy.base - DEBUG - syncing: mybucket/daily/foo.txt -> e:\daily\foo.txt, size_changed: False, last_modified_time_changed: True

to something like:

    ... awscli.customizations.s3.syncstrategy.base - DEBUG - syncing: mybucket/daily/foo.txt -> e:\daily\foo.txt, size_changed: 123 -> 123, modified time: 2015-10-21 14:48:28-07:00 -> 2015-10-21 05:16:32-07:00

kyleknap:
Looks good. Looks like the change to simplify the logic is safe. :ship:

PR #1592, opened 2015-10-26
Add example for kms create-alias

kyleknap:
There was some confusion about the value for ``--alias-name``.


jamesls:
:shipit:

PR #1593, opened 2015-10-27
Fix regression from client switch over

kyleknap:
The ``configservice subscribe`` command was using an attribute that no longer existed and was relying
on an internal attribute...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ Next Release (TBD)
   respectively.  The old arguments are still supported for backwards
   compatibility, but are no longer documented.
   (`issue 1599 <https://github.com/aws/aws-cli/pull/1599>`__)
+* bugfix:``aws configservice subscribe``: Fix an issue when creating a
+  new S3 bucket
+  (`issue 1593 <https://github.com/aws/aws-cli/pull/1593>`__)
 
 
 1.9.1
--- a/awscli/customizations/configservice/subscribe.py
+++ b/awscli/customizations/configservice/subscribe.py
@@ -151,7 +151,7 @@ class S3BucketHelper(object):
         return s3_bucket_exists(self._s3_client, bucket)
 
     def _create_bucket(self, bucket):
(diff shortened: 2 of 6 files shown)


jamesls:
:shipit:

Do we not have any functional tests for this?  It would be nice to have a way to catch if this breaks again, even though `.meta` is pretty much a support API.

kyleknap:
We do not. I can see if we can add functional tests since we do not have any integration tests and since integration tests would be tricky to add.

kyleknap:
@jamesls 
I added some functional tests. For the functional test that checks for an existing bucket and creates it when it does not exist, there is a limitation in ``BaseAWSCommandParamsTest`` where you cannot specify more than one status code (sort of similar to being able to provide more than one response via ``parsed_responses``, but there is nothing similiar to that for ``http_response``) which causes the entire subscribe command to fail when you set the status code to 404. However, I am able to capture the desired functionality for checking the existence of a bucket and then attempting to create it.

What do you think? I think this would be nice to have but out of the scope of this PR because this is fixing a regression.

jamesls:
@kyleknap am I correct in understanding that the functional test you've added would not have caught the issue fixed in this PR because we're not creating a bucket?

kyleknap:
It would have in the third test. The error happens before ``CreateBucket``: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/configservice/subscribe.py#L154 and I assert ``CreateBucket`` is called.

jamesls:
Oh I see.  I didn't notice the third test.  In that case, I'm not a huge fan of this because it doesn't really reflect reality.  It may also cause additional maintenance on us if we ever support multiple status codes in the base test class.

I'm really just after a test that uses a real client to ensure that we catch any interfaces breaking.  What if we just instantiate that S3 helper class with a real client, but mock out the responses?

kyleknap:
Finally all of the leaky patches are fixed. That was rough. Good to look at it one more time.

jamesls:
:shipit:

PR #1594, opened 2015-10-27
Adding better utime error messages.

### LINKED ISSUE #1322: aws s3 sync fails if user does not own all files
I am finding that `aws s3 sync` will fail if the active user does not have ownership of files in the sync target, regardless of what permissions are set for group and other.

**Normal result**

    $ aws s3 sync s3://example.com/ example/
    download: s3://example.com/file.txt to example/file.txt

**Problematic result**

To reproduce the problem, create an empty file that is world-writable but owned by root. aws s3 sync will fail to sync this file, even though the effective permissions allow the user to write to the file.

    $ mkdir example && touch example/file.txt
    $ sudo chown root example/file.txt && sudo chmod 777 example/file.txt

    $ aws s3 sync s3://example.com/ example/
    download failed: s3://example.com/file.txt to example/file.txt
    [Errno 1] Operation not permitted: '/example/file.txt'

    $ ls -l example/file.txt
    -rwxrwxrwx  1 root  wheel  54  7 May 17:55 example/file.txt

However, when running as the root user (`sudo aws s3 sync`) I cannot reproduce the problem. The file downloads as expected.

**Environment**

I am able to reproduce this issue on both machines I've tried it on:

* aws-cli/1.7.25 Python/2.7.6 Darwin/14.3.0 (Mac OS X 10.10)
* aws-cli/1.7.10 Python/2.7.6 Linux/3.13.0-40-generic (Ubuntu 14.04)

mtdowling:
This commit updates the s3 commands to give a more descriptive error message when attempting to set the utime of a file fails due to a "1" OS erro number (operation not permitted).

Note that if this error is encountered, the CLI will print the descriptive error message and continue downloading other files until finally exiting with an error.

Closes #1322

@jamesls @kyleknap @rayluo @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -353,6 +353,30 @@ def relative_path(filename, start=os.path.curdir):
         return os.path.abspath(filename)
 
 
+def set_file_utime(filename, desired_time):
+    """
+    Set the utime of a file, and if it fails, raise a more explicit error.
+
+    :param filename: the file to modify
+    :param desired_time: the epoch timestamp to set for atime and mtime.
+    :raises: SetFileUtimeError: if you do not have permission (errno 1)
+    :raises: OSError: for all errors other than errno 1
+    """
+    try:
+        os.utime(filename, (desired_time, desired_time))
+    except OSError as e:
+        # Only raise a more explicit exception when errno is 1
+        if e.errno is not 1:
+            raise e
+        raise SetFileUtimeError(
+            ("The file was downloaded, but attempting to modify the "
+             "utime of the file failed. Is the file owned by another user?"))
+
+
+class SetFileUtimeError(Exception):
+    pass
+
+
 class ReadFileChunk(object):
     def __init__(self, filename, start_byte, size):
         self._filename = filename
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -15,6 +15,7 @@ from datetime import datetime
 import mimetypes
 import hashlib
 import math
+import errno
 import os
 import sys
 from collections import namedtuple, deque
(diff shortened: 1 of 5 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -353,6 +353,30 @@ def relative_path(filename, start=os.path.curdir):
...
+    :param filename: the file to modify
+    :param desired_time: the epoch timestamp to set for atime and mtime.
+    :raises: SetFileUtimeError: if you do not have permission (errno 1)
+    :raises: OSError: for all errors other than errno 1
+    """
+    try:
+        os.utime(filename, (desired_time, desired_time))
+    except OSError as e:
+        # Only raise a more explicit exception when errno is 1
+        if e.errno is not 1:
```
kyleknap: Very small thing (it does not need to change). Just want to give a heads up. I usually avoid using ``is`` unless I am actually checking that the identity of an object is the identity of another object. It is fine for small integers like these, but for larger ones there may be edge cases:
```py
In [1]: 1000 is 10**3
Out[1]: False
```
Equality checks are usually safer if you are just checking values.

kyleknap:
Looks good had a small python comment. :ship:

awscli/customizations/s3/utils.py
mtdowling: Ah yeah... I really should have used equality here anyway because I *think* Python would automatically turn this into an identity check based on https://www.codementor.io/python/tutorial/stack-overflow-martijn-pieters-python-optimization. I'll change this and merge.

jamesls: Also probably should use a constant from `errno` instead of a hardcoded `2` for readability.

jamesls:
Have you confirmed the same errno on windows?  Is this problem possible on windows?

awscli/customizations/s3/utils.py
mtdowling: I've updated the PR to use `errno.EPERM` instead of `1`.

mtdowling:
os.utime is [supported on Windows](https://docs.python.org/2/library/os.html#os.utime), so I assume it could also fail under similar conditions where a user does not have permission to modify file attributes. `errno` is cross platform, so if the error occurs on Windows, it will still be `errno.EPERM `. Windows uses a subclass of `OSError` called `WindowsError` that introduces a `winerror` attribute that is Windows specific, but they map the `winerror` to the appropriate and normalized `errno` attribute. See https://docs.python.org/2/library/exceptions.html#exceptions.WindowsError.

jamesls:
Sounds good.  :shipit:

PR #1599, opened 2015-10-27
Add backcompat args for argument renames

jamesls:
In boto/botocore#703, there were some correction made
to the CamelCase->snake_case conversions.

However, in the CLI, we still need to support these arguments.

This fix for this is to:

* Add a copy of the correct arg into the arg table with the
  "wrong" name.
* Mark the "wrong" name entry as undocumented.

There were no unit tests for this module, so I added a few.
I also added some functional tests for some (but not all)
the rename cases, just to ensure the end to end functionality
is working as we expect.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,8 +4,15 @@ CHANGELOG
 
 Next Release (TBD)
 ==================
+
 * bugfix:``aws s3``: Fix some local path validation issues
   (`issue 1575 <https://github.com/aws/aws-cli/pull/1575>`__)
+* bugfix:``aws storagegateway``:  Fix ``--tape-ar-ns``,
+  ``--volume-ar-ns``, and ``--vtl-device-ar-ns`` to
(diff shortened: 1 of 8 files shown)


kyleknap:
The change looks good. The only thing I am curious about is how did you audit the arguments that needed an alias? We do not want to miss any to avoid breaking changes.

kyleknap:
:ship: otherwise

jamesls:
I linked to the gist in the original [pull request](https://github.com/boto/botocore/pull/703):  https://gist.github.com/jamesls/0d1005415d30d2f9f44a

Ran it against develop, ran it against the PR branch.  Looked at the diff.

I suppose if we wanted to be really sure I'd write the equivalent code that uses the arg tables to verify that we didn't miss anything because of customizations renaming args.

kyleknap:
Ok I think that is fine then. I do not think customization renamings is an issue because we chose what to rename it to as ``xform_name`` is really only used for model driven commands, which we do not control directly.

PR #1606, opened 2015-10-29
Apigateway

kyleknap:
Fix an issue with ``--uri`` not being black-listed. Noticed it when writing integration tests in this [file](https://github.com/kyleknap/botocore/blob/apigateway-fix/tests/integration/test_apigateway.py)


JordonPhillips:
:ship:

PR #1608, opened 2015-10-29
Survive a BucketAlreadyOwnedByYou error

rayluo:
This happens occasionally in our test cases, when the first create_bucket()
attempt succeeded but somehow the response never came back.
It is safe to ignore this error in our test case, because we would still have
a new bucket to complete other test cases.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -57,6 +57,7 @@ import awscli.clidriver
 from awscli.plugin import load_plugins
 from awscli.clidriver import CLIDriver
 from awscli import EnvironmentVariables
+import awscli.errorhandler
 
 
 # The unittest module got a significant overhaul
@@ -163,7 +164,16 @@ def create_bucket(session, name=None, region=None):
     params = {'Bucket': bucket_name}
     if region != 'us-east-1':
         params['CreateBucketConfiguration'] = {'LocationConstraint': region}
-    client.create_bucket(**params)
+    try:
+        client.create_bucket(**params)
+    except awscli.errorhandler.ClientError as e:
+        if e.error_code == 'BucketAlreadyOwnedByYou':
+            # This can happen in the retried request, when the first one
+            # succeeded on S3 but somehow the response never comes back.
+            # We still got a bucket ready for test anyway.
+            pass
+        else:
+            raise
     return bucket_name
 
 
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -57,7 +57,6 @@ import awscli.clidriver
 from awscli.plugin import load_plugins
 from awscli.clidriver import CLIDriver
 from awscli import EnvironmentVariables
-import awscli.errorhandler
 
 
 # The unittest module got a significant overhaul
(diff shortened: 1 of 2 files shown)


tests/unit/test_testutils.py
```diff
@@ -0,0 +1,34 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import mock
+
+from awscli.testutils import create_bucket
+from awscli.testutils import BaseCLIDriverTest
+
+ERROR_409 = b'''<?xml version="1.0" encoding="UTF-8"?>
```
JordonPhillips: Does this need to be taken outside the scope of the function? Also, you should probably put junk values for RequestID, HostID, and BucketName. I don't think anything can be done with them, but better safe than sorry (You don't use them anyway, so it's not a big deal).

JordonPhillips:
:+1: :ship:

awscli/testutils.py
```diff
@@ -57,6 +57,7 @@
 from awscli.plugin import load_plugins
 from awscli.clidriver import CLIDriver
 from awscli import EnvironmentVariables
+import awscli.errorhandler
```
jamesls: I would prefer we just disable the cli specific errorhandler for this client.  We eventually want to remove the `errorhandler` module entirely so new code shouldn't really be using it.

tests/unit/test_testutils.py
```diff
@@ -0,0 +1,36 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import mock
+
+from awscli.testutils import create_bucket
+from awscli.testutils import BaseCLIDriverTest
+
+
+class TestCreateBucket(BaseCLIDriverTest):
```
jamesls: We typically don't "test the tests".  I'm ok with not having this.  What do the rest of you think?

cc @kyleknap @mtdowling @JordonPhillips

awscli/testutils.py
rayluo: Did not know that plan. OK then. The next PR will unregister the errorhandler like this.

        client.meta.events.unregister(
            'after-call', unique_id='awscli-error-handler')

This line will not be as obvious as the `import awscli.errorhandler`, though. Which means we could probably forget to also remove this line, when we eventually remove the `errorhandler` module entirely. That being said, an extra `unregister('event', unique_id='an_nonexist_id')` will not bring substantial harm anyway.

tests/unit/test_testutils.py
rayluo: Personally I don't do it in a "black or white" way. I sometimes wrote tests for test utility when I feel a need, and sometimes - oh God of Test pls forgive me - skip test cases for simple function.

The funny story in this case is that, this test actually helped me, twice. At first it helped me to realize that, by default the exception is `awscli.errorhandler.ClientError` rather than `botocore.exceptions.ClientError`. And then it helped me just now to use `client.meta.events.unregister(...)` rather than `session.unregister(...)`. Those two errors won't be discovered by syntax check. Ouch. Ideally it shouldn't need a test case to get this seemingly-simple task right, but that is a different story anyway.

jamesls: I think if you can get an ACK from at least one other person let's keep the test.

tests/unit/test_testutils.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2013 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesls: 2015

jamesls:
:shipit:.  Would be nice to get someone else to +1 testutils.  I don't have a strong preference, though I slightly prefer it not being there.  Not a big deal either way.

tests/unit/test_testutils.py
kyleknap: Seems unnecessary to test the testutils as users of the SDK should never be using these unless submitting a PR in which tests are updated/added.

kyleknap: I'm +0 in removing it for the record.

PR #1619, opened 2015-11-04
Don't print tracebacks on Ctrl-C

jamesls:
This adds two new changes:

* Ctrl-C by the user does not print a traceback.  We just exit the process with a 130 RC
* Ctrl-C is ignored when piping help output to a pager.  This is needed so we avoid putting the terminal in a bad state if we immediately exit on a ctrl-c.  It's also consistent with the behavior of `man somecommand`, `git commit --help`, etc.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import sys
+import signal
 import logging
 
 import botocore.session
@@ -197,6 +198,12 @@ class CLIDriver(object):
                    '"aws configure".' % e)
             self._show_error(msg)
             return 255
+        except KeyboardInterrupt:
+            # Shell standard for signals that terminate
+            # the process is to return 128 + signum, in this case
+            # SIGINT=2, so we'll have an RC of 130.
+            sys.stdout.write("\n")
+            return 128 + signal.SIGINT
         except Exception as e:
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
--- a/tests/unit/test_clidriver.py
+++ b/tests/unit/test_clidriver.py
@@ -261,7 +261,7 @@ class TestCliDriver(unittest.TestCase):
         expected = {'log_level': logging.ERROR, 'logger_name': 'awscli'}
         self.assertEqual(driver.session.stream_logger_args[1], expected)
 
-    def test_ctrl_s_is_handled(self):
+    def test_ctrl_c_is_handled(self):
         driver = CLIDriver(session=self.session)
         fake_client = mock.Mock()
         fake_client.list_objects.side_effect = KeyboardInterrupt
--- a/tests/unit/test_clidriver.py
+++ b/tests/unit/test_clidriver.py
@@ -261,6 +261,15 @@ class TestCliDriver(unittest.TestCase):
         expected = {'log_level': logging.ERROR, 'logger_name': 'awscli'}
         self.assertEqual(driver.session.stream_logger_args[1], expected)
 
+    def test_ctrl_s_is_handled(self):
+        driver = CLIDriver(session=self.session)
+        fake_client = mock.Mock()
+        fake_client.list_objects.side_effect = KeyboardInterrupt
+        fake_client.can_paginate.return_value = False
+        driver.session.create_client = mock.Mock(return_value=fake_client)
+        rc = driver.main('s3 list-objects --bucket foo'.split())
+        self.assertEqual(rc, 130)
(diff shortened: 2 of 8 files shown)


tests/unit/test_clidriver.py
```diff
@@ -261,6 +261,15 @@ def test_error_logger(self):
         expected = {'log_level': logging.ERROR, 'logger_name': 'awscli'}
         self.assertEqual(driver.session.stream_logger_args[1], expected)
 
+    def test_ctrl_s_is_handled(self):
```
JordonPhillips: *ctrl_c

awscli/clidriver.py
```diff
@@ -197,6 +198,12 @@ def main(self, args=None):
                    '"aws configure".' % e)
             self._show_error(msg)
             return 255
+        except KeyboardInterrupt:
+            # Shell standard for signals that terminate
+            # the process is to return 128 + signum, in this case
+            # SIGINT=2, so we'll have an RC of 130.
+            sys.stdout.write("\n")
+            return 128 + signal.SIGINT
```
kyleknap: Out of curiosity what used to be the return value?

tests/unit/test_clidriver.py
```diff
@@ -261,6 +261,15 @@ def test_error_logger(self):
...
 
+    def test_ctrl_c_is_handled(self):
+        driver = CLIDriver(session=self.session)
+        fake_client = mock.Mock()
+        fake_client.list_objects.side_effect = KeyboardInterrupt
+        fake_client.can_paginate.return_value = False
+        driver.session.create_client = mock.Mock(return_value=fake_client)
+        rc = driver.main('s3 list-objects --bucket foo'.split())
+        self.assertEqual(rc, 130)
+
```
kyleknap: Can we also assert that we do not see any traceback in standard error?

tests/unit/test_help.py
```diff
@@ -117,6 +119,23 @@ def test_can_page_output_on_windows(self):
...
+                return self.mock_popen
+
+            def _is_pager_call(self, args):
+                return 'less' in args[0]
+
+        renderer = CtrlCRenderer()
+        renderer.mock_popen.communicate.return_value = ('rendered', '')
+        renderer.exists_on_path['groff'] = True
+        renderer.render('foo')
+
```
kyleknap: What does this test assert? It is not obvious to me.

kyleknap:
Implementation looks fine. Had a question and a couple of comments about the tests.

awscli/clidriver.py
jamesls: Because it was an uncaught exception that propagated all the way up the stack, the default python behavior is an RC of 1.

kyleknap: Ok that's fine. Do not really think users would be catching interrupts

mtdowling:
:ship:

tests/unit/test_clidriver.py
jamesls: Because this is a unittest, we'd never actually print the traceback to stderr.  The test runner would just catch the exception and log it in the error associated with the test result.

In order to test what you'd want I'd need to create an integration test, run an aws command, and send it a ctrl-c.

The only tricky part is that this is highly dependent on getting the timing right, we need to ctrl the process before it exists and after it's started and is making an API call.

tests/unit/test_help.py
jamesls: Good catch, it's missing an assertion.  I'll assert that the rendered input is sent to the pager despite receiving a ctrl-c.

tests/unit/test_clidriver.py
kyleknap: Could we could possibly use BaseAWSCommandParamsTest with a mocked session and the ``run_cmd()`` method to capture the stderr?

jamesls: No, there is no stderr to capture.

When a traceback gets all the way up the stack *python itself* will print the traceback for you before exiting.  It is impossible to test for stderr in this scenario in a unit test because we're not the one printing the traceback in the first place.

Take a look at my [latest commit](https://github.com/jamesls/aws-cli/commit/c09e9e7cd6f09c7bf5f993f9fcf76d13390f1f1f), where I add an integ test.  This is the only way to test this functionality if you want to verify that there's no Traceback in stdout/stderr.

jamesls:
@kyleknap Feedback incorporated.

tests/unit/test_clidriver.py
kyleknap: Ah makes sense forgot that that test does not use a process

kyleknap:
Looks good. :ship:

JordonPhillips:
:shipit:

PR #1622, opened 2015-11-04
Switch test code over to skip_if_windows

jamesls:
This fixes an issue with test_cli, where a test needed
to be annotated to be skipped on windows but was missed.

I've also added a test utility function for this and I've
updated all the existing skippable tests to use this decorator.
This also makes is easier in the future if we need to update
what qualifies as "not windows."

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -81,6 +81,22 @@ INTEG_LOG = logging.getLogger('awscli.tests.integration')
 AWS_CMD = None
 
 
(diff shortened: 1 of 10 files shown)


jamesls:
I should mention that [this](https://github.com/aws/aws-cli/pull/1622/files#diff-00cd37d3fc324762195eb40f56bb1102R358) is the specific test that needed to be tagged with skip if windows.

kyleknap:
:ship:

Issue #1466, opened 2015-08-25
using sse-kms when uploading objects


kbiton:
I use the aws s3 cli command to upload multiple object to s3 , I am utilising the parallel multipart build-in features. I noticed thar sse-kms is not a supported option when using aws s3 it looks like only sse-s3 is supported is there any particular reason for that ? I can use the high level API , aws s3api that being said this API (IMHO) does not support parallel & multipart uploads.
can you recommend on a workaround will support to SSE-KMS added to the cp options?

kyleknap:
It just has not been added it. Mainly because there may be some edge cases that we will have to take into account. For example, in order to use kms with s3, the requests must be signed using signature version 4.

I do not think that there is a good workaround, other than using ``s3api``, but like you said that does upload in parallel or automatically handle multipart. Marking this as feature request.

kbiton:
thank you

Issue #1377, opened 2015-06-09
Add SSE with Customer Key Support to S3 Sync


poldon:
Please add the ability to use SSE with a customer key to the aws s3 command, especially sync.  I see that s3api has support for customer keys, just not the high-level s3 commands.  Thanks!

kyleknap:
Marking as feature request.

PR #1487, opened 2015-09-02
Add SSE-C and SSE-KMS support to s3 subcommands

0xkag:
This pull request adds SSE-C and SSE-KMS support into awscli s3 subcommands like "aws s3 cp" and "aws s3 sync".

From my commit message:

> -- add SSE-C and SSE-KMS support to the aws s3 subcommands (SSE-{C,KMS} are two newer forms of S3 Server Side Encryption)
>
> -- the added SSE-C and SSE-KMS support coexists with existing SSE-S3 support without breaking the current meaning of the --sse command line option
>
> -- tested with (src, dst) of locals3, s3local, and s3s3, including copy from objects written as {S3,SSE-C,SSE-KMS,SSE-S3} to {S3,SSE-C,SSE-KMS,SSE-S3} both with the same and with different keys (for SSE-C and SSE-KMS)
>
> -- add a shell script to test different permutations of SSE (integration into the Python functional tests is a TODO) and s3local, locals3, s3s3 src and dst with aws s3 cp and aws s3 sync

There are a few open TODO items worth discussing, mostly around documentation and integrating a test script into the python test suite.  I may or may not have time to do these myself.  Regardless of the need for these, it is fully functional for all of the above, and I want to get this on your radar since I see open issues requesting additional SSE support (#1377, #1466).  (I have a need for SSE-KMS so I added it myself rather than wait.  I figured I might as well add SSE-C too while I was at it.)

> --
> Kyle George

PR #1623, opened 2015-11-05
Add support for --sse aws:kms and --sse-c

kyleknap:
So this builds on top of: https://github.com/aws/aws-cli/pull/1487.

The pull request adds support for SSE using KMS and SSE-C. To support this functionality, here are the changes required:
1) ``--sse`` can take now both ``AES256`` and ``aws:kms`` as values. If kms is specified, the clients default to sigv4. Note that if ``--sse`` is specified it still results in ``AES256``.
2) ``--sse-kms-key-id`` was added to allow users to specify a key id.
3) ``--sse-c`` and ``--sse-c-copy-source`` (for copies) was added to support SSE-C. This required the sibling parameters ``--sse-c-key`` and ``--sse-c-copy-source``. If you are missing one part of either pair, an error gets thrown. Also ``--sse-c-copy-source`` can only be used for copies.

As to tests, the integration are pretty exhaustive. I was trying to make sure I caught all of the different edge cases in terms of interacting with S3. The unit and functional tests are focused on more on covering some of the areas that I could not or was not efficient to cover with integration tests. Let me know if there are some tests that you would like to see added.

Fixes https://github.com/aws/aws-cli/issues/1377
Fixes https://github.com/aws/aws-cli/issues/1466


awscli/customizations/s3/filegenerator.py
```diff
@@ -320,7 +323,11 @@ def _list_single_object(self, s3_path):
         # instead use a HeadObject request.
         bucket, key = find_bucket_key(s3_path)
         try:
-            response = self._client.head_object(Bucket=bucket, Key=key)
+            params = {'Bucket': bucket, 'Key': key}
+            if self.sse_c:
```
jamesls: Well that's unfortunate.  Seems like we're missing a level of abstraction here.  Maybe it's worth abstracting these away as a parameter object so we don't have to change this class in the future if new params are required for head_object with new features.

Either that or I think it might be worth pulling this whole "metadata for a single object" concept out into its own object.  This would isolate future changes even more.

awscli/customizations/s3/fileinfo.py
```diff
@@ -59,7 +60,8 @@ def save_file(filename, response_data, last_update, is_stream=False):
         with open(filename, 'wb') as out_file:
             write_to_file(out_file, etag, md5, file_chunks)
 
-    if not _is_multipart_etag(etag) and sse != 'aws:kms':
+    if not _is_multipart_etag(etag) and sse != 'aws:kms' and \
```
jamesls: Can we just pull this out into a utils function that takes the entire response data?  Something like can_verify_checksum(response_data).  This is the third time we've needed to update this.  I'm looking to try to localize these changes going forward.  Seems like updating a utils function would be easier.

awscli/customizations/s3/fileinfo.py
```diff
@@ -253,6 +252,60 @@ def _handle_object_params(self, params):
         if self.parameters['expires']:
             params['Expires'] = self.parameters['expires'][0]
 
+    def set_sse_request_params(self, request_params, cli_params):
```
jamesls: Why do these need to be public?  Seems like an implementation detail we don't need to expose.

jamesls:
Overall seems ok.  I would prefer that we take the time to split out all the param handling logic in FileInfo into it's own class so we can test it more easily.  Seems like this is only going to get worse as we need to support more parameters.

awscli/customizations/s3/filegenerator.py
kyleknap: Yeah its really unfortunate. I am not sure about the second option. These reason we do ``HeadObject`` for a single object is that it is much cheaper than ``ListObjects``. We are not necessarily doing it to get metadata. I think I am going to do something along the lines of the first option.

awscli/customizations/s3/fileinfo.py
kyleknap: That is fine. Probably keep it in the ``FileInfo`` class.

kyleknap:
Alright, I think I addressed your feedback @jamesls. I split up the parameter mapping into its own utility class where the public methods map cli parameters to request parameters depending on what operation is being ran.

0xkag:
Kyle,

Just took a brief look at your changes.

In #1487 I required that someone globally set signature s3v4 for KMS to work.  In this request I see where you set signature s3v4 if someone has explicitly specified SSE w/ KMS.  That's certainly better, but it won't handle the case where someone is copying an object where the source is encrypted with SSE-KMS and the destination is not SSE-KMS.  As I said on #1487:

> I went through the code and looked at the feasibility of this. It looks non-trivial because of PUT Copy, where if you don't use SigV4 and the source object is SSE-KMS then it will fail. The only way to check is to HEAD first, but that's a lot of overhead. I think requiring SigV4 across the board should be considered.

I'm not saying that HEAD'ing every object before copy should be done, but I think that it at least needs to be documented or the failure mode handled with an appropriate error message.  I only saw one test that explicitly references "s3v4", and I don't think it covers this case.

kyleknap:
That is true. I do not think that we will be able to HEAD the object to determine if it is kms encrypted before doing a copy because that would adding an extra call that costs users time and money. I think the best option is to improve the error message on how to set sigv4 when you get kms/sigv4 errors from s3. This option will be nice as well because we would be able to apply it to both ``s3`` and ``s3api`` commands.

awscli/customizations/s3/filegenerator.py
```diff
@@ -116,7 +116,7 @@ class FileGenerator(object):
     ``FileInfo`` objects to send to a ``Comparator`` or ``S3Handler``.
     """
     def __init__(self, client, operation_name, follow_symlinks=True,
-                 page_size=None, result_queue=None):
+                 page_size=None, result_queue=None, request_parameters=None):
```
jamesls: Does it make sense to move page_size in the request_parameters dict with an entry for `ListObjects`?

awscli/customizations/s3/utils.py
```diff
@@ -517,3 +517,134 @@ def __new__(cls, message, error=False, total_parts=None, warning=None):
...
+
+class RequestParamsMapper(object):
+    """A utility class that maps CLI params to request params
+
+    Each method in the class maps to a particular operation and will set
+    the request parameters depending on the operation and CLI parameters
+    provided.
+    """
+    @classmethod
+    def map_put_object_params(cls, request_params, cli_params):
```
jamesls: A docstring (or a class docstring because it looks like they all have similar signatures) that shows a basic example as well as documents the params would be helpful.

jamesls:
:shipit:  Had some small feedback, otherwise looks good.  The RequestMapper stuff makes the code much easier to follow.

awscli/customizations/s3/filegenerator.py
kyleknap: Maybe? If I add it, I will probably will want to break the ``BucketLister`` interface to use ``request_parameters`` in lieu of the current arguments it takes. How about I add ``page_size`` to ``request_parameters`` whenever we decide that we need to use ``request_parameters`` because of a needed extra request parameter for ``ListObjects``?

jamesls: Ok sounds good.

kyleknap:
Update PR with docstring. Merging.

PR #1624, opened 2015-11-05
Adding IoT command args to save keys and certs

mtdowling:
This customization makes it easier to save various pieces of data
returned from iot commands that would typically need to be saved to a
file. This customization adds the following options:

- aws iot create-certificate-from-csr
  - --certificate-pem-outfile: certificatePem
- aws iot create-keys-and-certificate
  - --certificate-pem-outfile: certificatePem
  - --public-key-outfile: keyPair.PublicKey
  - --private-key-outfile: keyPair.PrivateKey

@jamesls @kyleknap @JordonPhillips @rayluo

### DIFF
--- a/awscli/customizations/arguments.py
+++ b/awscli/customizations/arguments.py
@@ -10,7 +10,25 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import os
+
 from awscli.arguments import CustomArgument
+import jmespath
+
+
+def resolve_given_outfile_path(path):
+    """Asserts that a path is writable and returns the expanded path"""
+    if path is None:
+        return
+    outfile = os.path.expanduser(os.path.expandvars(path))
+    if not os.access(os.path.dirname(os.path.abspath(outfile)), os.W_OK):
+        raise ValueError('Unable to write to file: %s' % outfile)
+    return outfile
+
+
+def is_parsed_result_successful(parsed_result):
+    """Returns True if a parsed result is successful"""
+    return parsed_result['ResponseMetadata']['HTTPStatusCode'] < 300
 
 
 class OverrideRequiredArgsArgument(CustomArgument):
@@ -51,3 +69,60 @@ class OverrideRequiredArgsArgument(CustomArgument):
         if name_in_cmdline in args:
             for arg_name in argument_table.keys():
                 argument_table[arg_name].required = False
+
+
+class StatefulArgument(CustomArgument):
+    """An argument that maintains a stateful value"""
+
+    def __init__(self, *args, **kwargs):
+        super(StatefulArgument, self).__init__(*args, **kwargs)
+        self._value = None
+
+    def add_to_params(self, parameters, value):
+        super(StatefulArgument, self).add_to_params(parameters, value)
+        self._value = value
+
+    @property
+    def value(self):
+        return self._value
+
+
+class QueryOutFileArgument(StatefulArgument):
+    """An argument that write a JMESPath query result to a file"""
+
+    def __init__(self, session, name, query, after_call_event,
+                 *args, **kwargs):
+        self._session = session
+        self._query = query
+        self._after_call_event = after_call_event
+        # Generate default help_text if text was not provided.
+        if 'help_text' not in kwargs:
+            kwargs['help_text'] = ('Saves the command output contents of %s '
+                                   'to the given filename' % self.query)
+        super(QueryOutFileArgument, self).__init__(name, *args, **kwargs)
+
+    @property
+    def query(self):
+        return self._query
+
+    def add_to_params(self, parameters, value):
+        value = resolve_given_outfile_path(value)
+        super(QueryOutFileArgument, self).add_to_params(parameters, value)
+        if self.value is not None:
+            # Only register the event to save the argument if it is set
+            self._session.register(self._after_call_event, self.save_query)
+
+    def save_query(self, parsed, **kwargs):
+        """Saves the result of a JMESPath expression to a file.
+
+        This method only saves the query data if the response code of
+        the parsed result is < 300.
+        """
+        if is_parsed_result_successful(parsed):
+            contents = jmespath.search(self.query, parsed)
+            with open(self.value, 'w') as fp:
+                # Don't write 'None' to a file -- write ''.
+                if contents is None:
+                    fp.write('')
+                else:
+                    fp.write(contents)
--- a/awscli/customizations/iot.py
+++ b/awscli/customizations/iot.py
@@ -1,0 +1,52 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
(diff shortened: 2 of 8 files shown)


mtdowling:
FYI: I fixed an issue with Python 3 and rebased.

awscli/customizations/arguments.py
```diff
@@ -51,3 +54,66 @@ def override_required_args(self, argument_table, args, **kwargs):
...
+
+    def save_query(self, parsed_result):
+        """Saves the result of a JMESPath expression to a file.
+
+        If the argument value was not set, then the file is not saved
+        or modified.
+        """
+        if self.value and self.query:
+            contents = jmespath.search(self.query, parsed_result)
+            with open(self.value, 'wb') as fp:
```
kyleknap: Should we create directories if needed?

awscli/customizations/arguments.py
```diff
@@ -51,3 +54,66 @@ def override_required_args(self, argument_table, args, **kwargs):
...
+                # Don't write 'None' to a file -- write ''.
+                if contents is None:
+                    fp.write(''.encode())
+                else:
+                    fp.write(contents.encode())
```
kyleknap: We should check the type here (i.e. if it is bytes or unicode). If it is unicode, let's encode it. If not, let it be. That will allow the class to be generalizable. Also, lets be explicit with what we are encoding with (`'utf-8'`)

awscli/customizations/iot.py
```diff
@@ -0,0 +1,64 @@
...
+  - ``--public-key-outfile``: keyPair.PublicKey
+  - ``--private-key-outfile``: keyPair.PrivateKey
+"""
+from awscli.customizations.arguments import QueryOutFileArgument
+
+
+class IoTWrapper(object):
+
+    def __init__(self, event_handler):
+        self._cert_arg = QueryOutFileArgument(
```
kyleknap: We are creating these arguments upon every command run, even if it is not iot related. Could we instantiate these in the arguments in the handler that gets registered to the events?

awscli/customizations/iot.py
```diff
@@ -0,0 +1,64 @@
...
+            'private-key-outfile', 'keyPair.PrivateKey')
+        event_handler.register(
+            'building-argument-table.iot.create-keys-and-certificate',
+            self._add_create_keys_and_cert_args)
+        event_handler.register(
+            'building-argument-table.iot.create-certificate-from-csr',
+            self._add_create_cert_from_csr_args)
+        save_events = ['after-call.iot.CreateCertificateFromCsr',
+                       'after-call.iot.CreateKeysAndCertificate']
+        for event in save_events:
```
kyleknap: It seems sort of weird that we are registering these events prior to we actually knowing that the arguments are present on the command line. I sort of like the pattern we do for ``streamingoutputargs.py`` where we register the event when we add it to the params.

awscli/customizations/iot.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _add_create_keys_and_cert_args(self, argument_table, **kwargs):
+        # This command uses every custom argument, so apply each.
+        argument_table['certificate-pem-outfile'] = self._cert_arg
+        argument_table['public-key-outfile'] = self._pub_key_arg
+        argument_table['private-key-outfile'] = self._private_key_arg
+
+    def _save_files(self, parsed, **kwargs):
+        # Do not save files when an error response is received.
+        if 'Error' not in parsed:
```
kyleknap: I think we should check the status code of the ``http_response`` so that it is less than 300 because it is more accurate of if a ClientError will be thrown or not.

tests/functional/iot/test_outfile.py
```diff
@@ -0,0 +1,98 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.testutils import BaseAWSCommandParamsTest
+import os
+
+
+class TestOutFileQueryArguments(BaseAWSCommandParamsTest):
+    def getpath(self, filename):
```
kyleknap: I would take a look at the ``FileCreator`` object from ``testutils``. It is pretty handy when it comes to creating and cleaning files up on the fly. I think the ``FileCreator.rootdir`` will be useful for you because it will get wiped out at the end of each test.

kyleknap:
It is a good start. I had some comments. I think the main comment I had was that pattern in which you added the arguments was a bit different than other arguments in the cli where most arguments are instantiated in a handler registered to 'building-argument-table' event and if that argument needs to register more handlers it usually does so in one of its methods like ``add_params``.

awscli/customizations/arguments.py
mtdowling: I modeled this customization after https://github.com/aws/aws-cli/blob/develop/awscli/customizations/iamvirtmfa.py, which didn't create directories, so I though to keep to that behavior.

awscli/customizations/arguments.py
mtdowling: Good idea. Will do.

awscli/customizations/arguments.py
kyleknap: Yeah just a thought I add. We can keep it as is, until that functionality is requested.

awscli/customizations/iot.py
mtdowling: I'm not sure I follow -- if the parsed response contains the `Error` key, then we should not be saving anything. That's the behavior I was going for here. Are you saying something else?

mtdowling:
I modeled this after https://github.com/aws/aws-cli/blob/develop/awscli/customizations/iamvirtmfa.py because I need similar semantics. I can look and see how this can be done differently. I think it will get more complicated, but I'll see what I can do.

awscli/customizations/iot.py
kyleknap: I was thinking something like:
```py
def _save_files(self, http_response, parsed, **kwargs):
    if http_response.status_code < 300:
        # continue saving files

```
This point is pretty nit picky but I liked it a tad more because it is explicitly checking the request was unsuccessful (I'm a +0 on this).

kyleknap:
Yeah that's a pretty old customization and has not seen many updates. I think for that customization it is alright given its a single argument, but when we are adding three new arguments, across a couple operations, and using the same style it gets a little hard to follow.

awscli/customizations/arguments.py
jamesls: Given that we're writing a file to their filesystem, seems like it's more reasonable to use whatever encoding the user may have configured rather than assume utf-8. We can just open the file via `r` and and let python handle the encoding.

awscli/customizations/iot.py
jamesls: I don't have a preference either way, but if you end up using the status code, I'd suggest pulling it from the `['ResponseMetadata']['HTTPStatusCode']` key rather than the `http_response`.

`http_response` was needed prior to adding this info in the `ResponseMetadata` key so I'd like to try to move event handlers off of this param.

mtdowling: Cool, I'll update and go with `['ResponseMetadata']['HTTPStatusCode'] `

mtdowling:
@kyleknap Feedback incorporated. I've updated the PR to only register arguments for the specific commands they target.

mtdowling:
@kyleknap Ok, more refactoring based on feedback. I'll rebase when this is good to go.

kyleknap:
Thanks! :ship:

PR #1625, opened 2015-11-05
Add read timeout argument

JordonPhillips:
This will set the socket read timeout on the default client, enabling calls to long Lambda functions. It is called with `--cli-read-timeout` which takes the timeout in seconds.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -21,6 +21,7 @@ from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
+from botocore.client import Config
 
 from awscli import EnvironmentVariables, __version__
 from awscli.formatter import get_formatter
@@ -141,7 +142,8 @@ class CLIDriver(object):
             default=option_params.get('default'),
             action=option_params.get('action'),
             required=option_params.get('required'),
-            choices=option_params.get('choices'))
+            choices=option_params.get('choices'),
+            cli_type_name=option_params.get('type'))
 
     def create_help_command(self):
         cli_data = self._get_cli_data()
@@ -656,10 +658,15 @@ class CLIOperationCaller(object):
             value is returned.
 
         """
+        if parsed_globals.read_timeout is not None:
+            config = Config(read_timeout=parsed_globals.read_timeout)
+        else:
+            config = Config()
         client = self._session.create_client(
             service_name, region_name=parsed_globals.region,
             endpoint_url=parsed_globals.endpoint_url,
-            verify=parsed_globals.verify_ssl)
+            verify=parsed_globals.verify_ssl,
+            config=config)
         py_operation_name = xform_name(operation_name)
         if client.can_paginate(py_operation_name) and parsed_globals.paginate:
             paginator = client.get_paginator(py_operation_name)
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -52,8 +52,13 @@
             "help": "<p>Do not sign requests.  Credentials will not be loaded if this argument is provided.</p>"
         },
         "ca-bundle": {
-                "dest": "ca_bundle",
-                "help": "<p>The CA certificate bundle to use when verifying SSL certificates. Overrides config/env settings.</p>"
+            "dest": "ca_bundle",
+            "help": "<p>The CA certificate bundle to use when verifying SSL certificates. Overrides config/env settings.</p>"
+        },
+        "cli-read-timeout": {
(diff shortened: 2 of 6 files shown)


mtdowling:
What's the motivation behind prefixing the argument with `cli`?

JordonPhillips:
To avoid polluting the namespace. If any service decides to include a variable called `read-timeout` the global argument will shadow that and we'll have to add painful customizations.

awscli/data/cli.json
```diff
@@ -52,8 +52,12 @@
             "help": "<p>Do not sign requests.  Credentials will not be loaded if this argument is provided.</p>"
         },
         "ca-bundle": {
-                "dest": "ca_bundle",
-                "help": "<p>The CA certificate bundle to use when verifying SSL certificates. Overrides config/env settings.</p>"
+            "dest": "ca_bundle",
+            "help": "<p>The CA certificate bundle to use when verifying SSL certificates. Overrides config/env settings.</p>"
+        },
+        "cli-read-timeout": {
```
jamesls: Oh.  I'm guessing that because the args are defined in json, you can't set `type: int` like you normally could with `add_argument`?

jamesls:
Can we move the positive integer validation down into botocore?  Seems like this is something worth having in botocore as well, so moving it into botocore would avoid having to have this logic exist twice.

Other suggestion would be to just add support for "type" in the cli.json.  Otherwise we'll need customizations for every int argument going forward.  I think the only reason we didn't add this at first was because there were no global integer arguments.

jamesls:
@JordonPhillips is this ready to be looked at again?  Were you able to make any progress on the `type: int` approach?

awscli/clidriver.py
```diff
@@ -656,10 +657,15 @@ def invoke(self, service_name, operation_name, parameters, parsed_globals):
             value is returned.
 
         """
+        if hasattr(parsed_globals, 'read_timeout'):
```
jamesls: `if parsed_globals.read_timeout is None`.  `read_timeout` should always be in parsed globals since it's defined in `cli.json`.

JordonPhillips:
I think I was misreading your comments, I had thought that you suggested I *either* add the int or let botocore handle the validation rather than both. It's going in.

awscli/clidriver.py
JordonPhillips: If I don't do it this way, tests fail:

`AttributeError: 'Namespace' object has no attribute 'read_timeout'`

kyleknap: Which test fails? Maybe the test needs to get updated. The global parsing should add ``None`` for it when missing.

awscli/data/cli.json
kyleknap: I thought add_argument handles int's, but I could be wrong.

@JordonPhillips 
I am fine with this approach as is.

kyleknap:
Looks fine with me. I agree with James though on that ``if`` check. We should not be compromise how we are writing the code based on how the tests are already written when we can just update the tests.

awscli/clidriver.py
JordonPhillips: `test_paginate_rc` and `test_ctrl_c_is_handled` in `test_clidriver`. It doesn't look like the namespace is being patched, so I don't know why it isn't showing up.

kyleknap: Interesting. Let's figure this out. If the if statement is not needed (only needed to make the test pass) let's remove it, keep it otherwise.

mtdowling:
LGTM. I think this PR could be squashed into a single commit before merging.

JordonPhillips:
Absolutely, that's the plan.

jamesls:
:shipit:

kyleknap:
LGTM :ship: after squashing commits.

tests/unit/test_clidriver.py
```diff
@@ -679,6 +683,13 @@ def test_aws_with_cacert_env_var(self):
         self.assertEqual(call_args[0], (mock.ANY, 'us-east-1'))
         self.assertEqual(call_args[1]['verify'], '/path/cacert.pem')
 
+    def test_aws_with_read_timeout(self):
+        self.assert_params_for_cmd(
+            'lambda invoke --function-name foo out.log --cli-read-timeout 90',
+            expected_rc=0)
+        call_args = self.create_endpoint.call_args
+        self.assertEqual(call_args[1]['timeout'], (60, 90))
```
rayluo: One minor thing. Shall we just check `self.assertEqual(call_args[1]['timeout'][1], 90)`? Just in case we would change the default connect_timeout into something other than 60 in future.

rayluo:
:ship:

Issue #1557, opened 2015-10-13
zsh autocomplete script removes already defined completions


jwhitley:
**Repro steps**

1.  Install aws-cli, e.g. via Homebrew
2.  Load it via `.zshrc`, e.g. using:

     ```
aws_cli_completion='/usr/local/share/zsh/site-functions/_aws'
[[ -r $aws_cli_completion ]] && source $aws_cli_completion
    ```
3.  Attempt to use a completion defined in `.zshrc` before the aws-cli completions are loaded.

**Actual results**

No completions loaded prior to the aws-cli completions work.  They were removed when the `_aws` script ran `compinit`.  Scripts that are intended to just establish a completion should not *ever* run `compinit`, for this very reason.  This is unexpected and rather bad behavior for a completion loader script.

**Expected results**

The zsh completions for `aws` are loaded, other completions remain as previously defined.

jamesls:
Thanks for reporting, looks like our zsh completer needs some updates.

mtdowling:
Fixed in https://github.com/aws/aws-cli/issues/1626

PR #1626, opened 2015-11-05
aws_zsh_completer.sh: do not run `compinit`

blueyed:
This will reset any manual `compdef`s from the user, and should only get
done once.


jamesls:
Fixes https://github.com/aws/aws-cli/issues/1557

jamesls:
Thanks!

Issue #1615, opened 2015-11-03
Elasticsearch Service `--advanced-options` Parsing Broken


bsandrow:
Per the documentation, I should be able to run this:

```console
$ aws es update-elasticsearch-domain-config --domain-name mylogs --region us-east-1 --advanced-options indices.fielddata.cache.size=35
```
What I get is this:

```console
$ aws es update-elasticsearch-domain-config --domain-name mylogs --region us-east-1 --advanced-options indices.fielddata.cache.size=35
Error parsing parameter '--advanced-options': Expected: '=', received: '.' for input:
indices.fielddata.cache.size=35
```

Right now, setting Advanced Options via the CLI only works if you use the `--cli-input-json` option (e.g.):

```console
$ aws es update-elasticsearch-domain-config --domain-name mylogs --region us-east-1 --cli-input-json '{"AdvancedOptions": {"indices.fielddata.cache.size": "35"}}'
```

Also, the documentation says that I should be pass `--advanced-options` like so:

```
--advanced-options indices.fielddata.cache.size=35 rest.action.multi.allow_explicit=false
```

But adding the second option (separated by a space) results in this error:

```
Unknown options: rest.action.multi.allow_explicit=false
```

(Running Version: `aws-cli/1.8.12 Python/2.7.10 Darwin/15.0.0`)

kyleknap:
Yeah looks like a bug in our shorthand syntax parsing. We recently redid it and at one point we considered making ``.`` meaningful way for expressing nested structures and members, but in the end, we decided against it.

tc-turner:
I was having this issue and was able to resolve it by removing leading spaces from `network.host: ` and `cluster.name:`

PR #1628, opened 2015-11-05
Make . a valid character key for shorthand

kyleknap:
I know we had discussions about making the ``.`` part of the shorthand syntax. This changes will prevent us from using it in a backward compatible way. But is needed for the shorthand syntax to be supported for this elasticsearch operation:
http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-configuration-api.html#es-configuration-api-actions-updateelasticsearchdomainconfig

However, I think we decided we are not going to include ``.`` in the syntax because it does not really simplify/condense the syntax. I just wanted to make sure everyone was aware of this.

Fixes https://github.com/aws/aws-cli/issues/1615

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+* bugfix:shorthand: Allow ``.`` as a valid key character.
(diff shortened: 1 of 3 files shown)


jamesls:
:shipit:

PR #1631, opened 2015-11-07
Copy over tox.ini from botocore and boto3

kyleknap:


jamesls:
:shipit: My same comments from boto3 apply, but if this is the best option then shipit.

kyleknap:
Yeah I think it is. It will at least ensure that people using ``tox`` are testing the same way that we test.

PR #1634, opened 2015-11-10
Making preview text more clear.

mtdowling:
The purpose of this change is to communicate more clearly that preview
services are in a preview state in the CLI and not necessarily a service
that is marked as preview (or pre-GA).


JordonPhillips:
:+1: :ship:

kyleknap:
:ship:

PR #1642, opened 2015-11-15
Add .spec file for building RPM from tarball

nkadel:
This is for the RPM building issue described at https://github.com/aws/aws-cli/issues/1643.

The  "setup.py bdist --formats=rpm" doesn't provide enough detail in the SRPM to handle dependencies. The attachedk spec file is based on it, and includes critical python-2.6.x dependencies to prevent confusion building and installing an RPM on older supported operating systems, such as RHEL 5 or its free rebuilds, CentOS 5 or Scientific Linux 5. I've also activated a BuildRequires for the "python-setuptools" package, whose absence also creates an RPM building problem if not manually installed.

I've published a .spec file at https://github.com/nkadel/nkadel-aws-cli/blob/develop/python-awscli.spec, and would be happy to toss in a merge requiest so that "rpmbuild -ta" could be used to build RPM's from the published tarbals. It takes a bit more work for RHEL 5, which requires the optional "python26" and a locally built "pythons25-setuptools" to compile the RPM. but I'd be happy to work with anyone with a good, secuire, public yum repository to help make those available..

### DIFF
--- a/python-awscli.spec
+++ b/python-awscli.spec
@@ -1,0 +1,418 @@
+%define name python-awscli
+%define realname awscli
+%define version 1.9.5
+%define unmangled_version 1.9.5
+%define release 0.1%{?dist}
+
+Summary: Universal Command Line Environment for AWS.
+Name: %{name}
+Version: %{version}
+Release: %{release}
+# Actual download URL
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the pull request.  Following up from the comment from https://github.com/aws/aws-cli/pull/1642, we would like to have these packaging spec files pushed upstream to the various linux  distros.

Issue #1442, opened 2015-08-10
s3 sync, s3 cp no space left on device errors handling


vinogradovs:
aws s3 sync and aws s3 cp to local disk don't handle free space left on device  errors

aws s3 sync puts IOError: [Errno 28] No space left on device errors but still tries to continue sync
aws s3 cp doesn't give any error, tries to continue coping and returns 0 exit code at the end

jamesls:
Can you confirm what version of the CLI you're using (`aws --version`)?

vinogradovs:
1.7.44,
Hm, it seems strange, but now exit code is 1 if the whole bucket is synced. I will check this on other machine later.
The other  problem is that if there are  a lot of files to sync, it is needed to wait until utility finishes to try syncing of all of them, and syncing can be going for hours or in some cases even days.

jamesls:
Not sure I understand.  Could you provide a script that demonstrates the issue.  Here's what I tried.  I'm seeing 0 rc's if the bucket is synced:

```
/tmp/foo $ aws s3 sync . s3://bucket/
upload: ./foo.txt to s3://bucket/foo.txt
/tmp/foo $ echo $?
0
/tmp/foo $ aws s3 sync . s3://bucket/
/tmp/foo $ echo $?
0
```

Note that if an error occurred during the syncing process a non zero will happen but we still keep trying to sync the remaining files for you.  Here's some additional docs on the return codes used by the CLI: http://docs.aws.amazon.com/cli/latest/topic/return-codes.html#cli-aws-help-return-codes  You can also see these docs by running `aws help return-codes`.

chumaitoe:
I can confirm that `aws s3 cp` onto a device with insufficient space results in a partial copy and return code 0.  aws-cli/1.7.44

vinogradovs:
Execuse me for long waiting,

simple script:
aws s3 sync   s3://mybucket/ /mnt/mount_point_with_no_space_left/test/
echo $#

The second one with cp
aws s3 cp --recursive  s3://mybucket/ /mnt/mount_point_with_no_space_left/test/
echo $#
In both cases I get 0 exit code, but also I get errors like this:

download failed: s3://mybucket//test/setup.log to ../mnt/mount_point_with_no_space_left/test/setup.log Could not create directory s3://mybucket/ /mnt/mount_point_with_no_space_left/test/test5: [Errno 28] No space left on device: 's3://mybucket/ /mnt/mount_point_with_no_space_left/test/test5'
[...]

PR #1646, opened 2015-11-16
Print failure when creating local file fails

### LINKED ISSUE #1069: "aws s3 cp" fails silently when user has no write permissions on local directory
I just spun up a new EC2 instance and formatted an ephemeral storage volume. When I attempted to copy a (very large) S3 object to the root directory of that volume, the command completed immediately, with no output at all. Eventually I determined that only root had write permissions on the root directory of the volume, and granted write permissions to ec2-user, after which the command completed successfully. It would have been very helpful to have been given some idea of the reason for the failure, or any output at all for that matter.

### LINKED ISSUE #1645: S3 command fails silently when copying large file to location without permission
This only happens when copying files larger than a certain size, but not sure what is the threshold. For smaller files, when copying to a directory I don't have permission, I correctly see the permission denied error and the command exits with status 1. But for large files the command just stops, without printing to the output and the exit status is 0.

Steps to reproduce:

    mkdir access-denied
    sudo chown root:wheel access-denied/
    aws s3 cp s3://some-bucket/large-file.gz access-denied/
    # Result: nothing in the output and exit code is 0
    # Expected: exit code of 1 and output showing premission denied

Looking at the code, it seems the S3 copy command behaves differently depending on the size of the file being copied. For smaller files it seem to just download the file in a single thread. But for larger files it starts multiple threads, each downloading parts of the file. The single threaded process correctly output the error, but the multi-threaded doesn't.

I suspect this might affect other S3 commands as well.

Pull request to follow.

phss:
Fixes #1645. May fix #1069 and #1442.

This only applies when downloading large files. See issue #1645 for more details.

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -260,8 +260,9 @@ class S3Handler(object):
         chunksize = find_chunksize(filename.size, self.chunksize)
         num_downloads = int(filename.size / chunksize)
         context = tasks.MultipartDownloadContext(num_downloads)
-        create_file_task = tasks.CreateLocalFileTask(context=context,
-                                                     filename=filename)
+        create_file_task = tasks.CreateLocalFileTask(
+            context=context, filename=filename,
+            result_queue=self.result_queue)
         self.executor.submit(create_file_task)
         self._do_enqueue_range_download_tasks(
             filename=filename, chunksize=chunksize,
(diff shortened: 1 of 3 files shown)


jamesls:
Looks good, I'm just going to add a quick integration test before merging.  Thanks for the PR.

jamesls:
Merged in your PR and added an additional integ test.  Tracking over in #1648 .

phss:
Thanks for merging @jamesls. And apologies as I completely forgot about permission issues on Windows.

PR #1649, opened 2015-11-17
Add more options for configuring timeouts

kyleknap:
Add global options for ``--cli-connect-timeout``, ``--no-cli-read-timeout``, ``--no-cli-connect-timeout``. Also needed to plumb these into the s3 commands.

### DIFF
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -59,6 +59,11 @@
             "dest": "read_timeout",
             "type": "int",
             "help": "<p>The maximum socket read time in seconds.</p>"
+        },
+        "cli-connect-timeout": {
+            "dest": "connect_timeout",
+            "type": "int",
+            "help": "<p>The maximum socket connect time in seconds.</p>"
         }
     }
 }
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -58,12 +58,12 @@
         "cli-read-timeout": {
             "dest": "read_timeout",
             "type": "int",
-            "help": "<p>The maximum socket read time in seconds. If the value is set to -1, the socket read will be blocking and not timeout.</p>"
+            "help": "<p>The maximum socket read time in seconds. If the value is set to 0, the socket read will be blocking and not timeout.</p>"
         },
         "cli-connect-timeout": {
             "dest": "connect_timeout",
             "type": "int",
-            "help": "<p>The maximum socket connect time in seconds. If the value is set to -1, the socket connect will be blocking and not timeout.</p>"
+            "help": "<p>The maximum socket connect time in seconds. If the value is set to 0, the socket connect will be blocking and not timeout.</p>"
         }
     }
 }
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -58,12 +58,12 @@
         "cli-read-timeout": {
             "dest": "read_timeout",
             "type": "int",
-            "help": "<p>The maximum socket read time in seconds.</p>"
+            "help": "<p>The maximum socket read time in seconds. If the value is set to -1, the socket read will be blocking and not timeout.</p>"
         },
         "cli-connect-timeout": {
             "dest": "connect_timeout",
             "type": "int",
-            "help": "<p>The maximum socket connect time in seconds.</p>"
+            "help": "<p>The maximum socket connect time in seconds. If the value is set to -1, the socket connect will be blocking and not timeout.</p>"
         }
     }
 }
--- a/tests/unit/customizations/test_globalargs.py
+++ b/tests/unit/customizations/test_globalargs.py
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from botocore.session import get_session
 from botocore.handlers import disable_signing
 import os
 
@@ -196,24 +197,32 @@ class TestGlobalArgsCustomization(unittest.TestCase):
 
     def test_cli_read_timeout(self):
         parsed_args = FakeParsedArgs(read_timeout='60')
-        session = mock.Mock()
+        session = get_session()
         globalargs.resolve_cli_read_timeout(parsed_args, session)
         self.assertEqual(parsed_args.read_timeout, 60)
+        self.assertEqual(
+            session.get_default_client_config().read_timeout, 60)
 
     def test_cli_connect_timeout(self):
         parsed_args = FakeParsedArgs(connect_timeout='60')
-        session = mock.Mock()
+        session = get_session()
         globalargs.resolve_cli_connect_timeout(parsed_args, session)
         self.assertEqual(parsed_args.connect_timeout, 60)
+        self.assertEqual(
+            session.get_default_client_config().connect_timeout, 60)
 
     def test_cli_read_timeout_for_blocking(self):
         parsed_args = FakeParsedArgs(read_timeout='-1')
-        session = mock.Mock()
+        session = get_session()
         globalargs.resolve_cli_read_timeout(parsed_args, session)
         self.assertEqual(parsed_args.read_timeout, None)
+        self.assertEqual(
+            session.get_default_client_config().read_timeout, None)
 
     def test_cli_connect_timeout_for_blocking(self):
         parsed_args = FakeParsedArgs(connect_timeout='-1')
-        session = mock.Mock()
+        session = get_session()
         globalargs.resolve_cli_connect_timeout(parsed_args, session)
         self.assertEqual(parsed_args.connect_timeout, None)
+        self.assertEqual(
+            session.get_default_client_config().connect_timeout, None)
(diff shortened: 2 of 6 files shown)


JordonPhillips:
Looks good to me :ship:

mtdowling:
What about just using 0 as the "no" option?

JordonPhillips:
@mtdowling I'm -0. I'd rather be more explicit

tests/unit/test_completer.py
```diff
@@ -73,7 +74,9 @@
           '--no-verify-ssl', '--no-paginate', '--no-sign-request', '--output',
           '--profile', '--starting-token', '--max-items', '--page-size',
           '--region', '--version', '--color', '--query', '--ca-bundle',
-          '--generate-cli-skeleton', '--cli-input-json', '--cli-read-timeout'])),
+          '--generate-cli-skeleton', '--cli-input-json', '--cli-read-timeout',
```
jamesls: We really need to fix this.  We shouldn't have to update the completer whenever we add new command line options.

tests/unit/customizations/test_globalargs.py
```diff
@@ -199,3 +199,34 @@ def test_cli_read_timeout(self):
...
+        globalargs.resolve_cli_read_timeout(parsed_args, session)
+        self.assertEqual(parsed_args.read_timeout, None)
+
+    def test_cli_connect_timeout_for_blocking(self):
+        parsed_args = FakeParsedArgs(no_connect_timeout=True)
+        session = mock.Mock()
+        globalargs.resolve_cli_connect_timeout(parsed_args, session)
+        self.assertEqual(parsed_args.connect_timeout, None)
+
+    def test_cli_read_timeout_for_blocking_overrides_read_timeout(self):
```
jamesls: It feels like to be consistent with other args in the CLI, we should have either of these two options:

a) The two args are mutex'd.
b) The last value (on the command line) wins.

Mutex'd seems more explicit to me so I have a slight preference.

jamesls:
Had a couple of questions.  Overall the change seems fine.

I think we're quickly getting to a point where we need to be able to set a session level config object used for all create_client calls rather than forcing customizations to pass these in.  I'd almost prefer that approach over supporting `config=` *just* for S3 customizations.

mtdowling:
I'm still strongly in favor of using `0` as the setting for "no timeout" instead of adding two new "--no" options that are mutually exclusive to their counterparts. Instead of four new options, we'd be introducing two: `--cli-connect-timeout`, `--cli-read-timeout`. Two options are easier to document and there's no need to worry about what happens or how to document if a user provides both, for example, a `--cli-read-timeout` and a `--no-cli-read-timeout`.

Using `0` as the "no timeout" setting is also very common in other extremely popular CLI tools...

1. wget has many different timeout options: 
   `--timeout=SECONDS         set all timeout values to SECONDS.`
   `--dns-timeout=SECS        set the DNS lookup timeout to SECS.`
   `--connect-timeout=SECS    set the connect timeout to SECS.`
   `--read-timeout=SECS       set the read timeout to SECS.`

   wget does not offer a `--no-timeout` or anything like that. They document their timeout options as accepting `0` to disable: `Setting a timeout to 0 disables it altogether.` (check out `man wget`).
2. cURL has a `--max-time` setting that can be set to `0` meaning there is no limit: http://curl.haxx.se/docs/manpage.html#-m (this translates to http://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html which uses 0 as "no limit").

kyleknap:
I originally added the ``--no`` options because I thought setting it to zero would be ambiguous. Does it turnoff timeout completely or does it make it such that a timeout error is thrown if the socket is not ready on the first select call? Given cURL and wget use ``0``, I feel that it would be better to stay with the norm of popular CLI tools. @jamesls what do you think? You actually originally suggested using it, but after talking we decided to change it to having ``--no`` arguments because it seemed more explicit.

rayluo:
I also like to comply with other popular CLI tools to have timeout=0 as "no timeout". Regarding to the ambiguity @kyleknap mentioned does conceptually exist, but in practice I guess people only care about either "no timeout at all" or "short-enough timeout as 1 second or even 0.1 second (by the way, do we support float?)", but no really need "immediately timeout". So it is not really a concern.

jamesls:
@kyleknap Works for me.

kyleknap:
So I updated the PR to do the following:
1) Set timeouts using the session's default client config
2) Use 0 as the value for timeout options when you want the socket to be set to blocking
3) Update ``test_completer.py`` so that we do not have to update it every time a new command and argument is added. I considered moving these to functional tests, but the implementation required too much knowledge of the internals. So I decided to just keep it in unit tests, as at least there is no longer a maintenance burden.

tests/unit/test_clidriver.py
```diff
@@ -688,7 +692,38 @@ def test_aws_with_read_timeout(self):
             'lambda invoke --function-name foo out.log --cli-read-timeout 90',
             expected_rc=0)
         call_args = self.create_endpoint.call_args
-        self.assertEqual(call_args[1]['timeout'][1], 90)
+        self.assertEqual(call_args[1]['timeout'], (60, 90))
```
rayluo: I incline to the original line, because it won't couple with a default value of another parameter (i.e. connect_timeout) which is not the test scope of current test case.

Or in case we really want to include the connect_timeout value in this test case, in order to prove this PR's implementation does not incorrectly change the other parameter's default value, I would suggest we refer to that value by importing it from botocore, rather than hardcoding it here.

Same comment applies to the following similar test cases.

rayluo:
Overall looks good. I just add minor comment for test cases.

By the way, I did not run tests locally for this branch, but I happen to notice that, the already merged PR #1625 does not apply read_timeout to `aws s3 ls --cli-read-timeout 123`. (We can add `--debug` and observe a line like this `Setting s3 timeout as (60, 60)`.) 

Is it because those S3 customized commands need special treatment? Perhaps same problem also occurs in this PR. Please confirm.

kyleknap:
I am not seeing the issue that you are seeing. This is the line in the debug log that I get from running:
```
$ aws s3 ls --cli-read-timeout 123 --cli-connect-timeout 40 --debug 
...
2015-11-30 10:21:54,707 - MainThread - botocore.endpoint - DEBUG - Setting s3 timeout as (40, 123)
```
The reason that I even did this was so that all of the custom commands would not have to special logic for checking timeout arguments as all commands share the same session

tests/unit/test_clidriver.py
kyleknap: I can update that for whenever a non-default value is relied on for timeout

awscli/data/cli.json
```diff
@@ -58,7 +58,12 @@
         "cli-read-timeout": {
             "dest": "read_timeout",
             "type": "int",
-            "help": "<p>The maximum socket read time in seconds.</p>"
+            "help": "<p>The maximum socket read time in seconds. If the value is set to 0, the socket read will be blocking and not timeout.</p>"
```
mtdowling: Can you update these two descriptions to be "and will not timeout"?

tests/unit/test_completer.py
```diff
@@ -13,115 +13,121 @@
...
-                          'describe-step', 'disable-hbase-backups', 'get',
-                          'install-applications', 'list-clusters',
-                          'list-instances', 'list-steps',
-                          'modify-cluster-attributes', 'modify-instance-groups',
-                          'put', 'remove-tags', 'restore-from-hbase-backup',
-                          'schedule-hbase-backup', 'socks', 'ssh',
-                          'terminate-clusters', 'wait']))
-]
+
+def get_completion_sets(driver):
```
mtdowling: This method is massive and hard to follow. It's better than the previous implementation in that it doesn't have to be updated as frequently, but it loses some of the declarative nature. If we are going to use this new approach, then can you document this method please? Maybe even breaking this method into smaller methods would help.

mtdowling:
I had a couple small comments (doc tweak and minor readability refactoring to a test). After those are addressed: :ship:

jamesls:
The completer tests are not what I had in mind.  I don't think these are as useful as the previous tests because we're actually writing code to automatically generate what the expected results should be.  This is effectively re-implementing chunks of the tab completer.

What I would propose is creating our own *fixed* set of fake command/arg tables and then, using the same declarative nature of the original tests, make assertions about how the tab completer should work.

So instead of approaching the problem as "how can we adapt our assertions based on the fact that the commands/args are constantly changing" we approach it as "how can we stop the command/args from constantly changing."  We're interested in testing the tab completion logic, so using the latest set of commands/args is unnecessary for these tests.

I don't think this should block the original feature request, so I'd be in favor of addressing the tab completer as a separate PR.

kyleknap:
Alright I will just blow that commit away and we can address it in another PR. There is a significant amount of more work in doing it from scratch.

awscli/data/cli.json
kyleknap: Where do you want me to add that? I have "...the socket read will be blocking and not timeout." at the very end for each

kyleknap:
Updated PR. Good to look at again.

jamesls:
Thanks.  :shipit:

PR #1650, opened 2015-11-18
Take into account if a glacier object is restored for s3 downloads

jamesls:
If a glacier object has been restored, we should be able
to download the file.

Fixes a regression introduced when we started ignoring the
GLACIER storage class.

This doesn't fix the issue for a recursive downloads/sync,
but that isn't possible because the Restore data isn't
in a ListObjects request.

### DIFF
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -1,4 +1,5 @@
 import os
+import re
 import logging
 import sys
 import time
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/fileinfo.py
```diff
@@ -279,10 +280,24 @@ def is_glacier_compatible(self):
...
-            if response_data.get('StorageClass') == 'GLACIER':
+            if response_data.get('StorageClass') == 'GLACIER' and \
+                    not self._is_restored(response_data):
                 return True
         return False
 
+    def _is_restored(self, response_data):
+        # Returns True is this is a glacier object that has been
+        # restored back to S3.
+        # 'Restore' looks like: 'ongoing-request="false", expiry-date="..."'
```
rayluo: So is the following effect equivalent to `return r'ongoing-request="false"' in response_data.get('Restore', '')`?

awscli/customizations/s3/fileinfo.py
```diff
@@ -279,10 +280,24 @@ def is_glacier_compatible(self):
...
+        result = re.search(r'ongoing-request="(.+?)"',
+                           response_data.get('Restore', ''))
+        if result is None:
+            return False
+        ongoing_request_value = result.group(1)
+        if ongoing_request_value == 'false':
+            return True
+        return False
```
rayluo: So is the above effect equivalent to this line?

    return r'ongoing-request="false"' in response_data.get('Restore', '')

jamesls: Nice.  Good catch, I'll update this.

jamesls:
@rayluo Feedback incorporated.

rayluo:
Thanks! :ship:

awscli/customizations/s3/fileinfo.py
```diff
@@ -1,4 +1,5 @@
 import os
+import re
```
kyleknap: Unused import

kyleknap:
Looks good. You now have an unused import ``re``. Otherwise :ship:

jamesls:
Rebased, squashed, merged via https://github.com/aws/aws-cli/commit/21b6bd228202343018fba2e9a554c917751895e1

PR #1652, opened 2015-11-20
Add support for arbitrary metadata with the cp command

JordonPhillips:
This will allow a `--metadata` flag for `aws s3 cp` with the same interface as `aws s3api put-object --metadata`.

Example Usage:
```sh
$ aws s3 cp ~/test.txt s3://test-bucket/ --metadata some=value
upload: test.txt to s3://test-bucket/test.txt
$ aws s3api head-object --bucket test-bucket --key test.txt
{
    "AcceptRanges": "bytes",
    "ContentType": "text/plain",
    "LastModified": "Fri, 20 Nov 2015 21:13:12 GMT",
    "ContentLength": 13,
    "ETag": "\"ETAG\"",
    "Metadata": {
        "some": "value"
    }
}
$ aws s3 cp ~/test2.txt s3://test-bucket/ --metadata '{"other":"value"}'
upload: test2.txt to s3://test-bucket/test2.txt
$ aws s3api head-object --bucket test-bucket --key test2.txt
{
    "AcceptRanges": "bytes",
    "ContentType": "text/plain",
    "LastModified": "Fri, 20 Nov 2015 21:15:04 GMT",
    "ContentLength": 13,
    "ETag": "\"ETAG\"",
    "Metadata": {
        "other": "value"
    }
}
```

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -307,9 +307,10 @@ METADATA = {
     },
     'help_text': (
         "A map of metadata to store with the objects in S3. This will be "
-        "applied to every object which is part of this request. When copying "
-        "between two s3 locations, the metadata-directive argument will default"
-        " to 'REPLACE' unless otherwise specified."
+        "applied to every object which is part of this request. In a sync, this"
+        "means that files which haven't changed won't receive the new metadata."
+        " When copying between two s3 locations, the metadata-directive "
+        "argument will default to 'REPLACE' unless otherwise specified."
     )
 }
 
@@ -401,6 +402,7 @@ def get_client(session, region, endpoint_url, verify, config=None):
     return session.create_client('s3', region_name=region,
                                  endpoint_url=endpoint_url, verify=verify,
                                  config=config)
+
 
 class S3Command(BasicCommand):
     def _run_main(self, parsed_args, parsed_globals):
@@ -675,8 +677,8 @@ class MvCommand(S3TransferCommand):
     USAGE = "<LocalPath> <S3Uri> or <S3Uri> <LocalPath> " \
             "or <S3Uri> <S3Uri>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 2, 'positional_arg': True,
-                  'synopsis': USAGE}] + TRANSFER_ARGS + [METADATA_DIRECTIVE,
-                                                         RECURSIVE]
+                  'synopsis': USAGE}] + TRANSFER_ARGS +\
+                [METADATA, METADATA_DIRECTIVE, RECURSIVE]
 
 class RmCommand(S3TransferCommand):
     NAME = 'rm'
@@ -696,7 +698,8 @@ class SyncCommand(S3TransferCommand):
     USAGE = "<LocalPath> <S3Uri> or <S3Uri> " \
             "<LocalPath> or <S3Uri> <S3Uri>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 2, 'positional_arg': True,
-                  'synopsis': USAGE}] + TRANSFER_ARGS
+                  'synopsis': USAGE}] + TRANSFER_ARGS + \
+                [METADATA, METADATA_DIRECTIVE]
 
 
 class MbCommand(S3TransferCommand):
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -298,6 +298,22 @@ EXPIRES = {
 }
 
 
+METADATA = {
+    'name': 'metadata', 'cli_type_name': 'map',
+    'schema': {
+        'type': 'map',
+        'key': {'type': 'string'},
+        'value': {'type': 'string'}
+    },
+    'help_text': (
+        "A map of metadata to store with the objects in S3. This will be "
+        "applied to every object which is part of this request. When copying "
+        "between two s3 locations, the metadata-directive argument will default"
+        " to 'REPLACE' unless otherwise specified."
+    )
+}
+
+
 METADATA_DIRECTIVE = {
     'name': 'metadata-directive', 'choices': ['COPY', 'REPLACE'],
     'help_text': (
@@ -649,7 +665,7 @@ class CpCommand(S3TransferCommand):
             "or <S3Uri> <S3Uri>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 2, 'positional_arg': True,
                   'synopsis': USAGE}] + TRANSFER_ARGS + \
-                [METADATA_DIRECTIVE, EXPECTED_SIZE, RECURSIVE]
+                [METADATA, METADATA_DIRECTIVE, EXPECTED_SIZE, RECURSIVE]
 
 
 class MvCommand(S3TransferCommand):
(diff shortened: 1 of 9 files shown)


jamesls:
What's the expected result for an s3->s3 copy?  Should we require that the user also has to provide the appropriate value for `--metadata-directive`?

JordonPhillips:
@jamesls Yes, it requires that `--metadata-directive` be set. I mention that in the doc string.

jamesls:
Yeah, that's more what I was getting at.  Would it ever be valid for an s3->s3 transfer where the user provides `--metadata` but not `--metadata-directive`?  If this is something that should always be required for s3->s3 sync, are there any downsides to setting metadata-directive=REPLACE for the user when we detect this case?

kyleknap:
I would be in favor of automatically setting it to ``REPLACE`` and making sure users do not have to worry about it, given there is no downsides.

JordonPhillips:
No, it would never be valid to use without `--metadata-directive`. I'll go ahead and default it.

awscli/customizations/s3/subcommands.py
```diff
@@ -649,7 +665,7 @@ class CpCommand(S3TransferCommand):
             "or <S3Uri> <S3Uri>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 2, 'positional_arg': True,
                   'synopsis': USAGE}] + TRANSFER_ARGS + \
-                [METADATA_DIRECTIVE, EXPECTED_SIZE, RECURSIVE]
+                [METADATA, METADATA_DIRECTIVE, EXPECTED_SIZE, RECURSIVE]
```
kyleknap: We should add ``METADATA`` to ``mv`` as well. @jamesls what is your thoughts on adding to ``sync`` as well? I think I would be in favor of adding it if someone had custom metadata that applied to all objects being synced.

awscli/customizations/s3/utils.py
```diff
@@ -564,6 +565,7 @@ def map_copy_object_params(cls, request_params, cli_params):
         """Map CLI params to CopyObject request params"""
         cls._set_general_object_params(request_params, cli_params)
         cls._set_metadata_directive_param(request_params, cli_params)
+        cls._set_metadata_params(request_params, cli_params)
```
kyleknap: We need to add it for multipart-uploads and copies as well. You can do this by setting it in ``map_create_multipart_upload``. It would be great to get an integration test as well to make sure multipart's are handled correctly.

awscli/customizations/s3/subcommands.py
```diff
@@ -298,6 +298,22 @@
...
+    'schema': {
+        'type': 'map',
+        'key': {'type': 'string'},
+        'value': {'type': 'string'}
+    },
+    'help_text': (
+        "A map of metadata to store with the objects in S3. This will be "
+        "applied to every object which is part of this request. When copying "
+        "between two s3 locations, the metadata-directive argument will need to"
+        "be set to REPLACE."
```
kyleknap: Update the docstring to remove the tidbit about needing to provide REPLACE for metadata-directive as we now autopopulate it.

kyleknap:
Looks good. Had a couple comments about the handling of multipart transfers and whether we should add it to the sync command.

JordonPhillips:
Adding it to sync would add another layer of complexity because we would have to check for metadata differences when deciding to sync. I'm not -1, but I think it would be enough work to need to be timeboxed.

awscli/customizations/s3/subcommands.py
jamesls: By supporting it in sync, we're simply saying that we'll set metadata to whatever you've specified if we need to transfer a file to S3 right?  Seems like it would make sense.

kyleknap: Yeah that was what I was thinking. We still do not support syncing metadata.

JordonPhillips: That's a much simpler goal, but we will need to explicitly say that in the documentation. An event handler would be the correct way to make this documentation modification, correct?

kyleknap: No you would just need to update the ``help_text`` for the ``MEATADATA`` argument.

JordonPhillips: Do you think what I have in now looks fine?

kyleknap: If we add it to sync, it would be good to add a note for the behavior with the sync command.

JordonPhillips:
Docs are all updated

awscli/customizations/s3/subcommands.py
```diff
@@ -680,7 +697,7 @@ class SyncCommand(S3TransferCommand):
     USAGE = "<LocalPath> <S3Uri> or <S3Uri> " \
             "<LocalPath> or <S3Uri> <S3Uri>"
     ARG_TABLE = [{'name': 'paths', 'nargs': 2, 'positional_arg': True,
-                  'synopsis': USAGE}] + TRANSFER_ARGS
+                  'synopsis': USAGE}] + TRANSFER_ARGS + [METADATA]
```
kyleknap: Probably would want to add ``METADATA_DIRECTIVE`` as well.

awscli/customizations/s3/utils.py
```diff
@@ -638,6 +641,13 @@ def _permission_to_param(cls, permission):
                          'read|readacl|writeacl|full')
 
     @classmethod
+    def _set_metadata_params(cls, request_params, cli_params):
+        if cli_params.get('metadata'):
+            request_params['Metadata'] = cli_params['metadata']
+            if not cli_params.get('metadata-directive'):
```
kyleknap: Will autofilling ``MetadataDirective`` work for ``CreateMultipartUpload`` and ``PutObject``? It does not look like it. May have to add edge cases for that.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -439,6 +439,20 @@ def test_copy_large_file_signature_v4(self):
...
+        bucket_name = self.create_bucket()
+        key = 'foo.txt'
+        filename = self.files.create_file(key, contents='')
+        p = aws('s3 cp %s s3://%s/%s --metadata keyname=value' %
+                (filename, bucket_name, key))
+        self.assert_no_errors(p)
+        response = self.head_object(bucket_name, key)
+        # These values should have the metadata of the source object
+        self.assertEqual(response['Metadata'].get('keyname'), 'value')
+
```
kyleknap: I would really like to see test cases (functional or integration-wise) that test supplying metadata for uploads and multipart transfers. It would answer my previous questions about metadata-directive.

kyleknap:
Almost there. There are still a couple edge cases that I do not think you handle. For these type of edgecases it never hurts to add an integration or functional test to make sure it works through and through.

awscli/customizations/s3/utils.py
```diff
@@ -638,6 +642,15 @@ def _permission_to_param(cls, permission):
                          'read|readacl|writeacl|full')
 
     @classmethod
+    def _set_metadata_params(cls, request_params, cli_params):
+        if cli_params.get('metadata'):
+            request_params['Metadata'] = cli_params['metadata']
+
+    @classmethod
+    def _auto_populate_metadata_directive(cls, request_params):
+        request_params['MetadataDirective'] = 'REPLACE'
```
kyleknap: We only want to replace`` MetadataDirective`` if it is not currently in ``request_params`` correct?

kyleknap: Also only autopopulate it if ``Metadata`` is present in ``request_params``. I think that is why the tests are failing.

kyleknap:
I think once you fix the comment I had. It should be good to ship.

JordonPhillips:
fixed

kyleknap:
Thanks! Looks good :ship:

PR #1654, opened 2015-11-21
Add unicode normalize flag (proof of concept).

### LINKED ISSUE #1639: Provide an option to perfom unicode normalization on local file names
## Summary

`aws s3 sync` doesn't play well with HFS+ unicode normalization on OS X. I suggest to add an option to normalize file names read locally in normal form C before doing anything with them.
## Reproduction steps
1. Create a file on S3 containing an accented character. For reasons that will become apparent later, do this on a Linux system.
   
   ```
   (linux) % echo test > test/café.txt
   (linux) % aws s3 sync test s3://<test-bucket>/test
   ```
2. Synchronize that file on a Mac.
   
   ```
   (OS X) % aws s3 sync s3://<test-bucket>/test test
   download: s3://<test-bucket>/test/café.txt to test/café.txt
   ```
3. Synchronize it back to S3.
   
   ```
   (OS X) % aws s3 sync s3://<test-bucket>/test test
   upload: test/café.txt to s3://<test-bucket>/test/café.txt
   ```
- Expected result: no upload because the file is identical locally and on S3: I was just sync'd!
- Actual result: the file is uploaded again.

At this point the file shows up twice in S3!

screen shot 2015-11-14 at 22 45 38
## Why this happens

Unicode defines two normal forms — NFC and NFD — for some characters, typically accented characters which are common in Western European languages and even occur in English.

The documentation of [unicodedata.normalize](https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize), the Python function that converts between the two forms, has a good explanation.

A quick illustration:

```
>>> "café".encode('utf-8')
>>> b'caf\xc3\xa9'
>>> unicodedata.normalize('NFC', "café").encode('utf-8')
>>> b'caf\xc3\xa9'
>>> unicodedata.normalize('NFD', "café").encode('utf-8')
>>> b'cafe\xcc\x81'
```

The default filesystem of OS X, [HFS+](https://en.wikipedia.org/wiki/HFS_Plus), enforces something that resembles NFD. (Let's say I haven't encountered the difference yet.)

Pretty much everything else, including typing on a keyboard on Linux or OS X, uses NFC. I'm not sure about Windows.

Of course this is entirely HFS+'s fault, but since OS X is a popular system among your target audience, I hope you may have some interest in providing a solution to this problem.

**What you can do about it**

I think a `--normalize-unicode` option (possibly with a better name) for `aws s3 sync` would be useful. It would normalize file names read from the local filesystem with `unicodedata.normalize('NFKC', filepath)`.

Its primary purpose would be to interact with S3 on OS X and have file names in NFC form on S3, which is what the rest of the world expects and will cause the least amount of problems.

I don't know `aws cli` well enough to tell which other parts could use this option. I just encountered the problem when trying to replace "`rsync` to file server" with "`aws s3 sync` to S3".

FWIW `rsync` provides a solution to this problem with the `--iconv` option. A common idiom is `--iconv=UTF8-MAC,UTF8` when rsync'ing from OS X to Linux and `--iconv=UTF8,UTF8-MAC` when rsync'ing from Linux to OS X. `UTF8-MAC` is how `rsync` calls the encoding of file names on HFS+.

However this isn't a good API to tackle the specific problem I'm raising here. This API is about the encoding of file names. The bug is related to Unicode normalization. These are different concepts. `UTF8-MAC` mixes them.

Thanks!

aaugustin:
Fix #1639.

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -13,6 +13,7 @@
 import os
 import sys
 import stat
+import unicodedata
 
 from dateutil.parser import parse
 from dateutil.tz import tzlocal
@@ -31,7 +32,7 @@ def is_special_file(path):
     """
     This function checks to see if a special file.  It checks if the
     file is a character special device, block special device, FIFO, or
-    socket. 
+    socket.
     """
     mode = os.stat(path).st_mode
     # Character special device.
@@ -116,11 +117,13 @@ class FileGenerator(object):
     ``FileInfo`` objects to send to a ``Comparator`` or ``S3Handler``.
     """
     def __init__(self, client, operation_name, follow_symlinks=True,
-                 page_size=None, result_queue=None, request_parameters=None):
+                 page_size=None, normalize_unicode=False,
+                 result_queue=None, request_parameters=None):
         self._client = client
         self.operation_name = operation_name
         self.follow_symlinks = follow_symlinks
         self.page_size = page_size
+        self.normalize_unicode = normalize_unicode
         self.result_queue = result_queue
         if not result_queue:
             self.result_queue = queue.Queue()
@@ -170,6 +173,8 @@ class FileGenerator(object):
         """
         join, isdir, isfile = os.path.join, os.path.isdir, os.path.isfile
         error, listdir = os.error, os.listdir
+        if self.normalize_unicode:
+            path = unicodedata.normalize('NFKC', path)
         if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
@@ -188,6 +193,8 @@ class FileGenerator(object):
                 listdir_names = listdir(path)
                 names = []
                 for name in listdir_names:
+                    if self.normalize_unicode:
+                        name = unicodedata.normalize('NFKC', name)
                     if not self.should_ignore_file_with_decoding_warnings(
                             path, name):
                         file_path = join(path, name)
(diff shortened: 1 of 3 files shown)


mtdowling:
I'm seeing a lot of broken tests for this PR. Can you look into those tests and rebase with develop before pushing? Thanks!

aaugustin:
Thanks for your feedback.

You might not realize that I am a paying customer of Amazon S3. As such, I have limited interest in contributing my work to Amazon for free.

I merely published a proof of concept (hence the title of the PR) that works for me, could be useful for others, and could serve as a basis for developing a more complete patch.

If you'd like to contract with my company for the development of such a patch, please get in touch — aymeric.augustin@fractalideas.com. Thanks!

hjacobs:
Actually I kind of agree with @aaugustin here, the only OSS AWS creates is client tools for paying AWS customers.

Amazon is a black hole for Free Software :black_circle: --- a lot goes in, nothing comes out :disappointed:

mtdowling:
Thanks for following up. I didn't realize that you were simply pushing this up as a suggestion for AWS to essentially take your suggestion and iterate on it. I was assuming you were submitting a pull request with the intent that we would perform a code review, iterate on the pull request, and ultimately merge the pull request into the repository.

I think it's great you pushed up a PR to give a rough idea on how this could address #1639. Given that you will not be iterating on this PR to get it in a state that we can merge it into the CLI, I'll go ahead and close it. We can still track the original issue in #1639, and we'll keep this PR in mind as we decide how to best address the issue.

aaugustin:
Yes, that's definitely the correct course of action at this point. This PR was just a slightly more convenient way to communicate the patch than copy-pasting it in the issue. I hope it helps anyway!

PR #1656, opened 2015-11-23
Fix write permissions tests

jamesls:
These tests currently fail if you run them as a superuser.  This is because the tests rely on being able to take away write permission for a particular directory and verify the expected behavior.

However if you're root, you'll always have access to this directory, regardless of the `chmod` call.

This fix skips the test if we detect you're running as a superuser.

The unit test has the same problem, but instead I patched out the open call to raise an exception, so we can verify the expected behavior instead of having to skip the test.  This also has the benefit that the unit test can now run on windows.


kyleknap:
:ship:

PR #1657, opened 2015-11-24
Handle out of range timestamps

JordonPhillips:


awscli/customizations/s3/utils.py
```diff
@@ -217,10 +218,23 @@ def get_file_stat(path):
...
+    except ValueError as e:
+        # Python's fromtimestamp raises value errors when the timestamp is out
+        # of range of the platform's C localtime() function. This can cause
+        # issues when syncing from systems with a wide range of valid timestamps
+        # to systems with a lower range. Some systems support 64-bit timestamps,
+        # for instance, while others only support 32-bit. We set the time to be
+        # UNIX epoch time so that file size will take precedence in sync.
+        LOG.warning("Attempting to access file with out of range timestamp. "
+                    "Passing epoch time as timestamp.")
+        update_time = datetime(1970, 1, 1)
```
mtdowling: Does setting this to be UNIX epoch time have some side effect when syncing files? Would setting this to None work too?

awscli/customizations/s3/utils.py
```diff
@@ -217,10 +218,23 @@ def get_file_stat(path):
...
+    try:
+        update_time = datetime.fromtimestamp(stats.st_mtime, tzlocal())
+    except ValueError as e:
+        # Python's fromtimestamp raises value errors when the timestamp is out
+        # of range of the platform's C localtime() function. This can cause
+        # issues when syncing from systems with a wide range of valid timestamps
+        # to systems with a lower range. Some systems support 64-bit timestamps,
+        # for instance, while others only support 32-bit. We set the time to be
+        # UNIX epoch time so that file size will take precedence in sync.
+        LOG.warning("Attempting to access file with out of range timestamp. "
```
jamesls: I think this should still trigger a non-zero RC.  This is still an error condition that we couldn't figure out the correct timestamp.  We just don't want to stop the entire sync process, which is what I believe happens now.

awscli/customizations/s3/utils.py
JordonPhillips: It will treat the file as being as old as possible, so it will never be synced unless it isn't already present or if the s3 lastmodified somehow gets set to lower. The warning and non-zero rc will alert the customer to the issue. Setting it to None would require a customization to sync strategies that rely on time, including any custom sync strategies that a customer might have.

mtdowling: Fair points. It feels hacky to set this to the epoch time, but it seems like the best we can do without a major rewrite of sync.

awscli/customizations/s3/filegenerator.py
```diff
@@ -173,6 +173,12 @@ def list_files(self, path, dir_op):
         if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
+                if last_update == EPOCH_TIME:
```
jamesls: Is there a cleaner way to do this?  It's valid to have a file with a timestamp of EPOCH_TIME, despite it being unlikely.

awscli/customizations/s3/utils.py
jamesls: Do we need the LOG warnings anymore?  The result_queue will handle printing out error messages.

awscli/customizations/s3/filegenerator.py
JordonPhillips: Changed from `==` to `is`, which will check to see if it refers to the same object.

awscli/customizations/s3/filegenerator.py
```diff
@@ -208,6 +214,12 @@ def list_files(self, path, dir_op):
                             yield x
                     else:
                         size, last_update = get_file_stat(file_path)
+                        if last_update is EPOCH_TIME:
```
kyleknap: I am fine with this approach but can we abstract this logic to a new method so we do not have to copy the logic twice?

awscli/customizations/s3/filegenerator.py
```diff
@@ -173,6 +173,12 @@ def list_files(self, path, dir_op):
         if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
+                if last_update is EPOCH_TIME:
+                    warning = PrintTask(
```
kyleknap: Use ``create_warning`` instead of ``PrintTask`` from ``utils``

awscli/customizations/s3/utils.py
```diff
@@ -217,10 +217,21 @@ def get_file_stat(path):
...
+
+    try:
+        update_time = datetime.fromtimestamp(stats.st_mtime, tzlocal())
+    except ValueError:
+        # Python's fromtimestamp raises value errors when the timestamp is out
+        # of range of the platform's C localtime() function. This can cause
+        # issues when syncing from systems with a wide range of valid timestamps
+        # to systems with a lower range. Some systems support 64-bit timestamps,
+        # for instance, while others only support 32-bit. We set the time to be
+        # UNIX epoch time so that file size will take precedence in sync.
```
kyleknap: Not sure if I am following this last sentence in the comment. We set it to epoch because it since it is the lowest valid timestamp value given that this timestamp error is raised on times before epoch time. Also for downloading sync's this setting of timestamps will cause the file to get downloaded because it is older than the object in s3.

tests/unit/customizations/s3/test_utils.py
```diff
@@ -386,14 +386,18 @@ def test_get_file_stat(self):
...
 
     def test_get_file_stat_error_message(self):
+        with mock.patch('os.stat', mock.Mock(side_effect=IOError('msg'))):
+            with self.assertRaisesRegexp(ValueError, 'myfilename\.txt'):
+                get_file_stat('myfilename.txt')
+
+    def test_get_file_stat_returns_epoch_on_invalid_timestamp(self):
         patch_attribute = 'awscli.customizations.s3.utils.datetime'
         with mock.patch(patch_attribute) as f:
             with mock.patch('os.stat'):
```
kyleknap: Can we remove the os.stat patch?

kyleknap:
Looks good. Had some small comments.

awscli/customizations/s3/filegenerator.py
```diff
@@ -173,6 +173,12 @@ def list_files(self, path, dir_op):
         if not self.should_ignore_file(path):
             if not dir_op:
                 size, last_update = get_file_stat(path)
+                if last_update is EPOCH_TIME:
```
jamesls: This is...interesting.  Remind me again the issues with just raising a "could not figure out the mtime" exception and catching it here?

awscli/customizations/s3/filegenerator.py
JordonPhillips: `create_warning` will say the file is skipped, which is not the case here. Didn't seem like there was much value in creating a new function that doesn't really abstract anything.

awscli/customizations/s3/filegenerator.py
JordonPhillips: We would lose size information. I suppose we could split up `get_file_stat` into two functions which each make a stat call and that would work.

tests/unit/customizations/s3/test_utils.py
JordonPhillips: Right now it's there so we don't have to create a file. We could certainly create one, but it provides no benefit and takes up more lines.

awscli/customizations/s3/filegenerator.py
jamesls: Ok, that makes sense.  I think using an identity instead of an equality check though is very subtle.  I'd prefer a more explicit sentinel of just "we couldn't figure out the time" and let the consuming code figure out what they want to do in those scenarios.

jamesls:
Just some small comments.  Also looks like there's a failing test on travis.

awscli/customizations/s3/utils.py
```diff
@@ -22,16 +22,16 @@
...
 from botocore.compat import unquote_str
 
 from awscli.compat import six
 from awscli.compat import PY3
 from awscli.compat import queue
 
-
 HUMANIZE_SUFFIXES = ('KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB')
 MAX_PARTS = 10000
+EPOCH_TIME = datetime(1971, 1, 1, tzinfo=tzutc())
```
jamesls: Just noticed this.  I think this is technically 1970, not 1971.

JordonPhillips:
@jamesls regarding the comment:

> I think using an identity instead of an equality check though is very subtle. I'd prefer a more explicit sentinel of just "we couldn't figure out the time" and let the consuming code figure out what they want to do in those scenarios.

The only consuming code is the file generator, which has no exposed API. I added in an explanatory comment, but I'm not sure what else is to be done without doubling our calls to `os.stat` I would definitely prefer catching an error, but I'm not sure it's worth the expense.

JordonPhillips:
Though now that I think about it, I could move the datetime conversion into `filegenerator.py` and catch the error there. We'd be passing along a timestamp instead of a datetime, but there are no other consumers. Would that be preferable?

kyleknap:
I'm fine as it is right now. :ship:

jamesls:
@JordonPhillips the fact that it's in utility function in a utility module suggests that it should be of more use to just a single consumer.  To be clear, all I was suggesting was to move the opinionated decision of "couldn't figure out the timestamp so let's use the oldest supported time we can." into the code that actually needs that behavior instead of baked into a utility function.

Right now, that's communicated by relying on an identity check for a default value.  Returning something that states that more explicitly seems like it  would clarify the code.

JordonPhillips:
@jamesls How about returning `None`? It seems reasonable to expect and is pretty explicit when you're checking for it.

jamesls:
:+1: Sounds good.

jamesls:
:shipit:  Thanks for incorporating all the feedback.

PR #1658, opened 2015-11-24
Remove customizations in BucketLister that attach encoding_type

JordonPhillips:
That functionality is moving to botocore as part of boto/botocore#726

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -449,50 +449,19 @@ class BucketLister(object):
         self._date_parser = date_parser
 
     def list_objects(self, bucket, prefix=None, page_size=None):
-        kwargs = {'Bucket': bucket, 'EncodingType': 'url',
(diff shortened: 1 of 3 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -449,50 +449,19 @@ def __init__(self, client, date_parser=_date_parser):
...
-                        content['LastModified'])
-                    yield source_path, content
-
-    def _decode_keys(self, parsed, **kwargs):
-        if 'Contents' in parsed:
-            for content in parsed['Contents']:
-                content['Key'] = unquote_str(content['Key'])
-
-
-class ScopedEventHandler(object):
```
mtdowling: By the way, this is a really nice abstraction that I hadn't seen before. I wonder if it would be useful to customers that use botocore and boto3?

mtdowling:
:ship: (after the other PR of course)

PR #1664, opened 2015-12-02
Add configure add-model command

kyleknap:
This is useful for adding custom service JSON models for easier consumption in the CLI and Boto3.


awscli/customizations/configure/addmodel.py
```diff
@@ -0,0 +1,102 @@
...
+
+from awscli.customizations.commands import BasicCommand
+
+
+def _get_endpoint_prefix_to_name_mappings(session):
+    # Get the mappings of endpoint prefixes to service names from the
+    # available service models.
+    prefixes_to_services = {}
+    for service_name in session.get_available_services():
+        service_model = session.get_service_model(service_name)
```
jamesls: I get that running this command is typically a one-off, but it seems excessive to load *every* single service JSON model just to copy a service model to a directory.  I understand why this was added but can we not just require the special cases require a `--service-name` argument?

I would think the "real" fix would be capturing the botocore specific renames in a JSON file somewhere in botocore.  This isn't the only place where this information was needed.

awscli/customizations/configure/addmodel.py
```diff
@@ -0,0 +1,102 @@
...
+        endpoint_prefix = service_model.endpoint_prefix
+        # Determine the service name
+        service_name = name_mappings.get(endpoint_prefix, endpoint_prefix)
+
+    api_version = service_model.api_version
+
+    # For the model location we only want the custom data path (~/.aws/models
+    # not the one set by AWS_DATA_PATH)
+    data_path = session.get_component('data_loader').CUSTOMER_DATA_PATH
+    return os.path.join(data_path, service_name, api_version, 'service-2.json')
```
jamesls: Seems like we should try to pull the model version from the JSON file first and only fall back to a hardcoded default of 2 if the key does not exist.

jamesls:
I don't think the integrations/.../test_configure.py file should be moved over to functional/ because it's actually shelling out to `aws` via subprocess, which we've typically categorized as an integration test.  It's only if you're staying in process via `run_cmd or assert_params_for_cmd` that it's considered functional.

awscli/customizations/configure/addmodel.py
kyleknap: Yeah that's fine with me, assuming this style of the model layout stays similar with future version of the model.

awscli/customizations/configure/addmodel.py
kyleknap: I am not a fan of having users know about the services whose name that deviate from the endpoint prefix because I believe that would force the user to know too much about the internals of botocore as they would have to know a service name has a different name than its endpoint prefix. Especially since with this current implementation, users do not need to even know anything about the contents of the model to get it to work. If we are to go down this route, I would much rather have ``--service-name`` be a required parameter as that would lessen the confusion of when service name needed and when its not needed.

As to the JSON file, I considered that as well, but that would add some maintenance burden to the process when we add new services. Furthermore with how I have it, the command will even identify previously added custom commands and know where to place it if this custom command has a different endpoint than what is named. As to other places where it is needed, can you point to places where this information is needed? I am having trouble thinking of any

At this point, I would prefer making ``--service-name`` required or keeping as is. If there are other locations where this endpoint mapping logic is needed, that option would become more favorable for me.

kyleknap:
Ahh I thought this test was just a test that did not get moved over when we added the functional test suite. I will add it back since we use ``aws()`` testutil.

awscli/customizations/configure/addmodel.py
kyleknap: I talked with you about this. Decided to check for its existence in ``session.get_available_services()``, then do the mapping creation if it does not exist. Here are some stats I got from running on my machine:

```
$ time aws configure add-model --service-model file://botocore/data/ec2/2015-10-01/service-2.json 

real	0m0.528s
user	0m0.329s
sys	0m0.119s

$ time aws configure add-model --service-model file://botocore/data/cloudwatch/2010-08-01/service-2.json

real	0m2.515s
user	0m1.707s
sys	0m0.226s

$ time aws configure add-model --service-model file://botocore/data/cloudwatch/2010-08-01/service-2.json --service-name cloudwatch

real	0m0.420s
user	0m0.308s
sys	0m0.110s
```
Based on these, I think this approach should be good to go with.

kyleknap:
@jamesls I updated based on feedback. Should be good to be looked at again.

tests/unit/customizations/configure/test_addmodel.py
```diff
@@ -0,0 +1,150 @@
...
+from botocore.loaders import Loader
+
+from awscli.customizations.configure.addmodel import get_model_location
+from awscli.testutils import unittest, FileCreator
+
+
+class TestGetModelLocation(unittest.TestCase):
+    def setUp(self):
+        self.session = get_session()
+        self.files = FileCreator()
```
jamesls: It doesn't look like this directory ever gets cleaned up.

tests/unit/customizations/configure/test_addmodel.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertEqual(
+            os.path.join(
+                self.data_loader.CUSTOMER_DATA_PATH,
+                self.non_matching_service, self.default_api_version,
+                'service-2.json'), model_location)
+
+    def test_get_model_location_of_nonexistant_service(self):
+        model_location = get_model_location(
+            self.session, self._create_service_definition(
+                'nonexistant', self.default_api_version))
```
jamesls: nonexistent

tests/functional/configure/test_addmodel.py
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+    def test_add_model(self):
+        cmdline = self.prefix + ' --service-model %s' % json.dumps(
+            self.service_definition, separators=(',', ':'))
+        self.run_cmd(cmdline)
+
+        # Ensure that the model exists in the correct location.
+        self.assertTrue(
+            os.path.join(
```
jamesls: This doesn't look right.  This will always return True won't it?

tests/functional/configure/test_addmodel.py
```diff
@@ -0,0 +1,59 @@
...
+
+from awscli.testutils import BaseAWSCommandParamsTest, FileCreator
+
+
+class TestAddModel(BaseAWSCommandParamsTest):
+    prefix = 'configure add-model'
+
+    def setUp(self):
+        super(TestAddModel, self).setUp()
+        self.files = FileCreator()
```
jamesls: Doesn't look like these files are getting cleaned up.

jamesls:
Looks good, just had some comments on the tests.

kyleknap:
I updated the test. Should be good again to be looked at.

jamesls:
:shipit:  Looks good.  Don't forget to update the changelog.

Issue #1667, opened 2015-12-04
`aws s3api create-bucket` not honoring `--region` argument


timkendrick:
I've just started using the AWS CLI, and it seems that whatever I try to create a bucket using the `aws s3api create-bucket` command, it ignores both the `--region` argument passed to the command, and the `region` setting in my `~/.aws/config`.

Example output below:

```bash
$ aws --version
aws-cli/1.9.11 Python/2.7.10 Darwin/14.5.0 botocore/1.3.11
```

```bash
$ aws s3api create-bucket --bucket bd354a50-2ec4-4b5c-b919-30f0581d65a3 --region eu-west-1
{
    "Location": "/bd354a50-2ec4-4b5c-b919-30f0581d65a3"
}
$ aws s3api get-bucket-location --bucket bd354a50-2ec4-4b5c-b919-30f0581d65a3
{
    "LocationConstraint": null
}
```

...however it all works fine when I use the `aws s3 mb` command:

```bash
$ aws s3 mb s3://158d3cbb-97fa-43af-a637-992b4e80dd57 --region eu-west-1
make_bucket: s3://158d3cbb-97fa-43af-a637-992b4e80dd57/
$ aws s3api get-bucket-location --bucket 158d3cbb-97fa-43af-a637-992b4e80dd57
{
    "LocationConstraint": "eu-west-1"
}
```

Full debug output below:

```bash
$ aws s3api create-bucket --bucket bd354a50-2ec4-4b5c-b919-30f0581d65a3 --region eu-west-1 --debug
2015-12-04 15:37:16,085 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/1.9.9 Python/2.7.10 Darwin/14.5.0 botocore/1.3.9
2015-12-04 15:37:16,086 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['s3api', 'create-bucket', '--bucket', 'bd354a50-2ec4-4b5c-b919-30f0581d65a3', '--region', 'eu-west-1', '--debug']
2015-12-04 15:37:16,086 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function add_scalar_parsers at 0x107d588c0>
2015-12-04 15:37:16,086 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function inject_assume_role_provider at 0x107d28758>
[...]

PR #1673, opened 2015-12-08
Add example for creating buckets not in us-east-1

kyleknap:
Fixes https://github.com/aws/aws-cli/issues/1667

### DIFF
--- a/awscli/examples/s3api/create-bucket.rst
+++ b/awscli/examples/s3api/create-bucket.rst
@@ -7,3 +7,18 @@ Output::
   {
       "Location": "/my-bucket"
   }
+
+
+The following command creates a bucket named ``my-bucket`` in the
+``eu-west-1`` region. Regions outside of ``us-east-1`` require the appropriate
+``LocationConstraint`` to be specified in order to create the bucket in the
+desired region::
+
+    $ aws s3api create-bucket --bucket my-bucket --region eu-west-1 --create-bucket-configuration LocationConstraint=eu-west-1 
+
+
+Output::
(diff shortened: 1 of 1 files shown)


rayluo:
:ship:

jamesls:
:shipit:

timkendrick:
Original reporter here - FYI this doesn't actually fix the issue I was talking about.

The misleading `--region` argument in the `create-bucket` example is [still there](https://github.com/kyleknap/aws-cli/blame/45c283ef3177ea9256c1d4bfab2b0f52a008a54b/awscli/examples/s3api/create-bucket.rst#L3)...

kyleknap:
It is unrelated to ``create-bucket``. It just specifies with what endpoint the request will be sent. S3 is a little special because even if you specify ``eu-west-1`` as the region, we will send it to the main s3 endpoint ``s3.amazonaws.com`` because it will reroute the request to the appropriate endpoint.

For the common use case, ``--region`` is not required to be the same as the region you are trying to create the bucket. However, it **is** required for some edge cases like in signature version 4 only regions like ``eu-central-1`` or if you are straight up using signature version 4 because signature version 4 requires the correct region to be specified and we will no longer be using the ``s3.amazonaws.com`` endpoint. So the example will break if I remove that ``--region`` argument for signature version 4 cases.

timkendrick:
Thanks for the explanation @kyleknap, that makes a lot of sense. Sorry for the confusion.

kyleknap:
No problem. Glad to help!

PR #1675, opened 2015-12-10
Add aws_bash_completer

xaka:
Just to be fair (there is one for zsh already, right?), consistent with other tools that provide bash completer script out of the box and to help other projects like Homebrew (https://github.com/Homebrew/homebrew/issues/46781).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,9 +69,12 @@ Command Completion
 
 The aws-cli package includes a very useful command completion feature.
 This feature is not automatically installed so you need to configure it manually.
-To enable tab completion for bash use the built-in command ``complete``::
+To enable tab completion for bash either use the built-in command ``complete``::
 
     $ complete -C aws_completer aws
+
+Or add ``bin/aws_bash_completer`` file under ``/etc/bash_completion.d``,
+``/usr/local/etc/bash_completion.d`` or any other ``bash_completion.d`` location.
 
 For tcsh::
 
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup_options = dict(
     packages=find_packages(exclude=['tests*']),
     package_data={'awscli': ['data/*.json', 'examples/*/*.rst',
                              'examples/*/*/*.rst', 'topics/*.rst',
-                             'topics/*.json']},
+                             'topics/*.json', 'bin/aws_bash_completer']},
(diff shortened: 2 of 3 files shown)


xaka:
P.S. It's not usual bin script, but special script recognized by bash auto completer, so I'm not quite sure about the proper location. Doesn't feel right to place under bin/. Any thoughts?

rayluo:
There is already a pre-existing `aws_completer` in `bin/` so it is fine. We will take a look into this PR. Thanks!

README.rst
```diff
@@ -69,9 +69,9 @@ Command Completion
 
 The aws-cli package includes a very useful command completion feature.
 This feature is not automatically installed so you need to configure it manually.
-To enable tab completion for bash use the built-in command ``complete``::
 
-    $ complete -C aws_completer aws
+For bash please refer to bin/aws_bash_completer.  Add that file under /etc/bash_completion.d,
```
jamesls: I would prefer to have the documentation for both approaches.  We can have users prefer the bash_competion.d script, but we should still document using `complete` directly, as it's still a valid way to enable this.

xaka: I don't have strong opinion on this one, in most cases package managers take care of it for you so you don't even need to touch anything. Do you want me to keep both?

jamesls: That would be my preference yes.

xaka: Alright. Now documentation covers both approaches.

setup.py
```diff
@@ -30,7 +30,7 @@
     packages=find_packages(exclude=['tests*']),
     package_data={'awscli': ['data/*.json', 'examples/*/*.rst',
                              'examples/*/*/*.rst', 'topics/*.rst',
-                             'topics/*.json']},
+                             'topics/*.json', 'bin/aws_bash_completer']},
```
xaka: @jamesls I'm not sure whether `bin/aws_bash_completer` belongs to `package_data` or to `scripts`. It's not executable file so I decided to keep it to `package_data` for now.

jamesls: It probably does make more sense as package data, but to be consistent with the other completers , I'll just go ahead and switch this over to scripts/.

jamesls:
Thanks for the pull request!  I'm going to make one small change (mentioned above) and merge.

PR #1681, opened 2015-12-15
new and updated ec2 examples

mwunderl:


awscli/examples/ec2/describe-hosts.html
```diff
@@ -0,0 +1,402 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
jamesls: Looks like an html file was accidentally included.  Could you remove this?

PR #1687, opened 2015-12-18
Ignore ProfileNotFound in assumerole cache injector

jamesls:
If a user has provided a profile that does not exist, trying to retrieve
components/config on the session will raise ProfileNotFound.  Sometimes
this is invalid:

"ec2 describe-instances --profile unknown"

and sometimes this is perfectly valid:

"configure set region us-west-2 --profile brand-new-profile"

Because we can't know (and don't want to know) whether the customer is
trying to do something valid, we just immediately return.  If it's
invalid something else up the stack will raise ProfileNotFound,
otherwise the configure (and other) commands will work as expected.

The previous behavior had a blanket "[except Exception](https://github.com/aws/aws-cli/blob/master/awscli/customizations/assumerole.py#L43)", but
I don't think this is necessary.

I also updated the tests to have better error messages.  They
were previously failing with "filename <foo> not found".

cc @kyleknap @rayluo @mtdowling @JordonPhillips

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -1,5 +1,10 @@
 import os
 import json
+import logging
+
+from botocore.exceptions import ProfileNotFound
+
+LOG = logging.getLogger(__name__)
 
 
 def register_assume_role_provider(event_handlers):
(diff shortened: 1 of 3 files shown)


kyleknap:
:ship:

PR #1693, opened 2015-12-23
Use the list index as key name when visiting list types

jamesls:
In the backwards compatible shorthand visitor, we need to
pass the list index as the key into the parent structure
when we visit sub elements in the list.  This is necessary
in the case where we need to modify a child element and
reinsert it back into the parent structure.


kyleknap:
:ship:

PR #1698, opened 2015-12-29
Update colorama version

dslackw:


kyleknap:
Looks fine to me.

kyleknap:
Here is the CHANGELOG for reference: https://github.com/tartley/colorama/blob/master/CHANGELOG.rst

kyleknap:
Actually after some testing. It looks like this version breaks us if you are trying to pipe color:
```
$ aws ec2 describe-instances --color on --output table | less -r
```

I do not think we will be able to take in 0.3.5. Will have to revert it.

c22:
As per #2037 this behavior was not intended behavior of colorama, see [changelog entry](https://github.com/tartley/colorama/blob/master/CHANGELOG.rst) for 0.3.4

PR #1710, opened 2016-01-07
No rst file should be executable or contain CRLF

Fale:
As title, some rst files were executable, which is wrong. Also some file did contain CRLF character, that is also wrong (and non-standard, since both things were present only on a small percentage of RST files)


Fale:
PS: It could make sense to move to 644, since 664 is a pretty uncommon permissions set (also, I don't see any real-world scenario for documentation where 664 makes more sense than 644).

JordonPhillips:
This makes sense, I think it's partially because some of the people who write docs use Windows. Could you rebase, and then we can merge?

Fale:
Ok, should be align with develop now :)

Fale:
Thanks a lot :)

PR #1716, opened 2016-01-11
Update ZipFile parameter documentation to be consistent.

### LINKED ISSUE #1704: lambda update-function-code --zip-file option inconsistent with create-function --zip-file option
The `aws lambda create-function --zip-file` option help says:

````
--zip-file  (blob) The path to the zip file of the code you are upload-
       ing. Example: fileb://code.zip
````

Not sure what the (blob) pertains to.

Anyway, the `aws lambda update-function-code --zip-file` option docs say:

````
--zip-file (blob)
          Based64-encoded .zip file containing your packaged source code.
````

... which is confusing and inconsistent. The `--zip-file` option for the `update-function` command should work the same as for the `create-function` command. Instead, I understand this documentation to mean that instead of providing the path to the zip file as the value of this option, the contents of the zip file need to be provided as a base64-encoded string, which seems horrible. It would be much better to provide a value like `fileb://code.zip`.

### PR DESCRIPTION
**Pull request by JordonPhillips**
When we pull ZipFile out of the Code argument in CreateFunction we give it a docstring relevant to the CLI, but we don't give that doc string in UpdateFunctionCode. This causes confusion for the customer because the service documentation mentions you have to base64 encode the .zip, which we handle for the customer.

Closes #1704

cc @rayluo @kyleknap @mtdowling @jamesls

### DIFF
--- a/awscli/customizations/awslambda.py
+++ b/awscli/customizations/awslambda.py
@@ -59,7 +59,7 @@ def _extract_code_and_zip_file_arguments(session, argument_table, **kwargs):
 
 def _modify_zipfile_docstring(session, argument_table, **kwargs):
     if 'zip-file' in argument_table:
-        argument_table['zip-file'].argument_model.documentation = ZIP_DOCSTRING
+        argument_table['zip-file'].documentation = ZIP_DOCSTRING
 
 
 def _should_contain_zip_content(value):
--- a/awscli/customizations/awslambda.py
+++ b/awscli/customizations/awslambda.py
@@ -23,10 +23,15 @@ ERROR_MSG = (
     "--zip-file must be a file with the fileb:// prefix.\n"
     "Example usage:  --zip-file fileb://path/to/file.zip")
 
+ZIP_DOCSTRING = ('<p>The path to the zip file of the code you are uploading. '
+                 'Example: fileb://code.zip</p>')
+
 
 def register_lambda_create_function(cli):
     cli.register('building-argument-table.lambda.create-function',
                  _extract_code_and_zip_file_arguments)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: awscli/customizations/awslambda.py [hunk 1]
```diff
@@ -38,9 +43,7 @@ def validate_is_zip_file(cli_argument, value, **kwargs):
 
 def _extract_code_and_zip_file_arguments(session, argument_table, **kwargs):
     argument_table['zip-file'] = ZipFileArgument(
-        'zip-file', help_text=('The path to the zip file of the code you '
-                               'are uploading. Example: fileb://code.zip'),
-        cli_type_name='blob')
+        'zip-file', help_text=ZIP_DOCSTRING, cli_type_name='blob')
```
rayluo: So will we see the "<p> ... </p>" wrapper surrounding the help text when running "aws ... help"?

JordonPhillips: No, it just makes it into a newline.
screen shot 2016-01-11 at 1 43 38 pm

awscli/customizations/awslambda.py
```diff
@@ -54,6 +57,11 @@ def _extract_code_and_zip_file_arguments(session, argument_table, **kwargs):
     )
 
 
+def _modify_zipfile_docstring(session, argument_table, **kwargs):
+    if 'zip-file' in argument_table:
+        argument_table['zip-file'].argument_model.documentation = ZIP_DOCSTRING
```
jamesls: I would prefer to treat the argument_model as read_only.  I don't think there's anywhere else in the code where we mutate a model.  What about making the `documentation` property settable on an argument object?

jamesls:
:shipit: assuming travis ci build is good.

PR #1718, opened 2016-01-12
Update sync.rst

stephenriddell:


awscli/examples/s3/sync.rst
```diff
@@ -81,4 +81,4 @@ Output::
 
 The following ``sync`` command syncs files between two buckets in different regions::
 
-    aws s3 sync s3://my-us-west-2-bucket s3://my-us-east-1-bucket --source-region us-west-2 --region us-east-1
\ No newline at end of file
+    aws s3 sync s3://my-us-west-2-bucket s3://my-us-east-1-bucket --source-region us-west-2 --region us-east-1
```
JordonPhillips: I'm curious why there is a line difference here, I'm not seeing it.

stephenriddell: Now that you note it, I'm equally curious. I only made the change to line 71, in the github editor, and clearly didn't review exactly what changes it thought occurred.

It looks like the final line, so I'm guessing that I completely messed up the last line by changing whether the file had an empty line at the end.

I'm not familiar with the best way to fix this. Do I make the correct change then make a new pull request?

JordonPhillips: No, if you can successfully amend it you push the update to this branch. I suggest you just amend the commit and force push (it's your own fork, so yolo).

stephenriddell: Okay then. Thanks for the help.

awscli/examples/s3/sync.rst
```diff
@@ -68,7 +68,7 @@ Output::
     upload: test2.txt to s3://mybucket/test2.txt
 
 The following ``sync`` command syncs files under a local directory to objects under a specified prefix and bucket by
```
jamesls: Wait, am I reading this right?

It looks like we have a copy/paste error.  We're actually downloading from s3 to local storage so both this sentence (line 70) as well as line 72 should be updated as well.

stephenriddell: The whole thing could be worded more clearly. Yes, this is downloading from s3 to local, with a sync command. The other sync command examples do say "sync {destination} to {source}" in their descriptions, maybe updating all the examples to say "sync {destination} to match {source}" or something similar would be clearer.

I only updated the description of the mechanism which was obviously wrong, rather than the description of what was trying to be accomplished, since the language used might already describe that (it's hard to be sure).

jamesls: Fair point.  We can address that as a separate PR.

PR #1719, opened 2016-01-13
Gracefully handle encoding errors when printing s3 keys

jamesls:
When displaying S3 key names via the high level
``s3`` commands, it's possible to encounter an S3 key that
cannot be encoded using the user's locale settings.

For example, if I upload a key on linux that contains
non-ascii characters, I'll upload the key to s3 with
utf-8 encoding.

If I then try to ``aws s3 ls`` on windows (which for example
might be using an active code page of 437), then the
ls command will crash with:

UnicodeEncodeError: 'charmap' codec can't encode character '\u2714' in position 0: character maps to <undefined>

Rather than crash, we instead want to make a best effort
attempt to print the key name.  The new behavior is to print
a placeholder character for any character that can't be encoded
into the target encoding.  Example output:

```
2016-01-12 13:43:34         11 sample_keyname?Foo.txt
```

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -312,18 +312,34 @@ def uni_print(statement, out_file=None):
     """
     if out_file is None:
         out_file = sys.stdout
-    # Check for an encoding on the file.
-    encoding = getattr(out_file, 'encoding', None)
-    if encoding is not None and not PY3:
-        out_file.write(statement.encode(out_file.encoding))
-    else:
-        try:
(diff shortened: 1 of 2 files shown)


jamesls:
Rebased again latest develop and fixed py26 failure.  Ready for review.

JordonPhillips:
:boat: I do worry that people will try to round-trip this, but I still think it's a good idea

kyleknap:
Looks good. :ship:

PR #1722, opened 2016-01-13
Added `iam` command

inokappa:
Hi,

I added `iam` as the following,

```
aws iam create-policy --policy-name my-policy --policy-document file://policy
```

Please confirm.
Thank you.

Yohei


JordonPhillips:
:+1: :shipit:

PR #1724, opened 2016-01-13
Describe and enforce multipart chunksize/threshold minimum.

JordonPhillips:
According to S3's [UploadPart docs](http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html):
> Each part must be at least 5 MB in size, except the last part.

This updates the S3 Configuration topic guide to mention that limitation, and then also the TransferConfig so that it is enforced. There were already tests enforcing a minimum, though it was blanket 1 across the board.

cc @rayluo @kyleknap @jamesls @mtdowling

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -17,7 +17,7 @@ import os
 import sys
 
 from awscli.customizations.s3.utils import (
-    find_chunksize, adjust_chunksize_to_upload_limits,
+    find_chunksize, adjust_chunksize_to_upload_limits, MAX_UPLOAD_SIZE,
     find_bucket_key, relative_path, PrintTask, create_warning)
 from awscli.customizations.s3.executor import Executor
 from awscli.customizations.s3 import tasks
@@ -27,9 +27,6 @@ from awscli.compat import queue
 
 
 LOGGER = logging.getLogger(__name__)
-# Maximum object size allowed in S3.
-# See: http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
-MAX_UPLOAD_SIZE = 5 * (1024 ** 4)
 
 CommandResult = namedtuple('CommandResult',
                            ['num_tasks_failed', 'num_tasks_warned'])
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -29,6 +29,7 @@ LOGGER = logging.getLogger(__name__)
 # Maximum object size allowed in S3.
 # See: http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
 MAX_UPLOAD_SIZE = 5 * (1024 ** 4)
+MIN_UPLOAD_CHUNKSIZE = 5 * (1024 ** 2)
 
 CommandResult = namedtuple('CommandResult',
                            ['num_tasks_failed', 'num_tasks_warned'])
@@ -203,14 +204,18 @@ class S3Handler(object):
                     )
                     self.result_queue.put(warning)
                 continue
-            elif is_multipart_task and not self.params['dryrun']:
-                # If we're in dryrun mode, then we don't need the
-                # real multipart tasks.  We can just use a BasicTask
-                # in the else clause below, which will print out the
-                # fact that it's transferring a file rather than
-                # the specific part tasks required to perform the
-                # transfer.
-                num_uploads = self._enqueue_multipart_tasks(filename)
+            elif is_multipart_task:
+                if self._is_upload(filename) and not \
+                        self._is_upload_chunksize_valid(filename):
+                    continue
+                if not self.params['dryrun']:
+                    # If we're in dryrun mode, then we don't need the
+                    # real multipart tasks.  We can just use a BasicTask
+                    # in the else clause below, which will print out the
+                    # fact that it's transferring a file rather than
+                    # the specific part tasks required to perform the
+                    # transfer.
+                    num_uploads = self._enqueue_multipart_tasks(filename)
             else:
                 task = tasks.BasicTask(
                     session=self.session, filename=filename,
@@ -234,6 +239,29 @@ class S3Handler(object):
                     return False
         else:
             return False
+
+    def _is_upload(self, filename):
+        if filename.operation_name in ['upload', 'copy']:
+            return True
+
+        if filename.operation_name == 'move':
+            if filename.src_type == 'local' and filename.dest_type == 's3':
+                return True
+            elif filename.src_type == 's3' and filename.dest_type == 's3':
+                return True
+
+        return False
+
+    def _is_upload_chunksize_valid(self, filename):
+        if self.chunksize < MIN_UPLOAD_CHUNKSIZE:
+            warning_message = ("Multipart chunksize for uploads must be at "
+                               "least 5MB, but is currently %s."
+                               % self.chunksize)
+            warning = create_warning(
+                    relative_path(filename.src), warning_message)
+            self.result_queue.put(warning)
+            return False
+        return True
 
     def _enqueue_multipart_tasks(self, filename):
         num_uploads = 1
@@ -476,8 +504,8 @@ class S3StreamHandler(S3Handler):
         # First we need to create a CreateMultipartUpload task,
         # then create UploadTask objects for each of the parts.
         # And finally enqueue a CompleteMultipartUploadTask.
-
         chunksize = self.chunksize
+
         # Determine an appropriate chunksize if given an expected size.
         if self.params['expected_size']:
             chunksize = find_chunksize(int(self.params['expected_size']),
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -16,8 +16,9 @@ import math
 import os
 import sys
 
-from awscli.customizations.s3.utils import find_chunksize, \
-    find_bucket_key, relative_path, PrintTask, create_warning
+from awscli.customizations.s3.utils import (
+    find_chunksize, enforce_chunksize_limits,
+    find_bucket_key, relative_path, PrintTask, create_warning)
 from awscli.customizations.s3.executor import Executor
 from awscli.customizations.s3 import tasks
 from awscli.customizations.s3.transferconfig import RuntimeConfig
@@ -29,7 +30,6 @@ LOGGER = logging.getLogger(__name__)
 # Maximum object size allowed in S3.
 # See: http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
 MAX_UPLOAD_SIZE = 5 * (1024 ** 4)
-MIN_UPLOAD_CHUNKSIZE = 5 * (1024 ** 2)
 
 CommandResult = namedtuple('CommandResult',
                            ['num_tasks_failed', 'num_tasks_warned'])
@@ -204,18 +204,14 @@ class S3Handler(object):
                     )
                     self.result_queue.put(warning)
                 continue
-            elif is_multipart_task:
-                if self._is_upload(filename) and not \
-                        self._is_upload_chunksize_valid(filename):
-                    continue
-                if not self.params['dryrun']:
-                    # If we're in dryrun mode, then we don't need the
-                    # real multipart tasks.  We can just use a BasicTask
-                    # in the else clause below, which will print out the
-                    # fact that it's transferring a file rather than
-                    # the specific part tasks required to perform the
-                    # transfer.
-                    num_uploads = self._enqueue_multipart_tasks(filename)
+            elif is_multipart_task and not self.params['dryrun']:
+                # If we're in dryrun mode, then we don't need the
+                # real multipart tasks.  We can just use a BasicTask
+                # in the else clause below, which will print out the
+                # fact that it's transferring a file rather than
+                # the specific part tasks required to perform the
+                # transfer.
+                num_uploads = self._enqueue_multipart_tasks(filename)
             else:
                 task = tasks.BasicTask(
                     session=self.session, filename=filename,
@@ -239,29 +235,6 @@ class S3Handler(object):
                     return False
         else:
             return False
-
-    def _is_upload(self, filename):
-        if filename.operation_name in ['upload', 'copy']:
-            return True
-
-        if filename.operation_name == 'move':
-            if filename.src_type == 'local' and filename.dest_type == 's3':
-                return True
-            elif filename.src_type == 's3' and filename.dest_type == 's3':
-                return True
-
-        return False
-
-    def _is_upload_chunksize_valid(self, filename):
-        if self.chunksize < MIN_UPLOAD_CHUNKSIZE:
-            warning_message = ("Multipart chunksize for uploads must be at "
-                               "least 5MB, but is currently %s."
-                               % self.chunksize)
-            warning = create_warning(
-                    relative_path(filename.src), warning_message)
-            self.result_queue.put(warning)
-            return False
-        return True
 
     def _enqueue_multipart_tasks(self, filename):
         num_uploads = 1
@@ -288,15 +261,14 @@ class S3Handler(object):
         return num_uploads
 
     def _enqueue_range_download_tasks(self, filename, remove_remote_file=False):
-        chunksize = find_chunksize(filename.size, self.chunksize)
-        num_downloads = int(filename.size / chunksize)
+        num_downloads = int(filename.size / self.chunksize)
         context = tasks.MultipartDownloadContext(num_downloads)
         create_file_task = tasks.CreateLocalFileTask(
             context=context, filename=filename,
             result_queue=self.result_queue)
         self.executor.submit(create_file_task)
         self._do_enqueue_range_download_tasks(
-            filename=filename, chunksize=chunksize,
+            filename=filename, chunksize=self.chunksize,
             num_downloads=num_downloads, context=context,
             remove_remote_file=remove_remote_file
         )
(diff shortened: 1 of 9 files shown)


JordonPhillips:
Integration tests aren't fully working yet

JordonPhillips:
Fixed the integration test issue, it seems there's an issue with calling put_object with a sizeable body (5MB in my case) shortly after bucket creation when using virtual host style addressing. I'll be looking into that issue as well, but for now I switched to path style addressing.

awscli/customizations/s3/transferconfig.py
```diff
@@ -36,6 +43,10 @@ class RuntimeConfig(object):
     def defaults():
         return DEFAULTS.copy()
 
+    @staticmethod
+    def minimus():
```
kyleknap: minimus --> minimums

tests/integration/customizations/s3/__init__.py
```diff
@@ -25,9 +26,16 @@ def make_s3_files(session, key1='text1.txt', key2='text2.txt'):
...
+
+    if size:
+        string1 = "*" * size
+        string2 = string1
+    else:
+        string1 = "This is a test."
+        string2 = "This is another test."
+
     client = session.create_client('s3', region_name=region)
+    client.meta.events.unregister('before-sign.s3', fix_s3_host)
```
kyleknap: Why do we need to unregister this?

kyleknap:
Wow that is a lot of changes for just a validation. Question: Could you just import the ``MINIMUMS`` global and override those value in the base test class since dictionaries are mutable? That would save you an awful bunch of changes I think and limit this PR to the change you added and the validation tests that you added. Let me know what you think?

tests/integration/customizations/s3/__init__.py
JordonPhillips: This is a workaround to the broken pipe issue, where we cannot send significant string bodies.

JordonPhillips:
@kyleknap I don't think we should have tests that use invalid data, convenient though it may be.

JordonPhillips:
Updated based on @kyleknap's feedback. Since multipart downloads have no chunksize restriction, and there is no real threshold restriction, the CLI will now validate the chunksize only when it is about to perform a multipart upload. It will then skip the file and print a warning. The result looks like this:

```
$ aws s3 cp --recursive . s3://test-bucket
warning: Skipping file ./elementaryos.iso. Multipart chunksize for uploads must be at least 5MB
upload: ./test.txt to s3://test-bucket/test.txt
upload: ./script.py to s3://test-bucket/script.py
Completed 2 of 2 part(s) with 1 file(s) remaining
```

Another option would be to increase the chunksize to 5MB under the hood.

awscli/customizations/s3/s3handler.py
```diff
@@ -299,6 +310,8 @@ def _enqueue_multipart_upload_tasks(self, filename,
         # First we need to create a CreateMultipartUpload task,
         # then create UploadTask objects for each of the parts.
         # And finally enqueue a CompleteMultipartUploadTask.
+        if not self._is_upload_chunksize_valid(filename):
```
kyleknap: I think this may mess up the progress bar because it is still being counted as a file to upload/copy even though it is being skipped with a part size of zero. It even looks like it in the updated output where it says that 2 of 2 parts is completed with 1 file remaining. I would recommend putting the logic in an if statement somewhere here: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/s3handler.py#L191. Mainly before the task gets counted here: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/s3handler.py#L220. It would also save you from having to duplicate the logic in both the copy and upload multipart enqueuing methods.

awscli/customizations/s3/tasks.py
```diff
@@ -436,16 +436,19 @@ def _queue_writes_for_stream(self, body):
...
-            offset = self._part_number * self._chunk_size + amount_read
-            LOGGER.debug("Submitting IORequest to write queue.")
-            self._io_queue.put(
-                IORequest(self._filename.dest, offset, current,
-                          self._filename.is_stream)
-            )
-            LOGGER.debug("Request successfully submitted.")
-            amount_read += len(current)
-            current = body.read(iterate_chunk_size)
+        try:
```
kyleknap: What caused the need to add the try except only to re-raise it?

tests/unit/customizations/s3/__init__.py
```diff
@@ -68,13 +70,16 @@ def assert_operations_for_s3_handler(self, s3_handler, tasks,
             self.assertEqual(self.operations_called[i][1], ref_operation[1])
 
 
-def make_loc_files(file_creator):
+def make_loc_files(file_creator, size=None):
```
kyleknap: The size argument is no longer needed here.

tests/integration/customizations/s3/test_s3handler.py
```diff
@@ -375,5 +389,45 @@ def test_bucket(self):
         self.assertNotIn(self.bucket, buckets_list)
 
 
+class S3HandlerTestBucket(unittest.TestCase):
```
kyleknap: Do we need this test since we have this one: https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L1083? I would prefer keeping tests out of this test suite because it is hard to work with and does not completely test end to end like ``test_plugin.py``

kyleknap:
Approach looks better. Had a few more comments to clean it up and fix a possible output issue.

tests/integration/customizations/s3/test_s3handler.py
JordonPhillips: I think this is an artifact of a bad rebase, I'll get rid of it.

tests/unit/customizations/s3/__init__.py
JordonPhillips: This is imported in the integration tests as well as the unit tests.

JordonPhillips:
```
$ aws s3 cp --recursive . s3://test-bucket
warning: Skipping file ./elementaryos.iso. Multipart chunksize for uploads must be at least 5MB
warning: Skipping file ./elementaryos2.iso. Multipart chunksize for uploads must be at least 5MB
upload: ./test.txt to s3://test-bucket/test.txt
upload: ./script.py to s3://test-bucket/script.py
```

tests/integration/customizations/s3/__init__.py
kyleknap: I do not think my comment got saved but I was going to say to do this just set addressing style to path for the client config instead of unregistering this event.

awscli/customizations/s3/s3handler.py
```diff
@@ -235,6 +240,28 @@ def _is_multipart_task(self, filename):
...
+            if filename.src_type == 'local' and filename.dest_type == 's3':
+                return True
+            elif filename.src_type == 's3' and filename.dest_type == 's3':
+                return True
+
+        return False
+
+    def _is_upload_chunksize_valid(self, filename):
+        if self.chunksize < MIN_UPLOAD_CHUNKSIZE:
+            warning_message = "Multipart chunksize for uploads " \
```
kyleknap: You might want to say what the provided chunksize is in the warning.

kyleknap:
Sorry about the couple bits of feedback that I missed. Also could you add a test to explicitly make sure a warning is thrown for a chunksize that is too small? Does not look like there was one added in the current diff. Otherwise looks good.

JordonPhillips:
Updated :+1:

kyleknap:
Changes look good. Could you update the test to use addressing_style path instead of unregistering the event when creating the client. Also lets get the tests to pass before merging this. But otherwise :ship:

JordonPhillips:
I ended up having to go a different route for testing, but it is tested

jamesls:
Just catching up on this, but I had a quick question about the expected behavior.  

We already upgrade the chunk value if their chunk size is too small and we'd exceed the part limit.  Seems reasonable to do the same thing if the chunk size value itself is too small.  Was there a reason not to do this?

JordonPhillips:
@jamesls Initially the idea was just to do validation, but upgrading under the hood could work too. Is there any reason why that might negatively impact a customer though? I can't think of one off the top of my head. Also want to know what @kyleknap thinks about that.

jamesls:
To expand on my previous comment, I'd suggest two options:

a) hard fail, the user provided a bad value, don't try and transfer anything until they fix this value.
b) upgrade to the lowest allowed minimum value.

I think this middle ground where "we'll try to transfer any non multipart files but skip multipart uploads" is the least intuitive of all three options and will potentially leave the destination in an avoidable inconsistent state.

JordonPhillips:
Changelog

* switched to auto-increasing the chunksize when necessary
* added logging statements whenever we change the chunksize
* added chunksize modification to unbound streaming uploads

Some more on that last point: the `find_chunksize` function required file size, which can't always be known in a streaming upload. Because of this, it was only run when we had an expected size. However, we still need to verify that the chunksize is within the size restrictions S3 imposes on uploads.

tests/unit/customizations/s3/test_utils.py
```diff
@@ -122,6 +123,15 @@ def test_small_chunk(self):
         size = 8 * (1024 ** 2)
         self.assertEqual(find_chunksize(size, chunksize), chunksize)
 
+    def test_small_chunk(self):
+        """
+        This test ensures that if the ``chunksize`` is below the minimum
+        threshold, noseit is automatically raised to the minimum.
```
kyleknap: ``noseit``? Not sure what you meant.

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -914,6 +915,25 @@ def test_upload_stream_with_expected_size(self):
...
+                multipart_threshold=1, multipart_chunksize=2))
+        s3handler.executor = mock.Mock()
+        fileinfo = FileInfo('filename', operation_name='upload',
+                            is_stream=True)
+        with MockStdIn(b'bar'):
+            s3handler._enqueue_multipart_upload_tasks(fileinfo, b'')
+        submitted_tasks = s3handler.executor.submit.call_args_list
+        # Determine what the chunksize was changed to from one of the
+        # UploadPartTasks.
+        changed_chunk_size = submitted_tasks[1][0][0]._chunk_size
```
kyleknap: Not a fan of accessing the internal ``_chunk_size``. Not sure there is a great way to grab this though without large amounts of rewrites...

tests/unit/customizations/s3/test_utils.py
JordonPhillips: Must have been typing with the wrong focus and didn't notice - oops!

tests/unit/customizations/s3/test_s3handler.py
JordonPhillips: Yeah, I just copied a previous test and made a few modifications.

kyleknap:
Looks good. I like how you broke up the utils functions and logged when changes were made. :ship:

tests/integration/customizations/s3/__init__.py
```diff
@@ -25,9 +26,16 @@ def make_s3_files(session, key1='text1.txt', key2='text2.txt'):
...
-    client = session.create_client('s3', region_name=region)
+
+    if size:
+        string1 = "*" * size
+        string2 = string1
+    else:
+        string1 = "This is a test."
+        string2 = "This is another test."
+
+    config = Config(s3={"addressing_style": "path"})
```
jamesls: Can you discuss why this is needed?  Your description mentioned you were having an issue with new buckets and large files, but I want to make sure we know *exactly* what was causing the test issues before switching over to path style.

JordonPhillips: The issue is boto/botocore#760. The issue is that we're forcing a large packet to an endpoint that is trying to tell us to re-direct. The service closes the connection since we continue to send data. This doesn't occur with path style since there are no DNS propagation requirements, and doesn't occur with small payloads because the service will just forward on small packets.

jamesls: Cool, so is it worth waiting until the botocore issue's resolved?

Long term, assuming the botocore issue is fixed, would we want to prefer path style vs. virtual hosted addressing for these integration tests?

awscli/topics/s3-config.rst
```diff
@@ -110,6 +110,7 @@ multipart_threshold
 -------------------
 
 **Default** - ``8MB``
+**Minimum** - ``5MB``
```
jamesls: We should probably update these docs and add a note that this will not apply to downloads.

awscli/customizations/s3/utils.py
```diff
@@ -283,17 +286,45 @@ def find_chunksize(size, current_chunksize):
...
+    elif chunksize < MIN_UPLOAD_CHUNKSIZE:
+        LOGGER.debug(
+            "Chunksize less than minimum chunksize. Setting to %s from %s." %
+            (human_readable_size(MIN_UPLOAD_CHUNKSIZE), chunksize_human))
+        return MIN_UPLOAD_CHUNKSIZE
     else:
         return chunksize
 
 
+def enforce_max_parts(size, current_chunksize):
```
jamesls: Can we add docstrings for this.  One of things that would help is to make it clear that this is computing a new chunk size (and not really enforcing chunk size).

It would be helpful for this chunk adjustment functions to indicate whether they are guaranteed to increase/decrease the chunk size.  For example, this function will only ever return the current chunk size or a greater chunk size, never a smaller chunksize.

awscli/customizations/s3/s3handler.py
```diff
@@ -476,8 +475,8 @@ def _enqueue_multipart_upload_tasks(self, filename, payload=None):
         # First we need to create a CreateMultipartUpload task,
         # then create UploadTask objects for each of the parts.
         # And finally enqueue a CompleteMultipartUploadTask.
+        chunksize = enforce_chunksize_limits(self.chunksize)
 
-        chunksize = self.chunksize
         # Determine an appropriate chunksize if given an expected size.
         if self.params['expected_size']:
             chunksize = find_chunksize(int(self.params['expected_size']),
```
jamesls: Can this be consolidated into the line below with the check for expected size?  It was a little hard to follow given `find_chunksize` also calls `enforce_chunksize_limits`, but is only called when an expected size is given.

jamesls:
Overall looks good.  Just had a few questions/comments.

tests/integration/customizations/s3/__init__.py
JordonPhillips: Removed

awscli/topics/s3-config.rst
JordonPhillips: :+1:

awscli/customizations/s3/utils.py
JordonPhillips: :+1:

awscli/customizations/s3/utils.py
```diff
@@ -278,22 +281,72 @@ def create_warning(path, error_message, skip_file=True):
...
+    parts in a multipart upload is not greater than the ``MAX_PARTS``, and that
+    the chunksize is not less than the minimum chunksize.
+
+    :param size: The size of the file to upload
+    :param current_chunksize: The currently configured chunksize
+    :return: If the given chunksize is valid, it is returned. Otherwise a valid
+        chunksize is calculated and returned.
+    """
+    size = adjust_chunksize_for_max_parts(size, current_chunksize)
+    return adjust_chunksize_to_upload_limits(size)
```
jamesls: It's been a while since I've looked at this so I want to make sure I understand this.

`adjust_chunksize_for_max_parts` only ever scales up the chunk size based on max number of allowed parts

`adjust_chunksize_to_upload_limits` can increase or decrease the chunk size based on max part_size.

So it's possible that `adjust_chunksize_to_upload_limits` can decrease the part size to something that would have been rejected/modified by `adjust_chunksize_for_max_parts` correct?

And the reason this is ok is because we're relying on the fact that something up the stack will reject the transfer anyways because at that point you've likely exceeded the max allowed file size in S3?

tests/integration/customizations/s3/__init__.py
```diff
@@ -25,9 +22,15 @@ def make_s3_files(session, key1='text1.txt', key2='text2.txt'):
...
-    client = session.create_client('s3', region_name=region)
+
+    if size:
+        string1 = "*" * size
+        string2 = string1
+    else:
+        string1 = "This is a test."
+        string2 = "This is another test."
+
+    client = session.create_client('s3', region_name=region, config=config)
```
jamesls: Doesn't this generate a `NameError`?  Where's `config` coming from?

awscli/customizations/s3/utils.py
JordonPhillips: Right. If the only way to satisfy max parts is to go over the single upload limit, it will by definition be above the max total size limit.

tests/integration/customizations/s3/__init__.py
JordonPhillips: Yeah, odd that the tests aren't failing on that.

JordonPhillips: Updated, but I'm gonna run the integration tests to make sure it's all good.

awscli/customizations/s3/utils.py
jamesls: For a pure API standpoint, I would expect a utility function like this to raise an exception if the constraints can't be met (instead of giving back a value that something somewhere else should reject), but I would be ok with documentation/comments explaining this for now.

JordonPhillips: Added an exception and a test for said exception.

jamesls:
:shipit:  Assuming travis is good.

PR #1727, opened 2016-01-14
Classic link dns support examples

mwunderl:


awscli/examples/ec2/describe-vpc-classic-link-dns-support.rst
```diff
@@ -0,0 +1,22 @@
...
+Command::
+
+  aws ec2 describe-vpc-classic-link-dns-support
+
+Output::
+
+  {
+    "Vpcs": [
+      {
+        "VpcId": "vpc-614cc409",
```
JordonPhillips: These look real, any reason they aren't like those in the other file (`vpc-88888888`)?

jamesls: I think the difference is that that the `vpc-888888` is for inputs and this is showing output.  Looking at other examples, this seems to be consistent, so I think this change is fine as is.

PR #1728, opened 2016-01-14
Cloudfront create distribution

rayluo:
Implement a new optional `--origin-domain-name` parameter for `aws cloudfront create-distribution`.

The following command shows an example for an S3 domain::

    aws cloudfront create-distribution \
      --origin-domain-name my-bucket.s3.amazonaws.com

Example in `aws cloudfront create-distribution help` has also been updated to reflect the new parameter.

### DIFF
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -90,6 +90,9 @@ class OriginDomainNameArgument(CustomArgument):
         origin_id = caller_reference(prefix='origin_id')
         item = {"Id": origin_id, "DomainName": value, "OriginPath": ""}
         if value.endswith('.s3.amazonaws.com'):
+            # We do not need to detect '.s3[\w-].amazonaws.com' as S3 buckets,
+            # because CloudFront treats GovCloud S3 buckets as custom domain.
+            # http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/setting-up-cloudfront.html
             item["S3OriginConfig"] = {"OriginAccessIdentity": ""}
         else:
             item["CustomOriginConfig"] = {
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -40,7 +40,7 @@ def register(event_handler):
     event_handler.register(
         'operation-args-parsed.cloudfront.create-distribution',
         validate_mutually_exclusive_handler(
-            ['origin-domain-name'], ['distribution-config']))
+            ['origin_domain_name'], ['distribution_config']))
 
 
 def caller_reference(prefix="cli"):
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -95,13 +95,13 @@ class OriginDomainNameArgument(CustomArgument):
             item["CustomOriginConfig"] = {
                 'HTTPPort': 80, 'HTTPSPort': 443,
                 'OriginProtocolPolicy': 'http-only'}
-        parameters['DistributionConfig'] = {  # The minimal config
+        parameters['DistributionConfig'] = {
             "CallerReference": caller_reference(),
             "Origins": {"Quantity": 1, "Items": [item]},
             "DefaultCacheBehavior": {
                 "TargetOriginId": origin_id,
                 "ForwardedValues": {
-                    "QueryString": True,
+                    "QueryString": False,
                     "Cookies": {
                         "Forward": "none"
                     }
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -33,6 +33,19 @@ def register(event_handler):
         'operation-args-parsed.cloudfront.create-invalidation',
         validate_mutually_exclusive_handler(['invalidation_batch'], ['paths']))
 
+    # Provides a --origin-domain-name for "aws cloudfront create-distribution"
+    event_handler.register(
+        'building-argument-table.cloudfront.create-distribution',
+        _add_origin_domain_name)
+    event_handler.register(
+        'operation-args-parsed.cloudfront.create-distribution',
+        validate_mutually_exclusive_handler(
+            ['origin-domain-name'], ['distribution-config']))
+
+
+def caller_reference(prefix="cli"):
+    return '%s-%s-%s' % (prefix, int(time.time()), random.randint(1, 1000000))
+
 
 def _add_paths(argument_table, **kwargs):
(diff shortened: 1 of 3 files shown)


awscli/customizations/cloudfront.py
```diff
@@ -50,14 +63,61 @@ def __init__(self):
...
+            'and --distribution-config are mututally exclusive.'
+        )
+        super(OriginDomainNameArgument, self).__init__(
+            'origin-domain-name', help_text=doc)
+
+    def add_to_params(self, parameters, value):
+        if value is None:
+            return
+        origin_id = caller_reference(prefix='origin_id')
+        item = {"Id": origin_id, "DomainName": value, "OriginPath": ""}
```
jamesls: Should this include some of the origin domain name as part of the `origin_id`?  The console does something similar here.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +63,61 @@ def __init__(self):
...
+        if value.endswith('.s3.amazonaws.com'):
```
jamesls: Won't this be an issue with Govcloud?

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +63,61 @@ def __init__(self):
...
+            item["S3OriginConfig"] = {"OriginAccessIdentity": ""}
+        else:
+            item["CustomOriginConfig"] = {
+                'HTTPPort': 80, 'HTTPSPort': 443,
+                'OriginProtocolPolicy': 'http-only'}
+        parameters['DistributionConfig'] = {  # The minimal config
```
jamesls: pep8, move the inline comment above the line of code.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +63,61 @@ def __init__(self):
...
+        super(OriginDomainNameArgument, self).__init__(
+            'origin-domain-name', help_text=doc)
+
+    def add_to_params(self, parameters, value):
+        if value is None:
+            return
+        origin_id = caller_reference(prefix='origin_id')
+        item = {"Id": origin_id, "DomainName": value, "OriginPath": ""}
+        if value.endswith('.s3.amazonaws.com'):
+            item["S3OriginConfig"] = {"OriginAccessIdentity": ""}
```
jamesls: This condition isn't tested (where a user provides an S3 bucket).

jamesls:
Just had some minor feedback.

tests/functional/cloudfront/test_create_distribution.py
```diff
@@ -0,0 +1,52 @@
...
+                'CallerReference': mock.ANY,
+                'Comment': '',
+                'Enabled': True,
+                'DefaultCacheBehavior': mock.ANY,
+                },
+            }
+        self.run_cmd(cmdline)
+        self.assertEqual(self.last_kwargs, result)
+
+    def test_both_distribution_config_and_origin_domain_name(self):
```
kyleknap: Let's add a test if just ``--distribution-config`` is provided to make sure it is used.

tests/functional/cloudfront/test_create_distribution.py
```diff
@@ -0,0 +1,52 @@
...
+                        'CustomOriginConfig': mock.ANY,
+                        'DomainName': 'foo.com',
+                        'Id': mock.ANY,
+                        'OriginPath': '',
+                    }]
+                },
+                'CallerReference': mock.ANY,
+                'Comment': '',
+                'Enabled': True,
+                'DefaultCacheBehavior': mock.ANY,
```
kyleknap: Maybe flush this value out since it is a dictionary, like you did for ``Items``, instead of doing ``mock.ANY``.

kyleknap:
Looks good. Agree with James' comments and had a couple more comments about the test as well.

awscli/customizations/cloudfront.py
rayluo: Deriving from domain name looks better, but there is a risk of duplicated ID. In the console, duplication is less a problem because it will provide a visual error message and the customer just need to type something else in the input box. But in CLI, there is no easy way for customer to get around this. So I would suggest to just use a basically random string here. Feel free to let me know if there is better solution.

awscli/customizations/cloudfront.py
rayluo: Looks like CloudFront treats S3 in GovCloud as custom domain anyway.

> "Due to the isolation of the AWS GovCloud (US) region, using CloudFront with your AWS GovCloud (US) resources is analogous to using CloudFront with a non-AWS origin server." Quoted from http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/setting-up-cloudfront.html

tests/functional/cloudfront/test_create_distribution.py
rayluo: OK. New test case is added now. https://github.com/aws/aws-cli/pull/1728/files#diff-73c3c69f0cfb79a5521960a3fe1ad3edR67

tests/functional/cloudfront/test_create_distribution.py
rayluo: We discussed this off-line. The `DefaultCacheBehavior` is not really the focus of these test cases, it will just contain whatever was provided as input. So we can choose to not explicitly write the content of DefaultCacheBehavior here, otherwise we will duplicate them in 3 different test cases (and there will be another 2 test cases in the update-distribution PR).

awscli/customizations/cloudfront.py
jamesls: @rayluo I don't think there's any risk of a duplicate ID, at least not anymore than your current approach.  I'm just suggesting to include the origin domain name in the origin_id, not to *only* use the domain name.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +63,64 @@ def __init__(self):
...
+            'The domain name for your origin. Note: --origin-domain-name '
+            'and --distribution-config are mututally exclusive.'
+        )
+        super(OriginDomainNameArgument, self).__init__(
+            'origin-domain-name', help_text=doc)
+
+    def add_to_params(self, parameters, value):
+        if value is None:
+            return
+        origin_id = caller_reference(prefix='origin_id')
```
jamesls: Also, this function should probably be renamed.  It's now no longer used for just `caller_preference`.

jamesls:
Just had some small feedback, otherwise looks good.

rayluo:
Replaced by #1736

PR #1730, opened 2016-01-14
Use dictionary form for CopySource arg

jamesls:
Ensures we never run into any issues trying to encoding
the S3 keys for the user when copying objects.

This should be a (mostly) transparent change to the CLI because the unit tests still run through the `before-call` handlers in the tests, so we should still see the serialized header version as the expected result.  Only small change to the tests was that we also perform URL encoding in `before-call` so there was a unicode test case that needed to be updated.

Depends on https://github.com/boto/botocore/pull/758.


kyleknap:
:ship:

PR #1731, opened 2016-01-14
Implement update-distribution --default-root-object

rayluo:
Implement a new optional `--default-root-object` parameter for `aws cloudfront update-distribution`.

The following command shows an example for changing default root object::

    aws cloudfront update-distribution --default-root-object index.html

Example in `aws cloudfront update-distribution help` has also been updated to reflect the new parameter.

### DIFF
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -17,6 +17,7 @@ import random
 import rsa
 from botocore.utils import parse_to_aware_datetime
 from botocore.signers import CloudFrontSigner
+from botocore.session import Session
 
 from awscli.arguments import CustomArgument
 from awscli.customizations.utils import validate_mutually_exclusive_handler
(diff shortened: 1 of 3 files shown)


awscli/customizations/cloudfront.py
```diff
@@ -58,6 +70,44 @@ def add_to_params(self, parameters, value):
...
+
+    def add_to_params(self, parameters, value):
+        if value is not None:
+            parameters[self.name] = value
+
+
+def _update_distribution(parsed_globals, call_parameters, **kwargs):
+    if ('default-root-object' in call_parameters
+            and 'DistributionConfig' not in call_parameters):
+        client = Session().create_client(
```
kyleknap: -1. Never create a new session. Use the existing session.

awscli/customizations/cloudfront.py
```diff
@@ -58,6 +70,44 @@ def add_to_params(self, parameters, value):
...
+            'default-root-object', help_text=doc)
+        self.session = session
+
+    def add_to_params(self, parameters, value):
+        if value is not None:
+            parameters[self.name] = value
+
+
+def _update_distribution(parsed_globals, call_parameters, **kwargs):
+    if ('default-root-object' in call_parameters
```
kyleknap: Why would this be in the ``call_parameters``?  It should only be in the ``parsed_args``. We should not be putting parameters into the call_parameters in the first place that has no chance to be used by the client.

awscli/customizations/cloudfront.py
```diff
@@ -58,6 +70,44 @@ def add_to_params(self, parameters, value):
                 }
 
 
+def _add_default_root_object(argument_table, **kwargs):
+    argument_table['distribution-config'].required = False
+    argument_table['default-root-object'] = DefaultRootObjectArgument(
+        kwargs['session'])
```
kyleknap: You can just add session to the signature to retrieve it. You do not need to pull it from kwargs.

kyleknap:
Nevermind, looking at wrong PR. I am going to look at the one that succeeds it and reapply my comments there.

rayluo:
Replaced by #1736

PR #1732, opened 2016-01-15
Fix bug in S3 handler for copying objects

jamesls:
Need to use the source key not the destination key.

This wasn't caught because the unit test was using the same
key name for source and destination.  I've updated the unit
tests so we can catch this going forward.

I've also confirmed the failing integration tests now pass.


JordonPhillips:
:boat:

kyleknap:
:ship:

PR #1734, opened 2016-01-18
Update requirements.txt for rsa

kyleknap:
Noticed that these were out of line when reading up on a recently discovered vulnerability in rsa that required a [version bump](https://bitbucket.org/sybren/python-rsa/src/7d8e02577dfc1829f15896cbf795a4cfe68a89bd/CHANGELOG.txt?at=default&fileviewer=file-view-default). Fortunately our [range dependency](https://github.com/aws/aws-cli/blob/develop/setup.py#L12) already handles it, but the ``requirements.txt`` did not match the one listed in the ``setup.py``.


JordonPhillips:
:boat:

PR #1736, opened 2016-01-19
Cloudfront create-distribution and update-distribution customization

rayluo:
Implement a new optional `--origin-domain-name` parameter and `--default-root-object` for `aws cloudfront create-distribution`. The following command shows an example for an S3 domain::

    aws cloudfront create-distribution \
      --origin-domain-name my-bucket.s3.amazonaws.com --default-root-object index.html

Implement a new optional `--default-root-object` parameter for `aws cloudfront update-distribution`. The following command shows an example for changing default root object::

    aws cloudfront update-distribution --default-root-object index.html

Example in `aws cloudfront create-distribution help` and `aws cloudfront update-distribution help` has also been updated to reflect the new parameters.

This PR replaces #1728 and #1731.

### DIFF
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -34,36 +34,31 @@ def register(event_handler):
         'operation-args-parsed.cloudfront.create-invalidation',
         validate_mutually_exclusive_handler(['invalidation_batch'], ['paths']))
 
-    update_distribution_args = [
-        ExclusiveArgument(
-            'default-root-object', 'distribution-config', help_text=(
-                'The object that you want CloudFront to return (for example, '
-                'index.html) when a viewer request points to your root URL.')),
-        ]
-    _setup_custom_params(
-        event_handler, 'update-distribution', update_distribution_args,
-        _update_distribution)
-
-    create_distribution_args = update_distribution_args + [
-        ExclusiveArgument(
-            'origin-domain-name', 'distribution-config',
-            help_text='The domain name for your origin.'),
-        ]
-    _setup_custom_params(
-        event_handler, 'create-distribution', create_distribution_args,
-        _create_distribution)
-
-
-def _setup_custom_params(event_handler, subcmd, custom_args, callback):
-    # Usage: _setup_custom_params(..., "create-distribution", [...], callable)
     event_handler.register(
-        'operation-args-parsed.cloudfront.%s' % subcmd,
+        'operation-args-parsed.cloudfront.create-distribution',
         validate_mutually_exclusive_handler(
-            [arg.py_name for arg in custom_args], ['distribution_config']))
+            ['default_root_object', 'origin_domain_name'],
+            ['distribution_config']))
     event_handler.register(
-        'building-argument-table.cloudfront.%s' % subcmd,
-        lambda **kwargs: _add_exclusive_arguments(args=custom_args, **kwargs))
-    event_handler.register('calling-command.cloudfront.%s' % subcmd, callback)
+        'building-argument-table.cloudfront.create-distribution',
+        lambda argument_table, **kwargs: argument_table.__setitem__(
+            'origin-domain-name', OriginDomainName(argument_table)))
+    event_handler.register(
+        'building-argument-table.cloudfront.create-distribution',
+        lambda argument_table, **kwargs: argument_table.__setitem__(
+            'default-root-object', CreateDefaultRootObject(argument_table)))
+
+    context = {}
+    event_handler.register('top-level-args-parsed', context.update)
+    event_handler.register(
+        'operation-args-parsed.cloudfront.update-distribution',
+        validate_mutually_exclusive_handler(
+            ['default_root_object'], ['distribution_config']))
+    event_handler.register(
+        'building-argument-table.cloudfront.update-distribution',
+        lambda argument_table, **kwargs: argument_table.__setitem__(
+            'default-root-object', UpdateDefaultRootObject(
+                context=context, argument_table=argument_table)))
 
 
 def unique_string(prefix='cli'):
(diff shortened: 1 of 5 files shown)


awscli/customizations/cloudfront.py
```diff
@@ -50,14 +86,83 @@ def __init__(self):
...
+            "Comment": "",
+        }
+        if 'default-root-object' in call_parameters:
+            call_parameters['DistributionConfig']['DefaultRootObject'] = \
+                call_parameters.pop('default-root-object')
+
+
+def _update_distribution(parsed_globals, call_parameters, **kwargs):
+    if call_parameters and 'DistributionConfig' not in call_parameters:
+        client = Session().create_client(
```
jamesls: Use the existing session, don't create a new one.

kyleknap: Agreed.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +86,83 @@ def __init__(self):
...
+            'cloudfront', region_name=parsed_globals.region,
+            endpoint_url=parsed_globals.endpoint_url,
+            verify=parsed_globals.verify_ssl)
+        response = client.get_distribution_config(Id=call_parameters['Id'])
+        call_parameters['IfMatch'] = response['ETag']
+        call_parameters['DistributionConfig'] = response['DistributionConfig']
+        if 'default-root-object' in call_parameters:
```
kyleknap: Why would this be in the ``call_parameters``? It should only be in the ``parsed_args``. We should not be putting parameters into the ``call_parameters`` in the first place that has no chance to be used by the client.

awscli/customizations/cloudfront.py
```diff
@@ -33,6 +34,41 @@ def register(event_handler):
...
+def _setup_custom_params(event_handler, subcmd, custom_args, callback):
+    # Usage: _setup_custom_params(..., "create-distribution", [...], callable)
+    event_handler.register(
+        'operation-args-parsed.cloudfront.%s' % subcmd,
+        validate_mutually_exclusive_handler(
+            [arg.py_name for arg in custom_args], ['distribution_config']))
+    event_handler.register(
+        'building-argument-table.cloudfront.%s' % subcmd,
+        lambda **kwargs: _add_exclusive_arguments(args=custom_args, **kwargs))
+    event_handler.register('calling-command.cloudfront.%s' % subcmd, callback)
```
jamesls: The typical way this is done with other customizations is to override `add_to_params` in a custom arg class.  Is there a reason this needs to be different from the other customizations and use `calling-command`?  Given you already have a custom arg class `ExclusiveArgument`, it doesn't seem like you gain anything by using `calling-command`.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +86,83 @@ def __init__(self):
...
+
+
+class ExclusiveArgument(CustomArgument):
+    DOC = '%s This argument and --%s are mututally exclusive.'
+
+    def __init__(self, name, exclusive_to, help_text=''):
+        super(ExclusiveArgument, self).__init__(
+            name, help_text=self.DOC % (help_text, exclusive_to))
+
+    def add_to_params(self, parameters, value):
```
kyleknap: I am debating if I like this idea anymore. I think it would make the code much cleaner if you just made ``add_to_params`` to just return and then rely on ``parsed_args`` in the handler that you register to ``calling-command`` event. The nice part about this approach is their is no more of this popping you have to do in the callbacks. Anybody else have ideas?

jamesls: I'm still missing why we don't want to just add this logic into add_to_params so all this logic is contained in the custom arg class.  This is how other customizations typically do this.  What's the advantage in doing this differently?

awscli/customizations/cloudfront.py
kyleknap: True. I like this more than the idea I suggested later on. Also thanks to the handler that does the mutual exclusive check it will eventually be validated so you should not have to worry about call params being clobbered. I would much prefer that the callbacks were added to ``add_to_params``. It should make the code even cleaner.

awscli/customizations/cloudfront.py
kyleknap: See your original comment you made. We are a little out of sync here, but I like your suggestion.

kyleknap:
Looks fine. The main thing for me is that there are a lot of handlers getting registered, which makes it very confusing to follow and at times seems a bit unnecessary. It would be much better if you baked the logic into the arguments that you create.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +86,83 @@ def __init__(self):
...
+            "CallerReference": unique_string(),
+            "Origins": {"Quantity": 1, "Items": [item]},
+            "DefaultCacheBehavior": {
+                "TargetOriginId": origin_id,
+                "ForwardedValues": {
+                    "QueryString": False,
+                    "Cookies": {"Forward": "none"},
+                },
+                "TrustedSigners": {
+                    "Enabled": True,
```
jamesls: What does this value mean?  I noticed that the console creates this with a value of `False` by default.  Does this matter?

awscli/customizations/cloudfront.py
rayluo: I know most ad-hoc logic can typically be done with other customizations is to override add_to_params in a custom arg class. In this case, I thought we need to collect multiple (2) parameters before we finalize the request. So, after some trial-and-error, I ended up using `calling-command` as the final checkpoint. If you guys know some other customization to do it differently, please give me a nudge. I'm more than happy to adjust it for a better way!

awscli/customizations/cloudfront.py
jamesls: I looked into this a little more.  I was initially confused because this didn't sync up with what the console says, but this will require you to use presigned cloudfront urls to access the content.

rayluo: It is a good point to maintain the parity with console, so I will change it to False in next update.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +81,103 @@ def __init__(self):
...
+
+
+class UpdateDefaultRootObject(CreateDefaultRootObject):
+    def __init__(self, context, argument_table):
+        super(UpdateDefaultRootObject, self).__init__(argument_table)
+        self.context = context
+
+    def add_to_params(self, parameters, value):
+        if value is not None:
+            client = self.context['session'].create_client(
```
kyleknap: Cool. I think we are fine here. I was worried for a second that the validation of the mutually exclusive arguments happens after ``add_to_params``, meaning we would have an unnecessary client call. But it isn't, so we are fine.

kyleknap:
Thanks for updating. Looks good. :ship:

awscli/customizations/cloudfront.py
```diff
@@ -33,6 +33,36 @@ def register(event_handler):
...
         validate_mutually_exclusive_handler(['invalidation_batch'], ['paths']))
 
+    event_handler.register(
+        'operation-args-parsed.cloudfront.create-distribution',
+        validate_mutually_exclusive_handler(
+            ['default_root_object', 'origin_domain_name'],
+            ['distribution_config']))
+    event_handler.register(
+        'building-argument-table.cloudfront.create-distribution',
+        lambda argument_table, **kwargs: argument_table.__setitem__(
```
jamesls: Why are we calling `__setitem__` directly?  That's almost never necessary.

awscli/customizations/cloudfront.py
```diff
@@ -50,14 +80,103 @@ def __init__(self):
...
+                'index.html) when a viewer request points to your root URL.'))
+
+    def add_to_params(self, parameters, value):
+        if value is not None:
+            parameters.setdefault(
+                'DistributionConfig', self.distribution_config_template())
+            parameters['DistributionConfig']['DefaultRootObject'] = value
+
+
+class UpdateDefaultRootObject(CreateDefaultRootObject):
```
jamesls: Can we update the docs for this to say that if you don't provide `distribution-config` we'll make a `get-distribution-config` call?

jamesls:
:shipit: Looks good.  One small comment before merging.

awscli/customizations/cloudfront.py
rayluo: To document the off-line conversation, the `__setitem__` here is needed because it is inside a lambda body.

awscli/customizations/cloudfront.py
rayluo: Done.

PR #1747, opened 2016-01-23
AWSCLI-1708 : Fix for user data string parameter instead of Base64 en…

rekhajoshm:
AWSCLI-1708 : Fix for user data string parameter instead of Base64 encoding
This is for enforced clarity.


kyleknap:
Sounds good. Thanks! Merging.

PR #1749, opened 2016-01-26
Feature/support private clusters

edgarklerks:
The use of a private cluster breaks the aws emr toolchain. The ssh subcommand will fail, because there is no public dns name associated with a private cluster. And in the subcommand describe-cluster the expectation is broken, that a cluster_id can be used to retrieve a dns name of the master node.

To reproduce the issues, start a cluster on a private subnet (as specified by the documentation of EMR), then run:

     aws emr ssh --cluster-id <cluster-id> --key-pair-file <your key>

It will fail, because it can't retrieve the dns name.

Furthermore there is no longer a way to translate a cluster_id into a hostname, when using a private cluster:

    aws emr describe-cluster --cluster-id <cluster-id>

MasterPublicDnsName is empty.

I added a field MasterPrivateDnsName to describe_cluster and let aws emr ssh use MasterPrivateDnsName as dns name in the case MasterPublicDnsName is an empty string.

### DIFF
--- a/awscli/customizations/emr/describecluster.py
+++ b/awscli/customizations/emr/describecluster.py
@@ -47,11 +47,16 @@ class DescribeCluster(Command):
             cluster_id=parsed_args.cluster_id,
             parsed_globals=parsed_globals)
 
+        master_private_dns = self._find_master_private_dns(
+            cluster_id=parsed_args.cluster_id,
+            parsed_globals=parsed_globals)
+
         constructed_result = self._construct_result(
             describe_cluster_result,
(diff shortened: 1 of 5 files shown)


kyleknap:
Thanks for the pull request. We will look into it and review it.

kyleknap:
@edgarklerks thanks for the pull request. There was some simplification that could have been done in terms of the architecture of the emr commands in fixing this issue so I just merged a PR that should fix the issue you are targeting in your PR. Let us know if you are still running into the issue.

PR #1757, opened 2016-01-29
Fix travis on a non-pr of master

kyleknap:
The environment variable is "false" if it is a PR. However that equates to
true as it is interpreted as a string

Same as this one: https://github.com/boto/botocore/pull/778


mtdowling:
:ship:

PR #1758, opened 2016-01-29
Add aws-exec-read ACL value for s3 commands

kyleknap:
Reference http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html


JordonPhillips:
:boat:

PR #1763, opened 2016-02-01
Add better description of sse-c keys

kyleknap:
Resolves ambiguity for whether the key should be base64 encoded


JordonPhillips:
:boat:

PR #1767, opened 2016-02-02
SAML forms-based authentication in CLI

rayluo:
Now customers can configure a profile to authenticate against a SAML Identity Provider (IdP). Only forms-based authentication is supported so far.
- You will need to configure your profile, which looks like the example in boto/botocore#783
- `saml_username` can be prompted by CLI, so it is optional in the configuration.
- ~~you don't have to specify the role_arn, CLI will prompt and ask for your input~~ You STILL NEED to specify the role_arn in config.

Relies on boto/botocore#783

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -3,6 +3,8 @@ import json
 import logging
 
 from botocore.exceptions import ProfileNotFound
+from botocore.compat import six
+from six.moves import input as raw_input
 
 LOG = logging.getLogger(__name__)
 
@@ -37,7 +39,40 @@ def inject_assume_role_provider_cache(session, **kwargs):
                   "JSONFileCache for assume-role.")
         return
     cred_chain.get_provider('assume-role').cache = JSONFileCache()
-    cred_chain.get_provider('assume-role-with-saml').cache = JSONFileCache()
+    saml_provider = cred_chain.get_provider('assume-role-with-saml')
+    saml_provider.cache = JSONFileCache()
+    saml_provider.role_selector = role_selector
+
+
+def role_selector(role_arn, roles):
+    if role_arn:
+        chosen_roles = [r for r in roles if r['RoleArn']==role_arn]
+        if not chosen_roles:
+            raise ValueError(
+                'Your specified role %s does not exist in IdP roles list: %s'
+                % (role_arn, [r['RoleArn'] for r in roles]))
+        role = chosen_roles[0]
+    else:
+        # So end user has not yet configured a role_arn.
+        # If IdP provides more than one role, ask the user to choose one,
+        # otherwise just proceed.
+        if len(roles) > 1:
+            options = [(str(i), role) for i, role in enumerate(roles, start=1)]
+            prompt = "\n".join(
+                "[%s]: %s" % (i, role['RoleArn']) for i, role in options)
+            choice = raw_input(
+                "Please choose the role you would like to assume:\n"
+                "(You can preconfigure the role_arn in your profile)\n"
+                "%s\n"
+                "Selection: " % prompt)
+            role = dict(options).get(choice)
+            if not role:
+                raise ValueError('You selected an invalid role index.')
+        elif roles:
+            role = roles[0]
+        else:
+            raise ValueError('There is no role to choose')
+    return role
 
 
 class JSONFileCache(object):
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -3,8 +3,6 @@ import json
 import logging
 
 from botocore.exceptions import ProfileNotFound
-from botocore.compat import six
-from six.moves import input as raw_input
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


awscli/customizations/assumerole.py
```diff
@@ -36,8 +38,41 @@ def inject_assume_role_provider_cache(session, **kwargs):
...
+        if not chosen_roles:
+            raise ValueError(
+                'Your specified role %s does not exist in IdP roles list: %s'
+                % (role_arn, [r['RoleArn'] for r in roles]))
+        role = chosen_roles[0]
+    else:
+        # So end user has not yet configured a role_arn.
+        # If IdP provides more than one role, ask the user to choose one,
+        # otherwise just proceed.
+        if len(roles) > 1:
```
jamesls: I don't think we should be prompting users to select a role here.  The CLI typically only prompts for things when required (i.e MFA tokens and soon the SAML passwords).  They should explicitly set which role they want to use in their config file.

I like the AWS powershell model for how they handle SAML.  Configuration is a separate concern.  If we want to update our `aws configure` command to prompt for SAML values, do an initial login and prompt you to pick which role to associate with the profile, that seems pretty reasonable.

awscli/customizations/assumerole.py
```diff
@@ -36,8 +38,41 @@ def inject_assume_role_provider_cache(session, **kwargs):
...
+            choice = raw_input(
+                "Please choose the role you would like to assume:\n"
+                "(You can preconfigure the role_arn in your profile)\n"
+                "%s\n"
+                "Selection: " % prompt)
+            role = dict(options).get(choice)
+            if not role:
+                raise ValueError('You selected an invalid role index.')
+        elif roles:
+            role = roles[0]
```
jamesls: Following up from the comment above, I also don't think we should silently pick the first role if only one role is available.  A user doesn't necessarily control which roles they can assume, and something as simple as adding a new role can potentially break them.

There's also no guarantee that the role you can assume could be swapped out by an admin so we'll now silently start assuming some different role the user may be unaware of.  I would expect in this case that we should error out.

I think explicitly requiring the role in the config + `aws configure` helper to make this process easy (by presenting them with a list of roles and asking them to select one) seems like the best way to address this, and cleanly separates concerns.

Does that seem reasonable?

rayluo: I have to admit that the current "choosing role on the fly" and "automatically choose the only available role" behavior was influenced by [Quint's post](http://blogs.aws.amazon.com/security/post/TxU0AVUS9J00FP/How-to-Implement-a-General-Solution-for-Federated-API-CLI-Access-Using-SAML-2-0). He did the same thing there.

If I were an end user in an organization, I would just use whatever role my administrator gave me, be it "dev" or "marketing" or something else, as long as it allows me to finish the job that I was supposed to accomplish by that role. So, switching role is not a concern for me (as an end user). I can be an "devops" today and administrator gives me a role to handle EC2 instances, I switched to marketing department tomorrow and my new role can access materials in the organization's S3 bucket.

So do we want to have this feature here?

jamesls: The AWS profiles have always maintained the property that they correspond to a single set of permissions (i.e one credential pair, one role).

I don't think it's unreasonable that if you want to use a different role, you have to rerun `aws configure`.  This is also what AWS powershell does (maps one profile to one role).  Seems like another reason to stay consistent.

If we separate concerns, the only downside I see is you have to run `aws configure ...` to reconfigure a role if it changes.  I think being explicit about this also serves as an acknowledgement to the user that they understand what role they're picking.

Also, let's say a user *wants* to explicitly set a role but doesn't know the full role ARN.  If we leave this change as is, the user has to make some arbitrary CLI call, look at the RoleArns from the printed list, then copy/paste that into their AWS config file.  If this is exposed via `aws configure` they can easily set that up and select the correct role.

jamesls:
Code change looks good, once we get the botocore change in and a clean run through travis.

One other thing I forgot, we should get the config-vars topic guide updated (http://docs.aws.amazon.com/cli/latest/topic/config-vars.html).  It currently has docs for assuming a role, we should update the docs to also document how to configure assume role with SAML.

kyleknap:
Yeah I am in the same vote as James here. Looks good but lets add some documentation in the help topic about it.

awscli/topics/config-vars.rst
```diff
@@ -225,6 +225,47 @@ Example configuration::
...
+
+The AWS CLI will automatically authenticate you against the SAML IdP, and then
+retrieve AWS credentials for you. You do not need to configure any credentials.
+
+You can specify the following configuration values for configuring an SAML role
+in the AWS CLI config file:
+
+* ``saml_endpoint`` - The URL of the login page of your SAML IdP
+* ``saml_authentication_type`` - The authentication type of your SAML IdP.
+  So far, the only valid setting is `form`, means forms-based authentication.
```
kyleknap: `form` --> ``form`` to make it a literal. Same thing for ``adfs`` and ``okta`` later in the guide.

awscli/topics/config-vars.rst
```diff
@@ -225,6 +225,47 @@ Example configuration::
...
+Markup Language (SAML) Identity Providers (IdP) which provides forms-based
+authentication.
+
+The AWS CLI will automatically authenticate you against the SAML IdP, and then
+retrieve AWS credentials for you. You do not need to configure any credentials.
+
+You can specify the following configuration values for configuring an SAML role
+in the AWS CLI config file:
+
+* ``saml_endpoint`` - The URL of the login page of your SAML IdP
```
kyleknap: Maybe make a point that it needs to be https

kyleknap:
Awesome. Had a couple small comments, but :ship:

lorengordon:
This looks pretty cool! Do you expect it will work to retrieve temporary credentials for AWS Directory Service users that are mapped to an IAM role?

atward:
Seems to be issues with this commit https://github.com/aws/aws-cli/pull/1767/commits/00158a27e1436a6f6c8d53cde9f77e24a88e7e24

I have same errors as reported by travis

```
$ aws sts get-caller-identity

Credential named assume-role-with-saml not found.
```

I would love to see this included in aws-cli, but don't know enough about the creds provider to assist.

jamesls:
Superceded via https://github.com/boto/botocore/pull/1316

PR #1769, opened 2016-02-03
Truncate file before writing role cache contents

### LINKED ISSUE #1684: JSON Cache for AssumeRoleProvider not truncating files
When we open a file for writing, if we're reusing the same file (same cache key) we don't truncate the file before writing.  If the second JSON response is smaller it will result in extra data at the end of the JSON document.

This will trigger a json parsing error, which raises a KeyError, which causes the cred provider to retrieve a new set of temporary credentials because it thinks the file is not in the cache.

jamesls:
Fixes #1684.


kyleknap:
:ship:

PR #1770, opened 2016-02-03
Add docs requirements file for generating docs

kyleknap:
Similar to what we do for boto3: https://github.com/boto/boto3/blob/develop/requirements-docs.txt.
More maintainable for scripting purposes.


requirements-docs.txt
```diff
@@ -0,0 +1,2 @@
+Sphinx==1.1.3
+-rrequirements.txt
```
jamesls: I know this is how boto3 does it, but do you know why we include requirements.txt?   For docs we should only need the `pip install -e .` and the `requirements-docs.txt` file.  Do we want to be pulling in dev versions of botocore when building docs?

kyleknap: It is so you can pull in the most recent botocore, which is needed for boto3 because the doc logic for botocore live in boto3. I would be fine with the ``-e`` option for the CLI because this logic is not really too coupled with botocore.

jamesls: Not sure I follow.  My issue was more that we're always building docs against the develop version of botocore instead of whatever the setup.py says you should use for botocore.  This means that `requirements-docs.txt` should *not* be used for anything other than the develop branch or the latest tag.  Otherwise, something like this would be wrong:

```
git checkout 1.9.0
pip install -r requirements-docs.txt
cd docs && make html
```

kyleknap: I was just explaining why we do it for boto3. It is because read the docs shows the develop branch of boto3 and so sometimes to build the docs off of the develop branch we may need the develop branch of botocore.

For the CLI we do not use the develop branch (latest release instead). So yes it makes more sense to do ``pip install -e .`` I will make the change

kyleknap:
Updated it to use ``-e .``

jamesls:
:shipit:  Looks good to me.

PR #1774, opened 2016-02-04
Fix test failures from mock 1.3.0 upgrade

jamesls:
This upgrades mock to version 1.3.0 which is what's used
in botocore and boto3.

This was causing subtle test falures depending on if you installed
CLI->botocore vs. botocore->CLI.

The latest version of mock, 1.3.0, is more strict in error validation.
Specifically, any mock method that starts with "assert" will now
raise AttributeErrors if they don't exist on the mock object.

Our code was using a couple of non-existent assertion methods that were
inadvertently passing (because they just returned a new mock object).


kyleknap:
:ship:

PR #1776, opened 2016-02-05
Remove instance id length check for opsworks customization

jamesls:
Needed to work with the longer Amazon EC2 instance ids.


kyleknap:
:ship:

PR #1780, opened 2016-02-09
Fix windows test for upload-build

kyleknap:
Also removed duplicate definition of tearDown

cc @jamesls @rayluo @JordonPhillips


rayluo:
:ship:

PR #1787, opened 2016-02-10
RST and XML files should not be executable

Fale:
as title


JordonPhillips:
I think we've had this issue before, it comes from doc writers using Windows git which may treat all files as executable by default. I wish the file mode was shown by default on PR's, but it is not. In the future, we should do the following:

* Look at the blame for PR's, which will show the filemode, and verify it is correct
* Advise doc writers to use `git update-index --chmod=-x <filename>` when this issue shows up

As for this CR specifically, that "`xml`" file is actually an `rst` file, which we will need to rename. With that in mind, could you collapse this into a single commit? Thanks!

Fale:
Hi,
I've proceded to squash the commits and to rename the rst file :)

JordonPhillips:
Thanks!

PR #1791, opened 2016-02-11
Now aws configure set foo http://www.example.com can store the url as-is

rayluo:
This is not immediately needed, but it will be useful when in future we add new setting which expects a URL.

Without this fix, setting URL will cause the content of that URL be written into the config file, which typically breaks the config file too. In the config file context, this is unlikely the desired behavior, so this is probably considered as a potential bug.


JordonPhillips:
Nice catch! :boat:

jamesls:
Please use the existing `tests/integration/customizations/test_configure.py` tests for this instead of creating new tests.

The integ tests should arguably moved to functional/, but for now let's not create another set of tests.

rayluo:
Ahh, I wish I found that set of integration tests earlier. It is now so easy to build a new test case on top of the current integration tests infrastructure. My previous attempt of building a functional test case from scratch was quite time-consuming.

Now I've switched to use integration test.

JordonPhillips:
Looks good! Definitely a much simpler change now.

kyleknap:
Looks good. :ship:

jamesls:
:shipit:

PR #1797, opened 2016-02-15
Break the configure model into several distinct files.

JordonPhillips:
Currently most of the configure code is in the `__init__.py` of the `configure` model, which is not so great. Since all the code for four commands is in one spot, it is difficult to scan and understand.

This PR does not change any functionality of the code whatsoever. It simply breaks each command into its own file in the configure module (and the same for the tests).


jamesls:
Out of curiosity, what are the perf differences?

kyleknap:
@jamesls you are asking for performance diff with respect to loading time of for any CLI command as there are few new modules now right? Small performance diffs with respect to running a ``configure`` command should not outweigh cleaning up code.

JordonPhillips:
@jamesls The performance difference is negligible in terms of the `configure` command performance. If your question is about overall performance, do you have a set of benchmarks you would like me to perform?

jamesls:
Ok, let me back up and add a little more context for my previous comment.

I'm personally a -0 to this change.  Having one class per subpackage is not common in our code base and seems like a very java-esque way to structure things (not that java is bad, it's just not python).  When I look at the current classes in this file:

```
class ConfigValue(object):
class SectionNotFoundError(Exception):
class InteractivePrompter(object):
class ConfigFileWriter(object):
class ConfigureListCommand(BasicCommand):
class ConfigureSetCommand(BasicCommand):
class ConfigureGetCommand(BasicCommand):
class ConfigureCommand(BasicCommand):
```

I see a handful of utility classes followed by 4 commands.  Seems straightforward to me.  I could maybe see breaking it into utils and commands subpackages.

That being said, if other's really find this useful I don't mind going with it.

The only thing that will make me go from -0 to -1 is if the perf differences are "big enough", which is why I initially asked.

We also have to consider that we load *every* customization right now when we run a CLI command, so this is slowing down every CLI call, not just configure.  This is also why I'm curious to see what the perf numbers are.  If it's negligible and you get another +1 then let's go with it.  Otherwise it's worth discussing.

Commands might include the time to run: an s3 command, an ec2 command, an iam command, etc.

rayluo:
1. It can be helpful to split a big enough source file into meaningful pieces. Just don't do it in a one-class-per-file way. We can use each new file as a natural container to group relevant sub-feature and its helper functions together. For example:

        # In __init__.py
        class SectionNotFoundError(Exception):
            # Technically only needed by ConfigFileWriter,
            # but it is an exception and we typically define exception at
            # package-wide level, so it is fine to leave it here in __init__.py
        class ConfigFileWriter(object):
            # Needed by both ConfigureCommand & ConfigureSetCommand.
            # It is fine to leave it in __init__.py or in a separated file

        # In list.py
        class ConfigValue(object):
            # Needed only by ConfigureListCommand, not yet moved to the list.py
        class ConfigureListCommand(BasicCommand):

        # In configure.py
        class InteractivePrompter(object):
            # Needed only by ConfigureCommand, already moved into configure.py
        class ConfigureCommand(BasicCommand):

        # In set.py
        class ConfigureSetCommand(BasicCommand):
            # Can be in a separated file since it does not fit in files above

        # In get.py
        class ConfigureGetCommand(BasicCommand):
            # Can be in a separated file since it does not fit in files above

   In that sense, I like the direction of this refactoring PR. Some minor tweaking may be needed, as suggested above. (BTW, I may also start similar refactoring in my previous CloudFront customization too.)


2. Performance test is generally an expensive work. So ideally we only bring that up when we have a reason to believe performance would be significantly affected. In general, we don't have to use performance test as a "tie breaker" to suggest a direction in a +0 or -0 situation. Doesn't +0 or -0 already mean "it is fine to go either way"? We can ask more team members to vote and therefore break the tie. (If, hypothetically, all team members are +0 or -0 for a proposal, we don't even need to use a :game_die: , we simply drop that proposal, and switch to more valuable items in our backlog.)

jamesls:
Yikes, all I wanted was quick confirmation about how much this impacts (or does not impact) performance, not to bikeshed over code layout.

I also disagree with you @rayluo that performance testing is expensive work.  I was asking for something very specific that should take about 10 minutes to produce.

At this point I'm half tempted to just do it myself, given the amount of time we've spent talking about this far exceeds the amount of time it would take to do some basic perf testing.

jamesls:
Had some time to look into this.  Here's what I had in mind.  Format is:

```
times,for,10,runs,in,seconds,..., -> mean, std-dev
```

```
# New feature branch
EC2 call:
1.67,1.11,1.15,1.12,1.25,1.13,1.11,1.09,1.15,1.24 -> 1.2, 0.17

IAM help:
0.76,0.69,0.68,0.70,0.71,0.71,0.69,0.69,0.70,0.69 -> 0.702, 0.02

IAM call:
1.34,0.94,0.95,0.97,0.89,0.92,0.94,0.93,0.93,0.92 -> 0.973, 0.131

configure get:
0.60,0.52,0.52,0.52,0.51,0.52,0.52,0.51,0.51,0.52 -> 0.525, 0.02


# Develop branch

EC2 call:
1.14,1.03,1.16,1.03,1.02,1.08,1.23,1.08,1.27,1.08 -> 1.11, 0.08

IAM help:
0.75,0.68,0.70,0.71,0.69,0.70,0.74,0.70,0.73,0.74 -> 0.714, 0.024

IAM call:
1.01,0.95,0.96,0.95,1.00,0.96,0.96,0.96,0.97,0.94 -> 0.966, 0.02

configure get:
0.53,0.53,0.54,0.52,0.52,0.51,0.52,0.52,0.52,0.51 -> 0.522, 0.009

```

So no, not a significant difference.

:shipit: from me.

awscli/customizations/configure/set.py
```diff
@@ -0,0 +1,104 @@
...
+    SYNOPSIS = 'aws configure set varname value [--profile profile-name]'
+    EXAMPLES = BasicCommand.FROM_FILE('configure', 'set', '_examples.rst')
+    ARG_TABLE = [
+        {'name': 'varname',
+         'help_text': 'The name of the config value to set.',
+         'action': 'store',
+         'cli_type_name': 'string', 'positional_arg': True},
+        {'name': 'value',
+         'help_text': 'The value to set.',
+         'action': 'store',
```
rayluo: You will probably have merge conflict. Please backport this bugfix https://github.com/aws/aws-cli/pull/1791

JordonPhillips: :+1:

rayluo:
:ship:

jamesls:
@JordonPhillips @rayluo I'm not seeing `test_set_with_a_url` from #1791.  Did that get merged in?

PR #1806, opened 2016-02-19
Handle empty sections when using configure set command

jamesls:
We were writing out a config value in the *next* section of the config file whenever there's an existing section that had no entries.

cc @kyleknap  @rayluo @JordonPhillips


jamesls:
Rebased against develop with the latest structural changes.  Should be good to look at now.

cc @kyleknap @rayluo @JordonPhillips

JordonPhillips:
:boat: LGTM

kyleknap:
:ship:

PR #1811, opened 2016-02-26
Fix formating issues for get-access-key-last-used documentation

tuxinaut:
- Remove trailing whitespace
- Replace tabs with spaces (fix indentation)


JordonPhillips:
I like it! Could you change to 4 spaces to match our other examples?

tuxinaut:
The indentation should be fine now. I think there was additionally a structure issue.

PR #1812, opened 2016-02-26
Avoid MD5

JordonPhillips:
Avoid MD5 if it isn't available.

### DIFF
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -6,13 +6,10 @@ from functools import partial
 import errno
 import hashlib
 
-from dateutil.parser import parse
-from dateutil.tz import tzlocal
-
-from botocore.compat import quote
-from awscli.customizations.s3.utils import find_bucket_key, \
-    uni_print, guess_content_type, MD5Error, bytes_print, set_file_utime, \
-    RequestParamsMapper
+from botocore.compat import MD5_AVAILABLE
+from awscli.customizations.s3.utils import (
+    find_bucket_key, guess_content_type, MD5Error, bytes_print, set_file_utime,
+    RequestParamsMapper)
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/fileinfo.py
```diff
@@ -48,22 +45,26 @@ def save_file(filename, response_data, last_update, is_stream=False):
...
         with open(filename, 'wb') as out_file:
-            write_to_file(out_file, etag, md5, file_chunks)
+            write_to_file(out_file, etag, file_chunks, md5)
 
-    if _can_validate_md5_with_etag(etag, response_data):
-        if etag != md5.hexdigest():
-            if not is_stream:
-                os.remove(filename)
-            raise MD5Error(filename)
+    if md5 and etag != md5.hexdigest():
```
jamesls: `md5 is not None`

awscli/customizations/s3/fileinfo.py
```diff
@@ -92,7 +93,7 @@ def write_to_file(out_file, etag, md5, file_chunks, is_stream=False):
     """
     body = b''
     for chunk in file_chunks:
-        if not _is_multipart_etag(etag):
+        if md5 and not _is_multipart_etag(etag):
```
jamesls: `md5 is not None`

tests/unit/customizations/s3/test_fileinfo.py
```diff
@@ -26,13 +25,38 @@
 from awscli.customizations.s3.fileinfo import TaskInfo
 
 
-class TestSaveFile(unittest.TestCase):
+class BaseMD5Test(unittest.TestCase):
```
jamesls: We typically only create a shared base class if it's going to be shared my multiple tests, but I only see `TestSaveFile` subclassing from it.  Was there a reason to create this base class?

awscli/customizations/s3/fileinfo.py
JordonPhillips: `None` is falsy, so is that necessary?

tests/unit/customizations/s3/test_fileinfo.py
```diff
@@ -95,6 +120,15 @@ def test_no_raise_md5_with_kms(self):
         # The file should have been saved.
         self.assertTrue(os.path.isfile(self.filename))
 
+    def test_no_raise_md5_when_md5_unavailable(self):
+        self.response_data['ETag'] = '"0"'
+        self.response_data['ServerSideEncryption'] = 'AES256'
```
jamesls: What's the significance of having this value in the response data?

jamesls:
Looks good, just had some small comments.

awscli/customizations/s3/fileinfo.py
jamesls: Yes.  Semantically, if you have a value of `None`, you need to differentiate between "provided a false like value" and "did not provide a value".

tests/unit/customizations/s3/test_fileinfo.py
JordonPhillips: To simplify `TestSaveFile` and allow for future tests to use MD5.

tests/unit/customizations/s3/test_fileinfo.py
JordonPhillips: It will cause `_can_validate_md5_with_etag` to pass, so that the only thing preventing MD5 is the `MD5_AVAILABLE` flag. That way if the test fails, it is because that flag did not get set or checked properly.

tests/unit/customizations/s3/test_fileinfo.py
```diff
@@ -30,18 +29,38 @@ class TestSaveFile(unittest.TestCase):
...
         self.response_data = {
             'Body': six.BytesIO(b'foobar'),
             'ETag': '"%s"' % etag,
         }
         self.last_update = datetime.now()
 
+        # Setup MD5 patches
+        self.md5_object = mock.Mock()
+        self.md5_object.hexdigest = mock.Mock(return_value=etag)
+        md5_builder = mock.Mock(return_value=self.md5_object)
```
jamesls: Unless it's more convenient to store these as instance attributes you can also just say:

```
md5_builder = mock.Mock()
md5_builder.return_value.hexdigest.return_value = etag
```

That makes it more clear that you're only storing mocks that you plan on asserting things on later.

jamesls:
:shipit:

PR #1813, opened 2016-02-27
add ElastiCache exmaples

quiver:
Added examples are:
* create-replication-group
* modify-cache-parameter-group
* modify-replication-group


kyleknap:
Thanks! Merging.

PR #1814, opened 2016-02-29
Add docs about when we calculate checksums

### LINKED ISSUE #1750: Does 'cp' perform data integrity checks and return non-zero exit code if md5 is incorrect?
Can't find this in the documentation.

jamesls:
Closes #1750 

cc @kyleknap @JordonPhillips


jamesls:
Oh interesting, this appears to be because of a new version of dateutil.  I'll dig in and see what we can do.

JordonPhillips:
:boat: When the dateutil thing gets solved

kyleknap:
Content looks good but question. Instead of adding this to the concepts file that is pretty much a getting started guide to the s3 commands, how about creating a new topic for this in like an ``S3 Command FAQ`` topic or something to avoid setting a trend where the getting started part does not get too cluttered with information that most users may not need or want to know about.

jamesls:
I agree with the idea of a FAQ, but I'm not so sure that this specific content falls under "things a user does not need or want to know about."  Checksum validation is probably one of the things we get asked frequently and something that seems like it should be documented on the main page rather than an FAQ.

That said, I don't feel that strongly about it.  What's your criteria for what falls in a FAQ vs. _concepts?

kyleknap:
The way I see it the _concepts.rst is mainly a getting started guide. Everything in this section is for getting your feet wet with using the s3 commands. So stuff that is about basic usage and is universal to the command set should be included here like (filter options, how path ordering affects behavior, recursive options).

The FAQ would be pretty much anything more technical then basic command usage and gets asked a lot. So I would see MD5 checksum/retries would go here. I would also see questions like "How do we determine if something needs to be synced?" and "Why do we not do a complete s3 to s3 copy?" would go here as well because it is a pretty meaty explanation and is a very common question.

jamesls:
I don't really see things like the "s3 help" page as a getting started guide (I'd say [this](http://docs.aws.amazon.com/cli/latest/userguide/using-s3-commands.html) is our getting started guide for S3).  I can see putting useful getting started stuff first, but I compare these pages more closely to reference/manual pages.  Compare, for example, `man rsync` or `git merge --help`, etc.  I think it makes more sense to follow this standard model of CLI man pages.

I see topic guides as more cross cutting concerns.  For example, `config-vars` doesn't really make sense in docs for a single command.  Same thing with `return-codes`.  An `S3 debugging/troubleshooting guide` also seems reasonable as a topic guide.

kyleknap:
Yeah I would like putting this in an ``S3 debugging/troubleshooting guide`` instead because isn't this content that you are adding applies to the s3api commands as well because the handler for content md5 on uploads is applied in botocore? You could pretty much take that whole upload section about md5 checking and make it refer to s3api and the content would still be valid.

jamesls:
Ok, so what should the topic name be? `aws help ____?`

kyleknap:
I am thinking ``s3-info``, ``s3-general``,  ``s3-debug``, or ``s3-troubleshoot``? I was also considering just ``s3``. Do you think calling it ``s3`` would be too confusing since there is already a ``aws s3 help`` command?

jamesls:
@kyleknap Moved this into an S3 FAQ topic guide.

kyleknap:
Thanks! :ship:

PR #1815, opened 2016-02-29
Refactor CLI completer

JordonPhillips:
The current completer relies on a heavy amount of shared state, which makes it difficult to effectively understand and test. This PR refactors the completer so that the functions accept parameters rather than relying on state. The functionality and logic is unchanged.

cc @kyleknap @jamesls

### DIFF
--- a/awscli/completer.py
+++ b/awscli/completer.py
@@ -19,19 +19,49 @@ LOG = logging.getLogger(__name__)
 
 class Completer(object):
 
-    def __init__(self):
-        self.driver = awscli.clidriver.create_clidriver()
-        self.main_hc = self.driver.create_help_command()
-        self.main_options = self._documented(self.main_hc.arg_table)
-        self.cmdline = None
-        self.point = None
-        self.command_hc = None
-        self.subcommand_hc = None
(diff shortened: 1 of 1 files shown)


jamesls:
You should be good to rebase against develop now.

awscli/completer.py
```diff
@@ -19,19 +19,49 @@
...
-        self.cmdline = None
-        self.point = None
-        self.command_hc = None
-        self.subcommand_hc = None
-        self.command_name = None
-        self.subcommand_name = None
-        self.current_word = None
-        self.previous_word = None
-        self.non_options = None
+    def __init__(self, driver=awscli.clidriver.create_clidriver()):
```
jamesls: You generally don't want to have default mutable arguments.  The common pattern is to default this to `None`, then check if `driver is None` and create a default value in that scenario.  Otherwise multiple instances may unknowingly share the same argument value and mutations in one instance will be reflected in instances of the other `Completer`.

jamesls:
Changes look good, just had a small comment.

jamesls:
:shipit:

PR #1816, opened 2016-02-29
Use realistic s3 key names in cloudtrail validation tests

jamesls:
This fixes the latest test failures from the dateutil 2.5 release.

Several cloudtrail unit/functional tests used
S3 key names such as ``abc`` instead of an S3 key name that
contained a date that we extract.  When this happened, we aren't
able to extract a timestamp from the key ``abc``, and as a result
we pass the empty string to dateutil's date parser.  In versions
prior to 2.5, dateutil would return the current date.  Now it
raises a ``ValueError``.

The fix is to update these cases to use a key name that contains
a date that can be parsed.

cc @kyleknap @JordonPhillips


JordonPhillips:
:boat:

kyleknap:
:ship:

PR #1821, opened 2016-03-01
Update tests after region and endpoint change

kyleknap:
cc @jamesls @JordonPhillips 

I was considering to remove the patching of the endpoint and make it rely on the client, but that would be a little bit out of the scope as I am just fixing the build. Let me know if you want me to switch the endpoint patching to be client patching.


jamesls:
:shipit:

PR #1822, opened 2016-03-01
Mark args as not required when aliasing

### LINKED ISSUE #1790: storagegateway commands having to do with volumes fail
It can't decide whether it wants `--volume-arns` or `--volume-ar-ns`
```
% aws --version                                                                                                                                                                                           
aws-cli/1.10.1 Python/2.7.10 Darwin/15.2.0 botocore/1.3.23
% aws storagegateway describe-stored-iscsi-volumes --volume-arns 'arn:aws:storagegateway:us-west-2:xxxxxxxxxx:gateway/sgw-xxxxxxxxxx/volume/vol-xxxxxxx'

  aws help
  aws <command> help
aws: error: argument --volume-ar-ns is required
% aws storagegateway describe-stored-iscsi-volumes --volume-ar-ns 'arn:aws:storagegateway:us-west-2:xxxxxxxxxx:gateway/sgw-xxxxxxx/volume/vol-xxxxxxx'                                                

  aws help
  aws <command> help
aws: error: argument --volume-arns is required
```

jamesls:
If you have a required argument that you'd like to rename,
we were previously copying the argument and injecting
the copy into the arg table under a different name.

The problem is that the ``required`` field is also copied.
This will require the user to provide both arguments, which
defeats the point of aliasing the argument we want to hide.

The change here is to not make aliased args required.  That
way the user can specify either version.

Fixes #1790

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -127,7 +127,8 @@ class CLIDocumentEventHandler(object):
             self._documented_arg_groups.append(argument.group_name)
         else:
             option_str = '%s <value>' % argument.cli_name
-        if not argument.required:
+        if not (argument.required
+                or getattr(argument, '_DOCUMENT_AS_REQUIRED', False)):
             option_str = '[%s]' % option_str
         doc.writeln('%s' % option_str)
 
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -42,8 +42,14 @@ def make_hidden_alias(argument_table, existing_name, alias_name):
     copy_arg = copy.copy(current)
     copy_arg._UNDOCUMENTED = True
     copy_arg.name = alias_name
-    copy_arg.required = False
-    current.required = False
+    if current.required:
+        # If the current argument is required, then
+        # we'll mark both as not required, but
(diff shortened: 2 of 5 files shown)


jamesls:
Rebased after the text fix, travis builds are good now.  Ready to look at again.

cc @kyleknap @JordonPhillips

kyleknap:
So botocore should catch that its required if users do not include it, but what about the documentation? I feel that we will get an issue where someone complains that in the docs it says it is not required but when they run the command and do not include the required argument a validation error is thrown. Is there a way we can make the documentation say it is required but not make it required through argparse?

jamesls:
I'm not sure changing our docs is the right approach here.  The behavior and docs should be consistent.  If we are going to mark something as required in our CLI docs, then it should have the same behavior that other required params have when they're not provided.

Let me see what our other options are.

FWIW there have been a few cases where a param has been required at the API level, but not marked as required in our service models.  I don't think we've ever had any complaints.

jamesls:
Ok, I think this should be ready for another look.

kyleknap:
Thanks! :ship:

PR #1826, opened 2016-03-02
Fix ordering issue with tests

jamesls:
Fixes an regression introduced in
16a59ac0b55a40aa81428844d48659ea76ea91f4.

The loaders are not being configured with the data path needed to
load the CLI data files.  This means that if you try to run any
test that used the base class where this loader was created,
you'll get DataNotFound errors.

However, this is a shared object to speed up tests.
As long as something comes along and configures the loader
properly (another base class does this properly), then the
other tests will pass as expected.

The fix here is to use the create_clidriver() function which
abstracts this logic for you.

Now you can run individual unit/functional tests without
getting these errors.

Verified that the unit/functional test run times are about the same.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -211,13 +211,8 @@ class BaseCLIDriverTest(unittest.TestCase):
         }
         self.environ_patch = mock.patch('os.environ', self.environ)
         self.environ_patch.start()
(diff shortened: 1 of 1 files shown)


kyleknap:
:ship:

PR #1827, opened 2016-03-03
Don't delete a bucket if we can't remove all the objects

jamesls:
We weren't checking the RC of the RmCommand.

cc @kyleknap @JordonPhillips


kyleknap:
It would be cool if we can make it easier to integrate the stubber into CLI functional test, but that change needs to happen in botocore to allow stubbers to be activated on a session. Otherwise, looks good. :ship:

PR #1828, opened 2016-03-03
Add option to double check if glacier objects are restored

JordonPhillips:
This adds the option to force glacier downloads by performing a HEAD on each glacier object in the list objects call.

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -64,7 +64,8 @@ class S3Handler(object):
             'content_language': None, 'expires': None, 'grants': None,
             'only_show_errors': False, 'is_stream': False,
             'paths_type': None, 'expected_size': None, 'metadata': None,
-            'metadata_directive': None, 'ignore_glacier_warnings': False
+            'metadata_directive': None, 'ignore_glacier_warnings': False,
+            'force_glacier_transfer': False
         }
         self.params['region'] = params['region']
         for key in self.params.keys():
@@ -186,7 +187,8 @@ class S3Handler(object):
                                          message=warning_message)
                 self.result_queue.put(warning)
             # Warn and skip over glacier incompatible tasks.
-            elif not filename.is_glacier_compatible():
+            elif not self.params.get('force_glacier_transfer') and \
+                    not filename.is_glacier_compatible():
                 LOGGER.debug(
                     'Encountered glacier object s3://%s. Not performing '
                     '%s on object.' % (filename.src, filename.operation_name))
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -392,6 +392,15 @@ IGNORE_GLACIER_WARNINGS = {
 }
 
 
+FORCE_GLACIER_TRANSFER = {
+    'name': 'force-glacier-transfer', 'action': 'store_true',
+    'help_text': (
+        'Forces a transfer request on all Glacier objects in a sync or '
+        'recursive copy.'
+    )
+}
+
+
 TRANSFER_ARGS = [DRYRUN, QUIET, INCLUDE, EXCLUDE, ACL,
                  FOLLOW_SYMLINKS, NO_FOLLOW_SYMLINKS, NO_GUESS_MIME_TYPE,
                  SSE, SSE_C, SSE_C_KEY, SSE_KMS_KEY_ID, SSE_C_COPY_SOURCE,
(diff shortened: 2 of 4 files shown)


JordonPhillips:
COMPLETER TEEESSSTTTSSS

jamesls:
Why not just add a flag that ignores the "is glacier check" (i.e add a flag that puts back the old behavior)?

We've purposefully not done HeadObject requests for each object historically, and I don't think this is something we should be doing now.

kyleknap:
I agree with James. We do not want to start heading objects. Just add back the old behavior.

JordonPhillips:
@jamesls @kyleknap So do you mean try to download it regardless of whether it is restored or not?

kyleknap:
Yes download no matter what.

JordonPhillips:
:+1:

JordonPhillips:
The tests failing are the completer tests. Rather than forcing them to work again, I'm going to wait until #1830 is merged.

JordonPhillips:
Rebased on top of #1830 so completer tests should no longer be failing.

awscli/customizations/s3/s3handler.py
```diff
@@ -186,7 +187,8 @@ def _enqueue_tasks(self, files):
                                          message=warning_message)
                 self.result_queue.put(warning)
             # Warn and skip over glacier incompatible tasks.
-            elif not filename.is_glacier_compatible():
+            elif not filename.is_glacier_compatible(
```
kyleknap: Why couldn't we just make the logic:
```py
elif not self.params.get('force_glacier_transfer') and not filename.is_glacier_compatible():
```
instead of adding it as a parameter to the ``is_glacier_compatible()`` method.

awscli/customizations/s3/subcommands.py
```diff
@@ -392,6 +392,15 @@
 }
 
 
+FORCE_GLACIER_TRANSFER = {
```
kyleknap: Do we want to name it ``--force-glacier-transfer`` or ``--force-glacier``? I think I would prefer ``--force-glacier`` because it is shorter. Also in your functional test, you use ``--force-glacier``. What do you think?

kyleknap:
Looks good. Just had a couple of comments.

awscli/customizations/s3/s3handler.py
JordonPhillips: Good point. Passing it through made sense when I was forcing a HEAD request, but condensing it makes more sense now.

awscli/customizations/s3/subcommands.py
JordonPhillips: I'm not crazy about `--force-glacier` because it isn't clear what you're forcing. `--force-glacier-transer` is slightly clearer. Even then, maybe something like `--force-glacier-attempt` would be better because it is clear that you are forcing the CLI to make an attempt with glacier.

kyleknap: That's fine then.

tests/functional/s3/test_cp_command.py
```diff
@@ -235,6 +235,45 @@ def test_cp_fails_with_utime_errors_but_continues(self):
...
+                    {'Key': 'foo/bar.txt', 'ContentLength': '100',
+                     'LastModified': '00:00:00Z',
+                     'StorageClass': 'GLACIER',
+                     'Size': 100},
+                ],
+                'CommonPrefixes': []
+            },
+            {'ETag': '"foo-1"', 'Body': six.BytesIO(b'foo')},
+        ]
+        cmdline = '%s s3://bucket/foo %s --recursive --force-glacier' % (
```
kyleknap: Should be ``--force-glacier-transfer``. ``--force-glacier`` only works because argparse is weird in that it can autocomplete arguments that were provided that we shortened.

tests/functional/s3/test_sync_command.py
```diff
@@ -78,6 +78,26 @@ def test_sync_to_non_existant_directory(self):
...
+                    {'Key': 'foo/bar.txt', 'ContentLength': '100',
+                     'LastModified': '00:00:00Z',
+                     'StorageClass': 'GLACIER',
+                     'Size': 100},
+                ],
+                'CommonPrefixes': []
+            },
+            {'ETag': '"foo-1"', 'Body': six.BytesIO(b'foo')},
+        ]
+        cmdline = '%s s3://bucket/foo %s --force-glacier' % (
```
kyleknap: Same thing here: should be ``--force-glacier-transfer``.

kyleknap:
Just had a small comment, but :ship: otherwise.

PR #1829, opened 2016-03-03
Test completer with subset to allow growing options

JordonPhillips:
This doesn't change the fact that the tests are not great, but at least gives them the flexibility to not break when more services / operations are added (only when they're removed).

cc @jamesls @kyleknap


jamesls:
I"m not really a fan of this.  This can hide the fact that completer could have bugs.

I'd rather just replace this with tests that don't depend on the actual existing services/args.

PR #1830, opened 2016-03-03
Refactor Completer tests to use fake data

JordonPhillips:
In addition, I found and fixed a bug where used arguments were not being filtered if there was no service name or partial service name.

cc @jamesls @kyleknap

### DIFF
--- a/awscli/completer.py
+++ b/awscli/completer.py
@@ -43,7 +43,7 @@ class Completer(object):
         if cmd_name is None:
             # If we didn't find any command names in the cmdline
             # lets try to complete provider options
-            return self._complete_provider(current_arg)
+            return self._complete_provider(current_arg, opts)
         elif subcmd_name is None:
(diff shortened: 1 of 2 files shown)


kyleknap:
Tests look much better. I was debating if I would rather see the existing driver used with patches or the one of the fake cli drivers that we already have, but that would make the tests a lot more verbose because you would have to be supplying service JSON models and hooking into the event system to provide your own custom commands. The only downside with your approach is if we change the internals of the driver or any of the constructors for the command objects, the tests will fail. In the end, I think I like the approach you went with more. :ship:

jamesls:
Much nicer.  Have you confirmed the code coverage numbers are the same/better than the previous tests?

JordonPhillips:
@jamesls Code coverage is the same (I wrote these by converting each item into at least one assertion). Line coverage is down by one (we don't use the default clidriver in the tests).

jamesls:
:shipit:  Perfect, thanks for checking.

Issue #1771, opened 2016-02-03
aws s3 cp errors when "-" stream to stdout option is used with a prefix


davidwarshaw:
aws-cli/1.10.1 Python/2.7.6 Darwin/14.3.0 botocore/1.3.23

When used with a prefix, "-" stream to stdout option errors with a 404, with or without `--recursive` flag.

```
> aws s3 ls s3://my-bucket/my-prefix --recursive
my-prefix/_SUCCESS
my-prefix/part-00000
my-prefix/part-00001
...
```

```
> aws s3 cp s3://my-bucket/my-prefix/part-00000 -
My data row 1
My data row 2
My data row 3
...
```

```
> aws s3 cp s3://my-bucket/my-prefix -
A client error (404) occurred when calling the HeadObject operation: Not Found
```

```
> aws s3 cp s3://my-bucket/my-prefix - --recursive
A client error (404) occurred when calling the HeadObject operation: Not Found
```

If this use of the option is not allowed, perhaps the documentation should be updated to reflect.

jamesls:
The first case: 
```
> aws s3 cp s3://my-bucket/my-prefix -
A client error (404) occurred when calling the HeadObject operation: Not Found
```
doesn't seem like the CLI can really do much with that.  We interpret that as you wanting to download the `s3://mybucket/my-prefix` key to stdout.  We have no way of knowing that you meant a prefix.  It's the same situation if you have `aws s3 cp s3://bucket/somekey /tmp/myfile`.  We assume you want to download the `somekey` file to `/tmp/myfile`.

The second case with the `--recursive` option on the other hand, seems like something we should be able to validate.  If you want to stream to stdout, then `--recursive` should not be allowed and we should error out.

davidwarshaw:
I agree with the reasoning to leave behavior as is for the case without `--recursive`.

When `--recursive` is used, I would have expected the behavior to be idiomatic with the other options and commands. Traverse the prefix in whatever the normal order is, streaming each key to stdout.

Barring this, I agree with erroring out at argument parsing with a good message, and adding a line to the docs for each option, specifying that they can't be used together.

PR #1839, opened 2016-03-07
Improve error message for streaming

kyleknap:
Raise an error if the user tries to use --recursive

Fixes https://github.com/aws/aws-cli/issues/1771

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1048,12 +1048,14 @@ class CommandParameters(object):
     def _validate_streaming_paths(self):
         self.parameters['is_stream'] = False
         if self.parameters['src'] == '-' or self.parameters['dest'] == '-':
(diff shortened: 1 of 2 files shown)


jamesls:
:shipit:

davidwarshaw:
Are the docs in the repo? It would be great if they were updated as well.

kyleknap:
Depends, what part of the docs were you looking to get updated?

davidwarshaw:
I suggest a note in the entry for "Downloading a S3 object as a local file stream", indicating the recursive flag is not valid in combo:
http://docs.aws.amazon.com/cli/latest/reference/s3/cp.html

davidwarshaw:
While we're being thorough, it should be "Downloading **an** S3 object as a local file stream". :)

davidwarshaw:
PR for doc update here: aws/aws-cli/pull/1840

PR #1844, opened 2016-03-11
[Fix] `s3 ls` command fails when UTF-8 is piped

malthejorgensen:
Piping the output of the `aws s3 ls` command when used on a bucket that has keys with UTF-8 characters in it, will raise an exception, exit the program and not list all files.

E.g. `aws s3 ls s3://<BUCKET-WITH-UTF8> | wc -l` will print the error `encode() argument 1 must be string, not None` and exit, upon reaching the UTF-8 key.

It fails because when stdout is being piped `sys.stdout.encoding` is `None`.

I am aware that the `PYTHONIOENCODING` environment variable can be set in order to change that, but it seems that the current code is trying to default to `ascii` and simply failing to default: [awscli/customizations/s3/utils.py, line 398](https://github.com/aws/aws-cli/blob/22a8404d8d064073034ffcb1641e0d320a74db53/awscli/customizations/s3/utils.py#L398) (latest commit on develop)

### DIFF
--- a/tests/unit/customizations/s3/test_utils.py
+++ b/tests/unit/customizations/s3/test_utils.py
@@ -478,6 +478,29 @@ class TestRequestParamsMapperSSE(unittest.TestCase):
              'SSECustomerKey': 'my-sse-c-key'})
 
 
+class MockPipedStdout(io.BytesIO):
+    '''Mocks `sys.stdout`.
+    We can't use `TextIOWrapper` because calling
+    `TextIOWrapper(.., encoding=None)` sets the ``encoding`` attribute to
+    `UTF-8`.
+    The attribute is also `readonly` in `TextIOWrapper` and `TextIOBase` so it
+    cannot be overwritten in subclasses. For these reasons we mock `sys.stdout`.
+    '''
+    def __init__(self):
+        self.encoding = None
+
+        super(MockPipedStdout, self).__init__()
+
+    def write(self, str):
+        # sys.stdout.write() will default to encoding to ascii, when its
+        # `encoding` is `None`.
+        if self.encoding is None:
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/s3/test_utils.py
```diff
@@ -486,6 +486,16 @@ def test_out_file_with_encoding_attribute(self):
...
 
+    def test_encoding_with_encoding_none(self):
+        '''When the output of the aws command is being piped,
+        the ``encoding`` attribute of ``sys.stdout`` is ``None``.'''
+        buf = io.BytesIO()
+        out = io.TextIOWrapper(buf, encoding=None)
+        uni_print(u'SomeChars\u2713\u2714OtherChars', out)
+        # Unicode characters get replaced with their
+        # UTF-8 byte value.
+        self.assertEqual(buf.getvalue(), b'SomeChars\xe2\x9c\x93\xe2\x9c\x94OtherChars')
```
jamesls: I'm probably missing something, but how does this get utf-8 encoded if your code change above sets the default to ascii?

malthejorgensen: It was just an observation – I actually don't know :S – maybe it's a Python default when encoding unicode strings to ascii.

malthejorgensen: Okay. I found out (I think). `TextIOWrapper` sets `encoding` to `UTF-8` when is passed `encoding=None`, so this test never hits the `UnicodeEncodeError` branch of `uni_print.

I'm gonna update this branch to fix it.

jamesls:
Thanks for updating.  Looks good to me.

PR #1846, opened 2016-03-11
Add api_versions support

kyleknap:
cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -363,8 +363,10 @@ class ServiceCommand(CLICommand):
 
     def _get_service_model(self):
         if self._service_model is None:
+            api_version = self.session.get_config_variable('api_versions').get(
+                self._service_name, None)
             self._service_model = self.session.get_service_model(
-                self._service_name)
+                self._service_name, api_version=api_version)
         return self._service_model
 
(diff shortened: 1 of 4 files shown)


awscli/clidriver.py
```diff
@@ -363,8 +363,10 @@ def _get_command_table(self):
 
     def _get_service_model(self):
         if self._service_model is None:
+            api_version = self.session.get_config_variable('api_versions').get(
```
JordonPhillips: The output of `get_config_variable` is `None` if they don't have versions set, so you'll need to check for that.

JordonPhillips:
Looks good, just one small comment.

awscli/clidriver.py
kyleknap: The default is an empty dictionary: https://github.com/boto/botocore/blob/develop/botocore/session.py#L83. So we can assume that value will always be a dictionary unless they messed up the way in which they specified the value in the config file like: ``api_versions = foo`` which will cause it to be a string.

JordonPhillips: Ah, I haven't pulled in that change yet. With that in mind, looks good.

JordonPhillips:
:boat:

PR #1847, opened 2016-03-11
Change groff command

### LINKED ISSUE #1798: groff 1.22.3 on smartos won't parse help text
Due to the way the latest groff 1.22.3 is installed on smartos (through pkgsrc), the traditional

`groff -man` will not parse man format pages correctly. I think this may be related to groff 1.22.3 in general, not the particular installation on smartos (illumos), but I do not have latest groff on either my mac or linux machines to test. In any case, as a result, there is empty help text when using `aws help`.

The fix, which should not interfere with other platforms, is to run `groff -m man`; you can change the line setting up the groff command in `help.py` to:

`cmdline = ['groff', '-m', 'man', '-T', 'ascii']`

In case someone has groff 1.22.3 installed on another platform it would be worth testing to see if the help pages are rendered correctly.

JordonPhillips:
Some systems have issues which cause `-man` to fail, while `-m man` still works, even though the first is an alias of the second.

Since this isn't really possible to write a test for, I ran both versions on the following systems to verify that no functionality is lost:

* OSX, with groff from pkgsrc, homebrew, and the default install
* FreeBSD
* Amazon Linux
* RHEL
* Ubuntu

Closes #1798 

cc @jamesls @kyleknap

### DIFF
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -106,7 +106,7 @@ class PosixHelpRenderer(PagingHelpRenderer):
         man_contents = publish_string(contents, writer=manpage.Writer())
         if not self._exists_on_path('groff'):
             raise ExecutableNotFoundError('groff')
-        cmdline = ['groff', '-man', '-T', 'ascii']
+        cmdline = ['groff', '-m', 'man', '-T', 'ascii']
         LOG.debug("Running command: %s", cmdline)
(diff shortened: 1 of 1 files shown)


kyleknap:
Thanks for doing the research! :ship:

PR #1859, opened 2016-03-18
Remove coveralls from README and CONTRIBUTING

JordonPhillips:
We no longer use coveralls, so it should no longer be mentioned in our docs.

cc @kyleknap @jamesls


kyleknap:
:ship:

PR #1864, opened 2016-03-21
Fix message generation on 5TB size limit

### LINKED ISSUE #1842: Exception thrown if uploaded file size > 5TB
We recently hit an exception when uploading a large file that by the looks of it should instead be reported to the user as a helpful message. The exception is being raised attempting to create that helpful message.

Version: aws-cli/1.10.11 Python/2.7.9 Windows/2012Server botocore/1.4.2

on: create_warning() got an unexpected keyword argument 'message' 
File "awscli\customizations\s3\s3handler.pyc", line 99, in call 
File "awscli\customizations\s3\s3handler.pyc", line 186, in _enqueue_tasks 
TypeError: create_warning() got an unexpected keyword argument 'message'

Looks like it is happening here:

https://github.com/aws/aws-cli/blob/c5fb60e1b49667d6f689bb8f8670ed4a1c01791a/awscli/customizations/s3/s3handler.py#L185

I think message= should be error_message= (although I don't know Python, so could be reading it wrong.

JordonPhillips:
Fixes #1842 

cc @jamesls @kyleknap

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -184,7 +184,7 @@ class S3Handler(object):
             if too_large and filename.operation_name == 'upload':
                 warning_message = "File exceeds s3 upload limit of 5 TB."
                 warning = create_warning(relative_path(filename.src),
-                                         message=warning_message)
+                                         warning_message)
                 self.result_queue.put(warning)
             # Warn and skip over glacier incompatible tasks.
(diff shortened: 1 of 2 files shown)


jamesls:
I know it's small but let's add a test for this.  It's likely why this went uncaught in the first place.

JordonPhillips:
Updated

kyleknap:
:ship:

Issue #1865, opened 2016-03-21
CreateDefaultRoles script is broken


andrewgross:
Hey,

I was trying to follow the tutorial [here](https://github.com/awslabs/data-pipeline-samples/tree/master/samples/RDStoRedshiftSqoop).  To setup permissions, I need to run `aws datapipeline create-default-roles`.  This call fails.  When you use `--debug`, you can see the following output:

```
2016-03-21 18:11:20,980 - MainThread - botocore.endpoint - DEBUG - Making request for <botocore.model.OperationModel object at 0x10acb9950> (verify_ssl=True) with params: {'body': {'Action': u'GetRole', 'RoleName': 'DataPipelineDefaultRole', 'Version': u'2010-05-08'}, 'url': u'http://iam.amazonaws.com/', 'headers': {'User-Agent': 'aws-cli/1.10.14 Python/2.7.6 Darwin/15.3.0 botocore/1.4.5'}, 'query_string': '', 'url_path': '/', 'method': u'POST'}
``` 

Note that the **`url`** key is not using SSL.  Naturally, the IAM service does not accept non-SSL requests, and our command fails.

I investigated this issue and discovered that the `CreateDefaultRoles` script [uses an old method signature when generating its IAM client](https://github.com/aws/aws-cli/blob/6f65ad7b19be0b3ac183f9799177d184f187de05/awscli/customizations/datapipeline/createdefaultroles.py#L53-L55).  It uses positional args instead proper keyword args.  This pushes an argument of `None` onto `use_ssl`, which will come out as boolean `False`.

I will submit a PR to fix the method signature.

kyleknap:
Merged PR. Closing issue.

PR #1866, opened 2016-03-21
Fix IAM Create Client Method Signature

andrewgross:
This code seems to have been written before additional `kwargs` were added to `create_client`.  A mispositioned argument means that it actually passes `use_ssl=None`, instead of an empty `endpoint_url`.  Switching to explicit keyword arguments fixes the problem.

See issues #1865


kyleknap:
Thanks! Looks good. Merging.

Issue #1835, opened 2016-03-07
Incorrect documentation for downloading RDS logs


cbandy:
Since #1617, the documentation indicates that the following command will download an entire RDS log file. However, it errors without downloading anything.

```
$ aws --version
aws-cli/1.10.8 Python/2.7.10 Darwin/15.3.0 botocore/1.3.28

$ aws rds download-db-log-file-portion --db-instance-identifier myinstance \
--starting-token 0 --log-file-name error/postgresql.log.2016-02-29-00 \
--output text

Bad starting token: 0
```

rayluo:
I haven't tried to reproduce this yet. But it sounds like relevant to our recent change to CLI pagination token. Perhaps @JordonPhillips can take a look.

cbandy:
I just found a workaround:

```
$ aws rds download-db-log-file-portion --db-instance-identifier myinstance \
--cli-input-json '{ "Marker": "0" }' --log-file-name error/postgresql.log.2016-02-29-00 \
--output text
```

kyleknap:
Yeah this is a regression due to pagination change. We are looking into how to handle these cases.

gorandev:
@cbandy Tried your workaround but it just spits 65M and the file ends with "[Your log message was truncated]"

JordonPhillips:
@gorandev @cbandy I just merged in a fix, it'll be available in the next release.

JordonPhillips:
In the meantime, using `--marker '0'` should do the trick.

gorandev:
@JordonPhillips The --marker '0' doesn't work, guess I'll have to wait for next release! :)

PR #1873, opened 2016-03-23
Remove reference to old starting token format in docs

JordonPhillips:
There is one last remaining bit of the old world left, the RDS docs mentioned in #1835. There is a pr up at boto/botocore#849 to fix the code bug there, but the doc here is a bit challenging. 

RDS is using the `Marker` value `0` to indicate that it will start at the beginning of the logs. I'm not sure how I feel about transforming our docs to show `--starting-token eyJNYXJrZXIiOiAiMCJ9` or showing `--Marker 0` instead. We could also potentially make another flag (`--start-at-beginning`) to suit this scenario. What do you guys think?

cc @kyleknap @jamesls

### DIFF
--- a/awscli/examples/autoscaling/describe-auto-scaling-groups.rst
+++ b/awscli/examples/autoscaling/describe-auto-scaling-groups.rst
@@ -50,4 +50,4 @@ To return a specific number of Auto Scaling groups, use the ``max-items`` parame
 
 If the output includes a ``NextToken`` field, there are more groups. To get the additional groups, use the value of this field with the ``starting-token`` parameter in a subsequent call as follows::
 
-    aws autoscaling describe-auto-scaling-groups --starting-token None___1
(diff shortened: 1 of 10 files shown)


jamesls:
:shipit:.  Regarding, RDS my preference is to keep it as is for now.

nrdlngr:
It looks like all of these examples have been converted to Unicode files in this PR; they have non-ascii dashes in them that break the HTML builds. The AWS CLI reference seems to be able to convert the characters properly, which is probably why it doesn't break the Travis CI build, but local HTML builds on Mac and Ubuntu are failing.

Is there a way we can block non-ascii files from being committed to the AWS CLI examples? Or add a test that at least checks that the files are ASCII and break the Travis CI build with an error if they are not?

jamesls:
Yeah, I'm going to add a test for this now to ensure this doesn't happen again.

Issue #613, opened 2014-01-27
[1.2.11][ec2][run-instances] secondary-private-ip-address-count - InvalidParameterCombination


marekjagielski:
When running command with --secondary-private-ip-address-count 1:

aws ec2 run-instances --image-id ami-eef31499 --security-group-ids sg-bp8eo6dd --user-data data --instance-type t1.micro --subnet-id subnet-14d4uc63 --iam-instance-profile Name=gateway --count 1 --secondary-private-ip-address-count 1 --region eu-west-1 --profile int

I get:

A client error (InvalidParameterCombination) occurred when calling the RunInstances operation: Network interfaces and an instance-level subnet ID may not be specified on the same request

However setting separately --secondary-private-ip-address-count 1:

aws ec2 assign-private-ip-addresses --network-interface-id eni-243a2357 --secondary-private-ip-address-count 1 --region eu-west-1 --profile int

everything goes fine.

jamesls:
Looks like we need to expand on https://github.com/aws/aws-cli/pull/222 and account for a subnet id being specified.

kyleknap:
Fixed via referenced merged PR.

PR #1874, opened 2016-03-23
Fix issue with network interfaces

kyleknap:
There were some customizations for ec2 run instances that used the NetworkInterfaces parameter but did not lower some of the legacy parameters to the NetworkInterfaces param causing server side validation errors.

If you do not do this, you get errors like this:
```
ClientError: A client error (InvalidParameterCombination) occurred when calling the RunInstances operation: Network interfaces and an instance-level subnet ID may not be specified on the same request
```

There is even a note about this: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/ec2runinstances.py#L83 so it looks like these parameters just got missed because it now works for me.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/ec2runinstances.py
+++ b/awscli/customizations/ec2runinstances.py
@@ -91,20 +91,27 @@ def _fix_args(params, **kwargs):
     # allows them to specify the security group by name or by id.
     # However, in this scenario we can only support id because
     # we can't place a group name in the NetworkInterfaces structure.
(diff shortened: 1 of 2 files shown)


kyleknap:
Fixes https://github.com/aws/aws-cli/issues/613 as well

JordonPhillips:
:boat:

jamesls:
:shipit:

PR #1883, opened 2016-03-30
Fix command name in IAM delete-role example

mwunderl:


jamesls:
I rebase against develop to pick up the new ascii check tests and merged via 7de185d

PR #1897, opened 2016-04-06
Propagate error on S3 range download failures

jamesls:
When we download a file in parts, the DownloadPartTask
was not adding a print task error on the result queue.
As a result, the request would appear to succeed, not print
any error message, and exit with an rc of 0.

I also audited all the ``except Exception`` clauses in tasks.py
and made sure every other place was adding an ``failed=True`` task.
DownloadPartTask was the only one missing this.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #1902, opened 2016-04-08
Unhide Route53 semantic pagination token

JordonPhillips:
The only way to list a single record set requires using this argument, so it should be un-hid from the docs.

cc @jamesls @kyleknap

### DIFF
--- a/awscli/customizations/paginate.py
+++ b/awscli/customizations/paginate.py
@@ -35,6 +35,9 @@ from awscli.arguments import BaseCLIArgument
 
 logger = logging.getLogger(__name__)
 
+# There are some services where the pagination params need to be unhiden
+PUBLIC_PAGINATION_TOKENS = {'route53': ['start-record-name']}
+
 
 STARTING_TOKEN_HELP = """
 <p>A token to specify where to start paginating.  This is the
@@ -112,7 +115,12 @@ def unify_paging_params(argument_table, operation_model, event_name,
         return
     logger.debug("Modifying paging parameters for operation: %s",
                  operation_model.name)
-    _remove_existing_paging_arguments(argument_table, paginator_config)
+
+    service_name = operation_model.service_model.service_name
+    public_tokens = PUBLIC_PAGINATION_TOKENS.get(service_name, [])
+    _remove_existing_paging_arguments(
+        argument_table, paginator_config, public_tokens)
+
     parsed_args_event = event_name.replace('building-argument-table.',
                                            'operation-args-parsed.')
     shadowed_args = {}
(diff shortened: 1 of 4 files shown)


awscli/customizations/paginate.py
```diff
@@ -35,6 +35,9 @@
 
 logger = logging.getLogger(__name__)
 
+# There are some services where the pagination params need to be unhiden
+PUBLIC_PAGINATION_TOKENS = {'route53': ['start-record-name']}
```
kyleknap: This may need to get updated if we ever want to support operation level filtering, but it is fine for now.

tests/unit/customizations/test_paginate.py
```diff
@@ -54,6 +54,8 @@ def test_customize_arg_table(self):
             'foo': mock.Mock(),
             'bar': mock.Mock(),
         }
+        # Default to false
+        argument_table['foo']._UNDOCUMENTED = False
```
kyleknap: Why was this line needed to be added? Especially since it was not done for ``'bar'``.

kyleknap:
It might be nice to add a functional test to make sure ``--start-record-name`` shows up in the ``help`` command. But otherwise :ship:

tests/unit/customizations/test_paginate.py
JordonPhillips: Because we are calling `assertTrue` on the response. Since the object is a mock, you'll get a mock back if it hasn't been touched, which is truthy. Without this, those `assertTrue` calls are meaningless. Maybe they should be `assertEqual` to avoid this in a more explicit way.

kyleknap: Make sense. It was just strange that bar did not have its ``_UNDOCUMENTED`` set to ``False`` as well.

jamesls:
I'm a little unclear why we need this.  The point of hiding this is to unify the pagination token.  `--starting-token` accomplishes this.  What's the reason for having two ways to do this?

JordonPhillips:
@jamesls Because this token is used outside of the pagination context to get a single resource record set.

jamesls:
There are several places where the concept exists of a non opaque pagination token that a customer could construct and the CLI has  decided to unify these names.

It doesn't make sense to have this be inconsistent with other APIs in the CLI.  If we're going to start to unhide these, we need to explore making this available in a consistent way, not special cased for this one API.  I would still consider specifying a starting token with a max items of 1 to be "pagination", and I think it makes sense to keep it unified and have them go through the pagination interface we expose rather than exposing the underlying pagination params.

JordonPhillips:
@jamesls I agree in the general case, but in this case there is no `GET` for the target resource. They are extracting a single, specific value. They can do this in one request by providing `start-record-name`, or in n requests by actually paginating to try to find the item. It could very well be argued that the service should supply a function that does just this, but in the meantime I think having a stop-gap such as this is beneficial to customers.

jamesls:
> or in n requests by actually paginating to try to find the item

What I am saying is there is no difference in the CLI between:

`aws route53 list-resource-record-sets --start-record-name my.domain.`

and

`aws route53 list-resource-record-sets --starting-token my.domain.`

It maps to the same param.  They can already do that right now without this change.  `--start-record-name` is not just to get a single record.  It is the *starting* record name.  If you want a single record then you have to specify max-items 1 along with the start record name.  Same thing when using `--starting-token`.

What I don't understand is if there's already a way to do this that's consistent with the existing paging model we have (and requires you to use paging parameters to restrict the result set to a single item anyways), what's the benefit in exposing the underlying paging param?

JordonPhillips:
@jamesls I looked more into this, and I think I agree with you now. I'll see what can be done to clear up customer confusion

PR #1905, opened 2016-04-11
Shutdown task executor before aborting multipart uploads

jamesls:
There's a race condition in which multipart uploads can still be
kept alive because aborting a multipart uploads can happen while
existing tasks are still being executed.  We must shutdown the
executor before we can then abort multipart uploads.

This order was correct in the normal shutdown case, but for
the case where a user 'Ctrl-C's the process or where unexpected
exceptions propogate back to the S3 handler, the order was reversed.

In other words, the race condition exposed two possibilities:

Proper cleanup:

```
    t1 ---|start_part|-------------|end_part|---------------------
                                                                 |  join
    t2 ---|start_part|-------------|end_part|---------------------
                                                                 |  join
  main -----------------|ctrl-c|---------------|abort_upload|-----
```
Bad cleanup:
```
    t1 ---|start_part|--------------------------------|end_part|--
                                                                 |  join
    t2 ---|start_part|--------------------------------|end_part|--
                                                                 |  join
  main -----------------|ctrl-c|--|abort_upload|------------------
```

New behavior:

```
    t1 ---|start_part|---------------|end_part|-+
                                                | join
    t2 ---|start_part|---------------|end_part|-+
                                                | join
  main -----------------|ctrl-c|----------------+-------|abort_upload|
```
There's also a gap in test coverage here, so I've added a test case
for this case.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -101,28 +101,35 @@ class S3Handler(object):
             self.executor.print_thread.set_total_files(total_files)
             self.executor.print_thread.set_total_parts(total_parts)
             self.executor.initiate_shutdown()
-            self.executor.wait_until_shutdown()
-            self._shutdown()
+            self._finalize_shutdown()
         except Exception as e:
             LOGGER.debug('Exception caught during task execution: %s',
                          str(e), exc_info=True)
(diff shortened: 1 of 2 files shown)


kyleknap:
:ship:

JordonPhillips:
:boat:

PR #1906, opened 2016-04-12
Fix bugs when aborting multipart uploads on failure

jamesls:
There were two issues I fixed when aborting multipart uploads.

First, we were not invoking the cancel callback passed to the
multipart upload context when a part task has requested that
an upload is cancelled.  This happens when the UploadPart API
call fails (due to a timeout for example).  This is because
the cancel callback is only invoked when the context state
is in _STARTED.  It doesn't account for contexts in the
_CANCELLED state that haven't had their uploads aborted.

The second issue is a less likely case.  We were only tracking
a multipart upload after we queued the complete multipart upload
task.  This means that if something in the Main
thread (which does the enqueuing) generates an exception after
the create upload task is queued but before the complete task
is queued, the S3 handler won't be tracking this upload yet
so it won't know to abort the upload.  Now we add an upload
to our list of tracked uploads as soon as we queue a create
upload task.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -136,20 +136,12 @@ class S3Handler(object):
         self._remove_pending_downloads()
 
     def _abort_pending_multipart_uploads(self):
-        # For the purpose of aborting uploads, we consider any
-        # upload context with an upload id.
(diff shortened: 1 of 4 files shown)


kyleknap:
:ship:

PR #1907, opened 2016-04-12
Validate build directory gamelift

kyleknap:
The build directory needs to exist and be nonempty to upload builds to GameLift.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/gamelift/uploadbuild.py
+++ b/awscli/customizations/gamelift/uploadbuild.py
@@ -41,6 +41,15 @@ class UploadBuildCommand(BasicCommand):
             endpoint_url=parsed_globals.endpoint_url,
             verify=parsed_globals.verify_ssl
         )
+        # Validate a build directory
+        if not validate_directory(args.build_root):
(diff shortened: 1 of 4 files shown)


awscli/customizations/gamelift/uploadbuild.py
```diff
@@ -102,6 +111,13 @@ def zip_directory(zipfile_name, source_root):
                     zf.write(full_path, relative_path)
 
 
+def validate_directory(source_root):
+    for path, dirs, files in os.walk(source_root):
```
jamesls: Is there a reason we can't just use `return os.path.isdir(source_root) and os.listdir(source_root)`?

jamesls:
:shipit:  Had one small comment, otherwise looks good.

awscli/customizations/gamelift/uploadbuild.py
kyleknap: Yeah I will add it.

kyleknap:
I added a comment. Merging.

PR #1908, opened 2016-04-13
Add new-change script for generating changelog entries

jamesls:
This is the first script to help out with making changelog entries easier to handle with parallel development work.

This script creates a JSON file in `.changes/next-release/` that contains: `{'type': ..., 'category': ..., 'description': ...}`.  That way there's no merge conflicts as new changes are added and it becomes easier to automatically for in pull requests.

You can either specify all these values as command line options, or if no args are provided it will open your editor and allow you to type in these values.

I had two follow up pull requests in mind.  One will be for taking everything in `.changes/next-release` and converting them to a single JSON file (e.g. `1.10.0`), and the other will be for generating the CHANGELOG file from the JSON files.

cc @kyleknap @JordonPhillips

### DIFF
--- a/scripts/new-change
+++ b/scripts/new-change
@@ -1,0 +1,201 @@
+#!/usr/bin/env python
+"""Generate a new changelog entry.
+
+Usage
+=====
+
+To generate a new changelog entry::
+
+    scripts/new-change
+
+This will open up a file in your editor (via the ``EDITOR`` env  var).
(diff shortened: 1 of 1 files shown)


jamesls:
Actually hold off on this.  I was going to send them as 3 separate PRs, but they're fairly small and somewhat related, so I'll send all three as a single PR.

PR #1914, opened 2016-04-14
Fix non deterministic test

jamesls:
This test was occasionally failing.  The test should
now reliably pass.  I've now ran it 1000 times and have not
seen any errors.

Follow up from https://github.com/aws/aws-cli/pull/1906

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #1915, opened 2016-04-14
Remove back-compat shorthand features from new services

JordonPhillips:
The way we currently handle the shorthand parser means that a service which has an argument that is modeled as a list of structures with a single member can never add new members. This makes what was a perfectly fine change into a breaking change. We can't change this behavior for existing services, but this will prevent the problem from cascading to new services.

For example, the `elb` operation `remove-tags` has such a list. You call it like so:

`aws elb remove-tags --load-balancer-names foo bar --tags baz bin`

You will still be able to call it like that. If, however, they added a new operation `modify-tags` with the same list structure, it would be called like so:

`aws elb modify-tags --load-balancer-names foo bar --tags Key=baz Key=bin`

Note that `--load-balancer-names` is a list of strings, not a list of structures.

cc @jamesls @kyleknap

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -21,6 +21,7 @@ from awscli import SCALAR_TYPES, COMPLEX_TYPES
 from awscli.paramfile import get_paramfile, ResourceLoadingError
 from awscli.paramfile import PARAMFILE_DISABLED
 from awscli import shorthand
+from awscli.utils import find_service_and_method_in_event_name
 
 
 LOG = logging.getLogger('awscli.argprocess')
@@ -247,11 +248,39 @@ def _is_complex_shape(model):
 
 class ParamShorthand(object):
 
+    def _uses_old_list_case(self, service_name, operation_name, argument_name):
+        """
+        Determines whether a given operation for a service needs to use the
+        deprecated shorthand parsing case for lists of structures that only have
+        a single member.
+        """
+        cases = {
+            'firehose': {
+                'put-record-batch': ['records']
+            },
+            'workspaces': {
+                'reboot-workspaces': ['reboot-workspace-requests'],
+                'rebuild-workspaces': ['rebuild-workspace-requests'],
+                'terminate-workspaces': ['terminate-workspace-requests']
+            },
+            'elb': {
+                'remove-tags': ['tags'],
+                'describe-instance-health': ['instances'],
+                'deregister-instances-from-load-balancer': ['instances'],
+                'register-instances-with-load-balancer': ['instances']
+            }
+        }
+        cases = cases.get(service_name, {}).get(operation_name, [])
+        return argument_name in cases
+
+
+class ParamShorthandParser(ParamShorthand):
+
     def __init__(self):
         self._parser = shorthand.ShorthandParser()
         self._visitor = shorthand.BackCompatVisitor()
 
-    def __call__(self, cli_argument, value, **kwargs):
+    def __call__(self, cli_argument, value, event_name, **kwargs):
         """Attempt to parse shorthand syntax for values.
 
         This is intended to be hooked up as an event handler (hence the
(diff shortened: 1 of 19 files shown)


awscli/argprocess.py
```diff
@@ -278,16 +301,22 @@ def __call__(self, cli_argument, value, **kwargs):
             be raised.
 
         """
+        split_event = event_name.split('.')[1:]
+        service_name = split_event[0]
+        operation_name = split_event[1]
```
jamesls: We can move this into the else clause so it only gets executed when we need to parse as shorthand.

awscli/argprocess.py
```diff
@@ -245,13 +245,36 @@ def _is_complex_shape(model):
     return True
 
 
+def _uses_old_list_case(service_name, operation_name, argument_name):
```
jamesls: Any reason this needs to be at the module level?  Seems like this is specific to the shorthand parsing class, so it makes sense in there.

awscli/clidocs.py
```diff
@@ -403,7 +403,10 @@ def _doc_input_structure_members(self, doc, argument_model, stack):
                 doc.style.new_line()
         doc.write('}')
 
-    def doc_option_example(self, arg_name, help_command, **kwargs):
+    def doc_option_example(self, arg_name, help_command, event_name, **kwargs):
+        split_event = event_name.split('.')[1:]
+        service_name = split_event[0]
+        operation_name = split_event[1]
```
jamesls: Duped from `argprocess`.  Worth extracting out into a utility method?

tests/functional/docs/test_help_output.py
```diff
@@ -177,6 +177,10 @@ def test_elb_help_output(self):
         # .. _`:
         self.assert_not_contains('.. _`:')
 
+    def test_shorthand_flattens_list_of_single_member_structures(self):
```
jamesls: Are there other functional tests that cover the remaining cases?  I think it's worth having functional tests at least one per service, just to have that extra safety net.

If there's existing assert_params_for_cmd functional tests for the other commands/params then I think we're good with functional tests.

jamesls:
Had some small comments.  I'll look over the tests shortly, but the approach seems reasonable.

tests/functional/docs/test_help_output.py
JordonPhillips: There is at least one for remove tags, I'll check the rest.

awscli/argprocess.py
JordonPhillips: This is used in both the shorthand parser and the shorthand doc generator, which are both in this module. I could put it as part of a super class for both of them to get it out of the module scope, though I'm not sure that makes much sense.

awscli/argprocess.py
```diff
@@ -245,13 +246,41 @@ def _is_complex_shape(model):
...
+    """
+    Determines whether a given operation for a service needs to use the
+    deprecated shorthand parsing case for lists of structures that only have
+    a single member.
+    """
+    if service_name == 'elb':
+        service_name = 'elasticloadbalancing'
```
kyleknap: A little curious. What event (I am assuming that is using endpointPrefix as service name) is this if statement needed for?

awscli/argprocess.py
kyleknap: Makes sense

tests/functional/docs/test_help_output.py
kyleknap: I agree with @jamesls. It would be nice to have.

awscli/argprocess.py
```diff
@@ -245,13 +246,41 @@ def _is_complex_shape(model):
...
+    cases = {
+        'firehose': {
+            'put-record-batch': ['records']
+        },
+        'workspaces': {
+            'reboot-workspaces': ['reboot-workspace-requests'],
+            'rebuild-workspaces': ['rebuild-workspace-requests'],
+            'terminate-workspaces': ['terminate-workspace-requests']
+        },
+        'elasticloadbalancing': {
```
kyleknap: I would prefer if we that we key by service/command name rather than endpoint. There is no guarantee that endpoint prefixes are not the same for different clients/commands.

kyleknap:
Looks good. I had a couple of comments as well.

awscli/argprocess.py
JordonPhillips: the doc events

JordonPhillips:
This PR is now dependent upon #1956. Otherwise I'll have to put in the line to correct `elasticloadbalancing` to `elb` back in.

tests/functional/docs/test_help_output.py
kyleknap: I think this comment still stands. Could we add functional tests (one for each service) to make sure the backcompat works.

kyleknap:
For me to get this merged, lets make sure all of the test pass (should get fixed when the event renaming gets merged) and functional tests are added for the back compat layer added for each hardcoded service.

JordonPhillips:
Added functional tests and rebased against #1956 so the tests will pass.

JordonPhillips:
I'll be squashing those commits before merge

kyleknap:
Looks good. Probably did not have to add a functional test for each operation command. I was just asking one for each service, but still good with me. :ship:

JordonPhillips:
Squashed the commits, just waiting on a shipit for #1956

Issue #1901, opened 2016-04-08
Doc strings for custom arguments are not being generated


JordonPhillips:
Something seems to have broken our doc strings for custom arguments. See [cp](http://docs.aws.amazon.com/cli/latest/reference/s3/cp.html) as an example vs [wayback machine](https://web.archive.org/web/20150809022927/http://docs.aws.amazon.com/cli/latest/reference/s3/cp.html?).

PR #1917, opened 2016-04-14
Fix invalid xml in the docs

kyleknap:
It caused the html parser to have an incorrect representation of the html tree which messed up the rendering of the docs. Here are the docs with the fix now:
screen shot 2016-04-14 at 1 53 44 pm

Fixes https://github.com/aws/aws-cli/issues/1901

cc @jamesls @JordonPhillips


JordonPhillips:
Nice find! :boat:

PR #1918, opened 2016-04-14
Fix issue with boolean argument translations

kyleknap:
When building the translation table in the docs, the CLI may use the
negative version of boolean argument, which messes up the meaning of some
of the docs.

Here is an example for ``dynamodb get-item`` and ``consistent-read`` argument:

**Before:**
screen shot 2016-04-14 at 3 35 44 pm

**After:**
screen shot 2016-04-14 at 3 35 18 pm

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -314,7 +314,8 @@ class OperationDocumentEventHandler(CLIDocumentEventHandler):
                     # results in the new proposed cli argument name. If it
                     # does, we assume we have the postive form of the argument
                     # which is the name we want to use in doc translations.
-                    if previous_mapping.startswith('no-') and \
+                    if cli_argument.cli_type_name == 'boolean' and \
+                            previous_mapping.startswith('no-') and \
                             cli_name == previous_mapping[3:]:
                         d[argument_name] = cli_name
                 else:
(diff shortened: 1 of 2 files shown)


JordonPhillips:
:boat:

awscli/clidocs.py
```diff
@@ -304,7 +304,21 @@ def build_translation_map(self):
...
+                if argument_name in d:
+                    previous_mapping = d[argument_name]
+                    # If the argument name is a boolean argument, we want the
+                    # the translation to default to the one that does not start
+                    # with --no-. So we check if the cli parameter currently
+                    # being used starts with no- and if stripping off the no-
+                    # results in the new proposed cli argument name. If it
+                    # does, we assume we have the postive form of the argument
+                    # which is the name we want to use in doc translations.
+                    if previous_mapping.startswith('no-') and \
```
jamesls: This seems like it has potential for false positives.  Can we restrict this check to be more specific?  What about checking if arg is an instance of the boolean argument?

kyleknap:
@jamesls I updated based on feedback.

jamesls:
:shipit:

Do you think this warrants a changelog entry?  If so, don't forget to add one before merging.

kyleknap:
Not so much because it is just a doc update.

PR #1922, opened 2016-04-15
Remove deprecated -e option for docker login

zackhsi:
```
Warning: '-e' is deprecated, it will be removed soon. See usage.
```

### DIFF
--- a/awscli/customizations/ecr.py
+++ b/awscli/customizations/ecr.py
@@ -52,6 +52,6 @@ class ECRLogin(BasicCommand):
         for auth in result['authorizationData']:
             auth_token = b64decode(auth['authorizationToken']).decode()
             username, password = auth_token.split(':')
-            sys.stdout.write('docker login -u %s -p %s -e none %s\n'
+            sys.stdout.write('docker login -u %s -p %s %s\n'
                              % (username, password, auth['proxyEndpoint']))
         return 0
--- a/awscli/examples/ecr/get-authorization-token.rst
+++ b/awscli/examples/ecr/get-authorization-token.rst
@@ -50,7 +50,7 @@ nothing with this information, but ``docker login`` required the email field).
 
 Command::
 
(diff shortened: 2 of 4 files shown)


jamesls:
Associated issue: #1926 

I just commented on that issue (https://github.com/aws/aws-cli/issues/1926#issuecomment-221278732), but it looks like there's issues regarding backwards compatibility of remove the `-e` option.

I'm following up with the ECR team, but for now I think we'll have to keep this unless there's a way to preserve backwards compatibility.

sentientmonkey:
We've filed an issue with docker and you can track it here: [#23390](https://github.com/docker/docker/issues/23390)

We will update this issue based on the outcome.

cauealves:
👍

neurolabs:
The -e flag is removed in docker 17.06, so aws ecr get-login is effectively broken.

Please merge this pull request.

alasdairnicol:
@neurolabs you can use the `--no-include-email` option introduced in #2619

alasdairnicol:
@jamesls I believe this should have been closed by #2619 along with #1926.

neurolabs:
works with `--no-include-email`, could be closed. Although I didn't find this option and resorted to `|sed 's/-e none//'` first. More documentation would probably be a good thing.

neurolabs:
Or maybe switch the default to `--no-include-email`? I would imagine that that would break stuff for less people than the other way around.

almorelle:
Any update on this ?

cauealvesbraz:
Hey @almorelle, 

After I updated my `aws-cli` the problems are solved.

I'm using this stack:
```bash
aws-cli/1.16.110 Python/3.7.2 Darwin/18.2.0 botocore/1.12.100
```

almorelle:
@cauealvesbraz I'm at aws-cli/1.16.90 Python/3.7.2 Darwin/18.2.0 botocore/1.12.80 which doesn't seem far behind but I'll try to upgrade anyway. thanks

PR #1924, opened 2016-04-18
Add example to list resource record sets by name

JordonPhillips:
This is to clear up some confusion with the route53 api in the CLI

cc @jamesls @kyleknap


kyleknap:
:ship:

Issue #1238, opened 2015-03-24
Unknown encoding on Mac OS X


freimer:
receiving unknown encoding message on Mac OS X even with PYTHONIOENCODING=utf-8 env var.  Will fix and do PR.

braddr:
This fixed the issue for me on osx 10.11 as well.  @freimer, how about submitting the fix as a pull request?

kjenney:
This fixed the issue for me on OSX 10.11.4. This needs to be merged, though.

blochberger:
Can be worked around when setting `LC_CTYPE=en_US.UTF-8` (was `UTF-8`).
I think that is a bug rather than a feature request.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

PR #1930, opened 2016-04-20
Handle unknown encoding on Mac OS X

freimer:
It's been a long time since I did this, but I got some requests to do a pull request to get this in the official code.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -46,7 +46,7 @@ if six.PY3:
 
         """
         if 'b' not in mode:
-            encoding = locale.getpreferredencoding()
+            encoding = locale.getpreferredencoding(do_setlocale=False)
         return open(filename, mode, encoding=encoding)
 
 else:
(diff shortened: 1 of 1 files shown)


jamesls:
Adding additional context here:

Original PR: https://github.com/aws/aws-cli/pull/1239
Issue: https://github.com/aws/aws-cli/issues/1238

awscli/compat.py
```diff
@@ -68,10 +70,15 @@ def get_stdout_text_writer():
...
+        encoding = locale.getpreferredencoding()
+        if encoding == '':
+            encoding = locale.getpreferredencoding(do_setlocale=False)
+        return codecs.getwriter(encoding)(sys.stdout)
 
     def compat_open(filename, mode='r', encoding=None):
         # See docstring for compat_open in the PY3 section above.
         if 'b' not in mode:
-            encoding = locale.getpreferredencoding()
+            encoding = locale.getpreferredencoding(do_setlocale=False)
```
jamesls: I think this shouldn't have the `do_setlocale` param right?

freimer: I believe that is what breaks it on Mac OS X.  The default is True, which returns a blank encoding causing the issue.  If it is set to False, then it works.  Perhaps it can be wrapped in an if like lines 73-75.  If I still have the code I'll do a new commit.

freimer: I actually did change this back lase June, but never pushed it.  It's up there now.

stealthycoin:
Closing this in accordance with the linked issue. You can set LC_ALL to resolve this issue which is the recommended solution. If you want to get this feature added to the CLI, you would need to open an issue, and collect 10 or more 👍 according to our new contributing process [described here](https://github.com/kdaily/aws-cli/blob/contrib-guide-proposal/proposals/contribution-guide.md#implementation-stage-1).

PR #1938, opened 2016-04-25
Fix string compatibility issue in scripts/new-change

JordonPhillips:
Python 3 does not have `string.letters` due to the ambiguous nature of
the word 'letters' when locales are taken into consideration. To get the
same effect, you can use `string.ascii_letters`, which is available on
all versions we support.

cc @kyleknap @jamesls

### DIFF
--- a/scripts/new-change
+++ b/scripts/new-change
@@ -47,7 +47,7 @@ import subprocess
 import argparse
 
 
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:

PR #1939, opened 2016-04-25
Add newlines to end of change files

JordonPhillips:
All files should have a newline before EOF, this changes the new-change
script to add them and backfills them into all existing change files.

cc @kyleknap @jamesls


jamesls:
:shipit:

PR #1940, opened 2016-04-25
Use a shared bucket across all test_plugin tests

jamesls:
This test adds a setup/teardown at the module level that
creates a bucket that all tests can use.  This change
was made for several reasons:

* Reduce number of resources created.  Considering multiple test runs
  on multiple python version, multiple platforms, etc.
* Have a more robust check in place for creating a bucket.  This
  better handles eventual consistency.
* Improve speed.  Rather than having to create/delete a bucket each
  test, we can reuse a bucket across all tests.  As part of this
  change, the shared bucket is emptied out before each test.
  From my testing, this shaves about 3 minutes off the test runs.

There are some tests that either require a specific bucket (a bucket
in a specific region) or multiple buckets.  This PR does not address
those cases.

cc @kyleknap @JordonPhillips

### DIFF
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -27,17 +27,71 @@ import socket
 import tempfile
 import shutil
 import copy
+import logging
 
 from awscli.compat import six
 from nose.plugins.attrib import attr
+import botocore.session
 
 from awscli.testutils import unittest, get_stdout_encoding
 from awscli.testutils import skip_if_windows
 from awscli.testutils import aws as _aws
 from awscli.testutils import BaseS3CLICommand
-from tests.integration.customizations.s3 import create_bucket as _create_bucket
 from awscli.customizations.s3.transferconfig import DEFAULTS
 from awscli.customizations.scalarparse import add_scalar_parsers
+
+
+# Using the same log name as testutils.py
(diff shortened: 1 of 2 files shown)


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -466,8 +527,10 @@ def test_copy_large_file_signature_v4(self):
     def test_copy_metadata(self):
         # Copy the same style of parsing as the CLI session. This is needed
         # For comparing expires timestamp.
+        # TODO: james - do we need this?  seems like we shouldn't be doing
+        # this.
         add_scalar_parsers(self.session)
```
kyleknap: Yeah I am not sure. Maybe because we create the session and a handler that is needed does not get added because it does not go through the entire CLI process?

kyleknap:
Looks good. :ship:

tests/integration/customizations/s3/test_plugin.py
jamesls: Ah crap, sorry about that.  I meant to remove that, it's more of an internal note for me.

I'd rather just use the real session via `create_clidriver().session`.  Given these are integration tests, that seems reasonable to use as realistic of a session as possible.

In the meantime I'll remove this.

PR #1943, opened 2016-04-28
Add --no-param-validation to disable parameter validation

jamesls:
cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/data/cli.json
+++ b/awscli/data/cli.json
@@ -15,6 +15,11 @@
             "dest": "verify_ssl",
             "help": "<p>By default, the AWS CLI uses SSL when communicating with AWS services.  For each SSL connection, the AWS CLI will verify SSL certificates.  This option overrides the default behavior of verifying SSL certificates.</p>"
         },
+        "no-param-validation": {
+            "action": "store_false",
+            "dest": "param_validation",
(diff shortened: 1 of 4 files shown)


awscli/data/cli.json
```diff
@@ -15,6 +15,11 @@
             "dest": "verify_ssl",
             "help": "<p>By default, the AWS CLI uses SSL when communicating with AWS services.  For each SSL connection, the AWS CLI will verify SSL certificates.  This option overrides the default behavior of verifying SSL certificates.</p>"
         },
+        "no-param-validation": {
+            "action": "store_false",
+            "dest": "param_validation",
+            "help": "<p>Disable parameter validation (enabled by default).  Specifying this parameter will disable parameter validation.  You typically should not need to disable parameter validation.</p>"
```
kyleknap: Might be nice to say what parts of validation gets ignored.

kyleknap:
I think I am fine with the global argument, but what's your thought on exposing it in the config? I feel that I would really only use this functionality once I have a working script. I feel that it may be tedious to have to go to every line in a working script where the CLI is used and add ``--no-param-validation``. Honestly, I am thinking that I would like to see it exposed in the config first and then as a global flag. The main downside that I see of doing this is that other SDKs may have to consider adopting it if they adopt ~/.aws/config.

jamesls:
Closing in preference to adding this in botocore: https://github.com/boto/botocore/pull/905

PR #1945, opened 2016-04-29
Flush stdout after input prompt.

### LINKED ISSUE #1925: Not properly outputting in MINGW
System: Windows 7 64bit

steps to reproduce:

1. Open mingw64 (git bash) 
2. Run `aws configure`
3. Observe no output
4. Hit enter 4 times
5. Observe output

It looks as if AWS CLI is thinking it is outputting text however it actually isn't.

image

JordonPhillips:
Some terminal emulators (such as mintty) will not write output to the terminal window until stdout is explicitly flushed. When running `aws configure` a user would see nothing but a new line. After pressing 'enter' four times, they would then see the output flushed all in a single line. This makes it very difficult to configure the command.

The solution is to call `flush` after every prompt. Since `raw_input` does not have an option to do this, we have to prompt and flush manually. Since we're already accessing stdout directly, it's easier to write to it directly since print adds some formatting that we don't want (namely, a newline at the end of the print).

Fixes #1925

cc @kyleknap @jamesls

I tested this out on my personal machine to verify that it fixes the issue. Amusingly, the default python REPL (for 2.7) falls prey to the same issue.

### DIFF
--- a/awscli/customizations/configure/configure.py
+++ b/awscli/customizations/configure/configure.py
@@ -15,7 +15,7 @@ import logging
 
 from botocore.exceptions import ProfileNotFound
 
-from awscli.compat import raw_input
+from awscli.compat import compat_input
 from awscli.customizations.commands import BasicCommand
 from awscli.customizations.configure.addmodel import AddModelCommand
 from awscli.customizations.configure.set import ConfigureSetCommand
@@ -39,7 +39,7 @@ class InteractivePrompter(object):
     def get_value(self, current_value, config_name, prompt_text=''):
         if config_name in ('aws_access_key_id', 'aws_secret_access_key'):
             current_value = mask_value(current_value)
-        response = raw_input("%s [%s]: " % (prompt_text, current_value))
+        response = compat_input("%s [%s]: " % (prompt_text, current_value))
         if not response:
             # If the user hits enter, we return a value of None
             # instead of an empty string.  That way we can determine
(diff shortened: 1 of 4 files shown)


jamesls:
Can you elaborate on the issue?  I'm not following what the bug is.  A terminal will never display anything until it's written to.  `.flush()` is an application level construct.  Why is the input being buffered for mintty?

jamesls:
Nevermind, I think I answered my own question:

https://code.google.com/archive/p/mintty/issues/212
https://code.google.com/archive/p/mintty/issues/56

which if I'm paraphrasing correctly means that we see the console as a pipe instead of a tty, hence the 4k buffer as the default.

awscli/customizations/configure/configure.py
```diff
@@ -39,7 +40,16 @@ class InteractivePrompter(object):
     def get_value(self, current_value, config_name, prompt_text=''):
         if config_name in ('aws_access_key_id', 'aws_secret_access_key'):
             current_value = mask_value(current_value)
-        response = raw_input("%s [%s]: " % (prompt_text, current_value))
```
jamesls: Would we ever not want this behavior?

Seems like we should fix this in compat rather than at the call site.

JordonPhillips:
Updated the comment to be more detailed and link to the [mintty github issue](https://github.com/mintty/mintty/issues/56).

kyleknap:
Looks good. :ship:

PR #1950, opened 2016-05-02
Direct Connect examples

mwunderl:


mwunderl:
The describe-locations API includes non-ascii characters in its output. It would be better if we could include the output as-is.

kyleknap:
Yeah that is unfortunate. I agree that this is something we should look to see if we can handle as we are able to handle unicode fine in the documentation in the model. Otherwise, merging PR.

mwunderl:
Thanks Kyle

PR #1956, opened 2016-05-04
Use service_name in process-cli-arg event

JordonPhillips:
Currently we use the endpoint prefix when generating this event, which means service whose names we have manually changed will emit events with the old name. For example, elb will emit as elasticloadbalancing. This makes registering event handlers confusing, especially since most other event handlers use the service name. Furthermore, multiple services can use the same endpoint, so this could be a source of bugs if not changed.

This changes the event from `process-cli-arg.endpoint-prefix.operation-name` to `process-cli-arg.service-name.operation-name`.

To make sure that no existing events handlers were broken, I wrote a script to find all the services whose service name is different from the endpoint prefix. Specifically, this only applies to modeled commands since CustomArgument does not inherit from CLIArgument. The result of that search is the following:

Service Name | Endpoint Prefix
------------ | ---------------
cloudwatch   | monitoring
dynamodbstreams | streams.dynamodb
efs | elasticfilesystem
elb | elasticloadbalancing
emr | elasticmapreduce
iot-data | data.iot
meteringmarketplace | metering.marketplace
ses | email

I then did a search through the source code to make sure we weren't registering any handlers against the process-cli-arg event for those services. Thankfully, we are not. Since the CLI isn't open for customization, we don't need to worry about trampling customers.

cc @kyleknap @jamesls

### DIFF
--- a/tests/unit/test_arguments.py
+++ b/tests/unit/test_arguments.py
@@ -10,8 +10,12 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import mock
+
 from awscli.testutils import unittest
 from awscli import arguments
+from botocore.model import StringShape, OperationModel, ServiceModel
+
 
 class DemoArgument(arguments.CustomArgument):
     pass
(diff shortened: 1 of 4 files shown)


kyleknap:
Thanks for looking compiling the table and looking through the CLI codebase. I have not looked into it too deeply (I will make a pass through to make sure I do not see any customizations), but one thing to keep in mind is that there are some plugins available like the awscl-cwlogs: https://pypi.python.org/pypi/awscli-cwlogs/1.0.0. Fortunately, cloudwatch logs client name matches the endpoint name so that should not be affected and its code stays strictly in the ``logs`` domain.

I think it would be worth the time to make sure there are no other plugins that could be affected and plan accordingly. In general, I am in favor of this change because we should not be using endpoint prefix for event names in the CLI codebase and I am in favor of it as long as we understand the impact and can account for it.

JordonPhillips:
@kyleknap I did a search of github and found no repos subscribing to any of those events. I also looked through pypi to find cli plugins and couldn't find any that subscribe to these events.

kyleknap:
I looked into it as well. I could not find any places where this event gets registered both in the CLI and other libraries/tools. The only other place that I can find endpoint_prefix used is in the ``load-cli-arg`` event. So it may be good to update this while we are at it.

For this change though :ship:. I would like to get @jamesls to look at it as well though to make sure that he is fine with it as well.

tests/unit/test_arguments.py
```diff
@@ -23,3 +26,36 @@ def test_can_set_required(self):
...
         arg.required = True
         self.assertTrue(arg.required)
+
+
+class TestCLIArgument(unittest.TestCase):
+    def setUp(self):
+        self.arg_name = 'foo'
+        self.op_name = 'bar'
+        self.service_name = 'baz'
+        self.argument_model = mock.Mock()
```
jamesls: I would prefer to either not have this test or to use real argument_model objects.  I don't think there's a lot of value in mocking out the model objects.

jamesls:
Change seems fine to me, had a small comment about the test.

JordonPhillips:
Updated the test

jamesls:
:shipit:

PR #1959, opened 2016-05-05
Adds example for dynamodb put-item conditional expression 

bchav:
Just a quick example for conditional expressions to help with syntactical questions.

### DIFF
--- a/awscli/examples/dynamodb/put-item.rst
+++ b/awscli/examples/dynamodb/put-item.rst
@@ -4,7 +4,7 @@ This example adds a new item to the *MusicCollection* table.
 
 Command::
 
-  aws dynamodb put-item --table-name MusicCollection --item file://item.json --return-consumed-capacity TOTAL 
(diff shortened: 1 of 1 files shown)


kyleknap:
Thanks! The content looks good to me.

I think the ReStructuredText formatting looks a little off when I look at your [branch](https://github.com/bchav/aws-cli/blob/develop/awscli/examples/dynamodb/put-item.rst) as everything that you added looks to be in a code block.

I think you should fix that by making sure that are two newlines after every the end of every codeblock (the ``::`` notation). Also, could you move the line ``If the key already exists, you should see:`` out of the ``Output`` code block? The output codeblock should be reserved to the error being thrown.

bchav:
Thanks, that took a bunch of trial and error to get formatting correct here :+1:

kyleknap:
Awesome. Looks much better. One last thing. Could you squash your commits into a single commit and repush the branch? There are a lot of "formatting" commits that not ideal to have in the commit history.

bchav:
Squashed. Let me know if everything checks out.

kyleknap:
Thanks! Merging.

PR #1970, opened 2016-05-09
Move EC2 customizations into a single directory

JordonPhillips:
There are several ec2 customizations cluttering up the customizations directory, so I moved them into their own directory to isolate them. PyCharm's refactoring cleans up imports when it goes in to change them, which is why handlers has so many changes.

cc @kyleknap @jamesls


kyleknap:
Looks good. Definitely easier to find ec2 customizations now. :ship:

PR #1971, opened 2016-05-09
Default ec2 page size to 1000

JordonPhillips:
If this value isn't set, EC2 will not paginate at all. Instead, they will
attempt to return all values to you at once. Naturally, this can cause
significant delay if you have a ton of resources.

Built off of #1970 

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/ec2/paginate.py
+++ b/awscli/customizations/ec2/paginate.py
@@ -1,0 +1,27 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+DEFAULT_MAX_RESULTS = 1000
+
+
+def set_max_results_default(parsed_args, parsed_globals, **kwargs):
+    """
+    In order to have EC2 return results you can paginate you need to inject
+    the `MaxItems` parameter. In the CLI we should be setting this by default
+    to avoid users having to wait exceedingly long times for the full results.
+    """
+
+    # The check for page size validates that the operation is a pagination
+    # operation.
+    if parsed_globals.paginate and hasattr(parsed_args, 'page_size') and \
+            parsed_args.page_size is None and parsed_args.max_results is None:
+        parsed_args.page_size = DEFAULT_MAX_RESULTS
--- a/awscli/handlers.py
+++ b/awscli/handlers.py
@@ -44,6 +44,7 @@ from awscli.customizations.ec2.decryptpassword import ec2_add_priv_launch_key
 from awscli.customizations.ec2.protocolarg import register_protocol_args
 from awscli.customizations.ec2.runinstances import register_runinstances
 from awscli.customizations.ec2.secgroupsimplify import register_secgroup
+from awscli.customizations.ec2.paginate import set_max_results_default
 from awscli.customizations.ecr import register_ecr_commands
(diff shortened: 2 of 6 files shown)


awscli/handlers.py
```diff
@@ -102,6 +103,7 @@ def awscli_initialize(event_handlers):
                             ec2_add_priv_launch_key)
     register_parse_global_args(event_handlers)
     register_pagination(event_handlers)
+    event_handlers.register('operation-args-parsed.ec2.*', set_max_results_default)
```
kyleknap: Should we be doing this for every ec2 operation? Does every ec2 operation have the "paginate and give me a pagination token only if you gave me a page size to start"?

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def set_max_results_default(parsed_args, parsed_globals, **kwargs):
+    """
+    In order to have EC2 return results you can paginate you need to inject
+    the `MaxItems` parameter. In the CLI we should be setting this by default
+    to avoid users having to wait exceedingly long times for the full results.
+    """
+    if parsed_globals.paginate and hasattr(parsed_args, 'page_size') and \
+            parsed_args.page_size is None and parsed_args.max_results is None:
```
kyleknap: I feel that the if statement should be:
``if parsed_globals.paginate and getatt(parsed_args, 'page_size', None):``

Why do we need to check for ``max_results`` is ``None``? I feel like this is an important use case where say someone only wanted the first 100 AMIs but we grab all of them and just return the first 100. It would be much better performance if we only grabbed 1000 and then iterated to 100.

tests/unit/customizations/ec2/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap: Surprising we did not have any unit test for the previous ec2 customizations... At least there were functional tests.

kyleknap:
Looks fine. Had a couple of questions though that I would like answered.

tests/unit/customizations/ec2/__init__.py
JordonPhillips: yeah....

awscli/customizations/ec2/paginate.py
JordonPhillips: I'm not sure I understand your objection regarding `max_results`. Max results is the limit key. If the customer specifies `--max-results 100` then EC2 will only return 100 results. We don't want to override anything the customer passes in deliberately.

awscli/handlers.py
JordonPhillips: `set_max_results` checks if the function has pagination by looking for `page_size`, which is only present on functions that both have pagination enabled, and also have a limit key. At that point in the stack, I have no access to the pagination config so that's the only way to check. I could potentially manually compile a list of pagination operations, but then that would break whenever EC2 adds more.

JordonPhillips:
I rebased against develop since #1970 got merged. I also slightly simplified the if statement and clarified some of the logic.

kyleknap:
Hmm looks like you have some failing tests.

JordonPhillips:
Yeah, I think I may have messed up something when I rebased.

awscli/customizations/ec2/paginate.py
JordonPhillips: Ended up not being able to simplify that if statement because the 'exists' check is important to weed out non-paginateable operations.

JordonPhillips:
I ended up not being able to use `getattr(parsed_args, 'page_size', None)` as Kyle suggested because I need to know if page size is present at all, and separately if it is `None`.

tests/unit/customizations/ec2/__init__.py
jamesls: If I had to guess, I think it was before the unit/functional test split where the existing tests were under unit/ but later recategorized as functional.

jamesls:
Change looks good to me.  Looks like this is just pending a clean run through travis.

JordonPhillips:
👍

kyleknap:
Makes sense to me. :ship:

PR #1973, opened 2016-05-11
Overlay streaming-specific config on non-streaming defaults

sgillies:
Without this change, users cannot configure multipart chunk size for streaming uploads.

When running `aws --debug s3 cp` I noticed that configuration of `multipart_chunksize` (as documented at http://docs.aws.amazon.com/cli/latest/topic/s3-config.html) had no effect for streaming uploads in awscli 1.10.26; no matter the value a chunk size of 8MB was always in effect and streaming uploads > 80GB were effectively prohibited. This PR makes larger streaming uploads possible.

### DIFF
--- a/tests/unit/customizations/s3/test_subcommands.py
+++ b/tests/unit/customizations/s3/test_subcommands.py
@@ -377,7 +377,7 @@ class CommandArchitectureTest(BaseAWSCommandParamsTest):
         self.patch_make_request()
         cmd_arc.run()
         output_str = "Completed 1 part(s) with ... file(s) remaining"
-        self.assertIn("nono", self.output.getvalue())
+        self.assertIn(output_str, self.output.getvalue())
 
     def test_error_on_same_line_as_status(self):
         s3_file = 's3://' + 'bucket-does-not-exist' + '/' + 'text1.txt'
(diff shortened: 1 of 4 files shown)


JordonPhillips:
Awesome! Could you please add tests to verify your changes?

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -356,6 +357,28 @@ def test_run_cp_put(self):
...
+        config = RuntimeConfig().build_config(
+            multipart_threshold=1, multipart_chunksize=2)
+        cmd_arc = CommandArchitecture(
+            self.session, 'cp', params, runtime_config=config)
+        cmd_arc.create_instructions()
+        self.patch_make_request()
+        cmd_arc.run()
+        output_str = "Completed 1 part(s) with ... file(s) remaining"
+        self.assertIn(output_str, self.output.getvalue())
+
```
sgillies: @JordonPhillips the output of this test varies depending on whether we're using Python 2 or 3, which suggests to me that it's not a useful test. I'm at a loss to test that a streaming cp is done using the configured chunk size, short of running it against a real AWS bucket and capturing debug logging.

sgillies:
@kyleknap @JordonPhillips does #2064 obsolete this PR and fix the issue?

sgillies:
Confirmed with 1.11.18 that multipart_chunksize can be configured for streaming copies.

PR #1977, opened 2016-05-18
Revert "Default ec2 page size to 1000"

kyleknap:
Reverts aws/aws-cli#1971

Need to revert this because it breaks the basic use case of:
```
aws ec2 describe-instances --instance-id i-ec221738 
```

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #1983, opened 2016-05-19
Allow OpsWorks registration via instance profile

githuesch:


githuesch:
Hold on, we'll try to put together a new example for the docs.

JordonPhillips:
⛵

JordonPhillips:
To capture some discussion we had: the parameter name `--use-instance-profile` is somewhat misleading as it will fall back to the default provider chain so it may not actually use the instance profile. A name like `--use-configured-credentials` or some such may be better.

Additionally, could you please use `scripts/new-change` to generate a changelog entry?

githuesch:
After some debate we're going back to the original proposal here. We made changes on our end, so when passing `--use-instance-profile`, the AWS OpsWorks agent will indeed try to use the instance profile and will fail if that's not sufficient. There's no falling back to other credential providers. In other words, we'd like to get this merged without further changes.

Here's the changelog entry you were asking for.

```
{
  "category": "opsworks",
  "type": "feature",
  "description": "Instead of always creating an IAM user for instance registration, allow using credentials from the instance profile"
}
```

JordonPhillips:
In that case, I think it's all good. The changelog entry should be included in the commit where the changelog script placed it.

JordonPhillips:
⛵

PR #1990, opened 2016-05-26
Update mb.rst

maxirus:
The 's3://' string is in the output as well.


JordonPhillips:
Seems so, thanks for the PR!

PR #1992, opened 2016-05-26
Add docs on addressing_style

jamesls:
The CLI has supported this for a while, but it never made it
to the CLI topic guides.

cc @kyleknap @JordonPhillips


JordonPhillips:
⛵

PR #1994, opened 2016-05-27
Fix regression with --no-paginate argument

### LINKED ISSUE #1993: Using `--no-paginate` with `aws cloudformation describe-stacks` throws error
Version: `aws-cli/1.10.34 Python/2.7.9 Linux/3.16.0-4-amd64 botocore/1.4.24`

When I attempt to run `aws cloudformation describe-stacks --no-paginate --region us-west-2` I receive:

    'Namespace' object has no attribute 'page_size'

This does not happen if I do not use the `--no-paginate` flag.

jamesls:
Not all operations that can paginate have a page_size so we need
to handle the case where that param is not provided.

Fixes #1993.

cc @kyleknap @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-Pagination.json
+++ b/.changes/next-release/bugfix-Pagination.json
@@ -1,0 +1,5 @@
+{
+  "category": "Pagination", 
+  "type": "bugfix", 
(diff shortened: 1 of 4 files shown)


JordonPhillips:
⛵

JordonPhillips:
⛵

PR #1998, opened 2016-06-02
Use default value getattr when checking for paging params

### LINKED ISSUE #1993: Using `--no-paginate` with `aws cloudformation describe-stacks` throws error

JordonPhillips:
When checking to make sure that no paging params are set, we
did not account for the possibility that page size isn't present
for a given operation. Thus, errors would be raised when
`--no-paginate` was supplied on operations without it. This sets
a default value to the getattr call to prevent the error.

Fixes #1993

cc @kyleknap @jamesls


JordonPhillips:
whoops, I'm blind

PR #2002, opened 2016-06-03
Always add uuid to change filename

JordonPhillips:
This further avoids merge conflicts by not having a default
name for a given service change type.

cc @kyleknap @jamesls


kyleknap:
:ship:

PR #2005, opened 2016-06-06
Suppress error output if there is no file association for .py

lorengordon:
Previously, `assoc .py` would output an extraneous error message if there was no file association for .py files: `File association not found for extension .py`.

This commit suppresses the `assoc` error output. With python already in the path, the aws cli executes just fine, even without the file association, so this message just clutters the output.

### DIFF
--- a/bin/aws.cmd
+++ b/bin/aws.cmd
@@ -9,7 +9,7 @@ for %%i in (cmd bat exe) do (
         call :SetPythonExe "%%~$PATH:j"
     )
 )
-for /f "tokens=2 delims==" %%i in ('assoc .py') do (
+for /f "tokens=2 delims==" %%i in ('assoc .py 2^> nul') do (
     for /f "tokens=2 delims==" %%j in ('ftype %%i') do (
         for /f "tokens=1" %%k in ("%%j") do (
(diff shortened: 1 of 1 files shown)


JordonPhillips:
To be clear, does the command still work if it finds no association?

lorengordon:
My use case here is that I do not have admin privs to my workstation, so I cannot install python normally, where it creates the association and adds itself to the path. Instead, I install python to a custom directory in my profile and load it into my path using my powershell profile. I've tried using `assoc` to create the association, but that requires admin privs. :(

This is pretty trivial, and the error message doesn't stop anything from actually executing, so if you want to just close the PR that's fine. It was just bothering me and I figured I'd share my solution with everyone else.

lorengordon:
@JordonPhillips @jamesls going on 9 months here for a one line change. If you don't want to merge it, and can't find time to test it, please feel free to just close it.

dstufft:
Closing due to https://github.com/aws/aws-cli/pull/2562.

cprogrammer1994:
As of 2017-12-18 I have this error:

```
File association not found for extension .py
```

PR #2014, opened 2016-06-10
Use urlopen from compat in gamelift.

JordonPhillips:
I noticed we were manually calling out urlopen rather than taking it
from compat.py. This seems to be the only instance where we're doing
this.

cc @kyleknap @jamesls


kyleknap:
:ship:

PR #2017, opened 2016-06-13
Remove legacy error handler

jamesls:
This PR cleans up some technical debt and removes the error legacy handling code used in the CLI.

The background on this:

In the pre-client era, you were always returned a response, even in the case of errors.  You then had to check if the response was an error response and take appropriate action.  To simplify some of the handlers, we wrote an error handler that would just raise an exception if it saw an error response.

This is no longer necessary in client code.  Clients will automatically raise an exception on errors.  However we never went back and removed the old error handler.  This code consists of primarily two changes:

* Converting the old exceptions to the new client error interface.
* Removing any code that unregistered the event handler.  This was how newer customizations could "work around" the technical debt and get real client errors.  We no longer need that.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/customizations/cloudtrail/subscribe.py
+++ b/awscli/customizations/cloudtrail/subscribe.py
@@ -14,7 +14,7 @@ import json
 import logging
 import sys
 
-from .utils import get_account_id, remove_cli_error_event
+from .utils import get_account_id
 from awscli.customizations.commands import BasicCommand
 from awscli.customizations.utils import s3_bucket_exists
 from botocore.exceptions import ClientError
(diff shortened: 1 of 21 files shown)


kyleknap:
Code changes look fine. What was your testing approach for some the ones that did not have great testing suites like CodeDeploy and OpsWorks? Did you try any of them manually?

jamesls:
Sort of manually.  Some of the customizations required preconditions before making the client calls in a try/except.  So I commented out the various pieces of code so it could make it to try/except client calls.  I then verified that it caught and handled the exception as expected.

kyleknap:
Sounds good. :ship:

JordonPhillips:
⛵

PR #2029, opened 2016-06-18
Remove enforcement of failure without region in sigv4

JordonPhillips:
This test will soon no longer be valid, as we are enabling
redirection for s3.

cc @jamesls @kyleknap


jamesls:
:shipit:

Issue #1742, opened 2016-01-22
Update colorama version


Fale:
It seems like the current aws-cli requires colorama >=0.2.5 and <=0.3.3, but the current colorama stable version in 0.3.6.
From tickets #1698 and #1700 I guess some fix will have to be executed before it's possible to update

PS: From colorama [changelog](https://github.com/tartley/colorama/blob/master/CHANGELOG.rst), it seems like from v0.3.3 they just fixed bugs, so it should not create any problem for the update (unless awscli was using a colorama bug as if it was a feature)

Fale:
Any update on this? In the mean time, colorama 0.3.7 has been released

Knio:
Hi,

Can you please mention what release on PyPI has this fix?

Fale:
@Knio not really sure which fix you are speaking about. If you mean accepting colorama <= 0.3.7, I guess 1.11.28, since https://github.com/aws/aws-cli/commit/e28212cfde9525781d9c5d2bc548ba4d03242ce1 was  part of that release

Knio:
That's what I was looking for, thanks!

Issue #47, opened 2013-02-09
For autoscalling terminate-instance-in-auto-scaling-group --should-decrement-desired-capacity should be optional


sjones4:
There should be some way to use the terminate-instance-in-auto-scaling-group operation without decrementing the desired capacity of a group.

$ aws autoscaling terminate-instance-in-auto-scaling-group --instance-id i-C60243E1 
usage: aws autoscaling terminate-instance-in-auto-scaling-group
       --instance-id INSTANCE_ID --should-decrement-desired-capacity
aws autoscaling terminate-instance-in-auto-scaling-group: error: argument --should-decrement-desired-capacity is required

$ aws autoscaling terminate-instance-in-auto-scaling-group --instance-id i-C60243E1 --should-decrement-desired-capacity false
Something is wrong.  We have leftover options
['false']

cewood:
Sorry to revisit such an old issue, but is this fix still valid? I'm currently experiencing this same issue with the following version of the cli: aws-cli/1.10.63 Python/2.7.12 Darwin/15.6.0 botocore/1.4.53.

> aws --region eu-west-1 autoscaling terminate-instance-in-auto-scaling-group --instance-id someinstancehere --should-decrement-desired-capacity
> 
> An error occurred (ValidationError) when calling the TerminateInstanceInAutoScalingGroup operation: Currently, desiredSize equals minSize (4). Terminating instance without replacement will violate group's min size constraint. Either set shouldDecrementDesiredCapacity flag to false or lower group's min size.

zerolaser:
@garnaat I got the following error when trying to terminate an instance in asg which has min size = desired capacity. Should it also have an option of decreasing the min size as well?
>Currently, desiredSize equals minSize. Terminating instance without replacement will violate >group's min size constraint. Either set shouldDecrementDesiredCapacity flag to false or lower >group's min size

PR #2037, opened 2016-06-26
Bump colorama version to <=0.3.7

c22:
Implements #1742 
Colorama differences between 0.3.3 and 0.3.7 are mostly bug fixes.
For what it's worth, I have been using colorama 0.3.7 with aws-cli for a while now without problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ docutils>=0.10
 -e git://github.com/boto/s3transfer.git@develop#egg=s3transfer
 -e git://github.com/boto/jmespath.git@develop#egg=jmespath
 nose==1.3.0
-colorama>=0.2.5,<=0.3.3
+colorama>=0.2.5,<=0.3.7
 mock==1.3.0
 rsa>=3.1.2,<=3.5.0
 wheel==0.24.0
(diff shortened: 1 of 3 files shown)


JordonPhillips:
Has the issue mentioned [here](https://github.com/aws/aws-cli/pull/1698) been resolved?

c22:
That behavior was a bug in colorama which was fixed in version 0.3.4

> Fix issue [#47](https://github.com/tartley/colorama/issues/47) and [#80](https://github.com/tartley/colorama/issues/80) - stream redirection now strips ANSI codes on Linux

In other words, it was never supposed to work that way.

c22:
Fixing the "issue" mentioned in #1698 is not insurmountable, it would require a few more changes, but IMO should not hold back the version bump and should be addressed as a feature request rather than a bug.

c22:
Initial investigations into allowing colorized output on a pipe when CLI is invoked "--color on" leads me to believe it could create issues on Win32 without doing some more refactoring.

I'm still in favor of the version bump, the minimum version is not changing. Users who want the old behavior can still use 0.3.3 and user of new versions of colorama who want ANSI characters preserved on non-TTY output can use a utility like unbuffer as a workaround:
`unbuffer aws ec2 describe-instances --color on --output table | less -r`

c22:
Thanks!

PR #2049, opened 2016-06-30
Use ClientError in EMR customizations

JordonPhillips:
When we removed the CLI error handling, we missed this case of
using ClientError since it was catching "Exception". This switches
to catching ClientError directly.

In the future, I think this entire command and its tests will need
to be refactored.

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/emr/createdefaultroles.py
+++ b/awscli/customizations/emr/createdefaultroles.py
@@ -148,7 +148,7 @@ class CreateDefaultRoles(Command):
 
         # Check if the default EC2 Role for EMR exists.
         role_name = EC2_ROLE_NAME
-        if self._check_if_role_exists(role_name, parsed_globals):
+        if self.check_if_role_exists(role_name, parsed_globals):
             LOG.debug('Role ' + role_name + ' exists.')
         else:
             LOG.debug('Role ' + role_name + ' does not exist.'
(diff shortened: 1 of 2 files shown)


awscli/customizations/emr/createdefaultroles.py
```diff
@@ -215,14 +215,15 @@ def _construct_role_and_role_policy_structure(
             list.append({'Role': response['Role'], 'RolePolicy': policy})
             return list
 
-    def _check_if_role_exists(self, role_name, parsed_globals):
```
jamesls: Why are we making this public?  If it's not intended to be called from other classes I don't think we should be changing internal vs. public for a bug fix.

JordonPhillips: The issue is that the command is exceedingly difficult to test. It needs to have some significant refactoring. At the moment though, we have a broken command.

kyleknap:
:ship: I would rather see this be exposed as a public method. It was bad that an internal method was being tested in the first place.

awscli/customizations/emr/createdefaultroles.py
jamesls: Just wanted to follow up here.  I see the change has been merged and I don't think we need to do anything else here, but I do want to make sure I clarify my thoughts for future changes we may need to make.  I realize I wasn't entirely clear on what I meant.

What I was trying to communicate earlier was that we should only make a method public because we want that to be part of the public interface, not because it makes our tests more "correct" because we're not testing internal methods.  In general for these types of changes I would prefer to just accept the fact that we're testing internal methods and not make it public.  Until such time where we consider rewriting it to be better structured. Consider:

* It's less code (and less risk) than having to find/replace the method name and ensure we change `_foo` to `foo`.  For these types of changes I think as minimal a change as possible is ideal.
* The method really is intended to be internal (i.e if we created another class in this module that was going to use this class, I wouldn't want it calling this internal-but-now-public method).

Yes, avoiding testing internal methods is ideal, but I don't think artificially changing public visibility so we can check the "not-testing-internal-methods" box seems correct.

Hope that helps clarify things going forward.

PR #2052, opened 2016-07-06
Don't use shared test bucket for website tests

jamesls:
This will produce false failures if the index_and_error
test runs before the index test.  To fix this is issue, I've
put back the old behavior for this test which is to create a bucket
per each of these tests.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #2053, opened 2016-07-06
Update to EFS GA Announcement

awsdan:
EFS is now Generally Available in 3 regions. This update is to reflect that in the language of the changelog.


kyleknap:
Closing out pull request as it was sent to the master branch. Furthermore, modifying the CHANGELOG will not help as we rely on the JSON files in .changes to generate the changelog. So if you want to update the changelog you would have to do it in the appropriate one there.

PR #2064, opened 2016-07-12
Port streaming copies to S3Transfer

JordonPhillips:
This will depend on the latest version of S3Transfer that supports uploading and downloading non-seekable file-like objects.

cc @kyleknap @jamesls


awscli/customizations/s3/s3handler.py
```diff
@@ -367,16 +380,14 @@ def _enqueue_upload_end_task(self, filename, upload_context):
...
     """
     This class is an alternative ``S3Handler`` to be used when the operation
     involves a stream since the logic is different when uploading and
     downloading streams.
     """
-    # This ensures that the number of multipart chunks waiting in the
-    # executor queue and in the threads is limited.
-    MAX_EXECUTOR_QUEUE_SIZE = 2
-    EXECUTOR_NUM_THREADS = 6
+    MAX_QUEUE_SIZE = 2
```
kyleknap: I do not think we should be restraining the transfer manager to these numbers as we now properly handle memory usage with the ``max_in_memory_upload_chunks``. I think we should just keep using the normal RuntimeConfig. We even have a PR open about this: https://github.com/aws/aws-cli/pull/1973.

awscli/customizations/s3/s3handler.py
```diff
@@ -385,151 +396,112 @@ def __init__(self, session, params, result_queue=None,
...
             # chunksize based on max parts and chunksize limits
-            chunksize = find_chunksize(int(self.params['expected_size']),
-                                       self.chunksize)
+            chunksize = find_chunksize(
+                int(self.params['expected_size']),
+                self.config.multipart_chunksize)
         else:
             # Otherwise, we can still adjust for chunksize limits
-            chunksize = adjust_chunksize_to_upload_limits(self.chunksize)
+            chunksize = adjust_chunksize_to_upload_limits(
```
kyleknap: Hmm we need to fix this in ``s3transfer``because we do not readjust the chunksize. It might be good to put a TODO to make sure we remove this once we we port it to s3transfer.

awscli/customizations/s3/s3handler.py
```diff
@@ -385,151 +396,112 @@ def __init__(self, session, params, result_queue=None,
...
-        # was needed.
-        self._enqueue_upload_single_part_task(
-            part_number=1, chunk_size=chunksize,
-            upload_context=upload_context, filename=filename,
-            task_class=tasks.UploadPartTask, payload=payload
-        )
+        # Set the size of the transfer if expected size is set.
+        expected_size = self.params.get('expected_size', None)
+        if expected_size is not None:
+            future.meta.size = expected_size
```
kyleknap: You should not be able to do that...https://github.com/boto/s3transfer/blob/develop/s3transfer/futures.py#L92 You need to use a subscriber and the ``provide_transfer_size`` method in the ``on_queued()`` method.

awscli/customizations/s3/s3handler.py
```diff
@@ -385,151 +396,112 @@ def __init__(self, session, params, result_queue=None,
...
-                break
-        # Once there is no more data left, announce to the context how
-        # many parts are being uploaded so it knows when it can quit.
-        upload_context.announce_total_parts(num_uploads)
-        return num_uploads
+        :return: A CommandResult representing the transfer status.
+        """
+        try:
+            future.result()
+            return CommandResult(0, 0)
```
kyleknap: Will there ever be an instance where the task gets warned?

awscli/customizations/s3/utils.py
```diff
@@ -406,19 +406,39 @@ def uni_print(statement, out_file=None):
     out_file.flush()
 
 
-def bytes_print(statement):
+def bytes_print(statement, stdout=None):
```
kyleknap: Why not make a ``fileout`` param with the default to ``sys.stdout``?

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -903,190 +904,124 @@ def test_remove_bucket(self):
...
+        expected_size = 6
+        self.params['expected_size'] = expected_size
+        handler = S3TransferStreamHandler(self.session, self.params)
+        file = FileInfo('-', 'foo-bucket/bar.txt', is_stream=True,
+                        operation_name='upload')
+
+        with MockStdIn(b'foobar'):
+            handler.call([file], self.transfer_manager)
+
+        self.assertEqual(self.transfer_future.meta.size, expected_size)
```
kyleknap: We should have a functional or integration test for this because this should not have passed.

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -903,190 +904,124 @@ def test_remove_bucket(self):
...
+        config = runtime_config(multipart_chunksize=1)
+        handler = S3TransferStreamHandler(
+            self.session, self.params, runtime_config=config)
+        file = FileInfo('-', 'foo-bucket/bar.txt', is_stream=True,
+                        operation_name='upload')
+
+        with MockStdIn(b'foobar'):
+            handler.call([file], self.transfer_manager)
+
+        self.assertTrue(
```
kyleknap: Helper methods for asserting that the chunksize is greater than the minimum and less than the maximum would be great here so that is more explicit. That took me some time to figure out that is what you were doing.

kyleknap:
Looks good. Approach looks fine. My main comments were about some edge cases and use cases.

awscli/customizations/s3/s3handler.py
JordonPhillips: Glacier status is never looked at.

awscli/customizations/s3/utils.py
JordonPhillips: This function deals specifically with stdout. It really wouldn't work using a different file object due to the lack of a `buffer` property.

tests/unit/customizations/s3/test_s3handler.py
JordonPhillips: In S3Transfer you mean? I don't think it makes sense to test functionality of a separate library.

kyleknap: No I mean a functional test for ``--expected-size``. To make sure that straight up works.

awscli/customizations/s3/s3handler.py
kyleknap: Sounds good.

awscli/customizations/s3/utils.py
kyleknap: I guess what I was meaning why not make sys.stdout the default then? You have the:
```py
if stdout is None:
    stdout = sys.stdout
```
and sys.stdout is not mutable so it should be fine as:
```py
def bytes_print(statement, stdout=sys.stdout):
```

awscli/customizations/s3/s3handler.py
```diff
@@ -367,169 +381,125 @@ def _enqueue_upload_end_task(self, filename, upload_context):
...
-                max_concurrent_requests=self.EXECUTOR_NUM_THREADS)
-        super(S3StreamHandler, self).__init__(session, params, result_queue,
-                                              runtime_config)
+        super(S3TransferStreamHandler, self).__init__(
+            session, params, result_queue, runtime_config)
+        self.config = create_transfer_config_from_runtime_config(
+            self._runtime_config)
+
+        # Restrict the maximum chunks to 1 per thread.
+        self.config.max_in_memory_upload_chunks = \
```
kyleknap: I am not really sure we would want to do this either. If they had something like 50 for max_request_concurrency, then memory usage could be really large. I realize we would not be utilizing all of the threads, but this should be faster anyways as we are no longer locking to 6 threads everytime and we should be protecting the user from too much memory usage as it is not completely explicit that increasing the number of threads would linearly increase the maximum memory usage by the chunksize.  I much rather expose this as a config variable down the road and keep the default (which is 10 right now).

awscli/customizations/s3/s3handler.py
```diff
@@ -367,169 +381,125 @@ def _enqueue_upload_end_task(self, filename, upload_context):
...
-        requested or there is no more left to pull in from stdin.  The
-        function wraps the data into a ``BytesIO`` object that is returned
-        along with a boolean telling whether the amount requested is
-        the amount returned.
+        params = {}
+        RequestParamsMapper.map_get_object_params(params, self.params)
+
+        future = manager.download(
+            fileobj=StdoutBytesWriter(), bucket=bucket,
+            key=key, extra_args=params)
```
kyleknap: Oh I just forgot, you may want to use the size subscriber here as well because the download will do a head object if a size is not provided and we already did a ``HeadObject`` in the ``FileGenerator`` I believe

awscli/customizations/s3/s3handler.py
JordonPhillips: One chunk per thread is what we are currently testing for in our integration tests, using the default fails that test even with the runtime margin. Are we sure we want to increase memory usage? I'm fine for it, but I think it warrants discussion.

awscli/customizations/s3/s3handler.py
JordonPhillips: I think you have to call `set_size_from_s3` to do that, which I'm not calling. I'll double check to make sure it isn't called elsewhere.

awscli/customizations/s3/s3handler.py
kyleknap: I am really confused as to why that test fails for one over the other then. Because if we are matching the number of threads (which is 10), the number of in memory chunks should be 10. However if we uncouple the two, the number of in memory chunks is still 10... Why is that? Or am I off in my logic somewhere?

awscli/customizations/s3/s3handler.py
kyleknap: Ok cool. I forgot we completely skirt the FileGenerator. I think we should be good here.

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -903,190 +906,152 @@ def test_remove_bucket(self):
...
+        subscribers = call_args.get('subscribers', [])
+        self.assertTrue(len(subscribers) == 1)
+
+        # Make sure that subscriber is the right kind
+        subscriber = subscribers[0]
+        self.assertIsInstance(subscriber, ProvideSizeSubscriber)
+
+        # Validate that the size on the subscriber is the expected size
+        self.assertEqual(subscriber.size, expected_size)
+
```
kyleknap: Maybe also assert that the ``future.meta.provide_transfer_size()`` got called with the proper value?

awscli/customizations/s3/s3handler.py
JordonPhillips: Right, so the file generator isn't used in the streaming transfers: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/subcommands.py#L955

Previously, the stream handler was calling the function I mentioned: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/s3handler.py#L408

kyleknap:
The only concern is about the max_in_memory_upload_chunks we should be using. Once we figure that out, we should be good.

awscli/customizations/s3/s3handler.py
kyleknap: Cool. Another thing that I was a little worried about is that since we combine head object and get object on the download(), I noticed we only map the get_object_params, but since those map_head_object and map_get_object are essentially the same we should be fine and not be leaving out any arguments.

It may be good to make a note about this though so we do not forget or map both of them anyways.

tests/unit/customizations/s3/test_s3handler.py
JordonPhillips: It wouldn't be, since we aren't actually calling out to `s3transfer`. The tests for `ProvideSizeSubscriber` cover that.

kyleknap: True. How hard would a functional or integration test be for this? I feel it may be good to have a quick test such that we make sure we have everything wired up correctly.

awscli/customizations/s3/s3handler.py
JordonPhillips: I'll investigate some more. For reference, [here](https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L1374) is the offending test.

kyleknap: Thanks!

kyleknap:
The recent commits look good to me.

awscli/customizations/s3/s3handler.py
```diff
@@ -367,169 +381,128 @@ def _enqueue_upload_end_task(self, filename, upload_context):
...
+        super(S3TransferStreamHandler, self).__init__(
+            session, params, result_queue, runtime_config)
+        self.config = create_transfer_config_from_runtime_config(
+            self._runtime_config)
+
+        # Restrict the maximum chunks to 1 per thread.
+        self.config.max_in_memory_upload_chunks = \
+            self.MAX_IN_MEMORY_CHUNKS
+
+    def call(self, files, manager=None):
```
jamesls: I would prefer to keep the same interface as `BaseS3Handler` and instead have the manager be passed in during the `__init__` if someone wants to override the manager (i.e in your tests).

tests/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from mock import Mock, patch
+
+from awscli.compat import six
+
+
+class MockStdIn(object):
```
jamesls: Testing helpers go in `awscli/testutils.py`.  Please move this to that file.  There's also a few existing helpers such as `capture_stdout()` that you might be able to refactor to share code.  They do a similar thing except with output.

awscli/customizations/s3/s3handler.py
```diff
@@ -367,169 +381,128 @@ def _enqueue_upload_end_task(self, filename, upload_context):
...
+            # TODO: remove when this happens in s3transfer
             # Otherwise, we can still adjust for chunksize limits
-            chunksize = adjust_chunksize_to_upload_limits(self.chunksize)
+            chunksize = adjust_chunksize_to_upload_limits(
+                self.config.multipart_chunksize)
+        self.config.multipart_chunksize = chunksize
 
-        num_uploads = '...'
+        stream_filein = sys.stdin
+        if six.PY3:
```
jamesls: Any py2/py3 differences should be moved over to `compat.py` so we don't have version conditionals throughout the codebase.  There's also I think `BinaryStdout` so perhaps we can consolidate code and reuse/refactor existing code to work for both input/output.

awscli/customizations/s3/utils.py
```diff
@@ -406,19 +407,39 @@ def uni_print(statement, out_file=None):
...
     """
     This function is used to properly write bytes to standard out.
     """
+    if stdout is None:
+        stdout = sys.stdout
+
     if PY3:
```
jamesls: It would be nice to clean up some of this technical debt if we're in the code anyways, and move the py2/py3 differences over to compat.py.

jamesls:
Looks like the error messages aren't being displayed in some cases.  Comparing this PR to a prior version of the CLI:

```
# v 1.10.0
$ echo 'hello world' | aws-1.10.0 s3 cp - s3://bad-bucket/foo
upload failed: ./- to s3://bad-bucket/foo A client error (NoSuchBucket) occurred when calling the PutObject operation: The specified bucket does not exist

# This PR
$ echo 'hello world' | aws s3 cp - s3://bad-bucket/foo
$

File: awscli/customizations/s3/s3handler.py (hunk 11 above)
Comment (JordonPhillips) [inline awscli/customizations/s3/s3handler.py]: I did that because `call` otherwise needs to create the manager itself. I guess since that's only used in testing it doesn't really matter though.

File: awscli/customizations/s3/s3handler.py (hunk 13 above)
Comment (JordonPhillips) [inline awscli/customizations/s3/s3handler.py]: So the `BinaryStdout` is a bit of a different beast intended to get around line-ending translation issues that arise with git. I'm not sure if it's super necessary to integrate on parts where the line ending is irrelevant.

**Comment by JordonPhillips** [comment]
@jamesls Not sure about the error messages not displaying. I'm adding a `PrintTask` to the result queue :\

**Comment by JordonPhillips** [comment]
Ah, nevermind, I got it

**Comment by JordonPhillips** [comment]
Nothing is actually reading from the result queue

File: awscli/compat.py [hunk 15]
```diff
@@ -48,7 +55,25 @@ def __enter__(self):
...
+
+
+def bytes_print(statement, stdout=None):
+    """
+    This function is used to write raw bytes to stdout.
+    """
+    if stdout is None:
+        stdout = sys.stdout
+
+    if PY3:
```
jamesls: In the future, please try to avoid `PY3` conditionals within a function like this.  The prevailing pattern in this module is to separate python2 vs. python3 code so there's specific versions of a function based on python version.  Checking whether or not we're in python3 is not something that needs to happen every time the function is called, it only needs to happen once.

awscli/compat.py
```diff
@@ -87,6 +114,8 @@ def compat_open(filename, mode='r', encoding=None):
 
     raw_input = raw_input
 
+    binary_stdin = sys.stdin
```
jamesls: Is there a reason we're not being consistent and having context managers for `BinaryStdin`, simliar to `BinaryStdout`?  Also, is the `msvcrt.setmode` not needed for binary stdin?  We use it for binary stdout on windows, so I would expect we'd need something similar for stdin.

What did you find when you were testing this?

awscli/customizations/s3/s3handler.py
kyleknap: Yeah there is not really any overlap with ``BinaryStdout``. I think the approach you went with is appropriate.

jamesls:
Overall looks fine.  Not using the `result_queue` for streams seems reasonable during this transition period.

I did have a question about `BinaryStdin`, which is the only remaining thing I'm curious about.

awscli/compat.py
kyleknap: Hmmm we only use this for the codecommit plugin: https://github.com/aws/aws-cli/commit/9f96039f4f24512c2375ff297f4f0d0e76ae8371.

It looks like we do not use it at all for the s3 commands, even for piping to standard out. Do you think we should start using it for both standard in and out? The integration test are not skipped on windows so I feel like we should be fine here as the content is checked for correctness: https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L1632

kyleknap:
I am good with all of the updates. But do you think we should add a functional test for errors on streams to make sure it gets printed out?

awscli/compat.py
jamesls: >  Do you think we should start using it for both standard in and out? 

I don't think we need to migrate everything over to `BinaryStdout` right now, but as we're consolidating things over to `compat.py`, I don't see any reason to have different ways of doing binary IO depending on if you're using output vs. input.  We should at least have a consistent way of doing things.

I see this as an opportunity to clean up technical debt.

JordonPhillips: The `BinaryStdout` context manager only exists to protect from line-ending translations that mess up the git commands that get printed out by codecommit. I think for data that we only want to present it isn't necessary (and may be incorrect). I'm not crazy about adding additional system calls that we don't need.

kyleknap: I guess for me, while I think trying to combine all of these is a good idea to reduce technical debt as it is strange that we have two different ways we print binary to standard out. I think I would prefer to minimize the amount of moving parts in this change as we have never relied on changing the mode of stdout or stdin on windows before with streaming s3 commands. I want to avoid as many regressions as possible since this change is quite a big change already.

Can we settle on adding a big todo note on combining/unifying these? I plan once ``s3transfer`` is in place to go through and just get rid of anything that may be technical debt because I have a feeling there will be a fair amount of technical debt left and the s3 codebase is already due for a cleanup.

JordonPhillips: I really don't think it's necessary to add that context manager to `bytes_print`. I tested it out on my windows machine and it appears to have no noticeable effect on the output unless you're trying to write a git command. We could add it, and I don't think it would break things, but it also doesn't help anything.

jamesls: @JordonPhillips just to clarify here, my original comment was about `binary_stdin`.

> Can we settle on adding a big todo note on combining/unifying these? 

@kyleknap as I mentioned in my first comment, I'm not suggesting using `BinaryStdout` right now.  I'm saying, if we're taking the effort to move things over to `compat.py` in an effort to clean things up, we should at least have a consistent API.

Having a consistent API is independent of how the API is implemented.  We just shouldn't have two APIs for doing similar things.  We have one way to do binary input/output period.

JordonPhillips:
This should be good to look at again.

awscli/customizations/s3/s3handler.py
```diff
@@ -400,6 +400,7 @@ def __init__(self, session, params, result_queue=None,
         # Restrict the maximum chunks to 1 per thread.
         self.config.max_in_memory_upload_chunks = \
             self.MAX_IN_MEMORY_CHUNKS
+        self.config.max_in_memory_upload_chunks
```
kyleknap: You are not setting anything here and you need to use ``config.max_in_memory_download_chunks``

JordonPhillips: foiled again by dot completion

JordonPhillips:
I rebased to get the updated s3transfer

kyleknap:
Thanks! :ship:

jamesls:
:shipit:  Thanks for updating.

JordonPhillips:
squashed some commits, will merge once tests pass

PR #2066, opened 2016-07-12
Review preview status from EFS

jamesls:
https://aws.amazon.com/about-aws/whats-new/2016/06/amazon-elastic-file-system-efs-is-now-generally-available/

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #2093, opened 2016-08-09
Add warning about streaming in PowerShell

JordonPhillips:
For those that are not intimately familiar with PowerShell, it may come as a
surprise when piping or redirecing results in data that isn't identical to
what they have stored in S3 or what they thought they stored. This adds a
warning to the CP streaming docs to let users know what is happening.

In more detail, what is happening is that the results become a .Net string
object, whose default encoding is a 16-bit encoding.

cc @kyleknap @jamesls

### DIFF
--- a/awscli/examples/s3/cp.rst
+++ b/awscli/examples/s3/cp.rst
@@ -140,6 +140,8 @@ Output::
 
 **Uploading a local file stream to S3**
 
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:  I was going to say we should just add this to our general docs on streaming input/output for S3, then realized we don't actually have any docs.  Probably worth adding documentation in the future.

PR #2100, opened 2016-08-11
Reintroduce streams port to s3transfer

jamesls:
This pulls in the original streams pull request and fixes the issue of hanging streaming uploads on windows.  The only new commit is https://github.com/aws/aws-cli/commit/2c483a12c2ccc1f5dd110f67f5f5c30ce8153d72 :

>Wrap sys.stdin to force non seekable behavior
>
>S3Transfer changes its upload logic based on whether or not
>a fileobj is seekable.
>
>On windows, seekable(sys.stdin) was returning True despite the
>fact that sys.stdin does not provide correct seekable behavior.
>
>This would result in streaming uploads hanging.
>
>The fix here is to ensure that seekable(fileobj) returns False
>for sys.stdin.  To accomplish this, sys.stdin is wrapped in
>an object that does not have a seek() method.  This ensures that
>S3Transfer will choose the correct code path, and streaming uploads
>on windows will not hang.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -239,6 +239,8 @@ class CLIDriver(object):
                                            format_string=LOG_FORMAT)
             self.session.set_stream_logger('awscli', logging.DEBUG,
                                            format_string=LOG_FORMAT)
+            self.session.set_stream_logger('s3transfer', logging.DEBUG,
+                                           format_string=LOG_FORMAT)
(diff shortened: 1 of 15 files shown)


kyleknap:
Awesome. Thanks for looking into this. :ship: assuming integration tests are fine.

JordonPhillips:
Thanks for looking into the issue! ⛵

PR #2101, opened 2016-08-11
Add results module and subscribers to add results

kyleknap:
So in integrating ``s3transfer`` with the CLI, one of the larger integration work we need done is the integration of subscribers, really the only way to get status on transfers, with the progress tracking and printing of the CLI.

This is just an initial part that implements the subscribers needed to produce namedtuple results that will be processed later by the components that track and print transfer progress. But in the end, I am trying to completely remove the current code base of ``executor.PrintThread()`` to make a more expandable, testable, and maintainable version of the general implementation.

So my implementation plan is as follows:

Introduce the following abstractions:

1) ``Results`` --> namedtuples that represent progress updates generated by subcriber callbacks
2) ``ResultSubscriber`` --> Subscribers that are invoked on the various status callbacks of ``s3transfer``. The subscriber will create the ``Results`` and then place it on a result queue to be consumed.
3) ``ResultProcessor`` --> A separate thread that pulls results off of the result queue. It will have components injected into it based on various parameters provided (i.e. ones that keep track of statistics and printing. By having the single thread queue consumer model, we do not have to worry about locking if the necessary components all live inside the single thread. The processor will pass the consumed result to each necessary component that was injected into it
4) ``ResultRecorder`` --> A component that is injected into the result processor that records results that was passed to it. It will record stuff like total bytes and files transferred for now.
5) ``ResultPrinter`` --> A component that is injected into the result processor that will print progress based on the result passed to it and the current numbers from the result recorder.

Here is a rough diagram of how this would look:
```
------------------------------------------------------------------------
ResultSubscribers (using on_queued, on_progress, on_done callbacks)
------------------------------------------------------------------------
    |
    |
    |
put Result in ResultQueue 
     |
---------------
ResultQueue
---------------
    |
    |
 get Result from ResultQueue
    |
---------------------------------------
ResultProcessor(threading.Thread)
    |
passes Result on
    |
    |
* component: ResultRecord (records result)
    |
passes Result on
    |
    |
* component: ResultPrinter (prints progress)
    |
    |
    passes Result onto any other components we may want to add...
----------------------------------------
```

Let me know what you think of this direction. Just wanted to give you a higher level picture of my plan to help with the review and not just put down a massive PR with all of the components. 

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -1,0 +1,124 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import logging
+from collections import namedtuple
+
+from s3transfer.subscribers import BaseSubscriber
+
+from awscli.customizations.s3.utils import relative_path
+
+
+LOGGER = logging.getLogger(__name__)
+
+QueuedResult = namedtuple(
+    'QueuedResult',
+    ['transfer_type', 'src', 'dest', 'total_transfer_size']
+)
+ProgressResult = namedtuple(
+    'ProgressResult',
+    ['transfer_type', 'src', 'dest', 'bytes_transferred',
+     'total_transfer_size']
+)
+SuccessResult = namedtuple(
+    'SuccessResult', ['transfer_type', 'src', 'dest']
+)
+FailureResult = namedtuple(
+    'FailureResult', ['transfer_type', 'src', 'dest', 'exception']
+)
+CommandResult = namedtuple(
+    'CommandResult', ['num_tasks_failed', 'num_tasks_warned'])
+
+
+class BaseResultSubscriber(BaseSubscriber):
+    TRANSFER_TYPE = None
+
+    def __init__(self, result_queue):
+        """Subscriber to send result notifications during transfer process
+
+        :param result_queue: The queue to place results to be processed later
+            on.
+        """
+        self._result_queue = result_queue
+        self._transfer_type = None
+
+    def on_queued(self, future, **kwargs):
+        src, dest = self._get_src_dest(future)
+        queued_result = QueuedResult(
+            self.TRANSFER_TYPE, src, dest, future.meta.size)
+        self._result_queue.put(queued_result)
+
+    def on_progress(self, future, bytes_transferred, **kwargs):
+        src, dest = self._get_src_dest(future)
+        progress_result = ProgressResult(
+            self.TRANSFER_TYPE, src, dest, bytes_transferred, future.meta.size)
+        self._result_queue.put(progress_result)
+
+    def on_done(self, future, **kwargs):
+        src, dest = self._get_src_dest(future)
+        try:
+            future.result()
+            self._result_queue.put(
+                SuccessResult(self.TRANSFER_TYPE, src, dest))
+        except Exception as e:
+            self._result_queue.put(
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/results.py
```diff
@@ -0,0 +1,124 @@
...
+        self._result_queue = result_queue
+        self._transfer_type = None
+
+    def on_queued(self, future, **kwargs):
+        src, dest = self._get_src_dest(future)
+        queued_result = QueuedResult(
+            self.TRANSFER_TYPE, src, dest, future.meta.size)
+        self._result_queue.put(queued_result)
+
+    def on_progress(self, future, bytes_transferred, **kwargs):
```
jamesls: This is new behavior right?  In the sense that we're now queueing much more granular progress events that we were previously queueing?  This should mean we can give much more detailed progress right?

awscli/customizations/s3/results.py
```diff
@@ -0,0 +1,124 @@
...
+        try:
+            future.result()
+            self._result_queue.put(
+                SuccessResult(self.TRANSFER_TYPE, src, dest))
+        except Exception as e:
+            self._result_queue.put(
+                FailureResult(self.TRANSFER_TYPE, src, dest, e))
+
+    def _get_src_dest(self, future):
+        raise NotImplementedError('must implement _get_src_dest()')
```
jamesls: We generally just use the method name in the msg.  The "must implement" is implied in the exception name.

tests/unit/customizations/s3/test_results.py
```diff
@@ -0,0 +1,204 @@
...
+        self.src = relative_path(self.filename)
+        self.dest = 's3://' + self.bucket + '/' + self.key
+        self.transfer_type = 'upload'
+        self.result_subscriber = UploadResultSubscriber(self.result_queue)
+
+    def test_on_queued(self):
+        self.result_subscriber.on_queued(self.future)
+        result = self.get_queued_result()
+        self.assert_result_queue_is_empty()
+        self.assert_correct_queued_result(result)
```
jamesls: I think we've chatted about this before, but I prefer the test logic is actually embedded in the test.  What specifically about on queued are you testing?  That should be apparent in the test.  Right now I have no idea what constitutes as "assert_correct_queued_result" should even mean.

These are named tuples, why not just compare them to the expected result?

awscli/customizations/s3/results.py
kyleknap: Yep it is new. And definitely we can make more detailed progress. For the ResultPrinter since there is no notion of "parts", I am planning to print progress in terms of human readable bytes.

jamesls:
Overall I like this direction.  I think breaking this down by class makes the logic much easier to follow.

I had a small comment about the test.  I'd prefer to see them much more direct.  Something like:

```
self.result_subscriber.on_queued(self.future)
result = self.get_queued_result()
self.assertEqual(result, QueuedResult(self.transfer_type, self.src, self.dest, ...))
```

Otherwise looks good to me.

awscli/customizations/s3/results.py
kyleknap: Makes sense. Was just use to what I was doing with ``s3transfer``. Suggestion is consistent in CLI codebase.

tests/unit/customizations/s3/test_results.py
kyleknap: Yeah I definitely could just call ``assertEqual``. I forgot that these were namedtuples where you can actually assert they are equal as opposed to something like the ``FileInfo`` objects.

kyleknap:
@jamesls I updated based on feedback. Should be good to look at again.

awscli/customizations/s3/results.py
```diff
@@ -0,0 +1,124 @@
...
+    TRANSFER_TYPE = None
+
+    def __init__(self, result_queue):
+        """Subscriber to send result notifications during transfer process
+
+        :param result_queue: The queue to place results to be processed later
+            on.
+        """
+        self._result_queue = result_queue
+        self._transfer_type = None
```
JordonPhillips: This is never used

JordonPhillips:
One unused variable, but ⛵  once it's removed.

awscli/customizations/s3/results.py
kyleknap: Right. Was an artifact of refactoring.

jamesls:
:shipit:  Thanks.

PR #2103, opened 2016-08-12
Add integration tests for describing EC2 instances

JordonPhillips:
This is to prevent any additional regressions in this common operation.

cc @kyleknap @jamesls

### DIFF
--- a/tests/integration/test_ec2.py
+++ b/tests/integration/test_ec2.py
@@ -10,8 +10,6 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-import botocore.session
-
 from awscli.testutils import unittest, aws
 
 
--- a/tests/integration/test_ec2.py
+++ b/tests/integration/test_ec2.py
@@ -1,0 +1,65 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import botocore.session
+
+from awscli.testutils import unittest, aws
+
+
+class TestDescribeInstances(unittest.TestCase):
+    @classmethod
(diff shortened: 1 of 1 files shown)


tests/integration/test_ec2.py
```diff
@@ -0,0 +1,65 @@
...
+
+class TestDescribeInstances(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Create an instance in the scope of a class so that only one needs to
+        # be created during the full test run.
+        cls.session = botocore.session.get_session()
+        cls.region = 'us-west-2'
+        cls.client = cls.session.create_client('ec2', cls.region)
+        response = cls.client.run_instances(
```
jamesls: -1 for me.

I feel this is a lesson learned from boto2.  We haven't spun up instances or dbs, etc. in any of the CLI/botocore code base up to this point.  This causes considerable maintenance burden both with flaky tests, cleanup failures, and time delays.  

I don't think the benefit of having this specific type of test outweighs the benefits.

JordonPhillips: That's a good point. The point of these tests is to verify the service's parameter handling, and EC2 conveniently provides `DryRun`. That should be sufficient.

JordonPhillips:
Updated to not create resources

kyleknap:
:ship: I like the fact that we are using dryrun now.

tests/integration/test_ec2.py
jamesls: Ooh, that's perfect!

jamesls:
:shipit:

Another option to consider in the future is the particular error code.  For example:

```
$ aws ec2 describe-instances --instance-ids i-12345666 --max-results 5 --no-paginate

An error occurred (InvalidParameterCombination) when calling the DescribeInstances operation: The parameter instancesSet cannot be used with the parameter maxResults

$ aws ec2 describe-instances --instance-ids i-12345666

An error occurred (InvalidInstanceID.NotFound) when calling the DescribeInstances operation: The instance ID 'i-12345666' does not exist
```

So in general, we can assert than we never get `InvalidParameterCombination` and just ensure we specify bogus instance ids (or bogus filter criteria, etc).

JordonPhillips:
That will be good to keep in mind for services that don't have DryRun

jamesls:
It depends on what you're trying to test.  `--dry-run` isn't going to catch param validation errors.  Consider:

```
$ aws ec2 describe-instances --instance-ids i-12345666 --max-results 5 --no-paginate

An error occurred (InvalidParameterCombination) when calling the DescribeInstances operation: The parameter instancesSet cannot be used with the parameter maxResults

$ aws ec2 describe-instances --instance-ids i-12345666 --max-results 5 --no-paginate --dry-run

An error occurred (DryRunOperation) when calling the DescribeInstances operation: Request would have succeeded, but DryRun flag is set.
```

In the example about, dry run says that request is fine, but we actually get a param validation error.  DryRun is just about whether or not you have appropriate permissions.

PR #2104, opened 2016-08-12
Result recorder

kyleknap:
The abstraction tracks statistic based on ``Results`` passed to it

It currently tracks the following:
* Expected number of files that will be transferred 
* Expected number of bytes that will be transferred
* Number of bytes transferred so far
* Number of files transferred so far
* Number of warnings received so far
* Number of bytes failed to be transferred so far

I was not sure if it made sense to track number of bytes failed to be transferred so far and considered just combining it with number of bytes transferred, but that seemed a little weird. In the end I will need to account for that number when printing progress. So I am up for any suggestions.

I am also up for any other statistics to track, but I was just trying to keep it small for now.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -22,20 +22,21 @@ from awscli.customizations.s3.utils import WarningResult
 LOGGER = logging.getLogger(__name__)
 
 
-def _create_new_result_cls(name, extra_fields=None):
+BaseResult = namedtuple('BaseResult', ['transfer_type', 'src', 'dest'])
+
+
+def _create_new_result_cls(name, extra_fields=None, base_cls=BaseResult):
     # Creates a new namedtuple class that subclasses from BaseResult for the
     # benefit of filtering by type and ensuring particular base attrs.
 
     # NOTE: _fields is a public attribute that has an underscore to avoid
     # naming collisions for namedtuples:
     # https://docs.python.org/2/library/collections.html#collections.somenamedtuple._fields
-    fields = list(BaseResult._fields)
+    fields = list(base_cls._fields)
     if extra_fields:
         fields += extra_fields
-    return type(name, (namedtuple(name, fields), BaseResult), {})
+    return type(name, (namedtuple(name, fields), base_cls), {})
 
-
-BaseResult = namedtuple('BaseResult', ['transfer_type', 'src', 'dest'])
 
 QueuedResult = _create_new_result_cls('QueuedResult', ['total_transfer_size'])
 
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -21,21 +21,31 @@ from awscli.customizations.s3.utils import WarningResult
 
 LOGGER = logging.getLogger(__name__)
 
-QueuedResult = namedtuple(
-    'QueuedResult',
-    ['transfer_type', 'src', 'dest', 'total_transfer_size']
-)
-ProgressResult = namedtuple(
-    'ProgressResult',
-    ['transfer_type', 'src', 'dest', 'bytes_transferred',
-     'total_transfer_size']
-)
-SuccessResult = namedtuple(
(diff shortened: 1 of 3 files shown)


awscli/customizations/s3/results.py
```diff
@@ -121,3 +122,77 @@ def _get_src_dest(self, future):
...
+            WarningResult: self._record_warning_result,
+        }
+
+    def record_result(self, result):
+        """Record the result of an individual Result object"""
+        self._result_handler_map.get(type(result), self._record_noop)(
+            result=result)
+
+    def _get_progress_cache_key(self, result):
+        return ':'.join([result.transfer_type, result.src, result.dest])
```
jamesls: Do you think it's an issue that this code relies on the property that any key in `_result_handler_map` must have a `transfer_type`, `src`, and `dest`.  There's not really a base type that we can say all `result` params must subclass from `X`.

I think there's potential that if we add more result types we might forget this requirement.

jamesls:
Overall I think it looks good.

Do all the callbacks invoked on the `ResultRecorder` object happen in the same thread?  As long as that's the case I don't think there's any issues.  Otherwise there might be some race conditions to handle.

awscli/customizations/s3/results.py
kyleknap: @jamesls
Yeah possibly. I can add a ``SubscriberResult`` base class that makes the three params required for any results produced from a ``ResultSubscriber``. Yet again I am not sure how many more result types that I will need to add. Right now I can really only see it being needed for dryrun results maybe (but that may require a different handler)? Otherwise it should not be required for all Results as stuff like ``WarningResults`` only requires a message. What do you think? I would be more on the side if I find myself adding any more result classes that require this, then add it?

kyleknap:
Yep the ResultRecorder will be used in only one thread so should not have to worry about concurrency issues.

awscli/customizations/s3/results.py
jamesls: That seems reasonable.  Even if a type check was added, it would still likely be an error case anyways, and the main benefit there would just be a better error message than `AttributeError`.

jamesls:
:shipit:

awscli/customizations/s3/results.py
kyleknap: Yeah I like the type check idea. Just a quick helpful reminder for the future. I will add that.

kyleknap:
@jamesls I just updated to subclass from a ``BaseResult`` class and added the type check. Let me know what you think about how I did it.

I was not sure what tests to add for the error case as it is more of a help case for us when we are developing and it is impossible to replicate unless I inject new custom behavior into the ResultRecorder, which would be weird for a test.

awscli/customizations/s3/results.py
```diff
@@ -21,21 +21,31 @@
...
+    # https://docs.python.org/2/library/collections.html#collections.somenamedtuple._fields
+    fields = list(BaseResult._fields)
+    if extra_fields:
+        fields += extra_fields
+    return type(name, (namedtuple(name, fields), BaseResult), {})
+
+
+BaseResult = namedtuple('BaseResult', ['transfer_type', 'src', 'dest'])
+
+QueuedResult = _create_new_result_cls('QueuedResult', ['total_transfer_size'])
```
jamesls: It's a little weird to have `_create_new_result_cls` implicitly depend on `BaseResult` which is defined above.  If the `BaseResult` line is moved below a `_create_new_result_cls` the function will fail.  What do you think about passing `BaseResult` to `_create_new_result_cls` to make this more explicit?

kyleknap: Makes sense I can add that. Would make it more extensible as well.

kyleknap:
@jamesls I updated the result class factory by allowing you to inject a base result class. Should be good to look at again.

jamesls:
:shipit:

Issue #1117, opened 2015-01-29
Non zero RC for S3 sync even if file is excluded


jamesls:
A file that would is `--exclude`'d should not trigger a non zero RC if we can't read the file.  For example:

```
/tmp $ mkdir demo
/tmp $ mkfifo demo/bad
/tmp $ touch demo/good
/tmp $ aws s3 sync demo/ s3://bucket/demo/ --exclude '*' --dryrun
warning: Skipping file /private/tmp/demo/bad. File is character special device, block special device, FIFO, or socket.
Completed 0 part(s) with ... file(s) remaining
/tmp $ echo $?
2
```

Given that `/tmp/demo/bad` was never going to be transferred in the first place (because it's been excluded via `--exclude '*'`, it seems odd that this will trigger a non zero RC.  I'd expect the RC to be 0 here.

ChrisSLT:
This was reported half a year ago - can you tell us what the delay is in fixing this?

ianderse:
Any news on this issue? Thanks!

deweysasser:
Perhaps it is time for someone to write a better tool.

gsmcwhirter:
I'm also interested in a fix to this issue. Any progress?

kachkaev:
Also facing this when trying to backup a docker volume. It contains a couple of sockets and this produces `warning: Skipping file /path/to/abcd.socket. File is character special device, block special device, FIFO, or socket.` Trying all sorts of values for `--exclude` did not help, even if I `--exclude "*"`.

Because `s3 sync` always returns a non-zero exit code, there is no way to use the command  in the automation tools like ansible unless executing something insane like `aws s3 sync [...args] || true`.

mcblair:
I am experiencing this as well - warnings for explicitly excluded files.

This is problematic because `s3 sync` is exiting with a non-zero exit code.

When automating `s3 sync` like this: [Simple S3 Resource for Concourse CI](https://github.com/18F/s3-resource-simple)

We exclude everything and only include the directories we want to sync. The result is that the expected files are uploaded but because of the non-zero exit code - the Concourse CI job using the above resource fails.

TBeijen:
``s3 cp`` seems to suffer from the same issue:

```
aws s3 cp --recursive --exclude '*' --include 'binlog.*' /var/lib/mysql/ s3://demo/bucket/
warning: Skipping file /var/lib/mysql/mysql.sock. File is character special device, block special device, FIFO, or socket.
[...]

Issue #1588, opened 2015-10-23
--exclude (still) inconsistent


drzraf:
```
$ cd ~
$ aws s3 sync / s3://bucket/test/ --exclude "bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "/bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "./bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "/bin/*"  --dryrun
$ cd /
$ aws s3 sync / s3://bucket/test/ --exclude "bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "/bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "./bin/*"  --dryrun
$ aws s3 sync / s3://bucket/test/ --exclude "/bin/*"  --dryrun
```

None of the above commands (whatever cwd is) will exclude /bin from being synch'ed
Using ``"*/bin/*"`` _would_ work, but it's unwanted since it would exclude other subdirectories like ``home/foo/comp/bin/``

mtdowling:
I played around with this, and I too see that the `--exclude` behavior is not working well (this possibly has to do with the fact that the root folder is being used).

thedukeness:
Perhaps this is related, I am using aws-cli/1.10.0 and cannot ever get a file like this to be included or excluded: ._3E53F853-DA82-4926-AC21-5C1096FB126C.MP3

aws s3 rm s3://foo/foo2/foo3 --include "*/._*" --exclude "*" --recursive --dryrun
aws s3 rm s3://foo/foo2/foo3 --include "*._*" --exclude "*" --recursive --dryrun
aws s3 sync /foo/foo2/foo3 s3://foo/foo2/foo3 --exclude "*/._*" --delete --dryrun

kirkmadera:
According to #548, it seems as though the behavior is to prepend each exclude/include pattern with the current working directory. I was able to verify this behavior with this test

Given the 
    test1.txt

    test2/
        test2.txt
        test3/
            test3.txt


`aws s3 sync . s3://examplebucket --exclude "test2/test3/*"`
test3.txt is excluded

`aws s3 sync . s3://examplebucket --exclude "test3/*"`
test3.txt is not excluded


[...]

Issue #1138, opened 2015-02-09
aws s3 sync traverses excluded directory - takes up a long time


ChrisSLT:
This has been discussed elsewhere but I wasn't sure what the response was.  I wanted to see if there would be an ability to stop SYNC from traversing an excluded directory because it takes up a lot of time.

When you do a sync from ~/ with --exclude "STAR" and --include "~/Documents/STAR" for example, the sync operation still takes many minutes to complete because it walks through every single file in every directory in the user's home folder.  I wish I knew enough to get into the actual code, but I'm wondering if you could stop the SYNC operation from doing this?

Thank you

jamesls:
I agree, we should stop traversing into a directory if it's excluded/included.  This is also related to #1117 which will throw warnings even if a file is excluded, which has the same root cause:  we're traversing into a directory when we could be skipping it.

I don't see any reason why we shouldn't change this behavior.

ChrisSLT:
Terrific, thank you

ChrisSLT:
I know you guys are working away on many fronts, but wondering if you could let us know a timeline or roadmap on fixing this issue?  thank you much

ChrisSLT:
Would you be able to let us know a timeline on this?

kyleknap:
It is difficult to put a specific timeline/date for this. Since this would fix a couple of bugs that have been reported to us, it is has a relatively high priority for us. However, implementation-wise it may take a while because it requires a good amount of refactoring on the internals of the s3 commands. Ultimately, the feature is near the top of our list of features to implement, but time has yet to be allocated to work on it. We can let you know when we do start working on this feature.

ChrisSLT:
Thank you.  Really appreciate it.

ChrisSLT:
Can you please give an indication as to when this issue will be fixed?  It creates a lot of complexity to have to have multiple scripts to get around this bug.  Thank you

ChrisSLT:
Can you tell me if this is intended to be fixed?

ChrisSLT:
Please update us on this issue.  Due to the bug, our team is getting pressure to move away from S3 sync.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

[...]

PR #2105, opened 2016-08-14
s3.sync: Do not stat files that don't match filters

clarete:
This probably helps on issue #1138.

I have a HUGE `~/src` directory that contains projects that I clone from various version control systems. Which makes me want to avoid backing them up on my personal S3 bucket to save some coins. For that, I used the `--exclude="src/*"` parameter (I also make sure the `aws` command is called from my `$HOME` directory, since I learned that the filters start matching from the current directory -- more details can be found in the #1588 issue). Although it worked, it still scanned and stat'ed the entire directory.

This patch changes the `FileGenerator` class to
1. take the filters from the command line as a parameter in its constructor
2. Use the `file_filter` parameter to match the path received by `.should_ignore_file` and figure out if the path received by the method is allowed by the filters.

This patch can be considered a work in progress because I only tested my specific need. It still needs proper tests. I didn't really write any because I wanted to ask if the strategy I chose is enough. If that's the case, I'd spend time writing the proper tests. Otherwise, I'd love to hear a better idea. I may be able to rework it if it's not too time consuming.

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -18,6 +18,7 @@ from dateutil.parser import parse
 from dateutil.tz import tzlocal
 from botocore.exceptions import ClientError
 
+from awscli.customizations.s3.fileinfo import FileInfo
 from awscli.customizations.s3.utils import find_bucket_key, get_file_stat
 from awscli.customizations.s3.utils import BucketLister, create_warning, \
     find_dest_path_comp_key, EPOCH_TIME
@@ -116,7 +117,8 @@ class FileGenerator(object):
     ``FileInfo`` objects to send to a ``Comparator`` or ``S3Handler``.
     """
     def __init__(self, client, operation_name, follow_symlinks=True,
-                 page_size=None, result_queue=None, request_parameters=None):
+                 page_size=None, result_queue=None, request_parameters=None,
+                 file_filter=None):
         self._client = client
         self.operation_name = operation_name
         self.follow_symlinks = follow_symlinks
@@ -127,6 +129,7 @@ class FileGenerator(object):
         self.request_parameters = {}
         if request_parameters is not None:
             self.request_parameters = request_parameters
+        self.file_filter = file_filter
 
     def call(self, files):
         """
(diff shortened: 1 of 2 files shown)


clarete:
I hope you guys like the idea, it also seems to fix #1117.

kyleknap:
@clarete I think the approach is reasonable. I think this is something we would pull in.

The only thing that I would be wary of is to make sure the file comparisons in sync do not get out of order when comparing the two order lists. But I do not think that is an issue because the comparison of files happens after the usual filtering and you are just pushing the filtering to happen even earlier in the process.

If you can get some unit tests up, we will take a more scrutinized look and make sure that none of the integration tests are not broken and proceed from there.

clarete:
Still interested, just didn't have time yet. Will post results as soon as my weird schedule allows. Sorry about that!

gravyboat:
@clarete Are you still planning on finishing this?

bugfood:
This patch still applies to current `develop`, albeit with a couple trivial merge conflicts. I tried it out and found it quite suitable. The largest benefit for my use case is that it makes `aws` no longer recurse deeply into directories that are excluded. It still stats the contents of excluded directories, but does not go into subdirectories thereof.

In my case, I am periodically syncing a large deep file tree from a netapp NFS share; netapp provides access to snapshots in `.snapshot/8hour.2020-07-26_1815`, `.8hour.2020-07-27_0215`, etc. for each snapshot. Naturally, I am excluding the snapshots, but with six snapshots and unpatched `aws`, the sync takes around seven times as long as it should (the source contents don't change substantially, so there is little actual data to transfer after the initial sync).

I have made a PR with a minor modification and an update to unit tests:
https://github.com/aws/aws-cli/pull/5425
I'll try to work on that until it passes the CI tests.

tim-finnigan:
Thank you for submitting this PR. We are going to use https://github.com/aws/aws-cli/pull/5425 to track this going forward, as that includes these changes along with a few tests and modifications as mentioned above. I just left a response [here](https://github.com/aws/aws-cli/pull/5425#issuecomment-1086413344) on that PR requesting some additional changes.

PR #2106, opened 2016-08-15
Set default page size for some EC2 operations

JordonPhillips:
When customers use max-results on EC2 operations, the entire
response will be returned and then the CLI will filter out what
the customer wants. This can cause significant pauses for what
seems like a simple request. This change will set the default
page size on describe-volumes and describe-snapshots if no
incompatible arguments are specified.

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/ec2/paginate.py
+++ b/awscli/customizations/ec2/paginate.py
@@ -1,0 +1,92 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+class EC2PageSizeInjector(object):
+
+    DEFAULT_TARGET_OPERATIONS = {
+        "describe-volumes": [],
+        "describe-snapshots": ['owner-ids', 'restorable_by_user_ids']
+    }
+
+    DEFAULT_GLOBAL_WHITELIST = [
+        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
+        'next_token', 'dry_run', 'starting_token'
+    ]
+
+    def __init__(self, default_page_size=1000, target_operations=None,
+                 global_whitelist=None):
+        """
+        :type default_page_size: int
+        :param default_page_size: The value to inject for page_size.
+
+        :type target_operations: dict
+        :param target_operations: A dictionary whose keys are operations and
+            whose values are parameters that are white-listed for that
+            operation. If a parameter is specified that is not in that list,
+            page_size will not be set.
+
+        :type global_whitelist: list of str
+        :param global_whitelist: A list of parameters that should be
+            whitelisted for every operation.
+        """
+        self._default_page_size = default_page_size
+        self._target_operations = target_operations
+        if self._target_operations is None:
+            self._target_operations = self.DEFAULT_TARGET_OPERATIONS
+        self._global_whitelist = global_whitelist
+        if self._global_whitelist is None:
+            self._global_whitelist = self.DEFAULT_GLOBAL_WHITELIST
+
+    def register(self, event_emitter):
+        """ Register `inject` for each target operation. """
+        event_template = "operation-args-parsed.ec2.%s"
+        for operation in self._target_operations.keys():
+            event = event_template % operation
+            event_emitter.register(event, self.inject)
+
+    def inject(self, event_name, parsed_args, parsed_globals, **kwargs):
+        """ Conditionally inject page_size. """
+        if not parsed_globals.paginate:
+            return
+
+        operation_name = event_name.split('.')[-1]
+
+        whitelisted_params = self._target_operations.get(operation_name, None)
+        if whitelisted_params is None:
+            return
+
+        whitelisted_params += self._global_whitelist
+        specified_params = self._get_specified_params(parsed_args)
+
+        for param in specified_params:
+            if param not in whitelisted_params:
+                return
+
+        parsed_args.page_size = self._default_page_size
+
+    def _get_specified_params(self, namespace):
+        attrs = dir(namespace)
+        params = []
+        for attr in attrs:
+            # The default value of generate_cli_skeleton is False where
+            # usually a boolean argument still defaults to None. Therefore
+            # it must be specifically ignored.
+            if attr == 'generate_cli_skeleton':
+                continue
+
+            if attr[0] != '_' and getattr(namespace, attr) is not None:
+                params.append(attr)
+
+        return params
+
--- a/awscli/customizations/ec2/paginate.py
+++ b/awscli/customizations/ec2/paginate.py
@@ -12,6 +12,10 @@
 # language governing permissions and limitations under the License.
 
 
+def register_ec2_page_size_injector(event_emitter):
+    EC2PageSizeInjector().register(event_emitter)
+
+
 class EC2PageSizeInjector(object):
 
     # Operations to auto-paginate and their specific whitelists.
@@ -20,50 +24,42 @@ class EC2PageSizeInjector(object):
     #    Value: List of parameters to add to whitelist for that operation.
     TARGET_OPERATIONS = {
         "describe-volumes": [],
-        "describe-snapshots": ['owner_ids', 'restorable_by_user_ids']
+        "describe-snapshots": ['OwnerIds', 'RestorableByUserIds']
     }
 
     # Parameters which should be whitelisted for every operation.
-    GLOBAL_WHITELIST = [
-        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
-        'next_token', 'dry_run', 'starting_token'
-    ]
+    GLOBAL_WHITELIST = ['NextToken', 'DryRun', 'PaginationConfig']
 
     DEFAULT_PAGE_SIZE = 1000
 
     def register(self, event_emitter):
-        """ Register `inject` for each target operation. """
-        event_template = "operation-args-parsed.ec2.%s"
-        for operation in self.TARGET_OPERATIONS.keys():
+        """Register `inject` for each target operation."""
+        event_template = "calling-command.ec2.%s"
+        for operation in self.TARGET_OPERATIONS:
             event = event_template % operation
-            event_emitter.register(event, self.inject)
+            event_emitter.register_last(event, self.inject)
 
-    def inject(self, event_name, parsed_args, parsed_globals, **kwargs):
-        """ Conditionally inject page_size. """
+    def inject(self, event_name, parsed_globals, call_parameters, **kwargs):
+        """Conditionally inject PageSize."""
         if not parsed_globals.paginate:
+            return
+
+        pagination_config = call_parameters.get('PaginationConfig', {})
+        if 'PageSize' in pagination_config:
             return
 
         operation_name = event_name.split('.')[-1]
 
-        whitelisted_params = self.TARGET_OPERATIONS.get(operation_name, None)
+        whitelisted_params = self.TARGET_OPERATIONS.get(operation_name)
         if whitelisted_params is None:
             return
 
         whitelisted_params += self.GLOBAL_WHITELIST
-        specified_params = self._get_specified_params(parsed_args)
 
-        for param in specified_params:
+        for param in call_parameters:
             if param not in whitelisted_params:
                 return
 
-        parsed_args.page_size = self.DEFAULT_PAGE_SIZE
-
-    def _get_specified_params(self, namespace):
-        attrs = dir(namespace)
-        params = []
-        for attr in attrs:
-            if attr[0] != '_' and getattr(namespace, attr) is not None:
-                params.append(attr)
-
-        return params
+        pagination_config['PageSize'] = self.DEFAULT_PAGE_SIZE
+        call_parameters['PaginationConfig'] = pagination_config
 
--- a/awscli/customizations/ec2/paginate.py
+++ b/awscli/customizations/ec2/paginate.py
@@ -16,7 +16,7 @@ class EC2PageSizeInjector(object):
 
     DEFAULT_TARGET_OPERATIONS = {
         "describe-volumes": [],
-        "describe-snapshots": ['owner-ids', 'restorable_by_user_ids']
+        "describe-snapshots": ['owner_ids', 'restorable_by_user_ids']
     }
 
     DEFAULT_GLOBAL_WHITELIST = [
@@ -79,12 +79,6 @@ class EC2PageSizeInjector(object):
         attrs = dir(namespace)
         params = []
         for attr in attrs:
-            # The default value of generate_cli_skeleton is False where
-            # usually a boolean argument still defaults to None. Therefore
-            # it must be specifically ignored.
-            if attr == 'generate_cli_skeleton':
-                continue
-
             if attr[0] != '_' and getattr(namespace, attr) is not None:
                 params.append(attr)
 
--- a/awscli/customizations/ec2/paginate.py
+++ b/awscli/customizations/ec2/paginate.py
@@ -28,7 +28,7 @@ class EC2PageSizeInjector(object):
     }
 
     # Parameters which should be whitelisted for every operation.
-    GLOBAL_WHITELIST = ['NextToken', 'DryRun', 'PaginationConfig']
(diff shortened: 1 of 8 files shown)


awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,92 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+class EC2PageSizeInjector(object):
+
+    DEFAULT_TARGET_OPERATIONS = {
+        "describe-volumes": [],
+        "describe-snapshots": ['owner-ids', 'restorable_by_user_ids']
```
kyleknap: Why does ``owner-ids`` have a hyphen and ``restorable_by_user_ids`` uses underscores?

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    DEFAULT_GLOBAL_WHITELIST = [
+        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
+        'next_token', 'dry_run', 'starting_token'
+    ]
+
+    def __init__(self, default_page_size=1000, target_operations=None,
```
kyleknap: Are the ``target_operations`` and ``global_whitelist`` needed? It is really only used for testing and if that is the case why can't you just keep the class variables, instantiate an instance, and then modify the list or dictionary for the test.

Another reason I am asking is that I find the ``target_operations`` interface to be strange. Strange in that it will be a dictionary. I would have more expected that to be a list of operations to inject into and then you would have you class variable ``PER_OPERATION_WHITELIST`` which will add specific parameters to the ``GLOBAL_WHITELIST`` for that operation if it is present in ``PER_OPERATION_WHITELIST``.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,92 @@
...
+        parsed_args.page_size = self._default_page_size
+
+    def _get_specified_params(self, namespace):
+        attrs = dir(namespace)
+        params = []
+        for attr in attrs:
+            # The default value of generate_cli_skeleton is False where
+            # usually a boolean argument still defaults to None. Therefore
+            # it must be specifically ignored.
+            if attr == 'generate_cli_skeleton':
```
kyleknap: Why is this needed? You whitelist it in the ``DEFAULT_GLOBAL_WHITELIST`` right?

tests/integration/test_ec2.py
```diff
@@ -13,23 +13,52 @@
...
+    def setUp(self):
+        self.prefix = 'ec2 describe-instances --region us-west-2'
+
+    def test_describe_instances_with_id(self):
+        command = self.prefix + ' --instance-ids malformed-id'
         result = aws(command)
-        expected_response = ('Request would have succeeded, '
-                             'but DryRun flag is set.')
-        self.assertIn(expected_response, result.stderr)
+        self.assertIn('InvalidInstanceID.Malformed', result.stderr)
```
kyleknap: Question: If you use MaxResults and provide a malformed instance id, what error gets propogated?  I have the same question for the rest of the test cases. Might be good to add a note or rename the tests as without any background knowledge it is not obvious why we are testing for that specific error, other than for the sake of a quick smoke test of the service.

Especially since for the case, where you have no instances matching the filter, EC2 will actually not complain about MaxResults being present:
```
$ aws ec2 describe-instances --filters Name=instance-id,Values=malformed --max-items 5
{
    "Reservations": []
}

```
That being said I am not sure if there is a great way to test the filter case.

kyleknap:
Approach looks fine.  My main comments were about the interface and questions about the test cases.

tests/integration/test_ec2.py
JordonPhillips: For the id case, the invalid combination error gets thrown. The filters case is harder to be exacting on, but the idea is to make sure it works generally even if the change doesn't impact it.

awscli/customizations/ec2/paginate.py
JordonPhillips: It seems strange to duplicate that information. I would be worried about future updates missing one location or another. Maybe the naming could be better?

awscli/customizations/ec2/paginate.py
```diff
@@ -14,44 +14,27 @@
...
-            whitelisted for every operation.
-        """
-        self._default_page_size = default_page_size
-        self._target_operations = target_operations
-        if self._target_operations is None:
-            self._target_operations = self.DEFAULT_TARGET_OPERATIONS
-        self._global_whitelist = global_whitelist
-        if self._global_whitelist is None:
-            self._global_whitelist = self.DEFAULT_GLOBAL_WHITELIST
+    DEFAULT_PAGE_SIZE = 1000
```
kyleknap: Oh I was fine with the ``page_size`` constructor parameter as it would make sense to individually tune each injector, but I am fine with it being a class variable for now and refactoring it if we need to choose different defaults for different operations

tests/integration/test_ec2.py
kyleknap: I wonder if it is possible if there are any other filter parameter combinations (other than instance-id) that can recreate the regression reliably. Otherwise I am about a zero on adding this test as there is not much value the test adds as even if we reintroduced the regression it would not catch it.

I doubt there will be a different filter parameter combination but it would be good to take a quick look.

awscli/customizations/ec2/paginate.py
JordonPhillips: It would need refactoring either way since before it was just setting page size. I agree though that there's potential that other operations may need more granularity.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
+        'next_token', 'dry_run', 'starting_token'
+    ]
+
+    DEFAULT_PAGE_SIZE = 1000
+
+    def register(self, event_emitter):
+        """ Register `inject` for each target operation. """
+        event_template = "operation-args-parsed.ec2.%s"
+        for operation in self.TARGET_OPERATIONS.keys():
```
jamesls: You don't need `.keys()`, you're not mutating the dict.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+    # Parameters which should be whitelisted for every operation.
+    GLOBAL_WHITELIST = [
+        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
+        'next_token', 'dry_run', 'starting_token'
+    ]
+
+    DEFAULT_PAGE_SIZE = 1000
+
+    def register(self, event_emitter):
+        """ Register `inject` for each target operation. """
```
jamesls: You have leading/trailing spaces on your docstrings which should be removed (pep257)

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+            event_emitter.register(event, self.inject)
+
+    def inject(self, event_name, parsed_args, parsed_globals, **kwargs):
+        """ Conditionally inject page_size. """
+        if not parsed_globals.paginate:
+            return
+
+        operation_name = event_name.split('.')[-1]
+
+        whitelisted_params = self.TARGET_OPERATIONS.get(operation_name, None)
```
jamesls: `,None` is a default value.  I think it's more idiomatic to not provide `None` here (use the default arg only if you have a different value value you want).

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+        specified_params = self._get_specified_params(parsed_args)
+
+        for param in specified_params:
+            if param not in whitelisted_params:
+                return
+
+        parsed_args.page_size = self.DEFAULT_PAGE_SIZE
+
+    def _get_specified_params(self, namespace):
+        attrs = dir(namespace)
```
jamesls: This is not a preferred way to do this.  `vars()` is far more common and used within our code base.  Please use `vars()` here.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+        params = []
+        for attr in attrs:
+            if attr[0] != '_' and getattr(namespace, attr) is not None:
```
jamesls: This will give you false positives because it's assuming that `None` is the unspecified values.  It doesn't take into account non None default values (i.e booleans that default to False, etc)

awscli/handlers.py
```diff
@@ -142,3 +143,4 @@ def awscli_initialize(event_handlers):
         register_create_keys_from_csr_arguments)
     register_cloudfront(event_handlers)
     register_gamelift_commands(event_handlers)
+    EC2PageSizeInjector().register(event_handlers)
```
jamesls: I'd prefer we stick to the more common pattern of having a single module level function, `register_*`, that handles the registration.  It's by far the common pattern used here, and it will also make it easier to refactor in the future if we start to persue deferred event registration.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+            return
+
+        whitelisted_params += self.GLOBAL_WHITELIST
+        specified_params = self._get_specified_params(parsed_args)
+
+        for param in specified_params:
+            if param not in whitelisted_params:
+                return
+
+        parsed_args.page_size = self.DEFAULT_PAGE_SIZE
```
jamesls: We should have debug logs that we're messing with the `parsed_args`.  This will make it easier to troubleshoot issues.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,69 @@
...
+    #    Key:   Operation
+    #    Value: List of parameters to add to whitelist for that operation.
+    TARGET_OPERATIONS = {
+        "describe-volumes": [],
+        "describe-snapshots": ['owner_ids', 'restorable_by_user_ids']
+    }
+
+    # Parameters which should be whitelisted for every operation.
+    GLOBAL_WHITELIST = [
+        'cli_input_json', 'generate_cli_skeleton', 'help', 'max_items',
```
jamesls: I don't get why some of these are whitelisted.  Why would `help` be a whitelisted parameter?  Same with `--generate-cli-skeleton`.  Why is this relevant for deciding if we inject a page_size param?

jamesls:
I'm not sure that trying to address this at the `parsed_args` layer is going to work.  For example, there's currently a regression here for `--cli-input-json`:

```
$ cat /tmp/foo.json
{
    "SnapshotIds": [
        "snap-05339dbe40d27bfa6"
    ]
}
$ aws ec2 describe-snapshots --cli-input-json file:///tmp/foo.json

An error occurred (InvalidParameterCombination) when calling the DescribeSnapshots operation: The parameter snapshotSet cannot be used with the parameter maxResults
```

It would be ideal if there was some unified way to handle this logic without having to also special case `cli-input-json`.

awscli/customizations/ec2/paginate.py
JordonPhillips: I added all the generic parameters present in the namespace. I suppose for those two it really doesn't matter since they'll prevent a request anyway.

awscli/customizations/ec2/paginate.py
JordonPhillips: Modeled params will default to `None` regardless of their type, so we only have to worry about that for custom params that get added to `parsed_args` specifically, not `parsed_globals`. Right now (for these operations) those are `help`, `cli_input_json`, and `generate_cli_skeleton`. Of those, only `generate_cli_skeleton` has a default value. Adding it to the whitelist takes care of that.

Data I have access to might change when I move to a different handler, but at this point I don't have any way to know what the default is.

JordonPhillips:
This is good to look at again

tests/unit/customizations/ec2/test_paginate.py
```diff
@@ -0,0 +1,115 @@
...
+        event_name = 'operation-args-parsed.ec2.foo'
+        injector.inject(
+            event_name=event_name,
+            parsed_globals=parsed_globals,
+            call_parameters=call_parameters
+        )
+        page_size = call_parameters.get('PaginationConfig', {}).get('PageSize')
+        self.assertIsNone(page_size)
+
+    def test_global_whitelist(self):
```
kyleknap: How does this test the global whitelist? It looks like it is testing operation level whitelist as this is returning None because ``baz`` is passed as a call_parameter and ``foo`` has an empty whitelist?

kyleknap:
I just had a comment about a test. But otherwise :ship:

tests/unit/customizations/ec2/test_paginate.py
JordonPhillips: The only whitelisted parameter is `bar`, any other provided parameters should cause it to not inject page size.

kyleknap: But what specific global whitelist parameter is causing the page size to not being injected? All I see is ``paginate=True``.

Is that what is making it not inject page size? If so  that is really strange when you compare this test and the last because the last test explicitly set paginate to False to get no injection and this one sets paginate to True to get no injection.

Could we just append bar to the current global whitelist and then pass a more explicit parameter in the parsed globals? Furthermore, if page_size is the one causing it to not get injected, it is ambiguous if it is sourced from the global or operation whitelist as your call parameters do not match the operation white list as well.

JordonPhillips: There is no specific parameter that's causing it to not be injected, that's the point of a whitelist: anything not in the list is rejected. `paginate` must be set for anything to be injected regardless of the state of the whitelists.

The global whitelist isn't a whitelist for `parsed_globals` it's a whitelist that is applied globally to all operations. Would it be less confusing to call it universal whitelist?

kyleknap: Oh that makes more sense. Yeah ``GLOBAL_WHITELIST`` definitely through me off because you had global parameters in it before. How do you think ``DEFAULT_PARAMETER_WHITELIST`` sounds? That would be much more clear to me.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,65 @@
...
+    # Format:
+    #    Key:   Operation
+    #    Value: List of parameters to add to whitelist for that operation.
+    TARGET_OPERATIONS = {
+        "describe-volumes": [],
+        "describe-snapshots": ['OwnerIds', 'RestorableByUserIds']
+    }
+
+    # Parameters which should be whitelisted for every operation.
+    GLOBAL_WHITELIST = ['NextToken', 'DryRun', 'PaginationConfig']
```
jamesls: Why is `DryRun` in the whitelist?  DryRun is just used for auth checks and doesn't return any data to paginate.  What problem did you run into if you didn't include this in the whitelist?

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,65 @@
...
+        if 'PageSize' in pagination_config:
+            return
+
+        operation_name = event_name.split('.')[-1]
+
+        whitelisted_params = self.TARGET_OPERATIONS.get(operation_name)
+        if whitelisted_params is None:
+            return
+
+        whitelisted_params += self.GLOBAL_WHITELIST
```
jamesls: Not really a problem because we're relying on the fact that this event is called once, but I find it weird that you're mutating the actual list in self.TARGET_OPERATIONS every time this event handler is called.  This means GLOBAL_WHITELIST gets append to the same list every time this handler is called.

awscli/customizations/ec2/paginate.py
```diff
@@ -0,0 +1,65 @@
...
+
+        for param in call_parameters:
+            if param not in whitelisted_params:
+                return
+
+        pagination_config['PageSize'] = self.DEFAULT_PAGE_SIZE
+        call_parameters['PaginationConfig'] = pagination_config
+
```
jamesls: You can remove this blank line

jamesls:
Had a few questions, but overall approach seems good to me.

awscli/customizations/ec2/paginate.py
JordonPhillips: Yeah, that should be changed.

awscli/customizations/ec2/paginate.py
JordonPhillips: There's no problem if it is excluded, but that list is the parameters where there will be no issue putting the paging param in. I'm 0 on keeping it, so I'll remove it if you'd like.

tests/integration/test_ec2.py
JordonPhillips: Those tests are there to verify that the common cases for this operation are covered since we're modifying it. I'll see if I can find a way to reproduce the issue, but either way I think they should be kept.

kyleknap:
Thanks :ship:

jamesls:
:shipit:

PR #2111, opened 2016-08-16
Add ResultPrinter

kyleknap:
This introduces an abstraction that will use ``ResultRecorder`` and ``Results`` received from the result queue to print out status statements.

Important Notes:
* The progress is slightly different now. Instead of printing parts we print number of bytes completed.
* There is no more ``...`` notation in progress. It will always print how many files/number of bytes it has currently seen go by. So there is a chance that the number of total expected bytes and number of remaining files can increase. This change I would like other's thoughts on. Not sure if it would be confusing to see the remaining files and total expected bytes to increase. Maybe I should add additional information to the progress statement when I know the count is not finalized? But I do know it is much more helpful than the ``...`` notation.
* To implement new types of progress, you create a new ``ResultPrinter`` class. I implemented  ``OnlyShowErrorsResultPrinter`` as an example.

Let me know what you think.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -11,11 +11,14 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
+import sys
 from collections import namedtuple
 
 from s3transfer.subscribers import BaseSubscriber
 
 from awscli.customizations.s3.utils import relative_path
+from awscli.customizations.s3.utils import human_readable_size
+from awscli.customizations.s3.utils import uni_print
 from awscli.customizations.s3.utils import WarningResult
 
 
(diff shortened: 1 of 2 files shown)


jamesls:
Can you show an example of the new output?

kyleknap:
@jamesls Yep here is a script you can use to reproduce the output:
```py
import time

from awscli.customizations.s3.results import QueuedResult
from awscli.customizations.s3.results import ProgressResult
from awscli.customizations.s3.results import SuccessResult
from awscli.customizations.s3.results import FailureResult
from awscli.customizations.s3.results import ResultRecorder
from awscli.customizations.s3.results import ResultPrinter


transfer_type = 'upload'
src = 'myfile'
dest = 's3://mybucket/mykey'
total_size = 1024 * 1024 * 20

recorder = ResultRecorder()
printer = ResultPrinter(recorder)

def process(result):
    recorder.record_result(result)
    printer.print_result(result)

queued = QueuedResult(transfer_type, src, dest, total_size)
process(queued)

chunk_size = 1024 * 1024
for i in range(0, total_size, chunk_size):
    progress = ProgressResult(transfer_type, src, dest, chunk_size, total_size)
    process(progress)
    time.sleep(0.2)

success = SuccessResult(transfer_type, src, dest)
process(success)
```

Here is a screenshot mid-progress:

screen shot 2016-08-16 at 9 51 49 am

Here is a screenshot once done:

screen shot 2016-08-16 at 9 52 02 am

Looking at the output. Looks like I forgot the parenthesis around file(s). So I added that. But let me know what you think

JordonPhillips:
I think I prefer showing the more exacting data, even if it changes. Would it be possible to show an approximate transfer rate as well? I know we have had several customer asks for that.

kyleknap:
Yeah we could calculate transfer rate as well. But that would require more work and I think is outside of the scope of this PR because I need to plumb it into every new abstraction that I have recently added (i.e. result subscribers, result recorder, and result printer).

kyleknap:
In general, I am trying to minimize the differences in porting all of this over. Once it is all in place, I think we are in much better place to actually implement features like displaying transfer speed.

kyleknap:
Also if you want to try it out with it all plumbed into the CLI. Here is a branch that I am working on: https://github.com/kyleknap/aws-cli/tree/integrate-cp that has it plumbed into ``cp``.

jamesls:
:shipit:

I spent some time trying out the integrate-cp branch.  I definitely like the total MB transferred progress, but after trying out a few large recursive copies, I'm not sure how I like the total number fluctuating.  On the one hand, it does more accurately represent what's happening internally, but on the other hand, I see customers asking questions about why this number keeps fluctuating.  Perhaps it might be worth (in a future PR) adding something that indicates this number is not the final number as you suggest.

I suppose with the way you've created these classes, it's much easier to write varying types of progress result classes.

kyleknap:
Yeah I like the idea of adding something like ``(still calculating total)`` to the output if we do not know the total is finalized.

PR #2114, opened 2016-08-18
Add ResultProcessor

kyleknap:
Abstraction is a thread that processes results in the result queue by
passing each result from the queue to the ResultRecorder and ResultPrinter.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -12,10 +12,13 @@
 # language governing permissions and limitations under the License.
 import logging
 import sys
+import threading
 from collections import namedtuple
 
 from s3transfer.subscribers import BaseSubscriber
 
+from awscli.compat import queue
+from awscli.customizations.s3.executor import ShutdownThreadRequest
 from awscli.customizations.s3.utils import relative_path
 from awscli.customizations.s3.utils import human_readable_size
 from awscli.customizations.s3.utils import uni_print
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/results.py
```diff
@@ -356,3 +359,43 @@ def _print_progress(self, **kwargs):
...
+                LOGGER.debug('Received result: %s', result)
+                try:
+                    self._process_result(result)
+                except Exception as e:
+                    LOGGER.debug(
+                        'Error processing result: %s', e, exc_info=True)
+            except queue.Empty:
+                pass
+
+    def _process_result(self, result):
```
jamesls: Is it worth just having a generic `ResultHandler` interface and having the `__init__` take a list of result handlers?  You could then either write adapters for the recorder/printer or update them to have the same interface.

What do you think?

kyleknap: Posssibly. I was actually considering doing that. The only reason I did not was because I did not see needing it in doing the port, but taking an arbitrary list of handlers is pretty flexible so I would be totally fine with making that the interface.

jamesls: Yeah, don't feel too strongly about it for this specific PR, it would just clean up the code a little bit:

```
for handler in self._result_handlers:
    handler.handle_result(result)
```

vs. having to check if a particular handler was `None` or not.

kyleknap: I'm updating it. I did not like the ``None`` check either.

jamesls:
:shipit:

I assume we're going to just change the existing recorder/printer to just have a `__call__(self, result):`?

kyleknap:
@jamesls I realized that as well when I was looking over the first commit I pushed. You just beat me to it. I went a head and changed the interface to use ``__call__()`` instead. So it would be good to get a look at the last commit I pushed.

jamesls:
:shipit:

PR #2117, opened 2016-08-19
Add needed subscribers for cp

kyleknap:
I am just trying to chunk out some of the code for review in integrating ``s3transfer`` with ``cp`` and ``cp --recursive``. This PR is mainly for the purpose of adding all of the additional logic that the CLI needs for performing transfers that is not by default in the ``TransferManager``.

Most of what I did was take logic from other parts of the code base and wrapped them into subscribers. For some of the ``on_queued`` subscribers, it is true that a subscriber is not necessarily needed but it is better to use a subscriber for some of the logic because if the logic is taking place before calling a ``TransferManager`` method there is a chance of 1) slowing down the enqueing of tasks and 2) having the entire transfer failing if there is an unhandled exception is thrown before being sent to the ``TransferManger``.

The custom logic that I implemented through subscribers include:
* Setting last modified time on downloads finishing. (Also note that this has a little different functionality now where instead of failing and not deleting the file when it cannot set utime, it now warns and does not delete the file, which I think is more appropriate)
* Injecting content type
* Creating directories for downloads where there is no presently existing directory

I also did a small amount of refactoring in the current code base as I saw places where I could improve the logic (i.e. creating ``OnDoneFilteredSubscriber`` and moving more error handling logic into ``guess_content_type``)

I would really recommend reviewing each of these commits one by one.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -19,6 +19,7 @@ import math
 import errno
 import os
 import sys
+import time
 from collections import namedtuple, deque
 from functools import partial
 
@@ -142,6 +143,10 @@ class MD5Error(Exception):
     """
     Exception for md5's that do not match.
     """
+    pass
+
+
+class CreateDirectoryError(Exception):
     pass
 
 
@@ -785,6 +790,60 @@ class OnDoneFilteredSubscriber(BaseSubscriber):
 
     def on_failure(self, future, e):
         pass
+
+
+class BaseProvideContentTypeSubscriber(BaseSubscriber):
+    """A subscriber that provides content type when creating s3 objects"""
+
+    def on_queued(self, future, **kwargs):
+        guessed_type = guess_content_type(self._get_filename(future))
+        if guessed_type is not None:
+            future.meta.call_args.extra_args['ContentType'] = guessed_type
+
+    def _get_filename(self, future):
+        raise NotImplementedError('must implement _get_filename()')
+
+
+class ProvideUploadContentTypeSubscriber(BaseProvideContentTypeSubscriber):
+    def _get_filename(self, future):
+        return future.meta.call_args.fileobj
+
+
+class ProvideCopyContentTypeSubscriber(BaseProvideContentTypeSubscriber):
+    def _get_filename(self, future):
+        return future.meta.call_args.copy_source['Key']
+
+
+class ProvideLastModifiedTimeSubscriber(OnDoneFilteredSubscriber):
+    """Sets utime for a downloaded file"""
+    def __init__(self, last_modified_time, result_queue):
+        self._last_modified_time = last_modified_time
+        self._result_queue = result_queue
+
+    def on_success(self, future, **kwargs):
+        filename = future.meta.call_args.fileobj
+        try:
+            last_update_tuple = self._last_modified_time.timetuple()
+            mod_timestamp = time.mktime(last_update_tuple)
+            set_file_utime(filename, int(mod_timestamp))
+        except Exception as e:
+            warning_message = (
+                'Successfully Downloaded %s but was unable to update the '
+                'last modified time. %s' % (filename, e))
+            self._result_queue.put(create_warning(filename, warning_message))
+
+
+class DirectoryCreatorSubscriber(BaseSubscriber):
+    """Creates a directory to download if it does not exist"""
+    def on_queued(self, future, **kwargs):
+        d = os.path.dirname(future.meta.call_args.fileobj)
+        try:
+            if not os.path.exists(d):
+                os.makedirs(d)
+        except OSError as e:
+            if not e.errno == errno.EEXIST:
+                raise CreateDirectoryError(
+                    "Could not create directory %s: %s" % (d, e))
 
 
 class NonSeekableStream(object):
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -818,7 +818,7 @@ class BaseProvideContentTypeSubscriber(BaseSubscriber):
             future.meta.call_args.extra_args['ContentType'] = guessed_type
 
     def _get_filename(self, future):
-        raise NotImplementedError('must implement _get_filename()')
+        raise NotImplementedError('_get_filename()')
 
 
 class ProvideUploadContentTypeSubscriber(BaseProvideContentTypeSubscriber):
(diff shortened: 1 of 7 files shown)


tests/unit/customizations/s3/test_utils.py
```diff
@@ -228,6 +240,20 @@ def test_tell_and_seek(self):
...
 
+class TestGuessContentType(unittest.TestCase):
+    def test_guess_content_type(self):
+        self.assertEqual(guess_content_type('foo.txt'), 'text/plain')
+
+    def test_guess_content_type_with_no_valid_matches(self):
+        self.assertEqual(guess_content_type('no-extension'), None)
+
+    def test_guess_content_type_with_unicode_error_returns_no_match(self):
+        with mock.patch('mimetypes.guess_type') as guess_type_patch:
```
kyleknap: I really do not like when I have to patch stuff like this (as well in os.makedirs), but I am not sure if there is any better way to do this. I am up for suggestions.

awscli/customizations/s3/utils.py
```diff
@@ -760,6 +760,33 @@ def on_queued(self, future, **kwargs):
         future.meta.provide_transfer_size(self.size)
 
 
+class OnDoneFilteredSubscriber(BaseSubscriber):
```
jamesls: Is there a reason for this not to be in s3transfer?

It's a little weird to me to add new public methods to a base class that look like they come from s3transfer but actually don't.

It might even make sense to just move this directly into `BaseSubscriber`.

awscli/customizations/s3/utils.py
```diff
@@ -760,6 +779,87 @@ def on_queued(self, future, **kwargs):
...
+class BaseProvideContentTypeSubscriber(BaseSubscriber):
+    """A subscriber that provides content type when creating s3 objects"""
+
+    def on_queued(self, future, **kwargs):
+        guessed_type = guess_content_type(self._get_filename(future))
+        if guessed_type is not None:
+            future.meta.call_args.extra_args['ContentType'] = guessed_type
+
+    def _get_filename(self, future):
+        raise NotImplementedError('must implement _get_filename()')
```
jamesls: `raise NotImplementedError('_get_filename')`

awscli/customizations/s3/utils.py
```diff
@@ -760,6 +779,87 @@ def on_queued(self, future, **kwargs):
...
+            self._result_queue.put(create_warning(filename, warning_message))
+
+
+class DirectoryCreatorSubscriber(BaseSubscriber):
+    """Creates a directory to download if it does not exist"""
+    def on_queued(self, future, **kwargs):
+        d = os.path.dirname(future.meta.call_args.fileobj)
+        try:
+            if not os.path.exists(d):
+                os.makedirs(d)
```
jamesls: Does the transfer manager guarantee that there will only ever be one `on_queued` callback invoked at any given time (across all threads)?

There was a race condition in the original code where another thread also going down this code path would try the `os.makedirs` and fail because some other thread had come along and already created the directory in between the `exists()` check.

Is this no longer an issue?

tests/functional/s3/test_cp_command.py
```diff
@@ -211,19 +211,6 @@ def test_no_metadata_directive_for_non_copy(self):
         self.assertEqual(self.operations_called[0][0].name, 'PutObject')
         self.assertNotIn('MetadataDirective', self.operations_called[0][1])
 
-    def test_cp_succeeds_with_mimetype_errors(self):
```
jamesls: Why was this test removed?

awscli/customizations/s3/utils.py
kyleknap: Yeah actually I think I want to add it to ``s3transfer``. When I was writing these subscribers, I just found it incredibly annoying that I had to keep adding this logic so I can see it being useful for others. I did not add it to ``s3transfer`` because if I add it there, then it would be a public interface for ``s3transfer`` with the CLI at least I can hide it from other developers from using, but yet again ``s3transfer`` is a work in progress. I just want to put a little more design thought into it. It was also a little difficult getting cross dependency of a PR in ``s3transfer`` with this ongoing PR.

What do you think? I am thinking either adding a TODO to make sure I backport this to ``s3transfer``, or do the backport now, wait till s3transfer pr gets merged, then update this PR. I think I am fine with either as both have their pros and cons.

I could also make the methods private if it is just the public part that you find strange.

awscli/customizations/s3/utils.py
kyleknap: Shoot forgot about that.

awscli/customizations/s3/utils.py
kyleknap: It is an issue. ``on_queued`` happens in the submission executor so it could still happen unless we drop the concurrency of the submission executor to one, but I am not sure we want to do that.

tests/functional/s3/test_cp_command.py
kyleknap: I moved it to this test: https://github.com/aws/aws-cli/pull/2117/files#diff-25df2ee263c482c355f21087bad3b373R250

It seemed like a really poor functional test: we were patching mimetypes in a functional test that really should have been in a unit test. If you want to keep the test, I am not opposed to adding it back, but I was just trying to clean some code up.

awscli/customizations/s3/utils.py
jamesls: I think a TODO note along with marking it internal would be reasonable for now.

The main confusing part was the fact that it followed the same naming convention `on_<event>` as all the built in subscriber methods but was actually tied to a specific subclass.  If these were named `_on_success`, that would at least be a clue that this is different that the built in subscribers.

kyleknap: Ahh that makes sense. Will update.

tests/functional/s3/test_cp_command.py
jamesls: Yeah I saw that test but it's not exactly testing the same thing.  This test also ensures that cp succeeds when we can't get the mimetype.  I still think that's still important to test in some manner given this test was written in response to a bug fix.

awscli/customizations/s3/utils.py
jamesls: Yeah I'd prefer to just account for the fact that the directory could have been created between the `exists()` and `makedirs()` check.

tests/functional/s3/test_cp_command.py
kyleknap: Alright I will add it back then.

jamesls:
Looks good, just had a few comments.

kyleknap:
@jamesls I updated. Let me know if I missed anything.

jamesls:
:shipit:  Looks good to me.

PR #2119, opened 2016-08-19
Use better random char generation for tests

jamesls:
We still occassionally run into issues where we're creating
pre-existing bucket names.  This gives a consistent interface
for creating random chars in all our tests.

cc @kyleknap @JordonPhillips

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -25,12 +25,12 @@ import copy
 import shutil
 import time
 import json
-import random
 import logging
 import tempfile
 import platform
 import contextlib
 import string
+import binascii
 from pprint import pformat
 from subprocess import Popen, PIPE
 
@@ -152,7 +152,7 @@ def temporary_file(mode):
 
     """
     temporary_directory = tempfile.mkdtemp()
-    basename = 'tmpfile-%s-%s' % (int(time.time()), random.randint(1, 1000))
+    basename = 'tmpfile-%s' % random_chars(8)
     full_filename = os.path.join(temporary_directory, basename)
     open(full_filename, 'w').close()
     try:
@@ -173,9 +173,7 @@ def create_bucket(session, name=None, region=None):
     if name:
         bucket_name = name
     else:
-        rand1 = ''.join(random.sample(string.ascii_lowercase + string.digits,
-                                      10))
-        bucket_name = 'awscli-s3test-' + str(rand1)
+        bucket_name = 'awscli-s3test-' + random_chars(10)
     params = {'Bucket': bucket_name}
     if region != 'us-east-1':
         params['CreateBucketConfiguration'] = {'LocationConstraint': region}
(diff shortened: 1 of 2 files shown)


JordonPhillips:
Why not use uuid and prepend 'awscli-s3test-'?

jamesls:
Because this is a more general purpose utility to give you control over the number of random chars you need vs. a fixed number in uuid.

It's also the same function used in botocore's tests, and I'd like to do a pass of botocore's tests to standardize on a single function as well.

We should have a standard way in tests to generate random identifiers.

kyleknap:
I wonder if it makes sense then to have one helper function that you use to generate a unique bucket name for all of the tests (i.e. have the logic ``"awscli-s3test-" + random_char(10)`` isolated to a helper method in testutils). So we do not have to update every spot if we need to update the logic again in the future and can ensure consistency?

Otherwise, I am fine with the approach to make the bucket names a bit more likely to be the same.

JordonPhillips:
@jamesls I get that, but I don't think we need that fine grained control. I would rather rely on a standards compliant uuid implementation than rolling our own thing that has features we don't need.

jamesls:
@kyleknap Fair point, I'll update.  I'd still like to keep the random_chars() because I would like to use this outside of s3 tests as well, but I can add a random_bucket_name() helper method and update the s3 tests to use the new function.

@JordonPhillips  I don't think the standards argument is relevant here.  It's literally calling `urandom` which is *the standard* way to generate random characters on a system. UUIDs are a higher level abstraction than just random chars, and in this particular case is unnecessary and has less capabilities.  Given their both one line functions, it makes sense to go with the one that has more capabilities so that if we ever do need more control, we don't have to go refactor this function again.

kyleknap:
Yeah agreed keep ``random_chars()`` but create a new helper that uses ``random_chars()`` for bucket name generation.

jamesls:
@kyleknap Added, ready for another look.

kyleknap:
Thanks. Looks good. :ship:

JordonPhillips:
Why would we need to control the specific number of characters?

jamesls:
Max length validation from services, APIs that require an exact length, etc.

jamesls:
I'm going to go ahead and merge this shortly.  I think all the concerns have been discussed:

* Using standards - `urandom` is a standard way of generating random bytes, albeit more low level than uuid
* Implementing features we don't need - The tests (even in this PR) are already using this capability of variable length.  Additionally, this has to be weighed against the tradeoffs.  The implicit assumption in that argument is that this approach comes with an unnecessary maintenance/support cost.  There's two counterpoints to that.  1) We're talking about one line functions.  This point would be valid if we were talking about 500 line roll-your-own vs. 1 line library call 2) This function has [been in botocore](https://github.com/boto/botocore/blob/develop/tests/__init__.py#L60) for over a year and has not had any additional maintenance burden.
* Consistency with botocore is a good thing.

Last call for comments, but I'd really prefer we move on to more pressing issues.

JordonPhillips:
I'm 0 on it at this point. 
⛵

PR #2120, opened 2016-08-19
Use BytesIO to fix py3 test failure

jamesls:
This test is skipped if you are running as root,
but on a dev machine, this test fails on python3.

Try this on py3: `nosetests tests/integration/customizations/s3/test_plugin.py:TestUnableToWriteToFile`

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

JordonPhillips:
⛵

Issue #2123, opened 2016-08-22
Support for colorized JSON output


alexejk:
It would be great if JSON output can be colorized to make viewing/working with JSON output easier. 

An example would be having all JSON keys in one color, strings values in another, and having number values in third. Very similar to what e.g http://www.bodurov.com/JsonFormatter/ would produce in terms of colors. Another example is: http://pygments.org/demo/5862267/

Possible one could simply approach coloring of JSON result by piping through e.g Pygments?
An example from SO (source: http://stackoverflow.com/questions/25638905/coloring-json-output-in-python)

```
from pygments import highlight, lexers, formatters
colorful_json = highlight(unicode(formatted_json, 'UTF-8'), lexers.JsonLexer(), formatters.TerminalFormatter())
print(colorful_json)
```

jamesls:
Thanks, I've added some thoughts/comments over in #2125.  Let's track the discussion over in the PR.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

PR #2125, opened 2016-08-22
Adding support for colorized JSON output

AlexejK:
As an attempt to provide support for #2123 this contains JSON colorization.
Basically the only difference is additional dependency and updates to the JsonFormatter to pipe through Pygments before writing out to the stream.

This PR also takes into consideration color flag for output and defaults to off for backwards compatibility.

There might be a more efficient way of doing this - to be frank, Python is not my strongest side.

### DIFF
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -85,15 +85,37 @@ class FullyBufferedFormatter(Formatter):
 
 class JSONFormatter(FullyBufferedFormatter):
 
+    def __init__(self, args):
+        super(JSONFormatter, self).__init__(args)
+
+        self.colorize = False
+        try:
+            if args.color == 'on':
+                self.colorize = True
+        except:
+            pass
+
     def _format_response(self, command_name, response, stream):
         # For operations that have no response body (e.g. s3 put-object)
         # the response will be an empty string.  We don't want to print
         # that out to the user but other "falsey" values like an empty
         # dictionary should be printed.
         if response != {}:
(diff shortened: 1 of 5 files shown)


jamesls:
Thanks for the pull request.  There's a previous discussion about this feature: https://github.com/aws/aws-cli/pull/10

Based on the feedback I've heard from people (including that thread):

* People are (understandably) hesitant to add big dependencies such as pygments.  Especially in this case because we're using such a small fraction of the available features needed from pygments
* I prefer not to have conditional behavior based on the existence of a package being installed or not (one of the original ideas in #10).
* If you do have pygments installed, you can already run `aws ec2 describe-instances | pygments -l json`.

Just to throw out another alternative, has anyone looked at how much work it is to write a colorized JSON lexer?  I think the assumption has been that it's too much code to be worth the trouble, but perhaps that's a reasonable compromise.  We already have `colorama` as a dep that we use to produce colorized output for `--output table`.

JordonPhillips:
As James mentioned, this isn't how we'd like to go about supporting this feature. I'm closing this out, but feel free to open a new PR that goes along with James's suggestions. Feature request opened at #2728.

aisipos:
Quote:
```I prefer not to have conditional behavior based on the existence of a package being installed or not (one of the original ideas in #10).```

How about if the dependency were conditional, but still required some command line flag to activate (eg, `--colorize`)? That way the output is predictable in the general case, but would show colorized output with the flag, or an error message indicating you to install pygments if it weren't installed.

PR #2126, opened 2016-08-22
Add Transfer Request Submitters

kyleknap:
This is the second to last PR that is needed to integrate ``cp`` and ``cp --recursive``. It also integrates the [PR involving subscribers](https://github.com/aws/aws-cli/pull/2117) so you will still see the diff from the that PR until it gets merged. I would recommend looking at the linked PR first but you can use this PR to understand how I am trying to integrate them.

The main reason for this abstraction is that the original ``S3Handler`` class was really large and had a lot of branching logic especially related to how enqueuing works which made it difficult to follow, difficult to expand, and difficult to test. It really seemed that it should have been abstraction by itself. By having the concept of individual transfer request submitters for each type of transfer, it helps solve these issues.

After this PR, I plan to make a version of ``S3Handler`` that glues all the result processor and the transfer submitters together and serves essentially as a dispatcher to the various transfer submitters depending on the fileinfo received.

Let me know what you think about this PR and my plan.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -26,9 +26,16 @@ from awscli.customizations.s3.executor import Executor
 from awscli.customizations.s3 import tasks
 from awscli.customizations.s3.transferconfig import RuntimeConfig, \
     create_transfer_config_from_runtime_config
+from awscli.customizations.s3.results import UploadResultSubscriber
+from awscli.customizations.s3.results import DownloadResultSubscriber
+from awscli.customizations.s3.results import CopyResultSubscriber
 from awscli.customizations.s3.utils import RequestParamsMapper
 from awscli.customizations.s3.utils import StdoutBytesWriter
 from awscli.customizations.s3.utils import ProvideSizeSubscriber
+from awscli.customizations.s3.utils import ProvideUploadContentTypeSubscriber
+from awscli.customizations.s3.utils import ProvideCopyContentTypeSubscriber
+from awscli.customizations.s3.utils import ProvideLastModifiedTimeSubscriber
+from awscli.customizations.s3.utils import DirectoryCreatorSubscriber
 from awscli.customizations.s3.utils import uni_print
 from awscli.compat import queue
 from awscli.compat import binary_stdin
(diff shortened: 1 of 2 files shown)


kyleknap:
Alright this PR is up to date now with the base of ``integrate-s3transfer`` should be good to take a look at. Let me know if you have any questions on the direction/design decisions. While it is a new abstraction, a lot of the changes are just adding tests for logic that we did not have unit tests for.

kyleknap:
Also for reference on how these will be used, here is the branch that actually integrates the transfer request submitters: https://github.com/kyleknap/aws-cli/commits/integrate-cp-v2. While that branch is pretty much done, I am trying to avoid sending a PR for that branch ight now because it depends on the other two PR's that I have out. All unit, functional, and integration test pass on the linked branch as well so it is fully functional if you want to try it out.

awscli/customizations/s3/s3handler.py
```diff
@@ -513,3 +521,187 @@ def _process_transfer(self, future):
...
+
+        :param transfer_manager: The underlying transfer manager
+        :param result queue: The result queue to use
+        :param cli_params: The associated CLI parameters passed in to the
+            command.
+        """
+        self._transfer_manager = transfer_manager
+        self._result_queue = result_queue
+        self._cli_params = cli_params
+        self._warning_hanlders = []
```
jamesls: Typo, but this appears not to be used.

awscli/customizations/s3/s3handler.py
```diff
@@ -513,3 +521,187 @@ def _process_transfer(self, future):
...
+    def __init__(self, transfer_manager, result_queue, cli_params):
+        """Submits transfer requests to the TransferManager
+
+        Given a FileInfo object and provided CLI parameters, it will add the
+        necessary extra arguments and subscribers in making a call to the
+        TransferManager.
+
+        :param transfer_manager: The underlying transfer manager
+        :param result queue: The result queue to use
+        :param cli_params: The associated CLI parameters passed in to the
```
jamesls: Would be helpful to document the type.  Is this the `Namespace` obj from argparse?

awscli/customizations/s3/s3handler.py
```diff
@@ -513,3 +521,187 @@ def _process_transfer(self, future):
...
+    def is_valid_submission(self, fileinfo):
+        """Checks whether it is valid to submit a particular FileInfo
+
+        :param fileinfo: The FileInfo to check if the transfer request
+            submitter can handle.
+
+        :returns: True if it can use the provided FileInfo to make a transfer
+            request to the underlying transfer manager. False, otherwise.
+        """
+        raise NotImplementedError('is_valid_to_submit()')
```
jamesls: `raise NotImplementedError('is_valid_submission')`

awscli/customizations/s3/s3handler.py
```diff
@@ -513,3 +521,187 @@ def _process_transfer(self, future):
...
+
+class UploadRequestSubmitter(BaseTransferRequestSubmitter):
+    REQUEST_MAPPER_METHOD = RequestParamsMapper.map_put_object_params
+    RESULT_SUBSCRIBER_CLASS = UploadResultSubscriber
+
+    def is_valid_submission(self, fileinfo):
+        return fileinfo.operation_name == 'upload'
+
+    def _add_additional_subscribers(self, subscribers, fileinfo):
+        subscribers.insert(0, ProvideSizeSubscriber(fileinfo.size))
```
jamesls: If there's a specific reason these need to be `.insert(0, )` instead of the typical `.append` it would be worth commenting why.

awscli/customizations/s3/s3handler.py
kyleknap: yeah I will remove it

awscli/customizations/s3/s3handler.py
kyleknap: It is a dictionary. I am still using the params dictionary provided to s3handler

awscli/customizations/s3/s3handler.py
kyleknap: It needs to be inserted because result subscribers pass size along and so the size needs to be provided before the result subscriber is called. I will add a comment.

awscli/customizations/s3/s3handler.py
```diff
@@ -513,3 +521,197 @@ def _process_transfer(self, future):
...
+                    self._result_queue.put(warning)
+                return True
+        return False
+
+
+class UploadRequestSubmitter(BaseTransferRequestSubmitter):
+    REQUEST_MAPPER_METHOD = RequestParamsMapper.map_put_object_params
+    RESULT_SUBSCRIBER_CLASS = UploadResultSubscriber
+
+    def is_valid_submission(self, fileinfo):
```
jamesls: Without seeing how these are going to be plumbed in, do you ever envision these being more that a dispatch based on the operation name in fileinfo?

I suspect the code's going to have some sort of linear scan through the available submitters util one returns `True`, but I wonder if a simple `operation_name: SubmitterCls` would be better.

kyleknap: I was considering doing it but I then tried to figure out how to handle streams because for streams you need to check ``operation_name`` and if ``is_stream``

kyleknap: Also reference this is how I am currently dispatching these: https://github.com/kyleknap/aws-cli/commit/d0f803a4df5b7b41d641f61ff3898d123a98f17c#diff-9498c040c28883ad2025c9c17119fdf4R558

Another edge case that I may run into is with ``mv``. For ``mv`` the operation is ``move`` so I need to know the direction to be able to resolve the completely correct transfer request submitter

jamesls: Yeah the O(n) I'm not too worried about because we shouldn't have too many of these (though I suppose it is wasteful to call `is_valid_submission` for (worst case) every possible submitter).  It's more the subtleties in the required ordering of submitters in order to ensure correct behavior.

FWIW, for streams would we not just have separate handlers (and therefore a separate dict of operation_name: SubmitterCls) for streams?  Not sure how you envision streams integrating, but I would assume we'd decide if we're in stream mode as high up the stack as possible and create the appropriate set of "stream" objects to use.  At that point, the rest of the code would all work the same.

jamesls:
Looks good to me, just had a few questions, but I'm also ok defering those until seeing a pull request that pulls everything together.

awscli/customizations/s3/s3handler.py
kyleknap: Yeah that could work for streams, but it still would not work for moves. I am not sure if I will be able to ever rely on an operation name to transfer request submitter. 

And yes with the design right now, I just need to make sure the streams are the highest up on the resolution chain.

In general if there was one big resolver chain I could see the following transfer request submitters:

* upload
* upload stream
* move upload
* download
* download stream
* move download
* copy
* copy download
* delete

So that is nine in all. Not sure if this will grow with like the addition of dryruns but I can easily seen nine at the least.

jamesls: If first on the list is highest priority wouldn't upload-stream go before upload?  If not then upload would return True even for an upload stream.  Either that or you have to make every handler "stream" aware and explicitly check that `operation_name == my_expected_operation and not stream_mode` which seems unfortunate for handlers that shouldn't have to care about anything stream related.

kyleknap: Oh that is not in order... If it was in order it would be:

    upload stream
    download stream
    move upload
    move download
    move copy
    upload
    download
    copy
    delete

Note that the only order that matters is that the stream logic is above the non streaming logic.

kyleknap:
@jamesls Yeah I think I would probably like to defer these questions till the PRs that integrates these into the various commands.

I think I have all of the smaller abstractions in place and I do not see those changing really. However, the gluing of them all together into a reasonable interface is something that may change over time. I am in favor of iterating on how they fit together if we start finding the design could be improved. It is just a little hard to tell what issues will be ran into if any in piecing these together. So I figure that starting from somewhere that is reasonable (from what can be told) and iterating until we figure out otherwise is a good approach.

Otherwise, I updated based on your previous comments.

jamesls:
:shipit:  Yeah seems fine to defer until the PR for integrating everything together.

PR #2130, opened 2016-08-23
Result updates based on CLI integration

kyleknap:
When working on integrating ``s3transfer`` into the ``cp`` there were a few more changes that I needed to make to get the functional/integration tests to pass:

1) Set the default out_file and error_file of the ResultPrinter to ``None`` instead of ``sys.stdout`` and ``sys.stdout`` and then set it when the ResultPrinter is instantiated. The problem with making it in the signature is that all of the functional tests could not patch out standard out and standard error because it happens after the class get defined. Note I was not too big of a fan of doing this but I am not sure if there is a better option to get the functional tests to pass with minimal changes to the functional tests.

2) Create an ``ErrorResult`` this is to handle the general exception case that is thrown anywhere outside of an individual transfer such as in the file generators

3) Noticed a duplication of the ``"warning: "`` prefix when printing. This is because ``create_warning`` already adds the "``warning: "`` prefix to the message it creates.

Let me know what you think.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -245,6 +245,8 @@ class ResultPrinter(BaseResultHandler):
     FAILURE_FORMAT = (
         '{transfer_type} failed: {src} to {dest} {exception}'
     )
+    # TODO: Add "warning: " prefix once all commands are converted to using
+    # result printer and remove "warning: " prefix from ``create_warning``.
     WARNING_FORMAT = (
         '{message}'
     )
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -54,6 +54,8 @@ SuccessResult = _create_new_result_cls('SuccessResult')
 
 FailureResult = _create_new_result_cls('FailureResult', ['exception'])
 
+ErrorResult = namedtuple('ErrorResult', ['exception'])
+
 CommandResult = namedtuple(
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/results.py
```diff
@@ -239,35 +246,40 @@ class ResultPrinter(BaseResultHandler):
         '{transfer_type} failed: {src} to {dest} {exception}'
     )
     WARNING_FORMAT = (
-        'warning: {message}'
+        '{message}'
```
jamesls: Would it make sense instead to move the `warning: ` and `error: ` into these format strings?

jamesls:
:shipit:  Had a question about format strings but otherwise looks good.

awscli/customizations/s3/results.py
kyleknap: I would actually prefer that. If I had the option I would:

1) Remove the prefix "warning: " from ``create_warning``
2) Add "warning: " to the format string

The only issue is that file generators still use ``create_warning`` and those warning still get delegated to the old stuff. How would a TODO right here sound?

Then for "error: " the only reason I did not do it was because it would deviate from existing behavior in the CLI. We do not prepend "error: " to these type of errors. I would be fine in adding it though if it makes sense?

kyleknap:
We talked about this but I added a todo note to update warning format once everything is using the new result printer and decided to not use "error: " because we currently do not use it, but we can add it if we want later. Merging.

PR #2131, opened 2016-08-23
Fix py2 test failure on windows

jamesls:
On python2 in windows you can't have environment variables
that are unicode types.  random_chars() returns unicode on py2,
which when added to 'tmpfile-%s' means the whole value is unicode.

Note: Windows on python3 you can't have env var values as bytes().

Verified failing tests now pass on windows.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #2132, opened 2016-08-23
Integrate cp with s3transfer

kyleknap:
This is the last pull request in integrating s3transfer with ``cp`` and ``cp --recursive``.

All of the test pass with minimal changes to the existing test suite so that is good. If you want to know why a particular test may have changed, I would recommend checking out the commit. I tried to be detailed on why I changed particular tests.

While this PR is smaller than some of my previous PR's, this is probably a good one to start trying to address any design feedback or ideas of the complete port of cp and how it affects ports of future commands.

For example, two larger design questions that I would like some conversation on include:

* How to manage submitters? Should these be injected into a general S3Handler? Should they be built into the S3Handler and you create various S3Handler subclasses? Or should there be one S3TransferHandler class with all of the submitters built in (which I chose because it is simplest given what is currently supported)?

* Thoughts on introducing builders for the various components in the future? For example having an S3TransferHandler builder and maybe eventually a FileGenerator, Filter, etc builders so that not all of this logic is not under CommandArchitecture. I only do that right now because the builder abstraction is not really needed as there is not too many different parameters that needs to be taken into account, but there may be in the future (you can already see it sort of forming with the result printer).

Let me know what you think.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -32,6 +32,7 @@ from awscli.customizations.s3.results import UploadResultSubscriber
 from awscli.customizations.s3.results import DownloadResultSubscriber
 from awscli.customizations.s3.results import CopyResultSubscriber
 from awscli.customizations.s3.results import ErrorResult
+from awscli.customizations.s3.results import CommandResult
 from awscli.customizations.s3.utils import RequestParamsMapper
 from awscli.customizations.s3.utils import StdoutBytesWriter
 from awscli.customizations.s3.utils import ProvideSizeSubscriber
@@ -45,9 +46,6 @@ from awscli.compat import binary_stdin
 
 
 LOGGER = logging.getLogger(__name__)
-
-CommandResult = namedtuple('CommandResult',
-                           ['num_tasks_failed', 'num_tasks_warned'])
 
 
 class BaseS3Handler(object):
@@ -526,8 +524,7 @@ class S3TransferStreamHandler(BaseS3Handler):
 
 
 class S3TransferHandler(object):
-    def __init__(self, transfer_manager, cli_params, result_queue,
-                 result_recorder, result_processor):
+    def __init__(self, transfer_manager, cli_params, result_command_recorder):
         """Backend for performing S3 transfers
 
         :type transfer_manager: s3transfer.manager.TransferManager
@@ -537,23 +534,23 @@ class S3TransferHandler(object):
         :param cli_params: The parameters passed to the CLI command in the
             form of a dictionary
 
-        :type result_queue: queue.Queue
-        :param result_queue: The queue to process results
-
-        :type result_recorder: ResultRecorder
-        :param result_recorder: Result recorder to get status of transfer
-
-        :type result_processor: ResultProcessor
-        :param result_processor: Result processor to process status updates
-            of transfer
+        :type result_command_recorder: ResultCommandRecorder
+        :param result_command_recorder: The result command recorder to be
+            used to get the final result of the transfer
         """
         self._transfer_manager = transfer_manager
-        self._result_queue = result_queue
-        self._result_recorder = result_recorder
-        self._result_processor = result_processor
+        # TODO: Ideally the s3 transfer handler should not need to know
+        # about the result command recorder. It really only needs an interface
+        # for adding results to the queue. When all of the commands have
+        # converted to use this transfer handler, an effort should be made
+        # to replace the passing of a result command recorder with an
+        # abstraction to enqueue results.
+        self._result_command_recorder = result_command_recorder
 
         submitter_args = (
-            self._transfer_manager, self._result_queue, cli_params)
+            self._transfer_manager, self._result_command_recorder.result_queue,
+            cli_params
+        )
         self._submitters = [
             UploadRequestSubmitter(*submitter_args),
             DownloadRequestSubmitter(*submitter_args),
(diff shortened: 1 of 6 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -940,6 +952,10 @@ def run(self):
         s3_stream_handler = S3TransferStreamHandler(
             self.session, self.parameters, result_queue=result_queue)
 
+        s3_transfer_hanlder = s3handler
```
jamesls: typo, hanlder

awscli/customizations/s3/subcommands.py
```diff
@@ -1008,6 +1024,37 @@ def run(self):
             rc = 2
         return rc
 
+    def _get_s3_transfer_handler(self, result_queue):
```
jamesls: Seems like it's worth pulling this into a separate class, both for testability, but also as a clean slate to start integrating more and more of the s3transfer logic over.

awscli/customizations/s3/s3handler.py
```diff
@@ -523,6 +525,80 @@ def _process_transfer(self, future):
...
+            )
+
+    def _shutdown(self):
+        self._result_queue.put(ShutdownThreadRequest())
+        self._result_processor.join()
+
+    def _enqueue(self, fileinfos):
+        for fileinfo in fileinfos:
+            for submitter in self._submitters:
+                if submitter.is_valid_submission(fileinfo):
```
jamesls: Seems like this isn't really about a "valid" submission.  It's more if the particular submitter understands the request / knows how to submit it.  What about names like `can_submit`, `can_handle`, etc instead of `is_valid_submission`?

awscli/customizations/s3/s3handler.py
```diff
@@ -523,6 +525,80 @@ def _process_transfer(self, future):
...
+
+        :type result_recorder: ResultRecorder
+        :param result_recorder: Result recorder to get status of transfer
+
+        :type result_processor: ResultProcessor
+        :param result_processor: Result processor to process status updates
+            of transfer
+        """
+        self._transfer_manager = transfer_manager
+        self._result_queue = result_queue
```
jamesls: Seems like there's a missing abstraction here.  You have a result queue, recorder, and processor, but at this level of abstraction, it seems too low level.  Especially in methods like `_shutdown()` where you have to coordinate how to shutdown the "result subsystem" by first enqueueing a shutdown request, then joining on a thread.  Seems like this can be packaged behind a class.

jamesls:
High level seems fine.  Still not entirely sold on the idea of iterating through every type of submitter for every single file we transfer, but I would agree that in its current form with only 3 submitters it's not really that big of a deal.

awscli/customizations/s3/subcommands.py
kyleknap: Cool. I was thinking that an abstraction was eventually going to be needed. Just was not sure when, but I'm totally with creating the abstraction now.

awscli/customizations/s3/s3handler.py
kyleknap: ``can_submit()`` is a better name I will do that.

awscli/customizations/s3/s3handler.py
kyleknap: Yeah it felt a little weird that I was passing in a result recorder and a result processor that already had been built with a result recorder to the initializer. How does an interface like this sound?
```py
class ResultProcessorManager:
    def __init__(self, result_queue, result_recorder, result_processor):
        self.result_queue = result_queue
    def start(self):
    def shutdown(self):
    def get_final_result(self):
    def __enter__(self):
    def __exit__(self, exc_type, *args):
```
That way all I would have to pass in is this result processor manager to the s3 transfer handler. If I do this, should I make it a separate PR? I feel like that may be nicer.

jamesls: Wouldn't you need a method to submit a result for processing if your'e no longer passing a `result_queue` to this this class?

Separate PR seems fine.

kyleknap: Possibly. So this is what I am planning. Let me know if that makes sense.

1) Change the interface of ``S3TransferHandler`` to:
```py
class S3TransferHandler(object):
    def __init__(self, transfer_manager, cli_params, result_queue):
```
2) Have the ``CommandResultRecorder`` live outside of ``S3TransferHandler`` because it eclipses more than the ``S3TransferHandler`` it covers the ``FileGenerators`` and any of the other abstractions. So I was going to have it do something like this:
```py
with command_result_recorder:
    fileinfos = filegenerator.call()
    s3_transfer_handler.call(fileinfos)
return command_result_recorder.get_command_result()
```

jamesls: Interesting.  I was going down the path of abstracting away results so that code doesn't have to deal directly with `result_queue`, which is a bit of an implementation detail.  Something like:

```
with command_result_recorder as r:
    fileinfos = ...
    s3_transfer_handler.call(fileinfos, r)

# In s3 transfer handler
def call(..., reporter):
    try:
       # ...
    except Exception as e:
        reporter.report_result(ErrorResult(exception=e))
```

Sounds like what you're shooting for is to just abstract away the stop/start/final_result part of results and keep submission as is. I think that's an improvement over what's there now, but I'm concerned that separating out the lifecycle of the reporting (start/stop) from the mechanism for reporting (result_queue) could lead to subtle bugs, e.g submission to results queue after result thread is shutdown.

kyleknap:
So I updated this PR based on the PR for CommandResultRecorder. So it would be good if https://github.com/aws/aws-cli/pull/2139 could get reviewed first before completing this one.

awscli/customizations/s3/s3handler.py
```diff
@@ -523,6 +525,118 @@ def _process_transfer(self, future):
...
+            result_queue, result_processor_handlers)
+        command_result_recorder = CommandResultRecorder(
+            result_queue, result_recorder, result_processor)
+
+        return S3TransferHandler(
+            transfer_manager, self._cli_params, command_result_recorder)
+
+    def _add_result_printer(self, result_recorder, result_processor_handlers):
+        result_printer = None
+        if self._cli_params.get('quiet'):
```
kyleknap: I wonder if this warrants having a ``ResultPrinterFactory``? It is near impossible to test this without patching or running somewhat of a functional test by seeing what gets printed out when the s3 transfer handler is called. We do have functional tests for this so I do not feel as bad for not having a unit test for this specific logic, but would be nice to unit test it.

kyleknap:
With the referenced PR merged and this one rebased. This should be good to look at again. I believe I have all of the previous feedback addressed.

awscli/customizations/s3/s3handler.py
jamesls: I think if we add any more result printer instances we should probably go ahead and create a separate `ResultPrinterFactory`.

jamesls:
:shipit:  Looks good to me.

awscli/customizations/s3/s3handler.py
kyleknap: Sounds like a plan!

PR #2133, opened 2016-08-24
Correct dir name in documentation about recursive copying to an S3 bucket

tedkornish:
I noticed this while reading the docs and, while it's entirely likely I'm wrong, it jumped out at me as a typo since I'm not sure how copying to `mybucket` would result in a file uploading to `mybucket2`. If this is unwarranted feel free to close the PR, but if not it could save someone else a couple minutes of confusion.


jamesls:
:shipit:  Thanks for catching this.

kyleknap:
:ship: Merging.

PR #2136, opened 2016-08-25
Add support for recording progress of unknown size

kyleknap:
I started looking into integrating the upload and download stream implementations into the cp integration: https://github.com/aws/aws-cli/pull/2132 and I ran into this as something I need to do the integration. I would recommend looking at the referenced PR first as while this PR is isolated from the referenced PR changes, that one is more important.

This is specifically needed for streaming uploads where the size may not be known a head of time but may be provided at any point. Currently the assumption is made that the total transfer size is known and is not ``None``, which will cause errors in trying to do arithmetic operation on ``None`` type objects. While it is not necessarily needed for supporting streaming uploads because we do not print out progress, it is useful if we ever wanted this statistic for something else like for the sake of tracking transfer speeds later on.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -163,12 +163,8 @@ class ResultRecorder(BaseResultHandler):
         self.expected_bytes_transferred = 0
         self.expected_files_transferred = 0
 
-        # Keys: result from _get_progress_cache_key()
-        # Values: number of bytes that have been transferred for that key
-        self._progress_cache = defaultdict(int)
-        # Keys: result from _get_progress_cache_key()
-        # Values: number of bytes that is expected for the particular transfer
-        self._total_size_cache = {}
+        self._ongoing_progress = defaultdict(int)
+        self._ongoing_total_sizes = {}
 
         self._result_handler_map = {
             QueuedResult: self._record_queued_result,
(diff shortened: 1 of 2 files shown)


awscli/customizations/s3/results.py
```diff
@@ -163,8 +164,11 @@ def __init__(self):
         self.expected_files_transferred = 0
 
         # Keys: result from _get_progress_cache_key()
-        # Values: number of bytes left for a particular transfer
-        self._progress_remaining_cache = {}
+        # Values: number of bytes that have been transferred for that key
+        self._progress_cache = defaultdict(int)
+        # Keys: result from _get_progress_cache_key()
```
jamesls: There is no longer a `_get_progress_cache_key()`.  

But more from a general coding practice standpoint, these comments aren't helpful.  I get that they key for the _process_cache dict is whatever is returned from `_get_process_cache_key()` because that's literally the name of the method.  I'd suggest either renaming these methods or just omitting these types of comments in general unless they're add insight not gleamed from reading the code.  They get out of date and don't help people reviewing the code.

awscli/customizations/s3/results.py
```diff
@@ -180,49 +184,78 @@ def __call__(self, result):
...
     def _record_progress_result(self, result, **kwargs):
-        self._progress_remaining_cache[
-            self._get_progress_cache_key(result)] -= result.bytes_transferred
-        self.bytes_transferred += result.bytes_transferred
+        bytes_transferred = result.bytes_transferred
+        self._account_for_updated_total_transfer_size_if_needed(result)
+        self._progress_cache[self._get_cache_key(result)] += bytes_transferred
+        self.bytes_transferred += bytes_transferred
+
+    def _account_for_updated_total_transfer_size_if_needed(self, result):
```
jamesls: This is a vague method name.  What specifically are you accounting for?  Can you rename it to be more clear?

tests/unit/customizations/s3/test_results.py
```diff
@@ -277,6 +291,54 @@ def test_multiple_progress_results(self):
...
 
+    def test_progress_result_with_no_known_transfer_size(self):
+        self.result_recorder(
+            QueuedResult(
+                transfer_type=self.transfer_type, src=self.src,
+                dest=self.dest, total_transfer_size=None
+            )
+        )
+
+        bytes_transferred = 1024 * 1024  # 1MB
```
jamesls: pep8: "Inline comments are unnecessary and in fact distracting if they state the obvious."

jamesls:
Small comments, but nothing major.

As an aside, now that I've seen the results.py module a few times, I'm finding the use of the `cache` names in the various dictionaries to be confusing.  They're not really a cache (i.e storing some fetched/computed result so subsequent uses don't have to refetch/recompute the value).  They're really used to keep track of progress and bytes transferred and these values get mutated on each `__call__`.  I have to keep reminding myself it's not actually a cache.

awscli/customizations/s3/results.py
kyleknap: Yeah I can remove these. I can also look to rename the name of variable/methods. How about?
``_progress_cache`` --> ``_ongoing_progress``
``_total_size_cache`` --> ``_ongoing_total_sizes``
 ``_get_cache_key()`` --> ``_get_result_key()``

awscli/customizations/s3/results.py
kyleknap: How about? ``_account_for_newly_discovered_total_transfer_size()`` Or is that too long?

kyleknap:
@jamesls. I think the naming conventions are a little bit better now with the update. Let me know what you think. I should have addressed all of the feedback.

jamesls:
:shipit:  I like the new names much better.

PR #2139, opened 2016-08-29
Add ComandResultRecorder

kyleknap:
The recorder uses the result queue, recorder, and processor to produce a ``CommandResult`` for the entirety of a command. This abstraction got brought up in this PR: https://github.com/aws/aws-cli/pull/2132. It seems that it will clean up the ``S3TransferHandler`` abstraction much more with this.

cc @jamesls @JordonPhillips


jamesls:
:shipit:  Looks good to me.

PR #2140, opened 2016-08-29
Add transfer benchmark script

JordonPhillips:
This adds a script that benchmarks all the cli transfer commands.
The commands are interleaved such that no additional setup is
necessary between command runs.

The weakness of this script is that it can't run individual commands,
though I think if we want that it would make sense as additional
scripts.

I also currently have it set to read from json defintions for cases,
but it may be wiser to restrict to a single test case per run and have
the additional arguments passed on the command line.

cc @jamesls @kyleknap

### DIFF
--- a/scripts/performance/benchmark-cp
+++ b/scripts/performance/benchmark-cp
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+from benchmark_utils import human_readable_to_bytes, summarize, clean
+from benchmark_utils import get_default_argparser, get_transfer_command
+from benchmark_utils import create_random_subfolder, benchmark_command
+
+
+def benchmark_cp(args):
+    destination = args.destination
+    if args.recursive:
+        destination = create_random_subfolder(destination)
+    command = 'cp %s %s' % (args.source, destination)
+    command = get_transfer_command(command, args.recursive, args.quiet)
+
+    def cleanup():
+        if not args.no_cleanup:
+            clean(destination, args.recursive)
+
+    benchmark_command(
+        command, args.benchmark_script,  args.summarize_script,
+        args.result_dir, args.num_iterations, args.dry_run,
+        cleanup=cleanup
+    )
+
+
+if __name__ == "__main__":
+    parser = get_default_argparser()
+    parser.add_argument(
+        '-s', '--source', required=True,
+        help='A local path or s3 path.'
+    )
+    parser.add_argument(
+        '-d', '--destination', required=True,
+        help='A local path or s3 path. A directory will be created in this '
+             'location to copy to in the case of a recursive transfer.'
+    )
+    benchmark_cp(parser.parse_args())
--- a/scripts/performance/benchmark-mv
+++ b/scripts/performance/benchmark-mv
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+from benchmark_utils import backup, copy, clean, get_default_argparser
+from benchmark_utils import create_random_subfolder, benchmark_command
+from benchmark_utils import get_transfer_command
+
+
+def benchmark_mv(args):
+    destination = args.destination
+    if args.recursive:
+        destination = create_random_subfolder(destination)
+    command = 'mv %s %s' % (args.source, destination)
+    command = get_transfer_command(command, args.recursive, args.quiet)
+    backup_path = backup(args.source, args.recursive)
+
+    def cleanup():
+        if not args.no_cleanup:
+            clean(destination, args.recursive)
+        clean(backup_path, args.recursive)
+
+    benchmark_command(
+        command, args.benchmark_script,  args.summarize_script,
+        args.result_dir, args.num_iterations, args.dry_run,
+        upkeep=lambda: copy(backup_path, args.source, args.recursive),
+        cleanup=cleanup
+    )
+
+
+if __name__ == "__main__":
+    parser = get_default_argparser()
+    parser.add_argument(
+        '-s', '--source', required=True,
+        help='A local path or s3 path.'
+    )
+    parser.add_argument(
+        '-d', '--destination', required=True,
+        help='A local path or s3 path. A directory will be created in this '
+             'location to move to in the case of a recursive transfer.'
+    )
+    benchmark_mv(parser.parse_args())
--- a/scripts/performance/benchmark_utils.py
+++ b/scripts/performance/benchmark_utils.py
@@ -1,0 +1,248 @@
+import s3transfer
+import os
+import subprocess
+import uuid
+import shutil
+import argparse
+import tempfile
+
+
+def summarize(script, result_dir, summary_dir):
+    """Run the given summary script on every file in the given directory.
+
+    :param script: A summarization script that takes a list of csv files.
+    :param result_dir: A directory containing csv performance result files.
+    :param summary_dir: The directory to put the summary file in.
+    """
+    summarize_args = [script]
+    for f in os.listdir(result_dir):
+        path = os.path.join(result_dir, f)
+        if os.path.isfile(path):
+            summarize_args.append(path)
+
+    with open(os.path.join(summary_dir, 'summary.txt'), 'wb') as f:
+        subprocess.check_call(summarize_args, stdout=f)
+
+
+def _get_s3transfer_performance_script(script_name):
+    """Retrieves an s3transfer performance script if available."""
+    s3transfer_directory = os.path.dirname(s3transfer.__file__)
+    s3transfer_directory = os.path.dirname(s3transfer_directory)
+    scripts_directory = 'scripts/performance'
+    scripts_directory = os.path.join(s3transfer_directory, scripts_directory)
+    script = os.path.join(scripts_directory, script_name)
+
+    if os.path.isfile(script):
+        return script
+    else:
+        return None
+
+
+def get_benchmark_script():
+    return _get_s3transfer_performance_script('benchmark')
+
+
+def get_summarize_script():
+    return _get_s3transfer_performance_script('summarize')
+
+
+def backup(source, recursive):
+    """Backup a given source to a temporary location.
+
+    :type source: str
+    :param source: A local path or s3 path to backup.
+
+    :type recursive: bool
+    :param recursive: if True, the source will be treated as a directory.
+    """
+    if source[:5] == 's3://':
+        parts = source.split('/')
+        parts.insert(3, str(uuid.uuid4()))
+        backup_path = '/'.join(parts)
+    else:
+        name = os.path.split(source)[-1]
+        temp_dir = tempfile.mkdtemp()
+        backup_path = os.path.join(temp_dir, name)
+
+    copy(source, backup_path, recursive)
+    return backup_path
+
+
+def copy(source, destination, recursive):
+    """Copy files from one location to another.
+
+    The source and destination must both be s3 paths or both be local paths.
+
+    :type source: str
+    :param source: A local path or s3 path to backup.
+
+    :type destination: str
+    :param destination: A local path or s3 path to backup the source to.
+
+    :type recursive: bool
+    :param recursive: if True, the source will be treated as a directory.
+    """
+    if 's3://' in [source[:5], destination[:5]]:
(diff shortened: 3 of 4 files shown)


jamesls:
It would be nice to improve the output to differentiate commands we're running vs. output from the tool.  I ran into an error trying out this command (not sure what), and the output makes it a little hard to track down:

```
aws s3 sync s3://bucket-name /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi
aws s3 cp s3://bucket-name/largefile /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
download: s3://bucket-name/largefile to ../../../../var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
aws s3 mv /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file s3://bucket-name/largefile
move: ../../../../var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file to s3://bucket-name/largefile
aws s3 sync s3://bucket-name /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi
aws s3 cp s3://bucket-name/largefile /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
download: s3://bucket-name/largefile to ../../../../var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
aws s3 mv /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file s3://bucket-name/largefile
move: ../../../../var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file to s3://bucket-name/largefile
aws s3 sync s3://bucket-name /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi
aws s3 cp /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file s3://bucket-name/largefile2

The user-provided path /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file does not exist.
aws s3 mv s3://bucket-name/largefile2 /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
An error occurred (404) when calling the HeadObject operation: Key "largefile2" does not exist
Completed 1 part(s) with ... file(s) remaining
aws s3 sync /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi s3://bucket-name
aws s3 rm s3://bucket-name/largefile2
An error occurred (404) when calling the HeadObject operation: Key "largefile2" does not exist
Completed 1 part(s) with ... file(s) remaining
aws s3 cp /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file s3://bucket-name/largefile2

The user-provided path /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file does not exist.
aws s3 mv s3://bucket-name/largefile2 /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file
An error occurred (404) when calling the HeadObject operation: Key "largefile2" does not exist
Completed 1 part(s) with ... file(s) remaining
aws s3 sync /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi s3://bucket-name
aws s3 rm s3://bucket-name/largefile2
An error occurred (404) when calling the HeadObject operation: Key "largefile2" does not exist
Completed 1 part(s) with ... file(s) remaining
aws s3 cp /var/folders/rb/4c1d0xbd18g_wxjcjlvb0whn1cjh5m/T/tmpDeFvAi/file s3://bucket-name/largefile2

```

scripts/performance/benchmark-s3-transfer
```diff
@@ -0,0 +1,243 @@
...
+        '--dry-run', default=False, action='store_true',
+        help='If set, commands will only be printed out, not executed.'
+    )
+    args = parser.parse_args()
+
+    script = args.benchmark_script
+    if script is None:
+        script = _get_benchmark_script()
+
+    if script is None:
```
jamesls: You could just fold all this into the `required=True`, and a custom `type=your_validator` in the `add_argument` call for `--benchmark-script`.

scripts/performance/benchmark-s3-transfer
```diff
@@ -0,0 +1,243 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'cases', help='A json file defining the test cases to run.'
+    ),
+    parser.add_argument(
+        '-b', '--benchmark-script',
+        help=('The benchmark script to run the commands with. This should be '
+              'from s3transfer.'))
+    parser.add_argument(
+        '-o', '--result-dir', default='results',
```
jamesls: This arg isn't honored.  Took me a while to figure out why I wasn't getting any results.

jamesls:
I keep getting the above error when I try this out.  I've tried several configurations (same bucket, different bucket), and it always seems to happen midway through.

I think I would prefer to make this more granular.  Perhaps the test case json files can specify which type of transfers they want tested so if you just want a small subset to test you don't have to test every single command per invocation.  It would also make troubleshooting easier.

JordonPhillips:
Looks like that error is due to the destination key being different from the source, which messes with the interleaving. I'll update it so only a destination bucket is provided.

scripts/performance/benchmark-cp
```diff
@@ -0,0 +1,93 @@
...
+            args.benchmark_script, command, '--output-file', out_file
+        ]
+        if not args.dry_run:
+            subprocess.check_call(benchmark_args)
+    
+    if not args.dry_run:
+        summarize(args.summarize_script, args.result_dir)
+
+        if not args.no_cleanup:
+            clean(args.destination, args.recursive)
```
jamesls: I think this should be wrapped in a `try/finally`.  I think we'd want to clean up even in the event that the bench mark scripts (or summarize scripts) fail.

scripts/performance/benchmark_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        subprocess.check_call(cp_args)
+        return
+
+    if recursive:
+        shutil.copytree(source, destination)
+    else:
+        shutil.copy(source, destination)
+
+
+def clean(destination, recursive):
```
jamesls: I think we need to be careful about this.  Especially in the `--recursive` case, if I had specified my `~/` directory as the dest and I was downloading objects from S3 that were broken down by prefix:

```
s3://bucket/foo
s3://bucket/bar
```

Then this script (with a source of `s3://bucket/`) would download to `~/foo` and `~/bar` if I was testing `s3->local`, but as part of cleanup it would `rm -rf` my entire home directory.

I think we should try to be more conservative in how my cleanup.  The ideal result is we only remove what we downloaded/uploaded, but I realize we would need to track that information somehow.  Is there a safer way we can clean up?

scripts/performance/benchmark-mv
```diff
@@ -0,0 +1,98 @@
...
+        if not args.dry_run:
+            subprocess.check_call(benchmark_args)
+            copy(backup_path, args.source, args.recursive)
+    
+    if not args.dry_run:
+        summarize(args.summarize_script, args.result_dir)
+
+        if not args.no_cleanup:
+            clean(args.destination, args.recursive)
+        clean(backup_path, args.recursive)
```
jamesls: This applies in general to all the calls to `clean()`.  I think we should make a best effort to cleanup even in the case of errors.

scripts/performance/benchmark-mv
```diff
@@ -0,0 +1,98 @@
...
+    
+
+def main():
+    parser = argparse.ArgumentParser()
```
jamesls: This looks like an exact duplicate in the other `benchmark-*` scripts.  Seems like a candidate for moving into `benchmark_utils`.

jamesls:
Overall seems to be working for me.  My only high level comment would be to move more stuff over to `benchmark_utils` to cut down on the duplication between scripts.

kyleknap:
As a heads up before I start commenting, I am trying your scripts out right now and I am going to make comments as I run into things so if I do not comment for a while there is a chance that I have more. I will let you know when I have no more comments.

scripts/performance/benchmark_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            return int(value)
+        except ValueError:
+            raise ValueError("Invalid size value: %s" % value)
+    else:
+        multiplier = SIZE_SUFFIX[suffix]
+        return int(value[:-len(suffix)]) * multiplier
+
+
+def summarize(script, directory, filename='summary.txt'):
```
kyleknap: I am not sure that I am a fan of adding the ``summarize.txt`` file because if I ever want to run ``summarize`` on that directory again, I would not be able to by doing ``/summarize my-results/*`` assuming my-results is the output directory. I would prefer the directory is reserved to the CSV's.

scripts/performance/benchmark_utils.py
JordonPhillips: We could create our own directory within the specified directory to isolate things, but that defeats half of the point of being able to specify your own location. We could list all the objects that would be copied, build those file paths, and then delete them. It would take much longer, but it would be safer.

JordonPhillips:
This is ready to look at again. I've addressed all the comments and also removed the monolithic benchmark script since its marginal utility is minuscule with the individual scripts in place.

scripts/performance/benchmark_utils.py
```diff
@@ -0,0 +1,176 @@
...
+def get_benchmark_script():
+    return _get_s3transfer_performance_script('benchmark')
+
+
+def get_summarize_script():
+    return _get_s3transfer_performance_script('summarize')
+
+
+def backup(source, recursive):
+    if source[:5] == 's3://':
```
kyleknap: This looks similar to ``get_random_subfolder``. Could we reuse that at all?

scripts/performance/benchmark_utils.py
```diff
@@ -0,0 +1,176 @@
...
+    'gb': 1024 ** 3,
+    'tb': 1024 ** 4,
+    'kib': 1024,
+    'mib': 1024 ** 2,
+    'gib': 1024 ** 3,
+    'tib': 1024 ** 4,
+}
+
+
+def human_readable_to_bytes(value):
```
kyleknap: This function is no longer used.

kyleknap:
To capture the offline discussion, I think we should consolidate a lot of this logic (especially when you iterate over and do the runs) into more general functions as it will allow us to easily update these scripts if we want to do something like add more args or fix a bug.

JordonPhillips:
Updated. Not 100% on the cleanup / upkeep bits, what do y'all think?

scripts/performance/benchmark_utils.py
```diff
@@ -0,0 +1,288 @@
...
+        subprocess.check_call(rm_args)
+    else:
+        if recursive:
+            shutil.rmtree(destination)
+        else:
+            os.remove(destination)
+
+
+def get_random_subfolder(destination):
+    """Get a random subdirectory of a given directory."""
```
jamesls: Might be worth clarifying/renaming this to make it clear we will *create* a new random subfolder under the provided destination for local files (vs. randomly picking an existing subdirectory).

jamesls:
:shipit:  Looks good to me.  I've been trying this out locally and it's been working great.

kyleknap:
Nice that cleaned it up a good amount. Looks good. I think ``human_readable_to_bytes`` is still around so you can delete that. But otherwise :ship:

PR #2143, opened 2016-08-31
Integrate streaming with new S3TransferHandler architecture

kyleknap:
Now that the architecture for cp is all put in place, I am starting to move functionality over to the new architecture. In this PR, I moved all of the functionality for uploading from and downloading to a stream to the new ``S3TransferHandler`` and removed code that was no longer needed.

Note that I am now making the assumption that ``s3transfer`` will now handle adjusting chunksize for the proper amount of parts.

Also I found a bug where we were printing streaming errors to stdout instead of stderr. Moving over to the new architecture fixed that and I updated the functional tests appropriately because they were asserting for the wrong behavior as well.

cc @jamesls @JordonPhillips


jamesls:
:shipit:  Looks good to me.

PR #2146, opened 2016-09-01
Explicitly translate supported confi values to TransferConfig

jamesls:
The previous implementation made the open-ended assumption that
runtime config translates 1-1 to transfer configs.
This also had the (unintended) side effect that everything in
transfer config was implicitly supported in the ~/.aws/config
file because the config is passed directly to `TransferConfig`.

This also generates errors if you pass an unknown option to the
`TransferConfig`, both for nonsensical values as well as valid
values supported in `~/.aws/config` but not in `TransferConfig`.
I noticed this because I had `addressing_style=path`
in my `~/.aws/config` file:

```
$ aws s3 cp foo s3://bucket/foo

__init__() got an unexpected keyword argument 'addressing_style'
```

I've updated the code to have a fixed list of supported config values.
This means that we need to explicitly update this list as we decide
to support for transfer config options.

cc @kyleknap @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-s3-56735.json
+++ b/.changes/next-release/bugfix-s3-56735.json
@@ -1,0 +1,5 @@
+{
+  "category": "``s3``", 
+  "type": "bugfix", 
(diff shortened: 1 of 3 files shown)


kyleknap:
Looks good. :ship: assuming test pass.

jamesls:
Ok I've branched off of develop and set develop as the base branch.

kyleknap:
Still :ship:

PR #2147, opened 2016-09-01
Integrate rm with s3transfer

kyleknap:
This pull request integrates ``rm`` with ``s3transfer`` and the new architecture. All existing tests pass without any changes needed.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -195,10 +195,8 @@ class ResultRecorder(BaseResultHandler):
                 'Any result using _get_ongoing_dict_key must subclass from '
                 'BaseResult. Provided result is of type: %s' % type(result)
             )
-        key_components = [result.transfer_type, result.src]
-        if result.dest is not None:
-            key_components.append(result.dest)
-        return ':'.join(key_components)
+        return ':'.join(
+            str(el) for el in [result.transfer_type, result.src, result.dest])
 
     def _pop_result_from_ongoing_dicts(self, result):
         ongoing_key = self._get_ongoing_dict_key(result)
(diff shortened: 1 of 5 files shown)


awscli/customizations/s3/results.py
```diff
@@ -326,16 +341,19 @@ def _print_noop(self, **kwargs):
 
     def _print_success(self, result, **kwargs):
         success_statement = self.SUCCESS_FORMAT.format(
-            transfer_type=result.transfer_type, src=result.src,
-            dest=result.dest)
+            transfer_type=result.transfer_type,
+            direction=self._get_direction(result)
```
jamesls: Can we rename `direction`?  "one way direction" sort of implies transfer in one direction, i.e local->s3.  Wasn't obvious to me what that meant.  What about something like `transfer_description`, `transfer_src_dest`, `transfer_locations`, etc.

awscli/customizations/s3/results.py
```diff
@@ -186,7 +195,10 @@ def _get_ongoing_dict_key(self, result):
                 'Any result using _get_ongoing_dict_key must subclass from '
                 'BaseResult. Provided result is of type: %s' % type(result)
             )
-        return ':'.join([result.transfer_type, result.src, result.dest])
+        key_components = [result.transfer_type, result.src]
```
jamesls: What about just `':'.join(str(el) for el in [...])` to avoid the `dest is None` special casing?

jamesls:
Pretty straightforward, looks good to me.  Had a few questions/comments.

awscli/customizations/s3/results.py
kyleknap: Yeah that works. I will update that.

awscli/customizations/s3/results.py
kyleknap: I like ``transfer_location``. I will do that.

kyleknap:
@jamesls it should be good to look at again. I think I addressed your feedback.

jamesls:
:shipit:

PR #2161, opened 2016-09-06
Move MB to a seperate command

JordonPhillips:
This will potentially need some work to integrate the result queue
if that is deemed necessary.

@kyleknap @jamesls

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -30,7 +30,7 @@ from awscli.customizations.s3.s3handler import S3Handler
 from awscli.customizations.s3.s3handler import S3TransferStreamHandler
 from awscli.customizations.s3.utils import find_bucket_key, uni_print, \
     AppendFilter, find_dest_path_comp_key, human_readable_size, \
-    RequestParamsMapper
+    RequestParamsMapper, split_s3_bucket_key
 from awscli.customizations.s3.syncstrategy.base import MissingFileSync, \
     SizeAndLastModifiedSync, NeverSync
 from awscli.customizations.s3 import transferconfig
@@ -750,12 +750,35 @@ class SyncCommand(S3TransferCommand):
                 [METADATA, METADATA_DIRECTIVE]
 
 
-class MbCommand(S3TransferCommand):
+class MbCommand(S3Command):
     NAME = 'mb'
     DESCRIPTION = "Creates an S3 bucket."
     USAGE = "<S3Uri>"
-    ARG_TABLE = [{'name': 'paths', 'nargs': 1, 'positional_arg': True,
-                  'synopsis': USAGE}]
+    ARG_TABLE = [{'name': 'path', 'positional_arg': True, 'synopsis': USAGE}]
+
+    def _run_main(self, parsed_args, parsed_globals):
+        super(MbCommand, self)._run_main(parsed_args, parsed_globals)
+
+        if not parsed_args.path.startswith('s3://'):
+            raise TypeError("%s\nError: Invalid argument type" % self.USAGE)
+        bucket, _ = split_s3_bucket_key(parsed_args.path)
+
+        bucket_config = {'LocationConstraint': self.client.meta.region_name}
+        params = {'Bucket': bucket}
+        if self.client.meta.region_name != 'us-east-1':
+            params['CreateBucketConfiguration'] = bucket_config
+
+        # TODO: Consolidate how we handle return codes and errors
+        try:
+            self.client.create_bucket(**params)
+            uni_print("make_bucket: %s\n" % bucket)
+            return 0
+        except Exception as e:
+            uni_print(
+                "make_bucket failed: %s %s\n" % (parsed_args.path, e),
+                sys.stderr
+            )
+            return 1
 
 
 class RbCommand(S3TransferCommand):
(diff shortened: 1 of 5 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -750,12 +750,35 @@ class SyncCommand(S3TransferCommand):
...
     USAGE = "<S3Uri>"
-    ARG_TABLE = [{'name': 'paths', 'nargs': 1, 'positional_arg': True,
-                  'synopsis': USAGE}]
+    ARG_TABLE = [{'name': 'path', 'positional_arg': True, 'synopsis': USAGE}]
+
+    def _run_main(self, parsed_args, parsed_globals):
+        super(MbCommand, self)._run_main(parsed_args, parsed_globals)
+
+        path = parsed_args.path
+        if path.startswith('s3://'):
```
jamesls: This isn't quite the exact same as the old behavior.  We require the user to provide the `s3://` prefix.

While we can discuss whether or not we should start allowing users to drop the `s3://` prefix where it's obviously implied, I don't think that should be the purpose of this PR.  The point here is simply to port this into it's own separate command *without* changing existing behavior.

jamesls:
I think there's also some cleanup that can happen in `CommandArchitecture` as it now longer needs to care about `make_bucket`.

awscli/customizations/s3/subcommands.py
kyleknap: +1

awscli/customizations/s3/subcommands.py
```diff
@@ -750,12 +750,35 @@ class SyncCommand(S3TransferCommand):
...
+            path = path[5:]
+        bucket, _ = find_bucket_key(path)
+
+        bucket_config = {'LocationConstraint': self.client.meta.region_name}
+        params = {'Bucket': bucket}
+        if self.client.meta.region_name != 'us-east-1':
+            params['CreateBucketConfiguration'] = bucket_config
+
+        try:
```
kyleknap: I would prefer we do not have branching logic from the ``ResultProcessor`` and ``CommandResultRecorder``. Would it be possible to use that to handle the printing and return code handling of a  command?

kyleknap:
Looks fine. I agree with James' comments as well. I had additional comments and questions as well.

awscli/customizations/s3/subcommands.py
JordonPhillips: Maybe, but it seems like overkill to spin up a new thread to watch a very short lived command.

kyleknap: That is true. So question. I wonder if we can make a lightweight version that uses the result recorder and result printer without having to spin up a thread and result queue? All I really want is that those two abstractions and probably the command result recorder being used to consolidate all of the logic.

JordonPhillips: It ended up not being that difficult to put in. I think it would add complexity to make a separate version just for this. I think we should either have it or not have it at all.

JordonPhillips:
updated

awscli/customizations/s3/subcommands.py
jamesls: I agree with @JordonPhillips on not having this at all.  This seems unnecessarily complex for what we need to happen specifically for `MbCommand`.

@kyleknap what's your motivation for using result processor / CommandResultRecorder?  We need to run a single command and print any errors that happen.  I can't see the original diff anymore, but that seems to just be a simple try/except clause is it not?

kyleknap: @jamesls this is the relevant diff/commit: https://github.com/aws/aws-cli/pull/2161/commits/4d99056fc8ea6500abf59e413e53945e4f4be613. So I do agree that the result processor, especially since a thread is spun up, is not really needed for this command. I think the main part of concern for me was that before with that commit we had two different places where we did the result printing logic and I was wondering if there was a way we could expose even just a helper function that consolidates this logic and have both the ``ResultPrinter`` and ``MbCommand`` can use it as if we need to update the printing logic ever, there is a good chance that we forget to update both places if we do not consolidate. Does that make sense?

jamesls: The logic is essentially:
```
try:
    some_api_call
    return success
except Exception as e:
    print_error(e)
    return failure
```

It's nice to not have to duplicate those lines, but it doesn't seem like that much an issue.  Either way I'd say that the `ResultPrinter` and associated classes is not the right way to handle simple cases like `Mb` and we'd need some other shared abstraction as you suggest.

So the question would be whether or not creating this abstraction should block getting this merged.  I'm ok with adding a comment that we may want to circle back and try to consolidate the code later.

kyleknap: No it should not be a blocker. I can settle with a note/todo comment saying it would be something nice to add later down the round.

kyleknap:
Looks good. @JordonPhillips based on previous discussion you can just remove the commit you added to use CommandResultRecorder and add a comment about consolidation. And it should be good.

JordonPhillips:
Added the todo and squashed commits

kyleknap:
Looks good. Thanks :ship:. Before merging, let's talk about when/where to merge this as it does not directly rely on ``integrate-s3transfer`` branch so it can be merged into that branch or the ``develop``.

JordonPhillips:
Will merge as soon as tests pass

PR #2164, opened 2016-09-08
Cache result kwargs for storing params for Results

kyleknap:
Some of the values needed to create the results are expensive especially if the progress callback is happening many times. So we want to record as many as the result kwargs as we can so the ``on_progress`` callbacks become essentially looking up values in a dictionary, creating a ``Result`` object, and placing it on the result queue. From the CLI side of the slow performance, this is pretty beneficial by not having to call ``relative_path`` for every progress update for uploads and downloads. Profiling this it looks like it definitely helps out with that.

cc @jamesls @JordonPhillips

### DIFF
--- a/tests/unit/customizations/s3/test_results.py
+++ b/tests/unit/customizations/s3/test_results.py
@@ -108,7 +108,15 @@ class TestUploadResultSubscriber(BaseResultSubscriberTest):
         # Simulate a queue result (i.e. submitting and processing the result)
         # before processing the progress result.
         self.result_subscriber.on_queued(self.future)
-        self.get_queued_result()
+        self.assertEqual(
+            self.get_queued_result(),
+            QueuedResult(
+                transfer_type=self.transfer_type,
+                src=self.src,
+                dest=self.dest,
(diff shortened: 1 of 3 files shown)


tests/unit/customizations/s3/test_results.py
```diff
@@ -105,6 +105,12 @@ def test_on_queued(self):
         )
 
     def test_on_progress(self):
+        # Simulate a queue result (i.e. submitting and processing the result)
+        # before processing the progress result.
+        self.result_subscriber.on_queued(self.future)
+        self.get_queued_result()
```
jamesls: Should you assert_equal on the returned result?

jamesls:
Nothing major, had a comment on a test.  Otherwise, pretty straightforward.

tests/unit/customizations/s3/test_results.py
kyleknap: Yeah does not hurt.

kyleknap:
@jamesls updated. Should be good to look at again.

jamesls:
:shipit:

PR #2165, opened 2016-09-08
Remove some logging statements

kyleknap:
Logging every single result in the result queue has non-insignificant slow down on the transfer speed and pollutes the debug logs as well. So now just log if errors are encountered or the shutdown request is sent.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #2166, opened 2016-09-09
Switch from coveralls to codecov

jamesls:
We've been using this in botocore for about 3 months now
and I think it's been working really well.

This is the same config as used in botocore.

cc @kyleknap @JordonPhillips


JordonPhillips:
⛵

kyleknap:
:ship:

PR #2168, opened 2016-09-12
Remove make bucket from s3handler integration tests

JordonPhillips:
cc @kyleknap @jamesls


jamesls:
Did you run this test?  This doesn't make any sense why we'd only remove make_bucket.  This won't actually work will it?

JordonPhillips:
Updated

JordonPhillips:
Updated

kyleknap:
Looks good. :ship:

jamesls:
:shipit:

PR #2172, opened 2016-09-14
Don't print bytes remaining if none are expected

JordonPhillips:
Previously when you were doing a recursive rm you would get a progress result like so: `Completed 0 Bytes/0 Bytes with 5 file(s) remaining.` This was somewhat confusing since rm doesn't actually delete bytes. So this removes the bytes part of the message when there is no expected byte transfer. I've also removed the period from the end of the message to be more consistent with the other messages. So now it would look like this: `Completed 1 file(s) with 5 file(s) remaining`

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -290,9 +290,13 @@ class ResultRecorder(BaseResultHandler):
 
 
 class ResultPrinter(BaseResultHandler):
-    PROGRESS_FORMAT = (
+    _FILES_REMAINING = "{remaining_files} file(s) remaining"
+    BYTE_PROGRESS_FORMAT = (
         'Completed {bytes_completed}/{expected_bytes_completed} with '
-        '{remaining_files} file(s) remaining.'
+        + _FILES_REMAINING
+    )
+    FILE_PROGRESS_FORMAT = (
+        'Completed {files_completed} file(s) with ' + _FILES_REMAINING
     )
     SUCCESS_FORMAT = (
         '{transfer_type}: {transfer_location}'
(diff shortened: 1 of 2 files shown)


kyleknap:
I thought we wanted to do something like:
```
Deleted 243 files with 1000 file(s) remaining
```
where we tracked the total number of files because with what you have you only have if there is a bunch of files to delete the number of files remaining will fluctuate and not be too helpful so it would nice to have some exact number in the progress.

I would also be fine with making it:
```
Completed 243 files with 1000 file(s) remaining
```
if you are worried about the special casing for deletes in the progress statement

JordonPhillips:
I like that second option better. Updated. The one thing I wasn't sure about is what we want to call 'completed' files; do we include failed transfers? Right now I'm including everything.

kyleknap:
Yeah let's include everything. I include failed bytes as of now so including failed files makes sense.

awscli/customizations/s3/results.py
```diff
@@ -400,23 +404,33 @@ def _add_progress_if_needed(self):
...
             self._result_recorder.files_transferred
         )
 
         # Create the display statement.
-        progress_statement = self.PROGRESS_FORMAT.format(
-            bytes_completed=bytes_completed,
-            expected_bytes_completed=expected_bytes_completed,
-            remaining_files=remaining_files
-        )
+        if self._result_recorder.expected_bytes_transferred > 0:
```
kyleknap: Might be good to make a comment on the reasoning why we make this if statement (i.e. explain that some transfers have zero size and it would be strange to display 0/0 Bytes)

JordonPhillips:
It now reads `Completed 1 file(s) with 5 file(s) remaining`

awscli/customizations/s3/results.py
```diff
@@ -400,23 +404,33 @@ def _add_progress_if_needed(self):
...
+            expected_bytes_completed = human_readable_size(
+                self._result_recorder.expected_bytes_transferred)
+
+            progress_statement = self.BYTE_PROGRESS_FORMAT.format(
+                bytes_completed=bytes_completed,
+                expected_bytes_completed=expected_bytes_completed,
+                remaining_files=remaining_files
+            )
+        else:
+            files_completed = self._result_recorder.files_transferred
```
kyleknap: You do not need to take into account ``files_failed`` and ``files_warned``. ``files_transferred`` takes into account successful transfers and failed transfers. We do not want to be including warned files because those correlate directly to number of warnings seen (we probably should rename it but not for this PR) and you want to avoid double counting as a file could have produced a warning but also been transferred.

JordonPhillips:
Updated

kyleknap:
Thanks! :ship:

PR #2173, opened 2016-09-14
Improve output when fatal errors encountered

kyleknap:
Now if a fatal error is encountered it will print out the fatal error and stop printing out any more updates it may receive as it cleans up. It was really confusing when all of the transfers
cancelled by Cntrl-C would show up in results as failures. Here is an example of how this looks now:
```

$ aws s3 cp aws-cli s3://kyleknap-dump/aws-cli --recursive
upload: aws-cli/.changes/1.10.13.json to s3://kyleknap-dump/aws-cli/.changes/1.10.13.json
upload: aws-cli/.changes/1.10.14.json to s3://kyleknap-dump/aws-cli/.changes/1.10.14.json
upload: aws-cli/.changes/1.10.11.json to s3://kyleknap-dump/aws-cli/.changes/1.10.11.json
upload: aws-cli/.changes/1.10.10.json to s3://kyleknap-dump/aws-cli/.changes/1.10.10.json
upload: aws-cli/.changes/1.1.2.json to s3://kyleknap-dump/aws-cli/.changes/1.1.2.json
upload: aws-cli/.changes/1.1.1.json to s3://kyleknap-dump/aws-cli/.changes/1.1.1.json
upload: aws-cli/.changes/1.10.17.json to s3://kyleknap-dump/aws-cli/.changes/1.10.17.json
upload: aws-cli/.changes/1.10.18.json to s3://kyleknap-dump/aws-cli/.changes/1.10.18.json
upload: aws-cli/.changes/1.10.2.json to s3://kyleknap-dump/aws-cli/.changes/1.10.2.json
upload: aws-cli/.changes/1.10.19.json to s3://kyleknap-dump/aws-cli/.changes/1.10.19.json
upload: aws-cli/.changes/1.10.16.json to s3://kyleknap-dump/aws-cli/.changes/1.10.16.json
upload: aws-cli/.changes/1.10.20.json to s3://kyleknap-dump/aws-cli/.changes/1.10.20.json
upload: aws-cli/.changes/1.10.15.json to s3://kyleknap-dump/aws-cli/.changes/1.10.15.json
upload: aws-cli/.changes/1.1.0.json to s3://kyleknap-dump/aws-cli/.changes/1.1.0.json
upload: aws-cli/.changes/1.10.12.json to s3://kyleknap-dump/aws-cli/.changes/1.10.12.json
fatal error: keyboard interrupt                       
$ 
```
I decided not to include the cleaning up bit because now Cntrl-C's are really fast and is not needed plus I think it looks a little better.

Note that this PR builds off of this one: https://github.com/boto/s3transfer/pull/56. So tests may fail until the linked PR is merged.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -17,6 +17,7 @@ from collections import namedtuple
 from collections import defaultdict
 
 from s3transfer.exceptions import CancelledError
+from s3transfer.exceptions import FatalError
 from s3transfer.subscribers import BaseSubscriber
 
 from awscli.compat import queue
@@ -95,9 +96,9 @@ class BaseResultSubscriber(OnDoneFilteredSubscriber):
     def _on_failure(self, future, e):
         result_kwargs = self._on_done_pop_from_result_kwargs_cache(future)
         if isinstance(e, CancelledError):
-            error_result_cls = ErrorResult
-            if 'KeyboardInterrupt' in str(e):
-                error_result_cls = CntrlCResult
+            error_result_cls = CntrlCResult
+            if isinstance(e, FatalError):
+                error_result_cls = ErrorResult
             self._result_queue.put(error_result_cls(exception=e))
         else:
             self._result_queue.put(FailureResult(exception=e, **result_kwargs))
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -58,6 +58,8 @@ FailureResult = _create_new_result_cls('FailureResult', ['exception'])
 
 ErrorResult = namedtuple('ErrorResult', ['exception'])
 
+CntrlCResult = _create_new_result_cls('CntrlCResult', base_cls=ErrorResult)
+
 CommandResult = namedtuple(
     'CommandResult', ['num_tasks_failed', 'num_tasks_warned'])
 
@@ -93,7 +95,10 @@ class BaseResultSubscriber(OnDoneFilteredSubscriber):
     def _on_failure(self, future, e):
         result_kwargs = self._on_done_pop_from_result_kwargs_cache(future)
         if isinstance(e, CancelledError):
-            self._result_queue.put(ErrorResult(exception=e))
+            error_result_cls = ErrorResult
+            if 'KeyboardInterrupt' in str(e):
+                error_result_cls = CntrlCResult
+            self._result_queue.put(error_result_cls(exception=e))
         else:
             self._result_queue.put(FailureResult(exception=e, **result_kwargs))
 
@@ -316,6 +321,7 @@ class ResultPrinter(BaseResultHandler):
     ERROR_FORMAT = (
         'fatal error: {exception}'
     )
+    CNTRL_C_MSG = 'cancelled: ctrl-c received'
 
     SRC_DEST_TRANSFER_LOCATION_FORMAT = '{src} to {dest}'
     SRC_TRANSFER_LOCATION_FORMAT = '{src}'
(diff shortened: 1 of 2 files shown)


JordonPhillips:
⛵  once the dependent prs get merged / this gets rebased. This is much better than seeing a bunch of `upload failed` messages.

kyleknap:
Everything has been merged in s3transfer so this PR and this got rebased so this should be good to look at.

jamesls:
Code wise this looks fine, but I wonder if the terminology is confusing here.  The error messages make sense from an implementation details standpoint but I think we should show something more user friendly because:

* While it's `fatal` in the sense that it stops the transfer, it's perfectly normal to ctrl-c out of a transfer if you don't want it to keep going and don't want to convey to users that something's gone wrong
* keyboard interrupt is super specific to a python developer

What about something like "cancelled: ctrl-c received"?

kyleknap:
@jamesls I would be fine with "cancelled: ctrl-c received" as it is more universal and human readable. 

Do you think we should be using "cancelled" for general errors that stop the entire transfer as well such as ListObjects failing?

jamesls:
Well in this case I was thinking we want to distinguish between an "expected fatal error" where a customer has explicitly asked to shutdown (via ctrl-c) vs. a truly fatal error like the `ListObjects` failing, which is unexpected and should make it clear to the user that something unexpected went wrong.

kyleknap:
Makes sense I will add some logic to differentiate between the two then in the result printer. Does ``"cancelled:"`` and ``"fatal error:"`` work? Or do you want something more explicit like ``"unexpected fatal error:"``? I think I like the latter more.

kyleknap:
@jamesls I updated based on feedback so it should be good to look at.

awscli/customizations/s3/results.py
```diff
@@ -91,7 +94,13 @@ def _on_success(self, future):
 
     def _on_failure(self, future, e):
         result_kwargs = self._on_done_pop_from_result_kwargs_cache(future)
-        self._result_queue.put(FailureResult(exception=e, **result_kwargs))
+        if isinstance(e, CancelledError):
+            error_result_cls = ErrorResult
+            if 'KeyboardInterrupt' in str(e):
```
jamesls: This seems fragile, shouldn't `s3transfer` expose a better way to disambiguate between these?  What about a `UserCanelledError`?

kyleknap:
@jamesls I just updated to rely on the new ``FatalError`` for ``ErrorResults`` instead of looking at the message of the error.

awscli/customizations/s3/results.py
```diff
@@ -57,6 +59,8 @@ def _create_new_result_cls(name, extra_fields=None, base_cls=BaseResult):
 
 ErrorResult = namedtuple('ErrorResult', ['exception'])
 
+CntrlCResult = _create_new_result_cls('CntrlCResult', base_cls=ErrorResult)
```
jamesls: Small detail, but the error messages use "Ctrl" which I think is a more common way to shorten "control".  I'd prefer to either use CtrlCResult or just the full out "ControlCResult"

awscli/customizations/s3/results.py
```diff
@@ -310,8 +320,9 @@ class ResultPrinter(BaseResultHandler):
         '{message}'
     )
     ERROR_FORMAT = (
-        '{exception}'
+        'fatal error: {exception}'
     )
+    CNTRL_C_MSG = 'cancelled: ctrl-c received'
```
jamesls: same thing, `ctrl` vs. `cntrl`.

kyleknap:
Updated the spelling. Test pass. Merging.

PR #2175, opened 2016-09-15
Refactor rb

JordonPhillips:
This pr extracts rb into its own command. I was able to remove a ***lot*** of code by doing so, so I broke the removal into several commits for ease of browsing. I also refactored the rb functional tests to more closely match other functional tests and added in several tests that either weren't present or lived in other places.

One thing I ran across was that rb had a validation on the s3 uri that only applied when using `--force`. I preserved that behavior, but I think we should consider expanding it to always apply or otherwise remove it entirely.

I decided not to try to shoehorn it into the new s3 transfer handler because that required abusing task submitters and would spin up threads for a single operation (`rm` is called directly, so its threads will be shutdown before it returns).

cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -25,7 +25,7 @@ from awscli.customizations.s3.comparator import Comparator
 from awscli.customizations.s3.fileinfobuilder import FileInfoBuilder
 from awscli.customizations.s3.fileformat import FileFormat
 from awscli.customizations.s3.filegenerator import FileGenerator
-from awscli.customizations.s3.fileinfo import TaskInfo, FileInfo
+from awscli.customizations.s3.fileinfo import FileInfo
 from awscli.customizations.s3.filters import create_filter
 from awscli.customizations.s3.s3handler import S3Handler
 from awscli.customizations.s3.s3handler import S3TransferHandlerFactory
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -799,14 +799,11 @@ class RbCommand(S3Command):
             raise TypeError("%s\nError: Invalid argument type" % self.USAGE)
         bucket, key = split_s3_bucket_key(parsed_args.path)
 
+        if key:
+            raise ValueError('Please specify a valid bucket name only.'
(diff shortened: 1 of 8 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -818,9 +789,46 @@ class RbCommand(S3TransferCommand):
...
 
+    def _run_main(self, parsed_args, parsed_globals):
+        super(RbCommand, self)._run_main(parsed_args, parsed_globals)
+
+        if not parsed_args.path.startswith('s3://'):
+            raise TypeError("%s\nError: Invalid argument type" % self.USAGE)
+        bucket, key = split_s3_bucket_key(parsed_args.path)
+
+        if parsed_args.force:
+            # TODO: always validate or never validate
```
kyleknap: Yeah this is interesting. So after thinking about it, I think we should apply the validation to both for consistency reasons. For the scenario where the person specified``s3://bucket/key`` for ``rb``, it is probably most likely they meant to specify ``rm``, in which case the ``rb`` command will fail. The only way this changes behavior is if the bucket was empty in the first place.

awscli/customizations/s3/subcommands.py
```diff
@@ -818,9 +789,46 @@ class RbCommand(S3TransferCommand):
...
+            # This is currently placed here to preserve behavior with the
+            # previous implementation which only validated key existence when
+            # --force is provided.
+            if key:
+                raise ValueError('Please specify a valid bucket name only.'
+                                 ' E.g. s3://%s' % bucket)
+            self._force(parsed_args.path, parsed_globals)
```
kyleknap: What is your thoughts on moving the ``self._force()`` to within the try except? That way we get ``failed remove_bucket: `` prepended to the runtime error message provided in ``self._force()``

JordonPhillips: The only issue is that this changes the rc on the force failing from 255 to 1

JordonPhillips: Capturing offline discussion: We should keep the rc since we document it, but we'll prepend `remove_bucket failed:` to the error message.

PR #2178, opened 2016-09-19
Port dryrun to s3transferhandler

JordonPhillips:
cc @kyleknap @jamesls

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -31,6 +31,7 @@ from awscli.customizations.s3.results import UploadStreamResultSubscriber
 from awscli.customizations.s3.results import DownloadStreamResultSubscriber
 from awscli.customizations.s3.results import DeleteResultSubscriber
 from awscli.customizations.s3.results import CommandResult
+from awscli.customizations.s3.results import DryRunResult
 from awscli.customizations.s3.results import ResultRecorder
 from awscli.customizations.s3.results import ResultPrinter
 from awscli.customizations.s3.results import OnlyShowErrorsResultPrinter
@@ -43,6 +44,7 @@ from awscli.customizations.s3.utils import ProvideUploadContentTypeSubscriber
 from awscli.customizations.s3.utils import ProvideCopyContentTypeSubscriber
 from awscli.customizations.s3.utils import ProvideLastModifiedTimeSubscriber
 from awscli.customizations.s3.utils import DirectoryCreatorSubscriber
+from awscli.customizations.s3.utils import format_fileinfo_src_dest
 from awscli.compat import queue
 from awscli.compat import binary_stdin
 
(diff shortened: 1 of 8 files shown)


kyleknap:
Yeah looking at this again. It would be cool if just had a ``DryRunResult`` that we put onto the ``result_queue`` and let our handlers figure out how to deal with them. Trying to fabricate the entire transfer lifecycle and use a different ``ResultPrinter`` seems strange and a little fragile if we iterate on ``Results``.

JordonPhillips:
Complete rewrite using a `DryRunResult`, the code is much nicer.

kyleknap:
So I really like the ``DryRunResult`` approach more. The one thing that seems still a little strange for me is the public ``utils`` functions to get the correct format are only used in one class. It may be better to have it internal to the class. Furthermore by making it internal to the class, you can take advantage of the class knowing the needed logic ahead of time (i.e. you will not need an if statement in the middle of the public function for delete). I think it would be cool to have the various ``RequestSubmitters`` implement there own get src and dest implementations to provide to the ``DryRun`` result provided. Then I think a stretch objective to make it even better would be figure out a way to consolidate that with ``_get_src_dest()`` methods of the ``ResultSubscribers`` with what you are doing in the ``RequestSubmitters``. However, I am not sure how much consolidation can be done there. Let me know what you think?

JordonPhillips:
I looked into consolidation, but they're using very different inputs and the actual transformation is quite simple so it ends up being far more cumbersome to try to mesh them together.

Updated to move that logic into the request submitters.

awscli/customizations/s3/s3handler.py
```diff
@@ -635,6 +645,23 @@ def _warn_glacier(self, fileinfo):
                 return True
         return False
 
+    def _format_fileinfo_src_dest(self, fileinfo):
+        """Returns formatted versions of a fileinfos source and destination."""
+        src = self._format_path(fileinfo.src, fileinfo.src_type)
+        dest = self._format_path(fileinfo.dest, fileinfo.dest_type)
+        return src, dest
+
+    def _format_path(self, path, path_type):
```
kyleknap: So I like what you did by making this internal and then implementing the ``DeleteRequestSubmitter._format_path``, but I feel like you can leverage the fact that we have already resolved to what type of transfer we are about to do. So I can see the following happen:

* Make ``BaseTransferRequestSubmitter._format_path`` raise a ``NotImplementedError`` and have them implemented in the child classes.
* Create helper methods for specific types of formating (i.e. ``_format_local_path()`` and ``_format_s3_path()``
* Remove the ``if path == '-'`` check in the ``_format_local_path()`` and just have the request submitters for those stream just return ``-``.

While it is a bit more code, I think it is beneficial because we would be minimizing the coupling between the FileInfo's and RequestSubmitters. With having a general method that generates the formatted path, we have to rely on more of the values from the ``FileInfo`` such as ``path_type`` when we really do not need to because we already know the transfer type ahead of time. In the future, if we were to refactor ``FileInfo`` (as I think it currently does too much and too many abstractions are coupled to it), we will not have to worry about affecting this functionality.

Let me know if this make sense?

tests/unit/customizations/s3/test_utils.py
```diff
@@ -42,13 +42,13 @@
     ProvideUploadContentTypeSubscriber, ProvideCopyContentTypeSubscriber,
     ProvideLastModifiedTimeSubscriber, DirectoryCreatorSubscriber,
     NonSeekableStream, CreateDirectoryError)
+from awscli.customizations.s3.fileinfo import FileInfo
```
jamesls: Was this accidentally included?

PR #2180, opened 2016-09-19
Explode part list given to os.path.join

JordonPhillips:
os.path.join wants multiple string values, but not an iterable like
string.join wants.

cc @kyleknap @jamesls


jamesls:
:shipit:

PR #2183, opened 2016-09-20
Add ci benchmark script

JordonPhillips:
I added several command line arguments because I wanted to allow some flexibility in testing out the script in various setups while still allowing it to run with no arguments on jenkins.

cc @kyleknap @jamesls

### DIFF
--- a/scripts/ci/run-benchmark
+++ b/scripts/ci/run-benchmark
@@ -9,6 +9,7 @@ from subprocess import check_call
 from datetime import datetime
 import random
 import argparse
+import shutil
 
 TEST_BUCKET = os.environ.get('PERF_TEST_BUCKET')
 WORKDIR = os.environ.get('PERF_WORKDIR', 'workdir')
@@ -45,8 +46,11 @@ def main(args):
     run_id = generate_run_id()
     results_dir = os.path.join(WORKDIR, 'results', run_id)
     os.makedirs(results_dir)
-    benchmark(args.bucket, results_dir, args.num_iterations)
-    print("RUN ID: " + run_id)
+    try:
+        benchmark(args.bucket, results_dir, args.num_iterations)
+        print("RUN ID: " + run_id)
+    except Exception:
+        shutil.rmtree(results_dir)
 
 
 def benchmark(bucket, results_dir, num_iterations=1):
--- a/scripts/ci/run-benchmark
+++ b/scripts/ci/run-benchmark
@@ -51,6 +51,7 @@ def main(args):
         print("RUN ID: " + run_id)
     except Exception:
         shutil.rmtree(results_dir)
+        raise
 
 
 def benchmark(bucket, results_dir, num_iterations=1):
--- a/scripts/ci/run-benchmark
+++ b/scripts/ci/run-benchmark
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+"""Script to benchmark several high level cli commands.
+
+As of now this benchmarks `cp` and `rm` with test cases for multiple 4kb files
+(default 10000 files) and a single large file (default 10gb, `cp` only).
+"""
+import os
+from subprocess import check_call
+from datetime import datetime
+import random
+import argparse
+
+TEST_BUCKET = os.environ.get('PERF_TEST_BUCKET')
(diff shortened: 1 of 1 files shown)


scripts/ci/run-benchmark
```diff
@@ -0,0 +1,121 @@
...
+    run_id = generate_run_id()
+    results_dir = os.path.join('results', run_id)
+    os.makedirs(results_dir)
+    benchmark(args.bucket, results_dir, args.num_iterations)
+    print("RUN ID: " + run_id)
+
+
+def benchmark(bucket, results_dir, num_iterations=1):
+    perf_dir = os.path.dirname(os.path.realpath(__file__))
+    perf_dir = os.path.join(perf_dir, os.pardir, 'performance')
```
jamesls: You can also just call `os.path.dirname()` twice instead of using `pardir`:

```
perf_dir = os.path.join(
    os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
    'performance')
```

scripts/ci/run-benchmark
```diff
@@ -0,0 +1,121 @@
...
+(default 10000 files) and a single large file (default 10gb, `cp` only).
+"""
+import os
+from subprocess import check_call
+from datetime import datetime
+import random
+import argparse
+
+TEST_BUCKET = os.environ.get('PERF_TEST_BUCKET')
+MANY_FILES_DIR = "files/10k"
```
jamesls: I think we should have a well defined place where we put both the local files we generate as well as the results dir.  Right now, this location depends on where you run the script.

I've been using `<reporoot>/workdir`, but I'm not tied to a particular name.  I do think we should say that everything related to running the perf scripts should use a single top level directory from the repo root regardless of what directory you're in when you run the script.

scripts/ci/run-benchmark
```diff
@@ -0,0 +1,121 @@
...
+LARGE_FILE_DIR = "files/10gb"
```
jamesls: Worth using `os.path.join` so these will need less changes when we get these running on windows.  Though you'll need to account for this when creating the `S3` paths.

scripts/ci/run-benchmark
```diff
@@ -0,0 +1,121 @@
...
+    if not os.path.exists(LARGE_FILE_DIR):
+        os.makedirs(LARGE_FILE_DIR)
+        run('caf gen --file-size %s --max-files 1 --directory %s' %
+            (file_size, LARGE_FILE_DIR))
+
+
+def main(args):
+    initialize_files(args.num_files, args.large_file_size)
+    run_id = generate_run_id()
+    results_dir = os.path.join('results', run_id)
```
jamesls: Same as my previous comment.  Let's turn these into absolute paths that don't depend on what your CWD is.  I'd suggest using `<reporoot>/workdir/results`.

scripts/ci/run-benchmark
```diff
@@ -0,0 +1,127 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-n', '--num-iterations', type=int, default=10,
+        help='The number of times to run each test.'
+    )
+    parser.add_argument(
+        '-b', '--bucket', default=TEST_BUCKET,
```
jamesls: One small thing I noticed.  The arg is called "bucket" but also requires the `s3://` prefix.  I like having this just be the bucket name and have the script inject the `s3://` prefix where needed.  Same thing with the env var `TEST_BUCKET`

scripts/ci/run-benchmark
```diff
@@ -45,8 +46,11 @@ def main(args):
     run_id = generate_run_id()
     results_dir = os.path.join(WORKDIR, 'results', run_id)
     os.makedirs(results_dir)
-    benchmark(args.bucket, results_dir, args.num_iterations)
-    print("RUN ID: " + run_id)
+    try:
+        benchmark(args.bucket, results_dir, args.num_iterations)
+        print("RUN ID: " + run_id)
+    except Exception:
+        shutil.rmtree(results_dir)
```
jamesls: Shouldn't this exception still propagate up the stack?

PR #2184, opened 2016-09-20
Add script to upload benchmark results

JordonPhillips:
Like with the perf script itself I wanted to leave room to use locally with custom parameters while also being able to run without any arguments on jenkins.

cc @kyleknap @jamesls

### DIFF
--- a/scripts/ci/upload-benchmark
+++ b/scripts/ci/upload-benchmark
@@ -5,6 +5,7 @@ import argparse
 from datetime import datetime
 from subprocess import check_call
 
+WORKDIR = os.environ.get('PERF_WORKDIR', 'workdir')
 DEFAULT_BUCKET = os.environ.get('PERF_RESULTS_BUCKET')
 DATE_FORMAT = "%Y-%m-%d-%H-%M-%S-"
 
@@ -42,7 +43,7 @@ def _is_result_dir_format(directory):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        '-d', '--directory', default='results',
+        '-d', '--directory', default=os.path.join(WORKDIR, 'results'),
         help='The directory where test results were placed.')
     parser.add_argument(
         '-s', '--s3-location', default=DEFAULT_BUCKET,
--- a/scripts/ci/upload-benchmark
+++ b/scripts/ci/upload-benchmark
@@ -21,5 +21,8 @@ if __name__ == "__main__":
     parser.add_argument(
         '-s', '--s3-location', default=DEFAULT_BUCKET,
         required=DEFAULT_BUCKET is None,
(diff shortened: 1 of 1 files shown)


scripts/ci/upload-benchmark
```diff
@@ -0,0 +1,25 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-d', '--directory', default='results',
+        help='The directory where test results were placed.')
+    parser.add_argument(
+        '-s', '--s3-location', default=DEFAULT_BUCKET,
+        required=DEFAULT_BUCKET is None,
```
jamesls: Not a huge fan of conditionally required based on the presence of a global env var, but I know we use this in a few other places as well.  It makes `--help` output confusing because it'll depend on whether or not you have this env var set.

JordonPhillips: True. How about specifically mentioning the env var in the help? That would clear up some of the confusion while keeping the validation.

JordonPhillips: I wish argparse had better support for conditionally required params though.

jamesls: Yeah that's fine.

scripts/ci/upload-benchmark
```diff
@@ -0,0 +1,28 @@
...
+import argparse
+from subprocess import check_call
+
+DEFAULT_BUCKET = os.environ.get('PERF_RESULTS_BUCKET')
+
+
+def main(args):
+    source = args.directory
+    destination = args.s3_location
+    check_call('aws s3 cp %s %s' % (source, destination), shell=True)
```
jamesls: I was trying this out and I was expecting to be able to upload the results directory.  However there's no `--recursive` here.  What did you have in mind for how this script would be used?

JordonPhillips: Ah, I had just been uploading the summary, but that doesn't work with the defaults. Now that we have the summary json though, do we want to keep each individual result?

jamesls:
I was playing around this a bit more and I also notice we don't clean up old results directories from the `run-benchmark` script.  I'm not sure we want to just to a recursive copy of `results/` each time.  Given these are sorted (mostly) by time, what if this script just found the latest resultsdir and uploaded that to S3?  That way you just need the bucket-name/path.

Thoughts?

JordonPhillips:
How about just using `sync` instead of `cp`? That way you can still upload multiple runs without doing anything special but it won't upload the same thing multiple times.

scripts/ci/upload-benchmark
jamesls: I'd think so.  I'd also like to get the raw data files so we can do additional analysis/visualization.

jamesls:
:shipit:

PR #2195, opened 2016-09-23
Inital commit of perfcmp script

jamesls:
This compares the summary values for 2 perf runs.
Right now you have to manually download the results
files, but eventually I'd like to add a mode that can
download from an S3 bucket.

cc @kyleknap @JordonPhillips

### DIFF
--- a/scripts/performance/perfcmp
+++ b/scripts/performance/perfcmp
@@ -1,0 +1,153 @@
+#!/usr/bin/env python
+"""Compare 2 perf runs.
+
+To use, specify the local directories that contain
+the run information::
+
+    $ ./perfcmp /results/2016-01-01-1111/ /results/2016-01-01-2222/
+
+"""
+import os
+import json
+import argparse
+
+from colorama import Fore, Style
+from tabulate import tabulate
+
+
+class RunComparison(object):
+
+    MEMORY_FIELDS = ['average_memory', 'max_memory']
+    TIME_FIELDS = ['total_time']
+    # Fields that aren't memory or time fields, they require
+    # no special formatting.
+    OTHER_FIELDS = ['average_cpu']
+
+    def __init__(self, old_summary, new_summary):
+        self.old_summary = old_summary
+        self.new_summary = new_summary
+
+    def iter_field_names(self):
+        for field in self.TIME_FIELDS + self.MEMORY_FIELDS + self.OTHER_FIELDS:
+            yield field
+
+    def old(self, field):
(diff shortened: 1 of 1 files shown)


scripts/performance/perfcmp
```diff
@@ -0,0 +1,153 @@
...
+        if field.startswith('std_dev_'):
+            field = field[len('std_dev_'):]
+        if field in self.MEMORY_FIELDS:
+            return self._human_readable_size(value)[0]
+        elif field in self.TIME_FIELDS:
+            return '%-3.2f' % value
+        else:
+            return '%.2f' % value
+
+    def _human_readable_size(self, value):
```
kyleknap: It is a bit unfortunate that this is repeated in multiple places now, but I am not sure if it would be good idea to be using methods cross repository scripts. So it should be fine.

scripts/performance/perfcmp
```diff
@@ -0,0 +1,153 @@
...
+                  'std_dev',
+                  Style.BRIGHT + 'delta' + Style.RESET_ALL]
+        rows = []
+        for field in comp.iter_field_names():
+            row = [field, comp.old(field), comp.old_suffix(field),
+                   comp.old_stddev(field), comp.new(field),
+                   comp.new_suffix(field), comp.new_stddev(field)]
+            diff_percent = comp.diff_percent(field)
+            diff_percent_str = '%.2f%%' % diff_percent
+            if diff_percent < 0:
```
kyleknap: Cool you fixed that from before.

scripts/performance/perfcmp
```diff
@@ -0,0 +1,153 @@
...
+    summary_json = os.path.join(benchmark_dir, 'summary.json')
+    with open(summary_json) as f:
+        return json.load(f)
+
+
+def main():
+    parser = argparse.ArgumentParser(description='__doc__')
+    parser.add_argument('oldrunid', help='Path to old run idir')
+    parser.add_argument('newrunid', help='Local to new run dir')
+    parser.add_argument('--local', action='store_true')
```
kyleknap: Is this argument used?

jamesls: No I'll remove.  I originally wanted a local mode as well as a remote mode that would download from s3, but I've found that I just download the data from s3 and always have it local.  It might be worth adding the remote mode in the future, but running `s3 cp` ended up not being a big deal for me.

kyleknap: yeah that is what I figured after I reread your overview.

PR #2208, opened 2016-09-29
Use a different utime for windows

kyleknap:
We run into WindowsErrors we when we try to set the utime that low so instead if an error occurs set the last modified time to -1 which will raise the warning as well for Windows. Also note that we cannot just use -1 for the entire test because it will not propogate an invalid timestamp warning on linux and cause the test to fail.

Note that I sent this to the integrate-s3transfer branch so we can get some clean test runs on that for Windows. Let me know if you want me to change that.

cc @jamesls @JordonPhillips

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -97,6 +97,18 @@ def skip_if_windows(reason):
     return decorator
 
 
+def set_invalid_utime(path):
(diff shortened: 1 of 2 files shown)


kyleknap:
I can also confirm my latest changes work on windows.

PR #2219, opened 2016-10-04
Detect params that prefix builtins and fix existing cases

JordonPhillips:
Argparse automatically expands prefixes of arguments, so something like
`--end` could expand to `--endpoint-url`. The parameter in question would
never be able to be passed in, even if the full parameter is also used.

This expands shadow detection to include prefix detection.

This also fixes the existing case, `end` in `route53domains`. While it
was initially just going to be `end`, it was expanded to include `start`
to keep the parameter naming consistent. `start` will continue to be
suppliable in addition to `start-time`. `end` was never suppliable in
the first place, so there is no effor to keep the alias.

cc @kyleknap @jamesls

### DIFF
--- a/.changes/next-release/bugfix-route53domains-89897.json
+++ b/.changes/next-release/bugfix-route53domains-89897.json
@@ -1,0 +1,5 @@
+{
+  "category": "route53domains",
+  "type": "bugfix",
+  "description": "Rename `--end` to `--end-time` to fix a bug relating to argparse prefix expansion. Alias `--start` to `--start-time` to maintain a consistent interface while keeping the old parameter."
+}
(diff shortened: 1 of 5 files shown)


JordonPhillips:
@jamesls Yes. I probably should have put that commit in front so it would be easier to test, but here's the output of cherry-picking it onto develop:

```
F
======================================================================
FAIL: Verify no command params are shadowed or prefixed by the built in param.
----------------------------------------------------------------------
  File "/Users/phjordon/.pyenv/versions/3.5.1/envs/cli/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/phjordon/workspace/aws-cli/tests/integration/test_cli.py", line 72, in test_no_shadowed_builtins
    raise AssertionError('\n' + '\n'.join(errors))
AssertionError: 
Shadowing/Prefixing a top level option: route53domains.view-billing.end
-------------------- >> begin captured logging << --------------------
[...]
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 1 test in 2.122s

FAILED (failures=1)
```

PR #2220, opened 2016-10-04
Add test for downloading an empty object

kyleknap:
Related to: https://github.com/boto/s3transfer/pull/65

cc @jamesls @JordonPhillips

### DIFF
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -498,7 +498,7 @@ class TestCp(BaseS3IntegrationTest):
             self.assertEqual(f.read(), contents)
 
     def test_download_empty_object(self):
-        bucket_name = self.create_bucket()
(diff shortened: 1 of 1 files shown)


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -497,6 +497,18 @@ def test_download_encrypted_kms_object(self):
         with open(local_filename, 'r') as f:
             self.assertEqual(f.read(), contents)
 
+    def test_download_empty_object(self):
+        bucket_name = self.create_bucket()
```
jamesls: We should be able to just use `_SHARED_BUCKET` instead of creating a new bucket.

kyleknap: Right forgot about it

kyleknap:
@jamesls updated to use the shared bucket

Issue #2233, opened 2016-10-14
[awscli/1.11.4] `aws s3 sync` failed to delete any local files


yoheimuta:
I think this issue is a bug or a breaking compatibility change (not describing anywhere), so please response it 🙇 

I read https://github.com/aws/aws-cli/blob/develop/CONTRIBUTING.rst#reporting-issues.

> A description of the problem/suggestion.

Our aws s3 sync command could not work well as before, after upgrading awscli from 1.11.3(any previous ones) to 1.11.4.

> The specific AWS CLI commands you are running. Please include debug logs for these commands by appending the --debug option to each command. Be sure to remove any sensitive information from the debug logs.

```
# Contents between the bucket and the local are different.
$ aws s3 ls s3://mtburn-test # empty
$ ll /tmp/test
total 0
0 -rw-r--r--  1 great  wheel     0B 10 14 17:00 will_be_delete.txt

# Expected to delete a local file, but failed
$ aws s3 sync \
s3://mtburn-test \
/tmp/test \
--delete; echo $?
delete failed: s3:///tmp/test/will_be_delete.txt Parameter validation failed:
Invalid bucket name "": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$"
1
```

> The AWS CLI version you are using aws --version.

```
$ aws --version
aws-cli/1.11.4 Python/2.7.9 Darwin/15.6.0 botocore/1.4.62
```

- To be sure, I could reproduce this issue in Ubuntu 14.04 and MacOSX 10.11.6.

> The first thing an AWS CLI developer will do is try to reproduce the issue you are seeing, so try to reduce your issue to the smallest possible set of steps that demonstrate the issue. This will lead to quicker resolution of your issue.

```
# Create a test bucket
$ aws s3 mb s3://mtburn-test
make_bucket: mtburn-test

# Create a local file
$ touch /tmp/test/will_be_delete.txt
$ ll /tmp/test/
total 0
0 -rw-r--r--  1 great  wheel     0B 10 14 17:00 will_be_delete.txt

# Install v1.11.3
$ sudo pip install awscli==1.11.3
...
$ aws s3 sync \
s3://mtburn-test \
/tmp/test \
--delete; echo $?
delete: ../../../tmp/test/will_be_delete.txt
0

# Install v1.11.4
$ sudo pip install awscli==1.11.4
...
$ touch /tmp/test/will_be_delete.txt
$ aws s3 sync \
s3://mtburn-test \
[...]

PR #2234, opened 2016-10-14
Enable local file deletes from fileinfos

kyleknap:
This is specifically done for sync --delete downloads. The
implementation is a bit quirky because the functionality is needed
for the CLI but does not really belong in s3transfer. Essentially
what happens if a local delete is seen, the CLI will submit the
transfer request and immediately delete the file in the main thread
as a part of the submission. While it is not ideal to be
doing that in the main thread because it will slow down the process,
it is acceptable as deleting a local file is pretty quick.

Fixes https://github.com/aws/aws-cli/issues/2233

cc @jamesls @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-s3syncdelete-69638.json
+++ b/.changes/next-release/bugfix-s3syncdelete-69638.json
@@ -1,0 +1,5 @@
+{
+  "category": "``s3 sync --delete``", 
+  "type": "bugfix", 
+  "description": "Fix regression where ``--delete`` would not delete local files"
+}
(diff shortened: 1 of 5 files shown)


kyleknap:
Correction on my assumption. I did some testing on this approach where I did ``sync s3 local --force`` where the s3 bucket was empty and the local file had a bunch of files. I found that this implementation was significantly faster like. I was seeing about 20%. The reason why this is faster is because we completely go around all of the locking primatives that slow down the parallelism. Since removing the file is really quick, there is not really benefit in trying to parallelize it as the locking mechanisms slow the entire process. Thus this even further supports why this functionality should not exists in s3transfer.

PR #2235, opened 2016-10-14
Fix a failing windows tests

kyleknap:
Our windows tests were failing. Was inappropriately using the full filename for matching a filename that was printed out with the relative path.

cc @jamesls @JordonPhillips


kyleknap:
Hmm Travis is taking a long time to start a build. It may be having issues. I am going to merge this. I have run the tests on multiple machines just fine. So I am confident that the tests will pass. If there are issues, we will catch them in our other CI's.

PR #2239, opened 2016-10-19
Port mv to s3transfer

JordonPhillips:
This is dependant on boto/s3transfer#69

The one sticking point here is telling the transfer manager that the transfer is a move. Right now it's plumbed through the fileinfo object, but I don't like how much had to be done to get it there. Other options:

* Make the fileinfo do the translation and set that boolean so that it doesn't need to be plumbed through.
* Have the transfer handler do the translation and pass it in.
* Do no translation at all and base the RequestSubmitters on the source type and destinaiton type

cc @kyleknap @jamesls

### DIFF
--- a/tests/unit/customizations/s3/test_utils.py
+++ b/tests/unit/customizations/s3/test_utils.py
@@ -41,7 +41,8 @@ from awscli.customizations.s3.utils import (
     StdoutBytesWriter, ProvideSizeSubscriber, OnDoneFilteredSubscriber,
     ProvideUploadContentTypeSubscriber, ProvideCopyContentTypeSubscriber,
     ProvideLastModifiedTimeSubscriber, DirectoryCreatorSubscriber,
-    NonSeekableStream, CreateDirectoryError)
+    DeleteSourceObjectSubscriber, DeleteSourceFileSubscriber,
+    DeleteCopySourceObjectSubscriber, NonSeekableStream, CreateDirectoryError)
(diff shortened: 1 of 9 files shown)


JordonPhillips:
Performance testing uncovered a bug: not deleting directories

scripts/performance/benchmark-mv
```diff
@@ -17,10 +17,14 @@ def benchmark_mv(args):
             clean(destination, args.recursive)
         clean(backup_path, args.recursive)
 
+    def upkeep():
+        clean(args.source, args.recursive)
```
kyleknap: Wasn't this commit in the other PR that you merged in?

tests/unit/customizations/s3/test_utils.py
```diff
@@ -749,6 +750,85 @@ def test_on_queued_failure_propogates_clear_error_message(self):
...
+
+class TestDeleteSourceFileSubscriber(unittest.TestCase):
+    def setUp(self):
+        self.tempdir = tempfile.mkdtemp()
+        self.filename = os.path.join(self.tempdir, 'myfile')
+        call_args = FakeTransferFutureCallArgs(fileobj=self.filename)
+        meta = FakeTransferFutureMeta(call_args=call_args)
+        self.future = mock.Mock()
+        self.future.meta = meta
+        self.subscriber = DirectoryCreatorSubscriber()
```
kyleknap: I do not think you need this.

PR #2248, opened 2016-10-22
Replace IAM with STS in CreateSubscription

charliejahchan:
IAM uses GetUser to extract the account the ID of the caller.
That does not work if the caller is assuming a role.
STS allows us to GetCallerIdentity, regardless of whether the
caller is a user or a role.

### DIFF
--- a/.changes/next-release/bugfix-cloudtrail-2075.json
+++ b/.changes/next-release/bugfix-cloudtrail-2075.json
@@ -1,0 +1,5 @@
+{
+  "category": "``cloudtrail``", 
+  "type": "bugfix", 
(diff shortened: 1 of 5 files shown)


charliejahchan:
There should be no change in behaviour and that should not break backwards compatibility since this particular STS call does not appear to be validated against the caller's IAM policy.

Tried it with the following policy:
{
  "Action": [
    "sts:GetSessionToken"
  ],
  "Effect": "Deny",
  "Resource": "*"
}

and it worked just fine.

charliejahchan:
Also realized I messed up my earlier comment.
s/SessionToken/CallerIdentity/

PR #2257, opened 2016-10-26
Fix typo (?) in sync documentation

mltsy:
I'm not sure if this is the only place this documentation is defined, but I couldn't find it anywhere else.  I noticed the error at http://docs.aws.amazon.com/cli/latest/reference/s3/sync.html

I'm hoping this will fix it, but it's not clear to me how those pages are compiled... ?


JordonPhillips:
Excellent, nice catch!

PR #2276, opened 2016-11-04
Set exit code on powershell

JordonPhillips:
Currently aws.cmd will not return a valid exit status in powershell.
So `echo $?` will always evaluate to `True` and `echo $LASTEXITCODE`
will always return `0`. This adds an explicit exit step which makes
sure that the return code is set properly.


jamesls:
Oh, should mention, this is probably worth adding to the changelog.

PR #2287, opened 2016-11-18
Add support for aliases

kyleknap:
Important to note that the tests will not pass until this PR gets merged: https://github.com/boto/botocore/pull/1088

This pull request adds alias support to the CLI (like git aliases). To define an alias, you need an ``~/.aws/cli/alias`` file, which is an INI file that defines aliases and their values. Here is an example of an alias file:
```
[toplevel]
my-alias = ec2 describe-instances
```
With this alias now, a user can type into the CLI:
```
$ aws my-alias
```
which will translate to running (from the same process):
```
$ aws ec2 describe-instances
```

External aliases can also be defined which let users leverage functionality from their shell. So if the alias file had the following definition:
```
[toplevel]
ls = !ls
```
So if a user types the following into the CLI:
```
$ aws ls
```
it will translate to running (in a subprocess):
```
$ ls
```

Past these examples, here are some important features/limitations that should be noted:

* The ``~/.aws/cli/alias`` file only has support for the ``[toplevel]`` section. Any other sections, will be ignored.
* Aliases defined under the ``[toplevel]`` section will only be applied at the service command level.
* Aliases can defined using multiple lines for readability:
  ```
  [toplevel]
  my-alias = 
     ec2 describe-regions \
        --query Regions
  ls =
     f() {
        ls .
     }; f
  ```
* Global and operation-level parameters can be provided inside of the alias. However ``--debug`` and ``--profile`` are not supported as their value gets set higher up the CLI stack and affect many of the other global parameters. In the case that those parameters are provided, the CLI will throw an error.
* Arguments entered past the alias will be appended to the expanded alias value.
* For subcommand aliases (i.e. not prefixed with ``!``), they must reference a service command and that service command must exist.
* Aliases are undocumented. So if you type ``help`` after a subcommand alias it will proxy to the command it it references.

It is also important to that this PR does not add support for:
* Auto-completeing aliases
* Commands for setting and getting aliases

Those should come in subsequent pull requests.

### DIFF
--- a/awscli/alias.py
+++ b/awscli/alias.py
@@ -1,0 +1,206 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import logging
+import os
+import shlex
+import subprocess
+
+from botocore.configloader import raw_config_parse
+
+from awscli.utils import emit_top_level_args_parsed_event
+
+
+LOG = logging.getLogger(__name__)
+
+
+class InvalidAliasException(Exception):
+    pass
+
+
+class AliasLoader(object):
+    ALIAS_FILENAME = os.path.expanduser(
+        os.path.join('~', '.aws', 'cli', 'alias'))
+
+    def __init__(self, alias_filename=None):
+        """Interface for loading and interacting with alias file
+
+        :param alias_filename: The name of the file to load aliases from.
+            This file must be an INI file. If no filename is provided,
+            the value of ALIAS_FILENAME will be used.
+        """
+        self._filename = alias_filename
+        if alias_filename is None:
+            self._filename = self.ALIAS_FILENAME
+        self._aliases = self._load_aliases()
+        self._cleanup_alias_values()
+
+    def _load_aliases(self):
+        if os.path.exists(self._filename):
+            return raw_config_parse(
+                self._filename, parse_subsections=False)
+        return {'toplevel': {}}
+
+    def _cleanup_alias_values(self):
+        aliases = self.get_aliases()
+        for alias in aliases:
+            # Beginning and end line separators should not be included
+            # in the internal representation of the alias value.
+            aliases[alias] = aliases[alias].strip(os.linesep)
+
+    def get_aliases(self):
+        return self._aliases.get('toplevel', {})
+
+
+class BaseAliasCommand(object):
+    _UNDOCUMENTED = True
+
+    def __init__(self, alias_name, alias_value):
+        """Base class for alias command
+
+        :type alias_name: string
+        :param alias_name: The name of the alias
+
+        :type alias_value: string
+        :param alias_value: The parsed value of the alias. This can be
+            retrieved from `AliasLoader.get_aliases()[alias_name]`
+        """
+        self._alias_name = alias_name
+        self._alias_value = alias_value
+
+    def __call__(self, args, parsed_args):
+        raise NotImplementedError('__call__')
+
+    @property
+    def name(self):
+        return self._alias_name
+
+    @name.setter
+    def name(self, value):
+        self._alias_name = value
+
+
+class ServiceAliasCommand(BaseAliasCommand):
+    UNSUPPORTED_GLOBAL_PARAMETERS = [
+        'debug',
+        'profile'
+    ]
+
+    def __init__(self, alias_name, alias_value, session, command_table,
+                 parser):
+        """Command for a `toplevel` subcommand alias
+
+        :type alias_name: string
+        :param alias_name: The name of the alias
+
+        :type alias_value: string
+        :param alias_value: The parsed value of the alias. This can be
+            retrieved from `AliasLoader.get_aliases()[alias_name]`
+
+        :type session: botocore.session.Session
+        :param session: The botocore session
+
+        :type command_table: dict
+        :param command_table: The command table containing all of the
+            possible service command objects that a particular alias could
+            redirect to.
+
+        :type parser: awscli.argparser.MainArgParser
+        :param parser: The parser to parse commands provided at the top level
+            of a CLI command which includes service commands and global
+            parameters. This is used to parse the service commmand and any
+            global parameters from the alias's value.
+        """
+        super(ServiceAliasCommand, self).__init__(alias_name, alias_value)
+        self._session = session
+        self._command_table = command_table
+        self._parser = parser
+
+    def __call__(self, args, parsed_globals):
+        alias_args = self._get_alias_args()
+        parsed_alias_args, remaining = self._parser.parse_known_args(
+            alias_args)
+        self._update_parsed_globals(parsed_alias_args, parsed_globals)
+        # Take any of the remaining arguments that were not parsed out and
+        # prepend them to the remaining args provided to the alias.
+        remaining.extend(args)
+        LOG.debug(
+            'Alias %r passing on arguments: %r to %r command',
+            self._alias_name, remaining, parsed_alias_args.command)
+        # Pass the update remaing args and global args to the service command
+        # the alias proxied to.
+        return self._command_table[parsed_alias_args.command](
+            remaining, parsed_globals)
+
+    def _get_alias_args(self):
+        try:
+            alias_args = shlex.split(self._alias_value)
+        except ValueError as e:
+            raise InvalidAliasException(
+                'Value of alias "%s" could not be parsed. '
+                'Received error: %s when parsing:\n%s' % (
+                    self._alias_name, e, self._alias_value)
+            )
+
+        alias_args = [arg.strip(os.linesep) for arg in alias_args]
+        LOG.debug(
+            'Expanded subcommand alias %r with value: %r to: %r',
+            self._alias_name, self._alias_value, alias_args
+        )
+        return alias_args
+
+    def _update_parsed_globals(self, parsed_alias_args, parsed_globals):
+        global_params_to_update = self._get_global_parameters_to_update(
+            parsed_alias_args)
+        # Emit the top level args parsed event to ensure all possible
+        # customizations that typically get applied are applied to the
+        # global parameters provided in the alias before updating
+        # the original provided global parameter values
+        # and passing those onto subsequent commands.
+        emit_top_level_args_parsed_event(self._session, parsed_alias_args)
+        for param_name in global_params_to_update:
+            updated_param_value = getattr(parsed_alias_args, param_name)
+            setattr(parsed_globals, param_name, updated_param_value)
+
+    def _get_global_parameters_to_update(self, parsed_alias_args):
+        # Retrieve a list of global parameters that the newly parsed args
+        # from the alias will have to clobber from the originally provided
+        # parsed globals.
+        global_params_to_update = []
+        for parsed_param, value in vars(parsed_alias_args).items():
+            # To determine which parameters in the alias were global values
+            # compare the parsed parameter to its default as specified by
+            # the parser. If the parsed value differs from the default value
+            # that global parameter must have been provided.
+            if self._parser.get_default(parsed_param) != value:
+                if parsed_param in self.UNSUPPORTED_GLOBAL_PARAMETERS:
+                    raise InvalidAliasException(
+                        'Global parameter "--%s" detected in alias "%s" '
+                        'which is not support in subcommand aliases.' % (
+                            parsed_param, self._alias_name))
+                else:
+                    global_params_to_update.append(parsed_param)
+        return global_params_to_update
+
+
+class ExternalAliasCommand(BaseAliasCommand):
+    """Command for `toplevel` external alias"""
+    def __call__(self, args, parsed_globals):
+        command_components = [
+            self._alias_value.lstrip('!')
+        ]
+        command_components.extend(args)
+        command = ' '.join(command_components)
+        LOG.debug(
+            'Using external alias %r with value: %r to run: %r',
+            self._alias_name, self._alias_value, command)
+        return subprocess.call(command, shell=True)
--- a/awscli/alias.py
+++ b/awscli/alias.py
@@ -248,7 +248,27 @@ class ServiceAliasCommand(BaseAliasCommand):
 
 
 class ExternalAliasCommand(BaseAliasCommand):
-    """Command for `toplevel` external alias"""
+    def __init__(self, alias_name, alias_value, invoker=subprocess.call):
+        """Command for external aliases
+
+        Executes command external of CLI as opposed to being a proxy
+        to another command.
+
+        :type alias_name: string
+        :param alias_name: The name of the alias
+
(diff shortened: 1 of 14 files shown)


awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def _update_parsed_globals(self, parsed_alias_args, parsed_globals):
+        global_params_to_update = self._get_global_parameters_to_update(
+            parsed_alias_args)
+        # Emit the top level args parsed event to ensure all possible
+        # customizations that typically get applied are applied to the
+        # global parameters provided in the alias before updating
+        # the original provided global parameter values
+        # and passing those onto subsequent commands.
+        emit_top_level_args_parsed_event(self._session, parsed_alias_args)
```
kyleknap: I decided to go with the approach of re-emitting the top level args event because it is really the only reliable way to recreate the final value of parsed globals as many of them contain customizations that modify the value.

tests/functional/docs/test_help_output.py
```diff
@@ -432,3 +433,54 @@ def test_operation_help_command_has_note(self):
...
             'intended for testing purposes only.')
+
+
+class TestAliases(BaseAWSHelpOutputTest):
+    def setUp(self):
+        super(TestAliases, self).setUp()
+        self.files = FileCreator()
+        self.alias_file = self.files.create_file('alias', '[toplevel]\n')
+
+        # TODO: If an environment variable is ever exposed to
```
kyleknap: Really did not like that I had to do this. My only other option would have been to expose an environment variable or additional configuration. I am up for ideas on this.

tests/functional/test_alias.py
```diff
@@ -0,0 +1,147 @@
...
+        cmdline = 'ec2 list-buckets'
+        self.assert_params_for_cmd(cmdline, {})
+        self.assertEqual(len(self.operations_called), 1)
+        self.assertEqual(self.operations_called[0][0].name, 'ListBuckets')
+
+    def test_external_alias(self):
+        # The external alias is tested by using mkdir; a command that
+        # is universal for the various OS's we support
+        directory_to_make = os.path.join(self.files.rootdir, 'newdir')
+        self.add_alias('mkdir', '!mkdir %s' % directory_to_make)
```
kyleknap: So I tested this on Windows to make sure that command will work. But does anyone have strong feelings about using a subprocess in a functional test. I feel it is sort of a gray area, but if anyone feels strongly about it, I can move it to an integration test. I made it a functional test because those get run more than integration tests and otherwise I am not too sure how I could test this functionally without patching out the subprocess.call, which would leak implementation detail.

stealthycoin:
We had this conversation last week and I forgot to add it here.

How do we want to handle circular aliases and aliases that replace legitimate command names?

kyleknap:
@stealthycoin 

For both types of aliases, they will always override a builtin command if they shadow one.

For circular aliases i.e.:
```
[toplevel]

ec2 = ec2 --region=us-west-2
```
The alias will proxy to the original command so essentially forcing ec2 to always use us-west-2 as its region.

For the case of external aliases, there is not much we can do as we run it a separate subprocess. It would be very difficult to detect if there is a circular alias as the aliased value can potentially have any value that is valid for a particular user's shell.

kyleknap:
Looks like test now pass should be good with the dependent PR merged. Should be good to look at.

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class InvalidAliasException(Exception):
+    pass
+
+
+class AliasLoader(object):
+    ALIAS_FILENAME = os.path.expanduser(
```
jamesls: Given this is non-mutable, we can just make this the default value rather than having to check if it's `not None` in the `__init__`.

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+            return raw_config_parse(
+                self._filename, parse_subsections=False)
+        return {'toplevel': {}}
+
+    def _cleanup_alias_values(self):
+        aliases = self.get_aliases()
+        for alias in aliases:
+            # Beginning and end line separators should not be included
+            # in the internal representation of the alias value.
+            aliases[alias] = aliases[alias].strip(os.linesep)
```
jamesls: Do we have to preserve whitespace in addition to line separators?  Seems easier to just use `.strip()` without `os.linesep`.  It's still possible to have `\r` in files on posix.

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def get_aliases(self):
+        return self._aliases.get('toplevel', {})
+
+
+class BaseAliasCommand(object):
```
jamesls: I think we should still be inheriting from `CLICommand`.  This ensures we have a common interface across all command objects in the CLI, though I realize you may have to fake some of the properties like `.arg_table`.

awscli/alias.py
jamesls: I don't think this is safe.  This is making the assumption (and future requirement) that all connected event handlers are idempotent, which I don't think is the case.

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+        # Retrieve a list of global parameters that the newly parsed args
+        # from the alias will have to clobber from the originally provided
+        # parsed globals.
+        global_params_to_update = []
+        for parsed_param, value in vars(parsed_alias_args).items():
+            # To determine which parameters in the alias were global values
+            # compare the parsed parameter to its default as specified by
+            # the parser. If the parsed value differs from the default value
+            # that global parameter must have been provided.
+            if self._parser.get_default(parsed_param) != value:
```
jamesls: I don't follow how the conclusion is drawn that differing default values means the param was provided.  Or are you saying it does not matter if they provide the default value?

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class ExternalAliasCommand(BaseAliasCommand):
+    """Command for `toplevel` external alias"""
+    def __call__(self, args, parsed_globals):
+        command_components = [
+            self._alias_value.lstrip('!')
+        ]
+        command_components.extend(args)
+        command = ' '.join(command_components)
```
jamesls: You can pass a list directly to `subprocess.call`.

awscli/alias.py
```diff
@@ -0,0 +1,206 @@
...
+                else:
+                    global_params_to_update.append(parsed_param)
+        return global_params_to_update
+
+
+class ExternalAliasCommand(BaseAliasCommand):
+    """Command for `toplevel` external alias"""
+    def __call__(self, args, parsed_globals):
+        command_components = [
+            self._alias_value.lstrip('!')
```
jamesls: Is `lstrip()` what we want here?  Or do you only want to remove the first `!` char (`self._alias_value[1:]` if the first char is a `!`).

awscli/clidriver.py
```diff
@@ -77,6 +81,7 @@ def __init__(self, session=None):
         self._cli_data = None
         self._command_table = None
         self._argument_table = None
+        self._command_table_with_aliases = None
```
jamesls: Do we need to have both forms of command tables with/without aliases?  It would seems simpler to just update the existing command table, but I might be missing something here (is it related to docs?)

tests/functional/docs/test_help_output.py
jamesls: Why not just pass in the filename and replace the existing alias loader in CLIDriver?  Seems ok to make this an instance attr.

tests/functional/docs/test_help_output.py
```diff
@@ -432,3 +433,54 @@ def test_operation_help_command_has_note(self):
...
+    def add_alias(self, alias_name, alias_value):
+        with open(self.alias_file, 'a+') as f:
+            f.write('%s = %s\n' % (alias_name, alias_value))
+
+    def test_alias_not_in_main_help(self):
+        self.add_alias('my-alias', 'ec2 describe-regions')
+        self.driver.main(['help'])
+        self.assert_not_contains('my-alias')
+
+    def test_alias_proxies_to_service_help(self):
```
jamesls: We might have chatted about this, but is this what we want here?  Seems like we'd maybe want to display the fact that this is an alias command.  I can see this being confusing just showing the help page for the underlying service command.  It also is odd that the `help` command will only work if you're proxying a service/operation with no args (otherwise you might get a parse error), so you can't tell if `aws myalias help` will work without knowing what the alias is.

tests/functional/test_alias.py
```diff
@@ -0,0 +1,147 @@
...
+        super(TestAliases, self).setUp()
+        self.files = FileCreator()
+        self.alias_file = self.files.create_file('alias', '[toplevel]\n')
+
+        # TODO: If an environment variable is ever exposed to
+        # modify the alias file location, remove this patch
+        # and use that environment variable because patching of specific
+        # classes properties should be avoided for functional tests but
+        # there is not really any other way to hook in...
+        self.alias_file_location_patch = mock.patch(
```
jamesls: Same comment as above, seems easier to just swap in an AliasLoader.

tests/functional/test_alias.py
```diff
@@ -0,0 +1,147 @@
...
+    def add_alias(self, alias_name, alias_value):
+        with open(self.alias_file, 'a+') as f:
+            f.write('%s = %s\n' % (alias_name, alias_value))
+
+    def test_subcommand_alias(self):
+        self.add_alias('my-alias', 'ec2 describe-regions')
+        cmdline = 'my-alias'
+        self.assert_params_for_cmd(cmdline, {})
+        self.assertEqual(len(self.operations_called), 1)
+        self.assertEqual(self.operations_called[0][0].name, 'DescribeRegions')
```
jamesls: It would be nice to verify we're also calling the right service, though I realize there's only one `DesribeRegions` call.

tests/functional/test_alias.py
```diff
@@ -0,0 +1,147 @@
...
+                        'RegionName': 'us-east-1'
+                    }
+                ]
+            }
+        ]
+        cmdline = 'my-alias'
+        stdout, _, _ = self.assert_params_for_cmd(cmdline, {})
+        self.assertEqual(len(self.operations_called), 1)
+        self.assertEqual(self.operations_called[0][0].name, 'DescribeRegions')
+        self.assertEqual(stdout.strip('\n'), 'us-east-1')
```
jamesls: should this just be `.strip()` to avoid windows issues?

tests/functional/test_alias.py
```diff
@@ -0,0 +1,147 @@
...
+        self.run_cmd('mkdir')
+        self.assertTrue(os.path.isdir(directory_to_make))
+
+    def test_external_alias_then_additonal_args(self):
+        # The external alias is tested by using mkdir; a command that
+        # is universal for the various OS's we support
+        directory_to_make = os.path.join(self.files.rootdir, 'newdir')
+        self.add_alias('mkdir', '!mkdir')
+        self.run_cmd('mkdir %s' % directory_to_make)
+        self.assertTrue(os.path.isdir(directory_to_make))
```
jamesls: It would also be nice to have a posix specific test for the IIFE aliases (as they're one of the more powerful use cases for aliases so it would be good to verify they work).

tests/unit/test_alias.py
```diff
@@ -0,0 +1,432 @@
...
+        alias_cmd = BaseAliasCommand('alias-name', 'alias-value')
+        self.assertEqual(alias_cmd.name, 'alias-name')
+        alias_cmd.name = 'new-alias-name'
+        self.assertEqual(alias_cmd.name, 'new-alias-name')
+
+
+class TestServiceAliasCommand(unittest.TestCase):
+    def setUp(self):
+        self.alias_name = 'myalias'
+        self.session = mock.Mock()
```
jamesls: Can you update this to use a `spec=`?

tests/unit/test_alias.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def create_command_table(self, services):
+        command_table = {}
+        for service in services:
+            command_table[service] = mock.Mock()
```
jamesls: Update to use spec=

tests/unit/test_alias.py
```diff
@@ -0,0 +1,432 @@
...
+        )
+        if extra_params:
+            for extra_param in extra_params:
+                parser.add_argument('--'+extra_param)
+        return parser
+
+    def test_alias_with_only_service_command(self):
+        alias_value = 'myservice'
+
+        command_table = self.create_command_table(['myservice'])
```
jamesls: It might be helpful to show how the values propagate here (use `alias_value` instead of the hard coded string).

tests/unit/test_alias.py
```diff
@@ -0,0 +1,432 @@
...
+
+        with self.assertRaises(InvalidAliasException):
+            alias_cmd([], FakeParsedArgs(command=self.alias_name))
+
+
+class TestExternalAliasCommand(unittest.TestCase):
+    def test_run_external_command(self):
+        alias_value = '!ls'
+        alias_cmd = ExternalAliasCommand('alias-name', alias_value)
+        with mock.patch('subprocess.call') as subprocess_call:
```
jamesls: Can we just have the command take a reference to `subprocess.call/Popen`?

awscli/clidriver.py
```diff
@@ -120,6 +131,26 @@ def _build_builtin_commands(self, session):
...
+    def _build_command_table_with_aliases(self):
+        command_table_with_aliases = OrderedDict()
+        command_table = self._get_command_table()
+        parser = self._create_parser(command_table)
+
+        for command_name, command_obj in command_table.items():
+            command_table_with_aliases[command_name] = command_obj
+
+        alias_loader = AliasLoader()
+        for alias_name, alias_value in alias_loader.get_aliases().items():
```
kyleknap: How does a ``build_alias_table()`` method on the loader sound?

awscli/clidriver.py
```diff
@@ -120,6 +131,26 @@ def _build_builtin_commands(self, session):
                                                     service_name=service_name)
         return commands
 
+    def _build_command_table_with_aliases(self):
```
kyleknap: See my explanation to the private attribute.

awscli/alias.py
kyleknap: That's fine with me if we expose an alias loader attribute on the clidriver. I just needed a way to hook into replace this default value.

awscli/alias.py
kyleknap: I would need to make sure there is no edge cases with multiline files with that, but not opposed if there are no issues.

awscli/alias.py
kyleknap: I'm not comparing differing default parameters. I am just taking the parameters that I parsed out of the alias (pre-modifications) and see if the value in the resulting namespace from the parsing of the alias value to see if that value was a default or not (i.e. I am unable to check for ``None`` or ``not`` to determine these). And if it was not default the user must have passed the parameter in the alias value.

awscli/alias.py
kyleknap: So I actually tried using a list but it causes my functional tests to fail if I do that. I looked online and the python [docs](https://docs.python.org/2/library/subprocess.html#subprocess.Popen) on shell for POpen say:
```
On Unix with shell=True, the shell defaults to /bin/sh. If args is a string, the string specifies
 the command to execute through the shell. This means that the string must be formatted
 exactly as it would be when typed at the shell prompt. This includes, for example, quoting or
 backslash escaping filenames with spaces in them. If args is a sequence, the first item
 specifies the command string, and any additional items will be treated as additional
 arguments to the shell itself. That is to say, Popen does the equivalent of:


Popen(['/bin/sh', '-c', args[0], args[1], ...])

```
So if I use a list, the extra arguments at the end need to be included in the first element which would essentially require the same join I am doing now.

PR #2293, opened 2016-11-21
Add --generate-cli-skeleton output

kyleknap:
Allows users to run a CLI command without sending a request over the wire. Inputs will be validated and a mocked response will be printed back to the user.

Implementation-wise, I re-used ``CLIOperationCaller`` but stubbed the client with ``ArgumentGenerator`` to get the closest validation and client response back (essentially we are removing the HTTP Request/Response part of the stack).

As to testing, I primarily focused on functional testing. I found that a lot more useful, because the setup required to get unit test working properly (i.e. mocked client and no disk writing for models) it would require a ton of mocking to the point that would be difficult to read/manage the session/client mocks and understand what is being tested. But if you would like to see more unit tests, I can look into adding them.

### DIFF
--- a/awscli/customizations/generatecliskeleton.py
+++ b/awscli/customizations/generatecliskeleton.py
@@ -13,8 +13,11 @@
 import json
 import sys
 
+from botocore import xform_name
+from botocore.stub import Stubber
 from botocore.utils import ArgumentGenerator
 
+from awscli.clidriver import CLIOperationCaller
 from awscli.customizations.arguments import OverrideRequiredArgsArgument
 
 
@@ -36,18 +39,20 @@ class GenerateCliSkeletonArgument(OverrideRequiredArgsArgument):
 
     The argument, if present in the command line, will prevent the intended
     command from taking place. Instead, it will generate a JSON skeleton and
-    print it to standard output. This JSON skeleton then can be filled out
-    and can be used as input to ``--input-cli-json`` in order to run the
-    command with the filled out JSON skeleton.
+    print it to standard output.
     """
     ARG_DATA = {
         'name': 'generate-cli-skeleton',
-        'help_text': 'Prints a sample input JSON to standard output. Note the '
-                     'specified operation is not run if this argument is '
-                     'specified. The sample input can be used as an argument '
-                     'for ``--cli-input-json``.',
-        'action': 'store_true',
-        'group_name': 'generate_cli_skeleton'
+        'help_text': (
+            'Prints a JSON skeleton to standard output without sending '
+            'an API request. If provided with no value or the value '
+            '``input``, prints a sample input JSON that can be used as an '
+            'argument for ``--cli-input-json``. If provided with the value '
+            '``output``, it validates the command inputs and returns a '
+            'sample output JSON for that command.'
+        ),
+        'nargs': '?', 'const': 'input',
+        'choices': ['input', 'output'],
     }
 
     def __init__(self, session, operation_model):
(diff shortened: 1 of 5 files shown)


kyleknap:
Right. I forgot that the ``ArgumentGenerator`` is incomplete as it provides ``None`` for timestamps. I have a branch for this where I generate timestamps for ``ArgumentGenerator``. I will get a PR sent up for that.

awscli/customizations/generatecliskeleton.py
```diff
@@ -59,29 +65,66 @@ def _register_argument_action(self):
...
 
+    def override_required_args(self, argument_table, args, **kwargs):
+        arg_name = '--' + self.name
+        if arg_name in args:
+            arg_location = args.index(arg_name)
+            try:
+                # If the argument immediately preceeding
+                # --generate-cli-skeleton is output, then do not mark
+                # all of the other arguments as not required because
+                # output allows does parameter validation.
```
jamesls: wording is confusing here.

awscli/customizations/generatecliskeleton.py
```diff
@@ -59,29 +65,66 @@ def _register_argument_action(self):
...
+    to make a stubbed client call for an operation command.
+    """
+    def _make_client_call(self, client, operation_name, parameters,
+                          parsed_globals):
+        method_name = xform_name(operation_name)
+        operation_model = client.meta.service_model.operation_model(
+            operation_name)
+        fake_response = {}
+        if operation_model.output_shape:
+            argument_generator = ArgumentGenerator(use_member_names=True)
```
jamesls: I think we may need to update the logic in botocore here, especially if we're going through the param validation.  For example:

```
$ aws dynamodb list-tables --generate-cli-skeleton output

Parameter validation failed:
Invalid length for parameter TableNames[0], value: 0, valid range: 3-inf
```

awscli/customizations/generatecliskeleton.py
```diff
@@ -36,18 +40,20 @@ class GenerateCliSkeletonArgument(OverrideRequiredArgsArgument):
...
-        'group_name': 'generate_cli_skeleton'
+        'help_text': (
+            'Prints a JSON skeleton to standard output without sending '
+            'an API request. If provided with no value or the value '
+            '``input``, prints a sample input JSON that can be used as an '
+            'argument for ``--cli-input-json``. If provided with the value '
+            '``output``, it validates the command inputs and returns a '
+            'sample output JSON for that command.'
+        ),
+        'nargs': '?', 'const': 'input',
```
jamesls: I missed the `const: input` part initially, can you move this to a separate line?

tests/functional/test_generatecliskeleton.py
```diff
@@ -0,0 +1,96 @@
...
+        #
+        # We assert only components of the response in case members
+        # are added in the future that would break an exactly equals
+        # assertion
+        skeleton_output = json.loads(stdout)
+        self.assertIn('Regions', skeleton_output)
+        self.assertEqual(
+            skeleton_output['Regions'][0]['RegionName'], 'RegionName')
+        self.assertEqual(
+            skeleton_output['Regions'][0]['Endpoint'], 'Endpoint')
```
jamesls: Looks like you're comparing the whole structure anyways.  Any reason not to just use `assertEqual(whole_thing, ...)`?

kyleknap: I left a comment above, but I am trying to avoid this test breaking if ec2 ever adds another member to its DescribeRegions output. Does that make sense? I can also just use the entire response if we feel comfortable that the operation will not change, but there is no guarantees.

awscli/customizations/generatecliskeleton.py
kyleknap: That's unfortunate. I will see if I can make the strings in list match the member names.

awscli/customizations/generatecliskeleton.py
kyleknap: I will update it.

awscli/customizations/generatecliskeleton.py
kyleknap: Yep I can do that.

tests/functional/test_generatecliskeleton.py
JordonPhillips: I think we have an assert subset hanging around somewhere

kyleknap: Do you know where? I would like to use that if I can.

JordonPhillips: Here it is: https://github.com/boto/botocore/blob/develop/tests/unit/__init__.py#L36

kyleknap: Hmm that looks like it does not handle lists...

kyleknap: @jamesls and I talked offline about this and agreed its fine not to assert the entire dictionary

kyleknap:
@jamesls @JordonPhillips Alright test now pass, incorporated feedback, and made sure the integration test for ``--generate-cli-skeleton`` now pass as well. So should be good to look at.

awscli/customizations/generatecliskeleton.py
```diff
@@ -59,29 +66,66 @@ def _register_argument_action(self):
...
             else:
-                skeleton = argument_generator.generate_skeleton(
-                    operation_input_shape)
-
-            # Write the generated skeleton to standard output.
-            sys.stdout.write(json.dumps(skeleton, indent=4))
-            sys.stdout.write('\n')
-            # This is the return code
-            return 0
+                argument_generator = ArgumentGenerator()
```
jamesls: Looks like my original comment was lost somewhere, but just to capture:

I think it would be nice if both cases just delegated out to a class to handle the main logic.  In the `for_output` case, the actual behavior is delegated to another class, but the else clause has the logic all inline.


```
if for_output:
    set_stuff_up()
    RealObject.handle_it()
else:
    set_stuff_up()
    RealObject2.handle_it()

kyleknap: Makes sense. I will add a todo before I merge.

kyleknap:
Added the todo comment on the refactor. Merging.

PR #2296, opened 2016-11-23
Add transfer speed reporting to s3 commands

kyleknap:
This adds transfer speed reporting by default to whenever there is bytes progress to the ``s3`` transfer commands. Here is how it looks:
```
$ aws s3 cp 5GB s3://mybucketfoo
Completed 9.0 MiB/5.0 GiB (6.1 MiB/s) with 1 file(s) remaining
```

Let me know what you think. I am up for suggestions on the output and whether to make it opt in or not. So go ahead and try out the branch.

I did performance tests on the implementation as well and there were no noticeable performance differences.

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -13,6 +13,7 @@
 import logging
 import sys
 import threading
+import time
 from collections import namedtuple
 from collections import defaultdict
 
@@ -50,7 +51,8 @@ def _create_new_result_cls(name, extra_fields=None, base_cls=BaseResult):
 QueuedResult = _create_new_result_cls('QueuedResult', ['total_transfer_size'])
 
 ProgressResult = _create_new_result_cls(
-    'ProgressResult', ['bytes_transferred', 'total_transfer_size'])
+    'ProgressResult', ['bytes_transferred', 'total_transfer_size',
+                       'timestamp'])
 
 SuccessResult = _create_new_result_cls('SuccessResult')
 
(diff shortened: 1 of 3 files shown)


awscli/customizations/s3/results.py
```diff
@@ -268,6 +276,9 @@ def _record_progress_result(self, result, **kwargs):
         self._ongoing_progress[
             self._get_ongoing_dict_key(result)] += bytes_transferred
         self.bytes_transferred += bytes_transferred
+        if result.timestamp > self.start_time:
+            self.bytes_transfer_speed = 1.0 * self.bytes_transferred / (
```
kyleknap: Sounds good.

awscli/customizations/s3/results.py
```diff
@@ -268,6 +277,9 @@ def _record_progress_result(self, result, **kwargs):
         self._ongoing_progress[
             self._get_ongoing_dict_key(result)] += bytes_transferred
         self.bytes_transferred += bytes_transferred
+        if result.timestamp > self.start_time:
```
jamesls: In what scenario would you have `result.timestamp` be less than `self.start_time`?

kyleknap: @jamesls 

Super unlikely but because we capture the start time in the result recorder and we capture timestamps in the subscriber, there is a chance that if a progress result gets created right after the queued result gets created that the timestamp on the progress result is less than the timestamp of when the result processor actually processes that initial queued result. I can put a comment in the code about this if that makes sense?

jamesls: Ok that makes sense.  A comment about that would be helpful.

kyleknap:
Comment and changelog entry added. Merging.

Issue #2314, opened 2016-12-09
`aws cloudformation package` generates invalid `TemplateURL`


jkit:
# Description

The `aws cloudformation package` replaces local paths with an `s3://` URL, however, nested stacks in a CloudFormation template require that the `TemplateURL` is specified as a `https://s3.amazonaws.com/...` URL.

# Environment

```sh
aws --version
aws-cli/1.11.27 Python/2.7.10 Darwin/16.1.0 botocore/1.4.84
```

# Repro Steps

## 1. Create A Simple Stack
```yaml
# simple-bucket.yml
Resources:
  SimpleBucket:
    Properties:
      AccessControl: Private
      BucketName: some-bucket
    Type: AWS::S3::Bucket
```

## 2. Create A Stack With a Nested Stack
```yaml
# nested-bucket.yml
Resources:
  NestedBucket:
    Type: "AWS::CloudFormation::Stack"
    Properties:
      TemplateURL: simple-bucket.yml
```

## 3. Generate New Template with `aws cloudformation package` 
```sh
aws cloudformation package \
  --template-file nested-bucket.yml \
  --s3-bucket templates \
  --output-template-file package-template.yml
```

This generates `package-template.yml`:

```yml
Resources:
  NestedBucket:
    Properties:
      TemplateURL: s3://templates/2605eb3b081bda5a3e38952c8e4b70a4.template
    Type: AWS::CloudFormation::Stack
```

## 4. Deploy Generated Template

```sh
aws cloudformation deploy \
  --template-file package-template.yml \
  --stack-name TestStack
```

This fails with:
```
The specified url must be an Amazon S3 URL.
```

# Workaround

If I change `TemplateURL` from `s3://templates/2605eb3b081bda5a3e38952c8e4b70a4.template` to `https://s3.amazonaws.com/templates/f8b1a0b3844e6eea8ff045565f5b0b7e.template` and rerun the `deploy` command. It works fine.

djamison:
@jkit's workaround worked for me.

kyleknap:
I can confirm this issue. We are working on a fix.

jamesls:
Should be fixed via https://github.com/aws/aws-cli/pull/2360

PR #2331, opened 2016-12-17
Fixed url bug for AWS CloudFormation stacks when packaged

dtimberlake:
This is my purposed fix for #2314.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -19,7 +19,7 @@ import contextlib
 import uuid
 from awscli.compat import six
 
-from six.moves.urllib.parse import urlparse, parse_qs
+from six.moves.urllib.parse import urlparse, parse_qs, urljoin
 from contextlib import contextmanager
 from awscli.customizations.cloudformation import exceptions
 from awscli.customizations.cloudformation.yamlhelper import yaml_dump, \
@@ -45,6 +45,13 @@ def is_s3_url(url):
         return True
     except ValueError:
         return False
+
+
+def is_https_s3_url(url):
+    if isinstance(url, six.string_types) \
+        and url.startswith("https://s3.amazonaws.com/"):
+        return True
+    return False
 
 
 def is_local_folder(path):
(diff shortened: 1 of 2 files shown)


awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -47,6 +47,13 @@ def is_s3_url(url):
         return False
 
 
+def is_https_s3_url(url):
+    if isinstance(url, six.string_types) \
+        and url.startswith("https://s3.amazonaws.com/"):
```
jamesls: Do we need to account for virtual host style addressing or does cfn not handle this either?  e.g `https://mybucket.s3.amazonaws.com/`?

jamesls:
Thanks again for all the cloudformation package/deploy fixes/enhancements!

awscli/customizations/cloudformation/artifact_exporter.py
dtimberlake: I didn't think so because their documentation states:

> The template must be stored on an Amazon S3 bucket, so the URL must have the form: https://s3.amazonaws.com/.../TemplateName.template

After testing it looks like CloudFormation does allow the virtual host style urls.

The two approaches I see here would be to simply test for `url.startswith('https')` or `re.match('https://[\w-]+.s3.amazonaws.com/', url)`, Thoughts?

dtimberlake: After testing, it looks like CloudFormation also supports urls with the `http` protocol as well. I think we need to support everything that's listed on [this page](http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingExamples).

Using `re.match("https?://[\w.\d\-]*s3[\d\w-]*\.amazonaws\.com/", url)` would theoretically match all on that page (unless I missed some) except CNAME urls. To support those we should just check for the `http` protocol and keep moving.

jamesls:
@sanathkr provided a fix via https://github.com/aws/aws-cli/pull/2360 that should address these concerns.  Let us know if doesn't work for you.

PR #2332, opened 2016-12-19
Fixed yaml parsing error for !GetAtt

dtimberlake:
When parsing `!GetAtt`, attributes should not be split on periods.

### Original:
`!GetAtt MyStack.Outputs.Arn`

### Parses to:
`{ "Fn::GetAtt": [ "MyStack", "Outputs", "Arn" ] }`

### Should be:
`{ "Fn::GetAtt": [ "MyStack", "Outputs.Arn" ] }`


jamesls:
Good catch, thanks for the pull request.

PR #2333, opened 2016-12-19
Handle race conditions in s3 sync a little better

### LINKED ISSUE #1841: [Errno 2] No such file or directory
I know this has been brought up before but I think I have found another scenario where this error can occur.

So I'm running the following command to sync files to S3.

`aws s3 sync ./source/ s3://bucket/destination/ --exclude '*' --include '*.mp3' --storage-class STANDARD_IA --no-follow-symlink`

Great, works fine. But every now and again the sync fails with the error:

`[Errno 2] No such file or directory`

I had a look around and saw people were talking about broken symlinks. I though it was quite strange seeing as my source is just a folder with MP3's - but I tried the `--no-follow-symlink` flag anyway. This still didn't work.

So I have a theory which I hope will be correct and result in a fix. When I run the sync command, it gathers a list of all the files in the directory then syncs them. This list doesn't change from the initial gathering. So I'm trying to sync a folder which regularly changes. So my theory is that while I'm trying to sync files, at the same time some of them are being deleted by their respective users, but because the initial file list is only retrieved at the beginning of the sync - it fails because the file doesn't exist anymore.

It would be nice just to ignore these failures. Let me know if I'm correct in thinking this.

2rs2ts:
The LBYL approach with `should_ignore_file` checking if `triggers_warning`
is `True` is susceptible to rare race conditions in which a file is
deleted after `should_ignore_file` calls `triggers_warning` but before
`get_file_stat` calls `os.stat`. When that happens, an `OSError` is raised.
This doesn't affect most people, but when you are using `s3 sync` on a
large directory that is constantly shifting (for example, a logging
directory being managed by logrotate, or a directory with a ton of tmp
files that come and go) this has a much higher chance of happening.

While my solution here does not address every possible error case, it
does address a few. I'm not sure how I can create a robust blanket
solution, but a little improvement is better than none at all, right?

Fixes #1841

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -207,18 +207,17 @@ class FileGenerator(object):
                         for x in self.list_files(file_path, dir_op):
                             yield x
                     else:
-                        stats = self._safely_get_file_stats(path, file_name=name)
+                        stats = self._safely_get_file_stats(file_path)
                         if stats:
                             yield stats
 
-    def _safely_get_file_stats(self, path, file_name=None):
-        file_path = path if file_name is None else os.path.join(path, file_name)
+    def _safely_get_file_stats(self, file_path):
         try:
             size, last_update = get_file_stat(file_path)
         except OSError:
             self.triggers_warning(file_path)
         else:
-            last_update = self._validate_update_time(last_update, path)
+            last_update = self._validate_update_time(last_update, file_path)
             return file_path, {'Size': size, 'LastModified': last_update}
 
     def _validate_update_time(self, update_time, path):
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -172,9 +172,13 @@ class FileGenerator(object):
         error, listdir = os.error, os.listdir
         if not self.should_ignore_file(path):
             if not dir_op:
-                size, last_update = get_file_stat(path)
-                last_update = self._validate_update_time(last_update, path)
-                yield path, {'Size': size, 'LastModified': last_update}
+                try:
+                    size, last_update = get_file_stat(path)
+                except OSError:
+                    self.triggers_warning(file_path)
+                else:
+                    last_update = self._validate_update_time(last_update, path)
+                    yield path, {'Size': size, 'LastModified': last_update}
 
             else:
                 # We need to list files in byte order based on the full
@@ -208,13 +212,17 @@ class FileGenerator(object):
                         for x in self.list_files(file_path, dir_op):
                             yield x
                     else:
-                        size, last_update = get_file_stat(file_path)
-                        last_update = self._validate_update_time(
-                            last_update, path)
-                        yield (
-                            file_path,
-                            {'Size': size, 'LastModified': last_update}
-                        )
+                        try:
+                            size, last_update = get_file_stat(file_path)
+                        except OSError:
+                            self.triggers_warning(file_path)
+                        else:
+                            last_update = self._validate_update_time(
+                                last_update, path)
+                            yield (
+                                file_path,
+                                {'Size': size, 'LastModified': last_update}
+                            )
 
     def _validate_update_time(self, update_time, path):
         # If the update time is None we know we ran into an invalid tiemstamp.
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -172,14 +172,9 @@ class FileGenerator(object):
         error, listdir = os.error, os.listdir
         if not self.should_ignore_file(path):
             if not dir_op:
-                try:
-                    size, last_update = get_file_stat(path)
-                except OSError:
-                    self.triggers_warning(file_path)
-                else:
-                    last_update = self._validate_update_time(last_update, path)
-                    yield path, {'Size': size, 'LastModified': last_update}
-
+                stats = self._safely_get_file_stats(path)
+                if stats:
+                    yield stats
             else:
                 # We need to list files in byte order based on the full
                 # expanded path of the key: 'test/1/2/3.txt'  However,
@@ -212,17 +207,19 @@ class FileGenerator(object):
                         for x in self.list_files(file_path, dir_op):
                             yield x
                     else:
-                        try:
-                            size, last_update = get_file_stat(file_path)
-                        except OSError:
-                            self.triggers_warning(file_path)
-                        else:
-                            last_update = self._validate_update_time(
-                                last_update, path)
-                            yield (
-                                file_path,
-                                {'Size': size, 'LastModified': last_update}
-                            )
+                        stats = self._safely_get_file_stats(path, file_name=name)
+                        if stats:
+                            yield stats
+
+    def _safely_get_file_stats(self, path, file_name=None):
+        file_path = path if file_name is None else os.path.join(path, file_name)
+        try:
+            size, last_update = get_file_stat(file_path)
+        except OSError:
+            self.triggers_warning(file_path)
+        else:
+            last_update = self._validate_update_time(last_update, path)
+            return file_path, {'Size': size, 'LastModified': last_update}
 
     def _validate_update_time(self, update_time, path):
         # If the update time is None we know we ran into an invalid tiemstamp.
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks for the pull request.  The idea seems reasonable to me.  What I would suggest though:

The code about catching `OSError` is duplicated a few times.  I think we should refactor the code to move all the file lookup stuff to a single method.  That way we can cut down on the duplicate code.

2rs2ts:
I can put this particular snippet in a closure, since the `else` case is the same in both cases, if that's what you mean, @jamesls. Otherwise I'm not sure what you mean; I'm new to this codebase, after all.

2rs2ts:
I'd like to follow up on this; can I get some clarification for my last question?

2rs2ts:
Bump.

2rs2ts:
@jamesls sorry to pester, but can you clarify your earlier comment so I can make my change? I don't want to take a stab in the dark.

2rs2ts:
Well, I ended up taking a stab in the dark :)

awscli/customizations/s3/filegenerator.py
```diff
@@ -208,13 +207,19 @@ def list_files(self, path, dir_op):
...
                             yield x
                     else:
-                        size, last_update = get_file_stat(file_path)
-                        last_update = self._validate_update_time(
-                            last_update, path)
-                        yield (
-                            file_path,
-                            {'Size': size, 'LastModified': last_update}
-                        )
+                        stats = self._safely_get_file_stats(path, file_name=name)
```
kyleknap: So for the ``_safely_get_file_stats`` can we just simplify the signature to just have ``path`` and just use ``file_path`` variable when we make this call instead of ``path`` and ``name``?

2rs2ts: No, you'll note that both the original implementation and my new one use the original `path`.

2rs2ts:
@kyleknap I'm down to give a test the ol' college try, but can I get some guidance on where it would belong? I'm guessing `tests/functional/s3/test_sync_command.py`, but if that's not right, let me know :)

awscli/customizations/s3/filegenerator.py
kyleknap: Sorry. I am still a little confused on how it is different.

So on line 199: https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3R199 there is:
``file_path = join(path, name)``

Then on line 215: https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3R215. The logic is:
``file_path = path if file_name is None else os.path.join(path, file_name)``

So then if you make line 210: https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3R210:
`` stats = self._safely_get_file_stats(file_path)``

It would produce the same result as the original line you are adding because the variable ``file_path`` on line 199 would be the same result as on line 215 with ``os.path.join(path, file_name)``.

tests/functional/s3/test_sync_command.py
```diff
@@ -164,3 +165,17 @@ def test_sync_with_delete_on_downloads(self):
...
+
+    def test_sync_skips_over_files_deleted_between_listing_and_transfer_attempt(self):
+        full_path = self.files.create_file('foo.txt', 'mycontent')
+        cmdline = '%s %s s3://bucket/' % (
+            self.prefix, self.files.rootdir)
+
+        # FileGenerator.list_files should skip over files that cause an OSError to be raised
+        # because they are missing when we try to get their stats.
+        with patch('awscli.customizations.s3.filegenerator.get_file_stat', side_effect=OSError()):
+            self.run_cmd(cmdline, expected_rc=0)
```
kyleknap: Interesting. I think the ``expect_rc`` for this would be ``2`` as a warning gets thrown and the file gets skipped. And that is happening because I think you need to mock out the response for ``ListObjects`` by setting ``self.parsed_responses``. Take a look at the example here:
https://github.com/aws/aws-cli/pull/2333/files?diff=unified#diff-b8e88c17b1981210afdf4b277e015fb2R85

awscli/customizations/s3/filegenerator.py
2rs2ts: I'm referring to these lines:

* [original](https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3L213)
* [new](https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3R221)

kyleknap: Oh... I see why you are doing that now! So the line you pointed to in the original is actually a bug...

In that line snippet we are passing ``path`` which is directory to the warning message, when we should be passing the ``file_path`` which is the actual filename that produced the ``last_update`` value. I guess it never went noticed because all it affects is the warning message in that it will print out the directory instead of filename in the warning.

I am all for fixing this to use ``file_path``. It would also simplify your code.

tests/functional/s3/test_sync_command.py
2rs2ts: Hm, should it cause the command to fail? I think if that is the case then my code does not resolve the issue I'm trying to resolve: https://github.com/aws/aws-cli/issues/1841

What we want is for these types of errors to be ignored. At least in my case, I have a script that runs after logrotate which syncs rotated logs to s3. Sometimes the logs get rotated away before they finish syncing – e.g. my pattern matches log1, log2, and log3, and while I am syncing log2, logrotate runs again and rotates log3 away, so by the time I get to it the file is gone and `s3 sync` fails. My script records when the `s3 sync` fails so that I have a record of possibly missing logs, so getting a return code of 2 in this case would not resolve my issue.

kyleknap: So unfortunately even if we gets this patch merged in, there is no guarantee that your sync script will always return 0. If you are rotating logs around while syncing them, there is a chance it gets listed out and queued but gets deleted before the upload begins causing that individual file to change.

However it does resolve the reference issue where removing an individual file mid-sync cause the entire sync command to fail. It is much better improvement to have it fail for that individual file instead of the entire sync.

Does that makes sense? Have you tried your patch on your current setup and do you get zero every time now?

kyleknap:
Recent commit to use ``file_path`` looks good. Let's now resolve the part about the tests.

tests/functional/s3/test_sync_command.py
2rs2ts: I agree, it's an improvement to sync as much as possible. The problem in https://github.com/aws/aws-cli/issues/1841 is that we want to ignore these errors - that is, we need at least need a way to reliably detect this particular error scenario so we can let our scripts consider the sync a success, if not have the `s3 sync` command itself exit with 0 – and I am not sure if I will be able to say I resolved that issue with this PR if the command exits with 2, because I don't know if there is a reliable way to detect this particular error scenario.

I have not tried this patch on my current setup. I'm not able to reliably reproduce the issue, so unless it's necessary to test it, I'd like to avoid the pain of doing so.

kyleknap: I don't think you need to test this in your current setup. I was just saying that even with this fix I do not think it will prevent all nonzero error codes from files being deleted mid-sync and the test as it sits now does not accurately reflect what will actually happen given the current state of the code.

Looking at the code again the issue is not that ``self.parsed_responses`` is not set. The issue is that when [trigger_warnings](https://github.com/aws/aws-cli/pull/2333/files#diff-2d6153b2f8310981a59f57a017f7a2a3R279) is called because of the ``OSError``, the file still exists and therefore fails to meet the check in ``trigger_warning`` for if the file exists.

As to https://github.com/aws/aws-cli/issues/1841 if this merged, that particular issue should be closed as it is only related to the OSError that crashes the ``sync`` command during the stating of the files. The general issue of suppressing errors and warnings for files that are getting modified/deleted mid-sync, that is a much larger problem, a lot harder to solve, and deserves its own conversation/issue.

2rs2ts: All right, I agree with your reasoning. I've updated the test.

2rs2ts:
👏 thank you

PR #2334, opened 2016-12-19
Include cross links to service API documentation

jamesls:
For reference, see ruby: http://docs.aws.amazon.com/sdkforruby/api/Aws/ACM/Client.html#add_tags_to_certificate-instance_method

This injects the link at the end of the description:

image

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -300,7 +300,7 @@ class ServiceDocumentEventHandler(CLIDocumentEventHandler):
 
 class OperationDocumentEventHandler(CLIDocumentEventHandler):
 
-    AWS_DOC_BASE = 'http://docs.aws.amazon.com/goto/WebAPI'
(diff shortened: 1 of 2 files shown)


awscli/clidocs.py
```diff
@@ -300,6 +300,8 @@ def doc_subitem(self, command_name, help_command, **kwargs):
 
 class OperationDocumentEventHandler(CLIDocumentEventHandler):
 
+    AWS_DOC_BASE = 'http://docs.aws.amazon.com/goto/WebAPI'
```
kyleknap: Should it be https instead?

jamesls: Yeah, I copied from ruby.  I'll get that to https.

jamesls:
Updated to use https.

@kyleknap, no I was not planning on linking to the service docs at this time.

PR #2347, opened 2016-12-28
Support Python 3.6

### LINKED ISSUE #2342: Support Python 3.6
Python 3.6 got released, and some distro (like Fedora) are swithcing to it.

JordonPhillips:
This adds python 3.6 to the tox configuration, the setup.py
classifiers, the travis configuration, and the readme.

Closes #2342

cc @kyleknap @jamesls @stealthycoin @dstufft


JordonPhillips:
I found the issue. Python's `fromtimestamp` can raise any of `ValueError`, `OSError`, or `OverflowError` depending on the platform and python version. We were only handling `OSError` and `ValueError`. This hadn't been an apparent issue, but it seems that python 3.6 on Travis' environment will throw `OverflowError`.

jamesls:
Interesting.  I wonder if that was intentional, seems like a breaking change from 3.5->3.6

JordonPhillips:
Yeah, I've been investigating that. The traceback wasn't very helpful since it doesn't appear to be a change with `dateutil`, but rather the standard library. I'm going to continue looking into it.

JordonPhillips:
Here's a minimal script to reproduce the error, it requires `python-dateutil`:

```python
from datetime import datetime

from dateutil.tz import tzlocal


bad_st_mtime = -100000000000.0
local_time_zone = tzlocal()

# Try converting a bad timestamp into a datetime
try:
    datetime.fromtimestamp(bad_st_mtime, tzlocal())
except ValueError:
    # This is the error that should be raised, so we ignore it
    pass
```

tests/unit/customizations/s3/test_utils.py
```diff
@@ -265,30 +265,29 @@ def test_get_file_stat(self):
...
-        patch_attribute = 'awscli.customizations.s3.utils.datetime'
-        with mock.patch(patch_attribute) as datetime_mock:
-            with temporary_file('w') as temp_file:
-                temp_file.write('foo')
-                temp_file.flush()
-                datetime_mock.fromtimestamp.side_effect = OSError()
-                size, update_time = get_file_stat(temp_file.name)
-                self.assertIsNone(update_time)
+    def test_returns_epoch_on_invalid_timestamp(self):
+        self.assert_handles_error(ValueError())
```
jamesls: I would prefer we use something more descriptive, `assert_handles_error` is pretty vague.

JordonPhillips:
* Note that the sample code only fails on some systems, such as Ubuntu 12.04

JordonPhillips:
I submitted a bug report to python: http://bugs.python.org/issue29100

JordonPhillips:
Renamed the assertion function

PR #2349, opened 2016-12-29
Use waiter to ensure S3 key exists before proceeding

jamesls:
This helps make the integration tests more robust.  It adds
a helper method that will use a waiter to ensure an object exists
at least min_successes times.

The two test classes I modified had test code that tried to download
an object immediately after uploading it.  This would cause
sporadic test failures.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -775,6 +775,16 @@ class BaseS3CLICommand(unittest.TestCase):
         response = client.head_object(Bucket=bucket_name, Key=key_name)
         return response
 
+    def wait_until_key_exists(self, bucket_name, key_name, extra_params=None,
+                              min_successes=3):
+        client = self.create_client_for_bucket(bucket_name)
+        waiter = client.get_waiter('object_exists')
+        params = {'Bucket': bucket_name, 'Key': key_name}
+        if extra_params is not None:
+            params.update(extra_params)
+        for _ in range(min_successes):
+            waiter.wait(**params)
+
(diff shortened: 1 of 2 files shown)


awscli/testutils.py
```diff
@@ -775,6 +775,16 @@ def head_object(self, bucket_name, key_name):
...
         return response
 
+    def wait_until_key_exists(self, bucket_name, key_name, extra_params=None,
+                              min_successes=3):
+        client = self.create_client_for_bucket(bucket_name)
+        waiter = client.get_waiter('object_exists')
+        params = {'Bucket': bucket_name, 'Key': key_name}
+        if extra_params is not None:
+            params.update(extra_params)
+        for _ in range(min_successes):
```
JordonPhillips: Why wait multiple times?

jamesls: Minimize failures.  We're making a best effort attempt to deal with eventual consistency here.

JordonPhillips: So it's possible for a waiter to exit successfully here, but then `get_object` still fails?

jamesls: Possible? yes.  Likely? no.

I'm ok with trading off a second or two in runtime to ensure the rare cases are even less likely to happen.

kyleknap:
Yeah I was thinking something along those lines.

jamesls:
Another way of looking at this problem is to add something to the operation you actually want to succeed: https://github.com/boto/botocore/issues/1114

PR #2356, opened 2017-01-03
Fixed a false statement about 0 value for the retention period of SWF execution

maksymkazakov:


stealthycoin:
The most up to date version of the cli (which at the time of writing is awscli-1.11.36 botocore-1.4.93) reflects the current docs [here](http://docs.aws.amazon.com/amazonswf/latest/apireference/API_RegisterDomain.html)

maksymkazakov:
@stealthycoin Wait, but that's a different page that you gave link to - http://docs.aws.amazon.com/amazonswf/latest/apireference/API_RegisterDomain.html
And I've pushed an update for CLI command, not for API action - http://docs.aws.amazon.com/cli/latest/reference/swf/register-domain.html

That page is lying at the moment, telling that 0 retention days = maximum retention period. Because of that I have created a domain that doesn't retain the history and due to existing constraints I can edit that setting anymore!

sjevs:
I agree with @maksymkazakov its confusing. @stealthycoin can you reopen or explain further?

PR #2360, opened 2017-01-04
Fix two bugs with CloudFormation package & deploy commands

jamesls:
1. `deploy` command must not override parameters with default values
2. `package` command must use Path-style S3 URL when packaging AWS::CloudFormation::Stack resource

Pulls in fixes from @sanathkr

### DIFF
--- a/tests/unit/customizations/cloudformation/test_s3uploader.py
+++ b/tests/unit/customizations/cloudformation/test_s3uploader.py
@@ -40,9 +40,10 @@ class TestS3Uploader(unittest.TestCase):
         self.transfer_manager_mock.upload = Mock()
         self.bucket_name = "bucketname"
         self.prefix = None
+        self.region = "us-east-1"
 
         self.s3uploader = S3Uploader(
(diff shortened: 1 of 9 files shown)


tests/unit/customizations/cloudformation/test_s3uploader.py
```diff
@@ -261,3 +262,46 @@ def test_make_url(self):
...
+                        region, self.bucket_name, key, version))
+
+        # Without versionId, that query parameter should be omitted
+        s3uploader = S3Uploader(self.s3client, self.bucket_name, region)
+        result = s3uploader.to_path_style_s3_url(key)
+        self.assertEqual(
+                result,
+                "https://s3-{0}.amazonaws.com/{1}/{2}".format(
+                        region, self.bucket_name, key))
+
```
kyleknap: Blank lines at the end of the file

jamesls: Ok I'll update that before merging.

PR #2368, opened 2017-01-06
Timestamp format

stealthycoin:
First draft of the cli_timestamp_format config option.

cc @jamesls @JordonPhillips @kyleknap @dstufft


awscli/customizations/scalarparse.py
```diff
@@ -36,13 +38,27 @@ def identity(x):
     return x
 
 
+def choose_default_formatters(timestamp_format):
+    parsers = dict(blob_parser=identity)
+    if timestamp_format == 'none':
+        # For backwards compatibility reasons, we replace botocore's timestamp
+        # parser (which parsers to a datetime.datetime object) with the
```
dstufft: Small nit, ``parses`` not ``parsers`` in "which parsers to a datetime.datetime object".

awscli/customizations/scalarparse.py
```diff
@@ -36,13 +38,27 @@ def identity(x):
...
+        # identity function which prints the date exactly the same as it comes
+        # across the wire.
+        parsers['timestamp_parser'] = identity
+    elif timestamp_format == 'iso8601':
+        # This is a noop because the underlying botocore behavior already uses
+        # iso8601
```
dstufft: This seems to be contradictory to the above comment, the default botocore parser parses it into a ``datetime.datetime`` object does it not? Assuming this is getting converted using ``str()`` that will get you an ``iso8601`` string but that is only true if every place you pass this in uses ``str()`` to convert the datetime. It's also not exactly ``iso8601`` since they replace the ``T`` separator with a empty space. That is allowed "by mutual agreement" but I typically see iso8601 as rendered with the ``T``.

awscli/customizations/scalarparse.py
```diff
@@ -36,13 +38,27 @@ def identity(x):
     return x
 
 
+def choose_default_parsers(timestamp_format):
+    parsers = dict(blob_parser=identity)
+    if timestamp_format == 'none':
+        # For backwards compatibility reasons, we replace botocore's timestamp
+        # parser (which parsers to a datetime.datetime object) with the
```
dstufft: Small nit but ``which parses`` not ``which parsers``.

stealthycoin: fixed.

awscli/customizations/scalarparse.py
```diff
@@ -36,13 +38,27 @@ def identity(x):
...
+        # identity function which prints the date exactly the same as it comes
+        # across the wire.
+        parsers['timestamp_parser'] = identity
+    elif timestamp_format == 'iso8601':
+        # This is a noop because the underlying botocore behavior already uses
+        # iso8601
+        pass
+    else:
+        raise ValueError('Unknown cli_timestamp_format value: %s, valid values'
+                         ' are \'none\' or \'iso8601\'' % timestamp_format)
```
jamesls: You can just use double quotes so you don't have to escape the single quotes.

tests/unit/customizations/test_scalarparse.py
```diff
@@ -31,9 +34,18 @@ def test_identity(self):
...
         factory.set_parser_defaults.assert_called_with(
             blob_parser=scalarparse.identity,
             timestamp_parser=scalarparse.identity)
+
+    def test_choose_default_formatters(self):
+        self.assertEqual(self.formatter_dict_both,
+                         scalarparse.choose_default_parsers('none'))
+        self.assertEqual(self.formatter_dict_blob_only,
+                         scalarparse.choose_default_parsers('iso8601'))
+        with self.assertRaises(ValueError):
```
jamesls: Let's separate out the error tests from the success test.

awscli/customizations/scalarparse.py
stealthycoin: Good call. JSONFormatter explicitly uses `.isoformat` where as TableFormatter and TextFormatter use `str()/unicode()` so it loses the `T` in those two formats.

JordonPhillips:
I believe you may have messed up a bit when rebasing

Issue #2366, opened 2017-01-06
CLI doesn't work on windows with unicode profile names


soukicz:
Hello,
CLI is using credentials from home directory but it doesn't work if this directory has unicode characters:

```
C:\Users\René Winkler>aws --version
Traceback (most recent call last):
File "aws", line 27, in <module>
File "aws", line 23, in main
File "awscli\clidriver.pyc", line 49, in main
File "awscli\clidriver.pyc", line 57, in create_clidriver
File "botocore\session.pyc", line 368, in full_config
File "botocore\configloader.pyc", line 105, in load_config
File "botocore\configloader.pyc", line 126, in raw_config_parse
File "botocore\exceptions.pyc", line 27, in __init__
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 12:   ordinal not in range(128)
```
https://stackoverflow.com/questions/37904689/unicodedecodeerror-when-calling-aws-cli

dstufft:
Are you able to see what version of botocore you have installed? If you've installed using pip then you should be able to see that information using ``pip list``. I believe that this should have been fixed by boto/botocore#861 so if you have a version older than 1.4.9 then upgrading should solve it. If you have a version of botocore >= 1.4.9 then it must be another issue.

soukicz:
It was installed from MSI package (https://s3.amazonaws.com/aws-cli/AWSCLI64.msi). I will try to uninstall and install it from pip.

NK74:
Same issue on Windows Server 2012 R2 with AWSCLI x64 (https://s3.amazonaws.com/aws-cli/AWSCLI64.msi).

> aws-cli/1.11.35 Python/2.7.9 Windows/2012Server botocore/1.4.92

> Traceback (most recent call last):
>   File "aws", line 27, in <module>
>   File "aws", line 23, in main
>   File "awscli\clidriver.pyc", line 54, in main
>   File "awscli\clidriver.pyc", line 223, in main
> UnicodeEncodeError: 'ascii' codec can't encode character u'\xde' in position 58: ordinal not in range(128)

![image](https://cloud.githubusercontent.com/assets/17825769/21882780/54a70ea6-d8ac-11e6-9a05-5a4b47a91fd7.png)

But AWS CLI work perfectly with unicode characters

dstufft:
I believe this should be fixed now with https://github.com/aws/aws-cli/pull/2374.

Issue #2370, opened 2017-01-08
DynamoDB batch-write-item UnicodeEncodeError


floschliep:
I'm trying to import a JSON file in DynamoDB using the following command:
`aws dynamodb batch-write-item --request-items file:///path/to/json/file.json`

Here's the error:
```
ParamError: Error parsing parameter '--request-items': Invalid JSON:
...
2017-01-08 15:31:55,068 - MainThread - awscli.clidriver - DEBUG - Exiting with rc 255

Traceback (most recent call last):
  File "/usr/local/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/local/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/Library/Python/2.7/site-packages/awscli/clidriver.py", line 54, in main
    return driver.main()
  File "/Library/Python/2.7/site-packages/awscli/clidriver.py", line 223, in main
    sys.stderr.write("%s\n" % e)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 1751-1755: ordinal not in range(128)
```

The error seems to be caused by Chinese characters, e.g. `针织品批发`.

How would I go about importing such a file? Can DynamoDB store those characters or only as unicode? Will I need to decode them after reading them from the database (assuming I stored them encoded)?

dstufft:
There appear to be two errors here. The first error is that your JSON is invalid, which unfortunately the error message is being obscured because when it tries to write out the exception message it fails with an encoding error. I've solved the exception that is masking the real exception in #2374, but perhaps trying to load the json file using:

```python
import json

json.loads(open("FILENAME").read())
```

Might indicate what the underlying issue is.

floschliep:
I can successfully load the JSON file without an error.
Though I noticed every string is prefixed with a `u` (as the file contains unicode), might this be an issue?

JordonPhillips:
I was able to put in the provided characters, so I'm inclined to think your json didn't match the dynamodb schema. That said, the printing bug should be dealt with.

JordonPhillips:
To be clear, does your file look like this?

```json
{
    "testing-table": [{
        "PutRequest": {
            "Item": {
                "numbah": {"N": "34322321231354354325342532452345"},
                "strnum": {"S": "针织品批发"}
            }
        }
    }]
}
```

or like this:

```json
{
    u"testing-table": [{
        u"PutRequest": {
            u"Item": {
[...]

Issue #1898, opened 2016-04-08
Piping to text: TypeError: encode() argument 1 must be string, not None


rpbaptist:
I was piping the output of `aws s3 ls` to a file and received this message:

`TypeError: encode() argument 1 must be string, not None`

After a search I solved it by setting `export PYTHONIOENCODING='utf-8'`

jamesls:
The underlying issue has also been fixed via https://github.com/aws/aws-cli/pull/1844.  You'll no longer get the `TypeError` even when `PYTHONIOENCODING` is set.

PR #2374, opened 2017-01-10
Output unicode instead of a byte string

dstufft:
This should fix the encoding error that is masking the real error in #2370. Unfortunately when I went to run tests on this, I found out that the latest version of nose breaks if you try to write unicode to ``sys.stdout`` (see https://github.com/nose-devs/nose/issues/680).

The test that I wrote for this is:

```python
def test_error_unicode(self):
        driver = CLIDriver(session=self.session)
        fake_client = mock.Mock()
        fake_client.list_objects.side_effect = Exception(u"☃")
        fake_client.can_paginate.return_value = False
        driver.session.create_client = mock.Mock(return_value=fake_client)
        rc = driver.main('s3 list-objects --bucket foo'.split())
        self.assertEqual(rc, 255)
```

We can either just let this go without a test, or pin ``nose`` to ``<1.3.0``.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -220,7 +220,7 @@ class CLIDriver(object):
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
             sys.stderr.write("\n")
-            sys.stderr.write("%s\n" % e)
+            sys.stderr.write(u"%s\n" % e)
             return 255
 
     def _emit_session_event(self):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -23,7 +23,7 @@ from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
 
 from awscli import EnvironmentVariables, __version__
-from awscli.compat import get_text_writer
+from awscli.compat import get_stderr_text_writer
 from awscli.formatter import get_formatter
 from awscli.plugin import load_plugins
 from awscli.commands import CLICommand
@@ -221,7 +221,7 @@ class CLIDriver(object):
         except Exception as e:
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
-            err = get_text_writer(sys.stderr)
+            err = get_stderr_text_writer()
             err.write("\n")
             err.write(six.text_type(e))
             err.write("\n")
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -23,6 +23,7 @@ from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
 
 from awscli import EnvironmentVariables, __version__
+from awscli.compat import get_text_writer
 from awscli.formatter import get_formatter
 from awscli.plugin import load_plugins
 from awscli.commands import CLICommand
@@ -220,14 +221,10 @@ class CLIDriver(object):
         except Exception as e:
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
-            sys.stderr.write("\n")
-            if six.PY3:
-                sys.stderr.write(six.text_type(e))
-            else:
-                sys.stderr.write(six.text_type(e).encode(
-                    getattr(sys.stderr, "encoding", None) or "ascii",
-                    "replace"))
-            sys.stderr.write("\n")
+            err = get_text_writer(sys.stderr)
+            err.write("\n")
+            err.write(six.text_type(e))
+            err.write("\n")
             return 255
 
     def _emit_session_event(self):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -26,6 +26,7 @@ from awscli import EnvironmentVariables, __version__
 from awscli.formatter import get_formatter
 from awscli.plugin import load_plugins
 from awscli.commands import CLICommand
+from awscli.compat import six
 from awscli.argparser import MainArgParser
 from awscli.argparser import ServiceArgParser
 from awscli.argparser import ArgTableArgParser
@@ -220,7 +221,13 @@ class CLIDriver(object):
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
             sys.stderr.write("\n")
-            sys.stderr.write(u"%s\n" % e)
+            if six.PY3:
+                sys.stderr.write(six.text_type(e))
+            else:
+                sys.stderr.write(six.text_type(e).encode(
+                    getattr(sys.stderr, "encoding", None) or "ascii",
+                    "replace"))
+            sys.stderr.write("\n")
             return 255
 
     def _emit_session_event(self):
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ docutils>=0.10
 -e git://github.com/boto/botocore.git@develop#egg=botocore
 -e git://github.com/boto/s3transfer.git@develop#egg=s3transfer
 -e git://github.com/boto/jmespath.git@develop#egg=jmespath
-nose==1.3.0
+nose==1.3.7
 colorama>=0.2.5,<=0.3.7
 mock==1.3.0
 rsa>=3.1.2,<=3.5.0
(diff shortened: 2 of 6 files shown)


JordonPhillips:
I think it's probably safe to pin nose for now, since we're the only ones that use those particular dependencies.

stealthycoin:
I'd rather test and pin nose than not test.

JordonPhillips:
Looks like the test fails on python 2.6

JordonPhillips:
Odd, stdout is reporting an encoding of `UTF-8` but is actually ascii.

dstufft:
This should fix #2366.

awscli/clidriver.py
```diff
@@ -220,7 +221,13 @@ def main(self, args=None):
             LOG.debug("Exception caught in main()", exc_info=True)
             LOG.debug("Exiting with rc 255")
             sys.stderr.write("\n")
-            sys.stderr.write("%s\n" % e)
+            if six.PY3:
```
kyleknap: So usually if there is a python 2/3 compatibility logic, the typical pattern in the CLI codebase is to add it to the ``compat`` module. Digging through the ``compat`` module, I found this commit that James made: https://github.com/aws/aws-cli/commit/8bb3f1aa9ab52f5dcb57e59cb45ff9b83c543457. So it looks like we have ran into this issue before, but with standard out. So it may be worth just copying that pattern or expanding the compat method to a general ``get_text_wrapper()`` function that you can provide stdin or stdout to.

requirements.txt
```diff
@@ -6,7 +6,7 @@ docutils>=0.10
 -e git://github.com/boto/botocore.git@develop#egg=botocore
 -e git://github.com/boto/s3transfer.git@develop#egg=s3transfer
 -e git://github.com/boto/jmespath.git@develop#egg=jmespath
-nose==1.3.0
+nose==1.3.7
```
kyleknap: What was the particular reason for updating the version of ``nose``?

dstufft: Um, im pretty sure it was the bug where nose always failed trying to output Unicode that I mentioned in the PR body.

awscli/clidriver.py
dstufft: I can do that.

dstufft:
Ok, I've implemented the review feedback. I also made the original compat shim more correct and robust against degenerate cases that it would have previously raised an exception during. Note, I had to skip the test on Python 2.6 because nose on Python 2.6 basically can't have anything non ASCII written to ``sys.stdout`` or ``sys.stderr``.

tests/unit/test_clidriver.py
```diff
@@ -279,6 +282,20 @@ def test_ctrl_c_is_handled(self):
...
+        # written to stdout/stderr, which makes it impossible for this test
+        # to actually succeed on 2.6.
+        if sys.version_info < (2, 7):
+            raise nose.SkipTest
+        driver = CLIDriver(session=self.session)
+        fake_client = mock.Mock()
+        fake_client.list_objects.side_effect = Exception(u"☃")
+        fake_client.can_paginate.return_value = False
+        driver.session.create_client = mock.Mock(return_value=fake_client)
+        rc = driver.main('s3 list-objects --bucket foo'.split())
```
kyleknap: So one idea that I have that should solve both skipping the python2.6 test (which we really try to avoid) and upgrading the version of nose is to patch ``sys.stderr`` and then assert the correct exception message got raised to the user? The only reason I suggest this is:

* Even though in the test you are asserting the rc, the exact reason why the rc was set to 255 is unknown unless you take a look at the error printed to standard error.
* While its arguably not the best practice to be patching ``sys.stderr`` for unit tests, it is something it is done all over this test file. So there is a precedent set. For example: https://github.com/aws/aws-cli/blob/develop/tests/unit/test_clidriver.py#L283. What would be better is if the CLIDriver allowed you to inject a file-handle for stdout and stderr so you do not have to patch to assert content written out, but that is unfortunately not the case.

Let me know what you think?

dstufft:
@kyleknap Well, the previous behavior can trivially error when ``locale.getpreferredencoding()`` returns ``None``, you can see a similar issue in https://github.com/pypa/pip/issues/4184. I'm not sure under what conditions it does that though, but we're hitting in pip while doing something similar so I think we should *at least* fall back to something if ``locale.getpreferredencoding()`` returns ``None`` or we should give a better error than:

```
>> codecs.getwriter(None)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/codecs.py", line 991, in getwriter
    return lookup(encoding).streamwriter
TypeError: must be string, not None
```

dstufft:
Also the current system does technically break the Python feature ``PYTHONIOENCODING``:

```
$ PYTHONIOENCODING=ASCII LANG=en_US.UTF-8 python -c "import sys,locale; print((sys.stdout.encoding, locale.getpreferredencoding()))"
('ASCII', 'UTF-8')
```

kyleknap:
@dstufft that's fair about defaulting to something not ``None`` given that straight will not work.  I have not seen the error in the CLI before that you linked to, but it looks like it could be an edge case.

Looking at past issues, it appears that that error you linked as appeared before in the CLI: https://github.com/aws/aws-cli/issues/1898, but it is related to the s3 commands, which has a completely different logic in handling encodings: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/utils.py#L272-L312. And it does not look at preferred encoding at all. So it is unfortunate the two are inconsistent.

I'm still a little hesitant on switching the primary mechanism for determining encoding to something different than ``locale.getpreferredencoding()`` without doing a good amount of research on the potential affects of it. How would you feel about trying ``locale.getpreferredencoding()`` first and then relying on the encoding for ``sys.stdout`` and then ``ascii`` as a last resort if ``locale.getpreferredencoding()`` returns ``None`` just for standard output? 

For standard error though, I think it is the appropriate approach to rely on the encoding of  ``sys.stderr`` because previously we were just doing a ``sys.stderr.write()`` for errors and so defaulting to ``locale.getpreferredencoding()`` may cause a change in behavior if we make that the first thing to try.

dstufft:
@kyleknap I think it would probable be better to either do ``sys.stdout.encoding`` first or not at all for ``sys.stdout``. Doing it in the wrong order is a bit strange but more importantly I think ``PYTHONIOENCODING`` will still be broken, but on certain edge cases it will not be broken (the edge case where ``locale.getpreferredencoding()`` returns ``None``) and I think it is a better place to be to be consistently broken rather than inconsistently broken.

*Personally* I think that checking ``sys.stdout.encoding`` first is the right thing to do, but if we don't want to change that then I'd go with just not checking that attribute at all for ``sys.stdout``.

dstufft:
Ok, I've implemented all of the feedback *except* for what to do about ``sys.stdout.encoding``. If @kyleknap would prefer not touching that path still then I'll throw a flag on ``_get_text_writer()`` that controls whether it inspects ``stream.encoding`` or not.

kyleknap:
@dstufft Yeah I rather not change the code path to avoid potential changes in behavior for stdout. I think not checking the ``sys.stdout.encoding`` is fine for consistency reasons. If we run into issues down the road related to ``locale.getpreferedencoding()``, then it would be good to revisit this logic. Not sure if @jamesls had any opinions on this. Otherwise the update you have so far looks good.

dstufft:
@kyleknap Ok. I was thinking about this too, and I think if we don't check the encoding attribute for ``sys.stdout`` we also shouldn't for ``sys.stderr``, otherwise we can get into a situation where ``sys.stdout`` and ``sys.stderr`` are using different encodings at the same time, which seems worse to me than just breaking things like ``PYTHONIOENCODING``.

dstufft:
Ok, this should all be working now without changing the existing behavior.

kyleknap:
@dstufft 
Yeah I think that makes sense. Another thing that I was thinking that might boost this line of thinking is that while the CLI is a python-based tool, the users of the CLI may not really have any knowledge of Python, as it is just an implementation detail that it is written in Python, and so to have them knowing about things like ``PYTHONIOENCODING`` may be a bit of a stretch if they run into encoding issues. Also controlling the encoding through locale settings is a bit more intuitive and common from a command line user perspective, which makes sense to rely on ``locale.getpreferredencoding()``.

Let me know if that makes sense.

dstufft:
Yea, AFAICT ``stream.encoding`` defaults to ``locale.getpreferredencoding()`` anyways, except in cases where someone has overridden it either by something like ``PYTHONIOENCODING`` or monkeypatching or something.

PR #2375, opened 2017-01-10
update register domain example

mwunderl:


dstufft:
This appears to be correct. When I select 0 the UI shows it is valid for zero days.

Issue #2369, opened 2017-01-07
awscli requires python-six but does not declare it


Fale:
Also, it would be nice to know which minimum version of six is required

dstufft:
``awscli`` does not itself depend on ``six``, but instead it uses the version of six that comes with ``botocore``. However, instead of depending on ``six``, ``botocore`` bundles a copy internally to use so that it does not need to declare a dependency. Inspecting ``botocore/vendored/six.py`` indicates that it is currently using v1.8.0.

Are you getting an error where it's failing to import six?

Fale:
yes, on machines with python-six 1.3 installed (see https://bugzilla.redhat.com/show_bug.cgi?id=1409681)

dstufft:
Yes, this does appear to be a bug, it looks like we actually *sometimes* use the bundled version of six and *sometimes* don't. It generally works because we pick up a transitive dependency on six via ``python-dateutil`` . This also appears to affect ``botocore`` as well.

I'm not 100% sure if we should fix this by rewriting all uses of ``six`` to use the one bundled in botocore or if we should just depend on ``six`` and stop bundling it. In either case we should probably be consistent.

Fale:
Uhm, why not removing the bundled version and depend on the "external" six? I mean, the huge majority of packages depend on six, so I guess 99% (if not all) the users already have it

dstufft:
That decision predates me, so I'm not sure!

Fale:
I'm sure that when it was done had reasons, but I wonder if today it still makes sense or not, and since that code needs to be touch either way, probably is a good moment to think about it

kyleknap:
Looks like it is a bad import in the ``cloudformation package/deploy`` commands. In general in the CLI codebase, if we need to use six, we should be importing from the [compat.py](https://github.com/aws/aws-cli/blob/develop/awscli/compat.py#L17). The problem is it is easy to forget that as ``python-dateutil`` has that dependency so if you import it from the wrong location, it still may work.

Here is the issue on why we vendored six: https://github.com/aws/aws-cli/issues/1051. It is related to how OS X pre-installs pip and sets its precedence before the six installed by pip.

I will look into updating the code so that import is correct.

dstufft:
@kyleknap Perhaps a linter should be added?

Fale:
@kyleknap that seems to be a 2014 problem. I wonder if it's still current or if it has been fixed in the meantime

kyleknap:
@dstufft 
[...]

PR #2393, opened 2017-01-18
Fix an import

kyleknap:
Was importing from indirect six dependency when should have been using the
vendored dependency or the urlparse module from compat

Fixes https://github.com/aws/aws-cli/issues/2369

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -19,7 +19,7 @@ import contextlib
 import uuid
 from awscli.compat import six
 
-from six.moves.urllib.parse import urlparse, parse_qs
+from awscli.compat import urlparse
 from contextlib import contextmanager
 from awscli.customizations.cloudformation import exceptions
 from awscli.customizations.cloudformation.yamlhelper import yaml_dump, \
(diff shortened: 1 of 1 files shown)


dstufft:
Should we be importing from the vendored dependencies in the test suite too? Because I see:

```
$ egrep -RiIn "^(import six|from six)" *
awscli/customizations/cloudformation/artifact_exporter.py:22:from six.moves.urllib.parse import urlparse, parse_qs
tests/functional/ec2/test_bundle_instance.py:18:from six.moves import cStringIO
tests/functional/ses/test_send_email.py:17:from six.moves import cStringIO
tests/integration/customizations/test_codecommit.py:20:from six import StringIO
tests/unit/customizations/cloudformation/__init__.py:14:import six, unittest
tests/unit/customizations/cloudformation/test_deploy.py:15:import six
tests/unit/customizations/codedeploy/test_push.py:18:from six import StringIO
tests/unit/customizations/test_codecommit.py:18:from six import StringIO
tests/unit/output/test_text_output.py:23:from six.moves import cStringIO
```

kyleknap:
Yeah let's clean those up as well. I am also inclined to add a linter for this if others are inclined as well because there is not really a good way to test this.

kyleknap:
@jamesls Yes, I plan to update the tests and also add the linter in this pull request. So keeping it in ``incorporating-feedback`` for now.

kyleknap:
We had some offline discussion about this, but in short, we decided to not block this PR on the creation of a linter as that may take some time to set up as that would be the project's first linter. Instead, I opened a separate issue to track the addition of a linter: https://github.com/aws/aws-cli/issues/2487

PR #2397, opened 2017-01-19
Add timeformatting option to cli

stealthycoin:
cc @dstufft @jamesls @kyleknap @JordonPhillips

### DIFF
--- a/.changes/next-release/feature-Configuration-22255.json
+++ b/.changes/next-release/feature-Configuration-22255.json
@@ -1,5 +1,0 @@
-{
-  "type": "feature",
-  "category": "Configuration",
-  "description": "A new configuration option cli_timestamp_format was added."
-}
--- a/.changes/next-release/feature-Configuration-22255.json
+++ b/.changes/next-release/feature-Configuration-22255.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "Configuration",
+  "description": "A new configuration option cli_timestamp_format was added."
+}
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@ aws-cli
 .. image:: https://travis-ci.org/aws/aws-cli.svg?branch=develop
    :target: https://travis-ci.org/aws/aws-cli
    :alt: Build Status
-   
+
 .. image:: https://badges.gitter.im/aws/aws-cli.svg
    :target: https://gitter.im/aws/aws-cli
    :alt: Gitter
@@ -186,25 +186,27 @@ In addition to credentials, a number of other variables can be
 configured either with environment variables, configuration file
 entries or both.  The following table documents these.
 
-=========== =========== ===================== ===================== ============================
-Variable    Option      Config Entry          Environment Variable  Description
-=========== =========== ===================== ===================== ============================
-profile     --profile   profile               AWS_DEFAULT_PROFILE   Default profile name
------------ ----------- --------------------- --------------------- ----------------------------
-region      --region    region                AWS_DEFAULT_REGION    Default AWS Region
------------ ----------- --------------------- --------------------- ----------------------------
-config_file                                   AWS_CONFIG_FILE       Alternate location of config
------------ ----------- --------------------- --------------------- ----------------------------
-output      --output    output                AWS_DEFAULT_OUTPUT    Default output style
------------ ----------- --------------------- --------------------- ----------------------------
-ca_bundle   --ca-bundle ca_bundle             AWS_CA_BUNDLE         CA Certificate Bundle
------------ ----------- --------------------- --------------------- ----------------------------
-access_key              aws_access_key_id     AWS_ACCESS_KEY_ID     AWS Access Key
------------ ----------- --------------------- --------------------- ----------------------------
-secret_key              aws_secret_access_key AWS_SECRET_ACCESS_KEY AWS Secret Key
------------ ----------- --------------------- --------------------- ----------------------------
-token                   aws_session_token     AWS_SESSION_TOKEN     AWS Token (temp credentials)
-=========== =========== ===================== ===================== ============================
+================ =========== ===================== ===================== ============================
+Variable         Option      Config Entry          Environment Variable  Description
+================ =========== ===================== ===================== ============================
+profile          --profile   profile               AWS_DEFAULT_PROFILE   Default profile name
+---------------- ----------- --------------------- --------------------- ----------------------------
+region           --region    region                AWS_DEFAULT_REGION    Default AWS Region
+---------------- ----------- --------------------- --------------------- ----------------------------
+config_file                                        AWS_CONFIG_FILE       Alternate location of config
+---------------- ----------- --------------------- --------------------- ----------------------------
+output           --output    output                AWS_DEFAULT_OUTPUT    Default output style
+---------------- ----------- --------------------- --------------------- ----------------------------
+ca_bundle        --ca-bundle ca_bundle             AWS_CA_BUNDLE         CA Certificate Bundle
+---------------- ----------- --------------------- --------------------- ----------------------------
+access_key                   aws_access_key_id     AWS_ACCESS_KEY_ID     AWS Access Key
+---------------- ----------- --------------------- --------------------- ----------------------------
+secret_key                   aws_secret_access_key AWS_SECRET_ACCESS_KEY AWS Secret Key
+---------------- ----------- --------------------- --------------------- ----------------------------
+token                        aws_session_token     AWS_SESSION_TOKEN     AWS Token (temp credentials)
+---------------- ----------- --------------------- --------------------- ----------------------------
+timestamp_format             cli_timestamp_format                        Ouput format of timestamps
+================ =========== ===================== ===================== ============================
 
 ^^^^^^^^
 Examples
--- a/README.rst
+++ b/README.rst
@@ -186,27 +186,27 @@ In addition to credentials, a number of other variables can be
 configured either with environment variables, configuration file
 entries or both.  The following table documents these.
 
-================ =========== ===================== ===================== ============================
-Variable         Option      Config Entry          Environment Variable  Description
-================ =========== ===================== ===================== ============================
-profile          --profile   profile               AWS_DEFAULT_PROFILE   Default profile name
----------------- ----------- --------------------- --------------------- ----------------------------
-region           --region    region                AWS_DEFAULT_REGION    Default AWS Region
----------------- ----------- --------------------- --------------------- ----------------------------
-config_file                                        AWS_CONFIG_FILE       Alternate location of config
----------------- ----------- --------------------- --------------------- ----------------------------
-output           --output    output                AWS_DEFAULT_OUTPUT    Default output style
----------------- ----------- --------------------- --------------------- ----------------------------
-ca_bundle        --ca-bundle ca_bundle             AWS_CA_BUNDLE         CA Certificate Bundle
----------------- ----------- --------------------- --------------------- ----------------------------
-access_key                   aws_access_key_id     AWS_ACCESS_KEY_ID     AWS Access Key
----------------- ----------- --------------------- --------------------- ----------------------------
-secret_key                   aws_secret_access_key AWS_SECRET_ACCESS_KEY AWS Secret Key
----------------- ----------- --------------------- --------------------- ----------------------------
-token                        aws_session_token     AWS_SESSION_TOKEN     AWS Token (temp credentials)
----------------- ----------- --------------------- --------------------- ----------------------------
-timestamp_format             cli_timestamp_format                        Ouput format of timestamps
-================ =========== ===================== ===================== ============================
+==================== =========== ===================== ===================== ============================
+Variable             Option      Config Entry          Environment Variable  Description
+==================== =========== ===================== ===================== ============================
+profile              --profile   profile               AWS_DEFAULT_PROFILE   Default profile name
+-------------------- ----------- --------------------- --------------------- ----------------------------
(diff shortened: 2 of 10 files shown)


tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,78 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Change the scalar response parsing behavior for the AWS CLI.
```
stealthycoin: Looks like I copy/pasted out a chunk of documentation with my license

JordonPhillips:
In particular, the readme and the [config vars](https://github.com/aws/aws-cli/blob/develop/awscli/topics/config-vars.rst) docs.

stealthycoin:
Documentation Screenshot:
screen shot 2017-01-19 at 4 02 32 pm

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,78 @@
...
+        self.environ = {
+            'AWS_DATA_PATH': os.environ['AWS_DATA_PATH'],
+            'AWS_DEFAULT_REGION': 'us-east-1',
+            'AWS_ACCESS_KEY_ID': 'access_key',
+            'AWS_SECRET_ACCESS_KEY': 'secret_key',
+            'AWS_CONFIG_FILE': '',
+            'AWS_SHARED_CREDENTIALS_FILE': '',
+        }
+        self.environ_patch = mock.patch('os.environ', self.environ)
+        self.environ_patch.start()
```
kyleknap: Hmm... Any reason you did not use ``BaseAWSCommandParamsTest``: https://github.com/aws/aws-cli/blob/develop/awscli/testutils.py#L338

There is some cleanup you are missing with removing patches and cleaning up the ``FileCreator`` files.

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_iso_parser(self):
+        response_dict = {
+            'headers': '',
+            'status_code': 200,
+            'body': b'0'
+        }
+        factory = self.driver.session.get_component('response_parser_factory')
+        parser = factory.create_parser('json')
+        mock_shape = mock.MagicMock()
```
kyleknap: Not a big fan of mocking here especially without a spec. As this is a functional test why aren't we using a stubbed client to get as close as possible to the actual workflow as possible or ``BaseAWSCommandParams.run_cmd``: https://github.com/aws/aws-cli/blob/develop/awscli/testutils.py#L424 and checking the output?

awscli/customizations/scalarparse.py
```diff
@@ -36,13 +39,29 @@ def identity(x):
     return x
 
 
+def iso_format(value):
+    return parse_timestamp(value).isoformat()
+
+
+def choose_default_parsers(timestamp_format):
```
kyleknap: I wonder if it makes sense to have this choose all of the parsers. It seems a little strange that it sets both the blob and timestamp options, but really you can only pass in a ``timestamp_format`` option. Wonder if it make sense to have a function:
```python
def add_timestamp_parser(session):
     # Sets the timestamp parser to the session based on the config
```
That way you can encompass all of the timestampy related functionality under one function and makes it easier to read.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -575,7 +575,7 @@ def test_copy_metadata_directive(self):
         response = self.head_object(bucket_name, new_key)
         # These values should have the metadata of the source object
         metadata_ref = copy.copy(metadata)
-        metadata_ref['Expires'] = 'Thu, 01 Jan 1970 00:00:00 GMT'
+        metadata_ref['Expires'] = '1970-01-01T00:00:00+00:00'
```
stealthycoin: Good point. Looks like the integration test is picking up the actual value I had in my config file which is bad.

tests/functional/test_timeformat.py
stealthycoin: All the stubbed out clients (that I found) mock out the chunk that does the parsing so they weren't very useful for this test. You provide a parsed_response and it skips the whole request/response/parsing cycle. This seemed a lot simpler than writing an entire mock wire response test class, all I really wanted to test was the parsing of a timestamp not a whole response.

tests/integration/customizations/s3/test_plugin.py
kyleknap: I think it is because you patch the environment and never return state in the functional tests and those are run first.

awscli/customizations/scalarparse.py
stealthycoin: I think the issue with that is the factory that botocore pulls the parsers from currently has to have its defaults set all at once with `set_parser_defaults` which means whatever is setting the timestamp parser also has to set the blob parser since we override that with `identity`. I didn't look too much into reworking the session's response parser factory.

kyleknap: Does it? Looking at the implementation: https://github.com/boto/botocore/blob/develop/botocore/parsers.py#L112. It does not appear to be the case. It only updates based on the kwargs pased in. So it should retain defaults unless you explicitly override them.

stealthycoin: Yup looks like I completely misread that. I'll reshuffle things to make more sense.

JordonPhillips:
Still need a readme update

awscli/customizations/scalarparse.py
```diff
@@ -36,13 +40,40 @@ def identity(x):
...
+    # This function can be called as the first handler for
+    # the session-initialized event, which happens before a
+    # profile can be created, even if the command would have
+    # successfully created a profile. Instead of crashing here
+    # on a ProfileNotFound the CLI should just use 'none'.
+    try:
+        timestamp_format = session.get_scoped_config().get(
+            'cli_timestamp_format',
+            'none')
+    except ProfileNotFound:
```
kyleknap: When does this exception happen and why are we catching this? I feel like this exception is one that is typically propogated to the user.

README.rst
```diff
@@ -186,25 +186,27 @@ In addition to credentials, a number of other variables can be
...
+---------------- ----------- --------------------- --------------------- ----------------------------
+ca_bundle        --ca-bundle ca_bundle             AWS_CA_BUNDLE         CA Certificate Bundle
+---------------- ----------- --------------------- --------------------- ----------------------------
+access_key                   aws_access_key_id     AWS_ACCESS_KEY_ID     AWS Access Key
+---------------- ----------- --------------------- --------------------- ----------------------------
+secret_key                   aws_secret_access_key AWS_SECRET_ACCESS_KEY AWS Secret Key
+---------------- ----------- --------------------- --------------------- ----------------------------
+token                        aws_session_token     AWS_SESSION_TOKEN     AWS Token (temp credentials)
+---------------- ----------- --------------------- --------------------- ----------------------------
+timestamp_format             cli_timestamp_format                        Ouput format of timestamps
```
kyleknap: Any reason this is ``timestamp_format`` here but ``cli_timestamp_format`` in the topic guide?

awscli/topics/config-vars.rst
```diff
@@ -81,6 +83,12 @@ The valid values of the ``output`` configuration variable are:
 * table
 * text
 
+``cli_timestamp_format`` controls the format of timestamps displayed by the AWS CLI.
+The valid values of the ``cli_timestamp_format`` configuration varaible are:
+
+* none - Display the timestamp exactly as recieved.
```
kyleknap: A little ambiguous. May want to include wording about it being the unmodified value received over the wire or from HTTP response.

tests/unit/customizations/test_scalarparse.py
```diff
@@ -30,10 +35,43 @@ def test_identity(self):
...
         session.get_component.assert_called_with('response_parser_factory')
         factory = session.get_component.return_value
+        expected = [call(blob_parser=scalarparse.identity),
+                    call(timestamp_parser=scalarparse.identity)]
+        self.assertEqual(factory.set_parser_defaults.mock_calls,
+                         expected)
+
+    def test_choose_none_timestamp_formatter(self):
+        session = Mock()
+        session.get_scoped_config.return_value.get.return_value = 'none'
```
kyleknap: Can we do the following to make this more specific and avoid too much mocking magic?
1) Use a botocore session as the spec for the mock
2) Set the get_scoped_config return_value to a dictionary with the timestamp key instead of mocking out the entire get method.

tests/functional/test_timeformat.py
kyleknap: Right so. Another option you can do is actually mock out the underlying session.send of the endpoint:
https://github.com/aws/aws-cli/blob/develop/tests/unit/test_testutils.py

awscli/customizations/scalarparse.py
kyleknap: Nevermind, I see the comment now. Might be good to move the comment to under the exception as I completely missed reading that comment

.changes/next-release/feature-Configuration-22255.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "Configuration",
```
kyleknap: Why are there two change log entries? Let's just use 61155 and mention the name ``cli_timestamp_format``

tests/functional/test_timeformat.py
kyleknap: Note for any other reviewers: ``BaseAWSCommandParamsTest`` is not being used because we need to mock out the wire response and patching ``make_request`` on endpoint also patches out the response parsing.

tests/unit/customizations/test_scalarparse.py
```diff
@@ -30,10 +36,89 @@ def test_identity(self):
...
+        session = Mock(spec=Session)
+        session.get_scoped_config.side_effect = ProfileNotFound(profile='foo')
+        factory = session.get_component.return_value
+        scalarparse.add_scalar_parsers(session)
         factory.set_parser_defaults.assert_called_with(
-            blob_parser=scalarparse.identity,
             timestamp_parser=scalarparse.identity)
+
+
+class TestCLITimestampParser(unittest.TestCase):
```
kyleknap: So this looks more like a functional test. Since this is a super set of what your other functional test you added tests, I would suggest just to remove ``TestParsers`` and add ``TestCLITimestampParser`` to/from ``tests/functional/test_timeformat.py``.

Also it would be great if there was a functional test for both ``none`` being set and having it not configured at all for completeness.

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,92 @@
...
+            'AWS_CONFIG_FILE': ''
+        }
+        self.environ_patch = mock.patch('os.environ', self.environ)
+        self.environ_patch.start()
+
+        self.wire_response = json.dumps({
+            'builds': [{
+                'startTime': 0,
+            }]
+        }).encode()
```
jamesls: I'd prefer we explicitly specify encodings here given we explicitly call out utf-8 when decoding in botocore.

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,92 @@
...
+import datetime
+from dateutil.tz import tzlocal
+from awscli.testutils import unittest, FileCreator, capture_output
+from awscli.clidriver import create_clidriver
+
+
+class TestCLITimestampParser(unittest.TestCase):
+    def setUp(self):
+        self.files = FileCreator()
+        self.environ = {
```
jamesls: This seems repeated from BaseCLIDriver.  Is there a reason to not reuse that class?

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,92 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesls: 2017

awscli/topics/config-vars.rst
```diff
@@ -81,6 +83,12 @@ The valid values of the ``output`` configuration variable are:
 * table
 * text
 
+``cli_timestamp_format`` controls the format of timestamps displayed by the AWS CLI.
+The valid values of the ``cli_timestamp_format`` configuration varaible are:
+
+* none - Display the timestamp exactly as received from the HTTP response.
+* iso8601 - Reformat timestamp using iso8601.
```
jamesls: Worth documenting it will use your local tz.

tests/functional/test_timeformat.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def test_iso(self):
+        self.environ['AWS_CONFIG_FILE'] = self.files.create_file(
+            'iso',
+            '[default]\ncli_timestamp_format = iso8601\n')
+        self.driver = create_clidriver()
+        expected_time = datetime.datetime.fromtimestamp(0).replace(
+            tzinfo=tzlocal()).isoformat()
+
+        with capture_output() as captured:
```
jamesls: The same structure here is duplicated across three tests.  Can we refactor some of this into testutils?  Seems similar to `BaseAWSCommandParamsTest.run_command`, but one layer lower than `Endpoint.make_request`.  Probably can do something similar to `run_cmd`.

tests/functional/test_timeformat.py
stealthycoin: Originally yes, but there has been some refactoring so I'll add something to testutils for this mocking `botocore.endpoint.Session.send` in general.

PR #2407, opened 2017-01-25
Refresh Bundled Installer

dstufft:
Refresh the bundled installer, moving over to primarily using wheels which will speed up installation of the installer as well as removes the need (or even attempt) of a compiler for PyYAML on the end user machine. In addition, it allows specifying alternative locations for awscli and botocore, which will typically be used to specify a branch to install botocore from.

An example of an alternate botocore branch usage would be:

```console
$ scripts/make-bundle --botocore 'git+https://github.com/JordonPhillips/botocore.git@extract-aggregate#egg=botocore'
```

### DIFF
--- a/scripts/make-bundle
+++ b/scripts/make-bundle
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """Script to create self contained install.
 
 The goal of this script is simple:
@@ -13,6 +13,8 @@ ecosystem.
 
 """
 
+from __future__ import print_function
+
 import argparse
 import os.path
 import re
@@ -24,13 +26,67 @@ import tempfile
 import zipfile
 
 
-if sys.version_info[:2] < (3, 5):
-    sys.exit("Must use Python >= 3.5 to run this script.")
-
-
 PYTHON_VERSIONS = ["2.6", "2.7", "3.3", "3.4", "3.5", "3.6"]
 
 VIRTUALENV_VERSION = "15.1.0"
+
+
+if sys.version_info[0] == 2:
+    PY2 = True
+    FileNotFoundError = OSError
+
+    _find_unsafe = re.compile(r'[^\w@%+=:,./-]').search
+
+    def shlex_quote(s):
+        """Return a shell-escaped version of the string *s*."""
+        if not s:
+            return "''"
+        if _find_unsafe(s) is None:
+            return s
+
+        # use single quotes, and put single quotes into double quotes
+        # the string $'b is then quoted as '$'"'"'b'
+        return "'" + s.replace("'", "'\"'\"'") + "'"
+
+    class subprocess_CompletedProcess(object):
+        """A process that has finished running.
+
+        This is returned by run().
+
+        Attributes:
+          args: The list or str args passed to run().
+          returncode: The exit code of the process, negative for signals.
+          stdout: The standard output (None if not captured).
+          stderr: The standard error (None if not captured).
+        """
+        def __init__(self, args, returncode, stdout=None, stderr=None):
+            self.args = args
+            self.returncode = returncode
+            self.stdout = stdout
+            self.stderr = stderr
+
+        def __repr__(self):
+            args = ['args={!r}'.format(self.args),
+                    'returncode={!r}'.format(self.returncode)]
+            if self.stdout is not None:
+                args.append('stdout={!r}'.format(self.stdout))
+            if self.stderr is not None:
+                args.append('stderr={!r}'.format(self.stderr))
+            return "{}({})".format(type(self).__name__, ', '.join(args))
+
+        def check_returncode(self):
+            """Raise CalledProcessError if the exit code is non-zero."""
+            if self.returncode:
+                raise subprocess.CalledProcessError(
+                    self.returncode,
+                    self.args,
+                    self.stdout,
+                    self.stderr,
+                )
+else:
+    PY2 = False
+    shlex_quote = shlex.quote
+    subprocess_CompletedProcess = subprocess.CompletedProcess
 
 
 class MissingPythonVersion(Exception):
@@ -44,14 +100,16 @@ def pip(python_version):
 
 
 def run_command(command, **kwargs):
-    print("  + {}".format(" ".join([shlex.quote(c) for c in command])))
+    print("  + {}".format(" ".join([shlex_quote(c) for c in command])))
 
     kwargs["stdin"] = subprocess.PIPE
     kwargs["stdout"] = subprocess.PIPE
     kwargs["stderr"] = subprocess.STDOUT
-    kwargs.setdefault("encoding", sys.stdout.encoding)
+    if not PY2:
+        kwargs.setdefault("encoding", sys.stdout.encoding)
 
-    with subprocess.Popen(command, **kwargs) as process:
+    process = subprocess.Popen(command, **kwargs)
+    try:
         # Close the stdin to make it an error for the subprocess to prompt.
         process.stdin.close()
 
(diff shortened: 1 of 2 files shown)


dstufft:
Note: Another benefit of this script is that there is no longer any needed to add any sort of overrides for a specific package on a particular python version (e.g. argparse on 2.6), it will automatically download all required dependencies for all versions of Python. The supported versions of python are hardcoded into the ``scripts/make-bundle`` script OR they can be overridden by specifying ``-p X.Y`` for each supported python  (e.g. ``scripts/make-bundle -p 2.7 -p 3.6``).

JordonPhillips:

scripts/make-bundle
```diff
@@ -12,171 +12,176 @@ interface for those not familiar with the python
...
+            output_dir=os.path.join(bundle_dir, "packages"),
+            pythons=pythons,
+        )
+
+        create_bootstrap_script(bundle_dir)
+
+        output = os.path.abspath(os.path.join(tmpdir, "awscli-bundle.zip"))
+        zip_dir(bundle_dir, output)
+
+        os.makedirs(output_dir, exist_ok=True)
```
dstufft: Yea and for ``subprocess.run()`` and ``tempfile.TemporaryDirectory()``.

scripts/make-bundle
```diff
@@ -12,171 +12,176 @@ interface for those not familiar with the python
...
+    with zipfile.ZipFile(output, "w", zipfile.ZIP_DEFLATED) as zfp:
+        for root, dirnames, filenames in os.walk(bundle_dir):
+            base = os.path.commonpath([tmpdir, root])
+            for filename in filenames:
+                path = os.path.join(root, filename)
+                zfp.write(path, arcname=os.path.relpath(path, base))
+
+
+def create_bundled_installer(aws_cli=".", botocore=None, pythons=None,
+                             output_dir="."):
```
dstufft: Resolved.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+                 "--no-cache-dir"] + requirements,
+                stdout=subprocess.PIPE,
+                env=env,
+                check=True,
+            )
+
+            # We need to rename the PyYAML file, we've disabled the building of
+            # it's C extension so it is now pure Python, HOWEVER, because of
+            # the way it works wheel still thinks it is a binary file.
+            # TODO: It'd be great if this wasn't required.
```
kyleknap: Yeah that's unfortunate. I am a little hesitant on relying on a regex search to determine the location of this file mainly because if the name of the file changes, it can cause issues. Is there any way we can better guarantee that we get a wheel file out of it?

scripts/make-bundle
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
dstufft: It wouldn't be very hard to add a small version check at the top and bail out early, I can do that np.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+        # We want to loop over every version of Python we support and download
+        # or build wheels for them.
+        first = False
+        for python in sorted(pythons, reverse=True):
+            print("Downloading Dependencies for {}".format(python))
+
+            # Create our virtual environment.
+            venv_path = os.path.abspath(os.path.join(tmpdir, python))
+            python_path = os.path.join(venv_path, "bin", "python")
+            subprocess.run(
```
dstufft: Ah yes, my dev box has them all so I didn't run into this. I could either just improve the error message *or* I could make it do like what tox does and just print a big warning if a specified Python does not exist. I personally lean towards keeping it a hard error but I don't feel _super_ strongly about that.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+                ["virtualenv", "-ppython{}".format(python), venv_path],
+                stdout=subprocess.PIPE,
+                check=True,
```
dstufft: So I was doing ``subprocess.PIPE`` largely to hide the output, but I didn't want to hide the ``stderr`` output so that it didn't silence warnings/errors. So I think that unless we're managing the ``stdout``/``stderr`` output in a way that is more complex than either not outputting it or outputting it then we don't want to set that here.

dstufft: So I guess the question is, do we think that the ``stdout`` output is useful? Personally I didn't think it was but I'm happy to change it, but either way we probably don't want to silence ``stderr``.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+            )
+
+            # Ensure that we have the latest versions of all of our tools
+            # installed in the virtual environment.
+            subprocess.run(
```
dstufft: I can bring the output back for sure.

scripts/make-bundle
kyleknap: I like hard error as well with improved error message.

scripts/make-bundle
kyleknap: Makes sense. I wonder if it makes sense to not hide standard output as well? For me the more output the better, unless it is way too much output.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+            )
+
+            if not first:
+                first = True
+                # We need to download and bundle virtualenv, but we need it to
+                # be the sdist since it won't actually be installed.
+                subprocess.run(
+                    [python_path, "-m", pip(python), "download",
+                     "--no-cache-dir", "--no-binary=virtualenv",
+                     "-d", output_dir, "virtualenv"],
```
dstufft: Hmm. Yes there is. I'm happy to pin it again.

scripts/make-bundle
kyleknap: So in the previous script, I believe we hid stdout as well. I think if we are printing out the commands we ran like before, that should be enough information for debugging.

scripts/make-bundle
dstufft: The name of the file *shouldn't* change without something like a new PEP, the filename format for a wheel is baked into the language of one of Wheel's PEPs.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+        )
+
+        create_bootstrap_script(bundle_dir)
+
+        output = os.path.abspath(os.path.join(tmpdir, "awscli-bundle.zip"))
+        zip_dir(bundle_dir, output)
+
+        os.makedirs(output_dir, exist_ok=True)
+        if os.path.exists(os.path.join(output_dir, os.path.basename(output))):
+            os.remove(os.path.join(output_dir, os.path.basename(output)))
```
dstufft: No, ``shutil.move`` won't overwrite.

scripts/make-bundle
```diff
@@ -12,171 +12,196 @@ interface for those not familiar with the python
...
+    parser.add_argument(
+        "--botocore",
+        metavar="REQUIREMENT",
+        help="A pip requirement for what botocore to install.",
+    )
+    parser.add_argument(
+        "--python", "-p",
+        metavar="Python",
+        action="append",
+        help=("A version of Python (X.Y) that this bundle will be valid for. "
```
dstufft: Good catch, I can do that.

scripts/make-bundle
kyleknap: Sounds good. Should be fine if it is specified in the PEP.

scripts/make-bundle
kyleknap: Yeah I think I would feel more comfortable if we did. If there is a new functionality that we want to pick up, bumping the version and testing out the new installers have not been difficult. There really has not been any issues in the past, but it is better to be safe.

scripts/make-bundle
kyleknap: Interesting. Makes sense then.

scripts/make-bundle
dstufft: I've added this.

dstufft:
Just a note, from the discussion with @kyleknap I've also added a ``--virtualenv`` parameter which will allow overriding the version of virtualenv used, but by default it will pin to the version specified at the top of the script.

dstufft:
Ok, this should now work on Python 2.7 and Python 3.

PR #2414, opened 2017-01-30
Adding example for deleting two Docker images from ECR

JasonSwindle:
Howdy,

This PR covers providing an example of deleting two or more Docker images from ECR.


jamesls:
Looks like the build is failing because this was sent to the master branch.  Could you resend this to the develop branch?  Thanks.

JasonSwindle:
Will do!  Thanks for the fast follow-up! :D

PR #2425, opened 2017-02-04
Some fixes to PEP8 code style

dslackw:


awscli/__init__.py
```diff
@@ -15,6 +15,8 @@
 ----
 A Universal Command Line Environment for Amazon Web Services.
 """
+
```
JordonPhillips: I'm pretty sure PEP8 doesn't require spaces between module level comments and import statements, though I could be wrong.

awscli/testutils.py
```diff
@@ -19,17 +19,19 @@
...
 import shutil
-import time
+# import time       imported but unused
 import json
 import logging
 import tempfile
 import platform
 import contextlib
-import string
+# import string     imported but unused
```
JordonPhillips: I would prefer that this just gets removed rather than commented out

awscli/testutils.py
```diff
@@ -783,12 +788,14 @@ def put_object(self, bucket_name, key_name, contents='', extra_args=None):
         if extra_args is not None:
             call_args.update(extra_args)
         response = client.put_object(**call_args)
+        # variable response assigned to but never used
```
JordonPhillips: Again, I would prefer this gets removed rather than adding a comment. As is this doesn't actually fix the issue.

awscli/testutils.py
koreno: yep, git is much better at keeping history than comments... :)

kdaily:
This looks like it's going to be hard to address the conflicts and the originating repository and user have been deleted. We have a few other open issues to address code style. Closing this.

Issue #1346, opened 2015-05-27
Unable to set `charset` when mime-types are guessed (S3)


gf3:
We are syncing a directory of various file types to an S3 bucket and `aws-cli` is correctly guessing the mime-types, however in our case it's important that it also append the `charset`. For example the guessed content-type for `index.html` might look like this:

```
Content-Type: text/html
```

But we'd like a way to be able to tell `aws-cli` that the `charset` for all the synced files is `UTF8` for instance:

```
Content-Type: text/html; charset=utf-8
```

Version: `aws-cli/1.7.26 Python/2.7.6 Darwin/14.4.0`

jdjkelly:
+1

chainlink:
:+1:

mathiasbynens:
:+1:

kieran:
:panda_face:

adammeghji:
:+1:

darcyclarke:
:+1:

kyleknap:
Marking as feature request. Any suggestions on how you would like to see it exposed in the CLI would be appreciated.

gf3:
@kyleknap perhaps via `--charset` switch? which would be appended to the guessed mime-type.

quiver:
You can explicitly set  `content-type` for `s3 cp/sync` and `s3api put-object` APIs.

For `s3 cp/sync`, use `--content-type` option.

```
$ aws s3 cp --content-type 'text/plain; charset=utf-8' index.html s3://BUCKET/index.html
upload: ./index.html to s3://BUCKET/index.html
$ aws s3api head-object --bucket BUCKET --key index.html
{
    "AcceptRanges": "bytes",
    "ContentType": "text/plain; charset=utf-8",
    "LastModified": "Thu, 28 May 2015 14:18:42 GMT",
    "ContentLength": 12,
    "ETag": "\"6f5902ac237024bdd0c176cb93063dc4\"",
    "Metadata": {}
}

$ aws s3 sync foo s3://BUCKET/foo  --content-type 'text/html; charset=utf-8'
upload: foo/index.html to s3://BUCKET/foo/index.html
$ aws s3api head-object --bucket BUCKET --key foo/index.html
{
    "AcceptRanges": "bytes",
    "ContentType": "text/html; charset=utf-8",
    "LastModified": "Thu, 28 May 2015 14:30:54 GMT",
    "ContentLength": 12,
    "ETag": "\"6f5902ac237024bdd0c176cb93063dc4\"",
[...]

PR #2426, opened 2017-02-05
Use UTF-8 encoding when guessing a MIME type of text/html for S3 uploads.

ezzatron:
This PR addresses what I suspect is the most common cause of #1346, which is that `aws s3` guesses the MIME type of `.html` files as `text/html`, without any encoding specification.

This change simply alters any "guessed" `text/html` MIME type to be guessed as `text/html; charset=utf-8` instead. It's a bit of a stop-gap measure to address #1346, but given that the issue is now almost 2 years old, I thought this PR might be worth making in any case.

In terms of relevant statistics to justify this PR, [this somewhat dubious report](https://w3techs.com/technologies/details/en-utf8/all/all) suggests that:

> **UTF-8 is used by 88.5%** of all the websites whose character encoding we know.

My gut feeling also tells me that UTF-8 is a more sensible guess for encoding in 2017, rather than omitting the encoding from the `text/html` MIME type which, in the HTTP specification, means `ISO-8859-1`.

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -335,7 +335,10 @@ def guess_content_type(filename):
     If the type cannot be guessed, a value of None is returned.
     """
     try:
-        return mimetypes.guess_type(filename)[0]
+        guessed_type = mimetypes.guess_type(filename)[0]
+        if guessed_type == 'text/html':
+            return 'text/html; charset=utf-8'
+        return guessed_type
     # This catches a bug in the mimetype libary where some MIME types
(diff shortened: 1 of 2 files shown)


stealthycoin:
Thanks for the submission.

cc @jamesls @JordonPhillips @kyleknap @dstufft

ajkerr:
I'd love to see this default of utf-8 extended to application/javascript. Webpack generates UTF-8 characters for things like &nbsp; entities, which don't render correctly by browsers unless the file encoding is application/javascript; charset=utf-8.

This honestly seems like a sane default to me.

Issue #2422, opened 2017-02-02
Documentation: opsworks describe-elastic-load-balancers example


dumityty:
Looks like the example output for the opsworks describe-elastic-load-balancers command (https://github.com/aws/aws-cli/blob/develop/awscli/examples/opsworks/describe-elastic-load-balancers.rst) is the same as that of the opsworks describe-apps command (https://github.com/aws/aws-cli/blob/develop/awscli/examples/opsworks/describe-apps.rst)

stealthycoin:
Thanks for finding that. Opened #2434

PR #2434, opened 2017-02-09
Replace example with one from the correct operation

stealthycoin:
cc @JordonPhillips @jamesls @kyleknap @dstufft


stealthycoin:
From #2422

Issue #2435, opened 2017-02-09
aws cloudformation deploy  'list' object has no attribute 'split'


Jormundir:
```
aws --version
aws-cli/1.11.45 Python/2.7.13 Darwin/15.6.0 botocore/1.5.8
```

After creating a cloudformation template, when I attempt to deploy it, I get the python error in the title:
```
aws cloudformation deploy --template-file _build/cloudformation.yaml --stack-name lambda

'list' object has no attribute 'split'
```

stealthycoin:
Can you run this command again with `--debug`. That will help track it down.

Jormundir:
Huh, with no changes, this is working today. I'm guessing something must have changed with an API response?

flaub:
Just ran into this:
```
$ aws --version
aws-cli/1.11.47 Python/2.7.13 Darwin/16.4.0 botocore/1.5.10
```

```
$ aws --debug cloudformation deploy --template-file stack-uat.yml --stack-name portal-uat --capabilities CAPABILITY_IAM help
2017-02-12 03:48:41,568 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/1.11.47 Python/2.7.13 Darwin/16.4.0 botocore/1.5.10
2017-02-12 03:48:41,568 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['--debug', 'cloudformation', 'deploy', '--template-file', 'stack-uat.yml', '--stack-name', 'portal-uat', '--capabilities', 'CAPABILITY_IAM', 'help']
2017-02-12 03:48:41,568 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function add_scalar_parsers at 0x105a75050>
2017-02-12 03:48:41,568 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function inject_assume_role_provider_cache at 0x10577cc08>
2017-02-12 03:48:41,569 - MainThread - botocore.loaders - DEBUG - Loading JSON file: /usr/local/lib/python2.7/site-packages/botocore/data/cloudformation/2010-05-15/service-2.json
2017-02-12 03:48:41,583 - MainThread - botocore.hooks - DEBUG - Event service-data-loaded.cloudformation: calling handler <function register_retries_for_service at 0x1053a37d0>
2017-02-12 03:48:41,583 - MainThread - botocore.handlers - DEBUG - Registering retry handlers for service: cloudformation
[...]

PR #2438, opened 2017-02-09
Fixing bug in cloudformation command caused by missing type-checking

sanathkr:
Fixes https://github.com/aws/aws-cli/issues/2435

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 
 import yaml
+import six
 from yaml.resolver import ScalarNode, SequenceNode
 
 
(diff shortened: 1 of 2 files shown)


stealthycoin:
cc @JordonPhillips @jamesls @kyleknap @dstufft

awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 
 import yaml
+import six
```
kyleknap: So for ``six`` could you import from ``awscli.compat`` instead?: https://github.com/aws/aws-cli/blob/develop/awscli/compat.py#L17 The CLI does not have a direct dependency on ``six`` which would be the case for this import, but instead relies on botocore's vendored version of ``six`` and imports that in the ``compat`` module

sanathkr: Done. Pushed the change

kyleknap:
Awesome. Merging.

Issue #2317, opened 2016-12-12
BodyS3Location from CloudFormation


kyledecot:
When looking at the CloudFormation documentation I see there is an option for providing an openapi (Swagger) definition but I don't see that there is an mention of it when doing `aws cloudformation create-rest-api help`. 

Is there a way of building an API using the cli while providing a openapi definition?

jamesls:
Are you talking about creating an API gateway rest api?  If so you can use the `import-rest-api` to provide a swagger definition: http://docs.aws.amazon.com/cli/latest/reference/apigateway/import-rest-api.html

kyledecot:
@jamesiri Yes, that's exactly what I'm talking about.

PR #2451, opened 2017-02-18
Don't package Serverless API DefinitionUri property if not present

sanathkr:
Similar to https://github.com/aws/aws-cli/issues/2317


kyleknap:
Thanks! Merging.

PR #2511, opened 2017-03-28
Add str representer

DiegoPomares:


awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -53,12 +54,22 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
     return {cfntag: value}
 
 
+def str_representer(dumper, node):
+
+    if re.match(r'[-+]?[0-9][0-9_]*', node):
```
sanathkr: I would rather quote all strings. If a value is parsed as string, it should be okay to quote them. What do you think?

sanathkr:
Should fix https://github.com/awslabs/serverless-application-model/issues/91

DiegoPomares:
https://github.com/aws/aws-cli/pull/2556

awscli/customizations/cloudformation/yamlhelper.py
sanathkr: @DiegoPomares Can we quote all strings by default? Many people are running into this bug. #2556 gives a workaround, but I want to fix the original problem as well for YAML

pettyjamesm:
Is this PR ever going to land? This PR is over a year old and I just hit this bug today.

ramblingbiped:
I ran into this issue today trying to package a template with an account id based mapping.  I've got an account id that starts with a zero, that when packaged, has the quotes removed resulting in the value not being interpreted as a string.

Any updates on this getting fixed?  It looks like it has been an identified problem for at least two years now:  https://github.com/awslabs/serverless-application-model/issues/91

ttyniwa:
https://github.com/aws/aws-cli/pull/2511#issuecomment-469923988

I'm also troubled by the same problem.

stealthycoin:
This has since been addressed. In CLI v2 we use the `ruamel` safe loader which should avoid these problems. For CLI v1 you can supply the `--use-json` flag.

Issue #2518, opened 2017-03-30
awscli fails when running the as part of a Daemon which does not have a stdin


markdessain:
Currently I am running jobs via the Apache Airflow Scheduler (https://airflow.incubator.apache.org/)

It runs background jobs as a Daemon with only stdout and stderr attached, but no stdin - https://github.com/apache/incubator-airflow/blob/master/airflow/bin/cli.py#L824-L829 / https://pagure.io/python-daemon/blob/master/f/daemon/daemon.py#_255

I have discovered that while running under Python3 the awscli will only work if the stdin is set due to - https://github.com/aws/aws-cli/blob/b5fe9872abc6006f2a3093d4d079d23024bc3e0a/awscli/compat.py#L75

Currently it raises `AttributeError: 'NoneType' object has no attribute 'buffer'` as stdin is `None`

-------------

A full example showing it working under Python2 but not Python3 can be found here:

```

# Works
# ========
# mkvirtualenv awscli_python2 -p python
# pip install awscli==1.11.68
# pip install python-daemon==2.1.1
# python aws_stdin.py

# Does not work
# ========
# mkvirtualenv awscli_python3 -p python3
# pip install awscli==1.11.68
# pip install python-daemon==2.1.1
# python aws_stdin.py


import os
import sys
import daemon
from subprocess import Popen, STDOUT, PIPE

stdout = open('out', 'w+')

ctx = daemon.DaemonContext(
    stdout=stdout
)

with ctx:
    sp = Popen(['aws'], stdout=PIPE, stderr=STDOUT)

    for line in iter(sp.stdout.readline, b''):
        print(line)

```

---------

```
b'Traceback (most recent call last):\n'
b'  File "/Users/markdessain/Envs/aws_stdin/bin/aws", line 19, in <module>\n'
b'    import awscli.clidriver\n'
b'  File "/Users/markdessain/Envs/aws_stdin/lib/python3.5/site-packages/awscli/clidriver.py", line 26, in <module>\n'
b'    from awscli.formatter import get_formatter\n'
b'  File "/Users/markdessain/Envs/aws_stdin/lib/python3.5/site-packages/awscli/formatter.py", line 19, in <module>\n'
b'    from awscli.table import MultiTable, Styler, ColorizedStyler\n'
b'  File "/Users/markdessain/Envs/aws_stdin/lib/python3.5/site-packages/awscli/table.py", line 19, in <module>\n'
b'    from awscli.compat import six\n'
[...]

PR #2519, opened 2017-03-30
possible solution for #2518

markdessain:
Details have been included in the issue #2518

This is to stop running `aws` from failing if no stdin is set when running from Python3. This is not an issue for Python2.

Thoughts?


markdessain:
Closed - details listed in Issue #2518

PR #2522, opened 2017-04-03
Alias mturk command

JordonPhillips:
A command was renamed at the botocore level, this is providing a
hidden alias to continue supporting the old spelling.

This requires boto/botocore#1172

### DIFF
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -80,7 +80,7 @@ def alias_command(command_table, existing_name, new_name, hide_current=True):
 
     :type hide_current: bool
     :param hide_current: Whether to hide the current command name or the new
-        command name. If True, the current name is hidden. If False, the ne
+        command name. If True, the current name is hidden. If False, the new
         name is hidden.
     """
     current = command_table[existing_name]
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -66,11 +66,30 @@ def rename_command(command_table, existing_name, new_name):
     del command_table[existing_name]
 
 
-def alias_command(command_table, existing_name, new_name):
-    """Moves an argument to a new name, keeping the old as a hidden alias."""
+def alias_command(command_table, existing_name, new_name, hide_current=True):
+    """Moves an argument to a new name, keeping the old as a hidden alias.
+
+    :type command_table: dict
+    :param command_table: The full command table for the CLI or a service.
+
+    :type existing_name: str
+    :param existing_name: The current name of the command.
+
+    :type new_name: str
+    :param new_name: The new name for the command.
+
+    :type hide_current: bool
+    :param hide_current: Whether to hide the current command name or the new
+        command name. If True, the current name is hidden. If False, the ne
+        name is hidden.
+    """
     current = command_table[existing_name]
-    _copy_argument(command_table, existing_name, new_name)
-    current._UNDOCUMENTED = True
+    new = _copy_argument(command_table, existing_name, new_name)
+
+    if hide_current:
+        current._UNDOCUMENTED = True
+    else:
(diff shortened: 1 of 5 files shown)


tests/functional/mturk/test_alias.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.testutils import BaseAWSCommandParamsTest
+
+
+class TestAlias(BaseAWSCommandParamsTest):
+    def test_alias(self):
+        # This command was aliased, both should work
+        command_tempalte = 'mturk %s --qualification-type-id foo'
```
dstufft: ``command_template`` not ``command_tempalte``.

awscli/customizations/utils.py
```diff
@@ -66,11 +66,30 @@ def rename_command(command_table, existing_name, new_name):
...
+
+    :type existing_name: str
+    :param existing_name: The current name of the command.
+
+    :type new_name: str
+    :param new_name: The new name for the command.
+
+    :type hide_current: bool
+    :param hide_current: Whether to hide the current command name or the new
+        command name. If True, the current name is hidden. If False, the ne
```
kyleknap: ne --> new

tests/unit/customizations/test_utils.py
```diff
@@ -43,23 +43,49 @@ def test_rename_command_table(self):
...
+            utils.alias_command(command_table, old_name, new_name, False)
+
+        self._assert_commands_exist(old_name, new_name, handler)
+
+        # Verify that the new isn't documented
+        self.driver.main(['help'])
+        self.assert_not_contains(new_name)
+        self.assert_contains(old_name)
+
+    def _assert_commands_exist(self, old_name, new_name, handler):
```
kyleknap: Can we break up this general ``_assert_commands_exist`` into several other ``_assert_*()`` functions? It seems to be a bit more than checking if it is just existing. If I wanted to check if it existed, I would just run ``help`` on the command and not necessarily all of the other help commands. In the end it will just make it a bit easier to read when going through the test.

awscli/customizations/utils.py
```diff
@@ -66,11 +66,30 @@ def rename_command(command_table, existing_name, new_name):
...
+    :type command_table: dict
+    :param command_table: The full command table for the CLI or a service.
+
+    :type existing_name: str
+    :param existing_name: The current name of the command.
+
+    :type new_name: str
+    :param new_name: The new name for the command.
+
+    :type hide_current: bool
```
kyleknap: I'm debating if I like the ``hide_current`` parameter. While I am fine with the approach of reusing the ``alias_command`` and having a parameter that decides what is undocumented, I was a little confused at first before reading the docstring as I did not realize that by setting it to False, it would not hide the other command as well. I wonder if there is a better name to indicate this toggling side effect?

JordonPhillips: Yeah, I'm really not sure what to go with on that one.

kyleknap: Maybe make it a string parameter called ``hide_which`` where you can provide the values ``'new'``, ``'existing'``, or maybe ``None`` later down the road if needed?

jamesls: We use `make_hidden_alias` for doing this same thing with parameters.  Shouldn't we just have a `make_hidden_command_alias` as well to be consistent?  I had to read the docs a few times to understand what `hide_current` meant.

JordonPhillips: That sounds good

JordonPhillips:
Good to look at again

PR #2537, opened 2017-04-10
Handle default profile in configure get and have more useful errors

### LINKED ISSUE #2536: Reading/writing EMR key_pair_file configuration options behaves oddly
Version:

```
$ aws --version
aws-cli/1.11.75 Python/2.7.10 Darwin/15.6.0 botocore/1.5.38
```

[It's suggested that one can set a default key_pair_file argument here](https://github.com/aws/aws-cli/blob/master/awscli/customizations/emr/ssh.py#L25) by running `aws configure set emr.key_pair_file <value>`

By that token, I would expect `aws configure get emr.key_pair_file` to retrieve this item and to exit with a exit code of 0.

```
$ aws configure set emr.key_pair_file /tmp/foo

$ cat config
[default]
emr =
    key_pair_file = /tmp/foo

$ aws configure get emr.key_pair_file

$ echo $?
1
```

As you can see, setting this and trying to retrieve it exits with a non-zero exit code which makes it a pain to check for this config item being set as part of shell scripts prior to do other EMR-based commands (such as create-cluster).

As an aside, trying to get the top level `emr` config item fails too;

```
$ aws configure get emr

expected a character buffer object
```

Additionally this item doesn't show up when `aws configure list` is run either;

```
$ aws configure list
access_key     REDACTED                    shared-credentials-file
secret_key      REDACTED                    shared-credentials-file
    region                <not set>             None    None

```

JordonPhillips:
Fixes #2536 

Basically, `configure get` didn't properly handle the default profile if you did something like `aws configure get s3.signature_version`. In addition to that, it gave an opaque error message if you tried to get an entire subsection.

While we could potentially support getting subsections, we would need to print them off in the same way that they are written to the config file, which could be tricky. For now we should at least have a better error message.

### DIFF
--- a/awscli/customizations/configure/get.py
+++ b/awscli/customizations/configure/get.py
@@ -76,6 +76,8 @@ class ConfigureGetCommand(BasicCommand):
             remaining = parts[2:]
         else:
             profile_name = self._session.get_config_variable('profile')
+            if profile_name is None:
+                profile_name = 'default'
             config_name = parts[0]
             remaining = parts[1:]
 
--- a/awscli/customizations/configure/get.py
+++ b/awscli/customizations/configure/get.py
@@ -21,7 +21,7 @@ class ConfigureGetCommand(BasicCommand):
     NAME = 'get'
     DESCRIPTION = BasicCommand.FROM_FILE('configure', 'get',
                                          '_description.rst')
-    SYNOPSIS = ('aws configure get varname [--profile profile-name]')
+    SYNOPSIS = 'aws configure get varname [--profile profile-name]'
     EXAMPLES = BasicCommand.FROM_FILE('configure', 'get', '_examples.rst')
     ARG_TABLE = [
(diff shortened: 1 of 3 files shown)


awscli/customizations/configure/get.py
```diff
@@ -30,31 +34,44 @@ class ConfigureGetCommand(BasicCommand):
...
+
+        if isinstance(value, six.string_types):
             self._stream.write(value)
             self._stream.write('\n')
             return 0
+        elif isinstance(value, dict):
+            # TODO: add support for this. We would need to print it off in
+            # the same format as the config file.
+            self._error_stream.write(
+                'varname must be a value name, not a section or sub-section '
```
kyleknap: I think we should be including the value of varname as it is not quite obvious that varname is the positional argument name in the help command. How does something like this sound?
``'varname: %s must reference a value, not a section or sub-section' % value``

PR #2556, opened 2017-04-19
Output format changes for aws-cloudformation-package command

DiegoPomares:
Use JSON as default format for output of "aws cloudformation package" command, added an flag for YAML.

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -53,12 +53,18 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
     return {cfntag: value}
 
 
+def str_representer(dumper, node):
+    return ScalarNode(tag='tag:yaml.org,2002:str', value=node, style='\'')
+    
+
 def yaml_dump(dict_to_dump):
     """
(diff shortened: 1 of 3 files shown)


sanathkr:
You should default to YAML and provide a use-json override. YAML dump was the old behavior and you should retain it for the majority of customers that are happy with it

sanathkr:
Also need unit tests

DiegoPomares:
YAML is the default output format now. Updated a unit test to get coverage back to 100%.

awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -53,12 +53,18 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
     return {cfntag: value}
 
 
+def str_representer(dumper, node):
```
sanathkr: We don't need this, do we? customers having trouble with YAML integers can use JSON. I don't want to uglify YAML with quotes everywhere

sanathkr:
Looks good. Thanks for doing.

sanathkr:
@jamesls @kyleknap Want to review as well?

dstufft:
Looks good, thanks!

PR #2562, opened 2017-04-24
Use console wrappers instead of our own commands

### LINKED ISSUE #2372: ImportError on any CLI command on Windows
Hi there,

I just ran `pip install awscli` in a new Python 2.7 virtual environment on Windows.  It installed successfully, but any command I run gives output like this:

```
$ aws ecr get-login --region us-east-1
  File "...\.tox\tools\Scripts\aws.cmd", line 50, in <module>
    import awscli.clidriver
ImportError: No module named awscli.clidriver
```

The package in the virtual environment seems fine and I can run this command successfully (no `ImportError`):

```
$ python -c "from awscli.clidriver import main; main()" ecr get-login --region us-east-1
Unable to locate credentials. You can configure credentials by running "aws configure".
```

Any idea what's wrong?

I'm using this version:

```
$ pip freeze | grep awscli
awscli==1.11.36
```

dstufft:
Fixes #2372

I could not find a reason why these commands were done the way they were, they date back to the very beginning of awscli and there's no information for why it was done that way. Given #2372 I think that it is a saner idea to just use these script wrappers (which have been used by pip and other projects for a very long time). Not only should this resolve #2372 but I believe it will just provide a better experience all around on Windows (Windows hardcodes the ``.exe`` extension in several places).


lorengordon:
If this gets merged, feel free to close my silly PR making a slight modification to aws.cmd, https://github.com/aws/aws-cli/pull/2005.

Issue #2564, opened 2017-04-24
Cloudformation Package does not work with already uploaded templates in nested stacks


jrstarke:
I'm using Cloudformation Stacks and SubStacks to deploy large parts of our infrastructural components, in combination with `aws cloudformation package` to package up the SubStack resources for deployment. 

Recently, we've started pushing AWS WAF out with our CloudFront distributions. To this, the stack for these distributions has a `AWS::CloudFormation::Stack` resource, which deploys the AWS WAF Preconfigured Protections (https://aws.amazon.com/answers/security/aws-waf-security-automations/). These protections include a CloudFormation template available at https://s3.amazonaws.com/solutions-reference/aws-waf-security-automations/latest/aws-waf-security-automations.template for creating the WAF WebACL and the associated rules.

The problem is, when I run `cloudformation package` on my top level stack, I get the error 

`TemplateURL parameter of AppFirewall resource is invalid. It must be a S3 URL or path to CloudFormation template file. Actual: /Users/jstarke/<redacted>/https:/s3.amazonaws.com/solutions-reference/aws-waf-security-automations/latest/aws-waf-security-automations.template`

Looking into this specific command, it appears that it checks if the template_path in the `TemplateURL` is an s3 url (https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/artifact_exporter.py#L64), but it only considers urls starting with `s3://`, where as the Cloudformation Stack documentation specifically requires TemplateURLs for existing resources to start with `https://s3.amazonaws.com/` (http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-stack.html#cfn-cloudformation-stack-templateurl).

jrstarke:
I've come up with what I think is the simplest reproduction of this. It involves a top level stack with a substack that uses the AWS WAF Preconfigured protections (So Stack -> Child Stack -> Grandchild Stack). I've made a gist here: https://gist.github.com/jrstarke/ab60f3713675c69d9377fbd1b5017459

When you run these templates with the following (replacing `<bucket_to_upload_to>` with a bucket you have access to):
`aws cloudformation package --template-file template1.yaml --s3-bucket <bucket_to_upload_to>`

dstufft:
Confirmed, thanks for the reproduction steps.

matheusmaximo:
I know this is a closed issue but just a point here, the docs say:
https://s3.region.amazonaws.com/bucket-name/sam-template-object

[...]

PR #2566, opened 2017-04-25
Adds support for S3 HTTPS urls

jrstarke:
Fixes https://github.com/aws/aws-cli/issues/2564

This adds logic for supporting https://s3.amazonaws.com as a supported s3 URL.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -328,7 +328,7 @@ class CloudFormationStackResource(Resource):
 
         template_path = resource_dict.get(self.PROPERTY_NAME, None)
 
-        if template_path is None or is_s3_url(template_path):
+        if template_path is None or is_s3_url(template_path) or template_path.startswith("https://s3.amazonaws.com/"):
             # Nothing to do
             return
 
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -60,28 +60,47 @@ def parse_s3_url(url,
                  object_key_property="Key",
                  version_property=None):
 
-    if isinstance(url, six.string_types) \
-            and url.startswith("s3://"):
+    if isinstance(url, six.string_types):
+        if url.startswith("s3://"):
+            # Python < 2.7.10 don't parse query parameters from URI with custom
+            # scheme such as s3://blah/blah. As a workaround, remove scheme
+            # altogether to trigger the parser "s3://foo/bar?v=1" =>"//foo/bar?v=1"
+            parsed = urlparse.urlparse(url[3:])
+            query = urlparse.parse_qs(parsed.query)
 
-        # Python < 2.7.10 don't parse query parameters from URI with custom
-        # scheme such as s3://blah/blah. As a workaround, remove scheme
-        # altogether to trigger the parser "s3://foo/bar?v=1" =>"//foo/bar?v=1"
-        parsed = urlparse.urlparse(url[3:])
-        query = urlparse.parse_qs(parsed.query)
+            if parsed.netloc and parsed.path:
+                result = dict()
+                result[bucket_name_property] = parsed.netloc
+                result[object_key_property] = parsed.path.lstrip('/')
 
-        if parsed.netloc and parsed.path:
-            result = dict()
-            result[bucket_name_property] = parsed.netloc
-            result[object_key_property] = parsed.path.lstrip('/')
+                # If there is a query string that has a single versionId field,
+                # set the object version and return
+                if version_property is not None \
+                        and 'versionId' in query \
+                        and len(query['versionId']) == 1:
+                    result[version_property] = query['versionId'][0]
 
-            # If there is a query string that has a single versionId field,
-            # set the object version and return
-            if version_property is not None \
-                    and 'versionId' in query \
-                    and len(query['versionId']) == 1:
-                result[version_property] = query['versionId'][0]
+                return result
 
-            return result
+        elif url.startswith("https://s3.amazonaws.com/"):
+            parsed = urlparse.urlparse(url)
+            query = urlparse.parse_qs(parsed.query)
+
+            folders = parsed.path.lstrip('/').split('/')
+
+            if len(folders) >= 2:
+                result = dict()
+                result[bucket_name_property] = folders[0]
+                result[object_key_property] = '/'.join(folders[1:])
+
+                # If there is a query string that has a single versionId field,
+                # set the object version and return
+                if version_property is not None \
+                        and 'versionId' in query \
+                        and len(query['versionId']) == 1:
+                    result[version_property] = query['versionId'][0]
+
+                return result
 
     raise ValueError("URL given to the parse method is not a valid S3 url "
                      "{0}".format(url))
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -82,7 +82,7 @@ def parse_s3_url(url,
 
                 return result
 
-        elif url.startswith("https://s3.amazonaws.com/"):
+        elif url.startswith("https://s3."):
             parsed = urlparse.urlparse(url)
             query = urlparse.parse_qs(parsed.query)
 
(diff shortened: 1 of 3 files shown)


dstufft:
This looks pretty good to me, although I would prefer it if it included some additional tests to ensure that this continues to function in the future. You can look at the tests in ``tests/unit/customizations/cloudformation/test_artifact_exporter.py`` for an example of how the ``s3://`` URLs are currently being tested.

dstufft:
Confirmed that this solves #2564 as well.

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -60,28 +60,47 @@ def parse_s3_url(url,
...
+                # If there is a query string that has a single versionId field,
+                # set the object version and return
+                if version_property is not None \
+                        and 'versionId' in query \
+                        and len(query['versionId']) == 1:
+                    result[version_property] = query['versionId'][0]
+
+                return result
+
+        elif url.startswith("https://s3.amazonaws.com/"):
```
lorengordon: There are actually AWS environments where the domain name is not `amazonaws.com`. How about checking for `https://s3.` instead?

jrstarke: Thanks @lorengordon, could you help me understand the usecase? Is it for uses like `s3-us-east-1` type things, or something else?

jrstarke:
@dstufft: Thanks for the pointer! I was looking for some reasonable tests I could add, so this is a huge help. 

Looks like we could add some `https://s3` urls to the [valid list](https://github.com/Giftbit/aws-cli/blob/cloudformation-package/tests/unit/customizations/cloudformation/test_artifact_exporter.py#L24). 

One thing I'm curious about, `https://s3-eu-west-1.amazonaws.com/bucket/key` is included in the [invalid list](https://github.com/Giftbit/aws-cli/blob/cloudformation-package/tests/unit/customizations/cloudformation/test_artifact_exporter.py#L34). Which, if any of these would not be valid?

```
        "https://s3.amazonaws.com/foo/bar",
        "https://s3.amazonaws.com/foo/bar/baz/cat/dog",
        "https://s3.amazonaws.com/foo/bar?versionId=abc",
        "https://s3.amazonaws.com/foo/bar/baz?versionId=abc&versionId=123",
        "https://s3.amazonaws.com/foo/bar/baz?versionId=abc",
        "https://s3.amazonaws.com/www.amazon.com/foo/bar",
        "https://s3.amazonaws.com/my-new-bucket/foo/bar?a=1&a=2&a=3&b=1",
        "https://s3-us-east-1.amazonaws.com/foo/bar",
        "https://s3.mydomain.com/foo/bar",

```

awscli/customizations/cloudformation/artifact_exporter.py
lorengordon: No, it's quite literally AWS S3 services (run by AWS) that exist on other domain names. I'm simply suggesting to modify the test to be:

```
elif url.startswith("https://s3."):
```

So that the feature works in such domains.

Another approach would be to rely somehow on the endpoint lookup via botocore.

jrstarke: Thank you for the feedback. I've incorporated this change.

jrstarke: Do you happen to have an example of this that I could include in the `valid` strings for the tests?

lorengordon: You can derive them from [endpoints.json in botocore](https://github.com/boto/botocore/blob/develop/botocore/data/endpoints.json). And actually looks like the `s3-` is required for some regions (us-gov-west-1), which is unfortunate.

* https://s3-fips-us-gov-west-1.amazonaws.com
* https://s3-us-gov-west-1.amazonaws.com
* https://s3.amazonaws.com.cn

You can also check the aws docs:

* http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html
* http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#cnnorth_region
* http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

jrstarke: Is it possible for us to simplify this at all? Given that this is for CloudFormation specifically, we might be able to look at a subset of the s3 urls. I know for `AWS::CloudFormation::Stack` it specifically says that it must be in the format of `https://s3.amazonaws.com/...`. I'll have a look at the docs for the other exportable resources, and see if they have any restrictions on their s3 urls.

jrstarke: Hello @lorengordon,

Do we need to support all specific based s3 urls, such as `https://s3-eu-west-1.amazonaws.com/bucket/key` (see https://github.com/aws/aws-cli/pull/2566/files#diff-e5123393ba45e32e56b53c9f8ccd7c08R189), or only the exceptional ones like Beijing, and Gov Cloud?

lorengordon: @jrstarke Someone on the aws-cli team would probably need to make that call, not me! :D

If it were me, I might try to lean on botocore somehow to query whether the provided https url is a valid endpoint. I'm not sure botocore has a method to do that directly, or if it would involve actually querying the bucket.

lorengordon: Maybe botocore's [`client.head_bucket()`](https://botocore.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.head_bucket) method, or possibly [`botocore.stub`](https://botocore.readthedocs.io/en/latest/reference/stubber.html#botocore-stub)

jrstarke: @dstufft: Would you be willing to weigh in on this?

jamesls:
cc @sanathkr Can you take a look?

sanathkr:
Thanks for doing this. Your problem is because the `is_s3_url` check  at https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/artifact_exporter.py#L331 fails and the code following it assumes path is a local file. Instead of going to the depths of parsing the URL (which can be tricky & needs to handle s3-region.amazonaws style too), I think you can simply test if value is a http URL in addition to `is_s3_url` check.

If someone specifies a HTTP path there, be it S3 or not, package command has no business there.

jrstarke:
Thanks for pointing that out @sanathkr 

Given that the documentation specifically says `The template must be stored on an Amazon S3 bucket, so the URL must have the form: https://s3.amazonaws.com/.../TemplateName.template`, should this like instead be 

```python
 if template_path is None or template_path.startswith("https://s3.amazonaws.com/") :
``` 

Given this is in `CloudFormationStackResource`, it should be safe, and only effect Cloudformation stacks. If you think that would be reasonable, I can verify that it solves the reproduction.

Thanks 😄

sanathkr:
@jrstarke Would't you need s3-region-name.amazonaws.com if the bucket was created in any region but "US Standard"? I understand it is more thorough to validate s3 URLs, but I would rather not bake the logic into CLI. Just handle what the CLI can and skip the rest is my thinking.

jrstarke:
@sanathkr I'm not actually sure. Our resources are primarily based out of `us-west-2`, and if I recall correctly, when I tried our resources with `https://s3-us-west-2.amazonaws.com/key` for a stack, it would complain at cloudformation, but when I switched it to `https://s3.amazonaws.com/key` it seemed to work ok. IIRC, when you try to access a key using `https://s3.amazonaws.com/key` and it's in a different region, it does a 307 redirect, and cloudformation picks it up correctly.

If we do want to specifically allow it if it's in `https://s3-region-name.amazonaws.com` form, would we want to check regexes, or is there something already built into boto core that we could use?

sanathkr:
Alright. Let's then check for startswith and call it done

prokilogrammer:
@jrstarke Can you get this in?

jrstarke:
@prokilogrammer: Thanks for the bump.  I've been out of the office for the last few weeks (and go back tomorrow) but I think the latest changes do what this needs 😃

sanathkr:
@jrstarke Great, thanks for making the changes. Looks good to be merged. @jamesls can you merge?

sanathkr:
@JordonPhillips can you merge?

JordonPhillips:
Added a changelog entry, merging

Issue #2460, opened 2017-02-23
aws cloudformation deploy not maintain stack parameters


agutierrezdapda:
With latest version of awscli, cloudformation deploy not maintain stack parameters.
aws cloudformation deploy get template defaults instead of stack parameters.

With aws-cli/1.11.19 and botocore/1.4.83 are working correctly.

kyleknap:
Could you post debug logs of the command you are running? You could do so by appending ``--debug`` to the command you ran.

agutierrezdapda:
[output.txt](https://github.com/aws/aws-cli/files/804098/output.txt)

stealthycoin:
Similar issue opened #2466 closing that one in favor of this one.

bsideup:
@stealthycoin you closed both with a cycle reference between them :D

kadrach:
I'm seeing this issue as well. @agutierrezdapda did this work for you with a previous version? Do you know which version I can fall back to?

oleyka:
Still an issue in 1.11.79.

dstufft:
Fixed in https://github.com/aws/aws-cli/pull/2573.

jyrkiput:
I'm not sure if this was fixed, when using example from https://github.com/aws/aws-cli/issues/2466 I value from stack is overridden with default in version 1.11.83
```
aws --version
aws-cli/1.11.83 Python/2.7.12+ Linux/4.10.0-21-generic botocore/1.5.46
```
I created following template
```
{
    "AWSTemplateFormatVersion" : "2010-09-09",
    "Description"              : "Test stack",
    "Parameters"               : {

        "TestParam2" : {
            "Description" : "Test param 2",
            "Type"        : "String",
            "Default": "exampledefault"
        }
    },
    "Resources" : {
        "examplebucket" : {
            "Type" : "AWS::S3::Bucket",
            "Properties": {
                "BucketName": { "Ref" : "TestParam2" }
            }
        }
    }
}
```
Then created a new stack with
```
aws cloudformation create-stack --stack-name aws-cli-test-s3 --template-body file://test.json
```
As expected, now I have S3 Bucket 'exampledefault'.
Then I updated the stack with '--parameters ParameterKey=TestParam2,ParameterValue=exampleoverride',
```
aws cloudformation update-stack --stack-name aws-cli-test-s3 --template-body file://test.json --parameters ParameterKey=TestParam2,ParameterValue=exampleoverride
```
[...]

PR #2573, opened 2017-04-28
cloudformation deploy: Use parameter default value only on stack creation

sanathkr:
There are four state combinations:
YES Override + YES Stack: Use overriden value
YES Override + NO Stack: Use overriden value

NO Override + YES Stack: Set UsePreviousValue to True
  ^^ This commit fixes this case. Earlier we were letting CloudFormation pickup default value from template

NO Override + NO Stack: Skip parameter to force CloudFormation to use template default value. UsePreviousValue=True will fail here because stack does not have any previous state.


sanathkr:
Fixes #2460

PR #2578, opened 2017-05-02
Use python -m to test CLI

dstufft:


awscli/testutils.py
```diff
@@ -639,11 +606,8 @@ def aws(command, collect_memory=False, env_vars=None,
     """
     if platform.system() == 'Windows':
         command = _escape_quotes(command)
-    if 'AWS_TEST_COMMAND' in os.environ:
-        aws_command = os.environ['AWS_TEST_COMMAND']
```
kyleknap: Just checking but is this environment variable used in any of the tests? I did a quick grep and did not see anything so I think we should be good.

dstufft: I couldn't find it referenced anywhere, it's entirely possible but searching for obvious sub strings did not bring anything up.

Master4Real:
[Federal_Register_Meeting_Notice_for_April_1_2024_CDAB_Meeting.pdf](https://github.com/aws/aws-cli/files/14888227/Federal_Register_Meeting_Notice_for_April_1_2024_CDAB_Meeting.pdf)

PR #2581, opened 2017-05-05
Revert "Use console wrappers instead of our own commands"

kyleknap:
Reverts aws/aws-cli#2562

### DIFF
--- a/awscli/__init__.py
+++ b/awscli/__init__.py
@@ -19,16 +19,6 @@ import os
 
 __version__ = '1.11.82'
 
-
-#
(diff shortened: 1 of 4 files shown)


duncm:
Were you able to find out why `aws.cmd` was used instead of an entry point, or was there any other particular reason for reverting?

`aws.cmd` doesn't work nicely with [pipx](https://github.com/pipxproject/pipx). My workaround is to manually change PythonExe at the top of the file each time awscli updates:

```
set PythonExe="%USERPROFILE%\.local\pipx\venvs\awscli\Scripts\python.exe"
```

PR #2602, opened 2017-05-12
Adding models now supports unicode models.

stealthycoin:
Adding models in py2 with unicode characters in them didn't work.

cc @jamesls @JordonPhillips @dstufft @kyleknap


stealthycoin:
Do we even want this?

PR #2605, opened 2017-05-15
add role-arn and notification-arns as parameters for cloudformation d…

stan3:
…eploy

PR #2606, opened 2017-05-15
cloudformation deploy - exit nonzero when no stack change

stan3:
Seems like a better default and works better with scripted deploy.  But perhaps shouldn't change the default behaviour?

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -186,13 +186,17 @@ class DeployCommand(BasicCommand):
     def deploy(self, deployer, stack_name, template_str,
                parameters, capabilities, execute_changeset, role_arn,
                notification_arns):
-        result = deployer.create_and_wait_for_changeset(
-                stack_name=stack_name,
-                cfn_template=template_str,
-                parameter_values=parameters,
-                capabilities=capabilities,
-                role_arn=role_arn,
-                notification_arns=notification_arns)
+        try:
+            result = deployer.create_and_wait_for_changeset(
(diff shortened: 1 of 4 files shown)


sanathkr:
@stan3 You have two different features in the PR. What is the other one about?

stan3:
@sanathkr this PR is committed on top of https://github.com/aws/aws-cli/pull/2605 as the two would otherwise conflict

sanathkr:
This looks good. Let's merge it 👍

sanathkr:
@kyleknap can we merge this?

mcblair:
Very interested in this.

jolexa:
Add label: `ready-to-merge` 🙏

JimmyDahlqvist:
Much needed feature.

spezam:
Any chance to get this merged?

JordonPhillips:
This had two issues: there were no added tests, and changing the default behavior is a breaking change. I submitted a pr that adds a commit to address those issues (while keeping @stan3 's commit so he keeps credit).

georgeismike:
@JordonPhillips what's needed to get this merged now?

thirstydeveloper:
We need this. Bad.

nafiux:
+1, I don't feel safe wrapping the command with || true, it should fail with a real error.

An idea could be add a flag to the CLI to return a 0 exit code is any change isn't detected by Cloudformation, this way you won't break your API (backward compatibility with your CLI), something like --skipNoChangeError

jochemvankessel:
Hello? someone? This should have been merged a long time ago already, right...? Why it isn't?

PR #2642, opened 2017-06-05
Remove inaccurate comment from zsh auto complete file

a8:
When I was using zsh + aws-cli it was a quick way to add auto
complete to zsh via the bash emulation. At the time I intended
to write native zsh auto completion.

However, I never found the time to do it and lost the motivation.
Thus, removing the comment.

Signed-off-by: Frank Becker <fb@alien8.de>


JordonPhillips:
Looks good to me, but you've based your pr off of master rather than develop. If you could fix that up I can get this merged in.

a8:
Sorry.

Dropping this one in favour of [#2645](https://github.com/aws/aws-cli/pull/2645) which is based on branch develop.

Issue #2639, opened 2017-06-05
Code Deploy - Unhandled exception - ZIP does not support timestamps before 1980 


jonathanchrisp:
## Overview
When running a deployment via circle-ci we've recently been getting the following error when running the `create_application_revision` command:

```
Unhandled exception
ZIP does not support timestamps before 1980
```

I couldn't find an existing issue with the repo. We haven't changed any configuration recently. This appears to have only started happening since yesterday, it was the first time we had the error on our builds.

We're running the following versions: 

```
aws-cli/1.11.97 Python/2.7.6 Linux/3.13.0-48-generic botocore/1.5.60
```

If anyone is able to point us in the right direction that would be greatly appreciated.

mgibas:
Same error for:
`aws cloudformation package ...`

```
Uploading to a5902e46b3516ee3f44caf6251079b5f  1846 / 1846.0  (100.00%)
Unable to upload artifact ./../async-handlers/donation-created-handler referenced by CodeUri parameter of DonationCreatedHandlerFunction resource.
ZIP does not support timestamps before 1980
```
even after downgrading to `1.11.79` (that was working some time ago) gives the same error.

[full build log](https://travis-ci.org/gitcheese/gitcheese.api/builds/239429253)

JordonPhillips:
It sounds like you have some files with invalid timestamps. This could be indicative of a larger issue, so I would recommend fixing them. Changing your version of the CLI won't impact this because the the error is raised in python itself.

arsenio:
I'm noticing the exact same error message with CircleCI today, during the create_application_revision command:

```create_application_revision /tmp/codedeploy_applications.json /tmp/codedeploy_revisions.json

create_application_revision loaded: {"applications":[{"region":"us-west-2","application_root":"/","revision_location":{"s3Location":{"bucket":"grove-applications","key":"<BUCKET_KEY>"},"revisionType":"S3"},"deployment_group":"staging","application_name":"<appname>"}]}
Bundling <appname> from /home/ubuntu/<appname>
Unhandled exception
ZIP does not support timestamps before 1980

((create_application_revision "/tmp/codedeploy_applications.json" "/tmp/codedeploy_revisions.json")) returned exit code 1
```

arsenio:
There's also a [Bug Report](https://discuss.circleci.com/t/zip-does-not-support-timestamps-before-1980/13110) open on CircleCI's support forums about this.

jonathanchrisp:
[...]

PR #2646, opened 2017-06-07
Use writestr to zip files with mtime before 1980

DiegoPomares:
Related: https://github.com/aws/aws-cli/issues/2639

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -185,14 +185,13 @@ def make_zip(filename, source_root):
                         full_path, source_root)
                     
                     stat_info = os.stat(full_path)
-                    if stat_info.st_atime < time_1980 or stat_info.st_mtime < time_1980:
-                        try:
-                            os.utime(full_path, (time_1980, time_1980))
-                        except OSError:
-                            raise ValueError('Cannot modify file timestamps for {} '
-                            '(ZIP does not support timestamps before 1980)'.format(full_path))
-
-                    zf.write(full_path, relative_path)
+                    if stat_info.st_mtime < time_1980:
+                        with open(full_path, mode='rb') as file:
+                            data = file.read()
+                            zf.writestr(relative_path, data)
+                            
+                    else:
+                        zf.write(full_path, relative_path)
 
     return zipfile_name
 
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -173,6 +173,8 @@ def zip_folder(folder_path):
 def make_zip(filename, source_root):
     zipfile_name = "{0}.zip".format(filename)
     source_root = os.path.abspath(source_root)
+    time_1980 = 315532800
+    
     with open(zipfile_name, 'wb') as f:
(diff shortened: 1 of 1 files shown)


eric-tucker:
I don't feel this should be necessary.  Uglify-JS needs to release a fix for

https://github.com/mishoo/UglifyJS2/issues/2054

JordonPhillips:
I'm -1 on modifying files without explicit consent. Maybe if there is a command line flag to opt in to this behavior, but even then I'm still hesitant to deliberately upload obviously corrupted files. With `cp` for instance we will skip files that have invalid timestamps.

stealthycoin:
Agreed with @eric-tucker and @JordonPhillips I would rather the user be alerted their file is corrupted rather than paint over it.

dstufft:
I also agree, we shouldn't be modifying someone's files, even one as minor as changing the mtime,  without their knowledge or consent.

jamesls:
(Just catching up on this now).  I suppose I have a basic question.  Why is the CLI `zip` command able to handle timestamps prior to 1980?  And is there any way we can do the same thing in python?  I'm also against modifying mtime's without the user opting in, but I'd like to know if there's a way we can just preserve timestamps to begin with.

```
$ ls -la
total 0
-rw-r--r--   1 james              0 Dec 31  1969 a
$ zip -r a.zip ./
  adding: a (stored 0%)
$ unzip -l a.zip
Archive:  a.zip
  Length     Date   Time    Name
 --------    ----   ----    ----
        0  12-31-69 16:00   a
 --------                   -------
        0                   1 file
$ rm a
$ ls -la
total 8
-rw-r--r--   1 james            152 Jun 12 10:35 a.zip
$ unzip a.zip
Archive:  a.zip
 extracting: a
$ ls -la
-rw-r--r--   1 james              0 Dec 31  1969 a
-rw-r--r--   1 james            152 Jun 12 10:35 a.zip
```

dstufft:
Well, zip files don't natively support it, the spec explicitly says the modification time is indicated by years since 1980. Zip files do support extra metadata being stored, so maybe the zip utility is storing the real mtime as an extra field?

DiegoPomares:
Seems like I failed to realize that modifying the timestamp was a big deal. I rewrote the whole thing, instead of modifying the timestamp in the source file, I switched to writestr [1] which adds the file to the zip with the current time.

The only caveat is that the whole file is read into memory before adding it to the zip, but I don't think that should be a big deal under most circumstances.

[1] https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.writestr

DiegoPomares:
Notice this patch is only for 'aws cloudformation package', but the zipfile module is used in several other places, like lambda and codedeploy. If you consider the writestr approach is good enough, perhaps it is best to write a wrapper module for zipfile to have consistency. I can code the whole thing as soon as you agree on a solution.

dstufft:
Does this have the effect that the mtime is set to the current time in the generated zip file if it was previously set to before 1980?

DiegoPomares:
dstufft@ that's right, it will only override the mtime for the file inside the zip, not the source file.

tim-finnigan:
There hasn't been any activity on this PR for a few years, and the initial maintainer response was against these changes. The original issue referenced (https://github.com/aws/aws-cli/issues/2639) seems to suggest the problem was caused by timestamp issues introduced by a third-party library that have since been fixed. I think this PR should be closed but can be revisited if necessary.

PR #2659, opened 2017-06-12
don't set UsePreviousValue if parameters doesn't exist (fixes #2599)

### LINKED ISSUE #2599: cloudformation deploy fails if new parameters are not overridden even if they have default value
After adding new parameters to the template, it has to be deployed with --parameters-override for all the new parameters even if they have default value, otherwise deploy fails with: 

> An error occurred (ValidationError) when calling the CreateChangeSet operation: Invalid input for parameter key Param2. Cannot specify usePreviousValue as true for a parameter key not in the previous template

To reproduce, deploy simple template:
```
AWSTemplateFormatVersion : '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: A hello world application.
Parameters:
  Param1:
    Default: 123
    Type: String
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      Handler: index.handler
      Runtime: nodejs4.3
      CodeUri: ./hello.zip
```

add some parameter to it:

```
...
Parameters:
  Param1:
    Default: 123
    Type: String
  Param2:
    Default: 123
    Type: String
...
```

Try to deploy again, it fails with aforementioned error.

stan3:
fixes #2599

### DIFF
--- a/awscli/customizations/cloudformation/deployer.py
+++ b/awscli/customizations/cloudformation/deployer.py
@@ -88,7 +88,6 @@ class Deployer(object):
         # Each changeset will get a unique name based on time
         changeset_name = self.changeset_prefix + str(int(time.time()))
 
-        changeset_type = "UPDATE"
         if not self.has_stack(stack_name):
             changeset_type = "CREATE"
             # When creating a new stack, UsePreviousValue=True is invalid.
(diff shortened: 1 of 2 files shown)


JordonPhillips:
Could you have a look at this @sanathkr ?

kienpham2000:
This is awesome, can we please review and merge? 

I'm having this exact issue as of version:
```
bash-3.2$ aws --version
aws-cli/1.11.148 Python/2.7.13 Darwin/16.7.0 botocore/1.7.6```

PR #2065, opened 2016-07-12
Bump colorama version max to 0.3.7

JordonPhillips:
This uses the `strip` argument to preserve the previous behavior
of colorizing when output is redirected.

Fixes #2043
Resolves #2037
Resolves #1742

cc @kyleknap @jamesls

PR #2666, opened 2017-06-14
- Support use of colorama up to 0.3.8

rjschwei:
+ colorama bugfix release 0.3.8 is available and contains no incompatible
    changes. There is no need to restrict use to less or equal 0.3.7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ def find_version(*file_paths):
 
 
 requires = ['botocore==1.7.7',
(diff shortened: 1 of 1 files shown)


JordonPhillips:
For some historical context, there had been a breaking change with colorama. This was fixed in #2065

rjschwei:
Superseded by https://github.com/aws/aws-cli/pull/2892/files

Issue #2494, opened 2017-03-16
Cannot delete s3 object with custom sse encryption


omencl:
When using command `aws --profile some_profile s3 cp /some_file s3://bucket/some_file --sse-c AES256 --sse-c-key XXX` file is successfully uploaded to server. Also copy back from bucket to local server works.
But this command `aws --profile some_profile s3 rm s3://bucket/some_file` returns
>  fatal error: An error occurred (400) when calling the HeadObject operation: Bad Request

Same object uploaded without sse-c can be deleted normally.

stealthycoin:
Can you add `--debug` to the command and give us the logs here?

omencl:
Here it is. I replace original bucket name with some foo name in logs.
[log.TXT](https://github.com/aws/aws-cli/files/850947/log.TXT)

Thank you!

omencl:
Note: now I am using s3api delete-object as a workaround

neelay-shah:
I am having the exact same issue.

stealthycoin:
Interesting, can yuo give the working debug logs for `s3api delete-object` and also @ mention me when you post them?

lvnyan05:
I had a same issue, too ! and I could avoid it as follows. 
```
$ aws s3 cp file s3://backet/test/ --sse-c AES256 --sse-c-key xxx
upload: ./file to s3://backet/test

$ aws s3 rm s3://backet/test/file
fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden

$ aws s3 rm s3://backet/test/ --recursive
delete: s3://backet/test/file
```
Best Regards.

neelay-shah:
Hi @stealthycoin 
Sorry for the delay, find the log in attachment. 
[aws-s3-rm-log.txt](https://github.com/aws/aws-cli/files/1022833/aws-s3-rm-log.txt)

stealthycoin:
Yep so this definitely looks like a bug to me.

PR #2667, opened 2017-06-14
Remove unnecessary HeadObject for single object rm

kyleknap:
This causes issues where the object is encrypted with SSE-C as you will get errors since HeadObject requires SSE-C key be provided.

Fixes https://github.com/aws/aws-cli/issues/2494

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -151,8 +151,8 @@ class FileGenerator(object):
 
     def _inject_extra_information(self, file_stat_kwargs, extra_information):
         src_type = file_stat_kwargs['src_type']
-        file_stat_kwargs['size'] = extra_information['Size']
-        file_stat_kwargs['last_update'] = extra_information['LastModified']
+        file_stat_kwargs['size'] = extra_information.get('Size')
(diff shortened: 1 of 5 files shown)


awscli/customizations/s3/filegenerator.py
```diff
@@ -151,8 +151,8 @@ def call(self, files):
 
     def _inject_extra_information(self, file_stat_kwargs, extra_information):
         src_type = file_stat_kwargs['src_type']
-        file_stat_kwargs['size'] = extra_information['Size']
-        file_stat_kwargs['last_update'] = extra_information['LastModified']
+        file_stat_kwargs['size'] = extra_information.get('Size')
```
jamesls: Is there a rationale for inserting the keys with `None` values vs. omitting them?

kyleknap: There is none. It does not matter if you provide `None` or omit them. No matter what one you chose, the end `size`  and `last_modified_time` properties on the `FileStat` object will be `None`

jamesls: Assuming that consuming code of `FileStat` can gracefully handle when these values are `None`, then this looks good to me.

PR #2671, opened 2017-06-16
Reduce some timing issues in integration tests

JordonPhillips:
With a sigv4 default the test in this case is completing much
faster. By the time we go to send the error signal to the socket
the copy may have already completed. This increases the size of
the object being transferred to prevent that from happening.

Additionally this makes the `key_exists` helper default to using
a waiter rather than relying on a single call to head object.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -819,6 +819,14 @@ class BaseS3CLICommand(unittest.TestCase):
         except (ClientError, WaiterError):
             return False
 
+    def key_not_exists(self, bucket_name, key_name, min_successes=3):
+        try:
+            self.wait_until_key_not_exists(
+                    bucket_name, key_name, min_successes=min_successes)
+            return True
+        except (ClientError, WaiterError):
+            return False
+
     def list_buckets(self):
         response = self.client.list_buckets()
         return response['Buckets']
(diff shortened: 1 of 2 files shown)


awscli/testutils.py
```diff
@@ -810,12 +811,12 @@ def get_key_contents(self, bucket_name, key_name):
         response = client.get_object(Bucket=bucket_name, Key=key_name)
         return response['Body'].read().decode('utf-8')
 
-    def key_exists(self, bucket_name, key_name):
-        client = self.create_client_for_bucket(bucket_name)
+    def key_exists(self, bucket_name, key_name, min_successes=3):
```
kyleknap: I'm a little hesitant to use the waiter for `key_exists()` because we use this to assert it does not exist as well, which would mean we would definitely have to wait 100 seconds for each time we are trying to assert the key does not exist.

JordonPhillips: We could have a `key_not_exists` and I can go replace instance of `key_exists` where necessary.

kyleknap: Yeah we talked offline about this. Sounds good to me.

JordonPhillips:
Cool. I'm currently running the full suite of tests across versions to make sure it's all good. I'll merge after that.

PR #2674, opened 2017-06-21
UTF-8 is default file encoding to read

sekido:
## My Environment

* Windows 7 Pro, Japanese (CP932)
* git bash 2.13.0
* AWS CLI via pip (aws-cli/1.11.109 Python/3.6.1 Windows/7 botocore/1.5.72)

## Problem

`aws cloudformation create-stack --template-body 'file://template.yml' --debug` raises an error.
template.yml is UTF-8 encoding.

```
  File "...\python36\lib\site-packages\awscli\paramfile.py", line 120, in get_file
    return f.read()
UnicodeDecodeError: 'cp932' codec can't decode byte 0x83 in position 68: illegal multibyte sequence
```

`get_file` calls `compat_open`, and `compat_open` opens file with platform dependent encoding, 'cp932' in Japanese environment, although the file is UTF-8 encoding.

```
$ python -i
Python 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 18:41:36) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
'cp932'
```

## How to fix

1. UTF-8 is default file encoding to read in `get_file`
2. Try to read the file with platform default encoding again if the file cannot read with UTF-8 encoding
3. Clear encoding parameter in `compat_open` because `open` doesn't receive "encoding" parameter in binary mode

### DIFF
--- a/tests/unit/test_paramfile.py
+++ b/tests/unit/test_paramfile.py
@@ -12,12 +12,13 @@
 # language governing permissions and limitations under the License.
 import platform
 
+import locale
 import mock
 from awscli.compat import six
 from awscli.testutils import unittest, FileCreator
 from awscli.testutils import skip_if_windows
 
-from awscli.paramfile import get_paramfile, ResourceLoadingError
+from awscli.paramfile import get_paramfile, get_file, ResourceLoadingError
 
 
 class TestParamFile(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tupans:
Grt file doesnt work

tests/unit/test_paramfile.py
```diff
@@ -62,6 +63,22 @@ def test_no_match_uris_returns_none(self):
...
+    def test_get_file_utf8_text(self):
+        contents = b'\xe9\x96\xa2\xe6\x88\xb8\xe3\x81\xa6\xe3\x81\x99\xe3\x81\xa8'
+        filename = self.files.create_file('foo', contents, mode='wb')
+        data = get_file('', filename, 'r')
+        self.assertEqual(data, u'\u95a2\u6238\u3066\u3059\u3068')
+        self.assertIsInstance(data, six.string_types)
+
+    @unittest.skipIf(locale.getpreferredencoding() != 'cp932',
+            'encoding fallback test with cp932 system')
+    def test_get_file_fallback_platformdefault_in_cp932(self):
```
JordonPhillips: This test will almost never be run. I would prefer to patch out the preferred encoding so that we can have this test be run every time the test suite is run.

sekido: Thank you for comment.  I fixed this testcase to run any environment.  Please review again.

joguSD:
I think this is reasonable. @JordonPhillips Do the updated tests look alright?

sekido:
Please let me know what is needed to merge.
This PR helps the internationalized environment people.

joguSD:
I think your PR is in a good spot for what it is attempting to do. We're concerned about the possibility of mojibake if we assume UTF-8 first. It's possible UTF-8 is the wrong encoding (but still decodes) and there would be no way for a user to opt out of this currently. I think we might need to reconsider the approach, or at the very least make this configurable in some way.

Also worth looking into is the fact that Python 3.7 will have a UTF-8 mode that can force UTF-8 encoding:
https://www.python.org/dev/peps/pep-0540/
https://github.com/python/cpython/blob/master/Lib/_bootlocale.py#L10-L14

kdaily:
Hi @sekido,

This has been incorporated into the AWS CLI v2 and is configurable using the environment variable `AWS_CLI_FILE_ENCODING`.

https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-encodingenvvar

PR #2677, opened 2017-06-21
Make the ctrl-c download test more robust

kyleknap:
There were some cases where the rc returned was 0 as the CLI process started or finished when the signal interrupt was sent. Also increased the sleep by a second to make sure we have enough time to hit the actual download process.


kyleknap:
Ran this integ test on a few different environments and all of them passed. Merging.

PR #2686, opened 2017-06-28
Reduce `cloudformation deploy` command polling delay

sanathkr:
Customers have asked for a faster deployment time.
See awslabs/serverless-application-model#125. This fix should
at least make the command complete as soon as cloudformation completes.

### DIFF
--- a/.changes/next-release/enhancement-Cloudformation-90843.json
+++ b/.changes/next-release/enhancement-Cloudformation-90843.json
@@ -1,0 +1,5 @@
+{
+  "category": "Cloudformation",
+  "type": "enhancement",
+  "description": "Reduce polling delay for `cloudformation deploy`."
+}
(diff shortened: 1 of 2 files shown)


sanathkr:
@JordonPhillips can I get a review on this? It might be a small change but could give good wins.

michaellieberherrr:
@sanathkr @JordonPhillips : This fix introduces an other issue, you run into "WaiterError: Waiter StackCreateComplete failed: Max attempts exceeded" if you have a template with for instance CloudFront resources (takes 10min+). I think the fix would be to increase the max retry attempts `waiter.config.max_attempts = 720` too, so that the max wait time will still be 60 mins in accordance to https://github.com/boto/botocore/blob/develop/botocore/data/cloudformation/2010-05-15/waiters-2.json#L24

sanathkr:
@michaellieberherrr good point. We should change max attempts as well. Can you send a PR for it?

Seblat5ch:
I have submitted the pull request before seeing this discussion. However rather than increasing the max attempts we could just tweak the polling interval. https://github.com/aws/aws-cli/pull/2761

sanathkr:
Nah, I would rather increase the max attempts. Could you make that change, instead? It optimizes for both long running tasks and quick tasks

Issue #2657, opened 2017-06-11
Botocore endpoint timeout not the same as the lambda timeout


lucasdf:
**Issue:**

I have an AWS Lambda function that is expected to run for a few minutes and it has a timeout of 4 minutes. I noticed that when I invoke this function using the CLI the function is always invoked multiple times, basically once every one minute. 

By using the "--debug" flag I noticed that botocore is setting the timeout of the endpoint as 60 seconds. So this is the cause of the multiple invocations: since the CLI reaches the timeout after one minute it sends a new request every minute, despite the fact that the lambda function's own timeout is higher than this value and the first invocation is still being processed.

**Questions:**
1. What is the expected behaviour of botocore/aws-cli: is it to use 60 seconds as timeout no matter what the lambda function timeout is, or is the expected behaviour to get the timeout value from the function definition? It seems to me that using the same timeout as the function should be the expected behaviour. If indeed it is then there seems to be something wrong when setting this value.
2. Is there a way to set the timeout value while invoking `aws lambda invoke`? `aws lambda invoke help` does not display any information regarding this.

**Commands:**
I am invoking the function using the following command:
`aws lambda invoke --function-name FUNCTION_NAME --payload 'JSON_PAYLOAD' output.json`
The debug flag shows the following message:
`MainThread - botocore.endpoint - DEBUG - Setting lambda timeout as (60, 60)`

**Versions:**
`aws-cli/1.11.93 Python/2.7.12 Linux/4.4.0-79-generic botocore/1.5.56`
`My lambda function is using Python 3.6`

**Reproducing the issue:**
It should be possible to reproduce it by creating a lambda function that just sleeps for more than 60 seconds and invoking it by using the CLI. 

**tl;dr:**
My function is being invoked multiple times because botocore is using 60 seconds as the timeout instead of using the four minutes that is defined in my lambda function. The expected behaviour should be to use the same timeout used by the lambda function. At least there should be a way for me to define the timeout value as a parameter when using `aws lambda invoke`.

lucasdf:
Full log: the function is invoked at 22:32:13. At 22:33:14 the timeout occur and the function is invoked again.

lucas@lucas-ubuntu:~/projects/app-shogun/app_shogun_lambda$ aws lambda invoke --debug --function-name test --payload '{"test_sleep": "true", "test_sleep_val": 80 }' output.json
[...]

PR #2689, opened 2017-06-29
Quote arguments in aliases

### LINKED ISSUE #2653: Alias does not pass quoted parameters correctly to the external command (shell)
Hi,

When using a shell function in aliases the parameters are not quoted correctly.

Here is a simple alias for testing (~/.aws/cli/alias):
```
[toplevel]

test-quotes =
  !f() {
    for i in "$@"; do
        echo "$i"
    done
  }; f
```

Testing the alias with `aws test-quotes foo bar "foo bar"`:

Desired output:
```
foo
bar
foo bar
```

Actual output:
```
foo
bar
foo
bar
```

When debugging the `ExternalAliasCommand`'s `__call__` method the `command_components` variable contains `['f() {\nfor i in "$@"; do\necho "$i"\ndone\n}; f', 'foo', 'bar', 'foo bar']` so the quoted parameter with a space is parsed correctly on input.

However when making the actual function call the parameter is not quoted correctly: `Using external alias 'test-quotes' with value: '!f() {\nfor i in "$@"; do\necho "$i"\ndone\n}; f' to run: 'f() {\nfor i in "$@"; do\necho "$i"\ndone\n}; f foo bar foo bar'`

One fix could be to use `quote` function from https://pypi.python.org/pypi/shellescape (backport of Python `shlex.quote` for Python versions 2.x & < 3.3) for `args` to get arguments escaped correctly but this creates a new dependency. (something like `command_components.extend(map(quote, args))`)

Version info: `CLI version: aws-cli/1.11.101 Python/2.7.13 Darwin/16.5.0 botocore/1.5.64`

JordonPhillips:
This quotes arguments passed in to aliases.

I've run the tests for this on Windows in the following shells: git bash, powershell, cmd, and cygwin.

Fixes #2653 
Also fixes https://github.com/aws/aws-cli/issues/2657#issuecomment-308846660

cc @jamesls @dstufft @kyleknap @stealthycoin

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -179,3 +179,81 @@ def compat_input(prompt):
     sys.stdout.write(prompt)
     sys.stdout.flush()
     return raw_input()
+
+
+def compat_shell_quote(s, platform=None):
+    """Return a shell-escaped version of the string *s*
+
+    Unfortunately `shlex.quote` doesn't support Windows, so this method
+    provides that functionality.
+    """
+    if platform is None:
+        platform = sys.platform
+
+    if platform == "win32":
+        return _windows_shell_quote(s)
+    else:
+        return shlex_quote(s)
+
+
+def _windows_shell_quote(s):
+    """Return a Windows shell-escaped version of the string *s*
+
+    Windows has potentially bizarre rules depending on where you look. When 
+    spawning a process via the Windows C runtime the rules are as follows:
+
+    https://docs.microsoft.com/en-us/cpp/cpp/parsing-cpp-command-line-arguments
+
+    To summarize the relevant bits:
+
+    * Only space and tab are valid delimiters
+    * Double quotes are the only valid quotes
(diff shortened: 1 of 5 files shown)


awscli/compat.py
```diff
@@ -179,3 +179,81 @@ def compat_input(prompt):
...
+    :param s: A string to escape
+    :return: An escaped string
+    """
+    if not s:
+        return '""'
+
+    buff = []
+    num_backspaces = 0
+    for character in s:
+        if character == '\\':
```
dstufft: One tiny nitpick (that you can feel free to ignore), but I think I would prefer this to be ``r'\'`` (here and throughout) if only because it doesn't require mentally processing the escape to parse what this code is doing.

JordonPhillips: I would agree, but unfortunately `r'\'` isn't a valid string since the backslash can still escape quotes in raw strings. Really frustrating because it makes the tests hard to read.

PR #2700, opened 2017-07-06
Improve the pagination help

joshbean:
This is my attempt to clarify the help for the AWS CLI's standardized pagination options.  The way to use these options is well documented in the AWS CLI User Guide, but the inline help didn't help me understand the difference between max-items and page-size.

### DIFF
--- a/awscli/customizations/paginate.py
+++ b/awscli/customizations/paginate.py
@@ -39,19 +39,37 @@ logger = logging.getLogger(__name__)
 STARTING_TOKEN_HELP = """
 <p>A token to specify where to start paginating.  This is the
 <code>NextToken</code> from a previously truncated response.</p>
+<p>For more information, see <a
+href="https://docs.aws.amazon.com/cli/latest/userguide/pagination.html"
+>Pagination</a> in the <i>AWS Command Line Interface User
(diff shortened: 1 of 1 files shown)


awscli/customizations/paginate.py
```diff
@@ -39,19 +39,37 @@
 STARTING_TOKEN_HELP = """
 <p>A token to specify where to start paginating.  This is the
 <code>NextToken</code> from a previously truncated response.</p>
+<p>For more information, see <a
```
JordonPhillips: I'm not sure I want to link to this page as it is a bit out of date. If it were updated I may be inclined, but I would prefer that it read "for detailed examples".

joshbean: I can make that change. How about "For usage examples, see ...". Does that work?

JordonPhillips: Sure, that works.

joshbean: Done. I just pushed a new commit with that change. Thanks for the fast review!

joshbean:
Bump. I'd love to get a second set of eyes on this. It would be really great to get this clarification into the official documentation.

PR #2705, opened 2017-07-12
Fixes ecr get-authorization-token example command

maazghani:
using uppercase `-D` for base64 as mentioned in documentation, single quotes around `--query` input


JordonPhillips:
Looks good, thanks for the PR!

PR #2708, opened 2017-07-12
Update zsh auto-complete docs and narrow `_bash_complete()` scope

indrajitr:
Changes:
- Make this auto-loadable when present in zsh `$fpath`
- Update docs suggesting alternate way of including completion using 'fpath'
  convention
- Narrow down the scope of `_bash_complete()` override, this is not necessary
  for zsh 5.0.3+, so override conditionally

_Edit: Updated to accommodate the fine set of feedback from @zmwangx that was provided [here](https://github.com/Homebrew/homebrew-core/pull/15727#issuecomment-317857731)._

### DIFF
--- a/bin/aws_zsh_completer.sh
+++ b/bin/aws_zsh_completer.sh
@@ -1,60 +1,79 @@
+#compdef aws
+
 # Source this file to activate auto completion for zsh using the bash
-# compatibility helper.  Make sure to run `compinit` before, which should be
-# given usually.
+# compatibility helper.
 #
-# % source /path/to/zsh_complete.sh
+# This can be done in either of the following ways:
 #
-# Typically that would be called somewhere in your .zshrc.
+#   - By explicitly by sourcing the script, typically somewhere in your .zshrc
+#     like so:
+#         % source /path/to/aws_zsh_completer.sh
+#
+#   - By making it available as '_aws' in your zsh '$fpath' like so:
+#         % ls -s /path/to/aws_zsh_completer.sh /usr/local/share/zsh/site-functions
+#     (assuming '/usr/local/share/zsh/site-functions' is already in '$fpath').
+#
+# Make sure to run `compinit` before sourcing this, which is usually a given.
 #
 # Note, the overwrite of _bash_complete() is to export COMP_LINE and COMP_POINT
 # That is only required for zsh <= edab1d3dbe61da7efe5f1ac0e40444b2ec9b9570
+# zsh version 5.0.3 onwards has this fully integrated. zsh relases prior to
+# that version do not export the required env variables!
 #
-# https://github.com/zsh-users/zsh/commit/edab1d3dbe61da7efe5f1ac0e40444b2ec9b9570
+# See: https://github.com/zsh-users/zsh/commit/edab1d3dbe61da7efe5f1ac0e40444b2ec9b9570
 #
 # zsh releases prior to that version do not export the required env variables!
 
-autoload -Uz bashcompinit
-bashcompinit -i
+# Load and initialize the completion system ignoring insecure directories.
+autoload -Uz bashcompinit && bashcompinit -i
 
-_bash_complete() {
-  local ret=1
-  local -a suf matches
-  local -x COMP_POINT COMP_CWORD
-  local -a COMP_WORDS COMPREPLY BASH_VERSINFO
-  local -x COMP_LINE="$words"
-  local -A savejobstates savejobtexts
+# Overwrite _bash_complete() for zsh versions less than 5.0.3 only
+autoload -Uz is-at-least
+if ! is-at-least 5.0.3; then
+  _bash_complete() {
+    local ret=1
+    local -a suf matches
+    local -x COMP_POINT COMP_CWORD
+    local -a COMP_WORDS COMPREPLY BASH_VERSINFO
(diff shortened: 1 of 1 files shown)


ilovezfs:
@joguSD @JordonPhillips any thoughts here?

ilovezfs:
@dstufft @stealthycoin any chance of getting this merged?

petemounce:
@dstufft @stealthycoin any chance of getting this merged, or a steer about what might be outstanding?

tim-finnigan:
Thanks for creating this PR and for your patience. Our team recently put out a [contribution guide](https://github.com/aws/aws-cli/blob/v2/proposals/contribution-guide.md) detailing improvements to the contribution process. I have created issue https://github.com/aws/aws-cli/issues/7851 for this PR. Some further discussion is needed regarding use cases and implementation approaches, and we can use that issue to track the discussion going forward.

indrora:
It's been a bit since I dug into the zsh completion logic but, it looks like this should improve completion performance since we're no longer using older fallback logic from bash on modern versions of zsh?

indrajitr:
Updated to resolve merge conflict with `develop` branch.

indrora (approved):
LGTM

indrora (approved):

bin/aws_zsh_completer.sh
```diff
@@ -1,60 +1,79 @@
...
-}
+    return ret
+  }
+fi
 
 complete -C aws_completer aws
+
+# Bind `aws` completion eagerly to `_bash_complete -C aws_completer` and
+# suppress stderr to avoid 'can only be called from completion function' warning
+_complete aws # 2> /dev/null
```
yan12125, line 79:
```suggestion
_complete aws 2> /dev/null
```
An extra change is needed to actually get rid of 'can only be called from completion function' warning

yan12125 (commented):
Thank you very much for the update! I cannot approve pull requests, though.

kyleknap:
@indrajitr Thanks for the PR. All of the references were helpful in providing context on trying to understand the rationale for the change. To make sure that I completely understand the rationale for the change, the change is mainly so a user can copy the `aws_zsh_completer.sh` file to wherever `zsh` looks for completions and auto-load the completion (i.e. does not have to run `source` on the script)? And specifically this is helpful for downstream distros (e.g. Homebrew) that install everything for the user related to a library (i.e. they will copy the completion files to the appropriate place on the system so that the completers are auto-loaded for the installed library)?

If the answers are yes to both of these questions, I'm open to the auto-loading change as it would help bring parity with the [`aws_bash_completer`](https://github.com/aws/aws-cli/blob/develop/bin/aws_bash_completer) that allows distros to just install that file to the appropriate location on the system to get auto-completion.

Another approach that I'm wondering if we should explore is just creating a new `aws_zsh_completer` file that is specifically to be installed to auto-enable zsh completers (assuming that is an established pattern) similar to the `aws_bash_completer`? Furthermore, this new file would follow all of the best practices for zsh completion files and not have any of <5.0.3 compatibility shims. The main reason that I'm suggesting exploring this is that when in 2013 the original `aws_zsh_completer.sh` script it was introduced, it was mainly to deal with the compatibility issues of zsh with our bash completer at that time. However, zsh has improved to no longer need that compatibility shim, and since, we have been recommending to [users to not even use the `aws_zsh_completer.sh` script](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html). That being said we cannot remove the `aws_zsh_completer.sh` script and want to be careful in making larger updates to it as users are still using it.

Thoughts?

tim-finnigan:
I'm going to mark this PR as Draft for now per the CLI [Contribution Guide](https://github.com/kdaily/aws-cli/blob/contrib-guide-proposal/proposals/contribution-guide.md) proposal submitted last year. Thanks again for your patience here as we're revisiting older PRs. 

As mentioned in the [comment](https://github.com/aws/aws-cli/pull/2708#issuecomment-1634941076) above some additional feedback and research is required before proceeding. We may also want to open an issue to track the feature request suggested above of creating a new `aws_zsh_completer` file.

yan12125:
> the change is mainly so a user can copy the `aws_zsh_completer.sh` file to wherever `zsh` looks for completions and auto-load the completion (i.e. does not have to run source on the script)?

Although I'm not the author and I cannot represent their intention, this is correct from what I understand.

> specifically this is helpful for downstream distros (e.g. Homebrew) that install everything for the user related to a library (i.e. they will copy the completion files to the appropriate place on the system so that the completers are auto-loaded for the installed library)?

Yes. For example, I've applied this patch to [aws-cli-v2](https://archlinux.org/packages/extra/any/aws-cli-v2/) package on Arch Linux, so Arch Linux users have auto-loaded completions.

tim-finnigan:
Checking in — after discussing this PR further with the team, we have decided to close it for the reasons mentioned in [this earlier comment](https://github.com/aws/aws-cli/pull/2708#issuecomment-1634941076). If anyone is still interested in this feature request, we recommend creating a new issue [using this template](https://github.com/aws/aws-cli/issues/new?assignees=&labels=feature-request%2Cneeds-triage&projects=&template=feature-request.yml&title=%28short+issue+description%29) outlining your use case and proposed solution. Further discussion and information gathering is needed prior to deciding upon potential implementations and paths forward here.

PR #2710, opened 2017-07-12
Add rotate-access-key command

loopingz:
To simplify rotating AccessKey to respect the 90 days principal i added a command to rotate the access key.

### DIFF
--- a/awscli/examples/configure/rotate-access-keys/_examples.rst
+++ b/awscli/examples/configure/rotate-access-keys/_examples.rst
@@ -24,7 +24,7 @@ will produce the following config file::
 
 Then the following commands::
 
-    $ aws --profile=test configure rotate-access-key
+    $ aws --profile=other configure rotate-access-key
 
 will produce the following config file::
 
--- a/awscli/examples/configure/rotate-access-keys/_examples.rst
+++ b/awscli/examples/configure/rotate-access-keys/_examples.rst
@@ -1,0 +1,37 @@
+Given an existing config file::
+
+    [default]
+    aws_access_key_id = default_access_key
+    aws_secret_access_key = default_secret_key
+
+    [other]
+    aws_access_key_id = other_access_key
+    aws_secret_access_key = other_secret_key
+
+The following commands::
+
+    $ aws configure rotate-access-key
+
+will produce the following config file::
+
+    [default]
+    aws_access_key_id = renew_access_key
+    aws_secret_access_key = renew_secret_key
+
+    [other]
+    aws_access_key_id = other_access_key
+    aws_secret_access_key = other_secret_key
+
+Then the following commands::
+
+    $ aws --profile=test configure rotate-access-key
+
+will produce the following config file::
+
+    [default]
+    aws_access_key_id = renew_access_key
+    aws_secret_access_key = renew_secret_key
+
+    [other]
+    aws_access_key_id = renew_other_access_key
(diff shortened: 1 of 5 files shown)


JordonPhillips:
Thanks for the contribution, I really like the idea! I'll have a look when I get some time.

loopingz:
Thanks i really hope it will make it to the next release, I'm trying to push the 90 days expiration on IAM Keys but without the tooling it will be harder to push

loopingz:
Any updates on this one ?

henvic:
Hi. Any news on this one?

rdp-cloud:
One point I would raise.. upon renewal, it would be preferred to 'disable' the previous access keys rather than remove them altogether. Removing them could have disasterous consequences if key management is less than ideal within an organisation.

Naturally, this would affect the renewal workflow as there are only 2 sets of keys available at any given time.

1. search for 'disabled' keys
2. remove any disabled keys
3. generate new keys 
4. set original 'active' key to disabled

loopingz:
@rdp-cloud The problem with disabling is that then it is in the way for the next rotation. With removing if i want to rotate every hour i can.

But I can add an option to not remove and disable instead

awscli/examples/configure/rotate-access-keys/_examples.rst
```diff
@@ -0,0 +1,37 @@
...
+    aws_access_key_id = renew_access_key
+    aws_secret_access_key = renew_secret_key
+
+    [other]
+    aws_access_key_id = other_access_key
+    aws_secret_access_key = other_secret_key
+
+Then the following commands::
+
+    $ aws --profile=test configure rotate-access-key
```
tingletech: `--profile=other`?

loopingz: Correct i ll fix this one

mmraz:
Is this still being considered?  It aggregates a set of steps we currently have to take sequenced by an external script.  Would really be nice to consolidate them down into a single call with a single return code to determine success or failure as an atomic operation.

loopingz:
@mmraz while waiting for AWS team (1 year and a half now) i did this small project: https://github.com/loopingz/yaws it has the two functions i added to awscli but never merged

loopingz:
Why this still has no review or reason, even while reaching to our enterprise support to get things move forward...

shaunhickson:
Thank you for this contribution. I’m very sorry that it took us so long to respond. As key rotation is something that has the potential to impact a number of areas across the AWS ecosystem, we are not able to accept this PR at this time. We are going to have an internal discussion regarding this and once we’ve identified what the next steps will be we will be in touch with more details. We will follow up with you as soon as we can.

mmraz:
> We are going to have an internal discussion regarding this and once we’ve identified what the next steps will be we will be in touch with more details

Can this feature be added to the aws-cli roadmap?  I check there regularly.

loopingz:
I do not understand how key rotation is problematic, it is even a part of white paper, trying to help users to achieve better security does not seem  problematic to me, and as it is just a command if you don't run it it does not rotate any keys by itself.

andymac4182:
@shaunhickson Any update on this? We are using https://github.com/rhyeal/aws-rotate-iam-keys to handle this for our local machines but would be great if it could be built in. We have a forced 90 day expiry on our personal access keys so it would be a lot of help to have this added in natively

PR #2715, opened 2017-07-14
Clarify presign region requirements

JordonPhillips:
cc @dstufft @stealthycoin

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -634,9 +634,12 @@ class WebsiteCommand(S3Command):
 
 class PresignCommand(S3Command):
     NAME = 'presign'
-    DESCRIPTION = ("Generate a pre-signed URL for an Amazon S3 object. "
-                   "This allows anyone who receives the pre-signed URL "
-                   "to retrieve the S3 object with an HTTP GET request.")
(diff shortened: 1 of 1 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -634,9 +634,13 @@ def _get_bucket_name(self, path):
 
 class PresignCommand(S3Command):
     NAME = 'presign'
-    DESCRIPTION = ("Generate a pre-signed URL for an Amazon S3 object. "
-                   "This allows anyone who receives the pre-signed URL "
-                   "to retrieve the S3 object with an HTTP GET request.")
+    DESCRIPTION = (
+        "Generate a pre-signed URL for an Amazon S3 object. This allows "
+        "anyone who receives the pre-signed URL to retrieve the S3 object "
+        "with an HTTP GET request. For sigv4 requests the --region parameter "
```
kyleknap: It is not necessary to use the ``--region`` parameter correct? They just have to make sure the region matches up with wherever the bucket is so it can be simply configured. It may be worth rewording this to make sure users have the correct region configured as opposed to saying `--region` should be provided.

JordonPhillips:
updated

PR #2720, opened 2017-07-20
Fix typo in configure examples

pallxk:


awscli/examples/configure/get/_examples.rst
```diff
@@ -24,12 +24,12 @@ The following commands would have the corresponding output::
...
 
     $ aws configure get profile.testing.aws_access_key_id
-    default_access_key
+    testing_access_key
 
     $ aws configure get preview.cloudsearch
     true
 
     $ aws configure get preview.does-not-exist
-    $
```
JordonPhillips: This is showing that there is no output, resulting in the prompt immediately appearing. I would prefer to keep this.

JordonPhillips:
Thanks for the PR! Taking a look.

awscli/examples/configure/get/_examples.rst
pallxk: I made the change as I saw there is always an empty line between prompts. I ran my test with PS1 set to `"\n$ "`. I'm also okay to keep this and amend my commit.

pallxk:
Rebased and amended to include most relevant changes to this PR.

joguSD:
Cool. Thanks for incorporating that feedback. Merging.

PR #2740, opened 2017-07-31
Fix unicode bug in s3 progress

### LINKED ISSUE #2738: Sync to s3 hangs forever - 'ascii' codec can't encode character
Think I'm experiencing an issue similar to #2575 attempting to sync from my Mac's local filesystem to s3:

    UnicodeEncodeError: 'ascii' codec can't encode character u'\u0303' in position 11: ordinal not in range(128)
    2017-07-29 00:49:40,775 - Thread-1 - awscli.customizations.s3.results - DEBUG - Error processing result QueuedResult(transfer_type='upload'

    File "/Users/jasonsturges/Library/Python/2.7/lib/python/site-packages/awscli/customizations/s3/results.py", line 604, in _process_result
        result_handler(result)
    File "/Users/jasonsturges/Library/Python/2.7/lib/python/site-packages/awscli/customizations/s3/results.py", line 241, in __call__
        result=result)
    File "/Users/jasonsturges/Library/Python/2.7/lib/python/site-packages/awscli/customizations/s3/results.py", line 267, in _record_queued_result
        self._get_ongoing_dict_key(result)] = total_transfer_size
    File "/Users/jasonsturges/Library/Python/2.7/lib/python/site-packages/awscli/customizations/s3/results.py", line 250, in _get_ongoing_dict_key
        str(el) for el in [result.transfer_type, result.src, result.dest])
    File "/Users/jasonsturges/Library/Python/2.7/lib/python/site-packages/awscli/customizations/s3/results.py", line 250, in <genexpr>
        str(el) for el in [result.transfer_type, result.src, result.dest])

Without the `--debug` option, it appears to hang forever at "0 files", repeatedly trying the same operations while pummeling the network.

Within the folder I'm attempting to sync, there's a folder with the 'ñ' character in the name.

Sync command is simply:

    aws s3 sync . s3://[bucket]/

Strange, as this folder exists in s3 - though I can't recall, it is possible that I initially uploaded it using the aws `cp` command instead of `sync` in the past.

Also strange is that adding `--dryrun` with sync works perfectly fine, correctly indicating a list of files to be uploaded.

Versions:

- aws-cli/1.11.94 Python/2.7.10 Darwin/15.6.0 botocore/1.5.57
- OS X El Capitan (10.11.6)

JordonPhillips:
Fixes #2738

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -22,7 +22,7 @@ from s3transfer.exceptions import CancelledError
 from s3transfer.exceptions import FatalError
 from s3transfer.subscribers import BaseSubscriber
 
-from awscli.compat import queue
+from awscli.compat import queue, six
 from awscli.customizations.s3.utils import relative_path
 from awscli.customizations.s3.utils import human_readable_size
 from awscli.customizations.utils import uni_print
@@ -246,8 +246,8 @@ class ResultRecorder(BaseResultHandler):
                 'Any result using _get_ongoing_dict_key must subclass from '
                 'BaseResult. Provided result is of type: %s' % type(result)
             )
-        return ':'.join(
-            str(el) for el in [result.transfer_type, result.src, result.dest])
+        properties = [result.transfer_type, result.src, result.dest]
+        return u':'.join(six.text_type(p) for p in properties)
 
     def _pop_result_from_ongoing_dicts(self, result):
         ongoing_key = self._get_ongoing_dict_key(result)
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -22,7 +22,7 @@ from s3transfer.exceptions import CancelledError
 from s3transfer.exceptions import FatalError
 from s3transfer.subscribers import BaseSubscriber
 
-from awscli.compat import queue, six
+from awscli.compat import queue, ensure_text_type
 from awscli.customizations.s3.utils import relative_path
 from awscli.customizations.s3.utils import human_readable_size
 from awscli.customizations.utils import uni_print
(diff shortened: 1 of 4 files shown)


awscli/customizations/s3/results.py
```diff
@@ -246,8 +246,8 @@ def _get_ongoing_dict_key(self, result):
                 'Any result using _get_ongoing_dict_key must subclass from '
                 'BaseResult. Provided result is of type: %s' % type(result)
             )
-        return ':'.join(
-            str(el) for el in [result.transfer_type, result.src, result.dest])
+        properties = [result.transfer_type, result.src, result.dest]
+        return u':'.join(six.text_type(p) for p in properties)
```
dstufft: Do we know if these are guaranteed to already be decoded and so they'll always be either ASCII only ``str`` *or* ``unicode`` objects? Because blindly casting to ``unicode`` on 2.7 will raise an error if the ``str`` contains bytes that are not ASCII encoded. e.g.:

```pycon
>>> unicode(u"☃".encode("utf8"))
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
```

If we know that isn't going to be an issue, then we're good and we can just merge this. If not, we might have to do something more here.

awscli/customizations/s3/results.py
```diff
@@ -246,8 +246,11 @@ def _get_ongoing_dict_key(self, result):
                 'Any result using _get_ongoing_dict_key must subclass from '
                 'BaseResult. Provided result is of type: %s' % type(result)
             )
-        return ':'.join(
-            str(el) for el in [result.transfer_type, result.src, result.dest])
+        key_parts = []
+        for result_property in [result.transfer_type, result.src, result.dest]:
+            if result_property is not None:
```
kyleknap: I think the ongoing key will change if this `None` check is used. Might be worth to double check, but from what I can tell it is fine to change as the key is only used to get the a value out of dictionary. If the key was used for anything else, there may be issues.

JordonPhillips: That's all it's used for, though I can have it add an empty string pretty easily.

kyleknap: I do not have an opinion either way. Pretty confident is not needed given we have a lot of tests for this and we both looked through the code. Plus if you want exact parity, the value would be `"None"` and not empty string, which is weird to rely on. So really the only thing that could break is if we are splitting on `':'` and expecting three elements, which we do not do anywhere in the code.

dfabreguette-ap:
+1 Is this merged in prod ? Not working on aws-cli/1.16.6 Python/3.5.3 Linux/4.9.0-6-amd64 botocore/1.11.6

PR #2746, opened 2017-08-03
Condition function doesn't support "Fn::" prefix

pmarques:
# Summary

Fix the AWS CLI cloudformation package command

# Details

When we use CLI tools to package a template it adds the function prefix *Fn::*. If we try to create a stack using the *package* cloudformation template it will thrown an error:

```
Template validation error: Template Error: Encountered unsupported function: Fn::Condition Supported functions are: [Fn::Base64, Fn::GetAtt, Fn::GetAZs, Fn::ImportValue, Fn::Join, Fn::Split, Fn::FindInMap, Fn::Select, Ref, Fn::Equals, Fn::If, Fn::Not, Condition, Fn::And, Fn::Or, Fn::Contains, Fn::EachMemberEquals, Fn::EachMemberIn, Fn::ValueOf, Fn::ValueOfAll, Fn::RefAll, Fn::Sub]
```

# References: 
 * [AWS Condition Documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html)

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -26,7 +26,7 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
     # Get the actual tag name excluding the first exclamation
     tag = node.tag[1:]
 
-    # Some intrinsic functions doesn't support prefix "Fn::"
+    # Some intrinsic functions doesn't support prefix "Fn::"
     prefix = "Fn::"
     if tag in ["Ref", "Condition"]:
         prefix = ""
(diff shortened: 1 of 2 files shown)


pmarques:
@JordonPhillips can you have a look in this one

awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -28,7 +28,7 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
 
     # All CloudFormation intrinsics have prefix Fn:: except Ref
```
joguSD: Can you update this comment to reflect the added tag?

joguSD:
One small comment, otherwise it looks alright to me.

awscli/customizations/cloudformation/yamlhelper.py
pmarques: Update. Changed to a more generic message.

PR #2747, opened 2017-08-03
Add no progress result printer to s3 transfer

joguSD:
This seemed relatively easy to implement and a lot of people have been asking for it.
This is basically a copy paste of the `OnlyShowErrorsResultPrinter` except that it still prints success messages. Just like the `--only-show-errors` flag you can now also pass `--no-progress` to omit the progress rendering. Not sure if I hit all the tests we need for this.

Implements: https://github.com/aws/aws-cli/issues/519

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -369,9 +369,13 @@ ONLY_SHOW_ERRORS = {'name': 'only-show-errors', 'action': 'store_true',
                         'output is suppressed.')}
 
 
-NO_PROGRESS = {'name': 'no-progress', 'action': 'store_true',
-                    'help_text': (
-                        'File transfer progress is not displayed.')}
+NO_PROGRESS = {'name': 'no-progress',
+               'action': 'store_false',
+               'dest': 'progress',
+               'help_text': (
+                   'File transfer progress is not displayed. This flag '
+                   'is only applied when the quiet and only-show-errors '
+                   'flags are not provided.')}
 
 
 EXPECTED_SIZE = {'name': 'expected-size',
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -369,6 +369,11 @@ ONLY_SHOW_ERRORS = {'name': 'only-show-errors', 'action': 'store_true',
                         'output is suppressed.')}
 
 
+NO_PROGRESS = {'name': 'no-progress', 'action': 'store_true',
+                    'help_text': (
+                        'File transfer progress is not displayed.')}
+
+
 EXPECTED_SIZE = {'name': 'expected-size',
                  'help_text': (
                      'This argument specifies the expected size of a stream '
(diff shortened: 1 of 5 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -369,6 +369,11 @@
                         'output is suppressed.')}
 
 
+NO_PROGRESS = {'name': 'no-progress', 'action': 'store_true',
```
jamesls: The `no-*` params typically have a `store_false` action with the destination being the param without the `no` prefix (`progress` in this case).  Also makes it less confusing if we ever add an explicit true version of this arg (`--progress`).

awscli/customizations/s3/subcommands.py
```diff
@@ -369,6 +369,11 @@
...
+                    'help_text': (
+                        'File transfer progress is not displayed.')}
```
jamesls: It's probably worth documenting that this arg only matter is the other two options aren't provided (quiet/only_show_errors)

c0reysc0tt:
I really hope this gets merged into an upcoming version soon. I manually made these changes to the three python files affected, and now the log files created by all my s3 cron scripts are useful and readable. As far as I can tell, it's a complete fix to a problem people have been complaining about for years. The --no progress option is documented in the help files the same as any other function and does what it says it does.  Thank you joguSD!

awscli/customizations/s3/subcommands.py
```diff
@@ -369,6 +369,15 @@
                         'output is suppressed.')}
 
 
+NO_PROGRESS = {'name': 'no-progress',
+               'action': 'store_false',
+               'dest': 'progress',
```
kyleknap: Not sure it is needed but all of the parameters like this have a `'default': True` in the dictionary. Not sure if it is needed. Do not think it is. But it would not hurt to be consistent.

awscli/customizations/s3/subcommands.py
```diff
@@ -369,6 +369,15 @@
...
+               'help_text': (
+                   'File transfer progress is not displayed. This flag '
+                   'is only applied when the quiet and only-show-errors '
```
kyleknap: Does the second sentence even matter? Based on the `--quiet` and `--only-show-errors` documentation, it will not show progress anyways. I would prefer to remove it in case we do not cause any confusion in the future if we add a new output parameter and forget to update this documentation.

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -1255,6 +1255,17 @@ def test_normal_output_only_show_errors(self):
...
         self.assertEqual('', p.stdout)
 
+    def test_normal_output_no_progress(self):
+        bucket_name = _SHARED_BUCKET
+        foo_txt = self.files.create_file('foo.txt', 'foo contents')
+
+        # Copy file into bucket.
+        p = aws('s3 cp %s s3://%s/ --no-progress' % (foo_txt, bucket_name))
+        self.assertEqual(p.rc, 0)
+        # Ensure success message was printed
```
kyleknap: It might be good to assert there was no progress statement as well in the standard output.

awscli/customizations/s3/subcommands.py
joguSD: The second sentence was added in response to some of James' feedback.

kyleknap: Oh its because it is conflicting. Yeah I'm fine with keeping it as is.

awscli/customizations/s3/subcommands.py
joguSD: The other two flags like this don't have a default either, I think it's fine to have no default.

tests/integration/customizations/s3/test_plugin.py
joguSD: Updated.

awscli/customizations/s3/subcommands.py
kyleknap: I was referring to all of the boolean flags that have `store_false` as the action. I just checked the docs and did it on the command line and its fine. Putting the `default` kwarg whose value is `True` is redundant.

Issue #2275, opened 2016-11-03
Cloudformation external parameter file do not support YAML file


Yaowenjie:
I used aws-cli to update my stacks, and want to use an external yaml file (dev.yaml) as parameters file like below:
```sh
aws cloudformation update-stack --stack-name test --template-body file://./api-test.yaml --parameters file://./dev.yaml
```
In `dev.yaml`, it only has:
```yaml

---
- ParameterKey: APICacheClusterSize
  ParameterValue: '1.6'
```
However, I got this Error message when I run the update command above:
```
Error parsing parameter '--parameters': Expected: '=', received: '-' for input:

---
- ParameterKey: APICacheClusterSize
^
  ParameterValue: '1.6'
```

If I use JSON file with same context as below, and use it instead of `dev.yaml`, it works fine:
```json
[
  {
    "ParameterKey": "APICacheClusterSize",
    "ParameterValue": "1.6"
  }
]
```

So, can the external parameters file for aws cloudformation be YAML file? or will...?

stealthycoin:
Hi, we currently don't have any plans to support YAML as part of the command line arguments. I will mark this as a feature request as I can see how it might be useful for some people.

The contents of the the file you provide to --parameters is loaded and dumped straight in as an argument so you can only have the file contents be in one of two formats:
Shorthand syntax:
`ParameterKey=string,ParameterValue=string,UsePreviousValue=boolean`
or JSON syntax:
```
[
    {
         "ParameterKey": "string",
         "ParameterValue": "string",
         "UsePreviousValue": true|false
     }
]
```

lorengordon:
I'll second this feature request.

jlm742:
I'll second it too

csgyuricza:
+1

dcw329:
Please. This is critical for us.

thijsdev:
+1

ddepaoli3:
absolutely +1

codeasone:
+1 allowing YAML formatted data for `--template-body` but not for `--parameters` is awkward

youngpm:
👍

amitlalani:
+1

andymotta:
+1

jdbass:
+1

mazerty:
since you can provide a template in yaml, you should be able to provide the parameters in yaml too
i think this feature is a must-have

KWSeamanCerner:
+1

nungster:
+100

[...]

PR #2749, opened 2017-08-06
Adding YAML support

lucioveloso:
#2275 Cloudformation external parameter file do not support YAML file

I changed the argprocess to be able to handle Json and Yaml files. This version provide options to use parameters with Json or Yaml.


lucioveloso:
Fixing impact in other projects.

PR #2752, opened 2017-08-07
Adding YAML support to aws-cli arguments

lucioveloso:
#2275 Proposing this new version to add YAML support to the aws-cli.

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -13,6 +13,7 @@
 """Module for processing CLI args."""
 import os
 import logging
+import yaml
 from awscli.compat import six
 
 from botocore.compat import OrderedDict, json
@@ -177,7 +178,7 @@ def _special_type(model):
 
 def _unpack_cli_arg(argument_model, value, cli_name):
     if is_json_value_header(argument_model):
-        return _unpack_json_cli_arg(argument_model, value, cli_name)
+        return _unpack_json_cli_arg(value, cli_name)
     elif argument_model.type_name in SCALAR_TYPES:
         return unpack_scalar_cli_arg(
             argument_model, value, cli_name)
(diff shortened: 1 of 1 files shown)


awscli/argprocess.py
```diff
@@ -188,13 +189,20 @@ def _unpack_cli_arg(argument_model, value, cli_name):
...
-    try:
-        return json.loads(value, object_pairs_hook=OrderedDict)
-    except ValueError as e:
-        raise ParamError(
-            cli_name, "Invalid JSON: %s\nJSON received: %s"
-            % (e, value))
+def _unpack_json_cli_arg(value, cli_name):
+    if value.lstrip().startswith("---"):
+        try:
+            return yaml.load(value)
```
epl: Strongly recommend changing all `yaml.load()` calls to `yaml.safe_load()` instead. For more details, please see https://pyyaml.org/wiki/PyYAMLDocumentation#loading-yaml

While aws-cli is _mostly_ used in situations where the input YAML is safe, it's likely that someone may build a layer on top of aws-cli which takes user input that affects the YAML and might cause a future security bug. Given there's no good reason why you'd want to load arbitrary Python objects, there's no harm done by this change.

Also see: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/yamlhelper.py

awscli/argprocess.py
```diff
@@ -188,13 +189,20 @@ def _unpack_cli_arg(argument_model, value, cli_name):
...
+        except Exception as e:
+            raise ParamError(cli_name, "Invalid JSON or YAML: %s\nReceived: %s"
```
epl: If the file starts with `---`, can the data ever be JSON? If not, I'd like to suggest that the error message is changed.

awscli/argprocess.py
monken: JSON is YAML. Any compliant YAML parser will be able to parse JSON. No need to implement hacks like looking for the dashes (which is not required to be valid YAML). Just throw out the JSON parser and use the YAML parser to also parse JSON.

Issue #2753, opened 2017-08-07
Invalid '-e' flag for docker login generated by ecr get-login


panga:
Problem:
The 'ecr get-login' command is generating an incompatible syntax for latest docker version.
Docker 17.06 doesn't support the deprecated '-e' flag anymore and give the following error:
```
unknown shorthand flag: 'e' in -e
See 'docker login --help'.
```

**This is a blocker for all my scripts using aws ecr get-login.**

Command:
`aws ecr get-login`

Result:
`docker login -u AWS -p <token> -e none https://<accountid>.dkr.ecr.us-east-1.amazonaws.com`

Versions:
aws-cli/1.11.129 Python/2.7.10 Darwin/16.7.0 botocore/1.5.92
Docker version 17.06.0-ce, build 02c1d87

'--email' flag removal PR: https://github.com/docker/cli/pull/143

joguSD:
You can remove that by specifying the `--no-include-email` flag. We default to including the email so we don't break customers still relying on it.

Duplicate of #2743, #2732, #2694, #2691, #2676, and a few others

AndrewRayCode:
@joguSD Looks incorrect, I'm getting
```
unknown flag: --no-include-email
```
Removing the `-e none` part of the aws login command entirely let me log in though

schematical:
+1 getting the same as @AndrewRayCode

tarikozket:
(@AndrewRayCode @schematical)

@joguSD means using `aws ecr get-login --no-include-email`. It worked perfectly for me!

aws-cli/1.16.102
Docker version 18.06.1-ce

Issue #2743, opened 2017-08-01
`asw ecr get-login` includes deprecated Docker flag


drewandersen:
Execute `aws ecr get-login --region=us-west-2` and receive the following error:

```
unknown shorthand flag: 'e' in -e
See 'docker login --help'.
```
The email flag was removed in release 17.06.0-ce: https://github.com/docker/cli/pull/143

joguSD:
You can remove that by specifying the `--no-include-email` flag. We default to including the email so we don't break customers still relying on it.

Duplicate of #2732, #2694, #2691, #2676, and a few others

Issue #2732, opened 2017-07-27
aws ecr get-login -e flag deprecated


max3-05:
According to the Docker changeset:

> `-e` and `--email` flags on docker login
> 
> Deprecated In Release: v1.11.0
> Removed In Release: v17.06
> 
> The docker login command is removing the ability to automatically register for an account with the target registry if the given username doesn’t exist. Due to this change, the email flag is no longer required, and will be deprecated.

Please, reflect this changes and remove obsolete `-e` flag from the output of `aws ecr get-login` command

JordonPhillips:
You can remove that by specifying the `--no-include-email` flag. We default to including the email so we don't break customers still relying on it.

Duplicate of #2694, #2691, #2676, and a few others

Issue #2718, opened 2017-07-17
Docker: deprecated --email flag is now removed


rylanhalteman:
The command `aws ecr get login` no longer provides a valid docker login command on the latest version of docker (17.06.0-ce).

The resulting command fails with the following output
```
❯ aws ecr get-login
❯ docker login -u AWS -p <private_key> -e none https://<snip>.dkr.ecr.us-east-1.amazonaws.com
unknown shorthand flag: 'e' in -e
See 'docker login --help'.
```

The `--email` option was deprecated previously, and removed in the latest version of docker https://docs.docker.com/release-notes/docker-ce/#17060-ce-2017-06-28.

This is breaking workflows for our developers, and should be fixed asap, as others will continue to upgrade to the latest version of docker.

Temporary fix: `aws ecr get-login | sed 's/\-e none//'`.  I will attempt a PR for this later today.

rylanhalteman:
I made this issue before digging into the code.  Looks like there's already an option to not include the email https://github.com/aws/aws-cli/blob/develop/awscli/customizations/ecr.py#L49-L56.

Closing.

PR #2768, opened 2017-08-16
ecr default --no-include-email to get-login command

msrai:
#2753 #2743 #2732 #2718 #2691 
flag -e/--email are deprecated since docker 1.11 and removed in docker 17.06.0-ce
https://docs.docker.com/release-notes/docker-ce/#17060-ce-2017-06-28


joguSD:
Might be a while before we can change this default.

JordonPhillips:
Due to backwards compatibility concerns we won't be able to change this default for quite a while.

PR #2770, opened 2017-08-16
Fix input option structure indentation for empty structures

joguSD:
This change ensures that the indentation is correct when closing a structure in input when that structure has no members. The previous logic would only correctly dedent when the structure had members. I tried adding some tests for this but got a little stuck as this is an issue only when we use the posix documentation rendering, but in all of the tests for this we mock the render class which ignores the document writers style so indentation ends up being ignored.


JordonPhillips:
Could you please add a test for this?

joguSD:
Added a test to validate `indent` and `dedent` calls are equal ensuring neutral indentation after writing a nested structure.

PR #2778, opened 2017-08-18
aws cloudformation deploy large template support

### LINKED ISSUE #2461: Allow aws cloudformation deploy to use template-url
As the [How to create serverless applications using AWS SAM](https://github.com/awslabs/serverless-application-model/blob/master/HOWTO.md) says : 

> SAM template is deployed to AWS CloudFormation by creating a changeset using the SAM template followed by executing the changeset... ...Alternatively, you can use aws cloudformation deploy CLI command to deploy the SAM template. Under-the-hood it creates and executes a changeset and waits until the deployment completes.

Nevertheless, a Cloudformation [create-change-set](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-change-set.html) allows to use the `--template-url` parameter, however the `aws cloudformation deploy` command does not. Please update the CLI to support `--template-url` while doing a `deploy`.

This is a feature already available on the AWS Management Console, where  you can deploy your stack using the URL that points to a template that is located in an S3 bucket.

AWS CLI version: 1.11.46

pwrmiller:
Currently the AWS SAM extensions for aws-cli only support CloudFormation templates smaller than 51200 bytes (this is a published limitation of CloudFormation). 

In order to support larger templates, we can first upload these templates to S3, then create the changeset for the deployment of the templates. This adds options to `aws cloudformation deploy` to facilitate the deployment of large templates (460,800 bytes).

This fixes #2461 by allowing `--s3-bucket` to be passed in `aws cloudformation deploy` along with `--template-file`. If the template is larger than 51200 bytes, the template is uploaded to the S3 bucket and the CloudFormation ChangeSet is created referencing the S3 URL. We use the included s3uploader of awscli/customizations/cloudformation the same way as it is used in SAM packaging. 

This also fixes issue number 151 in awslabs/serverless-application-model. 

@sanathkr has commented on both those issues, so might be a good resource for discussion on this pull request.

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -15,8 +15,11 @@ import os
 import sys
 import logging
 
+from botocore.client import Config
+
 from awscli.customizations.cloudformation import exceptions
 from awscli.customizations.cloudformation.deployer import Deployer
+from awscli.customizations.cloudformation.s3uploader import S3Uploader
 from awscli.customizations.cloudformation.yamlhelper import yaml_parse
 
 from awscli.customizations.commands import BasicCommand
@@ -59,6 +62,40 @@ class DeployCommand(BasicCommand):
                 'The name of the AWS CloudFormation stack you\'re deploying to.'
                 ' If you specify an existing stack, the command updates the'
                 ' stack. If you specify a new stack, the command creates it.'
+            )
+        },
+        {
+            'name': 's3-bucket',
+            'required': False,
+            'help_text': (
+                'The name of the S3 bucket where this command uploads'
+                ' your template if required.'
+            )
+        },
+        {
+            "name": "force-upload",
+            "action": "store_true",
+            "help_text": (
+                'Indicates whether to override existing files in the S3 bucket.'
+                ' Specify this flag to upload artifacts even if they '
+                ' match existing artifacts in the S3 bucket.'
+            )
+        },
+        {
+            'name': 's3-prefix',
+            'help_text': (
+                'A prefix name that the command adds to the'
+                ' artifacts\' name when it uploads them to the S3 bucket.'
+                ' The prefix name is a path name (folder name) for'
+                ' the S3 bucket.'
+            )
+        },
+
+        {
+            'name': 'kms-key-id',
+            'help_text': (
+                'The ID of an AWS KMS key that the command uses'
+                ' to encrypt artifacts that are at rest in the S3 bucket.'
             )
         },
         {
@@ -177,22 +214,41 @@ class DeployCommand(BasicCommand):
 
         parameters = self.merge_parameters(template_dict, parameter_overrides)
 
+        bucket = parsed_args.s3_bucket
+        if bucket:
+            s3_client = self._session.create_client(
+                "s3",
+                config=Config(signature_version='s3v4'),
+                region_name=parsed_globals.region,
+                verify=parsed_globals.verify_ssl)
+
+            s3_uploader = S3Uploader(s3_client,
+                                      bucket,
+                                      parsed_globals.region,
+                                      parsed_args.s3_prefix,
+                                      parsed_args.kms_key_id,
+                                      parsed_args.force_upload)
+        else:
+            s3_uploader = None
+
         deployer = Deployer(cloudformation_client)
         return self.deploy(deployer, stack_name, template_str,
                            parameters, parsed_args.capabilities,
                            parsed_args.execute_changeset, parsed_args.role_arn,
-                           parsed_args.notification_arns)
+                           parsed_args.notification_arns, s3_uploader)
+
 
     def deploy(self, deployer, stack_name, template_str,
                parameters, capabilities, execute_changeset, role_arn,
-               notification_arns):
+               notification_arns, s3_uploader):
         result = deployer.create_and_wait_for_changeset(
                 stack_name=stack_name,
                 cfn_template=template_str,
                 parameter_values=parameters,
                 capabilities=capabilities,
                 role_arn=role_arn,
-                notification_arns=notification_arns)
+                notification_arns=notification_arns,
+                s3_uploader=s3_uploader)
 
         if execute_changeset:
             deployer.execute_changeset(result.changeset_id, stack_name)
--- a/awscli/customizations/cloudformation/deployer.py
+++ b/awscli/customizations/cloudformation/deployer.py
@@ -18,6 +18,8 @@ import botocore
 import collections
 
 from awscli.customizations.cloudformation import exceptions
+from awscli.customizations.cloudformation.artifact_exporter import mktempfile
+
 from datetime import datetime
 
 LOG = logging.getLogger(__name__)
@@ -71,7 +73,7 @@ class Deployer(object):
 
     def create_changeset(self, stack_name, cfn_template,
                          parameter_values, capabilities, role_arn,
-                         notification_arns):
+                         notification_arns, s3_uploader):
         """
         Call Cloudformation to create a changeset and wait for it to complete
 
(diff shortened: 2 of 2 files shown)


sanathkr:
Thanks for doing this! I'll look at the PR

awscli/customizations/cloudformation/deployer.py
```diff
@@ -106,6 +108,26 @@ def create_changeset(self, stack_name, cfn_template,
...
+            if s3_uploader:
+                with mktempfile() as temporary_file:
+                    temporary_file.write(kwargs.pop('TemplateBody'))
+                    temporary_file.flush()
+                    url = s3_uploader.upload_with_dedup(
+                            temporary_file.name, "template")
+                    kwargs['TemplateURL'] = s3_uploader.to_path_style_s3_url(url.split('/')[-1])
+            else:
+                LOG.debug("Template size is too large and S3 bucket not given")
+                raise ValueError(("A template with a size greater than 51200 "
```
sanathkr: Can you use a custom exception like this instead? https://github.com/aws/aws-cli/blob/master/awscli/customizations/cloudformation/exceptions.py 

The idea is to keep user facing error messages in a separate file so we can always look for them in future.

sanathkr: For example usage, checkout https://github.com/aws/aws-cli/blob/50b4ccfb85acb714445cbb4a6fb0582ade8af2b0/awscli/customizations/cloudformation/artifact_exporter.py#L141

tests/unit/customizations/cloudformation/test_deployer.py
```diff
@@ -117,7 +118,7 @@ def test_create_changeset_success(self):
             "Capabilities": capabilities,
             "Description": botocore.stub.ANY,
             "RoleARN": role_arn,
-            "NotificationARNs": notification_arns
+            "NotificationARNs": notification_arns,
```
sanathkr: Are you missing s3_uploader here?

awscli/customizations/cloudformation/deployer.py
```diff
@@ -106,6 +108,26 @@ def create_changeset(self, stack_name, cfn_template,
             'Capabilities': capabilities,
             'Description': description,
         }
+
+        # CloudFormation requires template bodies to be less than 51200 bytes
+        # if they are deployed from a file. It supports much larger templates
+        # if the templates exist in s3. If this is a 'big' template, throw it
+        # into s3 first.
+        cfn_template_length = len(cfn_template.encode('utf-8'))
```
sanathkr: We need unit tests for this logic. It will be easier if you move this code into a separate method

awscli/customizations/cloudformation/deploy.py
```diff
@@ -62,6 +65,40 @@ class DeployCommand(BasicCommand):
             )
         },
         {
+            'name': 's3-bucket',
+            'required': False,
+            'help_text': (
+                'The name of the S3 bucket where this command uploads'
```
sanathkr: I would expand help string to precisely explain when you upload to S3, and how it is optional.

awscli/customizations/cloudformation/deployer.py
pwrmiller: I've modified the logic slightly - this might have issues if CloudFormation ever supports localized templates. Instead, I've moved this out to `deploy.py` and check the actual file of the template before loading it. 

I've also added appropriate unit tests for this condition.

awscli/customizations/cloudformation/deployer.py
pwrmiller: I've added in the custom exception as requested.

tests/unit/customizations/cloudformation/test_deployer.py
pwrmiller: These parameters don't accept an s3_uploader - these are used in the stubed call to boto3 create_change_set, which accepts a TemplateBody or a TemplateURL. A test has been added for deploying the change set using TemplateURL rather than TemplateBody. 

We now use a TemplateURL if an s3_uploader is passed, otherwise TemplateBody is used. By default, the s3_uploader is None, so we use TemplateBody by default (as it was before).

awscli/customizations/cloudformation/deploy.py
pwrmiller: Expanded as requested.

awscli/customizations/cloudformation/deploy.py
```diff
@@ -214,6 +215,10 @@ def _run_main(self, parsed_args, parsed_globals):
 
         parameters = self.merge_parameters(template_dict, parameter_overrides)
 
+        template_size = os.path.getsize(parsed_args.template_file)
+        if template_size > 51200 and not parsed_args.s3_bucket:
```
sanathkr: fantastic! thanks

tests/unit/customizations/cloudformation/test_deployer.py
```diff
@@ -142,7 +143,76 @@ def test_create_changeset_success(self):
...
+        self.stub_client.add_response("create_change_set", response,
+                                      expected_params)
+        with self.stub_client:
+            result = self.deployer.create_changeset(
+                    stack_name, template, parameters, capabilities, role_arn,
+                    notification_arns, s3_uploader)
+            self.assertEquals(response["Id"], result.changeset_id)
+            self.assertEquals("CREATE", result.changeset_type)
+
+        # Case 2: Stack exists. We are updating it
```
sanathkr: thanks for covering this case as well!

sanathkr:
@JordonPhillips Can we get this merged?

junleqian:
When this feature will get merged? This is really a tempting feature that saves me time from reading the nested stack documentation.

awscli/customizations/cloudformation/deployer.py
```diff
@@ -106,6 +108,17 @@ def create_changeset(self, stack_name, cfn_template,
...
+
+        # If an S3 uploader is available, use TemplateURL to deploy rather than
+        # TemplateBody. This is required for large templates.
+        if s3_uploader:
+            with mktempfile() as temporary_file:
+                temporary_file.write(kwargs.pop('TemplateBody'))
+                temporary_file.flush()
+                url = s3_uploader.upload_with_dedup(
+                        temporary_file.name, "template")
+                kwargs['TemplateURL'] = s3_uploader.to_path_style_s3_url(url.split('/')[-1])
```
junleqian: This URL creation should be changing based on the parameter _s3_prefix_. If not, the _CreateChangeSet_ step would get a _object does not exist_ error.

junleqian: Example error:

```
examplePrefix/examplehash.template  100000 / 100000  (100.00%)
An error occurred (ValidationError) when calling the CreateChangeSet operation: Template file referenced by https://s3.amazonaws.com/examplebucketname/examplehash.template does not exist.
```

junleqian: This seems to be a bug  in the _s3_uploader.to_path_style_s3_url_ method. Ideally, the prefix part of the result url should be:
```
base = "{0}/{1}/{2}".format(base, self.bucket_name, self.prefix)
```
when _self.prefix_ is present.

sanathkr: This is not a bug in to_path_style_s3_url, but a bug in the url.split. Parsing a S3 URL has more nuances. Correct implementation should follow this: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/artifact_exporter.py#L351-L357

pwrmiller:
@sanathkr and @junleqian, I've corrected the handling of the S3 URL using the approach you've mentioned. Thanks for pointing me in the right direction!

jrstarke:
@sanathkr Are we waiting for a future version to support this?

ctcsi:
@pwrmiller How long until this get's merged?

Dunedan:
Excited I installed 1.14.26 only to notice that this PR got reverted after being merged. 😞 
The commit message of the revert commit also doesn't make much sense to me. What's the status of this PR now?

PR #2784, opened 2017-08-22
clarify timestamp in s3 ls output

mwunderl:


mwunderl:
Where does the CLI get the timezone? I tried setting it in `/etc/sysconfig/clock` but still see UTC timestamps in the output.

kyleknap:
It gets the timezone on this line: https://github.com/aws/aws-cli/blob/develop/awscli/customizations/s3/subcommands.py#L564 when it calls `dateutil.tz.tzlocal()`. I would need to do more research on how that dependency actually resolves the timezone to use.

mwunderl:
OK. I've updated the text to match the intended behavior.

Issue #2806, opened 2017-09-01
Profile does not return region or output format if it contains a SPACE character


CliveJL:
The AWS CLI tool allows you to create a profile name with a space character in it, and writes the values to the credentials and config file as expected. However when getting that config using the quoted profile name, it doesn't return the region or default output format.

Setting the profile credentials and config:

    $ aws configure --profile 'Foo Bar'
    AWS Access Key ID [None]: foo
    AWS Secret Access Key [None]: bar
    Default region name [None]: eu-west-1
    Default output format [None]: json

    $ cat .aws/config
    [profile Foo Bar]
    output = json
    region = eu-west-1

    $ cat .aws/credentials
    [Foo Bar]
    aws_access_key_id = foo
    aws_secret_access_key = bar

Checking the profile credentials and config:

    $ aws configure --profile 'Foo Bar'
    AWS Access Key ID [****************foo]:
    AWS Secret Access Key [****************bar]:
    Default region name [None]:
    Default output format [None]:

CLI commands then prompt for a region when using the profile name. The **credentials** do still work, and are retrieved from the profile, but the **config** ,e.g. the region, has to be specified on the command line still:

    $ aws autoscaling describe-auto-scaling-groups --auto-scaling-group-names group_name --profile 'Foo Bar'
    You must specify a region. You can also configure your region by running "aws configure".

When performing the same steps above with a profile name without a space character (Foo_Bar), everything works as expected.

cwgem:
So after some debugging. Essentially the `configure` function for profile needs to check if the profile name has spaces and if so add single quotes around them. First off actual config loading happens in botocore and uses `shlex.split` on the profile section name:

https://github.com/boto/botocore/blob/develop/botocore/configloader.py#L254

To get the name of the profile. Unfortunately, this splits on whitespace which leads to your issue:

```
>>> import shlex
>>> test = 'profile Foo Bar'
>>> shlex.split(test)
['profile', 'Foo', 'Bar']
```

 Now given the following profile:

```
[profile 'Foo Bar']
output = json
region = eu-west-1
```

If I run this through the python interactive shell (botocore parses config files through `configparser`):

```
>>> config = configparser.RawConfigParser()
>>> config.read('config.ini')
['config.ini']
>>> config.sections()
["profile 'Foo Bar'"]
>>> shlex.split(config.sections()[0])
['profile', 'Foo Bar']
```

The quoting ensure that the profile name even with spaces comes up properly.

[...]

PR #2812, opened 2017-09-05
Fix issue #2806: Support for section names that contain spaces

gtseres:
Fixes issue #2806 

## Testing
```
(aws-cli) ubuntu@ubuntu-xenial:~/workspace/aws-cli/aws-cli$ cat ~/.aws/config
cat: /home/ubuntu/.aws/config: No such file or directory
(aws-cli) ubuntu@ubuntu-xenial:~/workspace/aws-cli/aws-cli$ aws configure --profile 'Foo Bar'
AWS Access Key ID [None]: foo
AWS Secret Access Key [None]: bar
Default region name [None]: us-west-1
Default output format [None]: json
(aws-cli) ubuntu@ubuntu-xenial:~/workspace/aws-cli/aws-cli$ aws autoscaling describe-auto-scaling-groups --auto-scaling-group-names group_name --profile 'Foo Bar'

An error occurred (InvalidClientTokenId) when calling the DescribeAutoScalingGroups operation: The security token included in the request is invalid.
(aws-cli) ubuntu@ubuntu-xenial:~/workspace/aws-cli/aws-cli$ cat ~/.aws/config
[profile 'Foo Bar']
output = json
region = us-west-1
(aws-cli) ubuntu@ubuntu-xenial:~/workspace/aws-cli/aws-cli$
```

This indicates that the section has been successfully fetched from the profile.

### DIFF
--- a/awscli/customizations/configure/writer.py
+++ b/awscli/customizations/configure/writer.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import os
+import shlex
 import re
 
 from . import SectionNotFoundError
@@ -52,6 +53,10 @@ class ConfigFileWriter(object):
 
         """
         section_name = new_values.pop('__section__', 'default')
+        parts = shlex.split(section_name)
+        head, rest = parts[0], ' '.join(parts[1:])
+        if ' ' in rest:
+            section_name = "{} '{}'".format(head, rest)
         if not os.path.isfile(config_filename):
             self._create_file(config_filename)
             self._write_new_section(section_name, new_values, config_filename)
--- a/awscli/customizations/configure/writer.py
+++ b/awscli/customizations/configure/writer.py
@@ -11,7 +11,6 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import os
-import shlex
 import re
 
 from . import SectionNotFoundError
@@ -53,10 +52,6 @@ class ConfigFileWriter(object):
 
         """
         section_name = new_values.pop('__section__', 'default')
-        parts = section_name.split(' ')
-        head, rest = parts[0], ' '.join(parts[1:])
-        if ' ' in rest:
-            section_name = "{0} \"{1}\"".format(head, rest)
         if not os.path.isfile(config_filename):
             self._create_file(config_filename)
             self._write_new_section(section_name, new_values, config_filename)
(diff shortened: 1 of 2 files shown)


cwgem:
For the tests I'd recommend the following:

* You'll want a unit test [here](https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/configure/test_writer.py), probably around where the profile with double quotes is (adding one for single quotes as well as the spaces case) 
* I'd also recommend one [here](https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/configure/test_get.py) for making sure parsing works okay
* Would recommend something [here](https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/test_configure.py#L124) for again making sure spaces in the profile works

awscli/customizations/configure/writer.py
```diff
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import os
+import shlex
```
joguSD: This is not being used.

awscli/customizations/configure/writer.py
```diff
@@ -52,6 +53,10 @@ def update_config(self, new_values, config_filename):
 
         """
         section_name = new_values.pop('__section__', 'default')
+        parts = section_name.split(' ')
+        head, rest = parts[0], ' '.join(parts[1:])
+        if ' ' in rest:
+            section_name = "{0} \"{1}\"".format(head, rest)
```
joguSD: Is there any reason you went with double quotes here?

joguSD:
Unfortunately, I'm not sure this approach will work as is and would likely require changes to botocore as well.
Here's a sample of a case that fails:
`aws configure --profile 'my test profile'`
The result from the above command will be:
Config:
```
[profile "my test profile"]
```
Credentials:
```
[my "test profile"]
```

The credentials will fail to be found as it's looking for `[my test profile]` as the section header in the credentials file. This is the behavior botocore currently uses. It might be better to do this conversion one layer up in the configure customization code and quote the string there before it's passed to the config writer.
https://github.com/aws/aws-cli/blob/e2295b022db35eea9fec7e6c5540d06dbd6e588b/awscli/customizations/configure/configure.py#L117-L119

Once that's done we'll need tests to verify that both the profile and credentials can be read and updated for profile names that contain multiple spaces (and potentially only spaces). 
For example:
One space
```
~/.aws/config
[profile " "]
~/.aws/credentials
[ ]
```
Two spaces
```
~/.aws/config
[profile "  "]
~/.aws/credentials
[  ]
```
Three or more words:
```
~/.aws/config
[profile "foo bar baz"]
~/.aws/credentials
[foo bar baz]
```
Would all be valid profiles names.

joguSD:
Thanks for the fix @gtseres!

I refactored your change a little bit to allow us to use it with the `set` command as well.
Will go in this PR: https://github.com/aws/aws-cli/pull/2835

PR #2813, opened 2017-09-07
Change default to not include email on ecr login

Kgirthofer:
The default should be to not include default as this has been depreciated.


JordonPhillips:
Due to backwards compatibility concerns we won't be able to change this default for quite a while.

shimont:
This is an issue now, I cant use the suggested `eval $(aws ecr get-login --region us-east-1)` anymore

PR #2814, opened 2017-09-07
Document locale requirements

JordonPhillips:
This adds a section to the config docs calling out the need to configure
the system locale.


JordonPhillips:
@kyleknap I wanted to limit specifics as much as possible. People can google "set locale on windows" to find those answers. My worry is that if we're too specific we'll get things wrong, or be out of date.

kyleknap:
Alright. It may be something we want to consider adding in the future if people continue to run into problems even with this snippet. Otherwise :ship:

PR #2835, opened 2017-09-15
Make configure/set command properly handle spaces in profiles

### LINKED ISSUE #2806: Profile does not return region or output format if it contains a SPACE character

joguSD:
Fixes #2806
 A big thanks to @gtseres for submitting the fix.

I refactored his fix a little bit to make it a function so we can unit test it and apply it to the `set` command as well.

### DIFF
--- a/awscli/customizations/configure/__init__.py
+++ b/awscli/customizations/configure/__init__.py
@@ -10,8 +10,12 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import string
+from botocore.vendored.six.moves import shlex_quote
+
 NOT_SET = '<not set>'
 PREDEFINED_SECTION_NAMES = ('preview', 'plugins')
+_WHITESPACE = ' \t'
 
 
 class ConfigValue(object):
@@ -36,3 +40,10 @@ def mask_value(current_value):
         return 'None'
     else:
         return ('*' * 16) + current_value[-4:]
+
+
+def profile_to_section(profile_name):
+    """Converts a profile name to a section header to be used in the config."""
+    if any(c in _WHITESPACE for c in profile_name):
+        profile_name = shlex_quote(profile_name)
+    return 'profile %s' % profile_name
--- a/tests/unit/customizations/configure/test_configure.py
+++ b/tests/unit/customizations/configure/test_configure.py
@@ -14,6 +14,7 @@ import os
 import mock
 
 from awscli.customizations.configure import configure, ConfigValue, NOT_SET
+from awscli.customizations.configure import profile_to_section
 from awscli.testutils import unittest
 from awscli.compat import six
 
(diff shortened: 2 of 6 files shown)


awscli/customizations/configure/__init__.py
```diff
@@ -36,3 +36,13 @@ def mask_value(current_value):
         return 'None'
     else:
         return ('*' * 16) + current_value[-4:]
+
+
+def profile_to_section(profile_name):
+    """Converts a profile name to a section header to be used in the config."""
+
+    if ' ' in profile_name:
```
jamesls: It's probably better to use `string.whitespace` to account for tabs an other weird edge cases.

awscli/customizations/configure/__init__.py
```diff
@@ -36,3 +36,13 @@ def mask_value(current_value):
...
+        section_format = 'profile "%s"'
```
jamesls: Given botocore uses `shlex.quote` we should probably use the corresponding `six.moves.shlex_quote` to be safe.

tests/unit/customizations/configure/test_configure.py
```diff
@@ -238,6 +239,19 @@ def test_dont_mask_unset_value(self):
         self.assertEqual(no_config.value, NOT_SET)
 
 
+class TestProfileToSection(unittest.TestCase):
```
stealthycoin: It would be nice to add a few more tests for cases that aren't just a single space.

awscli/customizations/configure/__init__.py
joguSD: I'm okay with making this a little more robust but I'm concerned about what we view as valid whitespace in a section header. A profile name with a newline in it is probably invalid.

jamesls: It's more about being coordinated with what `shlex.split` uses in botocore.

joguSD:
Switched it over to use `string.whitespace` and `shlex_quote`.

joguSD:
Switched it over to use `string.whitespace` and `shlex_quote`.

tests/unit/customizations/configure/test_configure.py
jamesls: Did this ever get incorporated?  You mean a test with multiple consecutive space right?  It would be nice to get a test added for that.

jamesls:
Looks good, looks like this is just missing @stealthycoin's feedback right?

tests/unit/customizations/configure/test_configure.py
joguSD: He meant other valid white space in reference to your previous comment. I added the test for tabs as well. I can add a test for multiple spaces though.

joguSD:
All feedback should be addressed now.

benkehoe:
Note this fix is not present in `ConfigFileWriter` used by `aws configure sso` in v2 https://github.com/joguSD/aws-cli/blob/v2/awscli/customizations/configure/writer.py

PR #2335, opened 2016-12-20
Added support for forcing files to be zipped when uploading

dtimberlake:
When packaging lambda functions with `aws cloudformation package`, the code must be zipped even if it's a single file. This is in contrast to other resources (i.e. stack templates or api swagger files). The error given by CloudFormation when deploying a template specifying a single code file (after running the package command) is `Could not unzip uploaded file. Please check your file, then try to upload again.`

To accommodate this difference, I've added support for a `FORCE_ZIP` attribute on a `Resource` to determine if it's property value should always be zipped. It accomplishes this by copying the file into a temporary directory, zipping the directory, uploading the zipped file, then discarding the temporary directory.

I have added the new `FORCE_ZIP` attribute to `LambdaFunctionResource` and `ServerlessFunctionResource`.

PR #2866, opened 2017-09-29
added support for forcing files to be zipped when uploaded

sanathkr:
Taking over existing Pull Request #2335. Two differences:
- Don't zip again if local file is already a zip file.
- Adds more unit tests


jamesls:
Merged via https://github.com/aws/aws-cli/commit/a61229c9f1bb53e519f71499ab0cc315737ce199

PR #2868, opened 2017-10-03
Update integration tests for profiles with spaces

joguSD:
This fixes the integration tests when run on windows.

This fix seems strange to me. I'm not sure why we perform the following on windows:
https://github.com/aws/aws-cli/blob/develop/awscli/testutils.py#L598-L605
It makes sense to replace the single quotes, but not necessarily to escape the double quotes.


jamesls:
The fix looks fine but can you elaborate on why it doesn't make sense to escape double quotes?

joguSD:
The commands as they were in the original tests would function on both *nix and windows, escaping the double quotes only on windows is what broke them. Switching to single quotes (which won't work on windows) so that the `_escape_quotes` function can replace them to double quotes resulting in the same string that was there in the original tests seems strange.

PR #2888, opened 2017-10-11
Undocument deprecated operations

joguSD:
This will undocument deprecated operations removing them from the documentation as well as auto-completion in an effort to make these operations less discoverable. 

These changes depend on https://github.com/boto/botocore/pull/1292 being merged. So the tests might fail initially.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -425,6 +425,8 @@ class ServiceOperation(object):
         self._lineage = [self]
         self._operation_model = operation_model
         self._session = session
+        if operation_model.deprecated:
+            self._UNDOCUMENTED = True
 
     @property
     def name(self):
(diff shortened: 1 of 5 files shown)


awscli/clidriver.py
```diff
@@ -425,6 +425,8 @@ def __init__(self, name, parent_name, operation_caller,
         self._lineage = [self]
         self._operation_model = operation_model
         self._session = session
+        if operation_model and operation_model.deprecated:
```
jamesls: Remind me again when `operation_model` can be a False value?

joguSD: The majority of unit tests for this class pass `{}` or `None`. In practice this will never be a False value. Would you rather all the tests be updated to pass a mock?

kyleknap: That would be better. I would prefer a mock that uses the `OperationModel` class as the value to `spec`.

joguSD: Sounds good, I'll update the tests.

PR #2889, opened 2017-10-12
Fix typos.

boone:


tests/unit/output/test_json_output.py
```diff
@@ -89,7 +89,7 @@ def test_jmespath_json_response(self):
     def test_zero_value_is_printed(self):
         # Even though the integer 0 is false-like, we still
         # should be printing it to stdout if a jmespath query
-        # evalutes to 0.
+        # evaluates to 0.
```
jamesls: It looks like this file inadvertently changed its mode from 0644 to 0755.  Could you change it back?

boone: @jamesls good catch, sorry about that, not sure how it happened. Fixed in f1fbba61. Will squash.

boone:
Should be good to go, thanks!

stealthycoin:
@boone Sorry for the delay. Looking at this again I see a few more, specifically:
*  tests/functional/s3/test_ls_command.py
* tests/functional/s3api/test_get_object.py

If you can get all the files back to 644 and squash we should be good to go!

boone:
@stealthycoin thanks for letting me know, I have fixed those up.

kdaily:
Closing in lieu of #6426 which rebased and fixed merge conflicts. Thanks again for your contribution!

PR #2890, opened 2017-10-12
Add tagging functionality during S3 upload

DennisTT:
Add the ability to set tags on S3 uploads, moves, copies, similar to the existing metadata parameter.


DennisTT:
Tests seem to be failing because it also requires the change here: https://github.com/boto/s3transfer/pull/96

m6a-UdS:
+1 :-)

PR #2914, opened 2017-10-19
fix typo in emr helptext.py

Pahtoe:


Pahtoe:
creates a conflict, so closing

PR #2922, opened 2017-10-25
Add option to return zero rc for empty deploy changeset

JordonPhillips:
Closes #2606

This adds a flag to allow users to have deploy return a zero rc in
the event that their deploy changeset is empty. This is something
we would want to look at changing the default for on a major version
bump.

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -186,13 +186,17 @@ class DeployCommand(BasicCommand):
     def deploy(self, deployer, stack_name, template_str,
                parameters, capabilities, execute_changeset, role_arn,
                notification_arns):
-        result = deployer.create_and_wait_for_changeset(
-                stack_name=stack_name,
-                cfn_template=template_str,
-                parameter_values=parameters,
-                capabilities=capabilities,
-                role_arn=role_arn,
-                notification_arns=notification_arns)
+        try:
+            result = deployer.create_and_wait_for_changeset(
+                    stack_name=stack_name,
+                    cfn_template=template_str,
+                    parameter_values=parameters,
+                    capabilities=capabilities,
+                    role_arn=role_arn,
+                    notification_arns=notification_arns)
+        except exceptions.ChangeEmptyError as ex:
+            sys.stdout.write("%s\n" % ex)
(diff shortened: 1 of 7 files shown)


JordonPhillips:
@sanathkr Presumably this would be alright with you since you approved the original PR, but I'd like to give you a chance to look.

awscli/customizations/cloudformation/deploy.py
```diff
@@ -181,18 +206,25 @@ def _run_main(self, parsed_args, parsed_globals):
...
+                    stack_name=stack_name,
+                    cfn_template=template_str,
+                    parameter_values=parameters,
+                    capabilities=capabilities,
+                    role_arn=role_arn,
+                    notification_arns=notification_arns)
+        except exceptions.ChangeEmptyError as ex:
+            if fail_on_empty_changeset:
+                raise
+            sys.stdout.write("%s\n" % ex)
```
dstufft: This concerns me slightly because I feel like there's a good chance of drift happening between here and where other errors are handled that could be troublesome. One thing I can see off the bat is I believe that previously errors here would get rendered using:

```
LOG.debug("Exception caught in main()", exc_info=True)
LOG.debug("Exiting with rc 255")
err = get_stderr_text_writer()
err.write("\n")
err.write(six.text_type(e))
err.write("\n")
```

We can already see there is a difference here in terms of writing to stderr, and additional logging. Perhaps it would be better to introduce a new exception type (or return type?) that wraps the original exception and is used to indicate a non-error exception, but allow us to move the processing of writing this out back up to clidriver where we can use the standard routines?

I'm just throwing this idea out there though, I'm not entirely sure how I personally feel about it yet.

btisdall: I think this feature is pretty essential for use of `deploy` in CI context. Any update @JordonPhillips ?

awisswolf:
Is there a reason this hasn't been merged? It would be great to have.

btisdall:
BTW @awisswolf until this, #2940 and #2599 get sorted I'm using https://github.com/widdix/cfn-create-or-update

nskoufis:
For future reference, this is behind the flag `--no-fail-on-empty-changeset`

PR #2923, opened 2017-10-26
Fix list-runs pagination bug

JordonPhillips:
This fixes a bug in list-runs which caused it to only be able to
handle a max of 100 runs.

### DIFF
--- a/.changes/next-release/bugfix-codepipeline-66643.json
+++ b/.changes/next-release/bugfix-codepipeline-66643.json
@@ -1,5 +1,0 @@
-{
-  "description": "Fixed a bug in codepipeline where list-runs could only handle 100 runs.",
-  "type": "bugfix",
-  "category": "codepipeline"
-}
--- a/.changes/next-release/bugfix-codepipeline-66643.json
+++ b/.changes/next-release/bugfix-codepipeline-66643.json
@@ -1,0 +1,5 @@
+{
+  "description": "Fixed a bug in codepipeline where list-runs could only handle 100 runs.",
+  "type": "bugfix",
+  "category": "codepipeline"
+}
--- a/awscli/customizations/datapipeline/__init__.py
+++ b/awscli/customizations/datapipeline/__init__.py
@@ -50,6 +50,7 @@ are separated by a space. For list type parameter values
 you can use the same key name and specify each value as
 a key value pair. e.g. arrayValue=value1 arrayValue=value2
 """
+MAX_ITEMS_PER_DESCRIBE = 100
 
 
 class DocSectionNotFoundError(Exception):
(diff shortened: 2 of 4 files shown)


awscli/customizations/datapipeline/__init__.py
```diff
@@ -400,9 +399,18 @@ def _list_runs(self, parsed_args, parsed_globals):
...
     def _describe_objects(self, pipeline_id, object_ids):
-        parsed = self.client.describe_objects(
-            pipelineId=pipeline_id, objectIds=object_ids)
-        return parsed
+        # DescribeObjects will only accept 100 objectIds at a time,
+        # so we need to break up the list passed in into chunks that are at
+        # most that size. We then aggregate the results to return.
+        objects = []
+        max_items_per_describe = 100
+        for i in range(0, len(object_ids), max_items_per_describe):
```
dstufft: I think this is still going to lop items off the end isn't it? If there are 202 items, then this range would return: ``[0, 100, 200]``, missing the last two?

awscli/customizations/datapipeline/__init__.py
```diff
@@ -400,9 +399,18 @@ def _list_runs(self, parsed_args, parsed_globals):
...
 
     def _describe_objects(self, pipeline_id, object_ids):
-        parsed = self.client.describe_objects(
-            pipelineId=pipeline_id, objectIds=object_ids)
-        return parsed
+        # DescribeObjects will only accept 100 objectIds at a time,
+        # so we need to break up the list passed in into chunks that are at
+        # most that size. We then aggregate the results to return.
+        objects = []
+        max_items_per_describe = 100
```
dstufft: It's normally customary to put constants at module scope with ``ALL_CAPS_NAMES``.

awscli/customizations/datapipeline/__init__.py
JordonPhillips: That would result in three requests: `0-99, 100-199, 200-202`

JordonPhillips: The tests should cover this since i'm testing a set of size 150.

dstufft: Ah yes, you're correct. I had the logic switched around in my head, thinking the range was returning the upper limit not the lower limit of the object IDs.

.changes/next-release/bugfix-codepipeline-66643.json
```diff
@@ -0,0 +1,5 @@
+{
+  "description": "Fixed a bug in codepipeline where list-runs could only handle 100 runs.",
```
jamesls: Data pipeline

.changes/next-release/bugfix-codepipeline-66643.json
```diff
@@ -0,0 +1,5 @@
...
+  "type": "bugfix",
+  "category": "codepipeline"
```
jamesls: datapipeline

Issue #2948, opened 2017-11-09
"aws cloudformation (deploy|wait)": describe-stack polling gets AWS throttling API errors


powersjcb:
The AWS API throttles requests after what appears to be a set number of API calls. We recently switched several workflows to heavily rely on `aws cloudformation (deploy|wait)`. These operations are consistently being throttled during peak usage from an excess number of `describe-stack` API requests.

As an alternative to AWS support limbo, waiting for a throttling increase, it would be nice to manually reduce throughput of waiters for tasks users know will take more time.

The failure:
`An error occurred (Throttling) when calling the DescribeStacks operation (reached max retries: 4): Rate exceeded`

Related ticket #1376

joguSD:
This seems reasonable marking as a feature request.
@sanathkr

sanathkr:
Yeah this is actually a good feature request. I will comment on the PR

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

sbe-arg:
IMO

aws cloudformation deploy | wait backend Describe commands should be by default unlimited and handling rate limit properly unless explicitly --set to stop at X count.

sbe-arg:
cross reference https://github.com/aws/aws-cli/issues/3665#issuecomment-635902479

tim-finnigan:
[...]

PR #2949, opened 2017-11-09
issues/2948: use variable waiter_config

powersjcb:
helps to alleviate #2948


oleyka:
Perhaps, `AWS_CF_WAIT_DELAY_*` would be better names for these envvars.

sanathkr:
I like the feature, but not a big fan of env-vars spread out deep inside the files. Can we add a real parameter instead and pass it down? That way is easy to discover, well documented & tested.

stealthycoin:
@powersjcb any interest in updating this PR per @sanathkr's feedback?

JoeMcGuire:
+1 for this feature. While we wait for it I've requested an API rate increase.

PR #2968, opened 2017-11-17
Add database abstraction for history recorder

stealthycoin:
Adds database connection and interaction abstractions as well as a
handler that can hook into botocore's HistoryRecorder
 * DatabaseConnection
 * DatabaseRecordReader
 * DatabaseRecordWriter
 * PayloadSerializer
 * DatabaseHistoryHandler

cc @jamesls @kyleknap @dstufft @JordonPhillips @joguSD

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -37,6 +37,12 @@ except ImportError:
     ZIP_COMPRESSION_MODE = zipfile.ZIP_STORED
 
 
+try:
+    import sqlite3
+except ImportError:
+    sqlite3 = None
+
+
 class NonTranslatedStdout(object):
     """ This context manager sets the line-end translation mode for stdout.
 
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -26,6 +26,7 @@ queue = six.moves.queue
 shlex_quote = six.moves.shlex_quote
 StringIO = six.StringIO
 urlopen = six.moves.urllib.request.urlopen
+binary_type = six.binary_type
 
 # Most, but not all, python installations will have zlib. This is required to
 # compress any files we send via a push. If we can't compress, we can still
--- a/awscli/customizations/history/db.py
+++ b/awscli/customizations/history/db.py
@@ -88,8 +88,6 @@ class DatabaseConnection(object):
 
 
 class PayloadSerializer(json.JSONEncoder):
-    _DEFAULT_TYPES = set(['str', 'int', 'list', 'dict', 'float'])
-
     def _process_CaseInsensitiveDict(self, obj, type_name):
         return dict(obj)
 
@@ -120,8 +118,6 @@ class PayloadSerializer(json.JSONEncoder):
 
     def default(self, obj):
         type_name = type(obj).__name__
-        if type_name in self._DEFAULT_TYPES:
-            return obj
         return getattr(
             self, '_process_%s' % type_name, self._unknown
         )(obj, type_name)
--- a/awscli/customizations/history/db.py
+++ b/awscli/customizations/history/db.py
@@ -22,6 +22,7 @@ from collections import MutableMapping
 from botocore.history import BaseHistoryHandler
 
 from awscli.compat import sqlite3
+from awscli.compat import ensure_text_type
 
 
 LOG = logging.getLogger('awscli.history.db')
@@ -211,7 +212,7 @@ class RecordBuilder(object):
         # The payload body key here is in bytes in python 3 and needs to be
         # decoded.
         if event_type in self._BYTES_BODY_PAYLOADS:
-            payload['body'] = payload['body'].decode('utf-8')
+            payload['body'] = ensure_text_type(payload['body'])
         return payload
 
     def build_record(self, event_type, payload, source):
--- a/awscli/customizations/history/db.py
+++ b/awscli/customizations/history/db.py
@@ -117,7 +117,6 @@ class DatabaseRecordWriter(object):
         if connection is None:
             connection = DatabaseConnection()
         self._connection = connection
-        self._identifier = None
 
     def write_record(self, record):
         # This method is not threadsafe by itself, it is only threadsafe when
@@ -126,18 +125,12 @@ class DatabaseRecordWriter(object):
         db_record = self._create_db_record(record)
         self._connection.execute(self._WRITE_RECORD, db_record)
 
-    def _get_identifier(self):
-        if self._identifier is None:
-            self._identifier = str(uuid.uuid4())
-        return self._identifier
-
     def _create_db_record(self, record):
-        uid = self._get_identifier()
         event_type = record['event_type']
         json_serialized_payload = json.dumps(record['payload'],
                                              cls=PayloadSerializer)
         db_record = (
-            uid,
+            record['command_id'],
             record.get('request_id'),
             record['source'],
             event_type,
@@ -191,6 +184,7 @@ class RecordBuilder(object):
     _BYTES_BODY_PAYLOADS = set(['HTTP_REQUEST', 'HTTP_RESPONSE'])
 
     def __init__(self):
+        self._identifier = None
         self._locals = threading.local()
 
     def _get_current_thread_request_id(self):
@@ -215,9 +209,16 @@ class RecordBuilder(object):
             payload['body'] = ensure_text_type(payload['body'])
         return payload
 
+    def _get_identifier(self):
+        if self._identifier is None:
+            self._identifier = str(uuid.uuid4())
+        return self._identifier
+
     def build_record(self, event_type, payload, source):
+        uid = self._get_identifier()
         payload = self._format_payload(event_type, payload)
         record = {
+            'command_id': uid,
             'event_type': event_type,
             'payload': payload,
             'source': source,
--- a/awscli/customizations/history/db.py
+++ b/awscli/customizations/history/db.py
@@ -1,0 +1,223 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
+import uuid
+import time
+import json
+import base64
+import threading
+
+from botocore.history import BaseHistoryHandler
+
+from awscli.compat import sqlite3
+from awscli.compat import ensure_text_type
+
+
+class DatabaseConnection(object):
+    _CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS records (
+      id TEXT,
+      request_id TEXT,
+      source TEXT,
+      event_type TEXT,
+      timestamp INTEGER,
+      payload TEXT
+    )"""
+    _ENABLE_WAL = """PRAGMA journal_mode=WAL"""
+    _DATABASE_NAME = 'history.db'
+    _DEFAULT_DATABASE_FILENAME = os.path.expanduser(
+        os.path.join('~', '.aws', 'cli', 'history', _DATABASE_NAME))
+
+    def __init__(self, db_filename=None):
+        if db_filename:
+            self._db_filename = db_filename
+        elif os.environ.get('AWS_HISTORY_PATH'):
+            self._db_filename = os.path.join(
+                os.environ.get('AWS_HISTORY_PATH'), self._DATABASE_NAME)
+        else:
+            self._db_filename = self._DEFAULT_DATABASE_FILENAME
+
+        if not os.path.isdir(os.path.dirname(self._db_filename)):
+            os.makedirs(os.path.dirname(self._db_filename))
+        self._connection = sqlite3.connect(
+            self._db_filename, check_same_thread=False)
+        self._ensure_database_setup()
+
+    def _ensure_database_setup(self):
+        self._create_record_table()
+        self._try_to_enable_wal()
+
+    def _create_record_table(self):
+        cursor = self._connection.cursor()
+        cursor.execute(self._CREATE_TABLE)
+        self._connection.commit()
+
+    def _try_to_enable_wal(self):
+        try:
+            cursor = self._connection.cursor()
+            cursor.execute(self._ENABLE_WAL)
+            self._connection.commit()
+        except sqlite3.Error:
+            # This is just a performance enhancement so it is optional. Not all
+            # systems will have a sqlite compiled with the WAL enabled.
+            pass
+
+    def cursor(self):
+        return self._connection.cursor()
+
+    def commit(self):
+        self._connection.commit()
+
+    @property
+    def row_factory(self):
+        return self._connection.row_factory
+
+    @row_factory.setter
+    def row_factory(self, row_factory):
+        self._connection.row_factory = row_factory
+
+
+class PayloadSerializer(json.JSONEncoder):
+    DEFAULT_TYPES = set(['str', 'int', 'list', 'dict', 'float'])
+
+    def _process_CaseInsensitiveDict(self, obj, type_name):
+        return dict(obj)
+
+    def _process_bytes(self, obj, type_name):
+        encoded = base64.b64encode(obj)
+        return ensure_text_type(encoded)
+
+    def _process_timestamp(self, obj, type_name):
+        return obj.isoformat()
+
+    def _unknown(self, obj, type_name):
+        return type_name
+
+    def default(self, obj):
+        type_name = type(obj).__name__
+        if type_name in self.DEFAULT_TYPES:
+            return obj
+        return getattr(
+            self, '_process_%s' % type_name, self._unknown
+        )(obj, type_name)
+
+
+class DatabaseRecordWriter(object):
+    _WRITE_RECORD = """
+    INSERT INTO records(id, request_id, source, event_type, timestamp, payload)
+    VALUES (?,?,?,?,?,?)"""
+    _REQUEST_LIFECYCLE_EVENTS = set(
+        ['API_CALL', 'HTTP_REQUEST', 'HTTP_RESPONSE', 'PARSED_RESPONSE'])
+    _START_OF_REQUEST_LIFECYCLE_EVENT = 'API_CALL'
+
+    def __init__(self, connection=None):
+        if connection is None:
+            connection = DatabaseConnection()
+        self._connection = connection
+        self._request_ids = {}
+        self._identifier = None
+
+    def write_record(self, record):
+        cursor = self._connection.cursor()
+        db_record = self._create_db_record(record)
+        cursor.execute(self._WRITE_RECORD, db_record)
+        self._connection.commit()
+
+    def _get_request_id(self, event_type):
+        if event_type == self._START_OF_REQUEST_LIFECYCLE_EVENT:
+            self._start_http_lifecycle()
+        if event_type in self._REQUEST_LIFECYCLE_EVENTS:
+            thread_id = threading.current_thread().ident
+            request_id = self._request_ids.get(thread_id, '')
+        else:
+            request_id = ''
+        return request_id
+
+    def _start_http_lifecycle(self):
+        thread_id = threading.current_thread().ident
+        new_identifier = str(uuid.uuid4())
+        self._request_ids[thread_id] = new_identifier
+
+    def _get_identifier(self):
+        if self._identifier is None:
+            self._identifier = str(uuid.uuid4())
+        return self._identifier
+
+    def _create_db_record(self, record):
+        uid = self._get_identifier()
+        event_type = record['event_type']
+        request_id = self._get_request_id(event_type)
+        json_serialized_payload = json.dumps(record['payload'],
+                                             cls=PayloadSerializer)
+        db_record = (
+            uid,
+            request_id,
+            record['source'],
+            event_type,
+            record['timestamp'],
+            json_serialized_payload
+        )
+        return db_record
+
+
+class DatabaseRecordReader(object):
+    _ORDERING = 'ORDER BY timestamp'
+    _GET_LAST_ID_RECORDS = """
+    SELECT * FROM records
+    WHERE id =
+    (SELECT id FROM records WHERE timestamp =
+    (SELECT max(timestamp) FROM records)) %s;""" % _ORDERING
+    _GET_RECORDS_BY_ID = 'SELECT * from records where id = ? %s' % _ORDERING
+
+    def __init__(self, connection=None):
+        if connection is None:
+            connection = DatabaseConnection()
+        self._connection = connection
+        self._connection.row_factory = self._row_factory
+
+    def _row_factory(self, cursor, row):
+        d = {}
+        for idx, col in enumerate(cursor.description):
+            val = row[idx]
+            if col[0] == 'payload':
+                val = json.loads(val)
+            d[col[0]] = val
+        return d
+
+    def yield_latest_records(self):
+        cursor = self._connection.cursor()
+        cursor.execute(self._GET_LAST_ID_RECORDS)
+        for row in cursor:
+            yield row
+
+    def yield_records(self, record_id):
+        cursor = self._connection.cursor()
+        cursor.execute(self._GET_RECORDS_BY_ID, [record_id])
+        for row in cursor:
+            yield row
+
+
+class DatabaseHistoryHandler(BaseHistoryHandler):
+    def __init__(self, writer=None):
+        if writer is None:
+            writer = DatabaseRecordWriter()
+        self._writer = writer
+
+    def emit(self, event_type, payload, source):
+        record = {
+            'event_type': event_type,
+            'payload': payload,
+            'source': source,
+            'timestamp': int(time.time() * 1000)
+        }
+        self._writer.write_record(record)
--- a/awscli/customizations/history/db.py
+++ b/awscli/customizations/history/db.py
@@ -15,12 +15,19 @@ import uuid
 import time
 import json
 import base64
+import datetime
 import threading
+import logging
+from collections import MutableMapping
 
 from botocore.history import BaseHistoryHandler
 
 from awscli.compat import sqlite3
 from awscli.compat import ensure_text_type
+from awscli.compat import binary_type
+
+
+LOG = logging.getLogger('awscli.history.db')
 
 
 class DatabaseConnection(object):
@@ -88,14 +95,22 @@ class DatabaseConnection(object):
 
 
 class PayloadSerializer(json.JSONEncoder):
-    def _process_CaseInsensitiveDict(self, obj, type_name):
+    def _encode_mutable_mapping(self, obj):
         return dict(obj)
 
-    def _process_bytes(self, obj, type_name):
-        encoded = base64.b64encode(obj)
-        return ensure_text_type(encoded)
+    def _encode_binary(self, obj):
+        # On PY3 a bytes like object that contains entirely valid utf-8 bytes
+        # will be routed here instead of the encode method. In the case of
+        # utf-8 bytes we want to make the stored data as readble as possible so
+        # first we try to decode it as utf-8, if that fails we will fall back
+        # to using a base64 encoded string.
+        try:
+            encoded = obj.decode('utf-8')
+        except UnicodeDecodeError:
+            encoded = ensure_text_type(base64.b64encode(obj))
+        return encoded
 
-    def _process_datetime(self, obj, type_name):
+    def _encode_datetime(self, obj):
         return obj.isoformat()
 
     def _unknown(self, obj, type_name):
@@ -103,7 +118,7 @@ class PayloadSerializer(json.JSONEncoder):
 
     def encode(self, obj):
         # The default method is not called in PY2 where the JSONEncoder thinks
-        # it can handle a bytes object becausae it can't tell it apart from a
+        # it can handle a bytes object because it can't tell it apart from a
         # str object. In PY3 this is handled by the _process_bytes method.
         # For PY2 we have to override the encode method where the encoding of
         # a bytes-like string will fail, catch the UnicodeDecodeError and
@@ -112,15 +127,19 @@ class PayloadSerializer(json.JSONEncoder):
             encoded = super(PayloadSerializer, self).encode(obj)
             return encoded
         except UnicodeDecodeError:
-            b64encoded = self._process_bytes(obj, 'bytes')
+            b64encoded = self._process_bytes(obj)
             encoded = self.encode(b64encoded)
             return encoded
 
     def default(self, obj):
-        type_name = type(obj).__name__
-        return getattr(
-            self, '_process_%s' % type_name, self._unknown
-        )(obj, type_name)
+        if isinstance(obj, binary_type):
+            return self._encode_binary(obj)
+        elif isinstance(obj, datetime.datetime):
+            return self._encode_datetime(obj)
+        elif isinstance(obj, MutableMapping):
+            return self._encode_mutable_mapping(obj)
+        else:
+            return type(obj).__name__
 
 
 class DatabaseRecordWriter(object):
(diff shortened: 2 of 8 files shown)


tests/functional/history/test_db.py
```diff
@@ -0,0 +1,199 @@
...
+            {
+                'source': 'TEST',
+                'event_type': 'zip',
+                'payload': '',
+                'timestamp': 6
+            }
+        ])
+        reader = DatabaseRecordReader(self.connection)
+        cursor = self._get_cursor()
+        cursor.execute(
```
kyleknap: I wonder if it makes sense to expose the ability to supply your own id to the writer intializer instead of having to do this? Then you can test for the other id pretty easily as well.

tests/unit/customizations/history/test_db.py
```diff
@@ -0,0 +1,711 @@
...
+class TestDatabaseRecordReader(BaseDatabaseRecordTester):
+    def setUp(self):
+        self.fake_connection = FakeDatabaseConnection()
+        self.reader = DatabaseRecordReader(self.fake_connection)
+
+    def test_row_factory_set(self):
+        self.assertEqual(self.fake_connection.row_factory,
+                         self.reader._row_factory)
+
+    def test_yield_latest_records(self):
```
kyleknap: We might want to flesh these out.

tests/unit/customizations/history/test_db.py
```diff
@@ -0,0 +1,711 @@
...
+        reloaded = json.loads(string_value)
+        self.assertEqual(original, reloaded)
+
+    def test_can_serialize_datetime(self):
+        now = datetime.datetime.now()
+        iso_now = now.isoformat()
+        string_value = json.dumps(now, cls=PayloadSerializer)
+        reloaded = json.loads(string_value)
+        self.assertEqual(iso_now, reloaded)
+
```
kyleknap: Should we add a test for the unknown types?

tests/unit/customizations/history/test_db.py
```diff
@@ -0,0 +1,711 @@
...
+
+class ThreadedRecordWriter(object):
+    def __init__(self, writer, fake_connection):
+        self._read_q = queue.Queue()
+        self._write_q = queue.Queue()
+        self._thread = threading.Thread(
+            target=self._threaded_record_writer,
+            args=(writer, fake_connection))
+        self._thread.start()
```
kyleknap: I would break this out to its own separate `start()` method. It is a little strange to spin up the actual thread in the initializer.

tests/unit/customizations/history/test_db.py
```diff
@@ -0,0 +1,711 @@
...
+        identifier, request_id = written_record[:2]
+        self.assertTrue(self.UUID_PATTERN.match(identifier))
+        self.assertTrue(self.UUID_PATTERN.match(request_id))
+
+    def test_can_write_http_request_record(self):
+        insert_stmt, written_record = self._write_record({
+            'event_type': 'HTTP_REQUEST',
+            'payload': {
+                'method': 'GET',
+                'headers': {},
```
kyleknap: Should this use the `CaseInsensitiveDict` class?

tests/unit/customizations/history/test_db.py
```diff
@@ -0,0 +1,711 @@
...
+                'streaming': False,
+                'headers': {},
+                'body': '...',
+                'status_code': 200,
+                'request_id': '1234abcd'
+            }))
+        )
+        identifier, request_id = written_record[:2]
+        self.assertTrue(self.UUID_PATTERN.match(identifier))
+
```
kyleknap: Do we want to assert for uid patttern on request id's?

tests/functional/history/test_db.py
```diff
@@ -0,0 +1,199 @@
...
+        ])
+
+        # Since the foo and bar events were written by the writer_a they all
+        # share an id. foo was written at time 3 which makes it the most
+        # recent, so when we call get_latest_records we should get the
+        # foo and bar records only.
+        reader = DatabaseRecordReader(self.connection)
+        records = set([record['event_type'] for record
+                       in reader.yield_latest_records()])
+        self.assertEqual(set(['foo', 'bar']), records)
```
kyleknap: Should we add tests to make sure the `yield_latest_records` and `yield_records` methods return items by timestamps?

tests/functional/history/test_db.py
```diff
@@ -0,0 +1,199 @@
...
+        }
+        records_to_write = 40
+        for _ in range(records_to_write):
+            writer.write_record(known_record_fields)
+
+        cursor = self._get_cursor()
+        cursor.execute("SELECT COUNT(*) FROM records;")
+        num_records = cursor.fetchone()
+        self.assertEqual(num_records[0], records_to_write)
+
```
kyleknap: I wonder if it makes sense to put some thread property based testing here in the functional tests. The idea is you would use a single db connection behind a lock and then have threads write to the db and at the end you would take the reader to make sure that all of the records were there and there was X distrinct request id's where X would be the number of http lifecycles multiplied by number of threads. Furthermore, you could do some more testing to make sure that when if you group them together by request id's all of the records are as expected.

tests/unit/customizations/history/test_db.py
stealthycoin: Yep.

tests/unit/customizations/history/test_db.py
stealthycoin: Missed these somehow.

tests/unit/customizations/history/test_db.py
stealthycoin: Yea I think also each test for threads may want a different number of threads so I think ill add a `_spin_up_n_threads()` type function that collects them all into a list which is joined on in the `tearDown()`

tests/unit/customizations/history/test_db.py
stealthycoin: I can't unless there was a prior API_CALL made. I can either change that logic or have an API_CALL be made first for all these tests.

tests/unit/customizations/history/test_db.py
stealthycoin: yes

tests/functional/history/test_db.py
stealthycoin: Yea this was a little annoying but I didn't see any reason to ever need that outside tests so I left it as is. In the event that is ever used in practice it could easy lead to colliding ids in a database which means you would start picking up records that didn't belong together.

tests/functional/history/test_db.py
stealthycoin: Yea that sounds good to me. I only have the one thread related test so there could definitely be a lot more.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+        return request_id
+
+    def _start_http_lifecycle(self):
+        thread_id = threading.current_thread().ident
+        new_identifier = str(uuid.uuid4())
+        self._request_ids[thread_id] = new_identifier
+
+    def _get_identifier(self):
+        if self._identifier is None:
+            self._identifier = str(uuid.uuid4())
```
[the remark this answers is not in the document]
stealthycoin: Yes my thought is that if we ever want to introduce multiple uids for some reason we can make multiple writers.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+        self._connection.commit()
+
+    def _get_request_id(self, event_type):
+        if event_type == self._START_OF_REQUEST_LIFECYCLE_EVENT:
+            self._start_http_lifecycle()
+        if event_type in self._REQUEST_LIFECYCLE_EVENTS:
+            thread_id = threading.current_thread().ident
+            request_id = self._request_ids.get(thread_id, '')
+        else:
+            request_id = ''
```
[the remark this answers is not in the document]
stealthycoin: Like put a None so its a null value in the database rather than an empty string? That seems reasonable.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+        db_record = self._create_db_record(record)
+        cursor.execute(self._WRITE_RECORD, db_record)
+        self._connection.commit()
+
+    def _get_request_id(self, event_type):
+        if event_type == self._START_OF_REQUEST_LIFECYCLE_EVENT:
+            self._start_http_lifecycle()
+        if event_type in self._REQUEST_LIFECYCLE_EVENTS:
+            thread_id = threading.current_thread().ident
+            request_id = self._request_ids.get(thread_id, '')
```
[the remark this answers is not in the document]
stealthycoin: Just for testing if I insert an HTTP record before inserting an API_CALL record then there will be no thread id. This is the same reason I do not check the thread_id in some of those tests, only in the ones where an API_CALL is inserted first.

I can either:
* Make it a direct access with [] and remove all tests that just insert a solitary http related record without first inserting an API_CALL record.
* Leave it
* Go back to what I was originally doing where if there is no request_id for an http record insertion it generates a new one then uses that one. Leaving API_CALL with the same behavior where it overwrites an existing request_id.

In any case the real thing won't use either behavior that is being used for tests so I don't really like either 1 or 3.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+    _START_OF_REQUEST_LIFECYCLE_EVENT = 'API_CALL'
+
+    def __init__(self, connection=None):
+        if connection is None:
+            connection = DatabaseConnection()
+        self._connection = connection
+        self._request_ids = {}
+        self._identifier = None
+
+    def write_record(self, record):
```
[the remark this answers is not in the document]
stealthycoin: Good call.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+            return encoded
+        except UnicodeDecodeError:
+            b64encoded = self._process_bytes(obj, 'bytes')
+            encoded = self.encode(b64encoded)
+            return encoded
+
+    def default(self, obj):
+        type_name = type(obj).__name__
+        return getattr(
+            self, '_process_%s' % type_name, self._unknown
```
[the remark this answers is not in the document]
stealthycoin: Yea that makes more sense.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,234 @@
...
+        # it can handle a bytes object becausae it can't tell it apart from a
+        # str object. In PY3 this is handled by the _process_bytes method.
+        # For PY2 we have to override the encode method where the encoding of
+        # a bytes-like string will fail, catch the UnicodeDecodeError and
+        # then call _process_bytes and retry the encoding.
+        try:
+            encoded = super(PayloadSerializer, self).encode(obj)
+            return encoded
+        except UnicodeDecodeError:
+            b64encoded = self._process_bytes(obj, 'bytes')
```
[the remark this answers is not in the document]
stealthycoin: No it only will encode the bytes object as base64.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def _try_to_enable_wal(self):
+        try:
+            cursor = self.cursor()
+            cursor.execute(self._ENABLE_WAL)
+            self._connection.commit()
+        except sqlite3.Error:
+            # This is just a performance enhancement so it is optional. Not all
+            # systems will have a sqlite compiled with the WAL enabled.
+            pass
```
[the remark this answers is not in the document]
stealthycoin: I can add debug logs yea.

awscli/customizations/history/db.py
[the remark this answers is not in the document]
kyleknap: Ok that is good then. Not sure if there is a test for it, but it would be good to make sure it gets caught.

awscli/customizations/history/db.py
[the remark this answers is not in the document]
kyleknap: Yep that is what I was thinking. I always feel like replacing empty string for None is a little ambiguous.

awscli/customizations/history/db.py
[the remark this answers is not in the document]
kyleknap: I think I am fine keeping it. It will also be more robust if there was an issue in the future where for whatever reason we did not previously put a request id. Like we do not want to lose the whole record.

tests/functional/history/test_db.py
kyleknap: Yeah that makes sense. It was just a thought I had when I was looking through the tests.

tests/unit/customizations/history/test_db.py
kyleknap: Ah I think we should just then at least for one of the `API_CALL` tests there is a pattern matching assertion for request_id if it is not already there.

awscli/compat.py
```diff
@@ -37,6 +37,12 @@
     ZIP_COMPRESSION_MODE = zipfile.ZIP_STORED
 
 
+try:
+    import sqlite3
+except ImportError:
+    sqlite3 = None
```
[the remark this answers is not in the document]
jamesls: This will happen before debug logs are configured though so I don't think it'd ever show up anywhere.  Plus you log if sqlite3 isn't available when you try to enable the customization.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+    )"""
+    _ENABLE_WAL = """PRAGMA journal_mode=WAL"""
+    _DATABASE_NAME = 'history.db'
+    _DEFAULT_DATABASE_FILENAME = os.path.expanduser(
+        os.path.join('~', '.aws', 'cli', 'history', _DATABASE_NAME))
+
+    def __init__(self, db_filename=None):
+        if db_filename:
+            self._db_filename = db_filename
+        elif os.environ.get('AWS_HISTORY_PATH'):
```
jamesls: I don't believe this was discussed.  I'd prefer to not add this env var for now.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+      event_type TEXT,
+      timestamp INTEGER,
+      payload TEXT
+    )"""
+    _ENABLE_WAL = """PRAGMA journal_mode=WAL"""
+    _DATABASE_NAME = 'history.db'
+    _DEFAULT_DATABASE_FILENAME = os.path.expanduser(
+        os.path.join('~', '.aws', 'cli', 'history', _DATABASE_NAME))
+
+    def __init__(self, db_filename=None):
```
jamesls: You can default to `_DEFAULT_DATABASE_FILENAME` if we're not going to support the `AWS_HISTORY_PATH` env var.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def default(self, obj):
+        if isinstance(obj, binary_type):
+            return self._encode_binary(obj)
+        elif isinstance(obj, datetime.datetime):
+            return self._encode_datetime(obj)
+        elif isinstance(obj, MutableMapping):
+            return self._encode_mutable_mapping(obj)
+        else:
+            return type(obj).__name__
```
jamesls: Remind me again why not the repr?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+    @row_factory.setter
+    def row_factory(self, row_factory):
+        self._connection.row_factory = row_factory
+
+
+class PayloadSerializer(json.JSONEncoder):
+    def _encode_mutable_mapping(self, obj):
+        return dict(obj)
+
+    def _encode_binary(self, obj):
```
jamesls: This seems hacky.  What were the specific history records that resulted in needing this method?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+            self._db_filename = db_filename
+        elif os.environ.get('AWS_HISTORY_PATH'):
+            self._db_filename = os.path.join(
+                os.environ.get('AWS_HISTORY_PATH'), self._DATABASE_NAME)
+        else:
+            self._db_filename = self._DEFAULT_DATABASE_FILENAME
+
+        if not os.path.isdir(os.path.dirname(self._db_filename)):
+            os.makedirs(os.path.dirname(self._db_filename))
+        self._connection = sqlite3.connect(
```
jamesls: From what I can tell we're committing after every sql statement so I'd prefer to simplify things by using using auto commit for the isolation level and not bothering with cursors/commits.  This code is already heavily coupled to sqlite3 anyways.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+            request_id,
+            record['source'],
+            event_type,
+            record['timestamp'],
+            json_serialized_payload
+        )
+        return db_record
+
+
+class DatabaseRecordReader(object):
```
jamesls: Where is the b64decode happening that undoes the b64encoding from the Writer?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+        cursor = self._connection.cursor()
+        db_record = self._create_db_record(record)
+        cursor.execute(self._WRITE_RECORD, db_record)
+        self._connection.commit()
+
+    def _get_request_id(self, event_type):
+        if event_type == self._START_OF_REQUEST_LIFECYCLE_EVENT:
+            self._start_http_lifecycle()
+        if event_type in self._REQUEST_LIFECYCLE_EVENTS:
+            thread_id = threading.current_thread().ident
```
jamesls: tls is the typical way to handle this.  Is there a reason you're not using that?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+    def write_record(self, record):
+        # This method is not threadsafe by itself, it is only threadsafe when
+        # used inside a handler bound to the HistoryRecorder in botocore which
+        # is protected by a lock.
+        cursor = self._connection.cursor()
+        db_record = self._create_db_record(record)
+        cursor.execute(self._WRITE_RECORD, db_record)
+        self._connection.commit()
+
+    def _get_request_id(self, event_type):
```
jamesls: This seems like the wrong level of abstraction to track this at.  I wouldn't expect a database record writer to know anything about request lifecycle events.

I would expect:

* Handler - bridge between history records and the backend storage
* DatabaseWriter - munges the records so they can be stored in the database (payload serialization, json encoding, etc).  Coordinate with:
* DBConnection - low level details of talking to sqlite3, dealing with transactions, etc.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,256 @@
...
+    def _row_factory(self, cursor, row):
+        d = {}
+        for idx, col in enumerate(cursor.description):
+            val = row[idx]
+            if col[0] == 'payload':
+                val = json.loads(val)
+            d[col[0]] = val
+        return d
+
+    def yield_latest_records(self):
```
jamesls: Small nitpick, can you change all the `yield_` to `iter_`?

awscli/customizations/history/db.py
stealthycoin: I had it that way initially I changed it due to feedback. I can put it back if you'd prefer. The docs for `ident` say its intended use was for keying a dictionary off of threads so I felt it wasn't too gross this way.

awscli/customizations/history/db.py
stealthycoin: My thought here was this is actually somewhat a property of the database given that it has its own column, and so the record writer that writes this particular schema already has to have knowledge about it separate from the simple write payload in a row logic. I could do this logic in botocore and put it as part of the payload.

tests/functional/history/test_db.py
```diff
@@ -0,0 +1,311 @@
...
+        cursor.execute("SELECT * FROM records;")
+        record = dict(cursor.fetchone())
+        for col_name, row_value in known_record_fields.items():
+            # Normally our reader would take care of parsing the JSON from
+            # the payload.
+            if col_name == 'payload':
+                record[col_name] = json.loads(record[col_name])
+                self.assertEqual(record[col_name], row_value)
+
+        self.assertIn('id', record.keys())
```
kyleknap: Usually you do not have to use `keys()`. `in` for a dict means to look in the keys for it already.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,244 @@
...
+            yield row
+
+    def iter_records(self, record_id):
+        cursor = self._connection.cursor()
+        cursor.execute(self._GET_RECORDS_BY_ID, [record_id])
+        for row in cursor:
+            yield row
+
+
+class RecordBuilder(object):
```
kyleknap: Did not see really a test class for the `RecordBuilder`. Do you think it makes sense to add some specifically related to making sure everything is formatted correctly, it handles payloads properly, and injects timestamps? It looks like you really have a lot of logic related to request_id's but not so much the other aspects of it.

tests/unit/customizations/history/test_db.py
```diff
@@ -617,6 +617,20 @@ def _get_request_id_for_event_type(self, builder, event_type):
         record = builder.build_record(event_type, {'body': b''}, '')
         return record.get('request_id')
 
+    def test_can_process_http_request_with_no_body(self):
+        builder = RecordBuilder()
+        try:
+            builder.build_record('HTTP_REQUEST', {}, '')
```
kyleknap: So it is not guaranteed that a body will be available? I think in my code I assume at least the `'body'` key will be present but maybe set as `None`

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+from awscli.compat import sqlite3
+from awscli.compat import ensure_text_type
+
+
+LOG = logging.getLogger('awscli.history.db')
+
+
+class DatabaseConnection(object):
+    _CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS records (
```
jamesls: Can you indent this?  It's easier to read if these don't align with class variables indentation level

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+      id TEXT,
+      request_id TEXT,
+      source TEXT,
+      event_type TEXT,
+      timestamp INTEGER,
+      payload TEXT
+    )"""
+    _ENABLE_WAL = """PRAGMA journal_mode=WAL"""
```
jamesls: Don't need triple quotes here.

awscli/customizations/history/db.py
jamesls: Yes, let's remove this from this PR and just default `db_filename` to `_DEFAULT_DATABASE_FILENAME` for now.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+    def __init__(self, db_filename=None):
+        if db_filename:
+            self._db_filename = db_filename
+        elif os.environ.get('AWS_HISTORY_PATH'):
+            self._db_filename = os.path.join(
+                os.environ.get('AWS_HISTORY_PATH'), self._DATABASE_NAME)
+        else:
+            self._db_filename = self._DEFAULT_DATABASE_FILENAME
+
+        if self._db_filename != ':memory:' and \
```
jamesls: If you have to have hard coded string checks like this, it's a pretty good indication that this is happening at the wrong level of abstraction.  I would prefer you move the makedirs stuff out of the db and let the db just deal with talking to sqlite3.  The thing that accepts a db_filename needs to ensure that this directory exists.  Then you can remove the `:memory:` check.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+           not os.path.isdir(os.path.dirname(self._db_filename)):
+            os.makedirs(os.path.dirname(self._db_filename))
+        self._connection = sqlite3.connect(
+            self._db_filename, check_same_thread=False, isolation_level=None)
+        self._ensure_database_setup()
+
+    def execute(self, query, *parameters):
+        self._connection.execute(query, *parameters)
+
+    def cursor(self):
```
jamesls: We discussed removing the cursor didn't we?  I'd suggest simplifying this interface and not exposing cursors at all.  We shouldn't need it.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+    def _create_record_table(self):
+        self.execute(self._CREATE_TABLE)
+
+    def _try_to_enable_wal(self):
+        try:
+            self.execute(self._ENABLE_WAL)
+        except sqlite3.Error:
+            # This is just a performance enhancement so it is optional. Not all
+            # systems will have a sqlite compiled with the WAL enabled.
+            pass
```
jamesls: Can you log this at debug?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class PayloadSerializer(json.JSONEncoder):
+    def _encode_mutable_mapping(self, obj):
+        return dict(obj)
+
+    def _encode_datetime(self, obj):
+        return obj.isoformat()
+
+    def _unknown(self, obj, type_name):
```
jamesls: Where is this used?

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+            return self._encode_datetime(obj)
+        elif isinstance(obj, MutableMapping):
+            return self._encode_mutable_mapping(obj)
+        else:
+            return repr(obj)
+
+
+class DatabaseRecordWriter(object):
+    _WRITE_RECORD = """
+    INSERT INTO records(id, request_id, source, event_type, timestamp, payload)
```
jamesls: indent please.

awscli/customizations/history/db.py
```diff
@@ -0,0 +1,246 @@
...
+            return request_id
+        return None
+
+    def _format_payload(self, event_type, payload):
+        # If the payload has a body key it can be in bytes, so it needs to be
+        # converted to utf-8 if possible so it can be serialized as JSON later.
+        if event_type in self._BYTES_BODY_PAYLOADS:
+            body = payload.get('body')
+            if body is not None:
+                payload['body'] = ensure_text_type(body)
```
jamesls: I don't think you answered my original question about this (I might have missed it though).  What specific instances are you running into for this?  Additionally, if a payload can be bytes, then it's possible that you can't utf-8 encode it.  It looks like you just removed the previous version's base64 encoding.  Are you certain we don't need it?

stealthycoin: The body key on http_request and http_response payloads is as far as I can tell always bytes if it is not None.

awscli/customizations/history/db.py
stealthycoin: For writing yes it uses the connection and autocommit mode. I do not see a way to read directly from the connection though, the __iter__ and all the fetch* commands are on the cursor only as far as I can tell.

jamesls: It's exactly the same as inserts: https://docs.python.org/2/library/sqlite3.html#using-shortcut-methods

awscli/customizations/history/db.py
```diff
@@ -25,38 +25,25 @@
...
+        CREATE TABLE IF NOT EXISTS records (
+          id TEXT,
+          request_id TEXT,
+          source TEXT,
+          event_type TEXT,
+          timestamp INTEGER,
+          payload TEXT
+        )"""
+    _ENABLE_WAL = 'PRAGMA journal_mode=WAL'
+    _DEFAULT_DATABASE_NAME = 'history.db'
```
kyleknap: Can probably remove this class variable now.

tests/unit/customizations/history/test_db.py
```diff
@@ -607,81 +599,89 @@ def test_can_serialize_unknown_type(self):
...
+
+    def test_does_create_record_with_correct_fields(self):
+        record = self.builder.build_record('type', 'payload', 'source')
+        self.assertEqual(record['event_type'], 'type')
+        self.assertEqual(record['payload'], 'payload')
+        self.assertEqual(record['source'], 'source')
+        self.assertTrue('command_id' in record)
+        self.assertTrue('timestamp' in record)
+
+    def test_can_process_http_request_with_noone_body(self):
```
kyleknap: typo: noone --> none

tests/unit/customizations/history/test_db.py
```diff
@@ -607,81 +599,89 @@ def test_can_serialize_unknown_type(self):
...
         try:
-            builder.build_record('HTTP_REQUEST', {}, '')
+            self.builder.build_record('HTTP_REQUEST', {'body': None}, '')
         except ValueError:
             self.fail("Should not raise value error")
 
-    def test_can_process_http_response_with_no_body(self):
-        builder = RecordBuilder()
+    def test_can_process_http_response_with_nono_body(self):
```
kyleknap: Remove this test. It is the copy of the last one now.

awscli/customizations/history/db.py
```diff
@@ -83,11 +79,53 @@ def _encode_mutable_mapping(self, obj):
     def _encode_datetime(self, obj):
         return obj.isoformat()
 
+    def _try_encode(self, obj):
```
kyleknap: Shouldn't this be `_try_decode()`? Maybe call it `_try_decode_bytes()`?

awscli/customizations/history/db.py
```diff
@@ -83,11 +79,53 @@ def _encode_mutable_mapping(self, obj):
...
+        try:
+            obj = obj.decode('utf-8')
+        except UnicodeDecodeError:
+            obj = '<Byte sequence>'
+        return obj
+
+    def _remove_non_unicode_stings(self, obj):
+        if isinstance(obj, str):
+            obj = self._try_encode(obj)
+        elif isinstance(obj, dict):
```
kyleknap: Hmm. Would it be possible to do this in place or would does that risk mutating the payload in the upstream code logic?

awscli/customizations/history/db.py
```diff
@@ -83,11 +79,53 @@ def _encode_mutable_mapping(self, obj):
...
 
+    def _try_encode(self, obj):
+        try:
+            obj = obj.decode('utf-8')
+        except UnicodeDecodeError:
+            obj = '<Byte sequence>'
+        return obj
+
+    def _remove_non_unicode_stings(self, obj):
+        if isinstance(obj, str):
```
kyleknap: What's your thoughts on using `binary_type` here instead?

tests/unit/customizations/history/test_db.py
```diff
@@ -598,6 +538,64 @@ def test_can_serialize_unknown_type(self):
...
+    def test_can_serialize_non_utf_8_bytes_type_in_list(self):
+        original = ['foo', b'\xfe\xed']
+        encoded = ['foo', '<Byte sequence>']
+        string_value = json.dumps(original, cls=PayloadSerializer)
+        reloaded = json.loads(string_value)
+        self.assertEqual(encoded, reloaded)
+
+    def test_can_serialize_non_utf_8_bytes_type_in_tuple(self):
+        original = ('foo', b'\xfe\xed')
+        encoded = ['foo', '<Byte sequence>']
```
kyleknap: Should this remain a tuple?

tests/functional/history/test_db.py
```diff
@@ -301,3 +299,168 @@ def test_can_read_most_recent_records(self):
...
+        '^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$',
+        re.I
+    )
+
+    def setUp(self):
+        self.db = DatabaseConnection(':memory:')
+        self.writer = DatabaseRecordWriter(connection=self.db)
+        self.handler = DatabaseHistoryHandler(writer=self.writer)
+
+    def _assert_last_record_structure(self, event_type, payload, source,
```
kyleknap: It's kind of hard to follow this. I honestly would prefer there be the following:

* `_get_last_record()` to get the last record and the payload already loaded.
* `_assert_record_is_correct_structure()` which does the assertion on the tuple
* Then have more helper assertion methods like `_assert_expected_event_type()`, `_assert_expected_source_type()`, and `_assert_expected_payload()`, `_assert_has_request_id()`

The problem now for me is the assertion is too vague as it is now and makes it hard to follow the assertions.

tests/unit/customizations/history/test_db.py
stealthycoin: No its being serialzied to JSON which turns it into a list.

awscli/customizations/history/db.py
stealthycoin: I match exactly what the JSON encoder does so I would be hesitant to deviate.

awscli/customizations/history/db.py
stealthycoin: That would probably mutate it. I added a few tests to check for mutation.

PR #2971, opened 2017-11-20
Add history show command

kyleknap:
This is based on John's PR: https://github.com/aws/aws-cli/pull/2968 and relies on the botocore PR as well: https://github.com/boto/botocore/pull/1315.

To start using the `history show` command. First run:
```
$ aws configure set cli_history enabled
```

Then run a command:
```
$ aws ec2 describe-regions
```

and then run:
```
$ aws history show
```
And you should get an output like this summarizing the entire command for you (providing no args displays the last comand ran). Here is sample output from my terminal:
screen shot 2017-11-19 at 9 48 55 pm

Let me know what you all think.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -50,12 +50,13 @@ from awscli.utils import emit_top_level_args_parsed_event
 LOG = logging.getLogger('awscli.clidriver')
 LOG_FORMAT = (
     '%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s')
+HISTORY_RECORDER = get_global_history_recorder()
 
 
 def main():
     driver = create_clidriver()
     rc = driver.main()
-    get_global_history_recorder().record('CLI_RC', rc, 'CLI')
+    HISTORY_RECORDER.record('CLI_RC', rc, 'CLI')
     return rc
 
 
@@ -200,10 +201,9 @@ class CLIDriver(object):
             # command table.  This is why it's in the try/except clause.
             self._handle_top_level_args(parsed_args)
             self._emit_session_event(parsed_args)
-            history_recorder = get_global_history_recorder()
-            history_recorder.record(
+            HISTORY_RECORDER.record(
                 'CLI_VERSION', self.session.user_agent(), 'CLI')
-            history_recorder.record('CLI_ARGUMENTS', args, 'CLI')
+            HISTORY_RECORDER.record('CLI_ARGUMENTS', args, 'CLI')
             return command_table[parsed_args.command](remaining, parsed_args)
         except UnknownArgumentError as e:
             sys.stderr.write("usage: %s\n" % USAGE)
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -21,6 +21,7 @@ from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
+from botocore.history import get_global_history_recorder
 
 from awscli import EnvironmentVariables, __version__
 from awscli.compat import get_stderr_text_writer
@@ -53,7 +54,9 @@ LOG_FORMAT = (
 
 def main():
     driver = create_clidriver()
-    return driver.main()
+    rc = driver.main()
+    get_global_history_recorder().record('CLI_RC', rc, 'CLI')
+    return rc
 
 
 def create_clidriver():
@@ -196,7 +199,11 @@ class CLIDriver(object):
             # general exception handling logic as calling into the
             # command table.  This is why it's in the try/except clause.
             self._handle_top_level_args(parsed_args)
-            self._emit_session_event()
+            self._emit_session_event(parsed_args)
+            history_recorder = get_global_history_recorder()
+            history_recorder.record(
+                'CLI_VERSION', self.session.user_agent(), 'CLI')
+            history_recorder.record('CLI_ARGUMENTS', args, 'CLI')
             return command_table[parsed_args.command](remaining, parsed_args)
         except UnknownArgumentError as e:
             sys.stderr.write("usage: %s\n" % USAGE)
@@ -228,13 +235,15 @@ class CLIDriver(object):
             err.write("\n")
             return 255
 
-    def _emit_session_event(self):
+    def _emit_session_event(self, parsed_args):
         # This event is guaranteed to run after the session has been
         # initialized and a profile has been set.  This was previously
         # problematic because if something in CLIDriver caused the
         # session components to be reset (such as session.profile = foo)
         # then all the prior registered components would be removed.
-        self.session.emit('session-initialized', session=self.session)
+        self.session.emit(
+            'session-initialized', session=self.session,
+            parsed_args=parsed_args)
 
     def _show_error(self, msg):
         LOG.debug(msg, exc_info=True)
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -278,7 +278,7 @@ def _windows_shell_quote(s):
     return new_s
 
 
-def get_popen_pager_cmd_with_kwargs(pager_cmd=None):
+def get_popen_kwargs_for_pager_cmd(pager_cmd=None):
     """Returns the default pager to use dependent on platform
 
     :rtype: str
@@ -296,4 +296,5 @@ def get_popen_pager_cmd_with_kwargs(pager_cmd=None):
         popen_kwargs = {'shell': True}
     else:
         pager_cmd = shlex.split(pager_cmd)
-    return pager_cmd, popen_kwargs
+    popen_kwargs['args'] = pager_cmd
+    return popen_kwargs
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import sys
+import shlex
 import os
 import platform
 import zipfile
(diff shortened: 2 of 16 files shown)


awscli/customizations/history/show.py
```diff
@@ -0,0 +1,454 @@
...
+         'default': 'detailed', 'help_text': (
+            'Specifies which format to use in showing the events for '
+            'the specified CLI command.')}
+    ]
+
+    def __init__(self, session, db_reader=None, output_stream_factory=None):
+        super(ShowCommand, self).__init__(session)
+        self._db_reader = db_reader
+        if db_reader is None:
+            connection = DatabaseConnection(
```
stealthycoin: This logic was removed from the DatabaseConnection class.

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def _is_xml(self, body):
+        try:
+            xml.dom.minidom.parseString(body)
+        except xml.parsers.expat.ExpatError:
+            return False
+        return True
+
+    def _strip_whitespace(self, xml_string):
+        xml_dom = xml.dom.minidom.parseString(xml_string)
```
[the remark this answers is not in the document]
kyleknap: I will add a comment here. But the main reason was because services like ec2 will return xml that is already pretty and as a result blindly making it pretty again adds even more whitespace and newlines on top of what was already there.

awscli/compat.py
```diff
@@ -257,3 +273,28 @@ def _windows_shell_quote(s):
         # quoted.
         return '"%s"' % new_s
     return new_s
+
+
+def is_windows():
```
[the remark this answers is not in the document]
kyleknap: I was mainly copying what was done in the `help` logic. I'm fine switching over to `os.name`.

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,454 @@
...
+    def _get_api_num(self, event_record):
+        request_id = event_record['request_id']
+        if request_id:
+            if request_id not in self._request_id_to_api_num:
+                self._request_id_to_api_num[
+                    request_id] = self._num_api_calls
+                self._num_api_calls += 1
+            return self._request_id_to_api_num[request_id]
+
+    def _format_value_with_description(self, value, value_description,
```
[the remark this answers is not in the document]
kyleknap: I will see if I can update that function name, but essentially I use this to format the values defined in the section definition

awscli/customizations/history/show.py
kyleknap: Yep I added that logic back in a recent commit

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,454 @@
...
+    def _format_http_body_xml(self, body):
+        stripped_body = self._strip_whitespace(body)
+        xml_dom = xml.dom.minidom.parseString(stripped_body)
+        return xml_dom.toprettyxml(indent=' '*4, newl='\n')
+
+    def _format_http_body_json(self, body):
+        obj = json.loads(body)
+        return self._format_dictionary(obj)
+
+    def _color_title(self, text):
```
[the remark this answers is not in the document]
kyleknap: I will see if I can abstract out a bit more. What specifically were you thinking would go into this component_name -> color mapping abstraction

kyleknap:
@stealthycoin I updated based on feedback and also pointed the PR to the `history` branch

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,463 @@
...
+                self._request_id_to_api_num[
+                    request_id] = self._num_api_calls
+                self._num_api_calls += 1
+            return self._request_id_to_api_num[request_id]
+
+    def _format_description(self, value_description):
+        return self._color_if_configured(
+            value_description + ': ', 'description')
+
+    def _format_value(self, value, event_record, value_format=None):
```
jamesls: This class is really hard to follow.  You have two levels of dispatch here:

* `_display -> _display_*`: no dynamic dispatch, calls a series of `_format_*` calls
* `_format_value -> _format_*`: dynamic dispatch

What I'd suggest for this PR:

* If you're going to use dynamic dispatch in this class, you need to make sure you use a unique prefix that cannot interfere with other methods in this class (consider if you called `_format_value` with `value_format='value'`).
* Use an entirely different prefix for things that use dynamic dispatch.  It's confusing that some `_format_*` calls are called by `_display_*` and some are called by `_format_value`.

Going forward in future code, some suggestions to consider:

* Consider separate classes for your dynamic dispatch.  The only other methods in the class should be methods used by the methods you dispatch to.
* Have the dynamic dispatch map to some public method (e.g `serialize() -> _serialize_{type_name}`, `visit() -> _visit_{node_type}`, `parse() -> _parse_{type}`, `format() -> _format_{name}`)

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def _format_dictionary(self, obj, event_record=None):
+        return json.dumps(obj=obj, sort_keys=True, indent=4)
+
+    def _format_http_body(self, body, event_record):
+        if not body:
+            return 'There is no associated body'
+        elif event_record['payload'].get('streaming', False):
+            return 'The body is a stream and will not be displayed'
+        elif self._is_xml(body):
```
jamesls: Can you add a TODO here?  From what I can tell, in the happy case where the body is xml, we're parsing the body *three* times (and serializing it twice).  This might be what's causing the large delays/cpu usage I'm seeing for various s3 commands that call `ListObjects`.

tests/integration/customizations/history/test_show.py
```diff
@@ -0,0 +1,90 @@
...
+        self.environ = os.environ.copy()
+        self.environ['AWS_CONFIG_FILE'] = self.files.create_file(
+            'config', (
+                '[default]\n'
+                'cli_history = enabled'
+            )
+        )
+        self.environ['AWS_DEFAULT_PROFILE'] = 'default'
+        self.environ['AWS_DEFAULT_REGION'] = 'us-west-2'
+        self.environ['AWS_HISTORY_PATH'] = self.files.rootdir
```
jamesls: I thought we changed this to have a `CLI` prefix (e.g `AWS_CLI_HISTORY_PATH`).  Was that feedback not incorporated?

tests/unit/customizations/history/test_show.py
```diff
@@ -0,0 +1,813 @@
...
+            'timestamp': 0,
+        }
+        self.formatter.display(api_event)
+        self.formatter.display(http_event)
+        collected_output = ensure_text_type(self.output.getvalue())
+        self.assertIn('[0] API call made', collected_output)
+        self.assertIn('[0] HTTP request sent', collected_output)
+
+    def test_display_http_response(self):
+        event = {
```
jamesls: These display tests all seem to have the same format.  You could cut down on duplication by having a custom assertion method:

```python
self.assert_output(
    for_event={
    },
    contains_output='...'
)
```

awscli/compat.py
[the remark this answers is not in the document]
kyleknap: No particular reason. I originally had it as `platform.system` because it was more heavily used in the codebase. I can switch though to `sys.platform`

awscli/customizations/history/show.py
kyleknap: Makes sense. To fix it, I decided to rip out the value specific format dispatchers all to its own class that the `DetailedFormatter` uses to pretty format the values for each section. Hopefully that will be clearer.

tests/integration/customizations/history/test_show.py
kyleknap: Hmmm. I thought I changed that too. I think the test is using the wrong value.

awscli/clidriver.py
```diff
@@ -53,7 +54,9 @@
 
 def main():
     driver = create_clidriver()
-    return driver.main()
+    rc = driver.main()
+    get_global_history_recorder().record('CLI_RC', rc, 'CLI')
```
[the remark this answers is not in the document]
kyleknap: It makes it easier to patch out in the tests. If I do not do this, I have to patch each module level instantiation of the `HISTORY_RECORDER` which is tedious. I put this in my last commit so let me know what you think.

kyleknap:
@stealthycoin @jamesls Alright I updated based on feedback. For the most part, I took in all your feedback.

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,486 @@
...
+        return formatted_value + '\n'
+
+    def _color_if_configured(self, text, component):
+        if self._colorize:
+            color = self._COMPONENT_COLORS[component]
+            return color + text + colorama.Style.RESET_ALL
+        return text
+
+
+class SectionValuePrettyFormatter(object):
```
jamesls: Cool, much easier to read.

awscli/customizations/history/show.py
```diff
@@ -0,0 +1,486 @@
...
+            yield process.stdin
+        finally:
+            process.communicate()
+
+    @contextlib.contextmanager
+    def _get_stdout_stream(self):
+        yield get_binary_stdout()
+
+    def _get_process_pager_kwargs(self):
+        kwargs = get_popen_kwargs_for_pager_cmd(
```
jamesls: I was playing around with this, and one thing I noticed is that we assume that if a user sets a pager via the `PAGER` env var then it can handle color.  I've always had my default `PAGER` set to `less`, so when I first ran the `show` command I had all the escape sequences in my term:

```
$ aws history show
```
image

I just used `--color on | less -R` to get what I wanted (paginated color), but I wonder if there's anything we can do here.  Just a guess, but I'd think `export PAGER=less` is a pretty common configuration.  What do you think?

kyleknap: @jamesls I wonder if we should just add -R onto the command if we detect that the value is just `less` with no arguments? @stealthycoin is runs into a similar scenario when using `-S` for the `list` command.

kyleknap: Also do you usually get color when you use `less`? I think there may be an environment value `LESS` that sets these flags.

jamesls: > Also do you usually get color when you use less? 

No I don't.

Capturing offline feedback:

* Don't use their `PAGER` var if set.
* Add support for our own env var for paging in the future.

PR #2985, opened 2017-11-26
History list

stealthycoin:
Based off of Kyle's show branch

### DIFF
--- a/awscli/customizations/history/commands.py
+++ b/awscli/customizations/history/commands.py
@@ -1,0 +1,63 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
+
+from awscli.compat import is_windows
+from awscli.utils import is_a_tty
+from awscli.utils import OutputStreamFactory
+
+from awscli.customizations.commands import BasicCommand
+from awscli.customizations.history.db import DatabaseConnection
+from awscli.customizations.history.constants import HISTORY_FILENAME_ENV_VAR
+from awscli.customizations.history.constants import DEFAULT_HISTORY_FILENAME
+from awscli.customizations.history.db import DatabaseRecordReader
+
+
+class HistorySubcommand(BasicCommand):
+    def __init__(self, session, db_reader=None, output_stream_factory=None):
+        super(HistorySubcommand, self).__init__(session)
+        self._db_reader = db_reader
+        self._output_stream_factory = output_stream_factory
+        if output_stream_factory is None:
+            self._output_stream_factory = OutputStreamFactory()
+
+    def _connect_to_history_db(self):
+        if self._db_reader is None:
+            connection = DatabaseConnection(self._get_history_db_filename())
+            self._db_reader = DatabaseRecordReader(connection)
+
+    def _get_history_db_filename(self):
+        filename = os.environ.get(
+            HISTORY_FILENAME_ENV_VAR, DEFAULT_HISTORY_FILENAME)
+        if not os.path.exists(filename):
+            raise RuntimeError(
+                'Could not locate history. Make sure cli_history is set to '
+                'enabled in the ~/.aws/config file'
+            )
+        return filename
+
+    def _should_use_color(self, parsed_globals):
+        if parsed_globals.color == 'on':
+            return True
+        elif parsed_globals.color == 'off':
+            return False
+        return is_a_tty() and not is_windows
+
+    def _get_output_stream(self, preferred_pager=None):
+        if is_a_tty():
+            env_pager = os.environ.get('PAGER')
+            if env_pager is not None:
+                preferred_pager = env_pager
+            return self._output_stream_factory.get_pager_stream(
+                preferred_pager)
+        return self._output_stream_factory.get_stdout_stream()
--- a/awscli/customizations/history/commands.py
+++ b/awscli/customizations/history/commands.py
@@ -55,9 +55,6 @@ class HistorySubcommand(BasicCommand):
 
     def _get_output_stream(self, preferred_pager=None):
         if is_a_tty():
-            env_pager = os.environ.get('PAGER')
-            if env_pager is not None:
-                preferred_pager = env_pager
             return self._output_stream_factory.get_pager_stream(
                 preferred_pager)
         return self._output_stream_factory.get_stdout_stream()
--- a/awscli/customizations/history/commands.py
+++ b/awscli/customizations/history/commands.py
@@ -36,6 +36,9 @@ class HistorySubcommand(BasicCommand):
             connection = DatabaseConnection(self._get_history_db_filename())
             self._db_reader = DatabaseRecordReader(connection)
 
+    def _close_history_db(self):
+        self._db_reader.close()
+
     def _get_history_db_filename(self):
         filename = os.environ.get(
             HISTORY_FILENAME_ENV_VAR, DEFAULT_HISTORY_FILENAME)
--- a/awscli/customizations/history/list.py
+++ b/awscli/customizations/history/list.py
@@ -1,0 +1,119 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import json
+import datetime
+
+from awscli.compat import default_pager
+from awscli.customizations.history.commands import HistorySubcommand
+
+
+class ListCommand(HistorySubcommand):
+    NAME = 'list'
+    DESCRIPTION = (
+        'Shows a list of previously run commands and their command_ids. '
+        'Each row shows only a bare minimum of details including the '
+        'command_id, date, arguments and return code. You can use the '
+        '``history show`` with the command_id to see more details about '
+        'a particular entry.'
+    )
+    _COL_WIDTHS = {
+        'id_a': 38,
+        'timestamp': 24,
+        'args': 50,
+        'rc': 0
+    }
+
+    def _run_main(self, parsed_args, parsed_globals):
+        self._connect_to_history_db()
+        raw_records = self._db_reader.iter_all_records()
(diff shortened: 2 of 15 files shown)


awscli/customizations/history/list.py
```diff
@@ -0,0 +1,413 @@
...
+        prefix = ' ' * offset
+        padding_right = field_width - len(value) - offset
+        postfix = ' ' * padding_right
+        aligned_value = prefix + value + postfix
+        return aligned_value
+
+    def _write_title(self, stream, title, column_widths, total_width):
+        title_space = total_width - 2  # Leave space for the sides
+        centered_title = self._align_center(title, title_space)
+        self._write(stream, '-' * total_width + '\n', 'delim')
```
kyleknap: Hmmm I think there is another layer of abstraction missing. I think it would make sense if we had two abstractions here: the `TableFormatter` that knows what is going to write and a `TableStyler` which is a helper class that the `TableFormatter` will call out to when writing. Having the the `TableFormatter` do the writes as well makes it a little difficult to read especially when the `write()` takes in a type as well.


I could imaging this `TableStyler` class look such as the following:
```python
class TableStyler(object):
      def __init__(self, ouput_stream, colorize=True):

      def column(self):
      # Writes the delimeter

     def row(self, length):
     # Writes the - for the row

    def align_cell_left(self, value, length):
    def align_cell_center(self, value, length):
```
That way the `TableFormatter` does not need to worry about the nitty gritty parts of writing out lower level spacing and components of the table.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,413 @@
...
+from awscli.customizations.history.constants import DEFAULT_HISTORY_FILENAME
+
+
+class ListCommand(BasicCommand):
+    NAME = 'list'
+    DESCRIPTION = (
+        'Shows a list of previously run commands and their command_ids. '
+        'Each row shows only a bare minimum of details including the '
+        'command_id, date, arguments and return code. You can use the '
+        '`history show` with the command_id to see more details about '
```
kyleknap: The help renders out to rst so you probably you double backticks.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def has_more_content(self):
+        return len(self._value) > 0
+
+    def get_line(self, space):
+        raise NotImplementedError('format_line')
+
+
+class StringTableCellFormatter(TableCellFormatter):
+    def get_line(self):
```
kyleknap: It would probably be good to include the space argument for this so a type check is not needed to know how many arguments should be passed in.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,456 @@
...
+        retval = self._value
+        self._value = ""
+        return retval
+
+
+class ArgumentTableCellFormatter(TableCellFormatter):
+    _TRUNCATED_ENDING = '...'
+    _PARAM_VALUE_INDENTATION = '  '
```
kyleknap: Super nit-pick but it may be easier to read these if it was a single space multiplied by the number of indentiation so lie `' ' * 2`. Or you can just repurpose the class variable to be an integer for the number of space indentations.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,456 @@
...
+    be split into a number of terminal rows.
+    This class is responsible for wrapping a table cell value and providing an
+    interface for asking for just 1 terminal row to write.
+
+    Now when rendering a table row, each column can be iterated over asking
+    its TableCellFormatter if it has a terminal row to write. This is repeated
+    until all TableCellFormatters indicate that they are out of content, then
+    the table row can be capped with a footer.
+    """
+    def __init__(self, value, space):
```
kyleknap: So I still am not following what the `value` and `space` are for. Could we add docstrings describing these in both the `__init__()` and `get_line(self, space)`?

tests/functional/history/test_list.py
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
```
kyleknap: It would be good to flesh this out just even it is a just a singe smoke test.

tests/unit/test_utils.py
```diff
@@ -120,3 +122,67 @@ def test_returns_none_if_event_is_too_short(self):
...
+        self.popen = mock.Mock(subprocess.Popen)
+        self.stream_factory = OutputStreamFactory(self.popen)
+
+    @mock.patch(
+        'awscli.utils.get_popen_kwargs_for_pager_cmd')
+    def test_pager(self, mock_get_popen_pager):
+        mock_get_popen_pager.return_value = {
+                'args': ['mypager', '--option']
+        }
+        with mock.patch('os.environ', {}):
```
kyleknap: I think we can get rid of this patch now that the environment has been hoisted and maybe move a test that relies on the environment variable to both `test_show.py` and `test_list.py`?

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,491 @@
...
+    ])
+
+    def _run_main(self, parsed_args, parsed_globals):
+        self._connect_to_history_db()
+        raw_records = self._db_reader.iter_all_records()
+        records = RecordAdapter(raw_records)
+        if not records.has_next():
+            raise RuntimeError('No recorded history.')
+
+        with self._get_output_stream('less -SR') as output_stream:
```
kyleknap: I just realized this from reading the tests but I do not think we can hard code it like this to always use `less -SR` as it will not work for windows. Mainly because windows does not have `less` installed, but I would recommend checking before changing it. It is not an issue though if that is case. I say in the `compat` module we just add the following logic that we can import into the `list` command:
```python
is_windows:
     default_pager = 'more'
else:
     default_pager = 'less -R'
```
From there we can just check if the default pager we got back is less and just clobber it with our own in `less`

tests/unit/test_utils.py
```diff
@@ -186,3 +191,12 @@ def test_exit_of_context_manager_for_pager(self):
     def test_stdout(self, mock_binary_out):
         with self.stream_factory.get_stdout_stream():
             self.assertTrue(mock_binary_out.called)
+
+    def test_can_silence_io_error_from_pager(self):
+        self.popen.return_value = MockProcess()
+        try:
+            with self.assertRaises(RuntimeError):
```
kyleknap: Why does this raise a `RuntimeError`? I would assume that this test only consist of the try,except,fail statement

kyleknap:
Also I think the tests are failing so we should get that updated.

tests/unit/test_utils.py
stealthycoin: The context manager never yields since I throw an IOError which in turn raises a RuntimeError which covers the IOError if I don't ignore it. I couldn't find a great seam otherwise for testing this. Either I let the yield happen and loose my chance to throw an IOError since there is no real process to throw it from, or I throw the IOError and also get a RuntimeError.

kyleknap: Ahh might be good to put a comment there then

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,499 @@
...
+
+        preferred_pager = self._get_preferred_pager()
+        with self._get_output_stream(preferred_pager) as output_stream:
+            use_color = self._should_use_color(parsed_globals)
+            styler = TableStyler(output_stream, colorize=use_color)
+            formatter = TableFormatter(self._COL_WIDTHS, styler)
+            formatter('history list', records)
+        return 0
+
+    def _get_preferred_pager(self):
```
jamesls: You're going to consolidate with what @kyleknap has for figuring out paging/color options once that's all merged correct?

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,499 @@
...
+        'Return Code': StringTableCellFormatter,
+    }
+
+    def create(self, col, value, space):
+        cls = self._COL_TO_FORMATTER_TYPE[col]
+        formatter = cls(value, space)
+        return formatter
+
+
+class TableStyler(object):
```
jamesls: There's already a table formatter used for the `--output table` format.  Why isn't that class being used?  Doesn't seem like we should be maintaining two separate classes.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,499 @@
...
+        # line and the parameter would fit we need to decide if we are
+        # going to skip it. If we do we need to put it back in the value
+        # so it can be
+        if param and line and fits:
+            if self._skip_param(line, candidate, space):
+                self._value.insert(0, candidate)
+                return True
+        return False
+
+    def get_line(self):
```
jamesls: I don't think this accounts for newlines in the values themselves which will mess up the format, e.g. an argument with a json value.

awscli/customizations/history/list.py
```diff
@@ -0,0 +1,499 @@
...
+        ('Arguments', 50),
+        ('Return Code', 13)
+    ])
+
+    def _run_main(self, parsed_args, parsed_globals):
+        self._connect_to_history_db()
+        raw_records = self._db_reader.iter_all_records()
+        records = RecordAdapter(raw_records)
+        if not records.has_next():
+            raise RuntimeError('No recorded history.')
```
jamesls: Please provide a better error message.

awscli/customizations/history/list.py
kyleknap: Yeah. I think I would prefer to make the change (if any) in this PR related to the `PAGER` env var instead of mine because this PR consolidates them.

awscli/customizations/history/list.py
stealthycoin: It doesn't really do what we want. I tried using that for awhile and I was just fighting the design since it does all sorts of magic under the hood for arbitrary payload shapes whereas we have a fixed schema and a fixed order we want to show, with specific colors. I found it much easier to get what I wanted when I removed the existing one and started from scratch.

awscli/customizations/history/list.py
stealthycoin: Yes.

awscli/customizations/history/list.py
jamesls: That doesn't make sense to me.  It doesn't do anything payload shapes.  It doesn't know anything about shapes or models.  Are you sure you're looking at `TableFormatter`?

I suppose I was being too subtle with my previous feedback.  Please consolidate this.  Unless you have concrete issues with why this can't be done (or point me at a branch where you tried), we shouldn't be maintaining two different ways to draw tables.

Either that or drop the table format.  I personally wish we would have have something similar to the the text based output so I can manipulate it with `cut`, `awk`, etc.  That's how I was hoping to use the list command.

awscli/customizations/history/commands.py
```diff
@@ -0,0 +1,63 @@
...
+    def _should_use_color(self, parsed_globals):
+        if parsed_globals.color == 'on':
+            return True
+        elif parsed_globals.color == 'off':
+            return False
+        return is_a_tty() and not is_windows
+
+    def _get_output_stream(self, preferred_pager=None):
+        if is_a_tty():
+            env_pager = os.environ.get('PAGER')
```
kyleknap: Talked offline about this, but I think we are just going to get rid of the functionality for now. I think if enough people want it we can do a `AWS_CLI_PAGER` parameter in the future for all features.

PR #2991, opened 2017-11-28
Fixes #2990 and #2518. Avoid AttributeError when awscli is invoked fr…

### LINKED ISSUE #2990: Reopen #2518: Prevent AttributeError when invoking awscli from python3 daemon
This is to reopen #2518. Details are there and will be submitting a new PR with an otherwise harmless, one-line fix.

emurphy:
…om python3 daemon.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -72,7 +72,7 @@ if six.PY3:
 
     raw_input = input
 
-    binary_stdin = sys.stdin.buffer
+    binary_stdin = sys.stdin.buffer if sys.stdin is not None else None
 
     def _get_text_writer(stream, errors):
         return stream


JordonPhillips:
I don't think simply making it `None` would be the best way to handle this. Any time we try to access it we would end up rasing an error like `'NoneType' object has no attribute 'write'` which isn't very helpful. I would prefer we add in a function `get_binary_stdin` with raises an exception with a descriptive method if stdin isn't available. That would delay the exception while also providing a helpful error message.

emurphy:
Making the `binary_stdin` variable `None` is consistent with `sys.stdin` being `None` and is far better than triggering an error. You are welcome to implement your preferred `get_binary_stdin` function. In the meantime, please merge #2991 so that the aws-cli works for this use case.

emurphy:
There may be a misunderstanding here. First, one thing to clarify is that I represent a customer of AWS. Our company pays hundreds of thousands of dollars for the privilege to use your services. It is my understanding that you are a paid employee or contractor of AWS.

Like other paying customers, we have encountered an error in the aws-cli software. We have helpfully, at our own expense, provided you with a minimal, lowest-possible-risk solution. Let us analyze the line of code to ensure there is no misunderstanding there…

Before merging this PR:
> binary_stdin = sys.stdin.buffer

After merging this PR:
> binary_stdin = sys.stdin.buffer if sys.stdin is not None else None

With a remedial understanding of how to program, anyone can understand the if statement.
Two input conditions are worth considering:
1. sys.stdin is None
2. sys.stdin is not None

In case #1:
In the Before case the program errors. aws-cli breaks, fails to meet a minimum requirement. 
In the After case the program continues, with the variable `binary_stdin` set to None. Not a problem, meets the minimum need.

In case #2, both Before and After have the same result. binary_stdin is safely set to sys.stdin.buffer. All is well.

I could challenge your statement that this fix is not “the best”, but I believe it is irrelevant. In the before case the software breaks; we are not getting from you what we pay for. In the after case your software works, we get what we pay for.

There are no drawbacks; there are no excuses for your inaction.
Please, merge it now.

I will not comment further. I will not change this PR. Our next step, if this issue is not addressed, will be to mention it to our AWS account manager.

Please, as your humble customer, I beg you… merge it now.

PR #2998, opened 2017-11-29
S3 commands ensure error RC takes precedence

joguSD:
If an S3 command encounters an error and a warning it will use the warning RC masking that the error happened. This adds a test for this case and ensures we use the correct RC.

### DIFF
--- a/.changes/next-release/bugfix-S3-3229.json
+++ b/.changes/next-release/bugfix-S3-3229.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "S3",
+  "description": "S3 commands that encountered both an error and warning will now return the RC for error instead of warning."
+}
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1082,7 +1082,7 @@ class CommandArchitecture(object):
         rc = 0
         if files[0].num_tasks_failed > 0:
             rc = 1
-        if files[0].num_tasks_warned > 0:
+        elif files[0].num_tasks_warned > 0:
             rc = 2
         return rc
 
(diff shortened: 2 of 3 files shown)


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -1055,6 +1055,33 @@ def test_bad_symlink(self):
                       p.stderr)
 
 
+class TestErrorAndWarning(BaseS3IntegrationTest):
```
kyleknap: What is your thoughts on moving this to a functional test? It would be better because we run the functional tests more frequently and we do not necessarily need to send the request over the wire to test for this.

joguSD: I originally looked into implementing this as a functional test but opted for an integration test because I wasn't sure if I could guarantee the order that the requests are made. All examples of the functional tests for these currently only have a single line of api request logic and not two. Basically, will the S3 transfer manager guarantee the order of the `put_object` calls?

kyleknap: So you could set the `max_concurrent_requests` rate to `1` to avoid that and it is fine that we would be overriding the config file because we would not be making an API requests. From there you could recreate a scenario that would trigger this. There are a probably a handful that you can do but one would be to do a recursive upload and make one of the files a special file type so it gets skipped and the other be normal and have the `PutObject` fail. This way you do not have to override the `max_concurrent_requests` either. The one downside is that we would have to skip for Windows with this approach specifically for the special file. But if you can think of something that does not skip Windows that would be ideal, maybe look around in the tests for cases where we skip?

kyleknap: Actually instead of a special file, you could use the `set_invalid_utime` from before and that should allows us to avoid skipping on Windows.

jamesls: Looks like this is just pending this feedback getting incorporated right?  Where did we land on this?

joguSD: Yep, gonna update this to become a functional test.

PR #3008, opened 2017-11-30
allowing colorama 0.3.9

kkdd:


stealthycoin:
Can you explain why you need the ceiling raised on the colorama version? We also could not take a change that modifies the botocore version as that is completely handled by automation. Could you open an issue explaining your reasoning for needing a different colorama version so we can discuss it there?

PR #3026, opened 2017-12-08
Also catch ValueError when handling deleted files

### LINKED ISSUE #2960: aws s3 sync racing with local files
I'm having an issue with S3 sync bombing when local files (which would be excluded anyway) are removed by another process.

I found issue #2333, and updated to the latest release, but the issue persists.  I think there is more that needs to be done to avoid these races.

matthiasr:
This try/except tries to handle failed stat calls due to files not being
there. However, before it can do so, get_file_stat has already
transformed any IOError into a ValueError. This is presumably necessary
for Python versions < 3.3 where there is no distinct FileNotFoundError,
so we can't just delete that mapping.

Fixes #2960.

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -214,7 +214,7 @@ class FileGenerator(object):
     def _safely_get_file_stats(self, file_path):
         try:
             size, last_update = get_file_stat(file_path)
-        except OSError:
+        except (OSError, ValueError):
             self.triggers_warning(file_path)
         else:
             last_update = self._validate_update_time(last_update, file_path)
--- a/tests/functional/s3/test_sync_command.py
+++ b/tests/functional/s3/test_sync_command.py
@@ -176,7 +176,7 @@ class TestSyncCommand(BaseAWSCommandParamsTest):
         # get their stats.
         def side_effect(_):
             os.remove(full_path)
-            raise OSError()
+            raise ValueError()
         with patch(
                 'awscli.customizations.s3.filegenerator.get_file_stat',
                 side_effect=side_effect
--- a/tests/functional/s3/test_sync_command.py
+++ b/tests/functional/s3/test_sync_command.py
@@ -166,7 +166,37 @@ class TestSyncCommand(BaseAWSCommandParamsTest):
 
         self.assertFalse(os.path.exists(full_path))
 
-    def test_sync_skips_over_files_deleted_between_listing_and_transfer(self):
+    # When a file has been deleted after listing,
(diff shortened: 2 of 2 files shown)


matthiasr:
An open question to me is why the [test](https://github.com/aws/aws-cli/blob/6601f809be07747e530de695459b812c5f1010ea/tests/functional/s3/test_sync_command.py#L169) does not show this problem at all. It passes before and after, at least on Python 3.6, so I left it alone.

groner:
As I understand it, the original get_file_stat() isn't called at all when patched, only the side_effect is called.  The test in question just isn't exercising the case where get_file_stat raises ValueError.

matthiasr:
Ah, that makes sense, and indeed if I change the test to raise a ValueError it failed before and passes now.

matthiasr:
I'll also rearrange the commits so that the failing test comes before the fix.

matthiasr:
hold on …

matthiasr:
(I'm trying to push the commits one-by-one so that Travis can show the legit build failure, but I pushed too quickly; will try again)

matthiasr:
Done! Ready for review.

stealthycoin:
@matthiasr Thanks for the PR! Could we get a separate test for the ValueError rather than replacing the old test? I would rather have a test in place for each.

matthiasr:
Done!

Now that we got this far … I'm wondering if it wouldn't be better to remove the IOError->ValueError translation after all. It gives the exception more context, but on Python 3.3+ this is not really necessary because a FileNotFoundError is emitted, and on Python 2.x it doesn't actually work.

From my reading of [PEP 3151](https://www.python.org/dev/peps/pep-3151/), Python 3.3+ treat IOError and OSError as the same; the actual error is a FileNotFoundError which is a subclass of the latter, so `get_file_stat` translates it into a ValueError as well. In my testing (but on limited platforms), `os.stat` in Python 2.6.6 and 2.7.14 both raise an `OSError: [Errno 2]`.

I'm thinking about scrapping this approach and instead removing the ValueError mapping, but that still leaves the possibility of two different exceptions being raised: an OSError on Python 2.x, or a FileNotFoundError on 3.x. I _think_ I know how to make the test reflect that …

matthiasr:
@stealthycoin I took a stab at the alternative approach in #3030. Only one of the two should go in, I leave it to you to choose. I think #3030 makes the error situation less confusing, while this one is more defensive.

joguSD:
@matthiasr thanks for the PR! I'm a bit conservative so I'm going to tend towards the defensive solution. Undoing the mapping like in #3030 removes one possible area where a `ValueError` is raised in place of an `IOError` but I would be concerned that this might happen in other places too. Either way, this fix looks good to me.

PR #3030, opened 2017-12-09
Remove mapping of exception from os.stat

### LINKED ISSUE #2960: aws s3 sync racing with local files

matthiasr:
This mapping has no effect on Python 2.x, because os.stat raises an
OSError if the file is not found, not an IOError. Python 3.3+ raise a
FileNotFoundError, so the mapping provides no additional context.

It does however throw off the handling of this error later, so this
change fixes #2960.

To make sure the error handling works correctly, throw the correct
exception depending on the python version running the test.

This is an alternative to #3026.

### DIFF
--- a/tests/functional/s3/test_sync_command.py
+++ b/tests/functional/s3/test_sync_command.py
@@ -176,7 +176,9 @@ class TestSyncCommand(BaseAWSCommandParamsTest):
         # get their stats.
         def side_effect(_):
             os.remove(full_path)
-            raise OSError()
+            # PEP 3151: Python 3.3+ translate this into a FileNotFoundError.
+            # This matches the behaviour of os.stat across Python versions.
+            raise OSError(2, "file not found", "foo.txt")
         with patch(
                 'awscli.customizations.s3.filegenerator.get_file_stat',
                 side_effect=side_effect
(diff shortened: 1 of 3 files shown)


tests/functional/s3/test_sync_command.py
```diff
@@ -176,7 +176,12 @@ def test_sync_skips_over_files_deleted_between_listing_and_transfer(self):
         # get their stats.
         def side_effect(_):
             os.remove(full_path)
-            raise OSError()
+            # PEP 3151: Python 3.3+ raise a FileNotFoundError, earlier versions
+            # an OSError from os.stat if the file has gone away.
+            try:
+                raise FileNotFoundError(2, "file not found", "foo.txt")
+            except NameError:
+                raise OSError(2, "file not found", "foo.txt")
```
groner: `OSError(errno.ENOENT, 'file not found')` will return `FileNotFoundError`, but only if called with the second argument.

matthiasr: Oh great, then I'll remove the try/except and adjust the comment (I would not remember this in half a year).

PR #3047, opened 2017-12-20
Wait for created buckets to exist in integ tests

JordonPhillips:
This enforces the use of waiters in the S3 integ tests to make sure
that buckets exist before we start using them. This will help with
issues where sometimes certain operations hit a NoSuchBucket error
even when other operations on the same bucket will succeed.


awscli/testutils.py
```diff
@@ -789,7 +792,13 @@ def put_object(self, bucket_name, key_name, contents='', extra_args=None):
...
         self.remove_all_objects(bucket_name)
         client = self.create_client_for_bucket(bucket_name)
-        response = client.delete_bucket(Bucket=bucket_name)
+
+        # Ironically there's a chance that, even though the bucket has been
+        # used several times, the delete will fail due to eventual
+        # consistency issues. So we need to make sure the bucket actually
+        # exists before deleting it.
+        self.wait_bucket_exists(bucket_name, min_successes=5)
+        client.delete_bucket(Bucket=bucket_name)
```
jamesls: This still isn't guaranteed to work, the issue is that despite other operations succeeding (List and potentially even HeadBucket), the *delete_bucket* call fails.  What about changing this slightly to minimize the chances of this failing:

1. Try to delete the bucket
2. Catch the NoSuchBucket error
3. If the bucket doesn't exist, then we're good.
4. Otherwise retry up to N times.

awscli/testutils.py
```diff
@@ -786,10 +789,20 @@ def put_object(self, bucket_name, key_name, contents='', extra_args=None):
...
+
+        # There's a chance that, even though the bucket has been used
+        # several times, the delete will fail due to eventual consistency
+        # issues.
+        for _ in range(attempts):
+            try:
+                client.delete_bucket(Bucket=bucket_name)
+                break
+            except client.exceptions.NoSuchBucket:
+                time.sleep(delay)
```
jamesls: This has the downside that we won't fast fail on invalid bucket names, or in the edge case where the bucket is actually deleted but we get a 404.  The step (3) that I suggested would allow us to short circuit this condition, with the caveat that HEAD bucket and DELETE bucket have to agree.

JordonPhillips: I don't think that's something we really have to worry about. There's no reason for us to be calling delete on a bucket multiple times. Introducing that condition re-introduces the risk of the different paths not having correct information.

JordonPhillips:
update: fixed an issue where the exception wasn't being re-raised when the attempts ran out

awscli/testutils.py
jamesls: I still think is worth having a fast fail in the case of errors, even if it just happens when a dev is working on a new test, shorter feedback loops really help.  A 25 second wait penalty seems unfortunate.

Like I mentioned, the *only* issue here is that both delete_bucket() and head_bucket() say the bucket doesn't exist, and yet the bucket still somehow exists.  In that case the test won't fail but we'll leak a bucket in our test runs.

If we were actually worried about that case we could add a similar `min_success` thing with our head_bucket calls.

awscli/testutils.py
joguSD: For 3) are you suggesting to do a head_bucket to determine if it exists?

PR #3062, opened 2017-12-28
Add reference to the universal param docs

JordonPhillips:
This adds a reference to the main doc page for the cli which
documents all the universal parameters.

This depends on boto/botocore#1350

Here's what the html docs will look like:

help

That link will take you [here](http://docs.aws.amazon.com/cli/latest/reference/).

Here's what the help pages will look like:

screen shot 2017-12-28 at 10 08 41 am

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -571,6 +571,9 @@ class OperationDocumentEventHandler(CLIDocumentEventHandler):
         doc.style.dedent()
         doc.style.new_paragraph()
 
+    def doc_options_end(self, help_command, **kwargs):
+        self._add_top_level_args_reference(help_command)
+
 
 class TopicListerDocumentEventHandler(CLIDocumentEventHandler):
     DESCRIPTION = (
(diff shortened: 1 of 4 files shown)


awscli/clidocs.py
```diff
@@ -340,6 +340,16 @@ def doc_description(self, help_command, **kwargs):
...
+        self._add_top_level_args_reference(help_command)
+
+    def _add_top_level_args_reference(self, help_command):
+        help_command.doc.writeln('')
+        help_command.doc.write("See ")
+        help_command.doc.style.internal_link(
+            title="'aws help'",
+            page='/reference/index'
+        )
+        help_command.doc.writeln(' for descriptions of universal parameters.')
```
jamesls: Maybe call these global parameters?

kyleknap:
I wonder if it makes sense if we put this at the end of the synopsis and options sections instead? If you are reading the manpage as is, I can see someone skipping over the snippet in the description and go straight to the synopsis/options section and miss out about the see more part in the description.

JordonPhillips:
I'm not a fan of that because the synopsis section currently only shows arguments with no other text. I would prefer to keep it that way.

kyleknap:
So how about both in the description and the end of the option section then?

JordonPhillips:
@kyleknap I added tests

kyleknap:
Yep. Still looks fine. Good catch.

PR #3071, opened 2018-01-06
Add --preserve-symlinks option for "deploy push" command

iansltx:
By default, symlinks are materialized into their targets when they are zipped for deploy. Some applications, e.g. PHP's Doctrine ORM vendor/bin scripts, rely on symlinks to be preserved to work, and are thus broken by the default deploy behavior if your PHP project's deployment archive includes its vendor directory (which it should).

This commit adds an option (default off for BC reasons) to preserve symlinks when creating the deployment archive. Used --ignore-hidden-files as a template for parameter implementation.

Tests have been updated with the extra argument, including asserting that you can't specify to both preserve and not preserve symlinks simultaneously. However, while I've confirmed that behavior works as expected in the scenario for which I needed this functionality, there isn't an automated test to that effect, partially because I'm not sure how that test would fare across all environments.

References:
https://gist.github.com/kgn/610907
http://www.mail-archive.com/python-list@python.org/msg34223.html

### DIFF
--- a/awscli/customizations/codedeploy/push.py
+++ b/awscli/customizations/codedeploy/push.py
@@ -95,6 +95,24 @@ class Push(BasicCommand):
             )
         },
         {
+            'name': 'preserve-symlinks',
+            'action': 'store_true',
+            'default': False,
+            'group_name': 'preserve-symlinks',
+            'help_text': (
+                'Optional. Set the --preserve-symlinks flag to preserve symlinks '
+                'when building your deployment archive; otherwise, set the '
+                '--no-preserve-symlinks flag (the default) to resolve symlinks '
+                'to their contents and include contents directly in the archive.'
+            )
+        },
+        {
+            'name': 'no-preserve-symlinks',
+            'action': 'store_true',
+            'default': False,
+            'group_name': 'preserve-symlinks'
+        },
+        {
             'name': 'description',
             'synopsis': '--description <description>',
             'help_text': (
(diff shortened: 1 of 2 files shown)


iansltx:
The failures on Python 3 here appear to be unrelated to the code I just pushed, but as Python isn't my primary language I don't mind being proven wrong.

iansltx:
Ah, Python 3 doesn't have a long type. Explains why the tests succeeded/failed the way they did. Will commit a fix as soon as I can figure out how to make it comparible with both versions.

iansltx:
Looks like that single-character change didn't break intended functionality on my Python 2 based environment, so if this test run passes we should be golden?

iansltx:
Single-character change didn't break my Python 2.7 based environment. Looks like the final obstacle is test coverage, which I'd appreciate help with.

justindho:
Hi @iansltx, thanks for the PR. Apologies for the delay in getting back to you.

Our team just put out a [contribution guide](https://github.com/aws/aws-cli/blob/v2/proposals/contribution-guide.md) detailing improvements to the contribution process. I have created issue #6989 for this PR. We will wait for the issue to receive 10 upvotes before moving this PR into the ready-for-review stage on our [project board](https://github.com/orgs/aws/projects/21/views/2). The rationale for this can be found in the [rationale section](https://github.com/aws/aws-cli/blob/v2/proposals/contribution-guide.md#rationale-upvotes) of the contribution guide.

For the time being, I will mark this PR as a draft.

RyanFitzSimmonsAK:
Hi @iansltx, thanks for creating this PR and for your patience. Could you please rebase this PR against the latest branch for the team’s further consideration going forward?

RyanFitzSimmonsAK:
Going to close this PR for the time being.  If someone later decides they want to pick this up, rebase it and it will be reopened for consideration.

PR #2620, opened 2017-05-22
feature/cfn-deploy-tag-support

Posnet:
I have added support to the `aws cloudformation deploy` command for tags.
This ensures that change sets and their stacks are tagged correctly on create or update.

PR #3106, opened 2018-01-23
Cfn deploy tag support

sanathkr:
Taking over #2620 and #3105. This PR modifies the tags parameter to be specified in `mykey=myvalue` format instead of the `{Key="mykey", Value="myvalue"}` format. This keeps the tags parameter format consistent with --parameter-overrides

### DIFF
--- a/tests/unit/customizations/cloudformation/test_deploy.py
+++ b/tests/unit/customizations/cloudformation/test_deploy.py
@@ -13,7 +13,7 @@
 import mock
 import tempfile
 import six
-from mock import patch, Mock, MagicMock
+from mock import patch, Mock, MagicMock, call
 import collections
 
 from awscli.testutils import unittest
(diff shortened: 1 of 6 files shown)


tests/unit/customizations/cloudformation/test_deploy.py
```diff
@@ -107,19 +111,30 @@ def test_command_invoked(self, mock_yaml_parse):
...
+                    None,
+                    not self.parsed_args.no_execute_changeset,
+                    None,
+                    [],
+                    None,
+                    fake_tags,
+                    True
+                )
+
+                print self.deploy_command.parse_key_value_arg.call_args_list
```
stealthycoin: Rogue print statement.

sanathkr:
I happened to force git push to my fork which broke CodeCov's diff-level coverage. I assume this isn't a problem since `deploy.py`, the file that I changed, has 100% coverage now

PR #3145, opened 2018-02-13
changing quotes to double quotes in example

soo-aws:
changing quotes to double quotes in example so it will work for users on windows per customer feedback.


soo-aws:
why would changing text affect the code coverage? that file wasn't changed and no tests were changed... This is my first doc update, did I do something wrong when submitting the PR?

PR #3016, opened 2017-12-05
Add request_payer parameter to sync, cp, and rm subcommands

sp-niemand:
Adds support for [Requester Pays buckets](http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html) to some `aws s3` subcommands.

Addresses several related issues:
https://github.com/aws/aws-cli/issues/2845
https://github.com/aws/aws-cli/issues/2557

Also this (?):
https://github.com/aws/aws-cli/issues/797

PR #3147, opened 2018-02-14
Add --request-payer to transfer commands

kyleknap:
This builds on this previous PR: https://github.com/aws/aws-cli/pull/3016

Compared to the original PR, there was a lot more lines added. For the most part, the logic from the original PR stayed intact. Just I added a lot of functional tests for this as practically all API requests used in a command will require the request payer header or the command will fail, and so I wanted to be sure that I caught all of the different scenarios (good thing I did too because I found a bug in s3transfer as well related to this). I also tested it locally on a bucket enabled for requester pays using an account that does not own the bucket and it was working well there. For anyone following/reviewing this PR, feel free to try it out locally as well and that will help double check I did not miss anything.

Also, note tests won't pass until this PR gets merged from `s3transfer`: https://github.com/boto/s3transfer/pull/103

### DIFF
--- a/.changes/next-release/enhancement-s3-9777.json
+++ b/.changes/next-release/enhancement-s3-9777.json
@@ -1,0 +1,5 @@
+{
+  "category": "``s3``", 
+  "type": "enhancement", 
+  "description": "Expose ``--request-payer`` to ``cp``, ``sync``, ``mv`, and ``rm`` commands."
+}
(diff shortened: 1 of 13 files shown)


aquanyc:
@kyleknap 

When do you think there will be a version we can pull for testing ?

kyleknap:
@aquanyc It was recently released in version [1.14.42](https://github.com/aws/aws-cli/blob/develop/CHANGELOG.rst#11442). So I would recommend upgrading to the latest version of the CLI to start using it.

PR #3160, opened 2018-02-21
Typo in glacier warning message

leijurv:
The warning message currently is: `You must restore the object to be able to the perform operation.` 

It should instead read: `You must restore the object to be able to perform the operation.`


stealthycoin:
Thanks for the fix!

PR #3167, opened 2018-02-27
Update describe-connections example

katerini:
Update example output to include more fields

### DIFF
--- a/awscli/examples/directconnect/describe-connections.rst
+++ b/awscli/examples/directconnect/describe-connections.rst
@@ -7,15 +7,18 @@ The following ``describe-connections`` command lists all connections in the curr
 Output::
 
   {
-      "connections": [
-          {
-              "ownerAccount": "123456789012", 
-              "connectionId": "dxcon-fg31dyv6", 
-              "connectionState": "requested", 
-              "bandwidth": "1Gbps", 
-              "location": "TIVIT", 
-              "connectionName": "Connection to AWS", 
-              "region": "sa-east-1"
-          }
-      ]
+    "connections": [
+        {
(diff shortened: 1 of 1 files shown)


awscli/examples/directconnect/describe-connections.rst
```diff
@@ -7,15 +7,18 @@ The following ``describe-connections`` command lists all connections in the curr
...
-      ]
-  }
\ No newline at end of file
+    "connections": [
+        {
+            "awsDevice": "EqDC2-123h49s71dabc", 
+            "ownerAccount": "123456789012", 
+            "connectionId": "dxcon-fguhmqlc", 
+            "lagId": "dxlag-ffrz71kw", 
+            "connectionState": "down",
```
stealthycoin: According to the docs its `Down` with a capital.

katerini: This output is from an actual describe call, so I think the docs probably need to be updated/corrected. I'll make a note to do this.
The reason for this update was because of a request to include a device ID in the example output (which is not returned for a connection that it's in the 'requested' state, as per the previous example). Not all parameters are always returned in a describe call; it depends on the state and configuration of the connection.

katerini:
Looking again at the API docs, the connection states are correctly defined (in lowercase).
https://docs.aws.amazon.com/directconnect/latest/APIReference/API_Connection.html

However, these values are not returned in the CLI reference. Only the description is returned, which uses normal writing convention (starting a sentence with an initial capital) to describe the states in more detail. 

With regards to adding examples to include other parameters, I've created a backlog item to add more in future. However, can update this particular example in the meantime? Thanks!

PR #3177, opened 2018-03-04
Fix datapipeline CLI args for multi-value params

fffergal:
Previously adding 1 or 2 values with the same key/id would work, but not 3. Test included, but made a new file as I couldn't find an existing file that looked appropriate.

I couldn't get tox working with Python 2.6 locally (problem with installing PyYAML), and this test failed in every other version `test_iso (tests.functional.test_timeformat.TestCLITimestampParser) ... FAIL` but hopefully OK.

### DIFF
--- a/awscli/customizations/datapipeline/__init__.py
+++ b/awscli/customizations/datapipeline/__init__.py
@@ -306,7 +306,10 @@ class ParameterValuesInlineArgument(CustomArgument):
                 key = argument_components[0]
                 value = argument_components[1]
                 if key in parameter_object:
-                    parameter_object[key] = [parameter_object[key], value]
+                    if isinstance(parameter_object[key], list):
+                        parameter_object[key].append(value)
+                    else:
+                        parameter_object[key] = [parameter_object[key], value]
                 else:
                     parameter_object[key] = value
             except IndexError:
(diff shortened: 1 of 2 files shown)


fffergal:
Is there anything I can do to help this get merged?

fffergal:
Let me know if there's anything you need from me to move on with this PR.

fffergal:
Is there anything else I can do before a reviewer looks at this? It's been a whille, shall I rebase and check it merges OK still?

fffergal:
What's the process for this PR now. Is there anything I need to do or shall I just leave it?

fffergal:
Hello,

I would like to remove the code from our tools that creates JSON files to work around this CLI arguments parsing bug. Unfortunately the workaround has spread to other tools too, and makes it a bit harder for new contributors to get familiar with the code. We've ruled out running our own fork of awscli as too difficult to manage, so would be good to fix this in the real awscli.

If there's anything I can do to help move this PR forward I would be only too happy, I'm invested in fixing the bug.

While I understand awscli is free and open source, the ease of use and comprehensive service support of awscli is part of AWS' value to us. It's a shame we can identify a fix but not get it merged. Could I be right in suspecting that DataPipeline is a lower priority now that Glue handles some of the same cases?

Thank you for any help.

Fergal Hainey

joguSD:
Verified the bug and the fix, thanks for the PR! I'll merge this later today when I get a chance.

fffergal:
Thank you very much @joguSD, no problem.

Am I right in thinking this is in release 1.15.62?

joguSD:
@fffergal Yep, this should have been released in 1.15.62.

PR #3186, opened 2018-03-08
Don't call `describe_stacks` in `aws cloudformation deploy`

shuber:
We're running into the following throttling errors when trying to call `aws cloudformation deploy` multiple times both sequentially across multiple deployment machines and in parallel on a single machine.

```
Waiting for changeset to be created..
An error occurred (Throttling) when calling the DescribeStacks operation (reached max retries: 4): Rate exceeded
Command exited with non-zero status 255
```

We've opened support case 4890801861 in regards to this issue where it was determined that the `aws cloudformation deploy` command is actually calling the cloudformation `describe_stacks` command under the hood to determine if a stack exists or not.

Since we have so many CloudFormation stacks the `describe_stacks` API can be slow and ineffecient so this pull request replaces those calls with ones to `describe_stack_resources` instead. It retains the same behavior for determining if the deploy should be performed with either a `create` or an `update` command for the following scenarios:

* If the stack does not exist then `create` is used
* If the stack exists in `REVIEW_IN_PROGRESS` state then `create` is used
* Otherwise if the stack already exists then `update` is used

Here are a couple links I found where this issue was previously reported.

* https://aws.uservoice.com/forums/598381-aws-command-line-interface/suggestions/33168517--aws-cloudformation-deploy-wait-describe-stack
* https://github.com/aws/aws-cli/issues/2948

### DIFF
--- a/awscli/customizations/cloudformation/deployer.py
+++ b/awscli/customizations/cloudformation/deployer.py
@@ -43,21 +43,18 @@ class Deployer(object):
         :return: True if stack exists. False otherwise
         """
         try:
-            resp = self._client.describe_stacks(StackName=stack_name)
-            if len(resp["Stacks"]) != 1:
-                return False
+            resp = self._client.describe_stack_resources(StackName=stack_name)
 
-            # When you run CreateChangeSet on a a stack that does not exist,
-            # CloudFormation will create a stack and set it's status
-            # REVIEW_IN_PROGRESS. However this stack is cannot be manipulated
-            # by "update" commands. Under this circumstances, we treat like
-            # this stack does not exist and call CreateChangeSet will
+            # When you run CreateChangeSet on a stack that does not exist,
+            # CloudFormation will create an empty stack and set its status to
+            # REVIEW_IN_PROGRESS. However this stack cannot be manipulated by
+            # "update" commands. Under these circumstances, we treat it like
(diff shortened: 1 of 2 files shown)


shuber:
Any feedback on this? Issues, concerns, comments? What can I do to help get this pushed thru?

joguSD:
cc @sanathkr

wjordan:
> Since we have so many CloudFormation stacks the `describe_stacks` API can be slow and ineffecient so this pull request replaces those calls with ones to `describe_stack_resources` instead.

Is there any public documentation, quotable correspondence from AWS support or anything else you can link/reference to publicly verify that `describe_stack_resources` is indeed more 'efficient' than `describe_stacks`, and would generate `Rate exceeded` throttling API errors less frequently? I'm mostly just curious for more details about the related AWS internals this PR is hinting at.

shuber:
@wjordan We currently have thousands of CloudFormation stacks and I can only confirm that we have not had any of these throttling errors after switching our logic to hit `describe_stack_resources` instead of `describe_stacks`.

mandeepbal:
I'm also seeing this issue on the latest version of the cli. We are running the deploy command in parallel from one machine. Will this PR be merged anytime soon?

stevemk14ebr:
Please merge this!

PR #3187, opened 2018-03-08
Fix get-lifecycle-policy-preview example command

mlapierre:
Just a minor correction for the get-lifecycle-policy-preview example. It was using the get-lifecycle-policy command.


kdaily:
It looks like this has subsequently been fixed. Closing!

PR #3193, opened 2018-03-10
Docs: Add Homebrew installation

### LINKED ISSUE #727: Install aws-cli using Homebrew
My [awscli formula](https://github.com/hydrajump/homebrew/blob/ec1379d4df6e436142ef3f84d56eb68a62dd7ffa/Library/Formula/awscli.rb) has been merged and is officially available with [Homebrew](http://brew.sh/).

It would be useful to add Homebrew as an installation option in the README.

As a suggestion something like this,
## Homebrew on Mac OS X

aws-cli can also be installed using the [Homebrew](http://brew.sh/) package manager on Mac OS X:

`brew install awscli`

If you want the development version of aws-cli:

`brew install awscli --HEAD`

After installation brew will provide instructions for installing completion and where to find the examples on how to use aws-cli.

dijonkitchen:
It's updated regularly in less than a few days: https://github.com/aws/aws-cli/issues/727#issuecomment-321920276

Resolves #727

Reference:
http://formulae.brew.sh/formula/awscli


README.rst
```diff
@@ -47,7 +47,7 @@ you can run::
 
 .. note::
 
-    On OS X, if you see an error regarding the version of six that came with
+    On macOS, if you see an error regarding the version of six that came with
```
ms-ati: Why change "OS X" to "macOS" here?

README.rst
```diff
@@ -63,6 +63,10 @@ awscli directory structure on your workstation, you can just run::
 If you want to run the ``develop`` branch of the CLI, see the
 "CLI Dev Version" section below.
 
+Althought not officially maintained by AWS, for macOS users, you can also use  `Homebrew`_::
```
ms-ati: Typo: "Althought"

Wouldn't a link to Homebrew be useful here?

ms-ati: Nevermind, I see it below. Deleting.

ms-ati: Typo: "Althought"

README.rst
dijonkitchen: That's what it's called now: 
https://www.apple.com/macos/high-sierra/
https://www.wired.com/2016/06/apple-os-x-dead-long-live-macos/

jhabdas: iOS, macOS, tvOS. Makes sense. Wonder what they'll call the successor to the iPhone X…

jhabdas:
@purp mentioned in comment linked by @dijonkitchen (awesome handle btw, now I'm hungry) the brew formula is updated no later than 3 days after the CLI updates. IMHO that's too long. When the CLI is updated it should be deployed everywhere instantly. The only way I see to do that right now is to not use Homebrew for CLI unless you're OK with the security and operational risks involved which, I imagine, most enterprises using the CLI would not be.

dijonkitchen:
@jhabdas can we just clarify that in the Readme instead of not including it at all. As @purp mentioned, on average it's around 14 hours; this or the 3 day timeline seems reasonable for most users and many are having problems with the usage outside of `brew`.

jhabdas:
@dijonkitchen If it were up to me (and I had any weight at Amazon—I don't) I'd be working to streamline the entire process. Something is obviously broken and pushing on this issue is definitely the way to go to help shake things out. Still thinking personally an evergreen AWS CLI wrapped in a docker container then wrapped into a brew thingie would be the way to go. Just thoughts though. Thank you for continuing to push. But what's the hold up which causes a 14 hour delay? Smells like manual effort.

dijonkitchen:
@kyleknap ?

jacoscaz:
IMHO, not having a reference to the brew package in the README makes things worse as users are less likely to find out the security implications of using the brew version when they eventually stumble upon it.

jhabdas:
I don't know what's worse, the lack of lip service from someone at Amazon or the fact they don't tend to their open source repos on GitHub in a timely manner. :trollface:

dijonkitchen:
@stealthycoin @joguSD @JordonPhillips ?

russelldavis:
Counterpoint: the [Elastic Beanstalk docs](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-install-osx.html) do mention Homebrew installation, but that's actually causing users to get an old version, as the [homebrew formula there](https://formulae.brew.sh/formula/aws-elasticbeanstalk) is 4 months old and 3 versions behind (3.14.6 vs the [latest 3.14.9](https://pypi.org/project/awsebcli/)). I'm completely in favor of installing via homebrew, but I think Amazon first needs to add their own process for automatically keeping the formulas up to date, rather than relying on manual work by outside parties.

ms-ati:
I'd like to 2nd @russelldavis. I think the best way to interact with the developer community on Mac OS X is for Amazon to take on the (extremely low lift) responsibility for maintaining and updating the Homebrew formula. 

As a positive example, this is how Heroku CLI is distributed, and it works extremely well.

tim-finnigan:
Thank you for creating this PR. I brought this up for discussion with the team and the consensus was that we don't plan to document support for unofficial installation methods. But we can continue using https://github.com/aws/aws-cli/issues/727 to track the feature request for Homebrew support.

PR #3197, opened 2018-03-13
Ensure python-dateutil version is Python 2.6 compatible

joguSD:
This force downloads `python-dateutil` v2.6.1 to ensure that we have a Python 2.6 compatible version of the package.

This ensures that no matter what version of Python is used to create the bundled installer v2.6.1 of `python-dateutil` will be present in the packages folder. Depending on the Python version used a more recent version of `python-dateutil` may additionally be present.


joguSD:
@pganssle This doesn't fix the version to 2.6.1. This script creates a bundled installer that includes all of the dependencies required to install the CLI. This ensures that version 2.6.1 is included regardless of what version of Python was used to generate the bundled installer. This does not mean that a newer version won't be present, just that 2.6.1 is *always* present and `pip` can choose the right version with the customer runs the bundled installer.

PR #3209, opened 2018-03-20
added support for using stdin and pipes

attie:
I'd like to use `aws` in a pipeline, as per [this question](https://superuser.com/q/1306071/707676).
It seems that `aws` will read the indicated file twice, using the second dataset for operation - in a pipeline, the second dataset will be empty...

I've implemented something basic that works - use the `stdin://` prefix/schema.

### DIFF
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -272,10 +272,28 @@ def get_pipe(prefix, path):
         PIPE_CACHE[path] = get_file(prefix, path, mode = 'r')
     return PIPE_CACHE[path]
 
+STDIN_CACHE = None
+def get_stdin(prefix, path):
+    """stdin can only be read once, to completion...
+
+    To facilitate this, we follow get_pipe()'s example, and maintain a cache.
+    To support Windows we do not use the "file" /dev/stdin, but rather sys.stdin
+
+    As we're not using get_file(), we need to re-implement exception handling
+    """
+    global STDIN_CACHE
+    try:
+        if STDIN_CACHE is None:
+            STDIN_CACHE = ''.join(sys.stdin.readlines())
+        return STDIN_CACHE
+    except (OSError, IOError):
+        raise ResourceLoadingError('Unable to read standard input: %s' % ( e ))
+
 LOCAL_PREFIX_MAP = {
     'file://': (get_file, {'mode': 'r'}),
     'fileb://': (get_file, {'mode': 'rb'}),
     'pipe://': (get_pipe, {}),
+    'stdin://': (get_stdin, {}),
 }
 
 
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -260,17 +260,22 @@ def get_uri(prefix, uri):
     except Exception as e:
         raise ResourceLoadingError('Unable to retrieve %s: %s' % (uri, e))
 
-STDIN_CACHE = None
-def get_stdin(prefix, path):
-    global STDIN_CACHE
-    if STDIN_CACHE is None:
-        STDIN_CACHE = get_file(prefix, prefix + '/dev/stdin', mode = 'r')
-    return STDIN_CACHE
+PIPE_CACHE = {}
+def get_pipe(prefix, path):
+    """Pipes can only be read once, to completion...
+
+    To facilitate this, we will maintain a cache of pipes that have been read.
+    Expected use includes /dev/stdin
+    """
+    global PIPE_CACHE
+    if path not in PIPE_CACHE:
+        PIPE_CACHE[path] = get_file(prefix, path, mode = 'r')
+    return PIPE_CACHE[path]
 
 LOCAL_PREFIX_MAP = {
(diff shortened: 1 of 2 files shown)


attie:
I'd appreciate ideas for testing this... would redirecting away from `/dev/stdin` to another pipe-like node be acceptable?
Of course I'd need to call `get_paramfile()` twice on input that is exhausted after the first call.

attie:
I've swapped the `stdin://` schema for `pipe://`... this is much more flexible, and also allows tests to be implemented nicely.

Example usage would be:
``` bash
cat ./mytask.json \
    | aws ecs register-task-definition --cli-input-json file:///dev/stdin
```

rianwouters:
Why not use ```''.join(sys.stdin.readlines())``` to read stdin so that it works on all platforms.

```
def get_stdin(prefix, path):
   STDIN_CACHE = None
    try:
        global STDIN_CACHE
        if STDIN_CACHE is None:
            STDIN_CACHE = ''.join(sys.stdin.readlines())
        return STDIN_CACHE
     except (OSError, IOError):
        raise ResourceLoadingError('Unable to read standard input')

LOCAL_PREFIX_MAP = {
    'file://': (get_file, {'mode': 'r'}),
    'fileb://': (get_file, {'mode': 'rb'}),
    'stdin:': (get_stdin, {}),
}
```

attie:
@rianwouters - I'm not sure how that's any better... I'm just reusing `get_file()` which is an existing function and does the job just fine!

This also removes support for other files other than `stdin` that also behave in a read-once way.

attie:
Also, who do I have to poke around here to get this merged? :stuck_out_tongue_closed_eyes:

rianwouters:
@attie does your solution work on Windows?

attie:
@rianwouters - fair point 👍

attie:
@rianwouters - fair point 👍 ... any thoughts on how to test this? (i.e: stdin)

rianwouters:
Please also consider https://github.com/aws/aws-cli/commit/43752cc#commitcomment-31448084

@attie For testing stdin, I believe you can just stub `sys.stdin.readlines` and use most of `test_does_use_file_prefix`

Something like:
```

stdinLines = ['stdin content']
def stdinReadlinesStub:
  r = stdinLines
  stdinLines = []
  return r

def test_does_use_stdin(self):
        stdinReadlines = sys.stdin.readlines
        sys.stdin.readlines = stdinReadlinesStub
        self.assert_param_expansion_is_correct(
            provided_param='stdin:',
            expected_param='stdin content'
        )
        sys.stdin.readlines = stdinReadlines
```

stealthycoin:
I don't think this is something we can do as a general feature in the CLI. The CLI has cases where it blocks execution and drops into an interactive mode with the user. Determining how to deal with that when one or more of the parameters are using `pipe://` would not be easy.

In that light I think it makes more sense to make a special case for `--cli-input-json` where it accepts `-` as a value that means "read stdin". Much like `aws s3 cp - s3://mybucket/stream.txt` will copy stdin to a bucket.

What do you think of that modification?

attie:
> Determining how to deal with that when one or more of the parameters are using pipe:// would not be easy.

This should be possible, though may be awkward - for example, `sudo` (and others) is able to correctly re-acquire the terminal for input:

```
$ sudo cat <<<"Hello"
[sudo] password for attie:
Hello
```

> In that light I think it makes more sense to make a special case for --cli-input-json where it accepts - as a value that means "read stdin".

I'm not convinced that's different?... admittedly it would be restricted to `--cli-input-json`, but can you be sure that this argument is mutually exclusive of interactive operation?

tim-finnigan:
Thanks for the PR and for your patience. Our team recently put out a [contribution guide](https://github.com/aws/aws-cli/blob/v2/proposals/contribution-guide.md) detailing improvements to the contribution process. I have created issue https://github.com/aws/aws-cli/issues/7850 for this PR. More discussion is needed here regarding use cases and possible implementations, and that is better suited for an issue than in the PR. For the time being, I will mark this PR as a draft.

RyanFitzSimmonsAK:
Hi @attie, thanks for creating this PR and for your patience. Could you please rebase this PR against the latest branch for the team’s further consideration going forward?

attie:
> Could you please rebase this PR against the latest branch ...

Done, but I've not re-tested... there are some issues with this implementation, dicussed above.

tim-finnigan:
Thanks for following up and continuing the discussion here. Because there are issues with the implementation as mentioned above, we have decided to close this PR for now. But we can continue using https://github.com/aws/aws-cli/issues/7850 to track the feature request. Others can 👍 that feature request and share use cases or proposed implementations there. I think more discussion is needed regarding the modification [suggested here](https://github.com/aws/aws-cli/pull/3209#issuecomment-1098522868) in an earlier comment, as well as the concerns shared in the subsequent [comment](https://github.com/aws/aws-cli/pull/3209#issuecomment-1105373435).

PR #3226, opened 2018-03-29
Update example for acm update-certificate-options

pdtcorrea:
fix the example for disabling CertificateTransparencyLoggingPreference


kdaily:
It looks like this has been addressed by the documentation writers. Thanks for bringing it to our attention!

PR #3242, opened 2018-04-05
Tolerate plugins that are not available

rhcarvalho:
When using multiple installs of the `aws-cli`, e.g., in multiple virtualenvs, Homebrew and system Python, some plugins may not be available in all of the installations.

Having the plugin listed in `~/.aws/config`, even if it is not intended to be used, prevents the `aws` command from working with an `ImportError` message.

This change makes the plugin system tolerate missing plugins.

### DIFF
--- a/awscli/plugin.py
+++ b/awscli/plugin.py
@@ -54,10 +54,13 @@ def _import_plugins(plugin_names):
     plugins = []
     for name, path in plugin_names.items():
         log.debug("Importing plugin %s: %s", name, path)
-        if '.' not in path:
-            plugins.append(__import__(path))
-        else:
-            package, module = path.rsplit('.', 1)
-            module = __import__(path, fromlist=[module])
-            plugins.append(module)
+        try:
+            if '.' not in path:
+                plugins.append(__import__(path))
+            else:
+                package, module = path.rsplit('.', 1)
+                module = __import__(path, fromlist=[module])
+                plugins.append(module)
+        except ImportError:
+            log.debug("Could not import plugin %s: %s", name, path)
     return plugins


rhcarvalho:
**Travis failures**

Same test fails in all Python 3.x versions:

- Python 3.3: FAIL: test_can_stream_to_file (tests.functional.s3api.test_select_object_content.TestGetObject)
https://travis-ci.org/aws/aws-cli/jobs/362877791#L7829
- Python 3.4: FAIL: test_can_stream_to_file (tests.functional.s3api.test_select_object_content.TestGetObject)
https://travis-ci.org/aws/aws-cli/jobs/362877793#L7829
- Python 3.5: FAIL: test_can_stream_to_file (tests.functional.s3api.test_select_object_content.TestGetObject)
https://travis-ci.org/aws/aws-cli/jobs/362877795#L7829
- Python 3.6: FAIL: test_can_stream_to_file (tests.functional.s3api.test_select_object_content.TestGetObject)
https://travis-ci.org/aws/aws-cli/jobs/362877796#L7825

rhcarvalho:
Apparently, the Travis failures here are because the develop branch is broken in https://travis-ci.org/aws/aws-cli/builds/362815143

rhcarvalho:
@jamesls looks like you were the only one to ever touch the `plugin.py` file and you're still actively contributing to the project. Could you please help me review this PR?

stealthycoin:
The main issue here is we do not want the case where a plugin fails to load and silently does the wrong thing. This could be very bad of the plugin was involved with credentials and you do something in the wrong account. So ignoring or just warning isn't really a viable general solution here. What might be a better option is providing some configuration option on a per-plugin level that says this plugin is allowed to not be present, so if you are ok with those risks you are opting into them, otherwise you are forced to have the plugin available. 

I will open an issue as a feature request where this can be tracked.

PR #3253, opened 2018-04-09
CodeCommit examples

AWS-SF:
Update the examples to reflect the CLI commands for pull requests, creating and editing files, and commenting.


AWS-SF:
Hi. I'm not sure why the checks failed, except that a number of pull requests had checks that failed right around the same time mine did. Could you please rerun the checks and/or advise how to work around this issue?

kyleknap:
Hey @AWS-SF thanks for the doc update. Would it be possible to do the following:

* Make sure there is only [one commit](https://github.com/aws/aws-cli/pull/3253/commits/00c6e0f55aa4dc579b8b49986bdd4f516cfd0ccf) in this PR. Right now it looks like the commits are out of sync with the develop branch so there is over 250 commits in this PR.
* Looks like the Travis build is failing as well because there is a snippet with non-ascii in it. Can you update it to not use ascii? We have the constraint because we are currently only able to render ascii text in the docs.

PR #3265, opened 2018-04-13
Add ONEZONE_IA to s3 transfer commands

### LINKED ISSUE #3252: ONEZONE_IA should be an allowed parameter to --storage-class
I'm using CLI v1.15.2 and according to the [docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/copy-object.html), `ONEZONE_IA` is one of the allowed parameters to `--storage-class`, yet when I use it, I get the error:

```
aws: error: argument --storage-class: Invalid choice, valid choices are:

STANDARD                                 | REDUCED_REDUNDANCY
STANDARD_IA
```

stealthycoin:
Add the ONEZONE_IA option to the argument --storage-class on the three
s3 transfer commands sync, mv and cp.

fixes #3252

### DIFF
--- a/.changes/next-release/enhancement-s3-99343.json
+++ b/.changes/next-release/enhancement-s3-99343.json
@@ -1,0 +1,5 @@
+{
+  "category": "s3", 
+  "type": "enhancement", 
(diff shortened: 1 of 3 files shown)


JordonPhillips:
ignore codecov, it's a false positive

PR #3285, opened 2018-04-24
Add lock to DatabaseRecordWriter

kyleknap:
This will allow the removal of the lock from the global history recorder as the DB writer is the only component currently that is definitely not thread safe and needs a lock. The reader may technically be not thread safe as well but it is only ever used by a single thread.

Related botocore PR that removes the lock: https://github.com/boto/botocore/pull/1447

Either this should get merged first or both roughly at the same time.


kyleknap:
I also did some manual testing with the `s3 cp` commands to make sure there are no concurrency issues. Works well for me. Merging.

PR #3287, opened 2018-04-25
Updated Longer ID examples

meiringa:
Updated examples for Longer IDs


joguSD:
Can you squash this down to one commit?

PR #3293, opened 2018-04-27
Support display of large man pages (Fix #3292)

### LINKED ISSUE #3292: help command blocks for larger content
Affected environment:
```
aws-cli/1.15.10 Python/2.7.10 Darwin/17.5.0 botocore/1.10.10
```
I'm actively tinkering using Python 3.5 but the behavior seems consistent between them (though I'd guess the OS is fairly significant).

Some help pages hang rather than rendering, such as:
`aws dynamodb scan help`

mwhipple:
Simple'ish approach. An overall nice solution may be to just pass control back to the shell sooner and allow the interactive mode buffering to kick in (but that seems more daunting).

### DIFF
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -15,6 +15,7 @@ import os
 import sys
 import platform
 import shlex
+import threading
 from subprocess import Popen, PIPE
 
 from docutils.core import publish_string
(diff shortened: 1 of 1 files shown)


mwhipple:
this is missing a couple things, adjustment incoming

mwhipple:
Closing as I won't be likely to isolate whether this is just a system misconfiguration in the near foreseeable future.

PR #3316, opened 2018-05-09
Fix --query statement in command example

stig:
Without this change the example outputs the parameter _value_
for both the Name and Value keys, which is not expected.


stig:
Ping @mwunderl since you created this file, and @dougtoppin since you added the example I'm correcting: I assume this is what you meant?

stig:
Ping @awstools

dougtoppin:
Sorry @stig , just saw your question.  Yes, the correct usage example should be what you suggested:

aws ssm get-parameters --names key1 key2 --query "Parameters[*].{Name:Name,Value:Value}"

PR #3331, opened 2018-05-17
Properly write profiles with spaces to shared credentials

### LINKED ISSUE #3314: aws configure get and aws configure set with multiword profile names are inconsistent
It seems that `aws configure set --profile "two words"` will add single quotes around the profile name, but `aws configure get --profile "two words"` will search for a profile name that does not have single quotes around the profile name.

These two methods should behave in a similar manner.

To reproduce:

```
$ aws --version
aws-cli/1.15.10 Python/3.6.5 Darwin/17.4.0 botocore/1.10.10
$ aws configure set aws_access_key_id test --profile "test profile"
$ aws configure get aws_access_key_id --profile "test profile"
The config profile (test profile) could not be found
$ aws configure get aws_access_key_id --profile "'test profile'"
test
```

joguSD:
This refactors the set command to delay turning the profile into a section until we know whether or not we're writing to the credentials file or to the shared config file.

Fixes #3314

### DIFF
--- a/awscli/customizations/configure/set.py
+++ b/awscli/customizations/configure/set.py
@@ -46,10 +46,14 @@ class ConfigureSetCommand(BasicCommand):
             config_writer = ConfigFileWriter()
         self._config_writer = config_writer
 
+    def _get_config_file(self, path):
+        config_path = self._session.get_config_variable(path)
+        return os.path.expanduser(config_path)
+
     def _run_main(self, args, parsed_globals):
         varname = args.varname
         value = args.value
-        section = 'default'
+        profile = 'default'
         # Before handing things off to the config writer,
         # we need to find out three things:
         # 1. What section we're writing to (section).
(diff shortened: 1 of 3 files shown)


awscli/customizations/configure/set.py
```diff
@@ -60,43 +64,44 @@ def _run_main(self, args, parsed_globals):
...
                 if len(parts) == 2:
                     value = {parts[1]: value}
             elif len(parts) == 2:
                 # Otherwise it's something like "set preview.service true"
                 # of something in the [plugin] section.
-                section, varname = parts
-        config_filename = os.path.expanduser(
-            self._session.get_config_variable('config_file'))
-        updated_config = {'__section__': section, varname: value}
+                profile, varname = parts
```
kyleknap: Talked about this offline but it's a little weird to be calling the predefined sections profiles. But sounds like it will be tricky to switch the names here so let's leave a comment for future reference.

awscli/customizations/configure/set.py
```diff
@@ -46,10 +46,14 @@ def __init__(self, session, config_writer=None):
...
+        return os.path.expanduser(config_path)
+
     def _run_main(self, args, parsed_globals):
         varname = args.varname
         value = args.value
-        section = 'default'
+        profile = 'default'
         # Before handing things off to the config writer,
         # we need to find out three things:
         # 1. What section we're writing to (section).
```
kyleknap: Just a reminder to change the comment to this:
```suggestion
         # 1. What section we're writing to (profile).
```

kyleknap (approved):
Looks fine still. Let's add a changelog entry to this before we merge it. :ship:

PR #3333, opened 2018-05-18
Added and updated Route 53 examples

skralik:


kyleknap:
Hi @skralik thanks for the contribution. Looking over the examples, one thing that I noticed is that there are a quite a few examples that either just show the syntax required or lack the output of what it looks like. Most CLI examples include actual inputs and outputs. Having the syntax is necessary, but the CLI should be auto-generating most of this. So it would be awesome if all of these examples are updated to display actual input and output of running a CLI command because it better helps supplement the documentation as it is really powerful to be able to copy, paste, modify (hopefully only a little), and run a command directly.

skralik:
Hi Kyle,

At least one Route 53 API command, ChangeResourceRecordSets, doesn't work well with autogenerated syntaxes. The doc includes one autogenerated syntax, but that syntax is completely unusable because no type of resource record set uses all of the elements in the syntax. I created 11 examples to cover all of the different types of records that you can create.

I took another look at the examples, and I'm not seeing any that aren't functional with just minor tweaks. I'm also including information that the CLI doc doesn't, including:
- The fact that you need to submit API requests for domain registration to the us-east-1 region.
- How you specify an input file. This is a regular source of pain for our customers, but either the CLI doc doesn't say anything about file:// or no one can find it. 
- The syntax of the content in the input file. (This is autogenerated for some CLI commands but not all.) This is not the same as the syntax for the command, and for commands that require a lot of elements in the input, it saves customers the trouble of figuring out what the JSON format is.

Few if any commands have anything useful in the output. In most cases, they either return the information that the command succeeded, or they regurgitate the information that you entered in the request, neither of which seemed useful enough to bother with given the 80+ commands that I was banging out examples for. In the remaining cases—get and list commands—they return the information that you asked for.

I don't have time right now to revisit those commands to add outputs; Route 53 is releasing two new services for re:Invent, and I'm slammed. Customers would get a lot of benefit from what I have written, but if these examples don't meet the requirements for CLI examples, I'll just have to try again after re:Invent.

Thanks.

Scott

PR #3357, opened 2018-06-05
Fixing a TypeError when running from context of library

trega3:
When running the S3 Sync from the context of a library, a TypeError occurs here because the BucketLister erroneously assumes the LastModified value is a string. Boto3's default behavior is for LastModified to be a datetime, and only in the case where one is running the aws cli from the context of a command line application is it a string.


kdaily:
Thanks again for the PR. Sorry for the delay in resolving, but after review this comment still holds here:

https://github.com/aws/aws-cli/issues/3358#issuecomment-395206882

Issue #2874, opened 2017-10-06
Add --no-overwrite option to aws s3 cp/mv


alexjurkiewicz:
It would be nice to have a convenience function `--no-overwrite` for `aws s3 cp/mv` commands, which would check the target destination doesn't already exist before putting a file into an s3 bucket.

Of course this logic couldn't be guaranteed by the AWS API (afaik...) and is vulnerable to race conditions, etc. But it would be helpful to prevent unintentional mistakes!

kyleknap:
Marking as a feature request. The tricky part if we did it in `cp` or `mv` is that the CLI may have to query S3 to see if the file exists before trying to upload it. So it may make more sense to add it to `sync` as it already does that.

sgrimm-sg:
I'd like to see this in `cp` and/or `mv` as well.

The reason I don't use `sync` for this right now is that `sync` has major performance problems if the destination bucket has a lot of existing files under the target directory.

When you run `aws s3 cp --recursive newdir s3://bucket/parentdir/`, it only visits each of the files it's actually copying.

When you run `aws s3 sync newdir s3://bucket/parentdir/`, it visits the files it's copying, but also walks the entire list of files in `s3://bucket/parentdir` (which may already contain thousands or millions of files) and gets metadata for each existing file.

On a sufficiently large destination bucket, `aws s3 cp --recursive` can take seconds and `aws s3 sync` can take hours to copy the same data.

Obviously fixing `sync` would be nice, but if adding a "check to see if the file already exists" query to `cp` is a more tractable problem than revamping the `sync` code to make it fast, it might make sense to do that instead.

smaslennikov:
I'm also very interested in this feature. An optional interactive prompt for overwriting files would also be nice to have.

shabeebk:
yes @sgrimm-sg, it makes sense. I am also interested to see CLI cp command which can actually handle these conditions.

jhoblitt:
It would be extremely useful for this to be an option on `aws s3 sync`.  `rsync` has this functionality available as `--ignore-existing`.  My preference would be try to use the same option names as `rsync` as I suspect there are a lot of folks already familiar with `rsync`.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

[...]

Issue #1449, opened 2015-08-15
aws s3 sync --existing


ChrisSLT:
Could you implement a feature like RYNC's --existing so that you can tell AWS to "skip creating files that do not exist yet on the destination"

Or is there a way to do this already?

The use case is you have just synced from machine 1 up to S3 and you have a huge dataset in S3.  Now you want to sync down from S3 to machine 2 so you're working with up to date files on machine 2.  But you only want to update files that are already on machine 2, you don't want to copy the huge dataset in S3 down.

Make sense?

jamesls:
Not currently possible to do with the CLI.  The closest thing you could is either:

a) only sync up to a specific prefix (s3://bucket/specific/dataset/)
b) only sync with `--exclude/--include` filters: `--exclude '*' --include '*.txt' s3://bucket/specific/prefix)

This could end up being a little tricky to implement efficiently, we may have to switch our internals to use HeadObject instead of ListObjects if `--existing` is specified.

Marking as a feature request.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

salmanwaheed:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  mkdirenv@gmail.com
    Domain salmanwaheed.info has exceeded the max emails per hour (151/150 (100%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from o6.sgmail.github.com ([192.254.113.101]:15527)
[...]

PR #3363, opened 2018-06-08
Add ignoreexisting to syncstrategy so files with the same name not sync'd

awuceres:
We would like a strategy so that once the files are sync'd once, it will not clobber local files as we modify them.

This strategy allows us to only transfer files that do not exist.

I might also be missing some places where I need to change the files.

### DIFF
--- a/awscli/customizations/s3/syncstrategy/ignoreexisting.py
+++ b/awscli/customizations/s3/syncstrategy/ignoreexisting.py
@@ -18,17 +18,17 @@ from awscli.customizations.s3.syncstrategy.base import BaseSync
 LOG = logging.getLogger(__name__)
 
 
-TRANSFER_ONCE = {
-    'name': 'transfer_once',
+IGNORE_EXISTING = {
+    'name': 'ignore_existing',
     'action': 'store_true',
     'help_text': ('If the file exists, do not sync.')
 }
 
 
-class TransferOnceSync(BaseSync):
+class IgnoreExistingSync(BaseSync):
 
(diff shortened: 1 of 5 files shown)


awuceres:
I have renamed transferonce to ignoreexisting to be more similar to rsync.

awuceres:
Comments please?  Is it something I'm missing?  Thanks.

awuceres:
Sorry to bother, but any comments?  
Is the code I wrote really bad, or I'm missing something, or you haven't gotten around to review yet, or you don't think this is a good feature, or should I file an issue to discuss this?
Much appreciated.

tim-finnigan:
Hi @awuceres, apologies that it has taken so long to get a response on this PR. I think it fell off the team's radar because it was not linked to a corresponding issue. I found a couple issues that look related:
* https://github.com/aws/aws-cli/issues/1449
* https://github.com/aws/aws-cli/issues/404
* https://github.com/aws/aws-cli/issues/2874

Although that last issue references `cp/mv`, the `sync` command is also mentioned several times throughout the issue (for example [here](https://github.com/aws/aws-cli/issues/2874#issuecomment-361306387)). You could try using the `--size-only` parameter as [suggested here](https://github.com/aws/aws-cli/issues/2874#issuecomment-710097606) as that might address your use case. 

All that being said, we want to consolidate related issues to help with tracking going forward. Please let us know if one of those existing issues or a new issue is the best place to track this feature request. This PR can be revisited in the future once we have a corresponding issue and more discussion regarding the use case and request.

PR #3365, opened 2018-06-08
Add region overrides to smoke test

JordonPhillips:
This lets us run most of the smoke tests in whatever region we want.

### DIFF
--- a/tests/integration/test_smoke.py
+++ b/tests/integration/test_smoke.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import re
+import os
 import random
 from nose.tools import assert_equal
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_smoke.py
```diff
@@ -108,13 +108,27 @@
...
+    'route53domains': 'us-east-1'
+}
+
+
+def _aws(command_string):
+    service = command_string.split()[0]
+    command = command_string
+    if service in REGION_OVERRIDES:
+        command += ' --region %s' % REGION_OVERRIDES[service]
+    return aws(command)
```
jamesls: aws() takes `env_vars` as a param.  why not just provide it there?

JordonPhillips: It doesn't make that much of a difference, but I can see where there could be some issues depending on how the smoke test is written. I'll get that changed.

JordonPhillips:
Updated

PR #3384, opened 2018-06-15
Paramfile handler object

stealthycoin:
Add configuration option to opt-out of url parameter fetching

By default the CLI will fetch the content from any string parameter that
starts with https:// or http:// and use the content as the parameter
rather than the URL itself. This adds a configuration option to block
this behavior and these two cases the same as any other string argument.

Specifically the newly added config option is the `cli_follow_urlparam`
option. It can be used to disable this behavior by adding the following
to the config file in `.aws/config`.

```
cli_follow_urlparam = false
```

If the value is set to `true` or if the config key is not present then the
default behavior is preserved.

### DIFF
--- a/.changes/next-release/enhancement-Argumentprocessing-46505.json
+++ b/.changes/next-release/enhancement-Argumentprocessing-46505.json
@@ -1,0 +1,5 @@
+{
+  "category": "Argument processing",
+  "type": "enhancement",
+  "description": "Added cli_follow_urlparam option in the config file which can be set to false to disable the automatic following of string parameters prefixed with http:// or https://. closes #2507 #3076 #2577. Further discussion #3398."
+}
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -12,11 +12,13 @@
 # language governing permissions and limitations under the License.
 import logging
 import os
+import copy
 
 from botocore.vendored import requests
 from awscli.compat import six
 
 from awscli.compat import compat_open
+from awscli.argprocess import ParamError
 
 
 logger = logging.getLogger(__name__)
@@ -121,7 +123,45 @@ class ResourceLoadingError(Exception):
     pass
 
 
-def get_paramfile(path):
+def register_uri_param_handler(session, **kwargs):
+    prefix_map = copy.deepcopy(LOCAL_PREFIX_MAP)
+    fetch_url = session.get_scoped_config().get(
+        'cli_follow_urlparam', 'true') == 'true'
+
+    if fetch_url:
+        prefix_map.update(REMOTE_PREFIX_MAP)
+
+    handler = URIArgumentHandler(prefix_map)
+    session.register('load-cli-arg', handler)
+
+
+class URIArgumentHandler(object):
+    def __init__(self, prefixes=None):
+        if prefixes is None:
+            prefixes = copy.deepcopy(LOCAL_PREFIX_MAP)
+            prefixes.update(REMOTE_PREFIX_MAP)
+        self._prefixes = prefixes
+
+    def __call__(self, event_name, param, value, **kwargs):
+        """Handler that supports param values from URIs."""
+        cli_argument = param
+        qualified_param_name = '.'.join(event_name.split('.')[1:])
+        if qualified_param_name in PARAMFILE_DISABLED or \
+                getattr(cli_argument, 'no_paramfile', None):
+            return
+        else:
+            return self._check_for_uri_param(cli_argument, value)
+
+    def _check_for_uri_param(self, param, value):
+        if isinstance(value, list) and len(value) == 1:
+            value = value[0]
+        try:
+            return get_paramfile(value, self._prefixes)
+        except ResourceLoadingError as e:
+            raise ParamError(param.cli_name, six.text_type(e))
+
+
+def get_paramfile(path, cases):
     """Load parameter based on a resource URI.
 
     It is possible to pass parameters to operations by referring
@@ -135,6 +175,10 @@ def get_paramfile(path):
     :param path: The resource URI, e.g. file://foo.txt.  This value
         may also be a non resource URI, in which case ``None`` is returned.
 
+    :type cases: dict
+    :param cases: A dictionary of URI prefixes to function mappings
+        that a parameter is checked against.
+
     :return: The loaded value associated with the resource URI.
         If the provided ``path`` is not a resource URI, then a
         value of ``None`` is returned.
@@ -142,7 +186,7 @@ def get_paramfile(path):
     """
     data = None
     if isinstance(path, six.string_types):
-        for prefix, function_spec in PREFIX_MAP.items():
+        for prefix, function_spec in cases.items():
             if path.startswith(prefix):
                 function, kwargs = function_spec
                 data = function(prefix, path, **kwargs)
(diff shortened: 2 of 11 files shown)


tests/unit/test_paramfile.py
```diff
@@ -10,14 +10,99 @@
...
+            prefixes = mock_get_paramfile.call_args[0][1]
+
+        self.assertEqual(len(prefixes), 4)
+        self.assertIn('file://', prefixes)
+        self.assertIn('fileb://', prefixes)
+        self.assertIn('http://', prefixes)
+        self.assertIn('https://', prefixes)
+
+    def test_uses_supplied_cases(self):
+        with mock.patch(
```
kyleknap: It would be great if we can minimize the use of patches. Specifically, it would be nice if we provide our own recording handler (or something like that) that we ensure gets called. This concerns are a little less for the default scenarios, but it may make sense to instead to switch some of these to functional tests (i.e. make sure each of the prefix types work).

.changes/next-release/enhancement-Argumentprocessing-46505.json
```diff
@@ -0,0 +1,5 @@
+{
+  "category": "Argument processing",
+  "type": "enhancement",
+  "description": "Added cli_follow_urlparam option in the config file which can be set to false to disable following of http"
```
kyleknap: Maybe we should update this to be "to disable following of parameter values that start with `http://` or `https://`"? Right now it is a little unclear of when it get follows.

tests/functional/test_paramfile.py
```diff
@@ -0,0 +1,186 @@
...
+logger = logging.getLogger(__name__)
+
+
+class FakeResponse(object):
+    def __init__(self, *args, **kwargs):
+        self.status_code = 200
+        self.text = 'function-name'
+
+
+class BaseTestCLIFollowParamURL(BaseCLIWireResponseTest):
```
kyleknap: There are not too many tests that use `BaseCLIWireResponseTest`. I would recommend using `BaseAWSCommandParamsTest` instead as that class gives you a lot more to work with. For example you do not have to set a wire response, and instead you can set a parsed response.

tests/functional/test_paramfile.py
```diff
@@ -0,0 +1,186 @@
...
+        self.files = FileCreator()
+        self.command_prefix = ['lambda', 'get-function', '--function-name']
+        self.start_patches()
+
+    def tearDown(self):
+        super(BaseTestCLIFollowParamURL, self).tearDown()
+        self.files.remove_all()
+        self.stop_patches()
+
+    def start_patches(self):
```
kyleknap: I prefer we avoid the pattern in general of starting the patches in the `setUp` and tearing them down in the `tearDown()`. Instead, let's either try to avoid patching at all cost for functional tests or use context managers. When they are in the `setUp` and `tearDown` it is pretty easy to forget to tear down a patch causing issues down the road when more tests get added.

tests/functional/test_paramfile.py
```diff
@@ -0,0 +1,186 @@
...
+        self.mock_open = mock_open(read_data='function-name')
+        self.open_patch = patch(
```
kyleknap: I would also prefer we just use the FileCreator for this instead of patching

tests/functional/test_paramfile.py
```diff
@@ -0,0 +1,186 @@
...
+    def stop_patches(self):
+        self.requests_patch.stop()
+        self.open_patch.stop()
+
+    def run_cmd_with_arg(self, argument):
+        command = self.command_prefix + [argument]
+        self.run_cmd(command)
+
+
+class TestCLIFollowParamURLDefault(BaseTestCLIFollowParamURL):
```
kyleknap: Hmmm. I think for these functional tests it would be better if we minimized the amount of assertions on mock we are making and be more explicit in each test case in what the inputs and outputs are. So specifically, I think we should try to copy the same style we do for the service-specific functional tests:
```python
class TestCLIFollowParamURLDefault(BaseTestCLIFollowParamURL):
    def test_with_http_prefix(self):
        command = "lambda get-function --function-name http://myremote.url"
        self.parsed_response = {}
        with mock.patch('awscli.paramfile.requests.get', spec=True) as mock_get:
            mock_get = FakeResponse(content='my-function-name')
            self.assert_params_for_cmd(command, {'FunctionName': 'my-function-name'}
            mock_get.assert_called_with('http://myremote.url')
```
Sure it is a little longer than what you currently have, but I find it a lot easier to read and understand. It also asserts that paramfile value were properly pulled into the client parameters. Would like to know your thoughts.

kyleknap:
Also it looks like the Travis build is failing.

.changes/next-release/enhancement-Argumentprocessing-46505.json
```diff
@@ -0,0 +1,5 @@
+{
+  "category": "Argument processing",
+  "type": "enhancement",
+  "description": "Added cli_follow_urlparam option in the config file which can be set to false to disable the automatic following of string parameters prefixed with http:// or https://"
```
jamesls: 1. Reference existing issues that this will close
2. We talked about creating a tracking issue where people could discuss this feature.  Did you create one?  If so, link that here as well.

awscli/paramfile.py
```diff
@@ -121,7 +123,45 @@ class ResourceLoadingError(Exception):
...
+        'cli_follow_urlparam', 'true') == 'true'
+
+    if fetch_url:
+        prefix_map.update(REMOTE_PREFIX_MAP)
+
+    handler = UriArgumentHandler(prefix_map)
+    session.register('load-cli-arg', handler)
+
+
+class UriArgumentHandler(object):
```
jamesls: s/Uri/URI/ pep8

mrburrito:
My two cents, since my major issue with this behavior is the SSM put-param in #2507.

The problem here is that you're treating this behavior as one-size-fits-all for the entire CLI when, in the majority of use cases -- local JSON parameters especially, you DO want to read from a remote URL and the user is expecting that.

The SSM, in particular, might be _the_ exception case where this should be inverted, though I'm sure there are others since I haven't exercised every command in the CLI.

stealthycoin:
@michaelize @mrburrito This is intended to give people a more reasonable way of opting out of this behavior in general, not in particular to SSM. SSM isn't the exception to the rule, there are quite a few https://github.com/aws/aws-cli/blob/develop/awscli/paramfile.py#L31-L119 and currently if we didn't hardcode it in that list you are out of luck, this gives people a way of unblocking themselves if/when we miss one.

On the flip side this doesn't block you from fetching urls even when disabled, since the CLI can be easily composed with other commands
```
$ aws ssm put-parameter --name foo --value $(curl http://...)
```

We can open a tracking feature request for more granular paramfile control configuration options: #3423

michaelize:
@stealthycoin Thanks for the explanation. It seems the current solution is good for now as it provides a consistent behavior for all the commands.

Bryksin:
Even though it is the solution and help, it still requires extra configuration outside of the SH script.
Basically, if I will do it for myself and share the sh script with my colleague it will not work for him because he doesn't have this option set to false!

Obviously I can automate and force set this option inside my SH script, but it is too much and my colleague might not be happy about it!

if user want the value of URL, he can use: 
```--value "$(curl <whatever_url>)"```

having it to resolve any URL by default is very not nice! and sorting it with additional config is also a not nice solution! 
possibly having extra argument like: 
```--plain-value "<url>"``` 
might workout, but I still believe it shouldn't resolve URL by default  as if u want to resolve it just use:
`"$(whatever you want to resolve)"`

PR #3385, opened 2018-06-16
Rename [wheel] section to [bdist_wheel] as the former is legacy

jdufresne:
For additional details, see:

https://github.com/pypa/wheel/blob/3dc261abc98a5e43bc7fcf5783d080aaf8f9f0cf/wheel/bdist_wheel.py#L127-L133

http://pythonwheels.com/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,4 +1,4 @@
-[wheel]
+[bdist_wheel]
 universal = 1
 
 


kdaily:
Since we do not support Python 2 anymore, this section isn't needed at all. Closing this and will open a separate one to remove the section altogether. Thanks!

PR #3402, opened 2018-06-22
Add configuration for disabling the URL param feature.

stealthycoin:
This change was initially reverted due to an integration tests failure (https://github.com/aws/aws-cli/pull/3401).
 
Specifically config commands often time have a scoped config that is not created at
the time initialize-session is called. In this case we just fall back to
the default behavior. This PR fixes that issue and adds unit tests to test all the config cases.


awscli/paramfile.py
```diff
@@ -125,8 +126,18 @@ class ResourceLoadingError(Exception):
 
 def register_uri_param_handler(session, **kwargs):
     prefix_map = copy.deepcopy(LOCAL_PREFIX_MAP)
-    fetch_url = session.get_scoped_config().get(
-        'cli_follow_urlparam', 'true') == 'true'
+    try:
```
kyleknap: Yeah I think it would be good if we moved one of the integration tests related to missing profile to be functional tests for this one so we can catch them in PR's.

jlforester:
In my opinion, this also needs exposed via a CLI option or environment variable so it can be managed without having to manage multiple profiles in the config file for when I want that behavior and when I don't.

PR #3416, opened 2018-06-28
fixing create cluster customization and examples

Pahtoe:
Changing items for Spot pricing, and opportunistic fixes to bring examples and descriptions up-to-date with current behavior. --jeffgoll@

### DIFF
--- a/awscli/customizations/emr/helptext.py
+++ b/awscli/customizations/emr/helptext.py
@@ -67,7 +67,7 @@ LOG_URI = (
 
 SERVICE_ROLE = (
     '<p>Specifies an IAM service role, which Amazon EMR requires to call other AWS services '
-    'such as EC2 on your behalf during cluster operation. This parameter '
+    'on your behalf during cluster operation. This parameter '
     'is usually specified when a customized service role is used. '
     'To specify the default service role, as well as the default instance '
     'profile, use the <code>--use-default-roles</code> parameter. '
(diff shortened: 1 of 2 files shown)


awscli/customizations/emr/helptext.py
```diff
@@ -67,7 +67,7 @@
 
 SERVICE_ROLE = (
     '<p>Specifies an IAM service role, which Amazon EMR requires to call other AWS services '
-    'such as EC2 on your behalf during cluster operation. This parameter '
+    'on your behalf during cluster operation. This parameter '
     'is usually specified when a customized service role is used. '
     'To specify the default service role, as well as the default instance '
     'profile, use the <code>--use-default-roles</code> parameter. '
```
stealthycoin: While we are at it how about we remove these extra `<code>` tags as well.

Pahtoe: The code tags actually work to provide formatting in the Web version, but only when they appear under the Synopsis. For example, see the text for the --release-version option at https://docs.aws.amazon.com/cli/latest/reference/emr/create-cluster.html. Where the confusion may be is that I removed the <code> tags from the TERMINATE_CLUSTERS string in this same file. That string is used in the description section here. https://docs.aws.amazon.com/cli/latest/reference/emr/terminate-clusters.html It looks like that formatting tag (along with the others) are not supported when the string is used as a description.

stealthycoin: Ah got it.

Issue #3131, opened 2018-02-06
package is always generating a new zip


rizidoro:
I build a Golang lambda function with the following template:

```
AWSTemplateFormatVersion : '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: Billing Api Create Application

Resources:
  BillingCreate:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: billing-create
      Handler: main
      CodeUri: ./build
      Runtime: go1.x
      Policies: AWSLambdaDynamoDBExecutionRole
```

Even when the code didn't change (Go build generates the same compiled code), `aws package` command generates a new zip file.

kyleknap:
Could you elaborate a little more on why it is an issue that a new zip is formed for every `package` command? It may be difficult to avoid making a new zip every time because the `package` command does an md5 of the zip it creates to see if it needs to reupload the code to s3 by comparing the two md5's.

rizidoro:
Imagine a scenario where you have a cloudformation file with more than one Lambda declared. 
For now, let's call it FN1 and FN2, one is at fn1.go file and the second at fn2.go.

I build both of them, which generates two binaries fn1  and fn2.

I run cloudformation package, and it generates 2 zip files and send them to S3.

One week later, I change the fn1 function, but not the fn2. My CI builds both of them, but only the first has a different MD5 (the second has the same MD5 as before). 

The problem here is the package command will generate a new zip for the second one too, even if the file did not changed, which causes all my cloudformation declared functions to be deployed.

jakul:
@rizidoro Can you download the zip files from S3, unzip them locally and diff them? Turns out I had one file which was actually different, because it included a "generated at" date which was being updated everytime I built the CloudFormation script

jakul:
You also need to check for timestamp differences amongst the files

jmassara:
I'm having the same problem as @rizidoro. I have cloudformation with more than one lambda. Each time I run package, it recreates the zip with a new md5 and uploads it - _regardless_ of whether or not the code actually changed.

If I download two separate zips uploaded to s3 during `package` and upzip them, the binaries are identical.

jmassara:
The issue is that `package` creates a zip file and does an md5 of that. If timestamps on files are different in the zip file, even if they are the same contents, the md5 is difrerent.  In the case of scripting languages, this is probably not an issue.  However with Go, each time you run `go build`, a new binary is created and thus a new timestamp.

[...]

PR #3436, opened 2018-07-09
Keep modification time when zipping single file to avoid new checksum

henrist:
The `cloudformation package` command will upload a zip, e.g. for source of a lambda function. When referencing a directory, the files stays in-place when being zipped. However, when referencing a single file, a temporary directory is created and the file is copied there before the directory is zipped.

Before, the file stats were not copied, leading to new modification times every time `cloudformation package` were run referencing a single file.

This change also copies the file stats, ensuring the zip file gets the same checksum and filename as previous runs, and avoids an necessary upload and change set of the CFN stack, unless the source file gets a new modification time or changes content.

Partly solves #3131. However, the source file still needs to have the same modification time as previous to avoid a new checksum. In a CI-environment the file will probably get a different modification time for every checkout. This can be mitigated by forcing a modification timestamp on the source file, e.g. by `touch -t '200001010000' file.py` before running `cloudformation package`.

I'm a bit unsure about the tests I've added. Don't like adding `time.sleep()`, but not sure how to test this otherwise.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -210,6 +210,7 @@ def copy_to_temp_dir(filepath):
     tmp_dir = tempfile.mkdtemp()
     dst = os.path.join(tmp_dir, os.path.basename(filepath))
     shutil.copyfile(filepath, dst)
+    shutil.copystat(filepath, dst)
     return tmp_dir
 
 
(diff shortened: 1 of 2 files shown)


stealthycoin:
Thanks for the PR! @sanathkr can you review this?

ashwgupt:
Hi Guys,

Is there any timelines on when will this fix be going in?

RyanFitzSimmonsAK:
Going to close this PR for the time being.  If someone later decides they want to pick this up, rebase it and it will be reopened for consideration.

PR #3444, opened 2018-07-12
Add template name defaults for SAM commands

Dunedan:
SAM strives to make deployment of serverless applications to AWS as
easy as possible. To make it even easier as it is right now, this commit
adds defaults for the template file names the `cloudformation package`
and `cloudformation deploy` commands take.

The default values for the template names are taken from the
[SAM Howto][1]. If there are better fitting values I'm of course
open to adjusting them.

Together with #3040 this commit would allow running
`cloudformation package` without the need to provide any arguments.

[1]: https://github.com/awslabs/serverless-application-model/blob/master/HOWTO.md

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -50,14 +50,6 @@ class DeployCommand(BasicCommand):
 
     ARG_TABLE = [
         {
-            'name': 'template-file',
-            'required': True,
-            'help_text': (
-                'The path where your AWS CloudFormation'
-                ' template is located.'
-            )
-        },
-        {
             'name': 'stack-name',
             'action': 'store',
             'required': True,
@@ -65,6 +57,15 @@ class DeployCommand(BasicCommand):
                 'The name of the AWS CloudFormation stack you\'re deploying to.'
                 ' If you specify an existing stack, the command updates the'
                 ' stack. If you specify a new stack, the command creates it.'
+            )
+        },
+        {
+            'name': 'template-file',
+            'required': False,
+            'default': 'packaged-template.yaml',
+            'help_text': (
+                'The path where your AWS CloudFormation template is located.'
+                ' Defaults to packaged-template.yaml.'
             )
         },
         {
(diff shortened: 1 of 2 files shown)


stealthycoin:
Thanks for the PR! @sanathkr thoughts?

stealthycoin:
Wait so just to make sure I understand this. The intention here is to make the default template name that comes out of `package` match the input of `deploy`? Package currently defaults to stdout though, so this doesn't seem like it makes it easier to use. Just changes the error you get. Am I misunderstanding?

Dunedan:
@stealthycoin This PR doesn't change the output location for the template generated by the `package`-command, so this still defaults to stdout. It just adds defaults for the input template locations for the `package` and `deploy` commands, which makes it possible to omit these parameters if the templates are already present at the expected locations.

ozbillwang:
@Dunedan 

I am not the contributor for this repo, but when go through your PR, I feel it isn't well organised, includes 
 some unnecessary changes

So if this is a one line change, just make it that line, why adjust the content sequence in json and touch other stuffs.

tim-finnigan:
Hi @Dunedan thanks for creating this PR and for your patience. Could you please rebase this PR against the latest branch for the team’s further consideration going forward?

Dunedan:
@tim-finnigan: Done. The changes are now rebased on `develop`. :slightly_smiling_face:

tim-finnigan:
Hello — thanks again for your patience and for rebasing. After further review with the team the consensus was that a tracking issue should be opened (I created one here: https://github.com/aws/aws-cli/issues/8015) to gather more feedback on this feature request. Also sharing this note mentioned there:
>For more context: the SAM CLI looks like for anything matching the pattern template `[yaml|yml|json]` for both its package and deploy command. If the AWS CLI were to introduce defaults we would want to copy what the SAM CLI does in order to avoid creating any new differences that end users have to keep track of. For now customers should 👍 this issue if they want to see this implemented.

In the meantime I will set this PR to Draft pending further review.

PR #3448, opened 2018-07-12
Add top level parameters for passing images to rekognition

stealthycoin:
Raw bytes cannot be passed to nested parameters. Rekognition has many
operations that accept an image as a nested parameter, as essentially a
tagged union between the raw local file, or an S3 location. In the CLI
the S3 location specification works, and the raw image bytes option does
not. This commit adds a new top level parameter that overwrites the
nested image bytes option, which allows for the use of the `fileb://`
prefix to load a file.

For example the detect-labels operation would look like this:

    aws rekognition detect-labels --image-file fileb://image.png

The new top level --image-file parameter fill in the Bytes value of the
existing --image parameter.

### DIFF
--- a/awscli/customizations/rekognition.py
+++ b/awscli/customizations/rekognition.py
@@ -1,0 +1,80 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.arguments import CustomArgument
+
+
+IMAGE_FILE_DOCSTRING = ('<p>The path to the image file you are uploading. '
+                        'Example: fileb://image.png</p>')
+IMAGE_DOCSTRING_ADDENDUM = ('<p>To specify a local file use <code>--%s</code> '
+                            'instead.</p>')
+
+
+class ParameterReplacement(object):
+    def __init__(self, source_param, target_param):
+        self.source_param = source_param
+        self.target_param = target_param
+
+
+FILE_PARAMETER_UPDATES = {
+    'compare-faces': [
+        ParameterReplacement('source-image', 'SourceImage'),
+        ParameterReplacement('target-image', 'TargetImage')
+    ],
+    'detect-faces': [ParameterReplacement('image', 'Image')],
+    'detect-labels': [ParameterReplacement('image', 'Image')],
+    'detect-moderation-labels': [ParameterReplacement('image', 'Image')],
+    'detect-text': [ParameterReplacement('image', 'Image')],
+    'index-faces': [ParameterReplacement('image', 'Image')],
+    'recognize-celebrities': [ParameterReplacement('image', 'Image')],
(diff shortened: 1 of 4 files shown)


awscli/customizations/rekognition.py
```diff
@@ -0,0 +1,80 @@
...
+        ParameterReplacement('source-image', 'SourceImage'),
+        ParameterReplacement('target-image', 'TargetImage')
+    ],
+    'detect-faces': [ParameterReplacement('image', 'Image')],
+    'detect-labels': [ParameterReplacement('image', 'Image')],
+    'detect-moderation-labels': [ParameterReplacement('image', 'Image')],
+    'detect-text': [ParameterReplacement('image', 'Image')],
+    'index-faces': [ParameterReplacement('image', 'Image')],
+    'recognize-celebrities': [ParameterReplacement('image', 'Image')],
+    'search-faces-by-image': [ParameterReplacement('image', 'Image')],
```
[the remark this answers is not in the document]
stealthycoin: That was my original plan. I went this route since it is simpler. I don't really have a strong opinion about this so I can be swayed to automatic vs partially automatic with the extra cease of `compare-faces` vs hard coded as it is now.

awscli/customizations/rekognition.py
```diff
@@ -0,0 +1,80 @@
...
+
+class ParameterReplacement(object):
+    def __init__(self, source_param, target_param):
+        self.source_param = source_param
+        self.target_param = target_param
+
+
+FILE_PARAMETER_UPDATES = {
+    'compare-faces': [
+        ParameterReplacement('source-image', 'SourceImage'),
```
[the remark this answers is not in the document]
stealthycoin: I was just leaving room to add stuff if I needed another mapping. I can swap it with a map now that it turns out I didn't need it.

awscli/customizations/rekognition.py
```diff
@@ -0,0 +1,79 @@
...
+FILE_PARAMETER_UPDATES = {
+    'compare-faces.source-image': 'source-image-bytes',
+    'compare-faces.target-image': 'target-image-bytes',
+    '*.image': 'image-bytes',
+}
+
+
+def register_rekognition_detect_labels(cli):
+    for target, new_param in FILE_PARAMETER_UPDATES.items():
+        operation, old_param = target.split('.')
```
jamesls: `rsplit('.', 1)`

awscli/customizations/rekognition.py
```diff
@@ -0,0 +1,79 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import re
+
+from awscli.arguments import CustomArgument
+
+
+IMAGE_FILE_DOCSTRING = ('<p>The path to the image file you are uploading. '
```
kyleknap: Probably should be reworded to "The content of the image file".

awscli/customizations/rekognition.py
```diff
@@ -0,0 +1,79 @@
...
+        argument_table[source_param].required = False
+        doc_addendum = IMAGE_DOCSTRING_ADDENDUM % new_param
+        argument_table[source_param].documentation += doc_addendum
+
+
+class ImageArgument(CustomArgument):
+    def __init__(self, name, source_param, **kwargs):
+        super(ImageArgument, self).__init__(name, **kwargs)
+        self._parameter_to_overwrite = reverse_xform_name(source_param)
+        print(source_param, self._parameter_to_overwrite)
```
kyleknap: Left over print statement.

kyleknap:
Looks like you addressed @jamesls comments as well. Should be good to merge.

PR #3440, opened 2018-07-10
Feature/cf package export local swagger includes transform

ekcrisp:
When providing a swagger definition for AWS::Serverless::Api resource using DefinitionUri you are unable to use cloudformation intrinsic functions. Several users have complained about this. One workaround is using AWS::Include transformation to reference a swagger doc from the inline property DefinitionBody where intrinsic functions can be used. This feature provides the same local file export for swagger documents referenced from an AWS::Include in the DefinitionBody as the existing DefinitionUri file export. Below are links to discussions where this feature has been mentioned.

https://github.com/awslabs/serverless-application-model/issues/160
https://github.com/awslabs/serverless-application-model/issues/8
https://github.com/awslabs/serverless-application-model/issues/345

Issue #2513, opened 2017-03-28
Allow `aws cloudformation package` to work with local paths in `AWS::Include` transforms


nikolay:

kyleknap:
Marking as feature request. @sanathkr what's your thoughts on this?

sanathkr:
Great idea. Let's do it. PRs please :-)

nikolay:
@kyleknap @sanathkr It looks like this won't be so easy to implement as you guys have made assumptions in the current code that won't work with `AWS::Include`, which can literally be placed anywhere.

nikolay:
Any luck? This would be a huge improvement of the CloudFormation workflow, which is the main obstacle of people adopting CloudFormation widely!

davidkelley:
Any update on this issue? It would be a fantastic feature to have.

lafiosca:
This would be a huge help for the use case from this example: https://github.com/awslabs/serverless-application-model/blob/master/examples/2016-10-31/api_swagger_cors/template.yaml

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

georgealton:
this would be a major benefit as @lafiosca mentions the exact same problem i'm trying to solve :)

ekcrisp:
@sanathkr @lafiosca @georgealton @nikolay 

[...]

PR #3454, opened 2018-07-16
Feature/use s3 uploader for cf package with include transforms

ekcrisp:
I recently opened a PR which added exports for include transform of inline swagger AWS::Serverless::Api resource. I've reconsidered and decided to add an implementation that will upload AWS::Include from anywhere in a cloud formation template for cloudformation package command. I have seen several discussions where users have asked for this feature, we would like to manage templates for various resources in isolation and also potentially reuse them in different stacks, this feature will help a lot towards this end. 

A few notes, I originally considered doing this on yaml parse or dump to avoid a pass through the entire template dictionary, however since JSON is also supported this would have been clunky, since it would be difficult to reuse parse hooks across pyyaml and json. It was impossible to add this without bypassing the existing "resource" export pattern since these can appear anywhere in the document, adding a handler that matched some key seemed reasonable to me, and I don't think we have to worry about cyclic references for yaml_parse return val. Lastly my python is not very strong and I've never used this unit testing framework, open to any suggestions you have

If this PR is approved I will reject and close the other one

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -405,16 +405,16 @@ EXPORT_DICT = {
 }
 
 
-def includes_transform_export_handler(template_dict, s3_uploader):
+def include_transform_export_handler(template_dict, uploader):
     if template_dict.get("Name", None) != "AWS::Include":
         return template_dict
-    includes_location = template_dict.get("Parameters", {}).get("Location", {})
-    if (is_local_file(includes_location)):
-        template_dict["Parameters"]["Location"] = s3_uploader.upload_with_dedup(includes_location)
+    include_location = template_dict.get("Parameters", {}).get("Location", {})
+    if (is_local_file(include_location)):
+        template_dict["Parameters"]["Location"] = uploader.upload_with_dedup(include_location)
     return template_dict
 
 GLOBAL_EXPORT_DICT = {
-    "Fn::Transform": includes_transform_export_handler
+    "Fn::Transform": include_transform_export_handler
 }
 
 
@@ -454,9 +454,10 @@ class Template(object):
         """
         for key, val in template_dict.iteritems():
             if key in GLOBAL_EXPORT_DICT:
-                val = GLOBAL_EXPORT_DICT[key](val, self.uploader)
+                template_dict[key] = GLOBAL_EXPORT_DICT[key](val, self.uploader)
             elif type(val) is dict:
                 self.export_global_artifacts(val)
+        return self.template_dict
 
 
     def export(self):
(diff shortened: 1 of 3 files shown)


ekcrisp:
https://github.com/aws/aws-cli/pull/3440 will be closed without merging if this is approved

ekcrisp:
@stealthycoin @sanathkr @jamesls @joguSD would appreciate a review from someone on your end when you get a chance, I'm going on vacation soon and would like to make any updates before then :)

this implements the feature requested here
https://github.com/aws/aws-cli/issues/2513

ekcrisp:
@sanathkr Just updated, thank you for reviewing!

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -432,6 +445,21 @@ def __init__(self, template_path, parent_dir, uploader,
...
         self.uploader = uploader
 
+    def export_global_artifacts(self, template_dict):
+        """
+        Template params such as AWS::Include transforms are not specific to 
+        any resource type but contain artifacts that should be exported,
+        here we iterate through the template dict and export params with a 
+        handler defined in GLOBAL_EXPORT_DICT
+        """
+        for key, val in iter(template_dict.items()):
```
kyleknap: Just a python thing. We can simplify this to:
```suggestion
        for key, val in template_dict.items():
```

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -432,6 +445,21 @@ def __init__(self, template_path, parent_dir, uploader,
...
+            if key in GLOBAL_EXPORT_DICT:
+                template_dict[key] = GLOBAL_EXPORT_DICT[key](val, self.uploader)
+            elif type(val) is dict:
+                self.export_global_artifacts(val)
+        return self.template_dict
```
kyleknap: Is it intentional that we are returning `self.template_dict`? I would assume it would instead be `return template_dict`. We probably we do not even need to return anything here because Python dictionaries are mutable so any updates made to the dictionary will reflect in the original dictionary passed in.

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -432,6 +445,21 @@ def __init__(self, template_path, parent_dir, uploader,
...
+        """
+        Template params such as AWS::Include transforms are not specific to 
+        any resource type but contain artifacts that should be exported,
+        here we iterate through the template dict and export params with a 
+        handler defined in GLOBAL_EXPORT_DICT
+        """
+        for key, val in iter(template_dict.items()):
+            if key in GLOBAL_EXPORT_DICT:
+                template_dict[key] = GLOBAL_EXPORT_DICT[key](val, self.uploader)
+            elif type(val) is dict:
```
kyleknap: We should be able to simplify this to:
```suggestion
            elif isinstance(val, dict):
```

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -432,6 +445,21 @@ def __init__(self, template_path, parent_dir, uploader,
...
+                self.export_global_artifacts(val)
```
kyleknap: Is it possible to have `Fn::Transforms` in lists as well? We may want to traverse through lists as well if that is the case.

PR #3480, opened 2018-07-30
add a simple sms example

ravsau:
I think this is a cool and easy use case for SNS publish.

### DIFF
--- a/awscli/examples/sns/publish.rst
+++ b/awscli/examples/sns/publish.rst
@@ -8,3 +8,8 @@ The following command publishes a message to an SNS topic named ``my-topic``::
   Second Line
 
 Putting the message in a text file allows you to include line breaks.
+
+
+Use the following command to send a quick SMS/text message::
+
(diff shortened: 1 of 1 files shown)


awscli/examples/sns/publish.rst
```diff
@@ -7,4 +7,9 @@ The following command publishes a message to an SNS topic named ``my-topic``::
...
   Second Line
 
-Putting the message in a text file allows you to include line breaks.
\ No newline at end of file
+Putting the message in a text file allows you to include line breaks.
+
+
+Use the following command to send a quick SMS/text message::
+
+   aws sns publish --phone-number +12012022034 --message " This is bob. My phone's dead. I'll be home soon."
```
stealthycoin: There is an extra space at the beginning of the message.

justindho:
Hi @ravsau,

Thanks for the PR, and apologies for taking so long to get back to you. There was an example added in #4765 that adds an example very similar to yours. As a result, I will be closing this PR.

PR #3521, opened 2018-08-24
Update CLI for service id event aliasing

JordonPhillips:
This fixes issues that sprung up as a result botocore switching to
using service id in events.

Related to boto/botocore#1524 , the tests won't pass without it.

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -245,7 +245,7 @@ def _is_complex_shape(model):
 
 class ParamShorthand(object):
 
-    def _uses_old_list_case(self, service_name, operation_name, argument_name):
+    def _uses_old_list_case(self, service_id, operation_name, argument_name):
         """
         Determines whether a given operation for a service needs to use the
         deprecated shorthand parsing case for lists of structures that only have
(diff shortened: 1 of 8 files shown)


awscli/argprocess.py
```diff
@@ -245,7 +245,7 @@ def _is_complex_shape(model):
 
 class ParamShorthand(object):
 
-    def _uses_old_list_case(self, service_name, operation_name, argument_name):
+    def _uses_old_list_case(self, service_id, operation_name, argument_name):
```
kyleknap: Could you elaborate on why we need to change shorthand parsing logic to use serviceId's? Looking at the code, `ParamShorthand` is only registered to the [process-cli-arg event](https://github.com/aws/aws-cli/blob/ee71e8cca0838a1505992ca9cd29add33386dfe6/awscli/handlers.py#L90) and that event only ever emits a [service_name](https://github.com/aws/aws-cli/blob/develop/awscli/arguments.py#L461). So I do not think we need to update this.

JordonPhillips: Because we alias both emit and register, so the events are still getting aliased. These in particular have to be changed because they're inspecting the event name. I could also refactor it to be explicit by having the event actually send the service id and operation name as kwargs.

PR #3537, opened 2018-09-04
Group IDs do not start with AKIA, they use AGPA

codeforsmiles:
This will help clear up confusion and inconsistencies in our docs.

AKIA are prefixes for Access Key IDs.

Definition list: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-prefixes


codeforsmiles:
Committing into develop instead of master

PR #3538, opened 2018-09-04
IAM Principal IDs do not start with AKIA, they use AIDA

codeforsmiles:
This will help clear up confusion and inconsistencies in our docs.

AKIA is a prefix for Access Key IDs.

Definition list: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-prefixes


codeforsmiles:
Committing into develop instead of master

PR #3549, opened 2018-09-07
s3 commands that list buckets will use ListObjectsV2

drocamor:
This PR makes the AWS CLI use [ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html) when doing operations in the s3 subcommand like ls, cp, and sync. The S3 API docs recommend using ListObjectsV2 instead of V1. 

I updated the code and the tests to use V2. It seems like there are some errors around the tests for requester pays features that I may have introduced, but I don't understand where they are coming from. If you have suggestions on how to fix those I can do that. Thanks!

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1081,7 +1081,8 @@ class CommandArchitecture(object):
     def _get_file_generator_request_parameters_skeleton(self):
         return {
             'HeadObject': {},
-            'ListObjects': {}
+            'ListObjects': {},
+            'ListObjectsV2': {}
         }
 
     def _map_request_payer_params(self, request_parameters):
@@ -1092,6 +1093,11 @@ class CommandArchitecture(object):
         )
         RequestParamsMapper.map_list_objects_params(
             request_parameters['ListObjects'], {
+                'request_payer': self.parameters.get('request_payer')
+            }
+        )
+        RequestParamsMapper.map_list_objects_params(
+            request_parameters['ListObjectsV2'], {
                 'request_payer': self.parameters.get('request_payer')
             }
         )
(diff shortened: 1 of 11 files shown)


jamesls:
The errors come from the fact that botocore has a customization to automatically urlencode/decode params.  These customizations are configured per operation (in botocore), so we'd need something similar for ListObjectsV2 now that we're not using ListObjects.

https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L881

We'd also need to do a thorough round of testing just to confirm there are no unintended negative side effects.

Let me know if you have any other questions.

drocamor:
Ah, ok. I will make a PR for botocore too and comment here when the tests are passing on my branches.

Agreed on the thorough testing.

Thanks!

drocamor:
OK, I had to make some Requester Pays related changes here and I made the encoding changes here: https://github.com/boto/botocore/pull/1552

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -103,9 +103,9 @@ def test_ls_command_for_bucket(self):
             .return_value.paginate
         # We should make no operation calls.
         self.assertEqual(call.call_count, 0)
```
kyleknap: The line that creates the `call` object should probably be changed to:
```
call = self.session.create_client.return_value.list_objects_v2
```
instead of:
```
call = self.session.create_client.return_value.list_objects
```

awscli/customizations/s3/subcommands.py
```diff
@@ -1095,6 +1096,11 @@ def _map_request_payer_params(self, request_parameters):
                 'request_payer': self.parameters.get('request_payer')
             }
         )
+        RequestParamsMapper.map_list_objects_params(
```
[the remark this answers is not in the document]
drocamor: 👍

tests/unit/customizations/s3/test_subcommands.py
drocamor: 👍

gaul:
Is there some specific reason to prefer ListObjects V2 over V1?  Some S3 implementations like S3Proxy do not implement the former.

Issue #3394, opened 2018-06-20
GraphQLSchema defines both 'Definition' and 'DefinitionS3Location'


fermayo:
This bug affects `aws-cli 1.15.23+`.

Given the following cloudformation template:
```
AWSTemplateFormatVersion: '2010-09-09'

Resources:
  GraphQLApi:
    Type: "AWS::AppSync::GraphQLApi"
    Properties:
      Name: Test
      AuthenticationType: "API_KEY"

  GraphQLSchema:
    Type: "AWS::AppSync::GraphQLSchema"
    Properties:
      ApiId: !GetAtt GraphQLApi.ApiId
      Definition: schema {}
```

Running `aws cloudformation package --template-file test.yaml --s3-bucket bucket` results in the following:
```
Uploading to 05e9ac726edc2118dbfb0b34faffc602  14043648 / 14043648.0  (100.00%)
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  GraphQLApi:
    Properties:
      AuthenticationType: API_KEY
      Name: Test
    Type: AWS::AppSync::GraphQLApi
  GraphQLSchema:
    Properties:
      ApiId:
        Fn::GetAtt:
        - GraphQLApi
        - ApiId
      Definition: schema {}
      DefinitionS3Location: s3://bucket/05e9ac726edc2118dbfb0b34faffc602
    Type: AWS::AppSync::GraphQLSchema
```

Note that both `Definition` and `DefinitionS3Location` are defined, when only `Definition` should be defined. Also note that nothing should be uploaded to S3 in this case either. This causes `cloudformation deploy` to fail with `Cannot set both the Definition and the DefinitionS3Location.`.

I believe this was introduced in https://github.com/aws/aws-cli/pull/3297

jamesls:
Thanks for reporting, cc @sanathkr, thoughts?

mkrn:
+1 affected by it. 

Also support for DefinitionS3Location  is not documented.

Downgraded to aws-cli/1.11.157 and using SAM CLI, version 0.4.0  solved it.  FYI

adrai:
With `aws-cli 1.15.22` it works.

I believe this regression was introduced in aws/aws-cli@c7a22bc.

danishbacker:
+1 affected by it. Downgrading AWS CLI as I was using CodePipeline/CodeBuild to deploy template.
But the making schema separate and using DefinitionS3Location solved the problem for me.

`Schema:
[...]

PR #3558, opened 2018-09-11
Don't upload and set GraphQL Schema Definition

txase:
If Definition is provided in-line, don't upload it and modify the
template. Previously, it would be uploaded, DefinitionS3Location would
be set, but Definition would also remain, leaving conflicting
properties.


txase:
Yes, it fixes that exact issue. I came across the bug on my own, so I didn’t know about #3394.

PR #3566, opened 2018-09-14
Add example for the backtrack-db-cluster command

randyurbano:
I want to add an example to the RDS CLI Reference for the backtrack-db-cluster command.


randyurbano:
If backtrack-db-cluster.rst is the correct name, then yes, please change it. Thanks.

randyurbano:
I changed the file name.

randyurbano:
Can you tell me why this pull request hasn't been merged? Is something incorrect?

PR #3592, opened 2018-09-19
[CLIv2] Change version to 2.0.0dev0

jamesls:
That way it's easier to identify if someone is trying out
the dev version of 2.0.

There's currently no plan to publish any of these pre-releases to pypi.


JordonPhillips:
There's also versions in `docs/source/conf.py`

PR #3610, opened 2018-09-27
[CLIv2] Change default timestamp format to iso8601

JordonPhillips:


awscli/customizations/scalarparse.py
```diff
@@ -46,19 +46,22 @@ def iso_format(value):
...
         # This function can be called as the first handler for
         # the session-initialized event, which happens before a
         # profile can be created, even if the command would have
         # successfully created a profile. Instead of crashing here
-        # on a ProfileNotFound the CLI should just use 'none'.
-        timestamp_format = 'none'
+        # on a ProfileNotFound the CLI should just use the default.
+        timestamp_format = default_format
+
     if timestamp_format == 'none':
```
jamesls: @stealthycoin has mentioned that we should consider changing this to `wire` to make this more obvious what it means.  Thoughts?

JordonPhillips: seems reasonable to me

stealthycoin: Wire is a pretty common term in the outside world as well.

kyleknap: I like `wire` as well. This behavior would be a more advanced feature in general. I think for people wanting to enable this would understand the meaning of `wire`, and I think `service-format` is a little ambiguous to me because I would read it as a service having a single specific timestamp format when that is not necessarily true as some services return different timestamp formats based on the operation.

JordonPhillips:
Updated

PR #3621, opened 2018-10-02
[CLIv2] Rebase against develop

JordonPhillips:
This is to pick up some bug fixes from the main branch


joguSD:
It *seems* like this rebase has commits that are already in the `v2` branch and they're going to show up twice now. I don't think it's common to rebase and then merge. You either rebase and force push or merge `develop` into `v2`, right?

PR #3640, opened 2018-10-08
Remove backwards compatibility aliases

stealthycoin:


stealthycoin:
Id rather leave them in case they are needed again.

joguSD:
@stealthycoin I think I'm with Kyle on this one. I'd rather remove the code if it's no longer used. We can always get it from a previous commit if we would like to use it again.

kyleknap:
Talked offline about this about, but I still think we should remove it. Given it's for v2, we should be striving to cleanup the code base as well which entails removing dead code. If we ever need to add this back, it should not be hard to track down this PR and add it back.

PR #3653, opened 2018-10-15
[CLIv2] Add indexing for model based completer

jamesls:
This is the start of the rewrite of the CLI autocompleter.  This first PR adds
the data model for handling auto completion based on the service model
definitions.  The big change from the current implementation is that rather
than do this dynmically, we instead generate an index separately, and then we
use the index to drive the auto completion process.

The original approach of traversing the `command_table` during the auto
completion process worked "ok" when the number of services was relatively small
5 years ago, but that approach is too slow given the number of services we now
have.  Also, importing the clidriver imports the majority of the AWS CLI as
well as botocore.  The import times for both of these packages has grown over
the years as we've added more functionality.  This is something we should look
at addressing separately, but regardless, we should try to avoid unnecessary
imports during the auto completion process.  The end goal is that we only
need to import sqlite3 in order to drive our auto completer.

This initial PR isn't enough to replace the existing auto completer.  There's
several other features needed that will be part of subsequence PRs once this
one gets merged into the `v2-autocomplete` branch.  We need an updated CLI
parser that's based on this generated index.  We also need to figure how/when
to plumb in this index generation.  I'm thinking we have an explicit command to
auto generate the index, and we also have some sort of smart detection where we
generate the index generation when we notice there's no index available (the
details still need to be worked out on that).  Another option is to check in
a pre-generated index before we cut CLI releases (in addition to the
auto generation heuristics).

I imagine the schema will need to change as we start adding more functionality
to the auto completer.  I'm looking for initial feedback on the overall
approach.

### DIFF
--- a/awscli/autocomplete/generator.py
+++ b/awscli/autocomplete/generator.py
@@ -1,0 +1,37 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Generates auto completion index."""
+from awscli.autocomplete import indexer
+
+
+def create_model_indexer(filename):
+    index = indexer.ModelIndexer(
+        indexer.DatabaseConnection(filename)
+    )
+    return index
+
+
+class IndexGenerator(object):
+    """Generates auto completion index.
+
+    This will generate an auto completion index for all the low level
+    indexer used by the CLI.  This object primarily delegates to other
+    objects that do the actual heavy lifting of generating auto completion
+    indices.
+
+    """
+    def __init__(self, model_indexer):
+        self._model_indexer = model_indexer
+
+    def generate_index(self, clidriver):
+        self._model_indexer.generate_index(clidriver)
--- a/awscli/autocomplete/indexer.py
+++ b/awscli/autocomplete/indexer.py
@@ -1,0 +1,118 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import logging
+import sqlite3
+
+
+LOG = logging.getLogger(__name__)
+
+
+class ModelIndexer(object):
+    def __init__(self, db_connection):
+        self._db_connection = db_connection
+
+    def generate_index(self, clidriver):
+        parent = 'aws'
+        self._db_connection.execute(
+            'INSERT OR REPLACE INTO command_table (command)'
+            'VALUES (:command)',
+            command=parent
+        )
+        command_table = clidriver.subcommand_table
+        self._generate_arg_index(command=parent, parent='',
+                                 arg_table=clidriver.arg_table)
+        self._generate_command_index(command_table, parent=parent)
+
+    def _generate_arg_index(self, command, parent, arg_table):
+        for name, value in arg_table.items():
+            self._db_connection.execute(
+                'INSERT INTO param_table '
+                '(argname, type_name, command, parent, nargs)'
+                ' VALUES (:argname, :type_name, :command, :parent, :nargs)',
+                argname=name, type_name=value.cli_type_name,
+                command=command, parent=parent,
+                nargs=value.nargs,
+            )
+
+    def _generate_command_index(self, command_table, parent):
+        for name, command in command_table.items():
+            self._db_connection.execute(
+                'INSERT INTO command_table (command, parent) '
+                'VALUES (:command, :parent)',
+                command=name, parent=parent,
+            )
+            self._generate_arg_index(command=name, parent=parent,
+                                     arg_table=command.arg_table)
+            self._generate_command_index(command.subcommand_table,
+                                         parent='%s.%s' % (parent, name))
+
+
+# This is similar to DBConnection in awscli.customization.history.
+# I'd like to reuse code, but we also have the contraint that we don't
+# want to import anything outside of awscli.autocomplete to ensure
+# our startup time is as minimal as possible.
+class DatabaseConnection(object):
+    _CREATE_CMD_TABLE = """\
+        CREATE TABLE IF NOT EXISTS command_table (
+          command TEXT,
+          parent TEXT REFERENCES command_table,
+          PRIMARY KEY (command, parent)
+        );
+    """
+    _CREATE_PARAM_TABLE = """\
+        CREATE TABLE IF NOT EXISTS param_table (
+          argname TEXT,
+          type_name TEXT,
+          command TEXT,
+          parent TEXT,
+          nargs TEXT,
+          FOREIGN KEY (command, parent) REFERENCES
+            command_table(command, parent)
+        );
+    """
+    _ENABLE_WAL = 'PRAGMA journal_mode=WAL'
+
+    def __init__(self, db_filename):
+        self._db_conn = None
+        self._db_filename = db_filename
+
+    @property
+    def _connection(self):
+        if self._db_conn is None:
+            self._db_conn = sqlite3.connect(
+                self._db_filename, check_same_thread=False,
+                isolation_level=None)
+            self._ensure_database_setup()
+        return self._db_conn
+
+    def close(self):
+        self._connection.close()
+
+    def execute(self, query, **kwargs):
+        return self._connection.execute(query, kwargs)
+
+    def _ensure_database_setup(self):
+        self._create_table()
+        self._try_to_enable_wal()
+
+    def _create_table(self):
+        self.execute(self._CREATE_CMD_TABLE)
+        self.execute(self._CREATE_PARAM_TABLE)
+
+    def _try_to_enable_wal(self):
+        try:
+            self.execute(self._ENABLE_WAL)
+        except sqlite3.Error:
+            # This is just a performance enhancement so it is optional. Not all
+            # systems will have a sqlite compiled with the WAL enabled.
+            LOG.debug('Failed to enable sqlite WAL.')
--- a/awscli/autocomplete/model.py
+++ b/awscli/autocomplete/model.py
@@ -1,0 +1,107 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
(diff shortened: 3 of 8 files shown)


tests/unit/autocomplete/test_indexer.py
```diff
@@ -0,0 +1,180 @@
...
+                        'list-objects': DummyCommand('list-objects'),
+                        'put-object': DummyCommand('put-object'),
+                    }
+                )
+            }
+        )
+        # Ideally we just use ':memory:', but that's for
+        # a specific sqlite3 connection so we'd have to
+        # update our interfaces to return/accept a db connection.
+        # I'd prefer not to have db connections as part of the
```
joguSD: How about defaulting to an in memory connection if the filename passed is `None`?

awscli/autocomplete/indexer.py
```diff
@@ -0,0 +1,118 @@
...
+          PRIMARY KEY (command, parent)
+        );
+    """
+    _CREATE_PARAM_TABLE = """\
+        CREATE TABLE IF NOT EXISTS param_table (
+          argname TEXT,
+          type_name TEXT,
+          command TEXT,
+          parent TEXT,
+          nargs TEXT,
```
kyleknap: This will probably be addressed in future PR's but what will be the upgrading plan when we need to add new argument data to the table (i.e. say we wanted to add `required`, `choices`, or `group_name`)? In the history db, we used a generalized `payload` field so that if new history fields were added it would not force a complete regeneration of the history db. However, I do not think that is applicable to the completer db as we will probably want the entire db to be regenerated.

tests/unit/autocomplete/test_indexer.py
```diff
@@ -0,0 +1,180 @@
...
+        self.indexer.generate_index(self.aws_command)
+        self.assertEqual(
+            set(self.query.arg_names(lineage=[], command_name='aws')),
+            set(['region', 'endpoint-url']),
+        )
+
+    def test_can_retrieve_service_params(self):
+        self.indexer.generate_index(self.aws_command)
+        self.assertEqual(
+            set(self.query.arg_names(lineage=['aws.ec2'],
```
kyleknap: Is this a valid way to use `lineage`? It will work because we are joining on `.` under the hood, but if we change it, there will be problems and I'm not sure we want the tests to reflect this.

awscli/autocomplete/generator.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
jamesls: Yes along with whatever auto mechanism triggers the "kick off autocompletion" process (assuming we add one)

awscli/autocomplete/model.py
```diff
@@ -0,0 +1,107 @@
...
+        return self._db_connection
+
+    def command_names(self, lineage):
+        """Return command names given a lineage.
+
+        This ``lineage`` is the same concept used in the
+        AWS CLI command classes, except that it explicitly
+        uses ``aws`` as the first parent.  This is the list of
+        parent commands for a given CLI command.  For example,
+        ``aws ec2 <here>`` has a lineage of ``['ec2']``.  The
```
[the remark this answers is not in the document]
jamesls: Yeah I'll update the doc example to be `['aws', 'ec2']` instead of `['ec2']`.

I think it's just a cleaner design that doesn't assume the "aws" prefix as parts of the value for lineage.  It makes the logic more generalized.  That way if you want to refer to the top level command you can use `lineage=[], command_name='aws'`.

awscli/autocomplete/indexer.py
```diff
@@ -0,0 +1,118 @@
...
+                'VALUES (:command, :parent)',
+                command=name, parent=parent,
+            )
+            self._generate_arg_index(command=name, parent=parent,
+                                     arg_table=command.arg_table)
+            self._generate_command_index(command.subcommand_table,
+                                         parent='%s.%s' % (parent, name))
+
+
+# This is similar to DBConnection in awscli.customization.history.
```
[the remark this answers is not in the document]
jamesls: It's also going to be challenging to remember to enforce that you can't import anything "heavyweight" in a `dbutils` module.  It's much easier to keep track of the fact that anything in `awscli/autocomplete/` should not import anything like `clidriver/botocore`.

It will also depend on how much logic is similar.  They've already diverged a bit and may likely diverge further.  Right now we'd only save about 10-20 lines of code.

awscli/autocomplete/indexer.py
jamesls: Yeah it doesn't matter for auto completion.  The entire db is regenerated per release (the service models are going to be different).  This really only affects us (CLI devs) if we want to make changes to the schema while testing against a specific version of the CLI.

tests/unit/autocomplete/test_indexer.py
jamesls: No it's not, good catch.  I'll update this.

jamesls:
@joguSD 
> Do we need to worry about some kind of migration process for when the schema changes?

No, separate index per version, similar to what the aws-shell does.  Kyle had the same question (https://github.com/aws/aws-cli/pull/3653#discussion_r225702609)

>If we do ship a command that generates the index 

To be clear, my suggestions weren't meant to be mutually exclusive.  We'll ship a command to generate the index, but I also want to investigate ways for customers to not have to run the command.  It's just meant as a way to either troubleshoot or regen the index for some reason.

> I think we saw a lot of pain/annoyance for the 'background' generation of the auto completion in the shell though, 

That doesn't ring a bell, can you remind me again?  I searched through the issues on the aws-shell and didn't find anything.  I can maybe see that being the case for the doc-based indexing but the model auto-completion is orders of magnitude quicker.

awscli/autocomplete/generator.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Generates auto completion index."""
+from awscli.autocomplete import indexer
+
+
+def create_model_indexer(filename):
```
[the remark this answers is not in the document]
jamesls: Moved.

awscli/autocomplete/indexer.py
```diff
@@ -0,0 +1,118 @@
...
+# I'd like to reuse code, but we also have the contraint that we don't
+# want to import anything outside of awscli.autocomplete to ensure
+# our startup time is as minimal as possible.
+class DatabaseConnection(object):
+    _CREATE_CMD_TABLE = """\
+        CREATE TABLE IF NOT EXISTS command_table (
+          command TEXT,
+          parent TEXT REFERENCES command_table,
+          PRIMARY KEY (command, parent)
```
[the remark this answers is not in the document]
jamesls: Not really sure what you mean by "pseudo composite parent".  We can't model the chain of parents as a composite key, it's arbitrary depth (or at least I don't know how you do it if it's possible).  The parent is a dotted reference of the command lineage up to the current command, similar to what's used in the command objects in the CLI today.

The existing composite primary key in the schema wasn't chosen for performance, it was chosen so the queries are simple (`select command from command_table where parent=:parent`).  With what you've proposed I think you'd have to use something like `where lineage regexp 'foo.bar.[^\.*]'` (I don't think that regex is actually correct but basically `foo.bar.*` but don't match `foo.bar.*.baz`).

tests/unit/autocomplete/test_indexer.py
jamesls: How does that help though?  That only works if we accept a db connection as an input param which now couples our indexer to a db connection.

awscli/autocomplete/indexer.py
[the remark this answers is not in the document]
joguSD: @jamesls Ah, that makes sense. By "pseudo composite" I meant the joint string separated by `.`, like we're doing for the parent right now, and not a real composite key in the SQL regard.

awscli/autocomplete/indexer.py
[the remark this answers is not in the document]
kyleknap: >> It's also going to be challenging to remember to enforce that you can't import anything "heavyweight" in a dbutils module... Right now we'd only save about 10-20 lines of code

That's fair enough. I wonder if there are other ways we can ensure that we do not import anything heavy weight into this module. An easy win would be put a comment at the top of the file that says to minimize imports (right now you have to scroll midway through the file to realize it). But I do wonder if there is any tests/checks (I would need to brainstorm examples) we can put in place for additional ensurance our imports are lightweight?

tests/unit/autocomplete/test_indexer.py
joguSD: The `ModelIndexer` is already coupled to `DatabaseConnection` though, so why does that matter? 

Do you mean the `ModelIndex` class taking a filename vs database connection? (Maybe the naming on these classes is a little confusing). Might be worth renaming `ModelIndexer` and `DatabaseConnection` to `SQLiteModelIndexer` and `SQLiteDatabaseConnection`. `ModelIndex` should be defined by the abstract interface:
```python
class ModelIndex(object):
    def command_names(self, lineage):
        pass

    def arg_names(self, lineage, command_name):
        pass

    def get_argument_data(self, lineage, command_name, arg_name):
        pass
```

And we have one concrete implementation of this: `SQLiteModelIndex`. Realistically, the current `ModelIndex` is already tied to a database connection implicitly (which is fine, but maybe the name should reflect that). If we're going to decouple `ModelIndex` from a database connection let's decouple it from IO entirely and not expect a filename either by having the constructor omitted from the interface.

tests/unit/autocomplete/test_indexer.py
```diff
@@ -0,0 +1,188 @@
...
+        # should be able to differentiate between this command and the
+        # one in ec2 via the ``parent`` arg.
+        s3_commands = self.aws_command.subcommand_table['s3'].subcommand_table
+        s3_commands['describe-instances'] = DummyCommand('describe-instances')
+        self.indexer.generate_index(self.aws_command)
+        # The 'ec2' version has several params in its arg table, but this
+        # 's3' version should have no params because we didn't add an arg
+        # table.
+        self.assertEqual(
+            set(self.query.arg_names(lineage=['aws.s3'],
```
kyleknap: I think we need to update this test case as well:
```suggestion
            set(self.query.arg_names(lineage=['aws', 's3'],
```

joguSD:
@jamesls I suppose I was conflating the documentation generation, looking at the issues I think that's actually what I was recalling (doc issues failing causing the shell to not launch). The only other issue I found was [this](https://github.com/awslabs/aws-shell/issues/162), which was more of a feature request to generate the indices without running the shell. 

At least personally, I was frustrated by having to regenerate the index every time the CLI version changed, as I hopped around CLI versions so much. Though, that was largely due to working on the CLI more than anything. I guess my point is: customers that update everyday shouldn't need to remember to generate the index everyday.

tests/unit/autocomplete/test_indexer.py
jamesls: That doesn't really answer my question though.  It sounds like you're saying "they're already somewhat coupled so why not couple them even more?".

Unless there's a concrete benefit as to why it make sense to do this, then this seems like unnecessary design at this point.

awscli/autocomplete/indexer.py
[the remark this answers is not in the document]
jamesls: Yeah I definitely have a few ideas, I just think they're much lower priority than getting everything implemented first.

tests/unit/autocomplete/test_indexer.py
joguSD: There's two benefits:
1) The interface for a `ModelIndex` is properly defined and isn't mixed with an implementation that's based on a database connection.
2) Allowing an object that represents the database connection to be passed instead decouples your implementation from any specific IO model (file vs memory), but does couple you more to the fact that a database connection is being used. I think this is fine as literally *every* method in the current implementation of the `ModelIndex` object is assuming you're dealing with a database connection. I don't see why you would want to decouple the `SQLiteModelIndex` object from a database connection. If you want to do something radically different create a different object that meets the already defined `ModelIndex` interface.

PR #3680, opened 2018-10-25
Add IMDS region loader to the session config chain

stealthycoin:
This commit adds a class that can be used to load the default region
which on an ec2 instance will be the region provided by the metadata
service. The CLI injects this loader into the region config loading
chain at the end to be used instead of the default None value used by
botocore's session.

This is a draft depending on the interface and approval of https://github.com/boto/botocore/pull/1591

cc @jamesls @kyleknap @JordonPhillips @dstufft @joguSD

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -10,25 +10,29 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import os
 import sys
 import signal
 import logging
 
 import botocore.session
 from botocore import __version__ as botocore_version
-from botocore.hooks import HierarchicalEmitter
 from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
 from botocore.history import get_global_history_recorder
+from botocore.configprovider import InstanceVarProvider
+from botocore.configprovider import EnvironmentProvider
+from botocore.configprovider import ScopedConfigProvider
+from botocore.configprovider import ConstantProvider
+from botocore.configprovider import ChainProvider
 
-from awscli import EnvironmentVariables, __version__
+from awscli import __version__
 from awscli.compat import get_stderr_text_writer
 from awscli.formatter import get_formatter
 from awscli.plugin import load_plugins
 from awscli.commands import CLICommand
-from awscli.compat import six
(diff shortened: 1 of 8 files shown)


awscli/clidriver.py
```diff
@@ -76,12 +77,25 @@ def _set_user_agent_for_session(session):
     session.user_agent_extra = 'botocore/%s' % botocore_version
 
 
+def _update_config_chain(session):
```
kyleknap: Talked offline about this but let's just make this an internal method and pull in the logic for configuration into this method instead of having the `awscli.configuration` module.

awscli/utils.py
```diff
@@ -17,10 +17,138 @@
...
+        should_use_instance_region = \
+            self._construct_use_instance_region_chain().provide()
+        if not should_use_instance_region.lower() == 'true':
+            return None
+
+        fetcher = self._get_fetcher()
+        region = fetcher.retrieve_region()
+        return region
+
+    def _construct_use_instance_region_chain(self):
```
kyleknap: Talked offline about this, but we can hold off on this for now. There is not really too much benefit in specifically opting out of this because you can opt-out by setting a region since it is last in the chain and if you really did not want to hit IMDS you would just set the environment variable to turn it off completely.

tests/unit/test_utils.py
```diff
@@ -203,3 +210,213 @@ def test_can_silence_io_error_from_pager(self):
                     pass
         except IOError:
             self.fail('Should not raise IOError')
+
+
+class TestInstanceMetadataRegionFetcher(unittest.TestCase):
```
kyleknap: Talked offline about this as well, but it would be great to get a functional test in for this.

tests/unit/test_utils.py
```diff
@@ -203,3 +210,155 @@ def test_can_silence_io_error_from_pager(self):
...
+        user_agent = 'my-user-agent'
+        self.add_get_region_imds_response()
+
+        InstanceMetadataRegionFetcher(
+            user_agent=user_agent).retrieve_region()
+
+        headers = self._send.call_args[0][0].headers
+        self.assertEqual(headers['User-Agent'], user_agent)
+
+    def test_non_200_response_for_region_is_retried(self):
```
kyleknap: Do we have a test for the happy 200 response case?

stealthycoin: Yea there is a test for when imds is not disabled.

https://github.com/aws/aws-cli/pull/3680/files/e7d666c14c84b774c8f8fa6bfe5b25118f05c962#diff-4b935d93f7e4f4a47577b93a23022fe3R264

PR #3681, opened 2018-10-25
Set region on session if region arg is provided

jamesls:
This fixes a bug where code that's calling session.create_client()
and not providing an explicit region value.  In this case we fall
back to pulling the region from the session config vars, which will
check env vars, config files, and session instance vars.  By setting
the region as a session instance var, this will ensure that clients
use the same region as specified from the `--region` arg.


jamesls:
@JordonPhillips Added a changelog entry.

PR #3685, opened 2018-10-25
Add filtering to history show command

stealthycoin:
Filters out the hash after Signature=<hash here>. Introduces a filter
abstraction that can be added to a particular Formatter section to
filter any content emitted by it.

### DIFF
--- a/awscli/customizations/history/filters.py
+++ b/awscli/customizations/history/filters.py
@@ -1,0 +1,18 @@
+import re
+
+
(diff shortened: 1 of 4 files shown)


stealthycoin:
I'm not too worried about it being stored long term since its not useful after a small time period. I'm mainly interested in fixing the case where someone gets a `history show` output and posts it right away without sanitizing that part.

PR #3695, opened 2018-11-01
[CLIv2] Add model based completer

jamesls:
This integrates the model index and the cli parser to create an auto-completer.  This includes:

* A model-based completer that uses the index to auto complete commands and parameters
* A class that's used to delegate to concrete completers so we can support for more types of auto-completers.
* A bug fix for the parser found when testing this model completer.  It was raising an `IndexError` in cases where it tried to consume a value for an argument, and there were cases where `last_fragment` was suppose to be an empty string but was `None` instead.

This should give us feature parity with the existing completer in the CLI.

### DIFF
--- a/awscli/autocomplete/completer.py
+++ b/awscli/autocomplete/completer.py
@@ -1,0 +1,102 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+class AutoCompleter(object):
+    """Main auto-completer object for the AWS CLI."""
+    def __init__(self, parser, completers):
+        self._parser = parser
+        self._completers = completers
+
+    def autocomplete(self, command_line, index=None):
+        parsed = self._parser.parse(command_line, index)
+        for completer in self._completers:
+            result = completer.complete(parsed)
+            if result is not None:
+                return result
+
+
+class BaseCompleter(object):
+    def complete(self, parsed):
+        """Attempt to autocomplete parsed on parsed result.
+
+        Subclasses should implement this method.
+
+        :param parsed: A ParsedResult from the auto complete CLI parser.
+        :return: This method should return one of two values:
+
+            * ``None`` - If the completer doesn't understand how to
+              complete the command, it should return ``None``.  This
+              signals to the ``AutoCompleter`` that it should move on
+              to the next completer.
+            * ``List[str]`` - If the completer is able to offer
+              auto-completions it should return a list of strings that
+              are valid suggestions for completing the command.  This
+              indicates to the ``AutoCompleter`` to immediately return
+              and to stop consulting other completers for results.
+        """
+        raise NotImplementedError("complete")
+
+
+class ModelIndexCompleter(BaseCompleter):
+
+    def __init__(self, index):
+        self._index = index
+
+    def complete(self, parsed):
+        if parsed.unparsed_items or parsed.last_fragment is None:
+            # If there's ever any unparsed items, then the parser
+            # encountered something it didn't understand.  We won't
+            # attempt to auto-complete anything here.
+            return
+        last_fragment = parsed.last_fragment
+        if last_fragment.startswith('--'):
+            # We could technically offer completion of options
+            # if the last fragment is an empty string, but to avoid
+            # dumping too much information back to the user, we only
+            # offer completions for options if the value starts with
+            # '--'.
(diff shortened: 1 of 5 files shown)


awscli/autocomplete/parser.py
```diff
@@ -150,6 +150,15 @@ def parse(self, command_line, location=None):
 
     def _consume_value(self, remaining_parts, option_name,
                        lineage, current_command):
+        # We have a special case where a user is trying to complete
+        # a value for an option, which is the last fragment of the command,
+        # e.g. 'aws ec2 describe-instances --instance-ids '
+        # Note the space at the end.  In this case we don't have a value
+        # to consume so we special case this and short circuit.
+        if remaining_parts == ['']:
```
kyleknap: It may be worth moving the empty string into a constant to help with readability and to help indicate its purpose in the parsing logic. Not sure about what name to give it though. Looking over the tests it would especially help in reading the difference between empty string and `None` for the `last_fragment`.

tests/unit/autocomplete/test_completer.py
```diff
@@ -0,0 +1,141 @@
...
+        mock_complete.complete.return_value = ['ec2', 'ecs']
+        auto_complete = completer.AutoCompleter(
+            self.parser, completers=[mock_complete])
+
+        results = auto_complete.autocomplete('aws e')
+        self.assertEqual(results, ['ec2', 'ecs'])
+        self.parser.parse.assert_called_with('aws e', None)
+        mock_complete.complete.assert_called_with(self.parsed_result)
+
+    def test_stops_processing_when_list_returned(self):
```
kyleknap: Do we want a test when both the first and second completers are set to return a value, but the first is the only one called?

awscli/autocomplete/completer.py
```diff
@@ -0,0 +1,120 @@
...
+            '--%s' % arg_name for arg_name in arg_names
+            if arg_name.startswith(fragment)]
+        # Global params apply to any scope, so if we're not
+        # in the global scope, we need to add completions for
+        # global params
+        self._inject_global_params_if_needed(parsed, results, fragment)
+        return results
+
+    def _inject_global_params_if_needed(self, parsed, results, fragment):
+        is_in_global_scope = (
```
kyleknap: I'm still not sure if I am following this. We only want to suggest global arguments if the current command is `aws` and there is no subcommands? So something like `aws --endpoint` would not trigger this if statement? Or is this because it is handled at the beginning of `_complete_options()`? It looks like it does get handled based on the tests.

tests/unit/autocomplete/test_completer.py
```diff
@@ -0,0 +1,141 @@
...
+        self.index.arg_names.return_value = ['region', 'endpoint-url']
+        self.assertEqual(self.completer.complete(parsed), ['--region'])
+
+    def test_can_combine_global_and_command_params(self):
+        parsed = parser.ParsedResult(
+            lineage=['aws', 'ec2'],
+            unparsed_items=[],
+            current_params={},
+            current_command='describe-instances',
+            last_fragment='--r',
```
kyleknap: May be a bit off-topic, but do we want to consider autocompleting for options when:

1) The last fragment is an empty string (i.e. not starting with `--`)
2) There are no more commands from the index
3) There are parameters available from the index.

So this would be:
```
aws ec2 describe-instances <tab>
```
and hopefully we would get all of the parameter options out of it?

awscli/autocomplete/completer.py
jamesls: What's happening is that we offer CLI options to auto complete based on your scope, e.g `aws ec2 describe-instances --` will autocomplete based on the params available for the `describe-instances` operation.  When you're in the global scope (`aws --`), the "current scope" *is* the global scope, so the only valid completions in that example would be global params.  However, when you're in the scope of a single operation, in addition to the params exposed by the operation we also need to offer global params because you can apply them whenever.

If we're already in the global scope we can skip this injection step because current scope == global scope.  If we didn't have this logic you'd be offered the same completions twice.

tests/unit/autocomplete/test_completer.py
jamesls: This was an explicit choice to avoid overwhelming the user with too many options.  I think it's pretty intuitive if you want CLI options, you can just type `--`.  There's  a comment that mentions this:

```
            # We could technically offer completion of options
            # if the last fragment is an empty string, but to avoid
            # dumping too much information back to the user, we only
            # offer completions for options if the value starts with
            # '--'.
```

jamesls:
@kyleknap Feedback incorporated.  I also added two changes:

* Move `InMemoryIndex` to `__init__.py` so the parser and the completer tests can reuse that abstraction.
* Change the interface to return a list of `CompletionResult` objects.  This allows us to include additional metadata about the completion as well as the starting index.

awscli/autocomplete/completer.py
kyleknap: That makes sense

tests/unit/autocomplete/test_completer.py
kyleknap: I can see that making sense. I would need to actually try a version of the integrated completer to know for sure. I can see it making sense though to do it for required parameters or having a workflow for figuring out the required parameters without having to access the help documentation or just running the command and seeing what validation errors you get. We can explore this more once everything is integrated and we can test it out more.

PR #3697, opened 2018-11-02
Retry smoke tests

JordonPhillips:
The smoke tests sometimes hit throttling hard enough that they fail,
particualrly when we run them in succession. This adds some retry with
delays on top.

### DIFF
--- a/tests/integration/test_smoke.py
+++ b/tests/integration/test_smoke.py
@@ -13,6 +13,7 @@
 import re
 import os
 import random
+import time
 from nose.tools import assert_equal
 
 from awscli.testutils import aws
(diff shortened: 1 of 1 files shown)


tests/integration/test_smoke.py
```diff
@@ -115,12 +116,18 @@
...
 
-def _aws(command_string):
+def _aws(command_string, max_attempts=1, delay=5, expected_rc=0):
     service = command_string.split()[0]
     env = None
     if service in REGION_OVERRIDES:
         env = os.environ.copy()
         env['AWS_DEFAULT_REGION'] = REGION_OVERRIDES[service]
+
+    for i in range(max_attempts - 1):
```
42B:
```python
for _ in range(max_attempts - 1):  # ?
```

tests/integration/test_smoke.py
```diff
@@ -115,12 +116,18 @@
 }
 
 
-def _aws(command_string):
+def _aws(command_string, max_attempts=1, delay=5, expected_rc=0):
```
kyleknap: I wonder if there is a better name we can use for `expected_rc`? Usually, I expect that it is making an assertion if an `expected_rc` is provided, but the value is being used to determine if we need a retry. So maybe something like `target_rc` or `desired_rc` that indicates we will stop making attempts until we hit that rc?

kyleknap:
Looks good. Just had a small comment.

tests/integration/test_smoke.py
JordonPhillips: sure

PR #3698, opened 2018-11-02
[CLIv2] Add end to end functionality for model completer

jamesls:
This allows you to actually test the completer end to end.  This likely isn't the real way we'd ship this, but to test this now:

```
# Recommend testing in a virtualenv to avoid perf issues with 'pip install -e'
$ virtualenv testcompleter && . testcompleter/bin/activate
$ pip install .
# This will create an index in ~/.aws/cli/cache/<version>.index
$ ./scripts/gen-ac-index
# For bash
$ complete -C aws_completer aws
```


jamesls:
Merged via 8295da85

PR #3702, opened 2018-11-06
[CLIv2] Initial commit of generating server side completions

jamesls:
This commit includes only the heuristics for generating
these completions.  It does not include the actual generated
modules.  That will be a large commit so I want to hold off on
sending that until this commit is reviewed.  Changes to this
script could change what we generate.


awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+                op_with_prefix_removed = op_name[len(prefix):]
+                return op_with_prefix_removed
+                break
+
+
+class BasicSingularize(object):
+    """Simple implementation of making a word singular.
+
+    Where possible, you should use nltk.stem.WordNetLemmatizer.
+    It's not included here as a default to avoid a hard dependenct
```
kyleknap: dependenct -> dependency

awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+                             resource_shape_name=resource_shape_name))
+            return r
+
+    def _remove_verb_prefix(self, op_name):
+        for prefix in self._RESOURCE_VERB_PREFIX:
+            # 'ListResources' -> 'Resources'
+            if op_name.lower().startswith(prefix):
+                op_with_prefix_removed = op_name[len(prefix):]
+                return op_with_prefix_removed
+                break
```
kyleknap: I don't think the `break` statement is needed here.

awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+                        if shape.type_name == 'list']
+        if len(list_members) != 1:
+            LOG.debug("Operation does not have exactly one list member, "
+                      "skipping: %s (%s)", op_name, list_members)
+            return
+        resource_shape_name = list_members[0]
+        list_member = output.members[resource_shape_name].member
+        op_with_prefix_removed = self._remove_verb_prefix(op_name)
+        if list_member.type_name == 'structure':
+            candidates = list_member.members
```
kyleknap: It might be good to break these if statements into separate internal methods for readability

awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+
+LOG = logging.getLogger(__name__)
+Resource = namedtuple('Resource', ['resource_name', 'ident_name',
+                                   'operation', 'jp_expr'])
+
+
+class ServerCompletionHeuristic(object):
+    # Prefix that typically indicates listing resources,
+    # e.g ListResources, DescribeResources, etc.
+    _RESOURCE_VERB_PREFIX = ('list', 'describe')
```
kyleknap: We may need to add `get`, but not sure how that affects the results. I know some services such as API Gateway use `get` instead of `list` and `describe`

awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+            resource = self._resource_for_single_operation(
+                op_name, service_model)
+            if resource is not None:
+                self._inject_resource(all_resources, resource)
+        return all_resources
+
+    def _generate_operations(self, op_names, resources, service_model):
+        # member_name -> resource
+        reverse_mapping = {}
+        for name, identifier_map in resources.items():
```
kyleknap: It would be nice to break this up into smaller internal methods for readability as well

awscli/autocomplete/autogen.py
```diff
@@ -0,0 +1,177 @@
...
+        # that has only one shape in the output that's of type list.
+        # It will then find that inspect that list member to find
+        # the best guess at which identifier maps to the resource.
+        # This is a best effort attempt and relies heavily on naming
+        # conventions.
+        op_model = service_model.operation_model(op_name)
+        output = op_model.output_shape
+        list_members = [member for member, shape in output.members.items()
+                        if shape.type_name == 'list']
+        if len(list_members) != 1:
```
kyleknap: How again do we model resources that have a required parameter in order to list (i.e. subresource)?

jamesls: Via the `parameters` key.  Right now we always generate an empty map so we're not taking advantage of this feature.

awscli/autocomplete/autogen.py
jamesls: I went ahead and added it.  It's not great but it's better than nothing.  Here's an example of api gateway: https://gist.github.com/jamesls/043bb24bd2a6efd427109f72be2d7eca

jamesls: Actually, I'll have to remove this for now.  Right now we make the assumption that the operations fall into two buckets, one for auto completing server side values, and the other for operations that *can* be auto-completed.  `Get` can be used for both cases so until there's more sophisticated logic here it can't be part of `_RESOURCE_VERB_PREFIX`.

jamesls:
@kyleknap Ok feedback incorporated.  I also added a new file `scripts/gen-server-completions` that's the main driver for generating these files.  I'll send a follow up PR for the generated files once we get this merged, it's a pretty large diff.

jamesls:
To simplify things for now, I added the inputParameters to the resources when they have required params, but I removed any operations that use a resource that has required params.  I expect to come back to this and try to auto-gen required params for operations, but for now this keeps things simple.

cc @kyleknap

Issue #3587, opened 2018-09-19
Tracking issue for AWS CLI v2


jamesls:
Version 1.0.0 of the AWS CLI was tagged just over 5 years ago. In that time, a lot has changed. We've received tons of feedback from customers. We've learned what features customers enjoy, and what parts of the CLI we'd like to change. There are changes we've wanted to make, but needed to wait until the next major version of the AWS CLI due to either the backwards incompatible nature or the large scope of the change.

Well, we're here now. This issue is to track all the work for CLI v2. We wanted to do this work on GitHub right from the beginning. Even though we're in the early stages of development, community feedback will play an important role in its development.

* Any AWS CLI v2 related issue/PR will be labeled with [v2](https://github.com/aws/aws-cli/labels/v2).
* We'll also reference this issue for any AWS CLI v2 related PRs so everyone can see all the changes we're making.
* All the development work will be in the [v2 branch](https://github.com/aws/aws-cli/tree/v2).

If you have an idea for CLI v2, whether it's a new feature or a change we couldn't make in v1 of the AWS CLI, please feel free to open a [GitHub issue](https://github.com/aws/aws-cli/issues/new) and let us know. You can also go through any issue labeled as [CLI v2](https://github.com/aws/aws-cli/labels/v2) and [+1](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) features that you'd like to see us implement.

Blog: https://aws.amazon.com/blogs/developer/aws-cli-v2-development/

MarkRose:
Highest our wish list would be to not require entering a 2fa code every hour when doing cross-account role-assumption despite the assumed role allowing for a longer session duration. Is there any chance #2748 can be looked at?

jamesls:
> Is there any chance #2748 can be looked at?

Yeah definitely.  Reading over #2748 it looks like the main thing we needed was service side support for longer role durations, which has since happened (https://aws.amazon.com/about-aws/whats-new/2018/03/longer-role-sessions/).  This should be possible now.

MarkRose:
That's awesome to hear. That's the one complaint I frequently get after implementing a centralized auth account.

tkelman:
Could the default setting on https://github.com/aws/aws-cli/pull/2619 be flipped now in v2 so `aws ecr get-login` works with recent docker versions?

dserodio:
[...]

PR #3706, opened 2018-11-08
[CLIv2] Support YAML output format

JordonPhillips:


setup.py
```diff
@@ -28,7 +28,8 @@ def find_version(*file_paths):
             'docutils>=0.10',
             'rsa>=3.1.2,<=3.5.0',
             's3transfer>=0.1.12,<0.2.0',
-            'PyYAML>=3.10,<=3.13']
+            'ruamel.yaml>=0.15.0,<0.16.0',
```
kyleknap: Should we be using 0.15.0? There's a note about staying away from it for now in their docs: https://yaml.readthedocs.io/en/latest/basicuse.html. May be worth putting a TODO about upgrading here.

tests/unit/output/test_yaml_output.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(self.parsed_response, parsed_output)
+
+    def test_empty_dict_response_prints_nothing(self):
+        self.parsed_response = {}
+        stdout = self.run_cmd('sts get-caller-identity', expected_rc=0)[0]
+        self.assertEqual(stdout, '')
+
+    def test_empty_list_prints_list(self):
+        self.parsed_response = []
+        stdout = self.run_cmd('sts get-caller-identity', expected_rc=0)[0]
```
kyleknap: For these ones should we be setting the `--output yaml` as well?

tests/unit/output/test_yaml_output.py
```diff
@@ -0,0 +1,84 @@
...
+        output = self.run_cmd(
+            'iam list-users --output yaml', expected_rc=0
+        )
+        parsed_output = yaml.safe_load(output[0])
+        self.assertEqual(self.parsed_response, parsed_output)
+
+    def test_empty_dict_response_prints_nothing(self):
+        self.parsed_response = {}
+        stdout = self.run_cmd('sts get-caller-identity', expected_rc=0)[0]
+        self.assertEqual(stdout, '')
```
kyleknap: Could you elaborate on why an empty dictionary is nothing but an empty list is still an empty list?

awscli/formatter.py
```diff
@@ -97,6 +97,30 @@ def _format_response(self, command_name, response, stream):
...
+        if not isinstance(response, (dict, set, list)):
+            # The yaml library will try to disambiguate when you dump a
+            # scalar value in a variety of ways. By default it will add
+            # an elipsis on a new line at the end of the file. While
+            # this is valid YAML and it does have its benefits,
+            # it's more useful to us to omit that. Setting indent to 0
+            # will cause it to omit the elipsis, but then it adds an
+            # extraneous newline at the beginning. So we dump to a
+            # string and trim it.
+            data = yaml.safe_dump(response, allow_unicode=True, indent=0)
```
kyleknap: What is the difference between `dump` and `safe_dump`? I tried looking through their docs with not much luck.

kyleknap:
Also it looks like the Travis build is failing

tests/unit/output/test_yaml_output.py
JordonPhillips: Keeping the same behavior as the json formatter. I imagine we don't want to print emtpy dicts because then all commands would have output, even if they return nothing. To get a list you'd have to be querying. I would be +0 on not printing empty lists.

setup.py
JordonPhillips: Might be some docs out of date. On the [readme](https://pypi.org/project/ruamel.yaml/) it says:

> For production systems you should pin the version being used with ``ruamel.yaml<=0.15``

More to the point, we need the newer versions to support 3.7. Would also like to use some of the new interface. How about we pin to a specific version for now?

sparrowt:
Looking forward to being able to use `--output yaml`, is there any timeline, even vaguely, for v2?

Perhaps https://github.com/aws/aws-cli/issues/3587 is the place to ask this, but I didn't want to spam such a popular issue.

agirlnamedsophia:
agreed, i'm also eager to know when i can use `--ouput yaml`!

PR #3714, opened 2018-11-09
[CLIv2] Add server side auto-completer

jamesls:
This PR support for a server side auto-completer along with several refactorings needed by this change.  I tried to keep the commits logical, but at a high level:

* Moved all the existing model-specific completer code in `autocomplete/local`.
* Updated the parser to support a `current_param` in the parsed result.  This lets us know if we're currently parsing values for a current parameter.
* Added a new subpackage in the completer for server side completions.

For the server side completer, I haven't added the indexing code for the `completions-1.json` files.  I've been using a simplified version that uses botocore session's directly for testing.  This PRs already fairly large, so I'll send the indexer code as a follow up PR.

The new content is primarily in `autocomplete/serverside/servercomp.py` + tests as well as `autocomplete/parser.py`


awscli/data/workmail/2017-10-01/completions-1.json
```diff
@@ -4,7 +4,7 @@
     "Aliase": {
       "operation": "ListAliases",
       "resourceIdentifier": {
-        "Aliases": "Aliases[]"
+        "Aliase": "Aliases[]"
```
kyleknap: Any ideas on how we got `Aliase` out of this?

awscli/autocomplete/serverside/servercomp.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ServerSideCompleter(BaseCompleter):
+    def __init__(self, completion_lookup, client_creator):
+        self._completion_lookup = completion_lookup
+        self._client_creator = client_creator
+
+    def complete(self, parsed):
+        """
+
+        :param parsed: The parsed CLI coommand the user has typed.
```
joguSD: I'm tempted to say leave this typo because it sounds like cool and command smashed together.

awscli/data/workmail/2017-10-01/completions-1.json
jamesls: So the bug was that we. weren't actually using the `nltk` singularizer to generate these values.  I think there's a tradeoff either way.  The NLTK version can't handle domain specific terms (e.g `Arns-> Arn`), but our version can't handle english edge cases.  I'd prefer to just rip out the nltk stuff for now until we can get it robust enough.  It does mean that cases like the above will still happen because we can't differentiate between `Names -> Name` vs. `Aliases -> Alias`.

awscli/autocomplete/serverside/servercomp.py
```diff
@@ -0,0 +1,148 @@
...
+    def _retrieve_remote_completion_data(self, parsed, completion_data):
+        # TODO: Handle global params that can affect how we create the client
+        # (region, endpoint-url, profile, timeouts, etc).
+        service_name = self._get_service_name_from_command(parsed)
+        client = self._client_creator.create_client(service_name)
+        py_name = xform_name(completion_data['operation'])
+        api_params = self._map_command_to_api_params(parsed, completion_data)
+        response = self._invoke_api(client, py_name, api_params)
+        if response:
+            return jmespath.search(completion_data['jp_expr'], response)
```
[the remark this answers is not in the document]
jamesls: The only way that would happen is if the top level key in the response was missing, for example: `{}` instead of `{"Instances": []}`.  I'm not aware of that ever happening in a response but I suppose it's technically possible.  I'll update the code to handle this.

jamesls:
@kyleknap feedback incorporated.

PR #3717, opened 2018-11-12
[CLIv2] Improve autogenerated completions

jamesls:
This updates the autogen module with better heuristics for generating autocompletions.

Note: The base branch is `v2-autocomplete`, so until https://github.com/aws/aws-cli/pull/3714 is merged, you'll see changes from that PR as well.  The last three commits are the changes associated with this PR:

* https://github.com/aws/aws-cli/commit/5169a98be227bfa041d4ae32fb4e3201a30cbc0a - Add a functional test to lint the completion files.  This will matter once we start hand writing these files, but also as a safety net to ensure the autogen module isn't generating invalid completions.
* https://github.com/aws/aws-cli/commit/cd2e10ad48cd06d5f102c764cfc3090d04830f39 - Autogen List/Describe operations.  I initially excluded these because I couldn't see a valid use case, but there's nothing technically wrong with completion information for List/Describe operations.  The main driver for this change was to allow us to autocomplete waiter params, which commonly use List/Describe operations.
* https://github.com/aws/aws-cli/commit/2937cdad2b5daec6a278d577bd703d856f798c70 - Change heuristic for finding resource Ids.  Some services such as `iam` have a resource such as `User` that has several response members that could be identifiers (`Arn, `UserId`, `UserName`).  We were previously picking the best one based on name similarity.  Now we generate identifiers for all output members on a struct, and then prune the identifiers at the end to remove any identifiers that aren't reference by an autocompletion operation.

Still not perfect, but I think these completions are better than what we have now.


awscli/data/ec2/2016-11-15/completions-1.json
```diff
@@ -1701,6 +1701,781 @@
...
+        ]
+      }
+    },
+    "DescribeRegions": {
+      "RegionNames": {
+        "completions": [
+          {
+            "parameters": {},
+            "resourceName": "Region",
+            "resourceIdentifier": "RegionName"
```
kyleknap: Hmmm for this one, I was not able to get the completions. Any idea on why? Are you able to get completions for `--region-names`?

awscli/data/cloudtrail/2013-11-01/completions-1.json
```diff
@@ -4,24 +4,228 @@
...
+            "resourceName": "Trail",
+            "resourceIdentifier": "Name"
+          }
+        ]
+      },
+      "S3BucketName": {
+        "completions": [
+          {
+            "parameters": {},
+            "resourceName": "Trail",
```
kyleknap: This does not look correct as well. I would assume this would end up being a Bucket resource.

tests/unit/autocomplete/test_autogen.py
```diff
@@ -556,3 +559,61 @@ def test_remove_operations_with_required_params(self):
...
+                'FooBarThingType': {
+                    'type': 'structure',
+                    'members': {
+                        # The DeleteFooBarThing accepts either an Id
+                        # or an Arn.
+                        'FooBarThingId': {'shape': 'String'},
+                        'FooBarThingArn': {'shape': 'String'},
+                        # However it doesn't accept a "Name", so this
+                        # identifier will be pruned from the response.
+                        'FooBarThingName': {'shape': 'String'},
```
kyleknap: It might be better to make the name of the member more explict in that it will be pruned (i.e. `FoorBarThingPrunedId`)

tests/functional/autocomplete/test_completion_files.py
```diff
@@ -0,0 +1,129 @@
...
+        for param_data in completions.values():
+            for comp_data in param_data['completions']:
+                resource_name = comp_data['resourceName']
+                resource_id = comp_data['resourceIdentifier']
+                assert_in(resource_name, resources)
+                identifiers = resources[resource_name]['resourceIdentifier']
+                assert_in(resource_id, identifiers)
+    # We also want to lint the 'resources' key as well.
+    for resource_data in resources.values():
+        assert_in(resource_data['operation'], service_model['operations'])
```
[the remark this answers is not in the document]
jamesls: We should yeah, do you want this as part of this PR?

awscli/data/ec2/2016-11-15/completions-1.json
```diff
@@ -1701,6 +1701,781 @@
...
+          }
+        ]
+      }
+    },
+    "DescribeClassicLinkInstances": {
+      "InstanceIds": {
+        "completions": [
+          {
+            "parameters": {},
+            "resourceName": "InstanceCreditSpecification",
```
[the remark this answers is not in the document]
jamesls: Yeah for a number of these changes, I don't think the existing approach is going to work, even with the modifications I've made.  The issue seems to be in the reverse mapping of resourceIdentifiers back to resource, and specifically that we can't assume that a resourceIdentifier for a resource (such as InstanceId) uniquely maps back to a single resource (Instance).  We need a smarter way to match up `DescribeClassicLinkInstances` to the appropriate resource type.  I'll have to think about how to best do this.

tests/functional/autocomplete/test_completion_files.py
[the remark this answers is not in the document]
kyleknap: If you already have something together, go ahead and add it to this PR. If not, we can put it in a followup PR

awscli/data/ec2/2016-11-15/completions-1.json
jamesls: Yeah it works for me, you'll need to regenerate the cache file.  Whenever new completions-1.json files are added/updated, you have to rerun `./scripts/gen-ac-index` again.

awscli/data/cloudtrail/2013-11-01/completions-1.json
jamesls: Conceptually yes, however we don't support cross resource references so it would have to be something you can get from the cloudtrail API.  In this case, it's a little weird as you arguably shouldn't be offering auto-completion of a `Create` operation using describe/list calls for the existing service.  That would only make sense if you were trying to create a resource that's similar to one of your existing resources.  I'm considering just having an excludes list for certain operation prefixes for now (e.g `Create`).

awscli/data/ec2/2016-11-15/completions-1.json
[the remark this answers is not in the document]
jamesls: Other thing to note, for this specific case one of the complications is that getting instance ids requires a `Reservations[].Instances[].InstanceId`, and we only go one level deep in the response structure.  We assume the identifier is a member of the structure of the associated list member in the response, e.g `Instances[].InstanceId`.

I'm actually thinking that ec2 might just need some handwritten overrides rather than trying to make the heuristics more sophisticated...

jamesls:
@kyleknap I incorporated most of the feedback except the EC2 case which I think is going to require some manual customizations, or significantly smarter heuristics.  Let me know what you think.

PR #3729, opened 2018-11-15
[CLIv2] Add logs tail command

kyleknap:
Allows you to grab and follow the most recent logs from all stream for a log group.

### DIFF
--- a/awscli/customizations/logs.py
+++ b/awscli/customizations/logs.py
@@ -283,7 +283,7 @@ class NoFollowLogEventsGenerator(BaseLogEventsGenerator):
 
 
 class FollowLogEventsGenerator(BaseLogEventsGenerator):
-    _TIME_TO_SLEEP = 1
+    _TIME_TO_SLEEP = 5
 
     def __init__(self, client, timestamp_utils, sleep=None):
         super(FollowLogEventsGenerator, self).__init__(client, timestamp_utils)
--- a/awscli/customizations/logs.py
+++ b/awscli/customizations/logs.py
@@ -1,0 +1,321 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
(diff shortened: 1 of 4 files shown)


awscli/customizations/logs.py
```diff
@@ -0,0 +1,321 @@
...
+class NoFollowLogEventsGenerator(BaseLogEventsGenerator):
+    def _filter_log_events(self, filter_logs_events_kwargs):
+        paginator = self._client.get_paginator('filter_log_events')
+        for page in paginator.paginate(**filter_logs_events_kwargs):
+            for log_event in page['events']:
+                yield log_event
+
+
+class FollowLogEventsGenerator(BaseLogEventsGenerator):
+    _TIME_TO_SLEEP = 1
```
jamesls: I would probably bump this up to at least 5 seconds.  We probably don't want to be running this command every second while we wait for more logs.

awscli/customizations/logs.py
```diff
@@ -0,0 +1,321 @@
...
+            return
+
+    def _do_filter_log_events(self, filter_logs_events_kwargs):
+        last_event_ids = []
+        while True:
+            response = self._client.filter_log_events(
+                **filter_logs_events_kwargs)
+            for event in response['events']:
+                # For the case where we've hit the last page, we will be
+                # reusing the most recent nextToken to continue polling. This
```
jamesls: Does this mean we can see the same nextToken twicein a row?  Wouldn't we get a `PaginationError` when we use the botocore paginators in the NoFollowLogEventsGenerator class because of duplicate next tokens?

awscli/customizations/logs.py
kyleknap: Yeah I can bump that up.

awscli/customizations/logs.py
kyleknap: Yes. Right now I'm not using the paginators in botocore. I have to do the pagination manually because paginators will stop after no pagination token is sent back in the response. Also based on the [docs](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html#CWL-FilterLogEvents-response-nextToken), it says that next tokens are vaild for 24 hours. So it should be fine to reuse them.

jamesls: > Yes. Right now I'm not using the paginators in botocore. 

I'm confused.  My question is about `NoFollowLogEventsGenerator`, which appears to be using the paginators (https://github.com/aws/aws-cli/pull/3729/files/1940387c79792b8b367618a8a87355fa02de14bb#diff-b85f7571d740176a2efa0d5e2e936d76R280).  Why isn't the duplicate next token not an issue in that case?

PR #3730, opened 2018-11-16
[CLIv2] Add core wizard runner

jamesls:
This adds the core engine for wizards which includes the planner and the executor.

Remaining work in follow up PRs:

* Integration with botocore loaders to discover available wizards for a service
* Custom CLI commands for any discovered wizards
* UI integration for prompting values with users (right now there's a placeholder prompter class that will definitely change)

### DIFF
--- a/awscli/customizations/wizard/core.py
+++ b/awscli/customizations/wizard/core.py
@@ -12,6 +12,8 @@
 # language governing permissions and limitations under the License.
 """Core planner and executor for wizards."""
 import jmespath
+import os
+
 from botocore import xform_name
 
 
@@ -136,6 +138,20 @@ class APICallStep(BaseStep):
         )
 
 
+class SharedConfigStep(BaseStep):
+    def __init__(self, config_api):
+        self._config_api = config_api
+
+    def run_step(self, step_definition, parameters):
+        if step_definition['operation'] == 'ListProfiles':
+            return self._config_api.list_profiles()
+        elif step_definition['operation'] == 'GetValue':
+            return self._config_api.get_value(
+                profile=step_definition['params'].get('profile'),
+                value=step_definition['params']['value'],
+            )
+
+
 class APIInvoker(object):
     """This class contains shared logic for the apicall step.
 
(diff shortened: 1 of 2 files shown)


awscli/customizations/wizard/core.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    def __init__(self, session):
+        self._session = session
+
+    def invoke(self, service, operation, api_params, plan_variables,
+               optional_api_params=None, query=None):
+        client = self._session.create_client(service)
+        resolved_params = self._resolve_params(
+            api_params, optional_api_params, plan_variables)
+        response = getattr(client, xform_name(operation))(**resolved_params)
```
joguSD: Might be worth adding a TODO here that we'll need to convert resolved parameters to the correct type, as everything read in from prompts is going to be a string. For now, this means we're limited to calling operations that only require string input. 

Might be worth exploring if typing in the wizard model or enforcing type via the service model is better (or maybe even both).

awscli/customizations/wizard/core.py
```diff
@@ -0,0 +1,237 @@
...
+    def run(self, plan):
+        self._parameters.clear()
+        steps_iter = self._steps(plan)
+        step = next(steps_iter)
+        while step is not self._STOP_RUNNING:
+            next_step_name = self._run_step(step)
+            step = steps_iter.send(next_step_name)
+        return self._parameters
+
+    def _steps(self, plan):
```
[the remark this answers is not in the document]
jamesls: Not a strong reason.  The main reason is that all the state variables related to keep tracking of the step transitions are local to the `_steps` method.  Otherwise you'd either have to mix them in as instance vars on the planner class or create a separate iterator object that tracks the state.

awscli/customizations/wizard/core.py
jamesls: Yeah good point.  I'll add a todo for now.  I think allowing type conversions in the wizard model would give the most control, but auto-converting based on the service model would be a nice convenience if you don't need special conversion behavior (which I would guess is the most common).

jamesls:
@joguSD pushed a change that switches from a coroutine to instance vars.  I think's a little more messy using instance vars but I don't feel that strongly either way.  Let me know which one you prefer.

awscli/customizations/wizard/core.py
```diff
@@ -24,38 +24,42 @@ def __init__(self, step_handlers):
         # These will hold the values we store from
         # running through the plan steps.
         self._parameters = {}
+        self._current_step_index = -1
```
joguSD: Seems a little strange to track these at the object level considering they're unique to a specific `run` call.

awscli/customizations/wizard/core.py
```diff
@@ -24,38 +24,42 @@ def __init__(self, step_handlers):
...
         self._parameters.clear()
-        steps_iter = self._steps(plan)
-        step = next(steps_iter)
+        self._current_step_index = -1
+        self._current_step_name = None
+        self._step_names = list(plan)
+        next_step_name = None
+        step = self._next_step(next_step_name, plan)
         while step is not self._STOP_RUNNING:
             next_step_name = self._run_step(step)
```
joguSD: Would it be better if `_run_step` always returned the next step? Inferring the next step when this value is `None` based off the index doesn't seem great.

Ideally, I think a step should always know what the next step is without the need for this to be tracked via external state. This should be easy to do just by populating the next step field for the implicitly ordered steps when we load the wizard. This removes the need for a `_STOP_RUNNING` sentinel value as `None` would imply there is no next step. This also vastly simplifies the `_next_step` function as there's no longer a difference between an explicit next step and an implicit next step.

jamesls: Capturing offline discussion:

After looking at both versions, I prefer the coroutine version because it keeps all that state local to a single method.

I don't think we should start pre-processing the wizards yet.  The wizard loading happens in a separate object, so we're now creating an implicit contract between what the runner accepts (i.e steps that always have a next step populated) vs what the spec actually supports (implicit next steps).  If anything you could have some sort of pre-processor that's self contained with a specific runner implementation that can fill in missing values (such as next steps) that make the runner logic simpler.  However, the only thing we'd have right now is filling in next steps, and I think it's much simpler to just keep this in a single method instead of introducing some sort of preprocessor object.

If the runner starts getting more complicated then I think we should reevaluate preprocessing.

PR #3742, opened 2018-11-21
[CLIv2] Add cli-dev command and wizard-dev command

jamesls:
This PRs adds the remaining functionality to actually run a wizard.  This includes several components.

* Adds the selectmenu from @joguSD, which is an extension over the built in prompt toolkit values.  This also includes pulling in prompt-tookit.
* Adds the runner and loader for wizards.
* Adds a `wizard-dev` command that allows you to run a wizard by directly specifying a path to the wizard file.
* Adds a top level `cli-dev` command.  This is a command that we can use to add misc. debugging and developer commands to the CLI.

### DIFF
--- a/awscli/customizations/devcommands.py
+++ b/awscli/customizations/devcommands.py
@@ -1,0 +1,34 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.customizations.commands import BasicCommand
+
+
+def register_dev_commands(event_handlers):
+    event_handlers.register('building-command-table.main',
+                            CLIDevCommand.add_command)
+
+
+# This is adding a top level placeholder command to add dev commands.
+# Each specific customization can add to this command via the
+# building-command-table event name.
+class CLIDevCommand(BasicCommand):
+    NAME = 'cli-dev'
+    DESCRIPTION = (
+        'Internal commands for AWS CLI development.\n'
+        'These commands are not intended for normal end usage. These '
+        'commands are used to help develop and debug the AWS CLI.\n'
+        'Do not rely on these commands, backwards compatibility '
+        'is not guaranteed.  Any of these commands under "aws cli-dev" '
+        'may be removed in future versions.\n'
+    )
+    SYNOPSIS = 'aws cli-dev [dev-command]'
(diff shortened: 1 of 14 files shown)


awscli/customizations/devcommands.py
```diff
@@ -0,0 +1,34 @@
...
+def register_dev_commands(event_handlers):
+    event_handlers.register('building-command-table.main',
+                            CLIDevCommand.add_command)
+
+
+# This is adding a top level placeholder command to add dev commands.
+# Each specific customization can add to this command via the
+# building-command-table event name.
+class CLIDevCommand(BasicCommand):
+    NAME = 'cli-dev'
```
kyleknap: Do we want to set `_UNDOCUMENTED` to `True` here to further emphasize the internal nature of it?

awscli/customizations/wizard/devcommands.py
```diff
@@ -0,0 +1,76 @@
...
+        'Internal command from developing, testing and debugging wizards.\n'
+        'This command is not intended for normal end usage. '
+        'Do not rely on this command, backwards compatibility '
+        'is not guaranteed.  This command may be removed in '
+        'future versions.\n'
+    )
+    ARG_TABLE = [
+        {'name': 'run-wizard',
+         'help_text': 'Run a wizard given a wizard file.',
+         'action': 'store',
```
kyleknap: Not sure if `action` or `cli_type_name` is needed. `store` and `string` should be the default for `action` and `cli_type_name` respectively.

awscli/customizations/wizard/selectmenu.py
```diff
@@ -0,0 +1,185 @@
...
+    menu.
+
+    :type max_height: int
+    :param max_height: The max number of items to show in the list at a time.
+
+    :returns: The selected element from the items list.
+    """
+    app_bindings = KeyBindings()
+
+    @app_bindings.add('c-c')
```
kyleknap: This causes issues for me on Python27. I'm getting the following output:
```
$ aws cli-dev wizard-dev --run-wizard file://role.yaml
Select type of trusted entity

Got 'c-c'
```

awscli/customizations/wizard/selectmenu.py
```diff
@@ -0,0 +1,185 @@
...
+from prompt_toolkit.layout import Layout, FloatContainer, Float
+from prompt_toolkit.layout.controls import UIControl, UIContent
+from prompt_toolkit.layout.screen import Point
+from prompt_toolkit.layout.dimension import Dimension
+from prompt_toolkit.key_binding.key_bindings import KeyBindings
+from prompt_toolkit.layout.margins import ScrollbarMargin
+from prompt_toolkit.layout.containers import ScrollOffsets, Window
+
+
+def select_menu(items, display_format=None, max_height=10):
```
kyleknap: It would also be nice if this was searchable via typing in a search query. It is pretty tedious when you can only scroll if you have a lot of options. How difficult would that be to add?

awscli/customizations/devcommands.py
jamesls: Personally no.  The documentation is really clear about this being an internal command that could be removed.

awscli/customizations/wizard/selectmenu.py
jamesls: Yeah I ran into that too the other day. cc @joguSD

awscli/customizations/wizard/selectmenu.py
jamesls: It's not really the intent of this widget to allow you to type something, it's specifically designed so you can't.

The way we were planning on handling this is at the UI layer that uses these widgets.  Basically the UI would be smart enough that if the number of choices is over a certain threshold it would switch to the default prompt-toolkit prompt that allows you to type.

kyleknap: Sounds fine to me

PR #3751, opened 2018-11-24
[CLIv2] Add custom completers for tail and ddb commands

kyleknap:
In order to get this working, I had to do a few things:

* Add support for positional parameters in our parser.
* Add support for loading custom completers
* Add support for the specific completers for `ddb put/select` and `logs tail`

### DIFF
--- a/awscli/autocomplete/custom.py
+++ b/awscli/autocomplete/custom.py
@@ -1,0 +1,21 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.customizations.dynamodb.autocomplete import add_ddb_completers
+from awscli.customizations.logs.autocomplete import add_log_completers
+
+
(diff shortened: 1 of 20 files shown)


jamesls:
In general I think this is good, just had a concern about the perf.  I'm seeing around anywhere from 38%-69% performance drops with this.  Is there anything simple we can do to keep performance similar to what it is now?  What if we add the custom completers last?  I think we should be careful not to affect completion times for existing commands.

For example:

**v2 branch**
```
$ python -m timeit -s "import awscli.autocomplete.main as m;" "m.autocomplete('aws ec2 describe-in')" > /tmp/results && tail -n 1 /tmp/results
100 loops, best of 3: 12.8 msec per loop
```
**PR**

```
$ python -m timeit -s "import awscli.autocomplete.main as m;" "m.autocomplete('aws ec2 describe-in')" > /tmp/results && tail -n 1 /tmp/results
10 loops, best of 3: 21.7 msec per loop
```

kyleknap:
I'm looking into it right now. I will put the customizations last in the list to be safe for now, but it looks like the reason it is getting slowed down is because I'm doing any additional select on the argument table looking for positional arguments on **every** command completion. I think I can be smarter about this and only do the query when it looks like there is a positional argument.

kyleknap:
@jamesls I think I fixed the issue. It was related to parsing of positional arguments. I was doing an additional select on the argument index for each command no matter if it was probably not a positional argument. These are the results I am seeing now:

**PR**
```
$ python -m timeit -s "import awscli.autocomplete.main as m;" "m.autocomplete('aws ec2 describe-in')" > /tmp/results && tail -n 1 /tmp/results
100 loops, best of 3: 15.9 msec per loop
```

**v2**
```
$ python -m timeit -s "import awscli.autocomplete.main as m;" "m.autocomplete('aws ec2 describe-in')" > /tmp/results && tail -n 1 /tmp/results
100 loops, best of 3: 16.1 msec per loop

```

PR #3753, opened 2018-11-24
[CLIv2]  Print region from imds in configure list command

kyleknap:
It looks like this when running from a clean EC2 instance:
```
$ aws configure list
access_key     ****************HDGD         iam-role    
secret_key     ****************rYB6         iam-role    
    region                us-west-2             imds    
```

### DIFF
--- a/awscli/customizations/configure/list.py
+++ b/awscli/customizations/configure/list.py
@@ -113,7 +113,12 @@ class ConfigureListCommand(BasicCommand):
         if region:
             # This makes the assumption that the only other possible place to
             # retrieve region in the CLI is via IMDS. If we add more region
-            # providers in the future this will have to be updated
+            # providers in the future this will have to be updated.
+            #
+            # TODO: Currently credentials use the value iam-role, but the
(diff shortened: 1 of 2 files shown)


awscli/customizations/configure/list.py
```diff
@@ -105,15 +105,35 @@ def _lookup_credentials(self):
...
+    def _lookup_region(self):
+        region = self._lookup_in_env_and_config('region')
+        if region:
+            return region
+        region = self._session.get_config_variable('region')
+        if region:
+            # This makes the assumption that the only other possible place to
+            # retrieve region in the CLI is via IMDS. If we add more region
+            # providers in the future this will have to be updated
+            return ConfigValue(region, 'imds', '')
```
jamesls: Can we add a todo is try and come up with consistent location names?  It seems odd that we have `iam-role` for the creds but `imds` for the region even though they basically come from the same source.

kyleknap: Sounds good. I'll add a todo before merging.

Issue #3759, opened 2018-11-26
GLACIER should be an allowed parameter to --storage-class


mourya92:
Amazon S3 now supports directly uploading objects to Glacier as per this snippet : 

[https://aws.amazon.com/s3/storage-classes/#Archive](url)
>.........You can upload objects directly to Amazon Glacier or use S3 Lifecycle policies.......... 

AWS CLI version : aws-cli/1.16.61 not yet supporting --storage-class=GLACIER in the high level command : aws s3 cp

` aws s3 cp direct_glacier_test s3://SAMPLE-BUCKET/direct_glacier_test --storage-class=GLACIER
usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help
aws: error: argument --storage-class: Invalid choice, valid choices are:

STANDARD                                 | REDUCED_REDUNDANCY                      
STANDARD_IA                              | ONEZONE_IA                              
INTELLIGENT_TIERING `

Upload Object to Glacier storage class : 
`aws s3api put-object --bucket SAMPLE-BUCKET --key direct_glacier_test --storage-class=GLACIER 
{
    "VersionId": "brdeDPuINHSul8OqiZCAp6s65zLFI4KX", 
    "ETag": "\"d41d8cd98f00b204e9800998ecf8427e\""
}`

Check the Object's storage class :
`aws s3api head-object --bucket SAMPLE-BUCKET --key direct_glacier_test
{
    "AcceptRanges": "bytes", 
    "ContentType": "binary/octet-stream", 
    "LastModified": "Mon, 26 Nov 2018 21:19:04 GMT", 
    "ContentLength": 0, 
    "VersionId": "brdeDPuINHSul8OqiZCAp6s65zLFI4KX", 
    "ETag": "\"d41d8cd98f00b204e9800998ecf8427e\"", 
    "StorageClass": "GLACIER", 
    "Metadata": {}
}`

Request to provide support for --storage-class=GLACIER in aws s3 cp command

jibowang:
why `put-object --storage-class=GLACIER` then `head-object`  command result :
```
{
    "AcceptRanges": "bytes", 
    "ContentType": "binary/octet-stream", 
    "LastModified": "Mon, 26 Nov 2018 21:19:04 GMT", 
    "ContentLength": 0, 
[...]

PR #3761, opened 2018-11-26
Add GLACIER to s3 transfer commands

mourya92:
Add the GLACIER option to the argument --storage-class on the three
s3 transfer commands sync, mv and cp.

#3759


joguSD:
Support for this went out in yesterday's release (v1.16.62).

PR #3792, opened 2018-12-11
Fix string formatting

daniel-s-ingram:


kdaily:
Thanks again @daniel-s-ingram, I rebased and merged this in #6437. Closing!

PR #3794, opened 2018-12-11
Added a colon for formatting

randyurbano:
Added a colon so that the example is formatted correctly in the output.


kdaily:
Looks like this change has been incorporated by the doc writers. Closing, thanks for bringing it to our attention!

PR #3800, opened 2018-12-15
AWS_REGION is not a supported variable

tb3088:
detritus like this only confuses people.


kdaily:
Hi @tb3088,

Quite a long delay here, and I apologize for that. 

The AWS CLI v2 does support the `AWS_REGION` variable via a region provider chain, where it takes precedence over `AWS_DEFAULT_REGION`.

https://github.com/aws/aws-cli/blob/v2/awscli/clidriver.py#L244-L264

I'll need to look further to see if this is something we'd change specifically for the CLI v1, or why it was used in these customizations in the first place.

tb3088:
I don't mind if you want to support BOTH of AWS_REGION and AWS_DEFAULT_REGION but ALL OF THE DOCS studiously mention ONLY AWS_DEFAULT_REGION. Only the JAVA implementation supports AWS_REGION (the OLD way of doing things) last I grep'd the source code. If AWS wants to support both, then DOCUMENT IT, and FIX ALL of the other language implementations to do so as well!

kdaily:
Hi @tb3088,

I agree that documentation should be up to date and consistent. `AWS_DEFAULT_REGION` was an AWS CLI only. As far as I know, all SDKs support `AWS_REGION`, some of them exclusively. Here's the documentation that I could find that explains how to get the region from the environment for all AWS SDKs:

- JS v3: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-region.html#setting-region-environment-variable
- JS v2: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html#setting-region-environment-variable
- Ruby: https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-setting-region
- Java v2: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/region-selection.html#automatically-determine-the-aws-region-from-the-environment
- Java v2: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html#automatically-determine-the-aws-region-from-the-environment
- Go v2: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-the-aws-region
- Go v1: https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-the-region
- .NET: https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-region-selection.html#globally

The only mentions of `AWS_DEFAULT_REGION` are for the Python SDK and the AWS CLI (which depends on the Python SDK):

- https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html
- https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-list

If you know of other documentation examples that show usage of `AWS_DEFAULT_REGION` outside of the AWS CLI or Python SDK, I would investigate them further!

tb3088:
331 $ git remote -vv
origin  https://github.com/aws/aws-sdk-ruby.git (fetch)

$ grep -ri aws_default_region | wc -l
296
$ grep -ri aws_region | wc -l
650


$ git remote -vv
origin  https://github.com/aws/aws-sdk-java-v2.git (fetch)
$ grep -ri aws_region | wc -l
45
$ grep -ri aws_default_region | wc -l
0


That's just ruby and Java. You'll find that ALL of the AWS SDKs are inconsistent in their use. If both are valid, then ALL code must behave the same way. And There should be EXACTLY ONE entrypoint that handles precedence of AWS_REGION over AWS_DEFAULT_REGION and ALL documentation fixed to comply with the actual code.

tim-finnigan:
Checking in as this topic came up recently here: https://github.com/boto/boto3/issues/3620#issuecomment-1462661383

> I'll share another documentation page referencing the cross-SDK compatibility of the `AWS_REGION` environment variable: https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html#feature-region-sdk-compat.
> 
> As noted there, boto3 and the AWS CLI v1 are exceptions in that they use `AWS_DEFAULT_REGION`. It was called out in the [v2 migration changes](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html) page that `AWS_REGION` was added for v2. It is an unfortunate inconsistency between the SDKs but as you suggested, it would be a breaking change to adjust the credential precedence at this point, and so I don't think this could be considered until the next major version of the SDK. Please let us know if you have any follow up questions or feedback.

If any documentation related to this is still inaccurate please let us know and we can escalate those issues to the appropriate team. But for the reasons mentioned above I think this PR should be closed.

PR #3805, opened 2018-12-17
Add adaptive timeout to 'ecs deploy' command

allisaurus:
Adjusts the `ecs deploy` command timeout to accommodate deployments with configured "wait" steps [[1](https://docs.aws.amazon.com/codedeploy/latest/APIReference/API_DeploymentReadyOption.html#CodeDeploy-Type-DeploymentReadyOption-waitTimeInMinutes)][[2](https://docs.aws.amazon.com/codedeploy/latest/APIReference/API_BlueInstanceTerminationOption.html#CodeDeploy-Type-BlueInstanceTerminationOption-terminationWaitTimeInMinutes)] that add up to more than 30 min (default wait).

New expected timeout is combined "wait" minutes + 10 minute buffer, up to 6 hours. 

## Tests

configured deployment wait less than 30 min (current behavior)
```
$~\aws-cli> aws ecs deploy --service testService --task-definition cd-task-def.json --codedeploy-appspec AppSpec.yml --cluster ecs-demo --codedeploy-deployment-group DpgECS-ecs-demo-testService --region us-east-1
Successfully registered new ECS task definition arn:aws:ecs:us-east-1:##########:task-definition/ecs-demo:137
Successfully created deployment d-TVXJ80###
Waiting for d-TVXJ80### to succeed...
Successfully deployed arn:aws:ecs:us-east-1:##########:task-definition/ecs-demo:137 to service 'testService'
$~>
```

configured deployment wait over 30 min (additional msg re: timeout)
```
$~\aws-cli> aws ecs deploy --service testService --task-definition cd-task-def.json --codedeploy-appspec AppSpec.yml --cluster ecs-demo --codedeploy-deployment-group DpgECS-ecs-demo-testService --region us-east-1
Successfully registered new ECS task definition arn:aws:ecs:us-east-1:#########:task-definition/ecs-demo:136
Successfully created deployment d-7BMEO7###
Waiting for d-7BMEO7### to succeed (will timeout after 45 minutes)...
Successfully deployed arn:aws:ecs:us-east-1:##########:task-definition/ecs-demo:136 to service 'testService'
$~>
```

configured deployment wait over 6 hours (default 30 wait + expected timeout):
```
$~\aws-cli> aws ecs deploy --service testService --task-definition cd-task-def.json --codedeploy-appspec AppSpec.yml --cluster ecs-demo --codedeploy-deployment-group DpgECS-ecs-demo-testService --region us-east-1
Successfully registered new ECS task definition arn:aws:ecs:us-east-1:##########:task-definition/ecs-demo:138
Successfully created deployment d-EHRQ6U###
Waiting for d-EHRQ6U### to succeed...

Waiter DeploymentSuccessful failed: Max attempts exceeded
$~>
```

### DIFF
--- a/awscli/customizations/ecs/deploy.py
+++ b/awscli/customizations/ecs/deploy.py
@@ -126,8 +126,9 @@ class ECSDeploy(BasicCommand):
             region_name=parsed_globals.region,
             verify=parsed_globals.verify_ssl)
 
-        self.wait_time = \
-            self._validate_code_deploy_resources(codedeploy_client)
+        self._validate_code_deploy_resources(codedeploy_client)
+        
+        self.wait_time = self._cd_validator.get_deployment_wait_time()
 
         self.task_def_arn = self._register_task_def(
             register_task_def_kwargs, ecs_client_wrapper)
@@ -204,7 +205,7 @@ class ECSDeploy(BasicCommand):
         validator = CodeDeployValidator(client, self.resources)
         validator.describe_cd_resources()
         validator.validate_all()
-        return validator.get_deployment_wait_time()
+        self._cd_validator = validator
 
 
 class CodeDeployer():
--- a/awscli/customizations/ecs/deploy.py
+++ b/awscli/customizations/ecs/deploy.py
@@ -127,7 +127,7 @@ class ECSDeploy(BasicCommand):
             verify=parsed_globals.verify_ssl)
 
         self._validate_code_deploy_resources(codedeploy_client)
-        
+
         self.wait_time = self._cd_validator.get_deployment_wait_time()
 
         self.task_def_arn = self._register_task_def(
@@ -210,10 +210,8 @@ class ECSDeploy(BasicCommand):
 
 class CodeDeployer():
 
-    MSG_WAITING = ("Waiting for {deployment_id} to "
-                   "succeed{custom_wait_msg}...\n")
-
-    MSG_CUSTOM_WAIT = " (will wait up to {wait} minutes)"
+    MSG_WAITING = ("Waiting for {deployment_id} to succeed "
+                   "(will wait up to {wait} minutes)...\n")
 
     def __init__(self, cd_client, appspec_dict):
         self._client = cd_client
@@ -294,15 +292,11 @@ class CodeDeployer():
 
     def wait_for_deploy_success(self, id, wait_min):
         waiter = self._client.get_waiter("deployment_successful")
-        wait_msg = ""
 
-        if wait_min > 30 and wait_min < MAX_WAIT_MIN:
-            wait_msg = self.MSG_CUSTOM_WAIT.format(wait=wait_min)
-
-        elif wait_min > MAX_WAIT_MIN:
+        if wait_min is not None and wait_min > MAX_WAIT_MIN:
             wait_min = MAX_WAIT_MIN
-            wait_msg = self.MSG_CUSTOM_WAIT.format(wait=wait_min)
-        else:
+
+        elif wait_min is None or wait_min < 30:
             wait_min = 30
 
         delay_sec = DEFAULT_DELAY_SEC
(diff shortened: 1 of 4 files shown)


awscli/customizations/ecs/deploy.py
```diff
@@ -122,7 +126,8 @@ def _run_main(self, parsed_args, parsed_globals):
             region_name=parsed_globals.region,
             verify=parsed_globals.verify_ssl)
 
-        self._validate_code_deploy_resources(codedeploy_client)
+        self.wait_time = \
+            self._validate_code_deploy_resources(codedeploy_client)
```
joguSD: It seems strange that the validator would return the wait time. I think I would prefer a separate call to get the wait time after validation has occured.

awscli/customizations/ecs/deploy.py
```diff
@@ -202,9 +204,16 @@ def _validate_code_deploy_resources(self, client):
...
         validator.validate_all()
+        return validator.get_deployment_wait_time()
 
 
 class CodeDeployer():
+
+    MSG_WAITING = ("Waiting for {deployment_id} to "
+                   "succeed{custom_wait_msg}...\n")
+
+    MSG_CUSTOM_WAIT = " (will timeout after {wait} minutes)"
```
joguSD: Is there any reason to not combine these messages and just print the amount of time to be waited every time?

awscli/customizations/ecs/deploy.py
```diff
@@ -309,6 +339,26 @@ def describe_cd_resources(self):
             raise exceptions.ServiceClientError(
                 action='describe Code Deploy deployment group', error=e)
 
+    def get_deployment_wait_time(self):
+
+        if (not hasattr(self, 'deployment_group_details') or
+                self.deployment_group_details is None):
+            return 0
```
joguSD: How do you feel about returning `None` here?

awscli/customizations/ecs/deploy.py
```diff
@@ -282,9 +291,30 @@ def update_task_def_arn(self, new_arn):
         appspec_obj[resources_key] = updated_resources
         self._appspec_dict = appspec_obj
 
-    def wait_for_deploy_success(self, id):
+    def wait_for_deploy_success(self, id, wait_min):
         waiter = self._client.get_waiter("deployment_successful")
-        waiter.wait(deploymentId=id)
+        wait_msg = ""
+
+        if wait_min <= 30 or wait_min > MAX_WAIT_MIN:
```
joguSD: If I'm understanding this correctly, if the `wait_min` is too low or too high we use the default waiter settings?
I would think if the max wait is too long, we would cap at the max wait. If it's too low we would use the min. 

I would think it works like this:
If there was no custom wait time information from the config we would use the default waiter settings.
If there is custom wait time information we would validate it based on the min/max and then use a customer waiter configuration. 

It's worth noting that if the default waiter configuration were to ever change the min of 30 here might become out of sync.

awscli/customizations/ecs/deploy.py
```diff
@@ -282,9 +291,30 @@ def update_task_def_arn(self, new_arn):
...
+            self._show_deploy_wait_msg(id, wait_msg)
+            waiter.wait(deploymentId=id)
+        else:
+            wait_msg = self.MSG_CUSTOM_WAIT.format(wait=wait_min)
+            delay_sec = DEFAULT_DELAY_SEC
+            max_attempts = (wait_min * 60)/delay_sec
```
joguSD: Let's pad the `/` with spaces.

awscli/customizations/ecs/deploy.py
allisaurus: Capping at max wait time makes sense. Also good point re: if the default waiter changes. I can change it so we're always calling `waiter.Wait()` with a `WaiterConfig`

awscli/customizations/ecs/deploy.py
allisaurus: Do you mean a separate API call or a separate private method in `_run_main` that retrieves the wait_time? Since the `CodeDeployValidator` retrieves all the info we need to determine the wait time, I'd rather do that latter, perhaps by keeping the validator around as an attribute of the main `ECSDeploy` obj so we can use it later. I'll add that change so you can LMK if it's enough of an improvement.

awscli/customizations/ecs/deploy.py
allisaurus: Original though was it may be helpful to be visually distinct from the default behavior, which is currently 30 min (vs. configured min + 10). However, I don't really see a downside to showing it all the time, and that would decrease the # of code paths we need to maintain & test. Will make that change.

allisaurus:
@joguSD addressed remaining feedback. Please let me know if there are additional comments/questions. Thank you!

allisaurus:
Hey @joguSD , any update or ETA on when this could get another look?

Issue #3807, opened 2018-12-19
Support "--date" option for S3 "cp" command


btbroot:
I would like to implement the functionality defined below.

The S3 service in AWS has an option to retain previous versions of objects. `aws-cli` includes a sub-command for working with s3 buckets: `aws s3`. Currently you can list, copy, remove, and sync files and directories between your local system and the s3 bucket. The command doesn't, however, provide a way to work with old versions of objects within a bucket.

This project is to add new functionality to the AWS CLI to allow for a date field in the `aws s3 cp` command. For example:

```aws s3 cp --date 'Tue Dec 11 13:04:01 PST 2018' --recursive s3://my.bucket.name/ /my/local/directory```

Without the `--date` option, the `aws s3 cp` command retrieves the most recent version of all specified objects. When this new `--date` option is specified the command should retrieve the state of the specified objects as they existed at the date and time specified.

JordonPhillips:
This would be really neat, especially if for some auditing reason you needed to know the state of your bucket at a given point in time. That said, it would likely be *tremendously* slow and expensive once `--recursive` comes into the mix.

There's a couple other features like that where implementation is feasible but execution is slow and costly. Maybe we could do something like give a warning and prompt for confirmation in those cases, and have a flag to turn that off for automation.

btbroot:
Thanks @JordonPhillips, I'll consider that

btbroot:
After a few tries, it's working as expected (the acceptance tests pass) and coverage is preserved, all checks are green.

Could anyone review?

tim-finnigan:
Checking in on this issue, thanks for your patience. I'm wondering about the intended use case and whether this would be the best approach. Couldn't you just use [aws s3api list-object-versions](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/list-object-versions.html) and [aws s3api copy-object](https://docs.aws.amazon.com/cli/latest/reference/s3api/copy-object.html) with the specified `versionId` values? Or are there other approaches that might address your use case @btbroot?

PR #3821, opened 2018-12-28
For "s3" command introduces new "--data" argument (in transfer argume…

btbroot:
Closes aws/aws-cli#3807

> Instead of selecting the last version of the object (or objects, for a recursive operation), selects the most recent 'version(s) by the date specified in ISO 8601 format. Makes sense only for a versioned bucket as a source, otherwise gets ignored.


btbroot:
All TravisCI checks are failing because of:

```The job exceeded the maximum log length, and has been terminated.```

Could anyone take a look?

PR #3823, opened 2019-01-02
Add hidden alias for old deploy commands

JordonPhillips:
There are a pair of commands for codedeploy that have github in their
names which used to use git-hub. This is fixed in botocore, and this
lets customers of the CLI continue to use the old command names.

Related to boto/botocore#1660


JordonPhillips:
This requires that botocore pr for the tests to pass

PR #3830, opened 2019-01-05
No longer special case Python 3.7

kevin-brown:
Now that the Xenial builds of Travis CI are officially supported, it makes sense to run ~all~ most builds on it. Python 3.7 is not supported on Trusty.

Sudo is no longer required to run builds on Xenial, so we can set this globally back to `false` and not have to worry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,22 @@
 language: python
-dist: xenial
 sudo: false
 
-python:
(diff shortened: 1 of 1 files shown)


kdaily:
We're no longer using Travis to run tests, they have been migrated to GitHub actions.

PR #3831, opened 2019-01-07
example command doesn't work without having a --cli-input-json parameter

hm2602:
before the filepath


stealthycoin:
Thanks for the PR, looks like this was already fixed.

PR #3890, opened 2019-01-26
For "s3" command introduces new "--date" argument (in transfer argume…

btbroot:
…nts group).

        'Instead of selecting the last version of the object (or objects, for '
        'a recursive operation), selects the most recent version(s) capped by '
        'the date specified in one of formats supported by dateutil '
        '(see: https://dateutil.readthedocs.io/en/stable/parser.html). '
        'Makes sense only for a versioned bucket as a source, '
        'otherwise gets ignored.'

Closes aws/aws-cli#3807


btbroot:
Rebase is needed

PR #3916, opened 2019-02-08
Update create-volume.rst

lindseyspirit:
Fixed command syntax.


kdaily:
Hi @lindseyspirit, apologies for the lengthy delay. This example has subsequently been changed. However the syntax `--parameter value` is correct and used extensively in the examples. 

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters.html

Closing this out!

PR #3918, opened 2019-02-10
Create list-stack-resources example.

TylerSustare:
This file will help show folks using the `list-stack-resources` command what is used for `--stack-name`


TylerSustare:
image

PR #3920, opened 2019-02-11
Updated docs to add --user-attributes to example command

DavidHoude:
Added --user-attributes as required by awscli to example.

`aws: error: argument --user-attributes is required
`


justindho:
Hi @DavidHoude, thanks for the PR. Apologies for taking so long for us to get back to you. It looks like this example command was updated in #4258. As a result, I will be closing this PR.

PR #3921, opened 2019-02-11
Remote Etag sync strategy

elayton:
Remote-to-remote syncs fail to copy if the source and destination files have the same byte size and if the destination file has a more recently modified date.  This patch fixes that with a new sync strategy and command line switch, --remote-etag, that will compare the ETag values of remote files and trigger a sync if they are different.

It will not matter with remote-to-local or local-to-remote syncs.


tim-finnigan:
Thanks for creating this PR and for your patience in waiting on a response. It looks like this one fell off of the team's radar, but I recently brought it up for discussion as we are cleaning up the backlog. Now that S3 has extended [checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/), the CLI team is more inclined to explore using those for improvements to sync strategy rather than using Etags. Further investigation into this is still needed but unfortunately this PR cannot be considered at this time.

PR #3924, opened 2019-02-12
Updated create a volume with tags example

lindseyspirit:
to show commands for Linux and Windows


justindho:
I spoke with @lindseyspirit internally, and this PR is no longer applicable. Closing the PR.

PR #3929, opened 2019-02-14
Updated output for describe-activations

awserik:
The output provided in the example is no longer accurate to that which is returned by the SSM service.  This request is to simply update the output of the example

### DIFF
--- a/awscli/examples/ssm/describe-activations.rst
+++ b/awscli/examples/ssm/describe-activations.rst
@@ -11,14 +11,15 @@ Output::
   {
     "ActivationList": [
         {
-            "IamRole": "AutomationRole",
-            "RegistrationLimit": 10,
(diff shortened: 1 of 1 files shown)


awscli/examples/ssm/describe-activations.rst
```diff
@@ -11,14 +11,15 @@ Output::
...
-            "IamRole": "AutomationRole",
-            "RegistrationLimit": 10,
             "ActivationId": "bcf6faa8-83fd-419e-9534-96ad14131eb7",
-            "ExpirationDate": 1487887903.045,
-            "CreatedDate": 1487801503.045,
-            "DefaultInstanceName": "MyWebServers",
-            "Expired": false,
+			"Description": "test",
+			"DefaultInstanceName": "MyWebServers",
+			"IamRole": "AutomationRole",
```
joguSD: Seems like the indentation is messed up here. Could we fix this up?

PR #3946, opened 2019-02-18
For "s3" command introduces new "--date" argument (in transfer argume…

btbroot:
…nts group).

        'Instead of selecting the last version of the object (or objects, for '
        'a recursive operation), selects the most recent version(s) capped by '
        'the date specified in one of formats supported by dateutil '
        '(see: https://dateutil.readthedocs.io/en/stable/parser.html). '
        'Makes sense only for a versioned bucket as a source, '
        'otherwise gets ignored.'

Closes aws/aws-cli#3807

### DIFF
--- a/awscli/customizations/s3/fileformat.py
+++ b/awscli/customizations/s3/fileformat.py
@@ -1,4 +1,5 @@
 # Copyright 2013 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2018 Transposit Corporation. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You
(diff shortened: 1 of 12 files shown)


btbroot:
@JordonPhillips, any chance this gets reviewed any time soon? Thanks!

btbroot:
Ping

tim-finnigan:
Checking in as there hasn't been any activity here in a while. There are now file conflicts that would need to be addressed in the PR. Also we'd want a response here on the related issue before proceeding: https://github.com/aws/aws-cli/issues/3807#issuecomment-1499488861. We can reopen that issue for further tracking of the feature request.

PR #3949, opened 2019-02-18
Created 4 new command examples and added examples or updated output to 8 other existing commands

awserik:
Created 4 new command examples and added examples or updated output to 8 other existing commands


jamesls:
Merged via c4f0988c (I rebased your commits to keep the commit message < 79 chars which is why it doesn't show up as merged).

PR #3953, opened 2019-02-18
Added new examples for 5 Session Manager commands

awserik:
Added examples for 5 Session Manager related commands.


awserik:
Closing as there is an issue with the existing RST files.  Will be recreating RST files and create a new PR.

PR #3985, opened 2019-03-05
Update EMR Configurations schema to support three-layer nested configurations

JunyangLi:
*Issue #, if available:*

*Description of changes:*
This change allows configurations given to instance groups and instance fleets
when creating an EMR cluster to be an up-to-three-layer nested structure with
the most inner layer of configuration always being empty. In this way, syntax
check of create-cluster command is consistent with current EMR API validation
for configuration objects.

### DIFF
--- a/awscli/customizations/emr/argumentschema.py
+++ b/awscli/customizations/emr/argumentschema.py
@@ -32,13 +32,23 @@ CONFIGURATIONS_CLASSIFICATION_SCHEMA = {
     "description": "Application configuration classification name",
 }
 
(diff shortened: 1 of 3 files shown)


tim-finnigan:
Hi @JunyangLi thanks for creating this PR and for your patience. Could you please rebase this PR against the latest branch for the team’s further consideration going forward?

tim-finnigan:
Closing PR as we have not heard back - this PR would require rebasing as mentioned above as well as a review from the EMR team.

PR #3986, opened 2019-03-05
awscli doesn't use the configurations of default section in .aws/config

t2y:
*Description of changes:*
I tried to configure s3 default settings with `~/.aws/config`, but awscli doesn't use them.

For example, I set as below configurations. That means requesting to virtual host style address with v2 signature. I know v2 signature will end soon. I just want to confirm to change default behavior since awscli uses v4 signature by default.

```bash
$ aws configure set default.s3.signature_version s3
$ aws configure set default.s3.addressing_style virtual
```

* https://docs.aws.amazon.com/cli/latest/topic/s3-config.html
* https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html

As the result, we can see these configurations in `~/.aws/config`.

```ini
[default]
s3 =
    signature_version = s3
    addressing_style = virtual
```

Then, I tried to run awscli for s3api get-object like this.

```bash
(awscli3) $ aws --debug --endpoint-url https://s3-ap-northeast-1.amazonaws.com --profile myaws --region ap-northeast-1 s3api get-object --bucket mybucket --key mykey data
```

I confirmed that this request use path style with v4 signature from debug log.

```
2019-03-06 08:34:27,860 - MainThread - botocore.client - DEBUG - Using S3 path style addressing.
...
2019-03-06 08:34:27,865 - MainThread - botocore.auth - DEBUG - Calculating signature using v4 auth.
```

I read awscli and botocore source code. It seems that botocore has an ability to load s3 specific configurations and merge some `Config` as needed.

* https://github.com/boto/botocore/blob/43689e64864368f6bf9d273acddbc58c4c215896/botocore/session.py#L708
* https://github.com/boto/botocore/blob/43689e64864368f6bf9d273acddbc58c4c215896/botocore/configloader.py#L117

So I think awscli should pass default configuration as `Config` to botocore client.

After I modified this, then I confirmed it works seeing the debug log.

```
2019-03-06 08:47:21,157 - MainThread - botocore.client - DEBUG - Using S3 virtual host style addressing.
...
2019-03-06 08:47:21,159 - MainThread - botocore.auth - DEBUG - Calculating signature using hmacv1 auth.
```

I'm not sure background of why awscli doesn't use default configuration, or how to enable default configuration when s3 sub command is used. Also, could you tell me if my operation is wrong.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -18,6 +18,7 @@ import botocore.session
 from botocore import __version__ as botocore_version
 from botocore.hooks import HierarchicalEmitter
 from botocore import xform_name
+from botocore.config import Config
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @t2y,

Sorry for the delay in responding to this. In your example command, you've specified a profile `--profile myaws` to be used. The CLI (and Python SDK) don't treat the `default` profile as a fallback, its settings are only used when it's specified as the profile to use (`--profile default`, or when not specifying any specific profile). This is not a behavior we are looking to change.

PR #4009, opened 2019-03-15
Disable Printing For Sensitive Configuration Inputs

mehmetoguzderin:
*Issue:* Configure command does not disable printing for sensitive configuration inputs. This can lead to problems when using aws-cli in computational notebook cells, where printed console values are stored by default and can be automatically committed to git repositories.

*Description of changes:* This pull request adds compat_getpass functionality and leverages the existing branching in get_value of configure to disable printing using new function.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -18,6 +18,7 @@ import platform
 import zipfile
 import signal
 import contextlib
+import getpass
 
 from botocore.compat import six
 #import botocore.compat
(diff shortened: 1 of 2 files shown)


mehmetoguzderin:
@JordonPhillips @jamesls Why does travis-ci give error in this one (stalls at prompt)? I tested the case it fails in Python 2.7 and Python 3.6 on Windows and Linux without problems and I think this one is necessary for setup related notebook projects.

PR #4019, opened 2019-03-20
Support for packaging GlueJob resources

### LINKED ISSUE #3570: Add `cloudformation package` support for AWS::Glue::Job resource type
Instead of having to upload Glue scripts to S3 manually, it would be nice if it could be done with `aws cloudformation package` command the same way it's done for `AWS::CloudFormation::Stack` or `AWS::Lambda::Function` resource types:

```yaml
Resources:
  MyGlueJob:
    Type: AWS::Glue::Job
    Properties:
      Name: MyGlueJob
      Command:
        ScriptLocation: ./gluescript.scala
        Name: glueetl
        ...
```

cfbarbero:
Fixes #3570 

The AWS::Glue::Job resource has a ScriptLocation property under a Command property.  This adds support for packaging that file. 

In order to support this the base Resource property was updated to support nested properties.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -106,17 +106,13 @@ def get_nested_property_value(resource_dict, property_name):
     :return:                Value of the property
     """
 
-    # Support nested properties by allowing '.' in the PROPERTY_NAME
-    if '.' in property_name:
-        sub_property_names = property_name.split('.')
-        property_value = resource_dict.get(sub_property_names[0], None)
-        for sub_property_name in sub_property_names[1:]:
-            if property_value:
-                property_value = property_value.get(sub_property_name, None)
-    else:
-        property_value = resource_dict.get(property_name, None)
-
-    return property_value
+    if '.' not in property_name:
+        return resource_dict.get(property_name, None)
+    
+    property_names = property_name.split('.', 1)
+    # default to empty dictionary so recursion has something to operate on
+    sub_property_dict = resource_dict.get(property_names[0], {}) 
+    return get_nested_property_value(sub_property_dict, property_names[1])
 
 
 def set_nested_property_value(resource_dict, property_name, new_value):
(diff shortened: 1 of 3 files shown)


awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -94,6 +94,55 @@ def parse_s3_url(url,
...
+
+    :param resource_dict:   Dictionary containing resource definition
+    :param property_name:   Property name of CloudFormation resource where this
+                            local path is present
+
+    :return:                Value of the property
+    """
+
+    # Support nested properties by allowing '.' in the PROPERTY_NAME
+    if '.' in property_name:
```
jfuss: Have you thought about doing some recursion here instead? Seems like a better way forward and would remove the explicit if else branch on "if . is in the string". 

You can use `property_name.split('.', 1)`. The base case would be the list with 1 item (might need a `None` check). I find the nest if's with loops like this generally harder to read.

cfbarbero: Thanks @jfuss.  I've updated the get_nested_property_value and set_nested_property_value to use recursion instead of the nested for/if structure.

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -94,6 +94,51 @@ def parse_s3_url(url,
                      "{0}".format(url))
 
 
+def get_nested_property_value(resource_dict, property_name):
```
kyleknap: For both the `get` and `set` methods, would it be possible to use [JMESPath](https://github.com/jmespath/jmespath.py) instead? It is a JSON query language we use in the CLI codebase. Specifically, we could replace calls to `get_nested_property_value` to: `jmespath.search(property_name, resource_dict)` and calls to `set_nested_property_value` to `botocore.utils.set_value_from_jmespath`. If we use these functions instead we should be able to cut down the amount of code we need to add and not have to worry about explicitly testing the helper functions as the new functions we would be calling are already tested.

cfbarbero: Sure, let me look into what that will take.  I've used jmespath with the --query output from the CLI, but never worked with the jmespath library.

cfbarbero: @kyleknap, I've refactored this to use the recommended methods.  Thanks for pointing that out, that significantly simplified this implementation!

PR #4030, opened 2019-03-27
setup.py: use read() function to find README.rst for long description

anonymouse64:
*Description of changes:*

There is already a helper function in setup.py for reading files, and if we use this helper, then loading this setup.py from a different directory will still work. As it is, I get a file not found error if I run python from a different directory.


JordonPhillips:
Looks good to me!

PR #4034, opened 2019-03-29
Update request-certificate.rst

joneps:
*Description of changes:*
Need to clarify that the validation domain must be the superdomain


kdaily:
Hi @joneps, thanks for the PR! I apologize for the lengthy delay. It looks like the example command was subsequently fixed by our doc writers, but they stayed with the description as a domain instead of a superdomain. Closing this out as it has been fixed. Thanks for bringing it to their attention!

PR #4060, opened 2019-04-08
Emit Non Aggregate Keys when Using Text Output

### LINKED ISSUE #3962: --output text results in "None" while --output json gives correct result on cloudformation changesets
Hi,

the following gives "CREATE_COMPLETE" which is correct

`aws cloudformation describe-change-set --change-set-name website-production-alb-a-1-0-7 --stack-name website-production-alb-a`

while this gives "None" which is incorrect

`aws cloudformation describe-change-set --change-set-name website-production-alb-a-1-0-7 --stack-name website-production-alb-a --output text`

`aws --version`
`aws-cli/1.16.110 Python/3.7.2 Darwin/18.2.0 botocore/1.12.100`

colleague has aws-cli/1.16.1 which also gives correct result on text.

dstufft:
Fixes #3962 

Changes the ``TextFormatter`` so that it will properly emit non aggregate keys in the response set upon the first iteration of a paginated response.

### DIFF
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -231,8 +231,12 @@ class TextFormatter(Formatter):
         try:
             if is_response_paginated(response):
                 result_keys = response.result_keys
-                for page in response:
-                    current = {}
+                for i, page in enumerate(response):
+                    if i > 0:
+                        current = {}
+                    else:
+                        current = response.non_aggregate_part
+
                     for result_key in result_keys:
                         data = result_key.search(page)
                         set_value_from_jmespath(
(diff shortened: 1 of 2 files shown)


kyleknap:
@dstufft The changes look fine to me. I also tested it out and it fixed the issue. Do you have any idea on why codecov is complaining we are missing a line?

dstufft:
Yea. I think I just need to add another page to my test.

dstufft:
I figured out why it wasn't testing the other path, my stored responses didn't have a ``NextToken``, so it wasn't actually consuming the second page of results. Adding that fixed it.

Issue #3553, opened 2018-09-09
An official Docker Image with the AWS CLI for use in CI/CD scenarios


matti:
I read Issue #3529 and #3291 and saw those were closed, with the only reaction hinting it was 'not that complicated'. However, the comment also acknowledged that doing this yourself would run the risk of being out of date. Apart from exactly that point, I would also like to point out that, for commercial users, having an official Amazon image is hugely preferential to "/aws-cli:latest".

In my case, I would be using this in a Google Cloud Build because it is far superior than AWS CodeBuild.

nscavell:
I am here because I too am looking for an official aws-cli docker image to use in my CI/CD environment. Instead I am now going to create ANOTHER pipeline to build a docker image which includes the aws cli when I could just reference the official one in my original pipeline.

Also...someone else gets it too https://hub.docker.com/r/google/cloud-sdk.

justnance:
@matti and @nscavell, Thank you for following up on this topic.  Sounds like there enough interest in this feature request to keep this open.  This issue will be used to track an official Docker image for the AWS CLI.  Please +1 it to show interest and help us prioritize this. 

Thanks.

matti:
Isn't +1 considered harmful ;)

Anyway this is the third (?) issue created about this...

Deamon:
👍  add my +1 

I have to create my own docker image to only include awsCLI in my CI. I would prefere an official one

tim-palmer:
+1

FraaJad:
+1

mathiewz:
+1

derickson82:
+1

riznob:
+1

robert-smith-nike:
+1

harsha91:
+1

ecimionatto:
+1

shaunforan:
+1

B0yc3y:
+1

Also for the mean time I just pushed this. Its Alpine 3.8 (Post fix) + aws cli.

https://hub.docker.com/r/boycey/alpine-aws-cli/

matti:
@justnance enough?

ina-stoyanova:
+1

lunarxlark:
+1

davidconde:
+1

nhoughto:
Another reason is when using an OS like coreOS that has no package management the idiomatic way of running things is via a container. I'm surprised the need for this would even be questioned, is an obvious omission. 👍

+1

[...]

PR #4062, opened 2019-04-08
Adding the Dockerfile

marcieltorres:
*Issue #3553*

*Description of changes:*
So, I'm creating the Dockerfile to `aws-cli`, it's a simple Dockerfile but it's very helpful to CI/CD :).

For some tests, I'm pushing the image to my account in the docker hub: `https://hub.docker.com/r/marcieltorres/aws-cli`

Would be nice if the AWS push this image to the docker hub official account :)

*Special thanks [@shikow](https://github.com/shikow)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM alpine:3.6
+
+RUN mkdir -p /awscli
+COPY . /awscli
+
+WORKDIR /
+
+RUN apk --no-cache update && \
+    apk --no-cache add python py-pip py-setuptools && \
+    pip install --upgrade pip && \
+    pip --no-cache-dir install -e /awscli/ && \
+    rm -rf /var/cache/apk/*


jdel:
I absolutely support the very much needed idea of having an official docker image of the `awscli` which is justified only by the need to run `aws ecr get-login` to pull from any image from ECR.

I would love to see the `awscli-cwlogs` plugin installed so we can also pipe arbitrary scripts output to CW with the same container. It would be very nice to have this on minimal linux distributions that have no python / cloudwatch agent.

I install it in a custom image with:

```
[...TRUNCATED...]
 && pip install --no-cache-dir awscli awscli-cwlogs \
 && aws configure set plugins.cwlogs cwlogs
[...TRUNCATED...]
```

Provided the EC2 instance has the proper role, the following would stream pretty much any output to CW:

```
$ echo "whoop whoop" | docker run -i --rm aws/awscli:latest aws logs push --region eu-west-1 --log-group-name any-name --log-stream-name test-cwlogs
```

jdel:
Also two ideas to make this a bit more future-proof: 
  - alpine:3.6 is currently [end of support on 01/05/2019](https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases) it would be better to use 3.9
  - python2 as we all know will be gone next year and it would be wise to setup the docker image to use directly python3. 

The following `Dockerfile` works pretty well for me. Happy to make a PR on your repo if needed.

```Dockerfile
FROM alpine:3.9

COPY . /awscli

RUN apk --no-cache update && \
    apk --no-cache add python3 && \
    pip3 --no-cache-dir install -e /awscli/ awscli-cwlogs && \
    aws configure set plugins.cwlogs cwlogs && \
    rm -rf /var/cache/apk/*
```

isaiahgrant:
Nice to see that there is already a PR out there!

Would you consider basing from [an official python image](https://hub.docker.com/_/python) like `python:3-alpine`?
This would save work during build time as we would no longer need to install python:
```
// We could potentially remove these lines.
apk --no-cache update && \
apk --no-cache add python py-pip py-setuptools && \
```

kdaily:
We have implemented a Docker container image for the AWS CLI v2. We do not have plans to support a Docker container image for the v1 client.

PR #4067, opened 2019-04-10
Update create-replication-task.rst

bpd0018:
The task identifier provided on the command line and the task identifier returned in the output should match.


justindho:
Hi @bpd0018,

Thanks for the PR. It looks like this example is now outdated and has been replaced by a new example in #5252. As a result, I will be closing this PR.

PR #4071, opened 2019-04-11
Special case docs for single element lists

jamesls:
Some params have lists that can only be a single
element (presumably to be forward compatible).  We can
detect this special case and provide better examples.
With this change, here's an example of what this looks like:

```
       --rebuild-workspace-requests (list)
          The WorkSpace to rebuild. You can specify a single WorkSpace.

       Shorthand Syntax:

          --rebuild-workspace-requests WorkspaceId1

       JSON Syntax:

          [
            {
              "WorkspaceId": "string"
            }
            ...
          ]

```

Previously we'd have `--my-arg Arg1 Arg2 Arg3`.

NOTE: This will fail until https://github.com/boto/botocore/pull/1712 is merged.

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -453,6 +453,10 @@ class ParamShorthandDocGen(ParamShorthand):
                 self._uses_old_list_case(
                     service_id, operation_name, cli_argument.name):
             member_name = list(model.member.members)[0]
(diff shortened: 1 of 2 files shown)


jamesls:
Yep looks like travis is good now.  Merging.

PR #4092, opened 2019-04-22
Unset termination protection in create-cluster

tokaaron:
The purpose of this change is to rely on the EMR service for determining the termination protection value when running the following CLI command:

```
aws emr create-cluster [...]
```

Currently the RunJobFlow `JobFlowInstancesConfig` model treats `TerminationProtected` as an optional value, but RunJobFlow requests from create-cluster will always send `false` for this field unless this is overridden in the CLI using `--no-termination-protected`.

This is a cosmetic fix to the EMR CLI.  The current behaviour of the CLI and EMR is not changing.

https://docs.aws.amazon.com/emr/latest/APIReference/API_JobFlowInstancesConfig.html

### DIFF
--- a/awscli/customizations/emr/createcluster.py
+++ b/awscli/customizations/emr/createcluster.py
@@ -210,30 +210,33 @@ class CreateCluster(Command):
                 parsed_args.auto_terminate is False):
             parsed_args.no_auto_terminate = True
 
-        instances_config['KeepJobFlowAliveWhenNoSteps'] = \
-            emrutils.apply_boolean_options(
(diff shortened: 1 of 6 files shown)


tim-finnigan:
Hi @tokaaron thanks for creating this PR and for your patience. Could you please rebase this PR against the latest branch for the team’s further consideration going forward?

tim-finnigan:
Closing PR as we have not heard back - this PR would require rebasing as mentioned above as well as a review from the EMR team.

Issue #2976, opened 2017-11-22
ruby-2.0 won't work for recent ubuntu


dictcp:
https://github.com/aws/aws-cli/blob/59e30779bc290a2f4f4d124c79144b8b77464700/awscli/customizations/codedeploy/systems.py#L216

we should not hardcode to install a ruby 2.0 while ignoring ruby >2.0, eg ruby2.3

Ref: 
https://github.com/aws/aws-codedeploy-agent/issues/61

joguSD:
@ravirajataws @feverLu

tim-finnigan:
Checking in, here is a related issue: https://github.com/aws/aws-codedeploy-agent/issues/174. After discussing this with a member of the CodeDeploy team I was told that the issue has now been addressed. For more guidance please refer to this comment: https://github.com/aws/aws-codedeploy-agent/issues/174#issuecomment-1116378405

PR #4115, opened 2019-04-29
Fix for Code Deploy Ruby2.0 dependency for Ubuntu 18 / Ubuntu 16

thejhnz:
***Issue #, if available:*** #2976 https://github.com/aws/aws-codedeploy-agent/issues/174

***Description of changes:*** Change package install to just "ruby", so that an on premises deploy install works on Ubuntu 16 and Ubuntu 18. 

Ruby should be installed anyway, as running the code deploy install script will do ruby checks. https://github.com/aws/aws-codedeploy-agent/blob/866632b98447312962e629c31da94e756127c620/bin/install#L79

Might be best to do something similar to the code deploy agent, by checking 2.*? Not sure of the implications of changing from "ruby2.0" to just "ruby" for older versions of Ubuntu.

### DIFF
--- a/awscli/customizations/codedeploy/systems.py
+++ b/awscli/customizations/codedeploy/systems.py
@@ -213,7 +213,7 @@ class Linux(System):
 class Ubuntu(Linux):
     def _update_system(self, params):
         subprocess.check_call(['apt-get', '-y', 'update'])
-        subprocess.check_call(['apt-get', '-y', 'install', 'ruby2.0'])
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating this PR - after discussing this issue with a member of the CodeDeploy team (for more info see https://github.com/aws/aws-cli/issues/2976#issuecomment-1524097240) this can now be closed.

PR #4147, opened 2019-05-13
Fix #3618 s3uploader writing to stdout

### LINKED ISSUE #3618: aws cloudformation prints S3 upload progress to STDOUT
When running `aws cloudformation package`, the output must not include the S3 upload progress.

It currently includes the combined output of the S3 file upload as well as packaged CloudFormation itself making it impossible to use via STDOUT.

See the full repro at https://github.com/dnagir/cf-package-stdout-messed-up

Workaround - use `--output-template-file` option.

cthulhuology:
The package command writes the template out to stdout.  Having the
status written on stderr results in the standard practice of piping
stdout to a template file to break.

eg.  aws cloudformation package --template-file cloudformation.yaml
--s3-bucket goehd-test > cloudformation.pkg

This results in the value

\rUploading ...

In the first line of the output with deploy complaining about
mapped values not being allowed in <unicode string>.

Redirecting this to stderr reverts the behavior to be compliant with
standard practice.

*Issue #3618*

*Description of changes:*

### DIFF
--- a/awscli/customizations/s3uploader.py
+++ b/awscli/customizations/s3uploader.py
@@ -220,8 +220,8 @@ class ProgressPercentage(BaseSubscriber):
         with self._lock:
             self._seen_so_far += bytes_transferred
             percentage = (self._seen_so_far / self._size) * 100
-            sys.stdout.write(
+            sys.stderr.write(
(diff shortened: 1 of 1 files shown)


justindho (approved):
Thanks, @cthulhuology!

PR #4148, opened 2019-05-13
Added region argument to eks update-kubeconfig

### LINKED ISSUE #4145: eks update-config - generates incorrect kube-config - missing aws-iam-authenticator command
Version: aws-cli/1.16.156 Python/2.7.5 Linux/3.10.0-957.12.1.el7.x86_64 botocore/1.12.146
kubectl : 1.12.7
iam-aws-authenticator: 1.12.7

```bash
aws --region eu-west-1 eks update-kubeconfig --name XXXXXX
```

it generates 

```
      args:
      - eks
      - get-token
      - --cluster-name
      -  XXXXXX
      command: aws ---> it should be aws-iam-authenticator
```

Obviously whatever the attempt with the get-token - it should not break the previous mode
or give me the choice
```
      args:
      - token
      - -i
      - xxxxx
      command: aws-iam-authenticator
```
The above makes kubectl invoke the aws cli but results to

```
You must specify a region. You can also configure your region by running "aws configure".
```

micahhausler:
*Issue #, if available:*

#4145

*Description of changes:*

The region of an EKS cluster will never change, so we add the --region
flag for users who have not setup a region via the config file or
environment variable. Fixes #4145

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -101,7 +101,7 @@ class UpdateKubeconfigCommand(BasicCommand):
     ]
 
     def _display_entries(self, entries):
-        """ 
+        """
         Display entries in yaml format
 
         :param entries: a list of OrderedDicts to be printed
@@ -248,7 +248,7 @@ class EKSClient(object):
                 client = self._session.create_client("eks")
             else:
                 client = self._session.create_client(
-                    "eks", 
+                    "eks",
                     region_name=self._globals.region,
                     endpoint_url=self._globals.endpoint_url,
                     verify=self._globals.verify_ssl
@@ -287,6 +287,9 @@ class EKSClient(object):
         Return a user entry generated using
         the previously obtained description.
         """
+
+        region = self._get_cluster_description().get("arn").split(":")[3]
+
         generated_user = OrderedDict([
             ("name", self._get_cluster_description().get("arn", "")),
             ("user", OrderedDict([
@@ -294,10 +297,12 @@ class EKSClient(object):
                     ("apiVersion", API_VERSION),
                     ("args",
                         [
+                            "--region",
+                            region,
                             "eks",
                             "get-token",
                             "--cluster-name",
-                            self._cluster_name
+                            self._cluster_name,
                         ]),
                     ("command", "aws")
                 ]))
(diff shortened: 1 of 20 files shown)


awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -287,17 +287,22 @@ def get_user_entry(self):
         Return a user entry generated using
         the previously obtained description.
         """
+
+        region = self._get_cluster_description().get("arn").split(":")[3]
```
kyleknap: Are we guaranteed to always have the arn available to us? In line below (294 in this diff), there is a `get()` method call with a default of `""` if `arn` is missing, which is why I ask.

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -287,17 +287,22 @@ def get_user_entry(self):
...
+
         generated_user = OrderedDict([
             ("name", self._get_cluster_description().get("arn", "")),
             ("user", OrderedDict([
                 ("exec", OrderedDict([
                     ("apiVersion", API_VERSION),
                     ("args",
                         [
+                            "--region",
+                            region,
```
kyleknap: When the region is set to a non-`us-east-1` region such as `us-west-2`, what is the endpoint url in the presigned url? If it is signed for the global endpoint (i.e.`sts.amazonaws.com`), we should be fine. But if it is using the regional endpoint (i.e. `sts.us-west-2.amazonaws.com`), then there may be some backwards compatiblity concerns if the account does not have the regional endpoints activated. I'm also assuming the previous executable`aws-iam-authenticator` only ever signed for the global STS endpoint.

awscli/customizations/eks/update_kubeconfig.py
micahhausler: Yes, this is computed based off of the cluster name, and will always be present.

awscli/customizations/eks/update_kubeconfig.py
micahhausler: For non-us-east-1, [`get_token`](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/eks/get_token.py#L95-L128) already uses the regional endpoint. This change is to match the behavior from `aws-iam-authenticator` where the customer might not have the AWS cli configured with a region

kyleknap: I just double checked this. We should be fine here. Looking at the url in the base64 encoded from running a command such as:
```
aws eks get-token --cluster-name foo --region us-west-2
```
The URL that gets presigned uses `https://sts.amazonaws.com/` and signs with `us-east-1`, which is what I would expect for backwards compatibility with `aws-iam-authenticator`.

PR #4149, opened 2019-05-14
Fixed link and updated formatting

ssennettau:
*Issue #, if available:*
Nil

*Description of changes:*
Updated the example document for the 'aws cloudformation update-stack' command. Fixed a link to AWS Documentation, added a new link, and capitalized the stack name to improve readability.

### DIFF
--- a/awscli/examples/cloudformation/update-stack.rst
+++ b/awscli/examples/cloudformation/update-stack.rst
@@ -1,16 +1,16 @@
 **To update AWS CloudFormation stacks**
 
-The following ``update-stack`` command updates the template and input parameters for the ``mystack`` stack::
+The following ``update-stack`` command updates the template and input parameters for the ``MyStack`` stack::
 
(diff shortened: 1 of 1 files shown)


justindho:
Hi @ssennettau,

Thanks for the PR, and apologies that it took so long to get back to you. It looks like a link to the docs for CloudFormation stack updates has since been added in the [update-stack example docs](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/update-stack.html#examples).

With regards to the casing for stack names used in the examples, we would like to maintain consistency between the different examples. For example, the examples for [describe-stacks](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/describe-stacks.html#examples) and [create-stack](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html#examples) use all lowercase strings, so we will stick with that convention for now. It would likely require a larger effort in the future to update the naming convention across all docs.

PR #4159, opened 2019-05-16
Fixed eks get-token to use provided session

### LINKED ISSUE #4163: aws eks get-token doesn't work with --profile
When going to use `aws --profile profilename eks get-token --cluster-name eksClusterName` I receive the following error `Unable to locate credentials. You can configure credentials by running "aws configure".`. However, when I set the ENVVAR of AWS_DEFAULT_PROFILE to the profile I want, the command works fine.

I verified other eks commands work fine with the --profile option. For example: `aws --profile profilename eks describe-cluster --name eksClusterName` works as expected.

aws --version                                             
aws-cli/1.16.160 Python/3.7.0 Darwin/18.2.0 botocore/1.12.150

micahhausler:
Fixes #4163
N/A

*Description of changes:*

When trying to use get-token with an MFA role and AWS_PROFILE set, it
would prompt for a MFA token on every call. The command now uses the
built-in session in order to leverage cached assume role credentials.

The command will still be able to assume an additional role by handling
`--role-arn`, but only for non-MFA roles.

### DIFF
--- a/tests/unit/customizations/eks/test_get_token.py
+++ b/tests/unit/customizations/eks/test_get_token.py
@@ -12,10 +12,11 @@
 # language governing permissions and limitations under the License.
 
 import mock
-from mock import patch
+from mock import patch, call
 import base64
 import datetime
 import botocore
+import json
 
 from awscli.testutils import unittest, capture_output
 from awscli.customizations.eks.get_token import (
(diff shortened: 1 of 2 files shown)


AirbornePorcine:
Please merge this soon, this is making working with EKS extremely painful :)

micahhausler:
We will update the EKS docs, but in order to use MFA with credential caching, you'll have to do the following:

```ini
[profile default]
region  = us-west-2
# creds from credential_process,  ~/.aws/credentials, or elsewhwere

[profile MFARole]
region  = us-west-2
mfa_serial = arn:aws:iam::012345678901:mfa/mfa-device-arn
source_profile = default
role_arn = arn:aws:iam::012345678901:role/AssumeableMFARole
```

The user will need to modify the `users` section of the kubeconfig with either `--profile` or `AWS_PROFILE` in the env section:

```yaml
users:
- name: arn:aws:eks:us-west-2:012345678901:cluster/my-cluster
  user:
    exec:
      args:
      # either this flag, or the env var below
      - --profile
      - MFARole
      - --region
      - us-west-2
      - eks
      - get-token
      - --cluster-name
      - my-cluster
      command: aws
      # env below, or --profile above
      env:
      - name: AWS_PROFILE
        value: MFARole
```

The `--role-arn` field on `aws eks get-token` will not be compatible with MFA roles

AirbornePorcine:
Hey, just a note - when I ran `aws eks update-kubeconfig` with a `--profile` parameter, it automatically added the AWS_PROFILE env portion you listed.

tests/unit/customizations/eks/test_get_token.py
```diff
@@ -103,3 +114,33 @@ def test_token_sess(self):
             token_no_prefix.encode()
         ).decode()
         self.assert_url_correct(decrypted_token, True)
+
+class TestGetTokenCommand(unittest.TestCase):
+
+    @patch('awscli.customizations.eks.get_token.uni_print')
```
JordonPhillips: You should use `testutils.capture_output` rather than patching `uni_print` directly.

tests/unit/customizations/eks/test_get_token.py
```diff
@@ -88,13 +96,16 @@ def test_token(self):
...
     def test_token_no_padding(self, mock_presigned_url):
-        generator = TokenGenerator(REGION, self._session_handler)
+        generator = TokenGenerator(REGION, self._session)
         tok = generator.get_token(CLUSTER_NAME, None)
         self.assertTrue('=' not in tok)
 
-    def test_token_sess(self):
-        generator = TokenGenerator(REGION, self._assuming_handler)
-        token = generator.get_token(CLUSTER_NAME, "RoleArn")
+    @patch.object(botocore.session.Session, 'create_client')
```
JordonPhillips: Rather than patching session, you should inject a mock session that returns the client you want.

tests/unit/customizations/eks/test_get_token.py
```diff
@@ -103,3 +114,33 @@ def test_token_sess(self):
             token_no_prefix.encode()
         ).decode()
         self.assert_url_correct(decrypted_token, True)
+
+class TestGetTokenCommand(unittest.TestCase):
+
+    @patch('awscli.customizations.eks.get_token.uni_print')
+    @patch.object(TokenGenerator, 'get_token', return_value='k8s-aws-v1.aHR0cHM6Ly9zdHMuYW1hem9uYXdzLmNvbS8=')
```
JordonPhillips: Rather than patching the token generator, you should inject a mock that gives you the behavior you want. I think the problem here is that the `TokenGenerator` class takes region as an argument to `__init__` instead of `get_token` which means you can't accept the `TokenGenerator` as an optional argument to construct `GetTokenCommand`. If you do a small refactor to fix that, then it will be easy to inject a very simple mock in this test.

tests/unit/customizations/eks/test_get_token.py
micahhausler: Is it ok to keep this patch? I'd have to inject a lot of mocks to not only handle the `create_client` case, but also mock all of the request signing required types, and I haven't figured out all of the numerous mocks that would be needed to be created and passed through. (`botocore.credentials.Credentials`, `session.get_component('event_emitter')`, etc)

tests/unit/customizations/eks/test_get_token.py
micahhausler: Done

tests/unit/customizations/eks/test_get_token.py
micahhausler: Done

tests/unit/customizations/eks/test_get_token.py
joguSD: I think this is difficult to patch as a consequence of the `SessionHandler` being a bit redundant now that we're using the default session provided by the CLI.

 Right now it's being used for two purposes:
1) Access to an `sts` client to make the assume role call
2) A vessel to carry the credentials object retrieved from assume role.

Furthering this, it seems a little strange that we would need to use the `RequestSigner` class directly (thus exposing more innards of the session / increasing complexity & testing burden).

From what I can tell it looks like we're doing this because we need to inject a custom header? If this is the only reason, simply registering a customization handler to the `sts` client on `before-sign` that injects this header seems to be a much more direct approach as it would allow us to just use the `generate_presigned_url` included in boto3.

I think there's quite a bit of refactoring / simplification we can do on this customization, though I'm leaning on the side of that being a follow up PR and not blocking this one.

PR #4171, opened 2019-05-21
New and updated examples for Amazon Chime

bisdavid:


awscli/examples/chime/batch-suspend-user.rst
```diff
@@ -0,0 +1,13 @@
+**To suspend multiple users **
```
jamesls: Looks like this is invalid rst:

```
Line 1: Inline strong start-string without end-string.
  **To suspend multiple users **
```

bisdavid: This is now fixed.

bisdavid:
This should be fixed.  Please merge.

PR #4180, opened 2019-05-22
Use random key name in s3 integ metadata tests

jamesls:
Helps reduce false positives.


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -563,8 +563,8 @@ def test_copy_metadata_directive(self):
         # For comparing expires timestamp.
         self.override_parser(timestamp_parser=identity)
         bucket_name = _SHARED_BUCKET
-        original_key = 'foo.txt'
-        new_key = 'bar.txt'
+        original_key = random_chars(6)
+        new_key = random_chars(6)
```
JordonPhillips: I know it's very unlikely that these will be duplicates, but why not just append a character to the new key to make sure?

jamesls: Updated.

Issue #4182, opened 2019-05-22
aws eks get-token regresses providing token expiration


tbarrella:
With
```
aws-iam-authenticator token -i <cluster>
```
the output includes an "expirationTimestamp" key in the token "status", but with
```
aws eks get-token --cluster-name <cluster>
```
that field is missing. As a result, aws-cli >1.16.154 undoes https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/160, which is necessary to prevent issues when using files from `update-kubeconfig` with Go clients for >15 minutes. Would it be hard to restore "expirationTimestamp"?

sgundapu:
@tbarrella is this what you are seeing?
https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/133#issuecomment-428336101

Is your application failing because of client-go is caching the token?
It makes sense for us to add the expirationTimestamp along with the token, since client side code is using that to refresh the token. I will look into adding that back.

tbarrella:
Yeah, with the same application I'd also observed what that comment describes using aws-iam-authenticator until switching from v0.3.0 to v0.4.0. Thank you!

justnance:
@tbarrella and @sgundapu - Thank you for posting this issue.  It appears the CLI does not have control over the resolution of this issue and it is being addressed under [issue #133](https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/133#issuecomment-428336101) under the kubernetes-sigs/aws-iam-authenticator repo.

tbarrella:
Sorry, I'm confused because issue #133 had already been closed and was addressed (with the fix released in aws-iam-authenticator v0.4.0) before aws-cli 1.16.155 was released. To maybe clarify or restate, another way to put the issue is that
```
aws eks update-kubeconfig
```
provides a kubeconfig with an exec command that no longer provides an expiration timestamp with its tokens as it did before 1.16.155. Since #133 is already closed, is there another open issue for fixing this behavior of awscli?

sgundapu:
@justnance there is some work on our end in the CLI to add "expirationTimestamp", I will close the issue once that is pushed out into the CLI and verified.

justnance:
@tbarrella - We are still working on getting this resolved. Thanks for your patience.

justnance:
Related PR #4141 has been merged but pending release.

sgundapu:
[...]

PR #4203, opened 2019-06-04
Add expirationTimestamp to get-token CLI

sgundapu:
https://github.com/aws/aws-cli/issues/4182

Add expirationTimestamp to get-token CLI

### DIFF
--- a/awscli/customizations/eks/get_token.py
+++ b/awscli/customizations/eks/get_token.py
@@ -14,7 +14,7 @@ import base64
 import botocore
 import json
 
-from botocore import session
+from datetime import datetime, timedelta
 from botocore.signers import RequestSigner
 from botocore.model import ServiceId
 
@@ -28,6 +28,8 @@ AUTH_SIGNING_VERSION = "v4"
 
 # Presigned url timeout in seconds
 URL_TIMEOUT = 60
+
+TOKEN_EXPIRATION_MINS = 14
 
 TOKEN_PREFIX = 'k8s-aws-v1.'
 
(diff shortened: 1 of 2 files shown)


awscli/customizations/eks/get_token.py
```diff
@@ -61,11 +62,18 @@ def _run_main(self, parsed_args, parsed_globals):
             parsed_args.role_arn
         )
 
+        # By default STS signs the url for 15 minutes so we are creating a
+        # rfc3339 timestamp with expiration in 14 minutes as part of the token, which
+        # is used by some clients (client-go) who will refresh the token after 14 mins
+        token_expiration = datetime.utcnow() + timedelta(minutes=14)
```
kokhang: Can you make the 14 a constant?

Also https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/160/files has a lot of things going on. Can you make sure we are doing something similar? So that we dont risk introducing more unexpected things

sgundapu: I will update it to a constant.

Yeah we seem to have covered whatever aws-iam-authenticator is doing on the client side.

awscli/customizations/eks/get_token.py
```diff
@@ -29,6 +29,8 @@
 # Presigned url timeout in seconds
 URL_TIMEOUT = 60
 
+TOKEN_EXPIRATION_MINS = 14
```
kyleknap: Can you elaborate on how `14` was chosen here? I looked at this PR: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/160 and while it was originally hardcoded `14` as well, it was updated to subtract `1` minute off of the original presigned URL expiration time. I wonder if we should follow suit here?

sgundapu: Unlike pre-signed URLs in S3, STS does not honor X-Amz-Expires header (which is part of the sigv4 auth headers) and they have a default timeout of 15 minutes.

kyleknap: Gotcha. Thanks for clarifying.

PR #4218, opened 2019-06-06
Fixed dev identified issue

browndjaz:
*Issue #, if available:*
I changed the parameter after --query based on dev feedback

### DIFF
--- a/awscli/examples/directconnect/describe-connection-loa.rst
+++ b/awscli/examples/directconnect/describe-connection-loa.rst
@@ -4,7 +4,7 @@ The following example describes your LOA-CFA for connection ``dxcon-fh6ayh1d``. 
 
 .. code::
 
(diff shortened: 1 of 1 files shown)


jamesls:
Hi, are you sure this change is correct?

The documentation for this shows that there is a `loa` top level structure:

https://docs.aws.amazon.com/cli/latest/reference/directconnect/describe-connection-loa.html

```
OUTPUT
       loa -> (structure)
          The   Letter  of  Authorization  -  Connecting  Facility  Assignment
          (LOA-CFA).

          loaContent -> (blob)
              The binary contents of the LOA-CFA document.

          loaContentType -> (string)
              The standard media type for the LOA-CFA document. The only  sup-
              ported value is application/pdf.

```

This is also what's documented in the service's API docs: https://docs.aws.amazon.com/directconnect/latest/APIReference/API_DescribeConnectionLoa.html

Can you please run this command and share the output without the `--query` argument so we can confirm?

Issue #4175, opened 2019-05-22
Do not recommend `sudo pip`


weskerfoot:
https://github.com/aws/aws-cli#installation

Line 42 of the README recommends using `sudo pip install awscli`.

This is likely to completely break your Python installation! It is considered bad practice to mix system Python packages with ones installed via pip. This should not be recommended. `--user` should be recommended first. If you want to mention the possibility of installing with `sudo pip` then it should also be mentioned that it *will* cause breakage.

JordonPhillips:
How it affects your system is highly dependent on the system and how you installed python. Even when it's a problem, the problem is that you're using pip at all.

For CLI v2 we're planning on distributing standalone installers for all platforms to avoid these and other issues.

bengaywins:
When you say standalone installers, do we mean distro packaging via repos? Or just like the current self-contained tarball one can download for 1.x? Personally I much prefer the former rather than the latter as it now allows proper package management.

morancj:
@bengaywins I strongly agree: I've opened [4842] for the consideration of @JordonPhillips and colleagues.

PR #4224, opened 2019-06-07
Changing sudo pip to pip in readme.rst file.

kshithijiyer:
fixes: #4175

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ The easiest way to install aws-cli is to use `pip`_ in a ``virtualenv``::
 
 or, if you are not installing in a ``virtualenv``, to install globally::
 
-    $ sudo pip install awscli
+    $ pip install awscli
 
 or for your user::
 
--- a/README.rst
+++ b/README.rst
@@ -55,7 +55,7 @@ you can run::
     On OS X, if you see an error regarding the version of six that came with
     distutils in El Capitan, use the ``--ignore-installed`` option::
 
-        $ sudo pip install awscli --ignore-installed six
+        $ pip install awscli --ignore-installed six
 
 
 This will install the aws-cli package as well as all dependencies.  You can


README.rst
```diff
@@ -39,7 +39,7 @@ The easiest way to install aws-cli is to use `pip`_ in a ``virtualenv``::
 
 or, if you are not installing in a ``virtualenv``, to install globally::
 
-    $ sudo pip install awscli
+    $ pip install awscli
```
merwok: This probably won’t work, plus it’s not right in this part of the doc (it’s the same as the command just before).

`pip install --user awscli` could be an alternative that works without virtual env and without messing us the system (or being denied)

kshithijiyer: So should we just remove it or leave it as it is?

merwok: I don’t understand the question.

`sudo pip install` is bad, `pip install` won’t work.  More thought is needed (e.g. with the `install --user` idea)

kdaily:
Hi @kshithijiyer,

Apologies for the delay. We're cleaning up our backlog according to our [new contribution process](https://github.com/aws/aws-cli/pull/6828). I understand the sentiment of this change, and agree with @merwok that just dropping the `sudo` is probably not the best. I would possibly just drop this command entirely and rely on suggesting that users use a virtual environment. The logic here is that's what the AWS CLI v1 installation guide also suggests that, as well as a [`--user` install](https://docs.aws.amazon.com/cli/v1/userguide/install-linux.html#install-linux-pip). It does not direct users to install system-wide without using an operating system's package manager.

I think we should draw a better distinction between installing for development purposes here in the GitHub repository,  and direct end users more towards the content in the Installation section of the AWS CLI User Guide.

Given the lag between opening this and now, I'm happy to make these changes, but I can leave it for you to do as well since it's a minimal change from what you already have. Let me know how you'd like to proceed!

kdaily:
After some further discussion, we'd be even happier to remove the entire installation section, save for things that a developer would use, and redirect all end users to the User Guide installation section.

tim-finnigan:
Given that we haven't heard back on this PR, I'm going to close it due to inactivity. As suggested above, there are broader improvements to the installation documentation that could be made here and I think that can be in a separate PR.

PR #4228, opened 2019-06-10
Fixed EC2 examples for modifying user data with the CLI

bisdavid:

### DIFF
--- a/awscli/examples/ec2/modify-image-attribute.rst
+++ b/awscli/examples/ec2/modify-image-attribute.rst
@@ -1,31 +1,38 @@
-**To make an AMI public**
+**Example 1: To make an AMI public**
 
-This example makes the specified AMI public. If the command succeeds, no output is returned.
+The following ``modify-instance-attribute`` example makes the specified AMI public. ::
 
-Command::
+    aws ec2 modify-image-attribute \
+        --image-id ami-5731123e \
+        --launch-permission "Add=[{Group=all}]"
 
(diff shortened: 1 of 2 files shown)


awscli/examples/ec2/modify-image-attribute.rst
```diff
@@ -1,31 +1,54 @@
...
+        --launch-permission "Remove=[{UserId=123456789012}]"
 
-  aws ec2 modify-image-attribute --image-id ami-5731123e --launch-permission "Remove=[{UserId=123456789012}]"
+**Example 5: To modify the user data attached to an instance**
+
+The following ``modify-instance-attribute`` example shows how to add the contents of a file to the UserData for an instance. Note that the Value string must be the actual string value; you cannot use the file:// syntax to pass the file. If you use this technique, you do not have to base64 encode the file's contents first. ::
+
+    aws ec2 modify-image-attribute \
+        --instance-id=i-09b5a14dbca622e76 \
+        --user-data Value="$(UserData.txt)"
```
kyleknap: For the `Value` part of this snippet, I'm not sure if that would work. Wouldn't we have to use `cat` to get the contents of the file?

bisdavid: That example was actually in the wrong command.  I moved it to modify-INSTANCE-attribute, and corrected the syntax.  I ran it to confirm that it runs correctly.

PR #4231, opened 2019-06-11
Allow 5.1 of PyYAML

dstufft:
Closes #4230 

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,7 @@ requires-dist =
         colorama>=0.2.5,<=0.3.9
         docutils>=0.10
         rsa>=3.1.2,<=3.5.0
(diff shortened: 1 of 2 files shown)


dstufft:
Yea, this is an adaption of that PR that cleans up the logic in ``setup.py`` so that it isn't dependent on the ordering of ``requires`` anymore.

asottile:
I'm a bit miffed, not going to lie -- that could have been a comment and an easy fix up if that truly was a blocking issue instead of ignoring my PR for months and then merging essentially an identical PR.

CumpsD:
Is there any way to easily get `awscli` to install if there is `PyYAML 5.1.1` installed system wide (via Arch `python-yaml` dist package)?

PR #4248, opened 2019-06-18
Make s3 tests wait for key to exist before fetching it

stealthycoin:
Before downloading the contents of a key the integration tests will now
wait for the object_exists waiter so succeed three times.


kyleknap (approved):
Looks good to me. :ship:

PR #4255, opened 2019-06-20
Create scripts for building AWS CLI pyinstaller exe and mac pkg file

stealthycoin:
The previous Makefile under macpkg has been broken out into two scripts
scripts/build/make-pyisntaller and scripts/build/make-macpkg. Tox has
been updated to contain envs for building artifacts, a mac pkg file can
be produced by running the following tox command:

  $ tox -e macpkg

This will create the file AWS-CLI-Installer.pkg under the dist folder.


scripts/installers/make-pyinstaller
```diff
@@ -0,0 +1,122 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        '--output-dir',
+        default=os.path.join(ROOT, 'dist'),
+        help=(
+            'The directory in which to write the output artifacts. By default '
+            'this will be the "dist" directory in the root of the awscli.'
+        )
+    )
+    parser.add_argument(
```
kyleknap: Thoughts on just having a `--no-cleanup` flag? If we want to just do both a `--cleanup` and `--no-cleanup` flag we should make them mutually exclusive arguments.

scripts/installers/make-pyinstaller
```diff
@@ -0,0 +1,122 @@
...
+
+def pyinstaller(specfile):
+    aws_spec_path = os.path.join(ASSET_DIR, specfile)
+    print(run('pyinstaller %s' % (aws_spec_path), cwd=ASSET_DIR))
+
+
+def copy_dist_files(output_dir, from_name, to_name, create):
+    src = os.path.join(ASSET_DIR, 'dist', from_name)
+    dst = os.path.join(output_dir, to_name)
+    print('maybe copying %s -> %s' % (src, dst))
```
kyleknap: This is sort of an awkward print statement to have. Honestly, we should print out where we can find the end artifact.

scripts/installers/make-pyinstaller
```diff
@@ -0,0 +1,122 @@
...
+    if src != dst:
+        print('Copying %s -> %s' % (src, dst))
+        if create:
```
kyleknap: Could you elaborate on why we need to use these two different functions and need the `create` parameter? I'd assume the logic would be check to make sure the output directory exists and if it does not exist then create the directory and finally do `shutil.copytree`.

kyleknap (commented):
Looks good. Just had some comments/questions.

scripts/installers/make-macpkg
```diff
@@ -0,0 +1,109 @@
...
+EXE_NAME = 'aws'
+TEMP_PKG_NAME = 'aws-cli.pkg'
+PKG_NAME = 'AWS-CLI-Installer.pkg'
+
+
+class BadRCError(Exception):
+    pass
+
+
+def run(cmd, cwd=None):
```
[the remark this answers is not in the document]
jamesls: How would this work?  Are you suggesting a `utils` in `./scripts/installers` that we `sys.path.append()`?  Seems like it could get tricky.

scripts/installers/make-macpkg
```diff
@@ -0,0 +1,109 @@
...
+    kwargs = {
+        'shell': True,
+        'stdout': subprocess.PIPE,
+        'stderr': subprocess.PIPE,
+    }
+    if cwd is not None:
+        kwargs['cwd'] = cwd
+    p = subprocess.Popen(cmd, **kwargs)
+    stdout, stderr = p.communicate()
+    rc = p.wait()
```
jamesls: You don't need `.wait()` if you call `.communicate()`.  You can just access `p.returncode` once `communicate()` returns.

scripts/installers/make-pyinstaller
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
```
jamesls: Can we call this script `make-exe`?  I'd rather name it after what it makes rather than the tool it uses.

tox.ini
```diff
@@ -7,3 +7,21 @@ skipsdist = True
 commands =
     {toxinidir}/scripts/ci/install
     {toxinidir}/scripts/ci/run-tests
+
+
+[testenv:pyinstaller]
```
jamesls: Same thing, can we rename this to not use `pyinstaller`?

scripts/installers/make-macpkg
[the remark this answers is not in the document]
kyleknap: That is what I was thinking. In general, I can see a `run()` method be something we use in a lot of our scripts and it doesn't make sense to keep repeating it.

scripts/installers/make-exe
```diff
@@ -0,0 +1,103 @@
...
+    parser.add_argument(
+        '--no-cleanup',
+        dest='cleanup',
+        action='store_false',
+        help=(
+            'Leave the build folder produced by pyinstaller. This can be '
+            'useful for debugging.'
+        ),
+    )
+    parser.set_defaults(cleanup=True)
```
kyleknap: I believe we should be able to just add `default=True` in the `add_argument()` call here instead of using `set_defaults`

kyleknap (approved):
Looks good. Just had that small comment. Otherwise, it should be good to merge.

kyleknap (approved):

PR #4260, opened 2019-06-23
eks update-kubeconfig: use current aws executable

pdvyas:
set the aws executable in kube config to be the one currently invoked.
This makes it not depend on $PATH and select the aws-cli from the right
environment

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 
 import os
+import sys
 import logging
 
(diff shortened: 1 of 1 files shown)


tim-finnigan:
@pdvyas thanks for creating this PR and for your patience. I brought this up with the team and they had some concerns with this implementation. This could potentially be an opt-in mechanism, but we need more discussion around use cases that this would address. If you are still interested in this feature could you please [create a feature request](https://github.com/aws/aws-cli/issues/new?assignees=&labels=feature-request%2Cneeds-triage&template=feature-request.yml&title=%28short+issue+description%29) for further tracking and discussion?

PR #4284, opened 2019-06-30
Added Windows OS support using PowerShell

dgomez407:
*Description of changes:*
The alias file syntax is designed for the
bash shell on linux. Running as-is causes
the following error:

$ aws myip
'f' is not recognized as an internal or external command,
operable program or batch file.

Using PowerShell with a few character
replacements allows the aliases to work.
The bash shell was considered but it needed
commands to end with a semicolon to work reliably
and it created an external dependency.

### DIFF
--- a/awscli/alias.py
+++ b/awscli/alias.py
@@ -277,6 +277,13 @@ class ExternalAliasCommand(BaseAliasCommand):
         ]
         command_components.extend(compat_shell_quote(a) for a in args)
         command = ' '.join(command_components)
+        if os.name == 'nt':
+            command = command.replace('\\\n', '')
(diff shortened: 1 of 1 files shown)


dgomez407:
Two forced pushes were amendments to a) remove redundant lines and b) be PEP8 compliant.

tim-finnigan:
Supporting aliases in PowerShell would require a deeper investigation into approaches beyond just replacing various characters. There hasn't been any recent activity on this PR or the corresponding issue (https://github.com/aws/aws-cli/issues/4285) which closed for staleness, but I'll reopen that issue to track further discussion.

PR #4295, opened 2019-07-07
Improved readablity and simplified logic to find first quote character.

hansrajdas:
*Description of changes:*
* Updated "_find_quote_char_in_part" function, previously it was scanning input
string multiple times however proposed logic does the same thing in single
iteration.
* Code branching is also reduced, only one if block is required.


hansrajdas:
Ping...suggestions or approvals?

justindho (approved):
@hansrajdas thanks for the contribution!

PR #4304, opened 2019-07-10
New KMS Example

bisdavid:

### DIFF
--- a/awscli/examples/kms/get-key-policy.rst
+++ b/awscli/examples/kms/get-key-policy.rst
@@ -1,0 +1,20 @@
+**To copy a key policy from one CMK to another CMK**
+
+The following ``get-key-policy`` example gets the key policy from one CMK and saves it in a text file. Then, it replaces the policy of a different CMK using the text file as the policy input.
+Because the ``--policy`` parameter of ``put-key-policy`` requires a string, you must use the ``--output text`` option to return the output as a text string instead of JSON.
+Before running these commands, replace the example key IDs with valid ones from your AWS account. ::
+
+    aws kms get-key-policy \
+        --policy-name default \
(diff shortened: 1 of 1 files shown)


awscli/examples/kms/get-key-policy.rst
```diff
@@ -0,0 +1,20 @@
+**To copy a key policy from one CMK to another CMK**
+
+The following ``get-key-policy`` example gets the key policy from one CMK and saves it in a text file. Then, it replaces the policy of a different CMK using the text file as the policy input.
+Because the ``--policy`` parameter of ``put-key-policy`` requires a string, you must use the ``--output text`` option to return the output as a simple text string.
```
juneb: Please delete "simple text". It's just a string.

bisdavid: My intent was to distinguish between **--output text** and **--output json**, which is also, technically a string.

bisdavid: I'll change it to "...return the output as a simple string instead of JSON."

juneb: How about "return the output as a text string instead of JSON."

supertat55:
unscribe

2019年7月12日(金) 7:48 June Blender <notifications@github.com>:

awscli/examples/kms/get-key-policy.rst
bisdavid: Ok.  Will do.

bisdavid: I changed it as you requested - however, JSON is also a text string, so I'm not sure that's the best way to explain it.

bisdavid:
This should be good now.  Please merge.

PR #4309, opened 2019-07-12
Add check manifest

stealthycoin:
Add scripts and configuration to check that data files are correctly added to the distributions.

Edit: What do people thing about adding something like this?


jamesls:
Yeah let's try this out.  Let's get travis passing first.

PR #4312, opened 2019-07-14
Add --query example for describe-instance-patches

jacklin213:
This commit adds an example usage of using --query to retrieve information based on "InstalledTime" as it is not a valid filter key.

Can also be used for patch "Title" as that is also not a valid filter key


jacklin213:
Opened Pull request in wrong branch, closing this one and re-opening in correct one

PR #4328, opened 2019-07-18
The correct event name is api-gateway.* not apigatewayv2.*

dstufft:

### DIFF
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -31,7 +31,7 @@ logger = logging.getLogger(__name__)
 # download the content (i.e TemplateURL in cloudformation).
 PARAMFILE_DISABLED = set([
     'api-gateway.put-integration.uri',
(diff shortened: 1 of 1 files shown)


jamesls:
Interesting, while this fixes the issue it semantically seems wrong that we use `api-gateway` for `apigatewayv2` event names.  This means that we aren't able to distinguish between the two commands if we needed different event handlers for a command/param that's in both `apigateway` and `apigatewayv2`.  I wonder why that is.

dstufft:
We're using endpointPrefix for the event name and the apigatewayv2 endpointPrefix is the same as the apigateway endpointprefix.

jamesls:
Interesting.  We should eventually switch these over to service ids similar to what we did for botocore.  In the meantime, given there's no overlap between `apigateway` and `apigatewayv2` I think we should go with this change.

jamesls (approved):

PR #4352, opened 2019-07-26
Delete extra whitespace

yoshiken:
A correction that does not affect the operation.

fix `W293 blank line contains whitespace` in pycodestyle


justindho (approved):
Thanks, @yoshiken!

PR #4353, opened 2019-07-26
Delete unused variable

yoshiken:
Removed unused variables. 

This will lead to a slight performance improvement.


justindho (approved):
Thanks, @yoshiken!

Issue #4350, opened 2019-07-25
pyyaml<=5.1 runs into issues with serial install on Python 3.6


aviveve:
I had an installed version of AWS CLI. But the pip install of serial failed since AWS CLI had a requirement that pyyaml <= 5.1. So I uninstalled awscli and reinstalled serial. It installed but installed PyYAML of 5.1.1.
Re-install of awscli now fails because of restrction yaml version.

Collecting serial
  Using cached https://files.pythonhosted.org/packages/1f/51/6a260c498162c37d0759f3759b7647a10d8d30caba1cfc9aa4b5b1f0d08b/serial-0.0.97-py2.py3-none-any.whl
Requirement already satisfied: iso8601>=0.1.12 in c:\programdata\anaconda3\lib\site-packages (from serial) (0.1.12)
Requirement already satisfied: future>=0.17.1 in c:\programdata\anaconda3\lib\site-packages (from serial) (0.17.1)
Requirement already satisfied: pyyaml>=3.13 in c:\programdata\anaconda3\lib\site-packages (from serial) (5.1.1)
Installing collected packages: serial
Successfully installed serial-0.0.97

C:\WINDOWS\system32>pip install awscli
Collecting awscli
  Using cached https://files.pythonhosted.org/packages/e3/56/398e5e1c1484b5e25da6eafbeca8ec2a69e25942485fe5c0aa6d3d173310/awscli-1.16.206-py2.py3-none-any.whl
Requirement already satisfied: botocore==1.12.196 in c:\programdata\anaconda3\lib\site-packages (from awscli) (1.12.196)
Requirement already satisfied: colorama<=0.3.9,>=0.2.5 in c:\programdata\anaconda3\lib\site-packages (from awscli) (0.3.7)
Requirement already satisfied: rsa<=3.5.0,>=3.1.2 in c:\programdata\anaconda3\lib\site-packages (from awscli) (3.4.2)
Collecting PyYAML<=5.1,>=3.10; python_version != "2.6" (from awscli)
  Downloading https://files.pythonhosted.org/packages/72/da/386634cb17f33b12278af017ccb3314cede4806af70285f396b8f6af1203/PyYAML-5.1-cp36-cp36m-win_amd64.whl (214kB)
     |████████████████████████████████| 215kB 1.6MB/s
Requirement already satisfied: docutils<0.15,>=0.10 in c:\programdata\anaconda3\lib\site-packages (from awscli) (0.13.1)
[...]

Issue #4243, opened 2019-06-15
Old version of wheel fails to build PyYAML on Mac Python 3.7


xerebz:
If I follow the instructions for building (pip install -r requirements.txt; pip install -e .), I get the following error on Mac using Python 3.7.3. The "assert tag == supported_tags[0]" is miscalculating the ABI (cpdarwin instead of cp-37m-darwin) which was fixed in the next version of wheel (0.25.0). I see that you're still trying to support Python 2.6 so the requirements.txt should be set to something like wheel>=0.25.0,<0.30.0 

```
pip3 install --force-reinstall --no-cache-dir pyyaml

Collecting pyyaml
  Downloading https://files.pythonhosted.org/packages/a3/65/837fefac7475963d1eccf4aa684c23b95aa6c1d033a2c5965ccb11e22623/PyYAML-5.1.1.tar.gz (274kB)
     |████████████████████████████████| 276kB 1.2MB/s
Building wheels for collected packages: pyyaml
  Building wheel for pyyaml (setup.py) ... error
  ERROR: Complete output from command /Users/alaileon/.local/share/virtualenvs/aws-cli-51GcqfER/bin/python3.7 -u -c 'import setuptools, tokenize;__file__='"'"'/private/var/folders/72/cxwx5fy11yg630rtc9tpwsqmw9lqlm/T/pip-install-swpywuf4/pyyaml/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/72/cxwx5fy11yg630rtc9tpwsqmw9lqlm/T/pip-wheel-ixq5fth5 --python-tag cp37:
  ERROR: running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.macosx-10.13-x86_64-3.7
  creating build/lib.macosx-10.13-x86_64-3.7/yaml
  copying lib3/yaml/scanner.py -> build/lib.macosx-10.13-x86_64-3.7/yaml
  copying lib3/yaml/error.py -> build/lib.macosx-10.13-x86_64-3.7/yaml
  copying lib3/yaml/constructor.py -> build/lib.macosx-10.13-x86_64-3.7/yaml
  copying lib3/yaml/composer.py -> build/lib.macosx-10.13-x86_64-3.7/yaml
[...]

Issue #4042, opened 2019-04-03
Please port to a newer version of PyYAML


glaubitz:
The current upstream release of PyYAML is 5.1 while aws-cli still depends on PyYAML <= 3.13.

Would be nice if aws-cli could be ported to the latest version of PyYAML that version dependency could be raised to 5.1 or higher.

kabbedijk:
The PR for this is already approved and waiting for merge.
https://github.com/aws/aws-cli/pull/4015#issuecomment-480319733

It is taking quite some time though :)

glaubitz:
Thanks. I'll look into cherry-picking the patch for openSUSE later today then.

justnance:
@glaubitz - Thanks for your post.   I have updated our internal collaboration on this issue.  We are tracking this work under #3660 and #2290. To avoid duplicate efforts I'm closing this issue.

asottile:
@justnance neither of those issues are relevant to allowing a newer pyyaml, could you please reopen this and/or review the patch in #4015?

SeppPenner:
The currently used PyYAML@3.13 contains a Arbitrary Code Execution issue. Check https://app.snyk.io/vuln/SNYK-PYTHON-PYYAML-42159 and https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18342 for more information. (Just as reference for the duplicate issue I added here: #4193).

rigogsilva:
UP, UP, UP. please!!!

justnance:
@asottile - Thanks for referencing #4015 and commenting in merged PR #4231.  I'm looking into confirming this issue is resolve before closing.  thanks.

asottile:
For 5.1 this is done, but now there's a 5.1.1

After the last snub in my previous PR I'm not really inclined to try and fix this one

justnance:
@asottile - Thanks again for the feedback.  I've confirmed the behavior with 5.1.1 under #4350 and labeled it has enhancement.  We are working on getting fixed under PR #4355. 

related to: #4350 #4243

joguSD:
The version range has been updated, closing this out.

PR #4355, opened 2019-07-28
Upgrading pyyaml<=5.1 to pyyaml<=5.1.1

justnance:
*Issue #, if available:*
fixes: #4350 #4042 #4243

*Description of changes:*
[PyYAML 5.1.1 released on June 6, 2019](https://pypi.org/project/PyYAML/#history) is not compatible with the latest version of the CLI.   

If you try to install PyYAML and the CLI is already installed, the following error returns: "PyYAML 5.1.1 which is incompatible."  This is due to PIP trying to install the latest PyYAML version (5.1.1).  Command used to install PyYAML is: 

pip3 install --force-reinstall --no-cache-dir pyyaml

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,6 +9,6 @@ requires-dist =
         docutils>=0.10,<0.15
         rsa>=3.1.2,<=3.5.0
         PyYAML>=3.10,<=3.13; python_version=="2.6"
-        PyYAML>=3.10,<=5.1;python_version!="2.6"
+        PyYAML>=3.10,<=5.1.1;python_version!="2.6"
         s3transfer>=0.2.0,<0.3.0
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -9,6 +9,6 @@ requires-dist =
         docutils>=0.10,<0.15
         rsa>=3.1.2,<=3.5.0
         PyYAML>=3.10,<=3.13; python_version=="2.6"
-        PyYAML>=3.10,<=5.1;python_version!="2.6"
+        PyYAML>=3.10,<=5.1.1;python_version!="2.6"
```
Surgo: I think `<5.2` is better. Maybe bump major version when backward incompatible changes.

PR #4365, opened 2019-07-31
Support caching web identity credentials

joguSD:
This updates the assume role file caching handler to inject the file cache onto both the assume role and assume role with web identity provider.


2ps:
Could you please revert this?  It causes issue #4371

Issue #2063, opened 2016-07-11
Add kms decrypt argument for base64 encoded input


theazureshadow:
The output for `aws kms encrypt` is a base64-encoded string. The input for `aws kms decrypt` is a binary string, which is not particularly bash-friendly. See #1100.

It would be useful if there was an additional `--ciphertext-base64` argument that could take that same base64 blob from `encrypt` and decrypt it correctly.

kyleknap:
Marking as a feature request.

Are you blocked on this? Could you chain the commands with ``base64`` to handle your use case? Like I do in this [blog post](https://blogs.aws.amazon.com/cli/post/TxLWWN1O25V1HE/Best-Practices-for-Local-File-Parameters)

theazureshadow:
I'm not blocked, but it is really non-obvious what's going wrong. We lost a bit of time when trying to validate that we'd encoded our data correctly. Seems like an easy win to be able to specify base64-encoded ciphertext.

w32-blaster:
I agree with this. There are use cases where it's not trivial to chain commands to get the correct result. It would help in many cases.

ajlanghorn:
👍 to this.

iadknet:
👍

victorparmar:
:+1:

rbm1337:
👍

JimmyDahlqvist:
+1

kid7st:
+1

sinogermany:
+1

techypaul:
+1, in the meantime for anyone else looking:

```
echo AQICAHhb....ZClShBC8dZpCg== | base64 --decode > /tmp/temp.enc
DB_PASS=aws kms decrypt --region eu-west-1 --ciphertext-blob fileb:///tmp/temp.enc --query Plaintext --output text | base64 --decode
sed -i -e "s/<<DB_PASS>>/${DB_PASS}/g" /var/www/mysite/DbSettings.php
```

gonzaloserrano:
+1

jordimirobruix:
+1

coboshm:
+1

ryanskinner:
+1

mlamothe-broadsoft:
+1

chi42:
+1

nikhithn:
+1

scf37:
[...]

PR #4366, opened 2019-07-31
Support base64-encoded parameters for binary types

joe-wright:
Add a "base64://" protocol for passing binary parameters, such as KMS
Decrypt's "CiphertextBlob", as a base64 string. Currently the only way
to specify a binary blob parameter is by writing the parameter value to
a file and providing the filepath prefixed with "fileb://".

*Issue #, if available:* https://github.com/aws/aws-cli/issues/2063

*Description of changes:* This allows the user to provide a binary blob parameter by prefixing its base64 representation with `base64://`. 

For example:
```
aws kms decrypt --ciphertext-blob base64://AQECAHitN+klEybSwklcQTh51VcQ6NVha5fJMHrZ5CwkWpD35AAAAGMwYQYJKoZIhvcNAQcGoFQwUgIBADBNBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDANWxamzybey0ZE8bQIBEIAgCqJfaUHxCYbRZ3Q53VH0wrLr1bD8bs7HYuytgIP62Y8=
```

Currently, the only way to provide a binary parameter is by saving it to disk and using `fileb://`:
```
echo -n 'AQECAHitN+klEybSwklcQTh51VcQ6NVha5fJMHrZ5CwkWpD35AAAAGMwYQYJKoZIhvcNAQcGoFQwUgIBADBNBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDANWxamzybey0ZE8bQIBEIAgCqJfaUHxCYbRZ3Q53VH0wrLr1bD8bs7HYuytgIP62Y8=' | base64 -d > ciphertext.blob
aws kms decrypt --ciphertext-blob fileb://ciphertext.blob
```

### DIFF
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -13,6 +13,7 @@
 import logging
 import os
 import copy
+import base64
 
 from botocore.awsrequest import AWSRequest
 from botocore.httpsession import URLLib3Session
@@ -232,6 +233,14 @@ def get_file(prefix, path, mode):
             path, e))
 
 
+def get_base64(prefix, value):
+    try:
+        return base64.b64decode(value[len(prefix):])
+    except Exception as e:
+        raise ResourceLoadingError('Unable to decode %s: %s' % (
+            value, e))
+
+
 def get_uri(prefix, uri):
     try:
         session = URLLib3Session()
@@ -247,6 +256,7 @@ def get_uri(prefix, uri):
 
 
 LOCAL_PREFIX_MAP = {
+    'base64://': (get_base64, {}),
     'file://': (get_file, {'mode': 'r'}),
     'fileb://': (get_file, {'mode': 'rb'}),
 }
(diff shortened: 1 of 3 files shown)


tim-finnigan:
Thank you for creating this PR. As mentioned [here](https://github.com/aws/aws-cli/issues/2063#issuecomment-565233049) in the corresponding issue: 

> In CLI v2 we're going with a slightly different approach: https://github.com/aws/aws-cli/pull/4748/
> 
> This will add a new parameter / config `cli_binary_format` which can currently either be set as `legacy` or `base64`. In `base64` mode blobs passed to the CLI are expected to be a base64 encoded string (no need for a `base64://`). Having an external configuration for this will allow us to add additional formats as we see fit for the various use cases that might arise for other/future services.

We recommend using AWS CLI v2 in which binary parameters are passed as base64-encoded strings [by default](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-binaryparam), and you can use [cli_binary_format](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-config-cli_binary_format) to configure how binary input parameters are interpreted.

PR #4369, opened 2019-08-01
Add Mac PKG installer uninstall and test script

stealthycoin:
Uninstalling a Mac PKG is not straightforward, so a utility script has
been added that can uninstall the AWS CLI. The installer tester script
has been updated with a pkg format option. This requires sudo access for
now to both install and uninstall the package.

Another tangential change to the postinstall script was made to keep
a manifest of extra files created by the installation process. This
file is checked at uninstall time and used to clean up any files not
registered with the PKG receipt (the receipt cannot be updated manually
by a postinstall script, so this must be a separate metadata file).
Any new files added in the future should be added to this manifest as
well.

### DIFF
--- a/scripts/installers/uninstall-mac-pkg
+++ b/scripts/installers/uninstall-mac-pkg
@@ -3,6 +3,7 @@
 import argparse
 import sys
 import os
+import re
 from datetime import datetime
 from subprocess import check_output
 from subprocess import CalledProcessError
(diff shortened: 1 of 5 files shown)


scripts/installers/uninstall-mac-pkg
```diff
@@ -0,0 +1,144 @@
...
+    _erase_files(list_files)
+    list_dirs = _get_dir_list(root_dir)
+    _erase_dirs(list_dirs)
+    _forget()
+    return 0
+
+
+def _get_root_dir():
+    lines = run('pkgutil --pkg-info %s /' % _PKG_ID, echo=False).split('\n')
+    volume_line = lines[2]
```
kyleknap: It would be nice to make this a regrex instead to pull out all of this information.

scripts/installers/uninstall-mac-pkg
```diff
@@ -0,0 +1,144 @@
...
+
+from utils import run
+from utils import BadRCError
+
+
+_PKG_ID = 'com.amazon.aws.cli2'
+
+
+def uninstall():
+    assert _is_installed() is True, 'Could not find AWS CLI installation.'
```
kyleknap: This should work by doing just `assert _is_installed()`.

scripts/installers/uninstall-mac-pkg
```diff
@@ -0,0 +1,144 @@
...
+def _get_file_list(root):
+    lines = run(
+        'pkgutil --only-files --files %s /' % _PKG_ID, echo=False).split('\n')
+    pkg_file_list = [os.path.join(root, line) for line in lines]
+    extra_files = _read_install_metadata(root)
+    return pkg_file_list + extra_files
+
+
+def _read_install_metadata(root):
+    metadata_path = os.path.join(root, '.install-metadata')
```
kyleknap: We should explain why we need to do some of these steps. It seems like there were a lot of gotchas in figuring this out and it would be good to capture why we need to do it.

PR #4381, opened 2019-08-06
Correct file paths for IOT Examples

bisdavid:
I put these examples in an incorrectly named folder, so none of them show up.  They've all been moved to the correct folders.


jamesls:
Interesting, we should probably add some tests to catch this in the future.  Ideally we'd fail the build so we'd know that there were issues before merging.

PR #4397, opened 2019-08-09
Support ':' char in shorthand syntax key names

### LINKED ISSUE #4348: codecommit: tag-resource does not permit `:` character in tag key
The following command incorrectly returns an error even though colon characters (`:`) in tag keys are legal:

```
aws codecommit tag-resource --resource-arn <arn> --tags namespace:KeyName=value
```

Expected behavior: a new tag is applied to the resource with the name `namespace:KeyName` and value `value`.

Actual behavior:

```
Error parsing parameter '--tags': Expected: '=', received: ':' for input:
namespace:KeyName=value
```

jamesls:
Fixes #4348.

### DIFF
--- a/.changes/next-release/enhancement-Shorthand-33683.json
+++ b/.changes/next-release/enhancement-Shorthand-33683.json
@@ -1,0 +1,5 @@
+{
+  "category": "Shorthand", 
+  "type": "enhancement", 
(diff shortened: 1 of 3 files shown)


kyleknap (approved):
:ship:

Issue #4240, opened 2019-06-14
aws s3 ls allows listing by file prefix, but aws s3 sync does not


benjamin-maynard:
Hi,

I have an S3 Bucket that has a range of files like this:

```
0000-abcdef-ghiklm-nopqrs
1000-abcdef-ghiklm-nopqrs
2000-abcdef-ghiklm-nopqrs
3000-abcdef-ghiklm-nopqrs
a000-abcdef-ghiklm-nopqrs
b000-abcdef-ghiklm-nopqrs
c000-abcdef-ghiklm-nopqrs
d000-abcdef-ghiklm-nopqrs
```

Each prefix has millions of files in the bucket.

If I want to list files by a prefix, I can do this really efficiently, like so:

* `aws s3 ls s3://my-bucket/1` - Will only list files beginning with `1`
* `aws s3 ls s3://my-bucket/2` - Will only list files beginning with `2`

This is efficient, as it doesn't return any files without the appropriate prefix. The API handles the list, only returning files that start with `1` or `2`

---

The same isn't possible for the `aws s3 sync` command.

`aws s3 sync s3://my-bucket/1* s3://my-new-bucket/` or `aws s3 sync s3://my-bucket/1 s3://my-new-bucket/` does not work.

The only way to do it is through adding a `--exclude` `--include` parameter for example `aws s3 sync s3://my-bucket/ s3://my-new-bucket/ --exclude "*" --include "1"`

But this is really inefficient, as it lists the whole contents of the bucket, and filters each item to determine whether it should match **client side**. This method makes perfect sense for normal include exclude operations like `--exclude "*" --include "*.txt"` but for prefixed files it is inefficient.

I know that prefixes should be used, but that isn't always possible. It is clear that the `aws s3 ls` command has the potential to return only files with a specific prefix, so could this behaviour be added into the sync command? Otherwise syncs between buckets take weeks instead of hours.

benjamin-maynard:
@JordonPhillips Any views on this?

inopinatus:
I have this problem also, with a CloudFront log bucket containing hundreds of thousands of files, and --include and --exclude are so slow as to be unusable.

kyleknap:
Marking it as a feature request. To support that style of syncing we would need to add a flag to sync. Otherwise, it would also make sense to make the `--include` and `--exclude` flags more efficient in terms of determining what we actually list in the bucket.

benjamin-maynard:
[...]

Issue #1454, opened 2015-08-17
aws s3 cp doesn't work with prefix


fabiocesari:
From a bucket containing millions of files, I want do download a few thousand based on their prefix. 

Note: using --exclude "*" --include "myprefix" works, but is impractical since it lists all the files and applies the filter afterwards.

The problem is aws s3 cp adds a trailing "/" after the prefix I specify. In the example below you can see that logs/2015-08-17 becomes logs/2015-08-17/ 
This additional "/" leads to an empty result set, since the keys in the bucket have the form logs/YYYY-MM-DD-HH-MM-SS-RANDOM).

> aws s3 cp s3://mybucket/logs/2015-08-17 . --recursive --dryrun --debug

2015-08-17 21:48:32,026 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/1.7.27 Python/2.7.6 Linux/3.14.13-c9, botocore version: 0.108.0
2015-08-17 21:48:32,026 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['s3', 'cp', 's3://mybucket/logs/2015-08-17', '.', '--recursive', '--dryrun', '--debug']
....
2015-08-17 21:48:32,088 - MainThread - botocore.endpoint - DEBUG - Making request for <botocore.model.OperationModel object at 0x7ff9d3ec3e50> (verify_ssl=True) with params: {'query_string': {u'prefix': u'logs/2015-08-17/', u'encoding-type': 'url'}, 'headers': {}, 'url_path': u'/mybucket', 'body': '', 'method': u'GET'}

rayluo:
The current CLI "aws s3 cp" behavior (coincidentally?) complies with the "cp" behavior. We will need to think whether or how to support the use case you mentioned.

geota:
+1 This issue bit me today.

What is the reason/use case for adding a trailing slash? 

I would like to filter on a partial prefix that does not correspond to a "directory". I was able to work around this by using bash to filter the files and piping it back to aws cp.

kyleknap:
@fabiocesari @geota 

We do read the exclude/include rules literally, and do not add the '/' to the end. To do what you want to do, you need to add a ``*`` to the end of the partial prefix, which means match all characters after. So a combination like ``--exclude '*' --include 'myprefix*'`` should work.

Let me know if you have any more questions.

fabiocesari:
[...]

PR #4423, opened 2019-08-20
Allow sync/cp/rm/mv S3 objects based on a prefix instead of using include/exclude filters

mourya92:
*Issue #1454 #4240 :*

**Description**
Use case : Download S3 server access logs of a certain day from the logging bucket. 

Example command : 
`aws s3 cp s3://example-bucketname/2019-07-22 . --recursive`

Today, the above command tries to download the log files under the "2019-07-22/" path(CLI adds / in the LIST API call). 

But, in fact there won't be such path(folder) in S3 bucket. The idea is to download all the S3 objects that match the prefix : 2019-07-22.

So, the solution provided today to users is to use include/exclude filters or use AWS Athena. This takes a lot of time if the bucket has billion log files. 

However, aws s3 ls behaves in a right way i.e., it shows the objects that match the prefix but, doesn't add a '/' in the LIST API call. 

So, this commit will fix this issue. 

If anyone wants to cp/sync/mv/rm an actual 'folder' from S3, they can simply specify a trailing slash '/' in the command. Even if they specify, CLI will still delete all objects that matches the prefix  and in 99% of the cases, the prefix will be same as the folder name. 

This issue has been faced by several users and removing the '/' will help them a lot. 

### DIFF
--- a/awscli/customizations/s3/fileformat.py
+++ b/awscli/customizations/s3/fileformat.py
@@ -47,18 +47,18 @@ class FileFormat(object):
         #     directory/objects under a common prefix or false when it
         #     is a single file
         dir_op = parameters['dir_op']
-        src_path = format_table[src_type](src_path, dir_op)[0]
+        src_path = format_table[src_type]('src', src_path, dir_op)[0]
         # :var use_src_name: True when the destination file/object will take on
         #     the name of the source file/object.  False when it
         #     will take on the name the user specified in the
         #     command line.
-        dest_path, use_src_name = format_table[dest_type](dest_path, dir_op)
+        dest_path, use_src_name = format_table[dest_type]('dest', dest_path, dir_op)
         files = {'src': {'path': src_path, 'type': src_type},
                  'dest': {'path': dest_path, 'type': dest_type},
                  'dir_op': dir_op, 'use_src_name': use_src_name}
         return files
 
-    def local_format(self, path, dir_op):
+    def local_format(self, typ, path, dir_op):
         """
         This function formats the path of local files and returns whether the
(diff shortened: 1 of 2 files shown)


ztane:
I do not think this is correct though. I'd prefer *another word* instead of the `--recursive` for that behaviour, say `--prefix`, because `--recursive` has been associated with certain behaviour.

p2pb:
This would be an highly valuable improvement (and arguably reflects the "true" (simple strings rather than nested structures) nature of S3 object keys).

jeremyplichtafc:
👍  please merge this in. I end up using this branch when I have to accomplish this and would rather have this be part of aws-cli master itself.

timtrinidad:
@ztane This is the behavior of `--recursive` when passed to `aws s3 ls`, so I could see the argument going either way.

ztane:
@timtrinidad yes but thousands of programs and assumptions would be broken. FWIW I want this fearure, but only by different name - prefix, because that's what it is

malthe:
This should also apply the other way, copy local files recursively to a bucket path and _prefix_.

indrora:
My oh my, this has had some time to stew. 

So far this seems to have a fairly positive response from the community -- which is important -- but what's unfortunately missing is feedback from the team. 

Additionally, there's no tests associated with this behavior to make sure it performs as expected. There's little documentation on what has changed and how, and that means I'm uncomfortable signing off on this. 

@ztane does bring up a good point that this does change an established behavior. Whether we use the established `--prefix` notation or create some new syntax here is up for debate. 

Until then, I'm going to close this PR. Open a feature request issue that addresses the two issues listed above and discuss the proposed solution there.

PR #4430, opened 2019-08-22
Case insensitive protocol

lathama:
Currently there is an ugly error if S3:// is used.

*Issue #, if available:*

*Description of changes:*
Case insensitive check on protocol


lathama:
Travis appears to be having issues.

PR #4433, opened 2019-08-22
Case insensitive check on protocol

lathama:
Currently there is an ugly error if S3:// is used.

*Issue #, if available:*

*Description of changes:*
String to lower() on protocol check


lathama:
@stealthycoin is there anything I need to do here?

stealthycoin:
@sanathkr thoughts?

lathama:
What is a reasonable time to wait on things like this? I think I did the right thing and fear I missed some important step. Please advise

lathama:
noop

Issue #4437, opened 2019-08-24
pinned colorama versions causing conflicts with other tools


ssbarnea:
It seems that the decision of pinning colorama is a never ending source of conflicts. Other tools or libraries do require newer versions because they want to avoid fixed bugs but aws-cli is overly conservative with its range. This cause multiple tickets and PR for updating the range which should have being limited to only major version not like we have not at patch level.
```
awscli 1.16.221 has requirement colorama<=0.3.9,>=0.2.5, but you have colorama 0.4.1.
```

Because colorama is a mature library that does not receive lots of development, I would go so far to say that it would be fine to even remove the ceiling completely.

kyleknap:
We are currently looking into updating the ceiling on this dependency. We have an existing GitHub issue tracking this: https://github.com/aws/aws-cli/issues/3724. I'd prefer to track it in the preexisting GitHub issue and we'll also take a look at the PR you opened. Closing this issue in favor of the referenced issue.

PR #4438, opened 2019-08-24
Correct upper bound on colorama dependency

ssbarnea:
Replace ever-outdated upper bound on colorama which caused conflicts
with other tools or libraries.

Newer one is based on semantic versioning is will make less likely to
cause conflicts, or bugs due to incompatibilities.

Fixes: #4437


stealthycoin:
Closing out of date PR.

Issue #4274, opened 2019-06-27
[v2] Please drop Python 2 support


methane:
Since awscli v2 can be packaged with Python 3, why Python 2 support is needed?

Python 2.7 will become EOL at 2020-01-01.  It's a very important milestone for library maintainers to drop Python 2 support.
But if Python 2 is still supported by awscli, it puts pressure to keep supporting Python 2 to library maintainers.

Please note that it is not a problem only for AWS, but also all maintainers who maintain libraries aws-cli depends on.

jamesls:
Hi, sorry for the delay here.

> Since awscli v2 can be packaged with Python 3, why Python 2 support is needed?

I think it's mostly an issue of timing.  We will eventually drop python 2 support, I think it's just a matter of how.  Given that the officially supported installation mechanisms for AWS CLI v2 will vendor python3, and pip will no longer be an officially recommended way of installing CLI v2, in theory the number of python2 users should drop as customers migrate over to CLI v2.  At the same time, we realize not everyone will want to use our supported installation mechanisms, and may prefer alternative mechanisms e.g. various linux distro packages, homebrew, docker, snap, etc., so we'd like to see how this plays out.

Once enough customers are no longer using python 2 with the AWS CLI, we can drop support for it.
There will also be a period of time where we'll have to support both AWS CLI v1 and v2, so I would imagine there may be some backporting between both versions.

Additionally, if pip is no longer being used to install the AWS CLI, then any of our dependencies would also see a corresponding drop in their pip download numbers as people migrate.  I've seen a number of discussions on other projects where a key metric in deciding python version support is number of installs via pip's metrics.  So it seems like not using pip for CLI v2 would effectively have the same desired result for package maintainers.

So AWS CLI v2 shouldn't exacerbate the python 2 issue.  And we won't be able to drop python 2 support in AWS CLI v1 until customers are no longer using python2 which we hope would happen as people migrate over to CLI v2.  Also, our dependencies have dropped support for a python version before we do, which makes an even stronger case for us to drop support for a python version.  For example, urllib3 no longer supports python2.6 even though we still support it (but I'd really really like to drop 2.6 support).

One last thing to keep in mind is none of this is set in stone.  Based on feedback, this all can change.  We haven't really had a discussion on py2/py3 support yet, so we'd love to hear what everyone thinks.

ssbarnea:
[...]

PR #4404, opened 2019-08-14
Improve testing

Jamim:
Hello,

I believe that it would be nice to have tests for modern Python versions and for PyPy.

These changes also:
 - declare support for Python 3.8 at `setup.py`
 - remove [deprecated](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) `sudo` option from `.travis.yml`
 - fix compatibility with Python 3.8
 - fix compatibility with PyPy
 - fix `SyntaxWarning` warnings

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

Regards!

PR #4439, opened 2019-08-24
Drop support for EOL python versions

ssbarnea:
This removes code related versions of python that reached EOL long time ago: python2.6, python3.3 and python3.4

Issue: #4274

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,7 @@
 language: python
 matrix:
   include:
-    - python: 2.6
-      dist: trusty
-      sudo: false
     - python: 2.7
-      dist: trusty
-      sudo: false
-    - python: 3.3
-      dist: trusty
-      sudo: false
-    - python: 3.4
       dist: trusty
       sudo: false
     - python: 3.5
--- a/README.rst
+++ b/README.rst
@@ -15,10 +15,7 @@ This package provides a unified command line interface to Amazon Web Services.
 
 The aws-cli package works on Python versions:
 
-* 2.6.5 and greater
 * 2.7.x and greater
-* 3.3.x and greater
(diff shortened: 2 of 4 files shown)


ssbarnea:
Oh, joy a badly configured codecov is doing mode damage than benefits.

jamesls:
Hi just wanted to follow up on this.  We've published our deprecation plan for 2.6 and 3.3 and plan to remove support Jan 10, 2020.  We wanted to give customers time to plan for this: https://aws.amazon.com/blogs/developer/deprecation-of-python-2-6-and-python-3-3-in-botocore-boto3-and-the-aws-cli/

ssbarnea:
Something is now right regarding planning because by the date you mentioned not only that py27 will also be obsoleted and unmaintained. I think that there is a chance that botocore will "support" only unsupported versions of python as at this moment most them are already dropped.

The reality is that botocore is even missing testing on py37 and py38, where is the planning on these? I see plans to drop support for versions what were already dropped by others but no real activity on even testing newer versions of python. Somehow I go the impression that aws-cli is stuck in a time loop that is few years behind everyone else.

I find that blog post misleading, I am curious who are these "many industry Python projects and packages continued to support..."? I know for sure that many such projects plan to drop support for py on 2020-01-01, few more aggressive ones already did that. My impression is that everyone else is moving at much greater speed than aws-cli repo

Jamim:
Hello @ssbarnea,

> The reality is that botocore is even missing testing on py37 and py38, where is the planning on these? I see plans to drop support for versions what were already dropped by others but no real activity on even testing newer versions of python.

Here it is: #4404

But sadly no one cares about that PR, so I completely agree with your opinion.

PR #4470, opened 2019-09-05
Add colorama tests

stealthycoin:


tests/functional/test_colorama.py
```diff
@@ -0,0 +1,222 @@
...
+
+import colorama
+from colorama import Fore
+from colorama import Back
+from colorama.winterm import WinTerm, WinStyle, WinColor
+
+
+@skip_if_windows('Posix color code tests')
+class TestPosix(unittest.TestCase):
+    _COLORAMA_KWARGS = {
```
jamesls: What do you think about having a constant in table.py for these colorama kwargs?  Seems like this can get out of sync if we change the kwargs we use in the table formatter and we may no longer be testing how we're using colorama.

tests/functional/test_colorama.py
```diff
@@ -0,0 +1,222 @@
...
+            content = Fore.BLUE + 'foo'
+            sys.stdout.write(content)
+            self.assertEqual(captured.stdout.getvalue(),
+                             '%sfoo%s' % (self._ANSI_FORE_BLUE,
+                                          self._ANSI_RESET_ALL))
+
+
+
+@unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")
+class WinTermTest(unittest.TestCase):
```
jamesls: I don't think these windows tests are particularly useful for us.  This seems to be testing the internals of how their winterm is implemented, and we don't directly use any of those classes in the CLI itself.

Assuming the `TestPosix` behavior also applies to windows (if you pipe the output you still get colorized output), I'd prefer to see similar high level tests for windows.  E.g. "test_colorama_does_not_strip_non_tty" etc., even if we have to restrict these tests to only run if we're on powershell (or cmd or whatever).

PR #4482, opened 2019-09-10
Move cliinputjson tests from integration to functional

stealthycoin:
Move cliinputjson tests from integration to functional


kyleknap (approved):
:ship:

PR #4510, opened 2019-09-18
Add support for organizations in validate-logs

kyleknap:
To validate logs for an organizational trail, an account id is needed and cannot be parsed from the arn. To account for this, we are adding an `--account-id` parameter and also search in the bucket differently if we know the trail publishing to an s3 bucket is an organizational one.

### DIFF
--- a/awscli/customizations/cloudtrail/validation.py
+++ b/awscli/customizations/cloudtrail/validation.py
@@ -29,6 +29,7 @@ from awscli.customizations.cloudtrail.utils import get_trail_by_arn, \
     get_account_id_from_arn
 from awscli.customizations.commands import BasicCommand
 from botocore.exceptions import ClientError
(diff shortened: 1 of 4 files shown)


awscli/customizations/cloudtrail/validation.py
```diff
@@ -100,6 +104,9 @@ def create_digest_traverser(cloudtrail_client, s3_client_provider, trail_arn,
     :param prefix: bucket: Key prefix prepended to each digest and log placed
         in the Amazon S3 bucket if it is different than the prefix that is
         currently associated with the trail.
+    :param prefix: account_id: The account id for which the digest files are
```
stealthycoin: `:param account_id:`

stealthycoin (approved):

PR #4529, opened 2019-09-25
New CLI Examples for AWS Signer

bisdavid:


awscli/examples/signer/list-signing-profiles.rst
```diff
@@ -0,0 +1,29 @@
+**To list all signing profiles**
+
+The following ``list-signing-profiles`` example displays details about all signing profiles for the account. ::
+
+    aws signer list-signing-profiles
+
+This command produces no output.
```
kyleknap: I think we can remove this sentence given it does produce output.

bisdavid: Whoops.  This is now fixed.  Thanks for catching!

PR #4552, opened 2019-10-02
Fix typo in delete-access-key.rst example

rymndhng:


stealthycoin:
Good catch, thanks!

PR #4603, opened 2019-10-22
[v2] Merge branch 'develop' into v2-develop-resync

jamesls:
This merges the latest changes in the develop branch
back into the v2 branch.

At this point, we can call the v2 branch stable and stop
rebasing/force pushing to v2.  All subsequent sync from
develop will require merging into v2.  This also lets us
review the changes and the merge conflict resolutions.

```
 Conflicts:
        MANIFEST.in
        awscli/__init__.py
        awscli/customizations/cloudformation/yamlhelper.py
        awscli/customizations/eks/update_kubeconfig.py
        awscli/handlers.py
        awscli/paramfile.py
        doc/source/conf.py
        requirements.txt
        scripts/make-bundle
        setup.cfg
        setup.py
```

NOTE: The diff is too big to show everything, but the actual conflict resolutions were fairly small.  I'd recommend going to the "FIles changed" tab and clicking on the files listed above if you want to review the files with conflicts.


jamesls:
Had to push a fix that removes a test.  We previously were supporting the `--start` param, but I don't think we should support this anymore.  Here's the commit message for reference:

```
Remove tests for start param in route53domains

History of this change:

* The original args are "start" and "end"
* We couldn't use "end" because of the "--endpoint-url" shadowing.
* We renamed these args to "--start-time" and "--end-time"

What broke is that last year we added pagination for this operation,
which adds a "--starting-token", which means that this arg now shadows
the "--start" arg.

Given this is v2, I think it's cleaner to just require that this value
must either be "--start-time" and "--end-time" and remove support for
"--start".
```

kyleknap (approved):
:ship:

PR #4604, opened 2019-10-22
[v2] Add support for building windows executables

jamesls:
* Upgrade to pyinstaller 3.5.  There was a bug in 3.4 that didn't
  build properly on windows when using a virtualenv.
* Explicitly use python when invoking scripts.  We can't rely on
  the shebang on windows.

Other than that everything else just works.

*Issue #, if available:*

*Description of changes:*


kyleknap (approved):
:ship:

stealthycoin (approved):

PR #4605, opened 2019-10-23
[v2] Don't pull from s3transfer/botocore from develop

jamesls:
Don't pull s3transfer/botocore from develop

Right now we specify the version from both s3transfer as well
as from the requirements.txt file.  Rather than pull from develop
we should just set the version from the setup.py.

I also synced the version of s3transfer to match the latest
minor version range.


kyleknap (approved):
:ship:

PR #4611, opened 2019-10-24
Update eks get-token to use client interface

kyleknap:
It ensures that we are not using any internals of botocore by using public interfaces. Also since we are using the client now, much of this was able to be refactored to functional testing.


joguSD (approved):
Assuming travis passes.

PR #4615, opened 2019-10-29
Rename executable to aws2

kyleknap:
Makes it easier to install/run side-by-side with V1 of the CLI. I also included a change to the uninstall script as the `pkg` utility can print out more than one whitespaces when printing out the information of the installed pkg and this would mess up the path to the executable when you run the `check` command.


macpkg/scripts/postinstall
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+EXE_NAME="aws2"
+COMPLETER_EXE_NAME="aws_completer"
```
jamesls: Should this be `aws2_completer`?

kyleknap: Right. Good catch

kyleknap:
@jamesls I updated based on your comment.

PR #4623, opened 2019-11-05
Include PyYAML in setup.py as well as requirements.txt

joguSD:
This syncs our PyYAML dependency with what is in v1 develop and also includes it in our `setup.py`. This was removed but is still used in some customization logic, so it needs to be in `setup.py` still.


stealthycoin:
We should put this back in `setup.cfg` as well.

stealthycoin (approved):

PR #4636, opened 2019-11-08
Strip whitespace in aws configure sso prompts

joguSD:
This adds additional logic to strip any extra white space we get from prompts while running `aws configure sso` to help protect against copy paste errors.

### DIFF
--- a/tests/unit/customizations/configure/test_sso.py
+++ b/tests/unit/customizations/configure/test_sso.py
@@ -80,6 +80,11 @@ class TestPTKPrompt(unittest.TestCase):
         self.prompter.get_value('', '', validator=validator)
         self.assert_expected_validator(validator)
 
+    def test_strips_extra_whitespace(self):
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/configure/test_sso.py
```diff
@@ -49,8 +49,11 @@ def test_user_hits_enter_returns_current(self):
         self.assertEqual(response, 'default_value')
 
     def assert_expected_completions(self, completions):
+        # The order of the completion list can vary becuase it comes from the
+        # dict's keys. Asserting that each expected completion is in the list
         _, kwargs = self.mock_prompter.call_args_list[0]
-        self.assertEqual(kwargs['completer'].words, completions)
+        for completion in completions:
```
kyleknap: We should also assert the length for the completions or I believe unittest has `assertItemsEqual` (Python 2) and `assertCountEqual` (Python 3) that we can use.

PR #4664, opened 2019-11-15
Fix timestamp related s3 tests failing on MacOS

stealthycoin:
On MacOS the file timestamp can no longer be invalid. This means our
tests that were checking for a warning by setting an invalid timestamp
no longer emit such a warning on newer Macs.

In one test it was specifically for invalid timestamps. This one is now
skipped on MacOS since it is not a valid case to check for anymore.

The other was a test to see if an error and a warning can be emitted at
the same time. It was using an invalid timestamp incidentally to emit a
warning. This test is run the same way on windows, and now skipped on
posix systems. The test was duplicated for posix systems and uses an
unreadable file to generate the warning.

### DIFF
--- a/tests/functional/s3/test_cp_command.py
+++ b/tests/functional/s3/test_cp_command.py
@@ -16,6 +16,7 @@ import os
 
 from awscli.testutils import BaseAWSCommandParamsTest
 from awscli.testutils import capture_input, set_invalid_utime
+from awscli.testutils import skip_if_windows, skip_if_macos
 from awscli.compat import six
 from tests.functional.s3 import BaseS3TransferCommandTest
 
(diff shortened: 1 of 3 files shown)


stealthycoin:
Since the CI system only runs on Linux here are the runs for mac and windows:
Windows:
```
C:\Users\Administrator\aws-cli [fix-macos-timestamp-tests ≡ +1 ~0 -0 !]> nosetests .\tests\functional\s3
.......S.....................................................................................................
----------------------------------------------------------------------
Ran 109 tests in 10.917s

OK (SKIP=1)
```

MacOS
```
(env3.6.5) ➜  aws-cli git:(fix-macos-timestamp-tests) n tests/functional/s3
........S................................................................................................S...
----------------------------------------------------------------------
Ran 109 tests in 8.939s

OK (SKIP=2)
```

tests/functional/s3/test_cp_command.py
```diff
@@ -615,7 +616,29 @@ def test_upload_unicode_path(self):
...
+        unreadable_path = self.files.create_file('foo.txt', 'foo')
+        self.files.create_file('bar.txt', 'bar')
+        os.chmod(unreadable_path, 0o222)
+        _, stderr, rc = self.run_cmd(command % self.files.rootdir, expected_rc=1)
+        self.assertIn('upload failed', stderr)
+        self.assertIn('warning: Skipping file', stderr)
+        self.assertIn('File/Directory is not readable.', stderr)
+
+    @skip_if_macos("On MacOS we cannot create an invalid timestamp.")
+    def test_cp_with_error_and_warning_windows(self):
```
kyleknap: Maybe we should name it `test_cp_with_error_and_invalid_timestamp_warning`? Mainly this test will run for Linux as well. We can then use a similar naming approach in the previous test for unreadable files.

PR #4677, opened 2019-11-18
[v2] Bump dependency of botocore to dev1

kyleknap:


kyleknap:
I pulled off the commit that bumps the version for the CLI to dev1. There is one more fix that I want to make before bumping the version. CI currently does not pass for PR's as we need to depend on the version that is currently set for the botocore v2 branch, which is `dev1`. Merging just that dependency update.

PR #4681, opened 2019-11-19
Fix use of `user-attributes` in cmd `cognito-idp admin-create-user`

sebastienfi:
*Issue #, if available:*
https://github.com/amazon-archives/amazon-cognito-identity-js/issues/532
https://forums.aws.amazon.com/thread.jspa?threadID=311766

*Description of changes:*
The previous version gives an incorrect example use of multiple user attributes. Should be separated with a space, not a comma as options' doc states.


kdaily:
Hi @sebastienfi, it looks like this has been addressed. Thanks!

PR #4699, opened 2019-11-22
[v2] Correctly handle pagination args passed in CLI input JSON

garnaat:
*Description of changes:*

Currently, if you pass pagination arguments via the ``--cli-input-json`` parameter they will be ignored.  This is because the pagination options are handled early in the process by analyzing the arguments passed via the command line.  However, the input JSON is handled after this and the values are inserted directly into the calling parameters.

To fix this, this PR adds an additional pagination handler to check for pagination args in call parameters which may have been inserted by using the --cli-input-json option.


awscli/customizations/paginate.py
```diff
@@ -161,11 +163,27 @@ def unify_paging_params(argument_table, operation_model, event_name,
...
+    #
+    # The reason we have to do the second is that someone could put
+    # operation-specific pagination args in the CLI input JSON file
+    # directly and this bypasses all of the CLI args processing.
     session.register(
         parsed_args_event,
         partial(check_should_enable_pagination,
                 list(_get_all_cli_input_tokens(paginator_config)),
                 shadowed_args, argument_table))
+    session.register(
```
kyleknap: I wonder if it makes sense to move this handler's logic to the `CliInputJSONArgument.add_to_call_parameters` handler and have `CliInputJSONArgument.add_to_call_parameters` turn off the `parsed_globals.paginate` flag? It looks like the `CliInputJSONArgument.add_to_call_parameters` and `check_should_enable_pagination_call_parameters` are both registered to the `calling-command` event, and we can isolate this logic to only when `--cli-input-json` parameter is used and so it does not need to be called for every paginatable command.

I do think it would make more sense to keep it in the `unify_paging_params` handler though if there are other instances where we set the `call_commands` and need to take account for account for pagination, but I do not believe there is right now. Thoughts?

garnaat: If I understand correctly, you are suggesting that we move this code to the `add_to_call_parameters` method of the `CLIInputJSONArgument` class.  You are correct that this handler is already being registered on the `calling-command` event we need.

I think the only downside I see to that is that the handler in `cliinputjson.py` would still need to know what pagination input tokens it should be looking for.  So, we would need some pagination-specific functionality in the code processing input JSON and that code would kind of duplicate what is already in `paginate.py`.

I think I have a slight preference to keep all of the pagination-related code in `paginate.py` but not for any really important reason.

kyleknap: That's fair. I'm fine for keeping it in `paginate.py`

PR #4702, opened 2019-11-25
[v2] Default command output to a pager

kyleknap:
This introduces logic that takes the response from the CLI and output it directly to a pager (think `git diff`). This makes it much easier to consume/navigate a response from the service if it is really large. In terms of behavior, it closely follows the behavior of `git`'s pager (except for using `more` on Windows):

* By default if on Linux/MacOS, use `less` as the pager and set the `LESS` environment variable to `FRX` in order to set the appropriate flags. For Windows, `more` is used with no additional environment variables.

* You can override the default pager with the following configuration options. These are in order of precedence:
  * `AWS_PAGER` environment variable
  * `cli_pager` shared config variable
  * `PAGER` environment variable

* If you set any of the configuration options to an empty string (e.g. `AWS_PAGER=""`). The CLI will not send the output to a pager.

*  You can also set flags when specifying the pager and those will combine with any env vars we set (e.g. `AWS_PAGER="less -S"` will essentially make it `less -FRXS`). The combining flags behavior is builtin into `less`. You can also negate flags we set by specifying it on the command line: (e.g. `AWS_PAGER="less -+F"` will deactivate the quit if one screen behavior)

* If you set the `LESS` env var, we will not clobber it with ours (e.g. `FRX`).

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -89,6 +89,21 @@ def skip_if_windows(reason):
     return decorator
 
 
+def skip_if_not_windows(reason):
+    """Decorator to skip tests should only be ran for windows.
+
+    Example usage:
+
+        @skip_if_not_windows("Not valid")
+        def test_some_windows_stuff(self):
+            self.assertEqual(...)
+
+    """
+    def decorator(func):
+        return unittest.skipIf(platform.system() != 'Windows', reason)(func)
+    return decorator
+
+
 def set_invalid_utime(path):
     """Helper function to set an invalid last modified time"""
     try:
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -89,12 +89,12 @@ def skip_if_windows(reason):
     return decorator
 
 
-def skip_if_not_windows(reason):
+def if_windows(reason):
     """Decorator to skip tests should only be ran for windows.
 
     Example usage:
 
-        @skip_if_not_windows("Not valid")
+        @if_windows("Only supported on windows")
         def test_some_windows_stuff(self):
             self.assertEqual(...)
 
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -16,11 +16,11 @@ import datetime
 import contextlib
 import os
 import sys
-import subprocess
+from subprocess import Popen, PIPE
 import logging
 
 from awscli.compat import six
-from awscli.compat import get_binary_stdout
+from awscli.compat import get_stdout_text_writer
 from awscli.compat import get_popen_kwargs_for_pager_cmd
 from botocore.utils import IMDSFetcher
 from botocore.configprovider import BaseProvider
@@ -257,13 +257,27 @@ def is_a_tty():
 
 
 class OutputStreamFactory(object):
-    def __init__(self, popen=None):
+    def __init__(self, session, popen=None, environ=None,
+                 default_less_flags='FRX'):
+        self._session = session
         self._popen = popen
         if popen is None:
-            self._popen = subprocess.Popen
+            self._popen = Popen
+        self._environ = environ
+        if environ is None:
+            self._environ = os.environ.copy()
+        self._default_less_flags = default_less_flags
+
+    def get_output_stream(self):
+        pager = self._get_configured_pager()
+        if is_a_tty() and pager:
+            return self.get_pager_stream(pager)
+        return self.get_stdout_stream()
 
     @contextlib.contextmanager
     def get_pager_stream(self, preferred_pager=None):
+        if not preferred_pager:
+            preferred_pager = self._get_configured_pager()
         popen_kwargs = self._get_process_pager_kwargs(preferred_pager)
         try:
             process = self._popen(**popen_kwargs)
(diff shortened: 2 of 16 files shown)


awscli/testutils.py
```diff
@@ -89,6 +89,21 @@ def decorator(func):
     return decorator
 
 
+def skip_if_not_windows(reason):
```
joguSD: Do we already have precedent for this name?

awscli/utils.py
```diff
@@ -277,11 +291,21 @@ def get_pager_stream(self, preferred_pager=None):
 
     @contextlib.contextmanager
     def get_stdout_stream(self):
-        yield get_binary_stdout()
+        yield get_stdout_text_writer()
```
joguSD: What's the context for this change?

joguSD: What's the context on changing this? Just because we can due to Python 3?

awscli/utils.py
```diff
@@ -277,11 +291,21 @@ def get_pager_stream(self, preferred_pager=None):
...
+        return self._session.get_component('config_store').get_config_variable(
+            'pager'
+        )
 
     def _get_process_pager_kwargs(self, pager_cmd):
         kwargs = get_popen_kwargs_for_pager_cmd(pager_cmd)
-        kwargs['stdin'] = subprocess.PIPE
+        kwargs['stdin'] = PIPE
+        env = self._environ.copy()
+        if pager_cmd.startswith('less') and 'LESS' not in env:
```
joguSD: Do you think it makes sense to check for `less` usage? If we set the `LESS` env var and they're not using `less` it doesn't matter right? My concern here is that perhaps someone is using less but the command name is different, such as an alias like `myless` or something.

I tried taking a look at what `git` does (both the source and some testting) and it *seemed* like they will just always set the `LESS` env var if it isn't already set (if the flag was set at build time).

awscli/utils.py
kyleknap: Most of the output through the CLIDriver is written out as text. The goal of this was to unify the output stream factory to always produce text streams. I think before when I wrote this class for the history commands, I did not realize that I could force the stdin for Popen to be text with `universal_newlines` and so I just went with using binary streams. If I did not make this change, I would have to either not reuse this class or make all of the CLI formatters to write out binary, which would be a larger change. Does that make more sense?

awscli/testutils.py
kyleknap: Not really. I could name it to `skip_if_macos_or_linux` if that is better? I know there is this PR that adds a new skip decorator for macos: https://github.com/aws/aws-cli/pull/4664 but there is not a precedent for a group of operating systems.

awscli/utils.py
kyleknap: If `git` always sets `LESS`, I think that is reasonable, and the alias argument makes sense too.

awscli/testutils.py
joguSD: I mostly dislike the sort of double-negative 'skip if not'. Something like `if_windows`, `on_windows`, `only_on_windows`.

awscli/utils.py
joguSD: Cool, makes sense.

drummerjoe:
This needs to be disabled by default...Cmon folks.

seekayel:
This is a breaking changes for scripts that use the cli. Please set the default to not use pager.

seren:
All of our scripts broke because of this. Very much a breaking change.

andrew-glenn:
The blast radius on this is huge. opt-in or bust.

dxinteractive:
Please disable this by default. Now bash scripts that use `aws` commands that previously ran automatically require manual intervention. I know you can opt out, but really?

samj:
This caused much consternation here today; I'm surprised an opaque breaking change like this made it into production.

rollulus:
Not cool. Make it opt in. It's a breaking change and it is simply not how cli's are expected to work - with `git diff` a weird exception.

andrew-glenn:
I've submitted a PR to maintain this feature while making it something a user needs to explicitly configure via `AWS_PAGER` or a config option, removing the `PAGER` env var from the list of providers.

https://github.com/aws/aws-cli/pull/5398

jonni-larjomaa:
The one who introduced this should have serious talk with their fellows. aws cli is used by non interactive programs to communicate with resources in the aws. At least as many others have said make this opt-in not opt-out ... breaking years old conventions is just bad engineering IMO!

OliverBailey:
100% remove it being opt_out. 

Anything like this should be opt_in.

KapitalEarth:
😥

suneel944:
> This introduces logic that takes the response from the CLI and output it directly to a pager (think `git diff`). This makes it much easier to consume/navigate a response from the service if it is really large. In terms of behavior, it closely follows the behavior of `git`'s pager (except for using `more` on Windows):
> 
> * By default if on Linux/MacOS, use `less` as the pager and set the `LESS` environment variable to `FRX` in order to set the appropriate flags. For Windows, `more` is used with no additional environment variables.
> * You can override the default pager with the following configuration options. These are in order of precedence:
>   
>   * `AWS_PAGER` environment variable
>   * `cli_pager` shared config variable
>   * `PAGER` environment variable
> * If you set any of the configuration options to an empty string (e.g. `AWS_PAGER=""`). The CLI will not send the output to a pager.
> * You can also set flags when specifying the pager and those will combine with any env vars we set (e.g. `AWS_PAGER="less -S"` will essentially make it `less -FRXS`). The combining flags behavior is builtin into `less`. You can also negate flags we set by specifying it on the command line: (e.g. `AWS_PAGER="less -+F"` will deactivate the quit if one screen behavior)
> * If you set the `LESS` env var, we will not clobber it with ours (e.g. `FRX`).

The worst idea ever..... I am struggling to make this aws cli to run, it just hangs in the place not at all compatible with github actions. I frustrated to the core.

PR #4711, opened 2019-11-27
[v2] Implement yaml-stream output format

kyleknap:
This adds a new output format `yaml-stream`. This format will print out each response in a large yaml list as they are outputted. In general if a command has an output, it will be encapsulated in a list and each response will be outputted as they are received in the case for pagination. So for example:

If I get a single response, it will look like this:
```
$ aws dynamodb list-tables --output yaml-stream --no-paginate
- TableNames:
  - ChaliceChatTable
  - Forum
```

Then a paginated response will look like this:
```
$ aws dynamodb list-tables --output yaml-stream --page-size 1
- LastEvaluatedTableName: ChaliceChatTable
  TableNames:
  - ChaliceChatTable
- LastEvaluatedTableName: Forum
  TableNames:
  - Forum
```
This output format pairs nicely with outputting to a pager because you can start looking over the responses coming back as the CLI is receiving them instead of having to wait for the entire pagination process to complete to see results.

### DIFF
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -182,7 +182,8 @@ class StreamedYAMLFormatter(Formatter):
 
     def _get_response_stream(self, response):
         if is_response_paginated(response):
-            return map(self._get_transformed_response_for_output, response)
+            return compat.imap(
+                self._get_transformed_response_for_output, response)
         else:
             output = self._get_transformed_response_for_output(response)
             if output == {}:
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -35,6 +35,13 @@ class Formatter(object):
     def __init__(self, args):
         self._args = args
 
+    def _get_transformed_response_for_output(self, response):
+        # Performs various transformations to the API response ready to
+        # be outputted such as removing response id's and performing the
+        # query
+        self._remove_request_id(response)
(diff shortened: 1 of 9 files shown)


awscli/formatter.py
```diff
@@ -120,17 +128,70 @@ def _format_response(self, command_name, response, stream):
...
+                return
+            finally:
+                # flush is needed to avoid the "close failed in file object
+                # destructor" in python2.x. See:
+                # http://bugs.python.org/issue11380).
+                self._flush_stream(stream)
+
+    def _get_response_stream(self, response):
+        if is_response_paginated(response):
+            return map(self._get_transformed_response_for_output, response)
```
jamesls: Interesting.  While technically this doesn't affect our v2 builds because we use python3, this does technically change the semantics of a streaming response in python2 because `map()` isn't lazy in py2.  It's an interesting question, but I'm inclined to say that as long as we still test on python2.7, we should try to preserve behavior where possible.  What do you think about adding a something in `compat.py` and falling back to `itertools.imap` for py2?

kyleknap: Good catch. I was not even thinking about Python 2 here. Yeah I'll add a compat for it.

Issue #4680, opened 2019-11-19
aws eks command does not respect --endpoint-url when used with --role-arn


yeolahm:
I am working on the EC2 instance which does not have Internet access at all and I do not wish to provide it.

I am trying to get the eks token with following command and I am successful when sts private endpoint (regional) is used
```
aws --region eu-west-1 eks get-token --cluster-name example-cluster --endpoint-url https://sts.eu-west-1.amazonaws.com --debug
```
However, when I use --role-arn parameter it tries to hit sts global endpoint (sts.amazonaws.com) even though --endpoint-url is passed.
```
 aws  --region eu-west-1 --endpoint-url https://sts.eu-west-1.amazonaws.com eks get-token --cluster-name example-cluster  --role-arn "arn:aws:iam::76867867667:role/example-role"  --debug
```
Below is the snippet from the end of AWS debug output
```
5df3027f065fd18984cf80951', 'X-Amz-Date': '20191119T133511Z', 'User-Agent': 'aws-cli/1.16.284 Python/2.7.14 Linux/4.14.88-88.76.amzn2.x86_64 botocore/1.13.20'}>
2019-11-19 13:35:11,798 - MainThread - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): sts.amazonaws.com:443
```

dggondol:
This PR should fix the issue https://github.com/aws/aws-cli/pull/4728

joguSD:
This should work by setting the following env var:
```
AWS_STS_REGIONAL_ENDPOINTS=regional
```
Docs: https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#aws-sts

Which will use the regional endpoint instead of the global one (without requiring the user to pass an explicit endpoint url). Let me know if that works for you.

PR #4728, opened 2019-12-03
Do not ignore the --endpoint-url in "aws eks get-token"

dggondol:
When using the parameter --role in a private subnet using the VPC private link for STS the assume-role will use the public STS endpoint and ignore the one defined in --endpoint-url.

*Issue #, if available:*

*Description of changes:*


kdaily:
This should work by setting the following env var:
```
AWS_STS_REGIONAL_ENDPOINTS=regional
```
Docs: https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#aws-sts
This flag was added relatively recently, so you may need to upgrade your CLI version.

This will use the regional endpoint instead of the global one (without requiring the user to pass an explicit endpoint url). Let me know if that works for you.

_Originally posted by @joguSD in https://github.com/aws/aws-cli/issues/4680#issuecomment-565239063_

PR #4748, opened 2019-12-10
Add CLI binary format parameter

joguSD:
This adds a new parameter / configuration to modify how the CLI interprets binary values passed as a command line argument and how binary values are represented in the output. The new argument is `--cli-binary-format` and can currently be set to `base64` to consistently treat binary input as a base64 encoded string for input and represented as a base64 encoded string in output. Additionally, this can be set to `legacy` to maintain the same behavior that the V1 of the CLI had which would treat strings literally which can be useful for some commands.

The motivation for this change is that treating parameters that map to binary blobs in the request literally significantly restricts what binary values can be provided and also does not allow for binary values present in responses to be provided as input to subsequent commands without additional processing.

### DIFF
--- a/awscli/customizations/binaryformat.py
+++ b/awscli/customizations/binaryformat.py
@@ -15,6 +15,7 @@ import binascii
 
 from botocore.exceptions import ProfileNotFound
 
+from awscli.compat import six
 from awscli.shorthand import ModelVisitor
 
 
@@ -52,7 +53,7 @@ class InvalidBase64Error(Exception):
 
 class Base64DecodeVisitor(ModelVisitor):
     def _visit_scalar(self, parent, shape, name, value):
-        if shape.type_name != 'blob' or not isinstance(value, str):
+        if shape.type_name != 'blob' or not isinstance(value, six.text_type):
             return
         try:
             parent[name] = base64.b64decode(value)
--- a/awscli/customizations/binaryformat.py
+++ b/awscli/customizations/binaryformat.py
@@ -13,28 +13,24 @@
 import base64
 import binascii
 
+from botocore.exceptions import ProfileNotFound
+
 from awscli.shorthand import ModelVisitor
 
 
 def add_binary_formatter(session, parsed_args, **kwargs):
     binary_format = parsed_args.cli_binary_format
     if binary_format is None:
-        binary_format = session.get_config_variable('cli_binary_format')
+        try:
+            binary_format = session.get_config_variable('cli_binary_format')
+        except ProfileNotFound:
+            binary_format = 'base64'
     BinaryFormatHandler(binary_format).register(session)
 
 
-class Base64DecodeVisitor(ModelVisitor):
-    def _visit_scalar(self, parent, shape, name, value):
-        if shape.type_name != 'blob' or not isinstance(value, str):
-            return
-        try:
-            parent[name] = base64.b64decode(value)
-        except binascii.Error:
-            raise RuntimeError('Invalid base64: "%s"' % value)
-
-
 def base64_decode_input_blobs(params, model, **kwargs):
-    Base64DecodeVisitor().visit(params, model.input_shape)
+    if model.input_shape:
+        Base64DecodeVisitor().visit(params, model.input_shape)
 
 
 def identity(x):
@@ -48,6 +44,20 @@ def _register_blob_parser(session, blob_parser):
 
 def register_identity_blob_parser(session):
     _register_blob_parser(session, identity)
+
+
+class InvalidBase64Error(Exception):
+    pass
+
+
+class Base64DecodeVisitor(ModelVisitor):
+    def _visit_scalar(self, parent, shape, name, value):
+        if shape.type_name != 'blob' or not isinstance(value, str):
+            return
+        try:
+            parent[name] = base64.b64decode(value)
+        except binascii.Error:
+            raise InvalidBase64Error('Invalid base64: "%s"' % value)
 
 
 class BinaryFormatHandler(object):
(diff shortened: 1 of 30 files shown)


awscli/customizations/scalarparse.py
```diff
@@ -77,6 +77,4 @@ def add_timestamp_parser(session):
 
 
 def add_scalar_parsers(session, **kwargs):
-    factory = session.get_component('response_parser_factory')
```
kyleknap: Would we add the binary input handler logic here? If not, it does not make sense to keep this module as `scalarparse.py` and instead switch to something timestamp specific.

awscli/data/cli.json
```diff
@@ -66,6 +66,13 @@
             "dest": "connect_timeout",
             "type": "int",
             "help": "<p>The maximum socket connect time in seconds. If the value is set to 0, the socket connect will be blocking and not timeout.</p>"
+        },
+        "cli-binary-format": {
+            "choices": [
+                "base64",
+                "legacy"
+            ],
+            "help": "<p>The formatting style to be used for binary blobs. If the format is set to legacy, the value passed to the AWS CLI for a blob field will be treated literally for input and binary blobs will be represented as a Base64 encoded string in output. The legacy format is consistent with V1 of the AWS CLI. The base64 format will treat the values passed to the AWS CLI for a blob field as a Base64 encoded string, decoding it into the appropriate binary value for input, and binary blobs will be represented as a Base64 encoded string in output.</p>"
```
kyleknap: Small nitpicks:

* It may be useful to describe the default first here as most people will probably using that.
* It also may be worth to note what the default is.
* In general, it may be worth not including the explanation on how the CLI does serialization and just keep it as if base64, blobs must be base64 encoded and if legacy blobs must not be base64 encoded. This will make the description and a little more succinct.

awscli/customizations/binaryformat.py
```diff
@@ -0,0 +1,85 @@
...
+    _register_blob_parser(session, identity)
+
+
+class InvalidBase64Error(Exception):
+    pass
+
+
+class Base64DecodeVisitor(ModelVisitor):
+    def _visit_scalar(self, parent, shape, name, value):
+        if shape.type_name != 'blob' or not isinstance(value, six.text_type):
```
kyleknap: Out of curiosity when is the second clause of the if statement ever reached?

joguSD: @kyleknap The most common right now would be `fileb://`. A handler earlier in the chain may have already converted the value to binary.

awscli/customizations/scalarparse.py
joguSD: I didn't put it here because binary format changes the input and output. I suppose it makes sense to change this to be timestamp specific now.

awscli/data/cli.json
joguSD: Updated

joguSD:
Think I addressed everything except:
> Update the auto-generated manpages to better indicate that you must base64 encode input

Right now we format this documentation as something like `Key=type` for both shorthand syntax and JSON syntax in the nested case. For example: `UploadPolicy=blob` or `"UploadPolicy": blob,`. Don't know if it would be more confusing to have something like `UploadPolicy="base64string", considering it will depend on the chosen `cli_binary_format`. Or did you have something else in mind for this?

kyleknap:
Right. I think we can keep it as `blob` for now.

awscli/examples/iot-data/update-thing-shadow.rst
```diff
@@ -3,8 +3,9 @@
 The following ``update-thing-shadow`` example modifies the current state of the device shadow for the specified thing and saves it to the file ``output.txt``. ::
 
     aws iot-data update-thing-shadow \
+        --cli-binary-format legacy \
```
kyleknap: I think for the examples where we are using the `--cli-binary-format` flag, we should convert these to just use `fileb://`. Mainly it feels weird to have examples where it uses a parameter value called `legacy` as `legacy` can be interpreted as something that should not be used. On a similar train of thought, I wonder if we should circle back on the value name here and to come up with a better name for the legacy binary format? Looking at this example, it seems completely reasonable to specify json in the command line and so it would seem strange to recommend using a value called `legacy` when we have no plans to deprecate it.

kyleknap: Building on ideas of what to call it instead what are your thoughts on renaming `legacy` to `raw-base64`?

awscli/examples/ses/send-raw-email.rst
```diff
@@ -2,7 +2,9 @@
 
 The following example uses the ``send-raw-email`` command to send an email with a TXT attachment::
 
-    aws ses send-raw-email --raw-message file://message.json
+    aws ses send-raw-email \
+    --cli-binary-format legacy \
```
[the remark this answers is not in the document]
joguSD: We can't use `fileb://` because the binary content is nested in in `Data`.

awscli/examples/iot-data/update-thing-shadow.rst
joguSD: The places I used `legacy` I did so because it's the most ergonomic option given the way the API is structured and the type of content that is trying to be passed. For this particular one I suppose I could move the contents to a file and then use `fileb://`.

wparad:
I understand the motivation for adding the new flag, but why the hell did we make it the default. Now everyone literally has to add an additional flag every time they want to encrypt something. For example I changed from v1 to v2 not to long ago, and I couldn't figure out for days why my decrypted blobs kept being base64 decoded automatically.

Why is the option "Don't do anything magic" not the default option? Put value into encrypt, get value out from decrypt. That should always be the default.

PR #4756, opened 2019-12-12
Refresh bundled installer to work with modern pip/venv

jamesls:
This updates the bundled installer generation to use modern versions of
virtualenv, and as a result, modern versions of pip, setuptools, etc.

Some of these changes may seem unnecessary, but here's why I had
to make all these changes.

The motivation for this change was that we couldn't upgrade our version
of dateutils in botocore because it used new functionality in its
build/install process that required a modern version of setuptools.  The
version we pull in comes from our virtualenv which is a fairly old
version.  So the first thing we had to do was upgrade our version of
virtualenv we use in the bundled installer.

That's where it all begins.

First there's py26 issues, specifically that newer virtualenvs (>=16)
don't support python2.6.  So we have to branch and decide which version
of virtualenv to use based on our python version.  That's not really a
problem because new versions of dateutils don't support py26 anyways so
it doesn't change anything for py26.  We just need to add more branching
to the bundled installer.

Next we have the fact that because we're pulling in a new virtualenv
for > py26, we pull in a new pip.  A new, major version of pip, that is
not backwards compatible with the version we were previously using.
Some of the arguments have been removed.  So we have to update how we're
calling pip.

Next up we have the fact that dateutils uses pep517 as part of its build
process, which is fairly new.  This introduces a few wrinkles for us.
First, it utilizes "isolated builds", which should contain "only the
standard library and any explicitly requested build-dependencies."
Normally this is a good thing, but it's problematic for us because this
is an offline installer and we don't want pip creating a new build
environment for us and downloading dependencies from pypi.  So we have to
disable that feature (that's the `--no-build-isolation`).  Fortunately,
all of dateutils build deps we already have installed in our main
installation venv we create so we're all all set there.

As part of dateutils build process, pip shells out to other processes to
invoke various hooks.  It uses the same virtualenv by using
`sys.executable` when it invokes its python hooks.  Except that there
appears to be a bug, only for py37 as far as I can tell, where
`sys.executable` is wrong and instead of pointing to the virtualenv
python, it instead points to the system python.  As you might expect,
this causes the build/install to fail.  I've filed an issue on the
virtualenv repo to see what they think about it.

Turns out, using the built in `venv` module works as expected, in that
`sys.executable` is correctly set.  But of course `venv` is only
available in py33+.  Which brings us to our next change, conditionally
use `python -m venv` on py3, otherwise fall back to one of the versions
of external `virtualenv` on py26/py27.

This brings us to our final change, `python -m venv` does not install
wheel whereas `virtualenv` does install wheel.  So we need to install
wheel, conditionally, based on python version.

Going forward, there's a few things we should consider:

* py26 is being dropped real soon so a good chunk of this code can be
  removed.
* Once we can ensure we're using a modern version of pip after dropping
  py26, we should try switching to wheels instead of sdists where
  possible.  If a package has a py2/py3 compatible pure python wheel,
  (python-dateutil) we should use that instead of the sdist.  That
  way we can avoid all these build issues and just use the pre-built
  wheel.  I opted not to do that here because I'm sure there's some
  gotcha that I'm not considering.

Verified I could generate a bundled installer that works on py26/py27/py37.
Also bumped the version of dateutils in botocore and verified I could generate
a bundled installer with v2.8.1 through our release process.
I'll send a follow up PR to botocore that bumps the max dateutils version.

### DIFF
--- a/scripts/assets/constraints-bundled.txt
+++ b/scripts/assets/constraints-bundled.txt
@@ -1,0 +1,1 @@
+python-dateutil==2.8.0
--- a/scripts/install
+++ b/scripts/install
@@ -19,6 +19,19 @@ PACKAGES_DIR = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), 'packages')
 INSTALL_DIR = os.path.expanduser(os.path.join(
     '~', '.local', 'lib', 'aws'))
+IS_PY26 = sys.version_info[:2] == (2, 6)
+IS_PY3 = sys.version_info[0] == 3
+if IS_PY26:
+    # --no-build-isolation isn't available in the pip
+    # version we use in py26.
+    INSTALL_ARGS = (
+        '--no-binary :all: --no-cache-dir '
+        '--no-index '
+    )
+else:
+    INSTALL_ARGS = (
+        '--no-binary :all: --no-build-isolation --no-cache-dir --no-index '
+    )
 
 
 class BadRCError(Exception):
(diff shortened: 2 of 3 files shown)


jamesls:
@kyleknap pointed out that using the built in venv module for older versions can result in pip versions that don't support the no build isolation, which means we can't use dateutils with the built in venvs in py3.

Rather than go down another rabbit hole of now including a fixed pip version, I think we should call it and just lock to an older version of dateutils in the bundled installer.

jamesls:
Should be ready to look at again.  I added a commit that pins to dateutils 2.8.0 until we can drop py26/py33 support, which should be happening soon.  From the commit message:

It doesn't seem like we can install dateutils
from source in python3.3 due to the fact that the version of
virtualenv/pip we'd need to leverage the build features dateutils
uses doesn't support python3.3.

This isn't normally an issue for people installing dateutils
through normal (online) pip because they can just use the prebuilt .whl
file which skips all the build steps.

I think once we drop py2.6/py3.3 support we can come back to this
again and:

1. Drop the constraints file.
2. Switch to pre-built wheels where possible (py2/py3 pure python).
   This should greatly simplify the install process.

PR #4776, opened 2019-12-18
Fix issue 3131: checksum on folder content rather than zip file

gpiccinni:
*Issue #3131 *

*Description of changes:*
Currently aws cloudformation package generates a zip archive containing the entire function code and dependencies and, after creating the zip archive computes the checksum of the zip and uses this value for the artifact name.
This implies that, when using dependencies into build, each time you build your code dependencies get downloaded and included in the zip file, the checksum computed on the zip file changes every time as the checksum of the zip includes file extended attributes (mtime and ctime) and dependencies files are always freshly installed.
The resulting checksum is different and the resulting cloudformation template contains a new artifact url and causes the function to be re-deployed even if there is no actual change in any of the code or dependencies file.

The changes computes the checksum on the entire folder content **before** creating the zip, this implies that the checksum represents the actual content of each function including dependencies.
Therefore this changes allows to trigger a function deployment only when there is an actual change in the code.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -27,7 +27,7 @@ from awscli.customizations.cloudformation import exceptions
 from awscli.customizations.cloudformation.yamlhelper import yaml_dump, \
     yaml_parse
 import jmespath
-
+import checksumdir
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Hyardaowyn:
Why has this not been merged yet?

KapitalEarth:
I was just about to ask the same question.

justindho:
Hi @gpiccinni, thanks for the PR. Apologies for the delay in getting back to you.

We likely would not accept taking a new dependency on `checksumdir`, as it has not been updated in almost 2 years. In addition, the [checksumdir repo](https://github.com/cakepietoast/checksumdir) does not have any tests around it and is not considered stable. If we were to pull in a new dependency, we'd need a compelling reason on why it is necessary.

If you are still interested in working on this PR, we ask that the PR branch be rebased off of the latest `develop` branch. In addition, we'll also probably want to add tests around making sure a deploy is not triggered if there was no actual code change and that a deploy is triggered if there are code changes.

Our team just put out a [contribution guide](https://github.com/aws/aws-cli/blob/v2/proposals/contribution-guide.md) detailing improvements to the contribution process. Until the feedback above is addressed, I'll be marking this PR as a draft.

tim-finnigan:
Closing this PR for reasons mentioned in the [earlier comment](https://github.com/aws/aws-cli/pull/4776#issuecomment-1139934734) and because there hasn't been any further discussion here.

PR #4788, opened 2019-12-20
Missing \ in examples for invoke

mrxder:
In the example calls for aws lambda invoke the \ char is missing before line break.


joguSD:
Thanks for this!

PR #4822, opened 2020-01-08
Fix dynamodb pagination bug for query and scan

joguSD:
If the last evaluated key returned contained a binary primary key
subsequent requests would incorrectly double encode this field when
providing it as the exclusive start key leading to inconsistent or
incomplete pagination results.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -506,7 +506,8 @@ class ServiceOperation(object):
             event,
             call_parameters=call_parameters,
             parsed_args=parsed_args,
-            parsed_globals=parsed_globals
+            parsed_globals=parsed_globals,
+            session=self._session,
         )
         # There are two possible values for override. It can be some type
         # of exception that will be raised if detected or it can represent
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -506,8 +506,7 @@ class ServiceOperation(object):
             event,
             call_parameters=call_parameters,
             parsed_args=parsed_args,
-            parsed_globals=parsed_globals,
-            session=self._session,
+            parsed_globals=parsed_globals
         )
         # There are two possible values for override. It can be some type
         # of exception that will be raised if detected or it can represent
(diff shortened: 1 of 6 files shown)


awscli/clidriver.py
```diff
@@ -506,7 +506,8 @@ def __call__(self, args, parsed_globals):
             event,
             call_parameters=call_parameters,
             parsed_args=parsed_args,
-            parsed_globals=parsed_globals
+            parsed_globals=parsed_globals,
+            session=self._session,
```
jamesls: I would prefer we do this without modifying clidriver.  In general, I think if there's an option to avoid addings more kwargs to an event we should take that option, provided it's not substantially more complex.

I'd prefer we have a class that stores a reference to the session on `session-initialized`, and then have another method that registers the `after-call` handler if needed.

awscli/customizations/dynamodb.py
```diff
@@ -0,0 +1,49 @@
...
+    # ensures that the continuation token is properly converted to binary to
+    # avoid double encoding the contination token.
+    last_evaluated_key = parsed.get('LastEvaluatedKey', None)
+    if last_evaluated_key is None:
+        return
+    for key, val in last_evaluated_key.items():
+        if 'B' in val:
+            try:
+                val['B'] = base64.b64decode(val['B'])
+            except (binascii.Error, TypeError):
```
jamesls: Thoughts on a hard failure here?  If this decode fails that indicates we received bad data from the service, or something else mucked with this value.  Either way, we can't guarantee that we're paginating correctly which brings us back to the inconsistent/complete pagination results.

FWIW if the base64 decode fails in the botocore parser layer (i.e we didn't mess with the blob parser defaults), this would propagate an exception.

awscli/clidriver.py
joguSD: Yeah I like that more, I'll try that out.

awscli/customizations/dynamodb.py
joguSD: Hmmm, yeah. I think hard failing is probably better than silently producing wrong results (potentially). This was mostly defensive in case something else ever parsed this into bytes already and the fact that it's ambiguous on Python 2. Though in that case one of the tests will probably fail so I think we can hard fail.

PR #4844, opened 2020-01-15
[v2] Cloudformation default no fail

stealthycoin:
Update cloudformation deploy default to not fail on empty changeset

Existing Tests were updated to be explicit about what their changeset
failure argument was and an additional test was added for the default case.

### DIFF
--- a/.changes/next-release/enhancment-cloudformation-66951.json
+++ b/.changes/next-release/enhancment-cloudformation-66951.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancment",
+  "category": "cloudformation",
+  "description": "Changes the default of cloudformation deploy to not fail on an empty changeset."
+}
--- a/.changes/next-release/enhancment-cloudformation-66951.json
+++ b/.changes/next-release/enhancment-cloudformation-66951.json
@@ -1,5 +1,0 @@
-{
-  "type": "enhancment",
-  "category": "cloudformation",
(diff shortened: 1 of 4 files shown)


.changes/next-release/enhancment-cloudformation-66951.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancment",
```
kyleknap: enhancment -> enhancement. On that note, I'm not sure if enhancement is the right label here. It is more of a breaking change. I'm also fine holding off on the changelog entry for now as we will likely document all of these breaking changes in the Changelog when v2 goes GA.

tests/unit/customizations/cloudformation/test_deploy.py
```diff
@@ -365,6 +365,24 @@ def test_deploy_does_not_raise_exception_on_empty_changeset(self):
             fail_on_empty_changeset=False
         )
 
+    def test_deploy_empty_changeset_does_not_raise_exception_by_default(self):
```
kyleknap: What's your thoughts on writing functional tests for these? I realize that there is no existing functional tests for this command set, but it would be really helpful to make sure everything is plumbed through correctly.

tests/functional/cloudformation/test_deploy.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
```
kyleknap: We can probably remove the shebang.

PR #4855, opened 2020-01-20
help config-vars: render groff tables

philpennock:
Fix missing `-t` flag to groff(1), to ask it to pre-process with tbl(1), which handles tables.  Needed for `aws help config-vars` at the very least.

### DIFF
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -109,7 +109,7 @@ class PosixHelpRenderer(PagingHelpRenderer):
         man_contents = publish_string(contents, writer=manpage.Writer())
         if not self._exists_on_path('groff'):
             raise ExecutableNotFoundError('groff')
-        cmdline = ['groff', '-m', 'man', '-T', 'ascii']
+        cmdline = ['groff', '-m', 'man', '-T', 'ascii', '-t']
         LOG.debug("Running command: %s", cmdline)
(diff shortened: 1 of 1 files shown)


aBurmeseDev:
Upon team discussion and testing, adding `-t` flag could potentially break older version and we need to find a better approach. Additionally, the format issue is only reproducible when `groff` is installed on the machine. We'll have to investigate further and meanwhile, I'm going to convert this PR to draft.

philpennock:
I'm slightly perplexed.  I'm fairly sure that groff took `-t` for inserting `tbl` back when I started learning Unix in the mid-90s.  Since this only affects installs where groff is available, I do not see what the compatibility issue is.

Checking the groff repo, I see that commit 9bd39c6872faa7f610a66ab2ff1894e7154a628b on 2001-04-18 gave the `nroff` wrapper script support for `-t`, so `-t` was present in the groff binary before then.  So that's 22 years old.

Checking the actual source code, `-t` was added in commit b04d345ed2 which is an import of release tarballs when constructing the repo, so it was in release 1.06 of groff, from 1992-09-01.  Making it now more than 31 years old.

I think the compatibility concerns are misplaced.

philpennock:
This is still a problem and I believe that I addressed compatibility concerns in my comment on 2023-09-27.  I think this is safe to merge.

Issue #4853, opened 2020-01-19
Importing ABC directly from collections module will be ImportError in Python 3.9


tirkarthi:
This was emitting deprecation warning from Python 3.4 and it will stop working in Python 3.9 . Reference : https://github.com/python/cpython/pull/10596 . There is already awscli/compat.py so it would be good to import ABC from `collections_abc` that has Python 2 and 3 compatibility.

Thanks for the library!

```
awscli/customizations/history/db.py
19:from collections import MutableMapping
tests/__init__.py
1:from collections import MutableMapping
2:from collections import Mapping
```

hugovk:
Please see PR https://github.com/aws/aws-cli/pull/4860.

PR #4860, opened 2020-01-21
Fix collections.abc imports for Python 3.9

hugovk:
*Issue #, if available:* Fixes https://github.com/aws/aws-cli/issues/4853.

*Description of changes:*

This was emitting deprecation warning from Python 3.4 and it will stop working in Python 3.9. Reference: [python/cpython#10596](https://github.com/python/cpython/pull/10596). There is already awscli/compat.py so import ABCs from `collections_abc` that has Python 2 and 3 compatibility.

---

And a heads up: you'll need to migrate off of Nose for testing, it's unmaintained and doesn't support Python 3.9: https://github.com/nose-devs/nose/issues/1099. I'd recommend pytest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,16 @@ matrix:
   include:
     - python: 2.7
       dist: trusty
-      sudo: false
     - python: 3.4
       dist: trusty
-      sudo: false
     - python: 3.5
       dist: trusty
-      sudo: false
     - python: 3.6
       dist: trusty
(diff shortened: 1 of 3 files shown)


fchorney:
I've been seeing this deprecation in tests for a long time. Glad somebody came here to supply a fix. Looks good! Hope it gets merged in

joguSD:
Thanks for the PR, this looks good to me!

PR #4874, opened 2020-01-24
adds `aws ecr get-login-password` customization

matthew-russo:
*Issue #, if available:*
fixes #4867 

*Description of changes:*
as proposed in https://github.com/aws/aws-cli/issues/4867
and previously discussed in
https://github.com/aws/aws-cli/issues/2875#issuecomment-433565983
https://github.com/aws/aws-cli/issues/3687#issue-374397564

this commit adds a new customization command for ECR that only
returns the password to login to a registry. this approach is
composable, is compatible with other container clients, allows
use of functionality like Docker's --password-stdin flag, and
is more resilient to changes in client APIs

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4876, opened 2020-01-24
New CLI exampeles for EC2 and ECR

bisdavid:

### DIFF
--- a/awscli/examples/ecr/get-authorization-token.rst
+++ b/awscli/examples/ecr/get-authorization-token.rst
@@ -1,58 +1,17 @@
 **To get an authorization token for your default registry**
 
-This example command gets an authorization token for your default registry.
+The following ``get-authorization-token`` example command gets an authorization token for your default registry. ::
 
-Command::
-
-  aws ecr get-authorization-token
+    aws ecr get-authorization-token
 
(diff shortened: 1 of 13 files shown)


awscli/examples/ecr/get-authorization-token.rst
```diff
@@ -1,58 +1,49 @@
...
+            }
+        ]
+    }
 
-**To get the decoded password for your default registry**
+**Example 2: To get the decoded password for your default registry**
 
-This example command gets an authorization token for your default registry and
-returns the decoded password for you to use in a ``docker login`` command.
+The following ``get-authorization-token`` example command gets an authorization token for your default registry and returns the decoded password for you to use in a ``docker login`` command.
```
kyleknap: So just today we added a new `get-login-password` command that essentially performs the logic for this example: https://github.com/aws/aws-cli/pull/4874. I'm wondering if we should actually just remove example 2 and example 3 here and point people to use the new command instead? Using the `get-login-password` command is going to be the recommended pattern going forward.

bisdavid:
Got permission from the service writer to remove the two examples. At a later date, he'll submit an example for the new get-login-password command.

Issue #4867, opened 2020-01-22
Proposal: aws ecr get-login-password


rpnguyen:
This is a proposal for a new AWS CLI command for ECR

```
$ aws ecr get-login-password
cGFzc3dvcmQ=
```

This command can be used in the following ways:

```
$ aws ecr get-login-password | docker login --username AWS --password-stdin 111111111111.dkr.ecr.us-west-2.amazonaws.com
Login Succeeded

$ docker login --username AWS --password "$(aws ecr get-login-password)" 111111111111.dkr.ecr.us-west-2.amazonaws.com
Login Succeeded
```

This idea has been previously proposed by @theY4Kman https://github.com/aws/aws-cli/issues/2875#issuecomment-433565983 and @kojiromike https://github.com/aws/aws-cli/issues/3687#issue-374397564

Issue #3687, opened 2018-10-26
eval aws ecr get-login should provide option to use password-stdin


kojiromike:
The instructions are pretty clear that `aws ecr get-login` prints credentials. Since these credentials are designed to be used in an `eval` context, the password can be observed elsewhere on the machine.

```bash
$ eval "$(aws ecr get-login --no-include-email)"
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
```

But recent versions of docker support a more secure approach using `--password-stdin`. You can use it now, but it's cumbersome:

```bash
$ read -sra cmd < <(aws ecr get-login --no-include-email)
$ pass="${cmd[5]}"
$ unset cmd[4] cmd[5]
$ "${cmd[@]}" --password-stdin <<< "$pass"
Login Succeeded
```

Or as a one-liner if you feel like manually composing the endpoint url:

```bash
$ aws ecr get-login --no-include-email | awk '{print $6}' | docker login -u AWS --password-stdin "$endpoint"
Login Succeeded
```

aws could support clean, secure docker login by providing more machine-parseable output, by providing a docker credentials-helper, or by providing the password separately from the rest of the command as in:

```bash
$ aws ecr get-password | eval "$(aws ecr get-login --password-stdin)"
```

gonrial:
Duplicated https://github.com/aws/aws-cli/issues/2875

justnance:
@gonrial - Thanks for pointing out this is a duplicated.  I'm going to close this issue so we can continue tracking the discussion under #2875.

PR #4886, opened 2020-01-28
[v2] Remove ecr get-login command

kyleknap:
The command was deprecated in v1 and replaced with the `ecr get-login-password` command


itajaja:
get-login was so useful :( where can I find the discussion with the reason for why it was removed?

pda:
The deprecation PR is here: https://github.com/aws/aws-cli/pull/4877 but it doesn't have much more information. The addition of `get-login-password` happened in https://github.com/aws/aws-cli/pull/4874 after being proposed in https://github.com/aws/aws-cli/issues/4867 with a bit more context in https://github.com/aws/aws-cli/issues/3687#issue-374397564
/cc @itajaja

rpnguyen:
Sorry that we did not make this more discoverable but we also put some information about the deprecation here: https://github.com/aws/containers-roadmap/issues/735

rpnguyen:
We also put some information about the deprecation here: https://github.com/aws/containers-roadmap/issues/735

itajaja:
alas, I guess it's too late now, but while `get-login` is surely more flexible, the old command was much easier to use, and I wish we could have have both. having to specify the aws account id is surely cumbersome.

rpnguyen:
You may also consider using the [ECR Docker credential helper](https://github.com/awslabs/amazon-ecr-credential-helper) which removes the need to call an authentication CLI command altogether.

PR #4890, opened 2020-01-30
[v2] Improve return code convention and consistency

joguSD:
This is a sweeping audit of mostly the customization commands in the AWS CLI to introduce a more nuanced return code convention and applying that convention consistently. This aims to have a little more granularity in the RCs used to convey broad categories of errors and to help distinguish between client side and server side errors.

The convention is:

* 0 Success
* 1 - 127 Reserved for custom commands to express granular exceptions
* 130 - Keyboard interrupt (Ctrl-C)
* 252 Client side failure to parse the command / make the request (Failing argparse, unknown parameter, failing botocore parameter validation, etc).
* 253 Configuration error. This indicates that the command invoked may not necessarily be wrong but the environment/configuration is invalid (No region, no credentials, etc).
* 254 Client error. This indicates there was a successful request/response with the service but the request failed. Indicates that the root issue is likely specific to the service
* 255 Catch all error. 255 specifically as a return code should not be relied on and a particular exception can move from 255 to a specific RC such as the above.

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -14,6 +14,7 @@ import argparse
 import sys
 from awscli.compat import six
 from difflib import get_close_matches
+from gettext import gettext
 
 
 HELP_BLURB = (
@@ -107,6 +108,21 @@ class CLIArgParser(argparse.ArgumentParser):
                         encoded.append(v)
                 setattr(parsed, arg, encoded)
         return parsed, remaining
+
+    def error(self, message):
+        """error(message: string)
+
+        NOTE: This is lifted directly from argparse, the only difference being
+        we use 252 for parsing errors.
+
+        Prints a usage message incorporating the message to stderr and
+        exits.
+        If you override this in a subclass, it should not return -- it
+        should either exit or raise an exception.
+        """
+        self.print_usage(sys.stderr)
+        args = {'prog': self.prog, 'message': message}
+        self.exit(252, gettext('%(prog)s: error: %(message)s\n') % args)
 
 
 class MainArgParser(CLIArgParser):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -58,6 +58,10 @@ from awscli.utils import emit_top_level_args_parsed_event
 from awscli.utils import write_exception
 from awscli.utils import OutputStreamFactory
 from awscli.utils import IMDSRegionProvider
+from awscli.constants import (
+    PARAM_VALIDATION_ERROR_RC, CONFIGURATION_ERROR_RC, CLIENT_ERROR_RC,
+    GENERAL_ERROR_RC,
+)
 from awscli.customizations.exceptions import ParamValidationError
 from awscli.customizations.exceptions import ConfigurationError
 
@@ -330,28 +334,28 @@ class CLIDriver(object):
             # client side validation at the botocore level.
             LOG.debug("Client side parameter validation failed", exc_info=True)
             write_exception(e, outfile=get_stderr_text_writer())
-            return 252
+            return PARAM_VALIDATION_ERROR_RC
         except UnknownArgumentError as e:
             sys.stderr.write("usage: %s\n" % USAGE)
             sys.stderr.write(str(e))
             sys.stderr.write("\n")
-            return 252
+            return PARAM_VALIDATION_ERROR_RC
         except ConfigurationError as e:
             # RC 253 represents that the command may be syntatically correct
             # but the environment or configuration is incorrect.
             LOG.debug("Invalid CLI or client configuration", exc_info=True)
             write_exception(e, outfile=get_stderr_text_writer())
-            return 253
+            return CONFIGURATION_ERROR_RC
         except NoRegionError as e:
             msg = ('%s You can also configure your region by running '
                    '"aws configure".' % e)
             self._show_error(msg)
-            return 253
+            return CONFIGURATION_ERROR_RC
         except NoCredentialsError as e:
             msg = ('%s. You can configure credentials by running '
                    '"aws configure".' % e)
             self._show_error(msg)
-            return 253
+            return CONFIGURATION_ERROR_RC
         except KeyboardInterrupt:
             # Shell standard for signals that terminate
             # the process is to return 128 + signum, in this case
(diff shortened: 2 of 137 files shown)


awscli/argparser.py
```diff
@@ -108,6 +109,21 @@ def parse_known_args(self, args, namespace=None):
...
+        we use 252 for parsing errors.
+
+        Prints a usage message incorporating the message to stderr and
+        exits.
+        If you override this in a subclass, it should not return -- it
+        should either exit or raise an exception.
+        """
+        self.print_usage(sys.stderr)
+        args = {'prog': self.prog, 'message': message}
+        self.exit(252, gettext('%(prog)s: error: %(message)s\n') % args)
```
kyleknap: I think it may be worth while if we add a `constants` module and start defining all of these error codes, especially because we are using these in more than one module.

awscli/clidriver.py
```diff
@@ -314,28 +320,51 @@ def main(self, args=None):
...
+            write_exception(e, outfile=get_stderr_text_writer())
+            return 252
         except UnknownArgumentError as e:
             sys.stderr.write("usage: %s\n" % USAGE)
             sys.stderr.write(str(e))
             sys.stderr.write("\n")
-            return 255
+            return 252
+        except ConfigurationError as e:
+            # RC 253 represents that the command may be syntatically correct
```
kyleknap: We should be documenting these in the topic guide on rc's: https://github.com/aws/aws-cli/blob/v2/awscli/topics/return-codes.rst

awscli/customizations/configure/configure.py
```diff
@@ -124,6 +124,7 @@ def _run_main(self, parsed_args, parsed_globals):
                 section = profile_to_section(profile)
                 new_values['__section__'] = section
             self._config_writer.update_config(new_values, config_filename)
+        return 0
```
kyleknap: For the commands that subclass from `BasicCommand` what is your thoughts on adding logic in the `main()` that if `_main()` does not return anything, the command will return a successful rc on your behalf? I'm mainly suggesting it because it is not necessarily obvious and can be easy to forget when writing these commands that an rc needs to be returned.

tests/unit/test_clidriver.py
```diff
@@ -632,13 +632,13 @@ def test_custom_command_schema(self):
         rc = driver.main(
             'ec2 foo --bar Count=4'.split())
 
-        self.assertEqual(rc, 255)
+        self.assertEqual(rc, 252)
 
         # Test extra unknown shorthand item
         rc = driver.main(
```
kyleknap: What's your thoughts on return 252 for invalid shorthand syntax as well? I noticed that the PR did not touch the `shorthand.py` module and it has some parsing exceptions: https://github.com/aws/aws-cli/blob/9a897c0c0de4c2c3908edb2d2d4e1efbf1e44d6d/awscli/shorthand.py#L57

kyleknap:
After doing an audit, here are some more places that we may want to update the RC's:

* `customizations/ec2/` - There are some `ValueError`s we missed
* `history list` - Throws `RuntimeError`  when `cli_history` is not enabled.
* `sso login` - Throws an error for not being configured.

I think there is a fair amount more in other command sets. Specifically, I did the following greps:

* Search for `expected_rc=255` in the functional test suite.
* Search for `raise ValueError` in the CLI codebase. Another idea to handle these one is to possibly just catch `ValueError`s in the driver, but I'm not sure if it is applicable for all `ValueErrors` raised.
* Search for `raise RuntimeError` in the CLI codebase. We may be able to catch these as well, but I'm less confidant about doing that compared to `ValueError`.

awscli/argparser.py
joguSD: Yeah, I was thinking about doing this. I'll break these out into constants.

tests/unit/test_clidriver.py
joguSD: I'll take a look at the shorthand stuff as well.

joguSD:
>     * `customizations/ec2/` - There are some `ValueError`s we missed

I'll do another once over on these. I might have missed some but some I didn't change intentionally. In general, I was leaving validation errors that happened *after* a client call occurred alone as I saw that as more of runtime behavior (and thus 255 by default) and not command validation error.

>     * `history list` - Throws `RuntimeError`  when `cli_history` is not enabled.
> 
>     * `sso login` - Throws an error for not being configured.

Are you expecting these to be raised as configuration errors? 

> 
> I think there is a fair amount more in other command sets. Specifically, I did the following greps:
> 
>     * Search for `expected_rc=255` in the functional test suite.
>     * Search for `raise ValueError` in the CLI codebase. Another idea to handle these one is to possibly just catch `ValueError`s in the driver, but I'm not sure if it is applicable for all `ValueErrors` raised.

>     * Search for `raise RuntimeError` in the CLI codebase. We may be able to catch these as well, but I'm less confidant about doing that compared to `ValueError`.

You'll need to point of specific instances and what the `RuntimeError` or `ValueError` actually represents. There's no consistency on what the exceptions are supposed to mean and a simple catch all of `ValueError` is 252 doesn't make sense. `RuntimeError` I'm intentionally not changing in most places because generally they indicate that something went wrong during the execution of the command and thus should be treated as 255 or explicitly mapped to something between 1-127.

tests/unit/test_clidriver.py
joguSD: Looks like our general argprocessing module has a `ParamError` exception (which is used to reraise the shorthand exceptions). I'll add this to the list of param errors we catch.

awscli/clidriver.py
joguSD: Updated it to reflect the new return codes. We should probably figure something out for codes 1-127. Right now it looks like we're trying to overload the general topic guide with some command specific definitions (the S3 ones) which probably won't work out long term.

awscli/topics/return-codes.rst
```diff
@@ -31,8 +31,23 @@ of a CLI command:
...
+* ``252`` -- Command syntax was invalid, an unknown parameter was provided, or
+  a paramter value was incorrect and prevented the command from running.
+
+* ``253`` -- The system environment or configuration was invalid. While the
+  command provided may be syntactically valid, missing configuration or
+  credentials prevented the command from running.
+
+* ``254`` -- The command was succesfully parsed and a request was made to the
+  specified service but the service returned an error. This will generally
+  indicate incorrect API usage or other service specific issues and is likely
```
kyleknap: The "is likely not an issue with the request being sent by the CLI" is sort of weird phrasing. Honestly, we can probably just remove that last part.

kyleknap:
Update looks good! Merging.

PR #4891, opened 2020-01-30
[v2] Rename aws2 back to aws

kyleknap:
The rationale is that there are relatively few breaking changes in AWS CLI v2 to warrant a new namespace for CLI v2 when it GA's. If a user wants to have the two CLI's side by side, they can always add an alias for one of them.

### DIFF
--- a/awscli/customizations/configure/sso.py
+++ b/awscli/customizations/configure/sso.py
@@ -111,9 +111,9 @@ def display_account(account):
 
 class ConfigureSSOCommand(BasicCommand):
     NAME = 'sso'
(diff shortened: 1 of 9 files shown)


joguSD:
Changes look fine. In the RC convention changes we added a `constants` module. Not sure if it makes sense to add the binary name to the constants and reference it?

kyleknap:
In the code base it could make sense for `sso` for example. However for the test and exe generation scripts, I'm not sure how feasible it is because the script may not be relying on `awscli` package being installed in order to run. This is especially true for any of the bash scripts.

PR #4908, opened 2020-02-04
[CLIv2] Backfill changelog entires for CLI v2

joguSD:
This backfills changelog entries for all of the bugfixes, enhancements, and features for CLI V2.


.changes/next-release/feature-configure-74317.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "configure",
+  "description": "Added the configure list-profiles command. See `#3735 <https://github.com/aws/aws-cli/issues/3735>`__."
```
kyleknap: For some of these commands, it may be worth adding the literal notation so it gets rendered to code when the rst document is formed.

.changes/next-release/feature-paramfile-57445.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
```
kyleknap: Talked offline, but for anything we are changing defaults or removing functionality, we should label it as `breaking-change` in order to make it easier to identify breaking changes when parsing through the changelog.

PR #4909, opened 2020-02-04
Fix README in pkg installer to dynamically update the CLI version

stealthycoin:
The readme version string will now always be up to date.


kyleknap (approved):
:ship:

nateprewitt (approved):

PR #4913, opened 2020-02-04
colorama version bump

stealthycoin:
dependabot does not do everything


nateprewitt:
Looks like we can't upgrade until we drop 3.4 support.

PR #4915, opened 2020-02-04
[v2] Remove unused py2 code

jamesls:
Most of these changes were isolated to compat, which is now
mostly a module for cross platform compat instead of python version
compatability.

The diff for compat looks off, but all I basically changed was:

```
if six.PY3:
    A
else:
    B
```

to:

```
A
```


kyleknap (approved):
🚢

stealthycoin (approved):

Issue #4719, opened 2019-11-29
Remove dependency on botocore's TextWriter


borlafu:
Is there any possibility we can remove the dependency on `botocore.docs.bcdoc.textwriter.TextWriter`?

I see it's being used here:
https://github.com/aws/aws-cli/blob/4ff0cbacbac69a21d4dd701921fe0759cf7852ed/awscli/help.py#L25

And that's blocking Botocore from removing it's dependency on Docutils:
https://github.com/boto/botocore/issues/900

This poses a licensing issue to many Botocore users, as well as kind-of a pain when dealing with dependency resolution.

jamesls:
I'd still like to remove this eventually, but we do need to have a suitable replacement for it in the meantime.  AFAIK no one's taken a crack at this, and it's not at the top of our backlog at the moment.  Open to other ideas though.

jwhitlock:
Would it be acceptable to move https://github.com/boto/botocore/blob/develop/botocore/docs/bcdoc/textwriter.py into this project, such as awclit/textwriter.py? ``TextWriter`` is unused on the botocore side, and that would be a first step to removing it (and docutils) from botocore.

borlafu:
I think that would be an acceptable way forward.

amw:
I believe this has already been completed.

kdaily:
This has been completed via https://github.com/aws/aws-cli/commit/5822bf405ba0f7c00a12b56c50ced2a5b66eae8c.

Thanks!

PR #4917, opened 2020-02-04
Copy textwriter.py from botocore project

jwhitlock:
Re issue #4719, [botocore/docs/bcdoc/textwriter.py](https://github.com/boto/botocore/blob/develop/botocore/docs/bcdoc/textwriter.py) is not used by botocore, and may be deleted in botocore to remove the docutils dependency. This PR imports the code into this project. There are no related tests to import.


kdaily:
Hi @jwhitlock, I think this will be taken care of by https://github.com/aws/aws-cli/pull/5538. Thanks!

PR #4932, opened 2020-02-07
Pop AWS_REGION from env vars in config test

jamesls:
We have to make sure we remove all env vars related to
configuring a region in this test.  We're verifying that we're
pulling the region from the config file when no env vars for
region exist.


joguSD (approved):
Assuming travis passes.

kyleknap (approved):
:ship:

Issue #4815, opened 2020-01-06
aws s3 ls s3:///random_nonsense returns true


stephenmsachs:
Calls to "aws s3 ls s3:///random_nonsense" will return true when called with three or more dashes, while two dashes provide the correct non-zero return value.

kyleknap:
I ran the command and it listed out my buckets. This is definitely not intentional. I believe it is happening because of this [line](https://github.com/aws/aws-cli/blob/f2d5a45151761f395a03b7f6a85801585caf5d7a/awscli/customizations/s3/subcommands.py#L477). Specifically because there are three slashes, it treats the bucket as a Falsy value and causes the buckets be listed out. Here we should just validate that the bucket field is not an empty string, especially if a key is provided.

PR #4936, opened 2020-02-08
Fix "aws s3 ls s3:///random_nonsense returns true"

wingkwong:
*Issue #, if available:*
#4815

*Description of changes:*
Currently running aws s3 ls s3:///random_nonsense returns all buckets. This change is to validate that the bucket field is not an empty string especially if a key is provided so that it returns false instead.

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -475,6 +475,8 @@ class ListCommand(S3Command):
             path = path[5:]
         bucket, key = find_bucket_key(path)
         if not bucket:
+            if key:
+                return 0
             self._list_all_buckets()
         elif parsed_args.dir_op:
             # Then --recursive was specified.


awscli/customizations/s3/subcommands.py
```diff
@@ -475,6 +475,8 @@ def _run_main(self, parsed_args, parsed_globals):
             path = path[5:]
         bucket, key = find_bucket_key(path)
         if not bucket:
+            if key:
+                return 0
```
vz10, line 479: Probably it'd better to raise a ParamValidation exception here instead of just successfully exit without any output. What do you think about checking both the key and the bucket instead?
```
if not bucket and not key:
    self._list_all_buckets()
```

vz10 (commented):
Looks good 👍
Just one small suggestion.

awscli/customizations/s3/subcommands.py
wingkwong: Is it expected?

```
aws s3 ls s3:///random_nonsense
```

Result:

```
Parameter validation failed:
Invalid bucket name "": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" or be an ARN matching the regex "^arn:(aws).*:s3:[a-z\-0-9]+:[0-9]{12}:accesspoint[/:][a-zA-Z0-9\-]{1,63}$"
```

Case: no bucket & no key

```
aws s3 ls s3:///
```

Result:

```
2021-03-31 17:16:51 bucket-1
2019-12-27 13:48:23 bucket-2
2020-09-04 13:36:46 bucket-3
```

vz10: Yes, I think this the best what we can do. In the first case it's unclear what customer actually needs and we'd better fail than try to guess, in the second one since input is technically empty we may show everything.

wingkwong: Got it. I'll update it tonight. Thanks a lot.

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -132,6 +132,26 @@ def test_ls_command_with_no_args(self):
             's3', region_name=None, endpoint_url=None, verify=None,
             config=None))
 
+    def test_ls_command_with_non_existsent_url(self):
```
justindho, line 135: Looks like there is a typo in the test name
```suggestion
    def test_ls_command_with_non_existent_url(self):
```

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -132,6 +132,26 @@ def test_ls_command_with_no_args(self):
...
+                                request_payer=None, page_size=None)
+        ls_command._run_main(parsed_args, parsed_global)
+        # We should make no operation calls.
+        call = self.session.create_client.return_value.list_buckets
+        self.assertEqual(call.call_count, 0)
+
+        # Verify get_client
+        get_client = self.session.create_client
+        args = get_client.call_args
+
```
justindho, line 150: There seems to be some extra whitespace appended on this line. Could we get this removed?

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -132,6 +132,26 @@ def test_ls_command_with_no_args(self):
             's3', region_name=None, endpoint_url=None, verify=None,
             config=None))
 
+    def test_ls_command_with_non_existsent_url(self):
+        ls_command = ListCommand(self.session)
+        parsed_global = FakeArgs(region=None, endpoint_url=None,
```
justindho, line 137: Can we fix the indentation of the following line to line up with this line? Similar to https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/s3/test_subcommands.py#L114-L115

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -132,6 +132,26 @@ def test_ls_command_with_no_args(self):
...
+                                    verify_ssl=None)
+        parsed_args = FakeArgs(dir_op=False, paths='s3:///random_nonsense',
```
justindho, line 139: Similar comment regarding indentation: https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/s3/test_subcommands.py#L116-L118

justindho (changes requested):
Hi @wingkwong, thanks for the contribution. Apologies for taking so long to get back to you. I just had a few small comments. Also, since it's been a few years since these changes were made, do you mind rebasing your changes onto `develop`?

wingkwong:
> Hi @wingkwong, thanks for the contribution. Apologies for taking so long to get back to you. I just had a few small comments. Also, since it's been a few years since these changes were made, do you mind rebasing your changes onto `develop`?

Sure. Will do.

aBurmeseDev:
Hi @wingkwong - this PR was recently brought up to team discussion and we'll need to do further investigation on this. Though we can't share a definite timeline, we now have a backlog item for the team and will post further updates here. In the meantime, I'm going to convert this to draft.

PR #4954, opened 2020-02-13
Make macOS PKG installer more flexible

stealthycoin:
This commit opens up the PKG installer to install the CLI
anywhere. During installation the user will now be prompted to install
in for all users, or just for the current user. Next it will show you a
choices screen with the install location displayed. Clicking on the
location allows you to select the `other...` option which will display a
file picker dialog to choose where the CLI is installed.

There are several caveats that I have not found suitable workarounds
for. First there can only be one global installation of the CLI. If you
do a second global installation in a different location, it will see
that you have a previous one and uninstall it first. If you choose the
same location it will simply overwrite the files.

This does not apply to current user only installations. If you choose
"install for me only" the CLI can be installed multiple times to
different locations. The caveat here is that it cannot be installed
outside the home directory. The installer app will not catch this
misconfiguration, and will throw an error after encountering a
permissions error.

Another thing to keep in mind about user directory installations is they
will not add a symlink to your PATH on your behalf. By default when you
do a global install, the PKG will install a symlink to the aws and
aws_completer binaries located in `/usr/local/bin`. Since a "for me
only" installation does not have permissions to that directory, and we
do not have a good way of guessing where you'd like us to put
symlinks (if any). We opt not to install any. This means if you want aws
on your PATH, you need to set up the symlinks yourself.

To see debug logs for an installation you can press cmd+L in the
installer GUI to open up the log pane. From there you can select
filtering options and save the log file after the installation is
done. The logs will also be written to `/var/log/install.log`.

In order to run an automated installation you can use the `installer`
command line tool. In order to run a global install (for all users of
this computer option in the GUI) invoke with `-target /` and `sudo`.

For example this will run a global install to the default location:
```
$ sudo installer -pkg AWS-CLI-Installer.pkg -target /
```

If you want to install to custom location it gets a little bit
tricker. This is required for a "me only" current user install since
the default location is `/user/local/` which won't work. To customize
the installation choices you need to use the `-applyChoiceChangesXML`
option. Which as the name implies takes an XML input file. Below I have
an example XML file that will change the install location to `~/aws`.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <array>
    <dict>
      <key>choiceAttribute</key>
      <string>customLocation</string>
      <key>attributeSetting</key>
      <string>/Users/stealthycoin/aws</string>
      <key>choiceIdentifier</key>
      <string>default</string>
    </dict>
  </array>
</plist>
```

It is important to note that the directory `/Users/stealthycoin/aws`
must already exist or the installer will exit with an error about a
malformed input document. Now we can invoke the `installer` command to
do a user only installation with this as the custom location. The main
differences are we do not need sudo to invoke it, and we change our
`-target` to `CurrentUserHomeDirectory`

```
$ installer -target CurrentUserHomeDirectory \
            -pkg AWS-CLI-Installer.pkg \
	    -applyChoiceChangesXML choices.xml
```

This will invoke the installer without elevated privileges, instruct it
you want to do a "me only" installation. And override the location with
our custom location `/Users/stealthycoin/aws`. Once this completes the CLI
should be installed at `/Users/stealthycoin/aws/aws-cli`. This
installation mode will not set up symlinks for you so it will not be
automatically added to your path.

### DIFF
--- a/macpkg/scripts/postinstall
+++ b/macpkg/scripts/postinstall
@@ -1,15 +1,28 @@
 #!/usr/bin/env bash
 EXE_NAME="aws"
 COMPLETER_EXE_NAME="aws_completer"
-sudo mkdir -p /usr/local/bin
-# Files created here are not stored in the bill of materials for the pkg.
-# This means we cannot find and remove them during an installation. We add a
-# custom .install-metadata file here which is read and used by our
-# scripts/installers/uninstall-mac-pkg to clean up any extra files created
-# here. If any new files are added, be sure to add them to the .install-metadata
-# file as well.
-sudo echo "/usr/local/aws-cli/.install-metadata" >> /usr/local/aws-cli/.install-metadata
-sudo ln -sf "/usr/local/aws-cli/$EXE_NAME" "/usr/local/bin/$EXE_NAME"
-sudo echo "/usr/local/bin/$EXE_NAME" >> /usr/local/aws-cli/.install-metadata
-sudo ln -sf "/usr/local/aws-cli/$COMPLETER_EXE_NAME" "/usr/local/bin/$COMPLETER_EXE_NAME"
-sudo echo "/usr/local/bin/$COMPLETER_EXE_NAME" >> /usr/local/aws-cli/.install-metadata
+PATH_TO_INSTALL="$2/aws-cli"
+METADATA_PATH="$PATH_TO_INSTALL/.install-metadata"
+SYMLINK_DIR="/usr/local/bin"
+ID=$(id -u)
+
+# Only install symlinks if we are installing with root permissions
+# this means we selected the install for everyone option.
+if [ "$ID" == "0" ]; then
+    echo "Global install detected. Installing symlinks in $SYMLINK_DIR."
+    sudo mkdir -p "$SYMLINK_DIR"
+    sudo echo "$METADATA_PATH" >> "$METADATA_PATH"
+
+    AWS_PATH="$PATH_TO_INSTALL/$EXE_NAME"
+    AWS_LINK="$SYMLINK_DIR/$EXE_NAME"
+    echo "Linking $AWS_LINK -> $AWS_PATH"
+    sudo ln -sf "$AWS_PATH" "$AWS_LINK"
+    sudo echo "$AWS_LINK" >> "$METADATA_PATH"
+
+
+    COMPLETER_PATH="$PATH_TO_INSTALL/$COMPLETER_EXE_NAME"
+    COMPLETER_LINK="$SYMLINK_DIR/$COMPLETER_EXE_NAME"
+    echo "Linking $COMPLETER_LINK -> $COMPLETER_PATH"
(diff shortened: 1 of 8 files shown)


macpkg/scripts/postinstall
```diff
@@ -1,15 +1,23 @@
...
+# Only install symlinks if we are installing with root permissions
+# this means we selected the install for everyone option.
+if [ "$ID" == "0" ]; then
+    echo "Global install detected. Installing symlinks in $SYMLINK_DIR."
+    sudo mkdir -p $SYMLINK_DIR
+
+    AWS_PATH="$PATH_TO_INSTALL/$EXE_NAME"
+    AWS_LINK="$SYMLINK_DIR/$EXE_NAME"
+    echo "Linking $AWS_LINK -> $AWS_PATH "
+    sudo ln -sf $AWS_PATH $AWS_LINK
```
kyleknap: We should encapsulate these with quotes. We should run this through `shellcheck` to make sure there are not anything else that we are missing.

macpkg/scripts/postinstall
```diff
@@ -1,15 +1,23 @@
...
+
+    COMPLETER_PATH="$PATH_TO_INSTALL/$COMPLETER_EXE_NAME"
+    COMPLETER_LINK="$SYMLINK_DIR/$COMPLETER_EXE_NAME"
+    echo "Linking $COMPLETER_LINK -> $COMPLETER_PATH"
+    sudo ln -sf $COMPLETER_PATH $COMPLETER_LINK
```
kyleknap: We should also start writing integration tests that actually run the installer in order to make sure that PKG functionality is not broken. Potential testing ideas would be:

* Test to make sure that this post-install script runs with sudo access.
* Be able to install it in a user location.

macpkg/scripts/postinstall
```diff
@@ -1,15 +1,23 @@
...
 EXE_NAME="aws"
 COMPLETER_EXE_NAME="aws_completer"
-sudo mkdir -p /usr/local/bin
-# Files created here are not stored in the bill of materials for the pkg.
-# This means we cannot find and remove them during an installation. We add a
-# custom .install-metadata file here which is read and used by our
-# scripts/installers/uninstall-mac-pkg to clean up any extra files created
-# here. If any new files are added, be sure to add them to the .install-metadata
-# file as well.
-sudo echo "/usr/local/aws-cli/.install-metadata" >> /usr/local/aws-cli/.install-metadata
```
kyleknap: Did we want to delete the `.install-metadata` file? This could be helpful when we add support for deleting installations via the installers.

kyleknap (commented):
Looks fine. Just capturing some of the offline feedback that I had.

kyleknap (approved):
Nice :ship:

PR #4959, opened 2020-02-13
Add docs for new retry config vars

jamesls:
From: https://github.com/boto/botocore/pull/1972


kyleknap (approved):
We should also add a changelog for this in the CLI.  Otherwise :ship:

Issue #4839, opened 2020-01-14
[v2] SSO login not launching browser


ericmillinder:
When using the packaged aws2 distributed at https://d1vvhvl2y92vvt.cloudfront.net/awscli-exe-linux-x86_64.zip, running the sso login fails to open the webbrowser. 

`$ /usr/local/aws-cli/v2/current/bin/aws2 sso login --profile <my profile> --debug`

```
2020-01-14 09:34:08,942 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/2.0.0dev3 Python/3.7.3 Linux/5.4.8-200.fc31.x86_64 botocore/2.0.0dev2
2020-01-14 09:34:08,942 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['sso', 'login', '--profile', 'appdev-admin', '--debug']
2020-01-14 09:34:08,943 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function add_timestamp_parser at 0x7fa442bde950>
2020-01-14 09:34:08,943 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function register_uri_param_handler at 0x7fa4434af048>
2020-01-14 09:34:08,943 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function add_binary_formatter at 0x7fa442b2a158>
2020-01-14 09:34:08,944 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function inject_assume_role_provider_cache at 0x7fa4434cdbf8>
2020-01-14 09:34:08,947 - MainThread - botocore.credentials - DEBUG - Skipping environment variable credential check because profile name was explicitly set.
2020-01-14 09:34:08,947 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function attach_history_handler at 0x7fa442d4dd08>
2020-01-14 09:34:08,947 - MainThread - botocore.hooks - DEBUG - Event session-initialized: calling handler <function inject_json_file_cache at 0x7fa442d8a730>
2020-01-14 09:34:08,957 - MainThread - botocore.loaders - DEBUG - Loading JSON file: /usr/local/aws-cli/v2/2.0.0dev3/dist/botocore/data/sso/2019-06-10/service-2.json
[...]

PR #4964, opened 2020-02-14
[v2] Fix issue with opening browser on moder linux distros

jamesls:
Don't use pyinstaller ld path for sso login

This ensures that the browser, launched as a subprocess, will load with the system libraries and not try to use any of the bundled libs that come from the pyinstaller exe.

As part of this change I added a commit that implements a context manager for swapping out the `LD_LIBRARY_PATH`.   For more info see:  https://pyinstaller.readthedocs.io/en/stable/runtime-information.html

I audited the rest of the places we use a subprocess (including transitive usage) and my initial testing shows that the `webbrowser` was the only place affected by this.  We should probably just swap out all of our usage of subprocess with one that swaps out `LD_LIBRARY_PATH` but that's more of a "best-practice"/future-proofing thing than it is fixing an actual issue.

Fixes https://github.com/aws/aws-cli/issues/4839

I tested this out on the same distro reported in #4839, Fedora 31.  I noticed that it took a long time to open a browser (the gif is sped up) but I can't tell if that's just xfce/firefox being slow or if it's related to me testing this in vmware.

sso-fedora31

### DIFF
--- a/.changes/next-release/bugfix-SSO-65835.json
+++ b/.changes/next-release/bugfix-SSO-65835.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "SSO",
+  "description": "Fix issue where browser would not open in modern linux distros (`#4839 <https://github.com/aws/aws-cli/issues/4839>`__)"
+}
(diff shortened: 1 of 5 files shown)


kyleknap (approved):
Looks good. We should also add a changelog. Otherwise :ship:

PR #4986, opened 2020-02-21
Replace pip with python -m pip

alanyee:
*Description of changes:*
Update install instructions to avoid these warnings:
```
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
```


stealthycoin:
Thanks for the contribution

PR #4993, opened 2020-02-24
[v2] Allow local installs for botocore

nateprewitt:
This change will allow passing an additional flag (`--botocore-path`) to each tox stage with a local path to botocore. With that, we can enable easier testing with unpublished versions and pin botocore for build consistency between tox stages.


nateprewitt:
Resolving in favor of #5055.

Issue #4949, opened 2020-02-12
aws-cli v2 bash auto_completer does not suggest `--cli-auto-prompt` as an option


ehammond:
The aws-cli v2 bash completion does not suggest `--cli-auto-prompt` as an option.

```
$ aws dynamodb delete-table --cli-<TAB><TAB>
--cli-connect-timeout  --cli-input-json       --cli-read-timeout

$ aws --version
aws-cli/2.0.0 Python/3.7.3 Linux/5.0.0-37-generic botocore/2.0.0dev4

$ echo $SHELL
/bin/bash
```

joguSD:
Confirmed. Marking this as a bug until we can generate a new autocompletion index that includes this.

thimslugga:
You can still install the aws-cli v2 with pip by specifying the github repo and v2 branch. I tested this on a a1 instance and it worked fine.

`
sudo su -
yum groupinstall -y "Development Tools"
yum install -y python3 python3-devel libffi-devel

python3 -m venv /opt/aws-cliv2/env --prompt aws-cliv2
source /opt/aws-cliv2/env/bin/activate

pip install git+https://github.com/boto/botocore.git@v2 --upgrade
pip install git+https://github.com/aws/aws-cli.git@v2 --upgrade

deactivate

test -f /usr/local/bin/aws && ln -s /opt/aws-cliv2/env/bin/aws /usr/local/bin/aws
test -f /usr/local/bin/aws_bash_completer && ln -s /opt/aws-cliv2/env/bin/aws_bash_completer /usr/local/bin/aws_bash_completer

aws --version
 aws-cli/2.0.0 Python/3.7.4 Linux/4.14.165-131.185.amzn2.aarch64 botocore/2.0.0dev4
`

stealthycoin:
Looks like this has been resolved

```
➜  aws dynamodb delete-table --cli-<TAB><TAB>
--cli-auto-prompt      --cli-binary-format    --cli-connect-timeout  --cli-input-json       --cli-input-yaml       --cli-read-timeout
```

PR #4999, opened 2020-02-25
[v2] Regenerated the autocomplete index

kyleknap:
Note that the index grew by a significant amount more than previous commits because we added  3-4 new global parameters. These get added to every single operation entry in the table. I think in the future though we should try avoid committing these to the repository given the size of the binary.


kyleknap:
I can confirm that it fixes this issue as well: https://github.com/aws/aws-cli/issues/4949

PR #5002, opened 2020-02-25
Fix issue with parsing a trailing empty line in codecommit helper

stealthycoin:
The spec for the IO format
https://git-scm.com/docs/git-credential#IOFMT allows for a trailing
newline character. Previously the codecommit credential helper would
crash if presented with input that ended in a blank line. Some third
party github libraries emit this format, causing them to be
incompatible with the CLI, but still spec-compliant.


kyleknap (approved):
:ship:

PR #5006, opened 2020-02-26
Added example for passing JSON string as Input parameter

tecmaverick:
*Issue #, if available: Many AWS users unable to figure out the syntax to pass JSON string to Input parameter.

*Description of changes:*
Added new example on how to pass JSON string as Input, when target is lambda function


kdaily:
Hi @tecmaverick,

Thanks for this PR. Since you made it, we have improved the documentation around quotations in strings, which is the difficulty with JSON text as a parameter value. Closing this out as this issue would apply in other operations that can take JSON as input.

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-quoting-strings.html#cli-usage-parameters-quoting-strings-containing

PR #5010, opened 2020-02-26
Add paramfile exception for apigatewayv2 integration-uri

stealthycoin:
This was added for the create-integration call but not for the
update-integration call. This PR brings them into sync.


kyleknap (approved):
:ship:

PR #5055, opened 2020-03-11
[v2] enable offline builds for deployments

nateprewitt:
We want to improve build reproducibility for the v2 installer and generate exact artifacts for what was used to create the build. This will allow us to optionally supply a curated set of tarballs/wheels to perform the installs without the need of an active PyPI server or internet connection.

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,0 +1,5 @@
+# Requirements we need to run our build jobs for the installers.
+# We create the separation for cases where we're doing installation
+# from a local dependency directory instead of requirements.txt.
+cryptography==2.8
+PyInstaller==3.5
--- a/scripts/install_deps.py
+++ b/scripts/install_deps.py
@@ -1,0 +1,74 @@
+import os
+
+from utils import cd, bin_path, run, virtualenv_enabled
+
+INSTALL_ARGS = "--no-binary :all: --no-build-isolation --no-cache-dir --no-index "
+PINNED_PIP_VERSION = '20.0.2'
+SETUP_DEPS = ("setuptools-", "setuptools_scm", "wheel")
+
+
+class InstallationError(Exception):
+    pass
+
+
+def get_package_tarball(package_dir, package_prefix):
+    package_filenames = sorted(
+        [p for p in os.listdir(package_dir) if p.startswith(package_prefix)]
+    )
+    if len(package_filenames) == 0:
+        raise InstallationError(
+            "Unable to find local package starting with %s prefix." % package_prefix
+        )
+    # We only expect a single package from the downloader
+    return package_filenames[0]
+
+
+def install_local_package(package_dir, package, pip_script="pip"):
+    with cd(package_dir):
+        run(
+            "%s install %s --find-links file://%s %s"
+            % (pip_script, INSTALL_ARGS, package_dir, package)
+        )
+
+
+def find_and_install_tarball(package_dir, package_prefix, pip_script="pip"):
+    tarball = get_package_tarball(package_dir, package_prefix)
+    install_local_package(package_dir, tarball, pip_script)
+
+
+def pip_install_packages(package_dir):
+    package_dir = os.path.abspath(package_dir)
+
+    # Setup pip to support modern setuptools calls
+    pip_script = os.path.join(os.environ["VIRTUAL_ENV"], bin_path(), "pip")
+    local_python = os.path.join(os.environ["VIRTUAL_ENV"], bin_path(), "python")
+
+    # Windows can't replace a running pip.exe, so we need to work around
+    run("%s -m pip install pip==%s" % (local_python, PINNED_PIP_VERSION))
+
+    # Install or update prerequisite build packages
+    setup_requires_dir = os.path.join(package_dir, "setup")
+    install_setup_deps(pip_script, setup_requires_dir)
+
+    find_and_install_tarball(package_dir, "awscli", pip_script)
+
+
+def install_setup_deps(pip_script, setup_package_dir):
+    # These packages need to be installed in this order before we
+    # attempt anything else in order to support PEP517 setup_requires
+    # specifications.
+
+    # We need setuptools >= 37.0.0 for setuptools_scm to work properly
+    for setup_dep in SETUP_DEPS:
+        find_and_install_tarball(setup_package_dir, setup_dep, pip_script)
+
+
+def install_packages(package_dir):
+    """Builds setup environment and installs copies of local packages for CLI"""
+
+    if not virtualenv_enabled():
+        raise InstallationError(
+            "Installation being performed outside of a virtualenv. Please enable before running."
+        )
+    else:
+        pip_install_packages(package_dir)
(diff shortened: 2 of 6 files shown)


scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+def cd(dirname):
+    original = os.getcwd()
+    os.chdir(dirname)
+    try:
+        yield
+    finally:
+        os.chdir(original)
+
+
+def run(cmd):
```
stealthycoin: Can we reuse `utils.run` here instead?

scripts/utils.py
```diff
@@ -47,3 +47,9 @@ def tmp_dir():
         yield dirname
     finally:
         shutil.rmtree(dirname)
+
+
+def _virtualenv_enabled():
```
stealthycoin: Lets make this public if it is used outside the utils module.

scripts/installers/make-exe
```diff
@@ -103,9 +104,24 @@ def main():
...
+    )
     args = parser.parse_args()
 
     output = os.path.abspath(args.output)
+    if args.src_dir:
+        print('Installing dependencies from local directory: %s' % args.src_dir)
+        install_packages(args.src_dir)
+    else:
+        run('pip install -r requirements.txt')
+        run('pip install .')
```
stealthycoin: Does this not need something like `pip_script` as well?

scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+        raise InstallationError(
+            "Installation being performed outside of a virtualenv. Please enable before running."
+        )
+    else:
+        try:
+            pip_install_packages(package_dir)
+        except Exception as e:
+            raise InstallationError(
+                "Error encountered when installing CLI dependencies", e
+            )
```
stealthycoin: Can we use [sys.exc_info](https://docs.python.org/3.7/library/sys.html#sys.exc_info) here to not lose the full traceback context.

scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+def _get_package_tarball(package_dir, package_prefix):
+    package_filenames = sorted(
+        [p for p in os.listdir(package_dir) if p.startswith(package_prefix)]
+    )
+    if len(package_filenames) == 0:
+        raise InstallationError(
+            "Unable to find local package starting with %s prefix."
+            % package_prefix
+        )
+    return package_filenames[-1]
```
stealthycoin: How come we are getting the last one from the list?

scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+
+def pip_install_packages(package_dir):
+    package_dir = os.path.abspath(package_dir)
+
+    pip_script = os.path.join(os.environ["VIRTUAL_ENV"], "bin", "pip")
+    run("%s install --upgrade pip" % pip_script)
+
+    setup_requires_dir = os.path.join(package_dir, "setup")
+    with cd(setup_requires_dir):
+        _install_setup_deps(pip_script, ".")
```
stealthycoin: Looks like we are both CDing to the directory and also passing current directory as an argument. Is there a reason for doing both? Can the directory be implied as current directory? Or can we not cd and pass in the directory path?

scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+            % package_prefix
+        )
+    return package_filenames[-1]
+
+
+def pip_install_packages(package_dir):
+    package_dir = os.path.abspath(package_dir)
+
+    pip_script = os.path.join(os.environ["VIRTUAL_ENV"], "bin", "pip")
+    run("%s install --upgrade pip" % pip_script)
```
stealthycoin: Given that we are trying to make a reproducible build, and pip has removed arguments before. What are your thoughts on locking this down as well. They also don't follow semver anymore so its difficult to trust a range.

scripts/install_deps.py
```diff
@@ -0,0 +1,112 @@
...
+import subprocess
+import sys
+
+from contextlib import contextmanager
+from utils import _virtualenv_enabled
+
+INSTALL_ARGS = "--no-binary :all: --no-build-isolation --no-cache-dir --no-index "
+
+
+class BadRCError(Exception):
```
stealthycoin: This is already defined in utils.

scripts/install_deps.py
nateprewitt: Moved cd over and am using the other util options now.

scripts/installers/make-exe
nateprewitt: The distinction here is one is done in a required active virtualenv and the default remains unchanged using whatever `pip` is available on the system path. We can change the `else` block to use the `pip_script` virtualenv as well, but then that means `make-exe` can't be run outside of one anymore.

scripts/install_deps.py
nateprewitt: This evolved from one of our other existing install scripts that was doing it this way. It looks like it was a 2.6 workaround James did that never got cleaned up. We can change to `0` without issue.

scripts/install_deps.py
nateprewitt: Yep agreed, let me find a version of 20 that's stable with the build infra and we'll just pin to that.

scripts/install_deps.py
nateprewitt: Yep, good catch, we can just pass in `setup_requires_dir` and avoid the `cd` step now. I think this originally had more logic that need us to be in the directory.

requirements-build.txt
```diff
@@ -0,0 +1,2 @@
+cryptography==2.8
```
kyleknap: Let's add a comment on why we need this file.

scripts/install_deps.py
kyleknap: Talked offline about it, but I'm fine with just not catching the exception here and let it propogate to help debugging. If we need to catch and run some logic on these later then we can go back and add this try except.

requirements-build.txt
stealthycoin: Agreed. It's not immediately obvious. In general I'd like to avoid what we have with the bundled installer where when we update our dependencies you have to magically know to update the make-bundle script file as well. It would be nice if we had some better way than a comment in the build requirements to signal where you need to update dependencies in strange places.

tox.ini
```diff
@@ -29,16 +28,14 @@ commands =
 [testenv:test-exe]
 basepython = python3.7
 deps =
-    -r{toxinidir}/requirements.txt
-    {toxinidir}
+    -r{toxinidir}/requirements-build.txt
```
kyleknap: Sorry did not notice this now till I tried it, but I think we need to add the lines from the `make-exe` script to the `test-exe` script:
```
run('pip install -r requirements.txt')
run('pip install .')
```
Specifically, we rely on `nose` to run the smoke tests and still import from the CLI/botocore codebase (even though we should probably update that in the future to require no dependencies)

kyleknap: Actually looking at both of these. I say we make the two changes:

* Do not make any changes to the `test-exe` environment.
* Do not specify `deps` for the `sign-exe`. Looking that one over I don't think it has any depdencies.

Given that the entire dependency closure for `requirements.txt` has stayed the same, I would not worry too much about these other environments. We just want to make sure the exe's can be built offline.

PR #5078, opened 2020-03-23
Add shorthand parser error for duplicate keys set on same object

stealthycoin:
Currently it is not an error in the shorthand parser if you specify a
second key/value pair with the same key as a previous one. Only one of
these values will actually be set, depending on the version of python
it could be either the first or second.

This is a disproportionately larger issue than it seems (not a big issue
for JSON) because of unfamiliarity with the shorthand parser
syntax. There are also commonly used ec2 APIs that accept a list of
objects in the form Key=key,Value=value Key=key2,Value=value2. This is
a list of two dictionaries each with two keys Key and Value.

A common issue people run into is to write the previous example as
Key=key,Value=value,Key=key2,Value=value2 separating the list items
with a comma as well as the dictionary entries. This is now one
dictionary with two keys named Key and two keys named
Value. Previously this would be accepted and silently drop the
duplicates. Instead now you will get the following error:
```
Error parsing parameter '--attributes': Second instance of key "Key" encountered for input:
Key=key1,Value=value1,Key=key2,Value=value2
                      ^
This is often because there is a preceding "," instead of a space.
```

### DIFF
--- a/awscli/shorthand.py
+++ b/awscli/shorthand.py
@@ -55,15 +55,8 @@ class _NamedRegex(object):
 
 
 class ShorthandParseError(Exception):
-    def __init__(self, value, expected, actual, index):
-        self.value = value
-        self.expected = expected
-        self.actual = actual
-        self.index = index
-        msg = self._construct_msg()
-        super(ShorthandParseError, self).__init__(msg)
 
-    def _construct_msg(self):
+    def _error_location(self):
         consumed, remaining, num_spaces = self.value, '', self.index
         if '\n' in self.value[:self.index]:
             # If there's newlines in the consumed expression, we want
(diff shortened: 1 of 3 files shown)


awscli/shorthand.py
```diff
@@ -149,9 +169,17 @@ def _parameter(self):
         # parameter = keyval *("," keyval)
         params = {}
         params.update(self._keyval())
+        last_index = self._index
         while self._index < len(self._input_value):
             self._expect(',', consume_whitespace=True)
-            params.update(self._keyval())
+            keyval = self._keyval()
+            key = list(keyval.keys())[0]
```
kyleknap: It might be good to add a note why we are doing this. It was definitely not obvious first looking at it.

PR #5107, opened 2020-04-03
[v2] Fix log tail bug when it doesn't pick up new log streams during polling

vz10:
*Issue #, if available:*

*Description of changes:*

Fix bug when we're using `logs tail` with `--follow` parameter. It wasn't picking newly created streams during polling.

### DIFF
--- a/awscli/customizations/logs/tail.py
+++ b/awscli/customizations/logs/tail.py
@@ -10,7 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from _collections import  defaultdict
+from collections import defaultdict
 from datetime import datetime, timedelta
 import re
 import time
@@ -294,6 +294,24 @@ class FollowLogEventsGenerator(BaseLogEventsGenerator):
             # KeyboardInterrupt propogate to the rest of the command.
             return
 
+    def _get_latest_events_and_timestamp(self, event_ids_per_timestamp):
+        if event_ids_per_timestamp:
+            # Keep only ids of the events with the newest timestamp
+            newest_timestamp = max(event_ids_per_timestamp.keys())
+            event_ids_per_timestamp = defaultdict(
+                set, {newest_timestamp: event_ids_per_timestamp[newest_timestamp]}
+            )
+        return event_ids_per_timestamp
+
+    def _reset_filter_log_events_params(self, fle_kwargs, event_ids_per_timestamp):
+        # Remove nextToken and update startTime for the next request
+        # with the timestamp of the newest event
+        if event_ids_per_timestamp:
+            fle_kwargs['startTime'] = max(
+                event_ids_per_timestamp.keys()
+            )
+        fle_kwargs.pop('nextToken', None)
+
     def _do_filter_log_events(self, filter_logs_events_kwargs):
         event_ids_per_timestamp = defaultdict(set)
         while True:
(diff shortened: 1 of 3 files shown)


awscli/customizations/logs/tail.py
```diff
@@ -294,20 +295,31 @@ def _filter_log_events(self, filter_logs_events_kwargs):
...
+                # reusing the newest timestamp of the received events to keep polling.
+                # This means it is possible that duplicate log events with same timestamp
+                # are returned back which we do not want to yield again.
                 # We only want to yield log events that we have not seen.
-                if event['eventId'] not in last_event_ids:
+                if event['eventId'] not in event_ids_per_timestamp[event['timestamp']]:
+                    event_ids_per_timestamp[event['timestamp']].add(event['eventId'])
                     yield event
-            last_event_ids = [event['eventId'] for event in response['events']]
+            if event_ids_per_timestamp.keys():
```
kyleknap, line 311: For dictionaries, we should be fine doing `if event_ids_pert_timestamp`. Specifically, the check is already based on the existence of keys for dictionaries.

awscli/customizations/logs/tail.py
```diff
@@ -294,20 +295,31 @@ def _filter_log_events(self, filter_logs_events_kwargs):
...
+                # Store ids of the events with the newest timestamp only
+                newest_timestamp = max(event_ids_per_timestamp.keys())
+                event_ids_per_timestamp = defaultdict(
```
kyleknap, line 314: Thinking it over, one thing that I'm not sure about is if CloudWatch logs preserve timestamp order across pages. Specifically, I'm concerned about if the second to last page has an event with a timestamp that is newer than the events in the last page. With this logic, we would store the events to skip in the newest timestamp of the last page. However, that means if we restart the filter with the timestamp from the last page, instead of the actual newest timestamp that is on the second to last page, we may reprint the event associated to the timestamp on the second to last set because we are not storing it in our set. 

Unless there is documentation to say otherwise, it may make sense to just keep a `most_recent_event_timestamp` that we check and update if needed for every event. We can then store any event that matches the timestamp in a set like we are doing in the `defaultdict` and then clear the set if we find a newer timestamp and store its events. Thoughts?

awscli/customizations/logs/tail.py
```diff
@@ -294,20 +295,31 @@ def _filter_log_events(self, filter_logs_events_kwargs):
...
+                    set, {newest_timestamp: event_ids_per_timestamp[newest_timestamp]}
+                )
             if 'nextToken' in response:
                 filter_logs_events_kwargs['nextToken'] = response['nextToken']
             else:
+                # Remove nextToken and update startTime for the next request
```
kyleknap, line 320: I'm not sure if this feedback will apply if we decide to update the startTime logic, but the `_do_filter_log_events` is getting a little difficult to read. It may be worth starting to break down the more complicated logic into smaller internal methods. For example with this code, I can see there being a:

* `_reset_filter_log_events_params()` which would do the logic of adding a new startTime and popping off the next token.
* `_update_latest_events_and_timestamp()` which would do the book keeping logic that keeps track of what the latest timestamp and events are.

There may be more methods we can come up with, but a good indication if it may be worth adding one is if you find yourself needing to add a comment on why a line of code is needed as often times the method name can do the documenting for us.

awscli/customizations/logs/tail.py
vz10: We will always keep only the newest timestamp in the dict and when we get to the next page we will add new timestamps to it and after that we again will keep only the newest one so we will always have the newest timestamp.

kyleknap: Right. Yeah I did not read the code right. I was thinking that the `event_ids_per_timestamp` was recreated for each page instead of reused. I think if we can break the various places we have bookkeeping logic into smaller internal methods, that will help a fair amount in following it.

PR #5157, opened 2020-04-25
[v2] Add environment variable to set encoding used for text files.

vz10:
*Issue #, if available:* 
5086

*Description of changes:* 
Add ``AWS_CLI_ENCODING`` environment variable to set encoding used for text files.

### DIFF
--- a/.changes/next-release/feature-encoding-19887.json
+++ b/.changes/next-release/feature-encoding-19887.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "encoding",
+  "description": "Add environment variable to set encoding used for text files. fixes `#5086 <https://github.com/aws/aws-cli/issues/5086>`__"
+}
--- a/.changes/next-release/feature-encoding-19887.json
+++ b/.changes/next-release/feature-encoding-19887.json
@@ -1,5 +1,5 @@
 {
-  "type": "feature",
+  "type": "enhancement",
   "category": "encoding",
   "description": "Add environment variable to set encoding used for text files. fixes `#5086 <https://github.com/aws/aws-cli/issues/5086>`__"
 }
--- a/tests/functional/test_paramfile.py
+++ b/tests/functional/test_paramfile.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
+import os
 
 from mock import patch, ANY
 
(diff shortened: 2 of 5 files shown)


tests/unit/test_compat.py
```diff
@@ -137,3 +139,15 @@ def test_ignore_signal_sigtstp(self):
         with ignore_user_entered_signals():
             self.assertEqual(signal.getsignal(signal.SIGTSTP), signal.SIG_IGN)
             os.kill(os.getpid(), signal.SIGTSTP)
+
+
+class TestGetPreferredEncoding(unittest.TestCase):
```
kyleknap, line 144: It would be interesting as well to see if we can come up with a functional test here as well to make sure the encoding is used when `file://` is used. Specifically, it would be nice if we added a test case in `tests/functional/test_paramfile.py` that takes in a file that has a specific encoding, we set the `AWS_CLI_FILE_ENCODING`, and call a CLI command that includes `file://` and make sure that the file was interpreted correctly or did not fail to read the file. I'm not entirely sure what encoding to use in testing this, but the main reason that I'm suggesting it is so that if we refactor the code at all to where the paramfile logic does not use the `compat.getprefferedencoding()`, we will still be able to catch if we are not respecting `AWS_CLI_FILE_ENCODING`.

.changes/next-release/feature-encoding-19887.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
```
kyleknap, line 2: I still think this would be better to label it as an `enhancement`. Specifically, `feature` typically warrants a larger feature for users to try out and thus a minor version bump of the CLI, but this addition is more to re-enable a work around that we had for v1, but with a different name. For `enhancement`s they only result in patch version bumps.

tests/functional/test_paramfile.py
```diff
@@ -67,3 +68,24 @@ def test_does_use_fileb_prefix(self):
...
+        with patch.dict(os.environ, {'AWS_CLI_FILE_ENCODING': 'utf-8'}):
+            self.assert_param_expansion_is_correct(
+                provided_param=param,
+                expected_param='經理'
+            )
+
+    def test_does_use_encoding_cp1251(self):
+        path = self.files.create_file('foobar.txt', '經理')
+        param = 'file://%s' % path
+        with patch.dict(os.environ, {'AWS_CLI_FILE_ENCODING': 'cp1251'}):
```
kyleknap, line 87: We should be able to set the env var in the `self.environ` property on the test as we already subclass from this [class](https://github.com/aws/aws-cli/blob/0933b44a9784913784773d21aad9ab5e72884190/awscli/testutils.py#L346) that patches it for us.

PR #5169, opened 2020-05-01
authorize-security-group-ingress.html examples example code update

amhammett:
The first example within https://docs.aws.amazon.com/cli/latest/reference/ec2/authorize-security-group-ingress.html#examples has double '\' which results in error "Unknown option: \" 


kdaily:
Hi @amhammett,

Thanks for bringing this to our attention. It looks like it has been fixed by our doc writers. Closing this out!

PR #5188, opened 2020-05-08
add user_agent_extra to 'ecs deploy' clients

allisaurus:
To identify calls made by `ecs deploy`

--

### DIFF
--- a/awscli/customizations/ecs/deploy.py
+++ b/awscli/customizations/ecs/deploy.py
@@ -16,7 +16,7 @@ import json
 import os
 import sys
 
-from botocore import compat
+from botocore import compat, config
 from botocore.exceptions import ClientError
 from awscli.compat import compat_open
 from awscli.customizations.ecs import exceptions, filehelpers
@@ -109,6 +109,8 @@ class ECSDeploy(BasicCommand):
     MSG_SUCCESS = ("Successfully deployed {task_def} to "
                    "service '{service}'\n")
 
+    USER_DATA_EXTRA = 'customization/ecs-deploy'
+
     def _run_main(self, parsed_args, parsed_globals):
 
         register_task_def_kwargs, appspec_obj = \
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/ecs/test_ecsclient.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # set global args
+        self.global_args = Namespace()
+        self.global_args.region = 'us-east-1'
+        self.global_args.endpoint_url = None
+        self.global_args.verify_ssl = None
+
+    def test_client_config(self):
+        self.test_client = ECSClient(
+            self.session, None, self.global_args, ECSDeploy.USER_DATA_EXTRA)
```
stealthycoin, line 37: I would change the `ECSDeploy.USER_DATA_EXTRA` constant here to be hardcoded in the test. The way it is now if the value got changed accidentally in the variable definition then the test would pass without complaint and the value would be wrong.

awscli/customizations/ecs/deploy.py
```diff
@@ -399,13 +402,17 @@ def validate_deployment_group(self):
 
 
 class ECSClient():
-    def __init__(self, session, parsed_args, parsed_globals):
+
+    def __init__(self, session, parsed_args, parsed_globals, user_data_extra):
         self._args = parsed_args
+        self._custom_config = config.Config(
+            **{'user_agent_extra': user_data_extra})
```
stealthycoin, line 409: Same comment as above.

awscli/customizations/ecs/deploy.py
```diff
@@ -109,22 +109,25 @@ class ECSDeploy(BasicCommand):
     MSG_SUCCESS = ("Successfully deployed {task_def} to "
                    "service '{service}'\n")
 
+    USER_DATA_EXTRA = 'customization/ecs-deploy'
```
stealthycoin, line 112: `USER_AGENT_EXTRA` here instead?

allisaurus:
thanks for taking a look @stealthycoin ! this has been updated, per feedback. Let me know if we need anything else to push this over the finish line.

PR #5199, opened 2020-05-13
Remove SimpleDB from preview

xssfox:
SimpleDB was released in 2008
SimpleDB support was add in aws-cli in 2014

I think 6 years is enough testing for a service - lets see SimpleDB come out of preview :)

### DIFF
--- a/awscli/customizations/preview.py
+++ b/awscli/customizations/preview.py
@@ -38,7 +38,6 @@ logger = logging.getLogger(__name__)
 
 
 PREVIEW_SERVICES = [
-    'sdb',
 ]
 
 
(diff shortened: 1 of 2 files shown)


xssfox:
Looks like some of the preview tests assume `sdb` is in preview. What's the best approach. We could move a lesser user service into preview like `workmail`

xssfox:
Some of the builds failed due to GitHub rate limiting. Not sure how to retrigger a build :)

kdaily:
Hi! SimpleDb is not suggested for use anymore - it's existence here is for legacy reasons.

xssfox:
Hi @kdaily 

I appreciate your time to look into this. Is SimpleDB deprecation an official AWS stance as the product page on the AWS website does not suggest such (https://aws.amazon.com/simpledb/) and creation of domains still works fine. Several other services no longer receive updates but are very much out of preview status on the CLI and remain functional. What work would be required to remove its preview flag?

Issue #4039, opened 2019-04-02
aws s3 sync and glacier deep archive - InvalidObjectState


nokiomanz:
Hi,

Following the announcement of glacier deep archives. I modified my lifecycle policy to first move to glacier and then move to deep archives. On the following day. Object older than x days had been moved to deep archive.

I have a script that run on that bucket that rsync file back locally for double backup. For object that were glacier I added to my command --ignore-glacier-warnings which made my script run fine and return without error. Like this
aws s3 sync --delete --ignore-glacier-warnings s3://mybucket/myfolder/ /localbackup/

With the addition of deep archive, I get the following message.
An error occurred (InvalidObjectState) when calling the GetObject operation: The operation is not valid for the object's storage class
My script still runs fine. But I get a bunch of this error for each object in glacier deep arvice.

I tried updating my awscli to latest version following this
https://docs.aws.amazon.com/cli/latest/userguide/install-bundle.html
Still with the same result.

If I run my rsync command on a specific object that is under glacier deep archives I don't get any message and it does nothing since it just can't.

swetashre:
@nokiomanz - Thank you for reporting the behavior. I am able to reproduce the issue and marking it as a bug . The sync command does not currently works with Deep Archive class storage objects. You are getting the error because Deep Archive objects stored in s3 cannot be downloaded/copied without restoring them.

vz10:
Hi @nokiomanz ,

In AWS CLI version 2.0.15 this issue has been fixed.

Could you please try it out and confirm that it resolved your issue.

PR #5205, opened 2020-05-15
Mute warnings for glacier deep archive objects

vz10:
*Issue #, if available:*  #4039

*Description of changes:*
 Mute deep glacier archive objects warnings when --ignore-glacier-warnings option enabled

### DIFF
--- a/tests/functional/s3/test_cp_command.py
+++ b/tests/functional/s3/test_cp_command.py
@@ -364,6 +364,20 @@ class TestCPCommand(BaseCPCommandTest):
         self.assertEqual(self.operations_called[0][0].name, 'HeadObject')
         self.assertIn('GLACIER', stderr)
 
+    def test_warns_on_deep_arhive_incompatible_operation(self):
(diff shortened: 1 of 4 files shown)


tests/functional/s3/test_cp_command.py
```diff
@@ -393,6 +393,21 @@ def test_turn_off_glacier_warnings(self):
         self.assertEqual(self.operations_called[0][0].name, 'HeadObject')
         self.assertEqual('', stderr)
 
+    def test_turn_off_glacier_warnings_for_deep_archive(self):
```
stealthycoin, line 396: We have added tests for the negative cases. What do you think about adding tests for the positive cases? I see there are a lot of warning tests for ensuring that a warning is present for `GLACIER`. To get full testing parity, we should probably add these for `DEEP_ARCHIVE` as well.

vz10: Sounds reasonable.

PR #3139, opened 2018-02-08
Append a newline to config files if needed

joguSD:
Fixes #2996

This will check if the config file ends in a newline, if not it appends one before writing a new section to avoid writing an invalid config file. This pretty much only happens on windows.

PR #5206, opened 2020-05-15
ensure there's a newline before a new profile

rich:
Fixes an error that corrupts config files when creating a new profile on a config file that didn't have a newline at the end of the file.

We have some scripts that make changes to our developers' local cli config file. In many cases, these files have been edited, one way or another, outside of `aws configure` and this seems to have resulted in there not being a newline at the end of the file.

The writer code doesn't ensure correctness of the file and when a new profile is passed to it, it blindly appends to the file. Instead of trying to make a larger change ensuring overall correctness of the config file, I tried to keep it minimal and check for this single error case.

*Description of changes:*
before writing a new section, the last character of the file is checked to see if it's a newline. If it's not, a newline is written. The rest of the method is unchanged.

### DIFF
--- a/awscli/customizations/configure/writer.py
+++ b/awscli/customizations/configure/writer.py
@@ -77,7 +77,17 @@ class ConfigFileWriter(object):
             pass
 
     def _write_new_section(self, section_name, new_values, config_filename):
-        with open(config_filename, 'a') as f:
+        with open(config_filename, 'a+') as f:
+            position = f.tell()
+
+            if position > 0:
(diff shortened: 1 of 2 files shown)


rich:
@kyleknap mind taking a look or directing it to the right person? Thanks!

kdaily:
Hi @rich, thanks for submitting and I apologize for the lengthy wait. This looks like a duplicate of #3139 - I'm going to check on the status of that and try to get it merged in soon.

PR #5207, opened 2020-05-15
Use IMDSv2 for autodiscovering EC2 region

vz10:
*Issue #, if available:* #4976

*Description of changes:*
Use IMDSv2, if it supported by instance, for auto-discovering EC2 region


tests/functional/test_clidriver.py
```diff
@@ -27,6 +27,17 @@
 from awscli.clidriver import create_clidriver
 
 
+class RegionCapture(object):
+    def __init__(self):
+        self._region = None
```
stealthycoin, line 32: Make this public since we are accessing it from outside the class (in the test).

PR #2322, opened 2016-12-13
Adding support for skipping automatic property uploads

dtimberlake:
This adds the ability to specify that a resource should not have its property automatically uploaded when run through `aws cloudformation package` if the property is not specified in the original template.

It also adds this functionality to `ApiGatewayRestApiResource`.

This is in response to #2320

PR #5222, opened 2020-05-21
Support exporting DefinitionS3Location in AWS::StepFunctions::StateMachine

shunjd:
*Issue #, if available:*
N/A

*Description of changes:*
AWS StepFunctions just released the new property `DefinitionS3Location` which is similar to [BodyS3Location](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-s3location.html) in AWS::ApiGateway::RestApi.

### DIFF
--- a/awscli/customizations/cloudformation/artifact_exporter.py
+++ b/awscli/customizations/cloudformation/artifact_exporter.py
@@ -331,7 +331,6 @@ class GraphQLSchemaResource(Resource):
     # Necessary to support Definition
     PACKAGE_NULL_PROPERTY = False
 
-
 class AppSyncResolverRequestTemplateResource(Resource):
     RESOURCE_TYPE = "AWS::AppSync::Resolver"
     PROPERTY_NAME = "RequestMappingTemplateS3Location"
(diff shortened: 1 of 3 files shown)


kamisbox:
Do you think support for Serverless StateMachine DefinitionUri can be added too?

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -331,7 +331,6 @@ class GraphQLSchemaResource(Resource):
     # Necessary to support Definition
     PACKAGE_NULL_PROPERTY = False
 
-
```
kyleknap, line 334: Let's keep the extra newline. For Python, classes should be separated by two newlines.

tests/unit/customizations/cloudformation/test_artifact_exporter.py
```diff
@@ -82,7 +83,7 @@ def test_all_resources_export():
             "class": GraphQLSchemaResource,
             "expected_result": uploaded_s3_url
         },
-
+
```
kyleknap, line 86: Let's remove the extra spaces that got added here.

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -417,6 +416,15 @@ class ServerlessRepoApplicationLicense(Resource):
     PACKAGE_NULL_PROPERTY = False
 
 
+class StepFunctionsStateMachineDefinitionResource(ResourceWithS3UrlDict):
+    RESOURCE_TYPE = "AWS::StepFunctions::StateMachine"
+    PROPERTY_NAME = "DefinitionS3Location"
+    BUCKET_NAME_PROPERTY = "Bucket"
+    OBJECT_KEY_PROPERTY = "Key"
+    VERSION_PROPERTY = "Version"
+    PACKAGE_NULL_PROPERTY = True
```
kyleknap, line 425: Should the `PACKAGE_NULL_PROPERTY` be set to `False` here? Specifically, it matches what the ApiGateway resource does: https://github.com/aws/aws-cli/blob/2998af688919ea8c770a225067dafcc8e6ab4d78/awscli/customizations/cloudformation/artifact_exporter.py#L378. Also based on the issue where it got added: https://github.com/aws/aws-cli/pull/2322, it seems needed as users could define the step function definitions inline in the template so we would not necessarily always want to be uploading to S3 and setting the S3 property.

shunjd: Yes, it should be `False`, thanks for pointing it out.

PR #5247, opened 2020-05-30
Improve Dockerfile to accept executables filename as an argument

vz10:
Improve Dockerfile and make_docker script to accept executables filename as argument.

*Issue #, if available:*

*Description of changes:*
To make it possible to use the same Dockerfile to build images from different sources for x86 and Arm platforms

### DIFF
--- a/scripts/installers/make-docker
+++ b/scripts/installers/make-docker
@@ -18,14 +18,14 @@ DIST_DIR = os.path.join(ROOT, 'dist')
 DEFAULT_EXE_ZIP = os.path.join(DIST_DIR, 'awscli-exe.zip')
 DEFAULT_DOCKER_OUTPUT = os.path.join(DIST_DIR, 'aws-cli-docker.tar')
 DEFAULT_TAGS = ['amazon/aws-cli']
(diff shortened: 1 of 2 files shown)


scripts/installers/make-docker
```diff
@@ -18,14 +19,14 @@ DIST_DIR = os.path.join(ROOT, 'dist')
...
-EXPECTED_DOCKERFILE_EXE_NAME = 'awscli-exe-linux-x86_64.zip'
 
 
 def make_docker(exe, output, tags):
     _ensure_docker_is_available()
     with tmp_dir() as build_context:
         _make_build_context(build_context, exe)
-        _docker_build(build_context, tags)
+        # Use ntpath.basename to extract filename for the path
+        _docker_build(build_context, tags, ntpath.basename(exe))
```
kyleknap, line 29: Is there a specific reason we are using `ntpath` instead of `os.path.basename` here? `ntpath` is specific to windows while `os.path` should behave appropriately based on the platform we are running on.

vz10: No specific reason. I used to use it before and it worked fine on both platforms (Linux and Win) so I decided to use it in case we'll start creating Win docker images in the future. But I think `os.path.basename` should do the same so at the of the day it doesn't matter.

PR #5254, opened 2020-06-02
Pin wcwidth dependency to <0.2.0

stealthycoin:


setup.py
```diff
@@ -31,6 +31,8 @@ def find_version(*file_paths):
     's3transfer>=0.3.0,<0.4.0',
     'ruamel.yaml>=0.15.0,<0.16.0',
     'prompt-toolkit>=2.0.0,<3.0.0',
+    'wcwidth<0.2.0',
+    'prompt-toolkit>=2.0.0,<3.0.0',
```
kyleknap, line 35: This looks like a duplicate line. We should probably remove the `prompt-toolkit>=2.0.0,<3.0.0` line above it.

Issue #5245, opened 2020-05-29
YAML format mismatch between aws-cli and CloudFormation service


iamed2:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution

A comment in a previous issue mentioned this but the issue itself was unrelated.

**Describe the bug**
`aws cloudformation package` generates YAML 1.2, but the CloudFormation service expects YAML 1.1. This results in `aws cloudformation package` changing the string `"yes"` to the bare word `yes`, which is then interpreted by the CloudFormation service as `true`.

**SDK version number**
`aws-cli 2.0.17`

**Platform/OS/Hardware/Device**
macOS 10.14.6

**To Reproduce (observed behavior)**
Attached is a simple template [`yes-repro.yaml`](https://github.com/aws/aws-cli/files/4704307/yes-repro.yaml.txt) that will produce the problematic output when run through `aws cloudformation package`. Deploying the packaged template will result in the custom resource failing, and deploying the template without packaging will successfully create the stack.

**Expected behavior**
I expect a `"yes"` to stay quoted in the output of `aws cloudformation package`. 

**Logs/output**
I've attached [template output](https://github.com/aws/aws-cli/files/4704312/yes-repro-packaged.yaml.txt) from `aws cloudformation package`.

**Additional context**
This was not an issue with awscli v1 because it generated YAML 1.1 and maintained the quoting of `"yes"`.

kdaily:
Hi @iamed2, 

This should be fixed now by #5257, available in [version 2.0.19](https://github.com/aws/aws-cli/blob/v2/CHANGELOG.rst#2019).

iamed2:
Thanks!

PR #5257, opened 2020-06-03
[v2] Always quote YAML boolean strings for cloudformation package

kyleknap:
CloudFormation treats unquoted values that are YAML 1.1 native booleans as booleans, rather than strings. In YAML 1.2, the only boolean values are "true" and "false" so values such as "yes" and "no" when loaded as strings are not quoted when dumped. This logic ensures that we dump these values with quotes so that CloudFormation treats these values as strings and not booleans.

Fixes: https://github.com/aws/aws-cli/issues/5245

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -70,9 +70,9 @@ def _add_yaml_1_1_boolean_resolvers(dumper_cls):
     # that we dump these values with quotes so that CloudFormation treats
     # these values as strings and not booleans.
     boolean_regex = re.compile(
-        '^(?:y|Y|yes|Yes|YES|n|N|no|No|NO|'
-        'true|True|TRUE|false|False|FALSE'
-        '|on|On|ON|off|Off|OFF)$', re.X
+        '^(?:yes|Yes|YES|no|No|NO'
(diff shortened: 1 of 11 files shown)


awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -60,13 +62,31 @@ def _dict_representer(dumper, data):
...
+    # when loaded as strings are not quoted when dumped. This logic ensures
+    # that we dump these values with quotes so that CloudFormation treats
+    # these values as strings and not booleans.
+    boolean_regex = re.compile(
+        '^(?:y|Y|yes|Yes|YES|n|N|no|No|NO|'
+        'true|True|TRUE|false|False|FALSE'
+        '|on|On|ON|off|Off|OFF)$', re.X
+    )
+    boolean_first_chars = list(u'yYnNtTfFoO')
+    dumper_cls.add_implicit_resolver_base(
```
vz10, line 78: What the difference between this resolver and default `implicit_resolvers` from `yaml/resolver.py`?

kyleknap: There is no difference in the regex in that diff. However, I had to make a change to remove the single character yes's and no's because CloudFormation does not support that as a valid boolean YAML value.

PR #5269, opened 2020-06-05
S3 cares not for gender.

TheSkorm:
Requester doesn't need to be male, female or even human to use this command. Updated to reflect.


stealthycoin:
Looks good.

PR #5295, opened 2020-06-16
Remove duplicate see help link for global params in waiter docs

stealthycoin:
The waiter help documentation does not include an options section, but
the handler for generating it does not override and suppress the
option_end documentation event. The default handler is used instead
and emits a global parameters link, the same as the previous
description section. This results in the waiter docs having two links
to the global parameters help page at the end of the description
section.

*Issue #, if available:*

*Description of changes:*


stealthycoin:
merged in cdf59d092977e5886e228089fdb8ad0d611cf20f

PR #5320, opened 2020-06-26
EKS get-token - Use STS regional endpoint

kokhang:
This change will make EKS get-token command to generate getCallerIdentity pre-signed URL using the STS regional endpoint.

Currently, this command invokes sts using the global us-east-1 endpoint.


kokhang:
@kyleknap  @stealthycoin, can you take a look at this PR? Thank you.

kokhang:
Closing this PR. In order to make STS use the regional endpoint, set  the env `AWS_STS_REGIONAL_ENDPOINTS=regional`.  See https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#aws-sts.

PR #5350, opened 2020-07-07
Add support for nested input structures to doc generation

vz10:
*Issue #, if available:*

*Description of changes:* Improve AWS CLI docs to include documentation strings for parameters in nested input structures

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -168,7 +168,9 @@ class CLIDocumentEventHandler(object):
             argument.argument_model, argument.cli_type_name)))
         doc.style.indent()
         doc.include_doc_string(argument.documentation)
-        self._document_enums(argument, doc)
+        if hasattr(argument, 'argument_model'):
+            self._document_enums(argument.argument_model, doc)
+            self._document_nested_structure(argument.argument_model, doc)
         doc.style.dedent()
         doc.style.new_paragraph()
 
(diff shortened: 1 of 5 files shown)


awscli/clidocs.py
```diff
@@ -186,18 +190,47 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
...
+                    doc.style.li('``%s``' % enum)
+                doc.style.end_ul()
+
+    def _document_nested_structure(self, model, doc, stack):
+        """Recursively documents parameters in nested structures"""
+        shape_name = getattr(model, 'name', None)
+        if not shape_name or shape_name in stack:
+            return
+        stack.append(shape_name)
+        if isinstance(model, StructureShape):
```
kyleknap, line 210: For identifying shape types, let's use the `type_name` attribute as it is a bit more consistently in this module and gives us a bit more freedom if we want to change class names in the future in botocore.

awscli/clidocs.py
```diff
@@ -186,18 +190,47 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
...
+            self._output_struct_data(model.members, doc, stack)
+        if isinstance(model, ListShape):
+            self._output_list_data(model, doc, stack)
+        stack.pop()
```
kyleknap, line 214: We should also account for map shapes. Here is how we do so in the output side: https://github.com/aws/aws-cli/blob/6a9c20c7b94f4bcab0ac577f7babfc35dd34cffc/awscli/clidocs.py#L561-L567

awscli/clidocs.py
```diff
@@ -186,18 +190,47 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
...
-            model = argument.argument_model
-            if isinstance(model, StringShape):
-                if model.enum:
-                    doc.style.new_paragraph()
-                    doc.write('Possible values:')
-                    doc.style.start_ul()
-                    for enum in model.enum:
-                        doc.style.li('``%s``' % enum)
-                    doc.style.end_ul()
+        if isinstance(model, StringShape):
```
kyleknap, line 195: It probably makes sense to document enums as well when they are in nested structures. Right now it looks like we do not account for that.

doc/source/_static/override.css
```diff
@@ -0,0 +1,3 @@
+code {
+    white-space: normal;
```
kyleknap, line 2: What was the reasoning for adding this override?

awscli/clidocs.py
```diff
@@ -186,18 +190,47 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
...
+        doc.style.end_ul()
+
+    def _output_list_data(self, data, doc, stack):
+        doc.style.start_ul()
+        doc.style.start_li()
+        doc.write('(%s) ' % data.member.type_name)
+        doc.parser.feed(data.member.documentation)
+        self._document_nested_structure(data.member, doc, stack)
+        doc.style.end_li()
+        doc.style.end_ul()
```
kyleknap, line 233: We should add tests to make sure we are correctly documenting nested input structures. I think we can added cases to this test file: https://github.com/aws/aws-cli/blob/develop/tests/unit/test_clidocs.py and add cases that the expected documentation is produced at least for (there may be more):
1. Nested structures
2. Nested maps
3. Nested lists
4. We handle shapes that can result in hitting circular recursions.

awscli/clidocs.py
```diff
@@ -186,18 +190,47 @@ def doc_relateditem(self, help_command, related_item, **kwargs):
...
+            self._output_struct_data(model.members, doc, stack)
+        if isinstance(model, ListShape):
+            self._output_list_data(model, doc, stack)
+        stack.pop()
+
+    def _output_struct_data(self, members, doc, stack):
+        doc.style.start_ul()
+        for name, shape in members.items():
+            doc.style.start_li()
+            doc.write('**%s** (%s) -- ' % (name, shape.type_name))
```
kyleknap, line 220: So one thing that I noticed is that the format of nested structure is different than the `Output` section. Is there a particular reason we chose one over the other? I think at the very least we should make sure the format is consistent (e.g. either both use the boto3 style of nested parameter documentation or the aws cli output nested style.) Honestly, it may make sense if we just pull out the logic that we use for the output and generalize it for use between the input and output nested structures as it looks like there is not much difference between the two. That way if we decide we do not like the format in general, we can update it once and it will apply to both places (e.g inputs and outputs) where we document nested structures.

awscli/clidocs.py
vz10: In all the cases I've seen enums were already documented in `.documentation` so it will be duplication

doc/source/_static/override.css
vz10: It doesn't brake long line of code if it doesn't have any EOL symbols in it which looks not good.

PR #5377, opened 2020-07-14
Fix s3 bug when request payer not passed to subcommands

vz10:
*Issue #, if available:*

*Description of changes:*
Fix `s3 cp` bug when `request-payer` parameter wasn't passed to subcommands

### DIFF
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -442,6 +442,16 @@ class RequestParamsMapper(object):
         cls._set_request_payer_param(request_params, cli_params)
 
     @classmethod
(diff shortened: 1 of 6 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -441,6 +441,16 @@ def map_get_object_params(cls, request_params, cli_params):
         cls._set_sse_c_request_params(request_params, cli_params)
         cls._set_request_payer_param(request_params, cli_params)
 
+    @classmethod
```
kyleknap, line 444: We should probably update the unit tests to add cases for these new methods. There should be a `TestRequestParamsMapperRequestPayer` class that can be updated in `tests/unit/customizations/s3/test_utils.py`

PR #5400, opened 2020-07-22
Add human readable os name

vz10:
*Issue #, if available:*

*Description of changes:*
Add human readable os name. 
* add distro as a dependency to detect linux distribution
* add tests

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -16,6 +16,8 @@ import sys
 import signal
 import logging
 
+import distro
+
 import botocore.session
 from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
@@ -106,18 +108,36 @@ def _get_distribution_source():
     return 'bundle'
 
 
-def _get_human_readable_os_name():
-    return 'os_name'
+def _get_distribution():
+    distribution = None
+    if platform.system() == 'Linux':
+        distribution = _get_linux_distribution()
+    return distribution
 
+
+def _get_linux_distribution():
+    linux_distribution = 'unknown'
+    try:
+        linux_distribution = distro.id()
+        version = distro.major_version()
+        if version:
+            linux_distribution += '.%s' % version
+    except Exception:
+        pass
+    return linux_distribution
 
 def _set_user_agent_for_session(session):
     session.user_agent_name = 'aws-cli'
     session.user_agent_version = __version__
-    session.user_agent_extra = '%s/%s.%s' % (
+    # user_agent_extra on linux will look like "rpm/x86_64.Ubuntu.18"
+    # on mac and windows like "sources/x86_64"
+    session.user_agent_extra = '%s/%s' % (
         _get_distribution_source(),
-        _get_human_readable_os_name(),
         platform.machine()
     )
+    linux_distribution = _get_distribution()
+    if linux_distribution:
+        session.user_agent_extra += ".%s" % linux_distribution
 
 
 class CLIDriver(object):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -11,12 +11,12 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import os
+import platform
 import sys
 import signal
 import logging
 
 import botocore.session
-from botocore import __version__ as botocore_version
 from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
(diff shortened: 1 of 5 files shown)


awscli/clidriver.py
```diff
@@ -106,18 +108,32 @@ def _get_distribution_source():
...
+    try:
+        distribution = distro.id()
+        version = distro.major_version()
+        if distribution and distribution != 'darwin':
+            linux = distribution
+            if version:
+                linux += '.%s' % version
+    except Exception:
+        linux = 'linux'
+    return linux
```
kyleknap, line 122: So for this logic, I was thinking something more like this:
```python
def _get_distribution(self):
    distribution = None
    if platform.system() == 'Linux':
        distribution = self._get_linux_distribution()
    return distribution

def _get_linux_distribution(self):
    linux_distribution = None
    try:
        linux_distribution = distro.id()
        version = distro.major_version()
         if version:
             linux_distribution += '.%s' % version
    except Exception:
        return linux_distribution
```
The main changes that I was thinking:
1. It is a bit more defensive for non-Linux systems in that we do not even call the `distro` library for Windows and Mac. While there was an if statement for mac in the PR, I'm not sure if we can guarantee that the if statement will still apply if we upgrade versions of the library and say they add more support for non-Linux systems.
2. If for whatever reason we cannot get a major version for the linux distribtution, we would still be able to include the distribution id.
3. If we cannot figure out the distribution in general, we should just not add it to the user agent. The general `linux` value is a bit of an ambiguous value if you are not familiar with the code to know that means the distribution is unknown. I would be fine with either just not including it or have a more specific value like `unknown`.

tests/functional/test_clidriver.py
```diff
@@ -103,6 +103,21 @@ def test_imds_region_is_used_as_fallback_with_v2_support(self):
         self.driver.main(['ec2', 'describe-instances'])
         self.assertEqual(capture.region, 'us-mars-2')
 
+    def test_session_user_agent_extra(self):
+        # if "distro" failed during getting linux version we'll have
+        # just "linux" in the user agent which shouldn't happen
+        self.assertFalse("linux" in self.session.user_agent_extra)
```
kyleknap, line 109: For this test, I would prefer if we just use the unit test to catch this as we can control when the exception is thrown. While we don't expect the exception to be thrown, it would be weird if a user was running the test on a machine that did throw the exception and now they cannot get the tests to pass.

Also if the distribution id had `linux` in the name, it would also cause the test to fail which also something we would not want.

tests/unit/test_clidriver.py
```diff
@@ -737,6 +739,32 @@ def test_idempotency_token_is_not_required_in_help_text(self):
...
+
+    def test_user_agent(self, *args):
+        driver = create_clidriver()
+        user_agent_extra_pattern = re.compile(
+            r'^.+/[a-z1-9_]+(\.[a-z]+)?(\.[1-9]+)?$'
+        )
+        self.assertIsNotNone(user_agent_extra_pattern.match(
+            driver.session.user_agent_extra))
+        # check that distro didn't fail
+        self.assertFalse('linux' in driver.session.user_agent_extra)
```
kyleknap, line 760: We may want to remove this assertion if the distro has `linux` in the name in case the distribution id has `linux`.

tests/unit/test_clidriver.py
```diff
@@ -737,6 +739,32 @@ def test_idempotency_token_is_not_required_in_help_text(self):
...
+        user_agent_extra_pattern = re.compile(
+            r'^.+/[a-z1-9_]+\.amzn\.1$'
+        )
+        self.assertIsNotNone(user_agent_extra_pattern.match(
+            driver.session.user_agent_extra))
+
+    def test_user_agent(self, *args):
+        driver = create_clidriver()
+        user_agent_extra_pattern = re.compile(
+            r'^.+/[a-z1-9_]+(\.[a-z]+)?(\.[1-9]+)?$'
```
kyleknap, line 755: Are there any distro id's that use numbers or capital letters? We may want to lax that constraint as well.

tests/functional/test_clidriver.py
```diff
@@ -103,6 +103,21 @@ def test_imds_region_is_used_as_fallback_with_v2_support(self):
         self.driver.main(['ec2', 'describe-instances'])
         self.assertEqual(capture.region, 'us-mars-2')
```
kyleknap, line 105: I don't think it was included in the parent commit, but we should have unit test for the machine platform and make sure `source` gets included in the user agent extra. We can also add that in another PR, but something I noticed.

tests/unit/test_clidriver.py
vz10: According to the `distro.id()` docs it always lowercase and the only think I missed is underscore.

tests/functional/test_clidriver.py
vz10: I'll add `source` check in the PR where I'll actually read it form file

PR #5410, opened 2020-07-24
Add distribution source to the user_agent

vz10:
*Issue #, if available:*

*Description of changes:*
Add distribution source to the user_agent

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import json
 import os
 import platform
 import sys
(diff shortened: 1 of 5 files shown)


awscli/clidriver.py
```diff
@@ -105,7 +107,13 @@ def create_clidriver():
 
 
 def _get_distribution_source():
-    return 'bundle'
+    metadata_file = os.path.join(os.environ['AWS_DATA_PATH'],
```
kyleknap, line 110: Instead of using `AWS_DATA_PATH` here, we should probably use the path to the `awscli/data` directory. The `AWS_DATA_PATH` indicates where to look for additional models to load so it is possible for users to set it and therefore this will not point to this file. Typically if we want to rely on the `AWS_DATA_PATH` env var, we should probably go through a botocore loader, but in this case I'm not sure if we need to be relying on it.

awscli/data/metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "distribution_source": "sources"
```
kyleknap, line 2: For the value here, let's use `source` instead of plural to indicate its a source build. It sort of reads weird by making it plural.

Issue #4079, opened 2019-04-16
EKS update-config overwrites user details when using multiple roles against the same cluster


jkpl:
I have an EKS cluster where I manage user access using IAM roles. This is roughly what my map roles would look like:

```yaml
- rolearn: arn:aws:iam::123456789123:role/k8s-admin
  username: admin
  groups:
    - system:masters
- rolearn: arn:aws:iam::123456789123:role/k8s-developer
  username: developer
  groups:
    - developers
```

To use `kubectl` with EKS, I need to assume the right role. Therefore, I'll have to provide the role ARN when updating the kubeconfig:

```shell
aws eks update-kubeconfig --name mycluster --role-arn arn:aws:iam::123456789123:role/k8s-admin --alias mycluster-admin
```

So far so good. However, when I try to add my second role to the same kubeconfig...

```shell
aws eks update-kubeconfig --name mycluster --role-arn arn:aws:iam::123456789123:role/k8s-developer --alias mycluster-developer
```

...the user information for the previous config update is overwritten. You can see this in the kubeconfig file:

```yaml
apiVersion: v1
clusters;
- cluster:
    certificate-authority-data: <base64>
    server: https://XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.YYY.us-east-1.eks.amazonaws.com
  name: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
contexts:
- context:
    cluster: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
    user: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
  name: mycluster-admin
- context:
    cluster: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
    user: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
  name: mycluster-developer
current-context: mycluster-developer
kind: Config
preferences: {}
users:
- name: arn:aws:eks:us-east-1:123456789123:cluster/mycluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - token
      - -i
      - eks
      - -r
      - arn:aws:iam::123456789123:role/k8s-developer
      command: aws-iam-authenticator
      env: null
```

Both contexts remain in the config file, but there's only one user which is attached to both contexts. There should be two users with distinct role ARNs.

I haven't verified this from the code, but I'm guessing the user is overwritten because the same user name is used for both updates.

swetashre:
[...]

PR #5413, opened 2020-07-26
Use 'alias' as kube-context username when arg is passed

sarmad-abualkaz:
*Issue #, if available:* #4079

*Description of changes:*
We have a similar issue to #4079, whereby admin users want to perform tests on eks clusters with both developer-like and admin permissions. We use alias's as our way to distinguish between contexts. 

With current setup, the following two commands executed one after the other would overwrite the user entry since its sharing the same user-name (i.e. the cluster ARN is always the user-name):
`aws eks update-kubeconfig --name mycluster --role-arn arn:aws:iam::123456789123:role/k8s-admin --alias mycluster-admin`
`aws eks update-kubeconfig --name mycluster --role-arn arn:aws:iam::123456789123:role/k8s-dev --alias mycluster-dev`

I feel like an easy win would be to use the `--alias` name as the user-name for the context, since the alias is anyway being used as the kube-context name. With this change if the `alias` arg isn't passed, the user-name would continue to be named as before, i.e. cluster-ARN.

@swetashre, since the issue is assigned to you, I was wondering what you think of this approach? 

Thanks. 

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -116,6 +116,7 @@ class UpdateKubeconfigCommand(BasicCommand):
         client = EKSClient(self._session,
                            parsed_args.name,
                            parsed_args.role_arn,
+                           parsed_args.alias,
                            parsed_globals)
         new_cluster_dict = client.get_cluster_entry()
         new_user_dict = client.get_user_entry()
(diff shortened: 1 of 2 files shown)


sarmad-abualkaz:
I'm just curious, is anyone actually looking at this PR? 

The PR has been open for more than 2 years and I have not received any feedback. If I'm being honest, I've never had such a bad 'open-source' collab experience. 

If this is not useful or no longer required, let me know and I can close it. If you have some feedback, also let me know and I can improve on the changes. 

Thanks

splichy:
@kdaily @nateprewitt can you look at this please?

yuxiang-zhang:
Kindly note that this feature has been added with the `--user-alias` flag, available since `v1.27.98` and `v2.11.6`.

PR #5415, opened 2020-07-27
Add customization command and subcommands to the user_agent

vz10:
*Issue #, if available:*

*Description of changes:*
Add customization command and subcommands to the user_agent

### DIFF
--- a/awscli/customizations/commands.py
+++ b/awscli/customizations/commands.py
@@ -177,7 +177,8 @@ class BasicCommand(CLICommand):
                     cli_argument.argument_model, value)
 
             setattr(parsed_args, key, value)
-
+        if hasattr(self._session, 'user_agent_extra'):
+            self._add_customization_to_user_agent()
         if hasattr(parsed_args, 'help'):
             self._display_help(parsed_args, parsed_globals)
         elif getattr(parsed_args, 'subcommand', None) is None:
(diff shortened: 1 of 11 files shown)


awscli/customizations/commands.py
```diff
@@ -309,6 +310,14 @@ def _raise_usage_error(self):
         ) % lineage
         raise ParamValidationError(error_msg)
 
+    def _add_customization_to_user_agent(self):
+        if 'command' in self._session.user_agent_extra:
```
kyleknap, line 314: So I'm a little hesitant on doing this if check because:

1) The command could technically have `command` in its name causing us to inappropriate append the last element in the lineage to it. I could not find an example of where this is a problem now but does not mean it could be a problem in the future.

2) The presence of `command` does not necessarily mean we are building up the command component of the user agent as some other logic could have come along and appended to the user agent.

I'm thinking it may make sense to change this to a regex that searches for `command/` and replaces its value component with the current command lineage (and not just appending the last element of the lineage). Then if the regex does not find anything then we just fall to the logic in the if statement.

awscli/customizations/commands.py
```diff
@@ -177,7 +177,8 @@ def __call__(self, args, parsed_globals):
                     cli_argument.argument_model, value)
 
             setattr(parsed_args, key, value)
-
+        if hasattr(self._session, 'user_agent_extra'):
```
kyleknap, line 180: I'm wonder if we should add this to the auto-generated commands too. One side effect of only doing it in this class is that you can get in weird cases like with the `wait` commands:
```
$ aws s3api wait bucket-exists --bucket foo --debug
```
where because the `wait` component is subclassed from a custom command class but the `bucket-exists` class subclasses from the `ServiceOperation` class, it misses the `bucket-exists` command in the user agent: `command/s3api.wait`

awscli/customizations/commands.py
vz10: I think it makes sense to look for `command` + `/` but I'd prefer not to user regexps substitutions because now user_agent is sort of immutable because we don't change anything in the current one just add to the end

kyleknap: That's fair on the regex idea.

PR #5427, opened 2020-07-28
Update docker and make-exe build to inject distribution source 

vz10:
*Issue #, if available:*

*Description of changes:*
Update docker and make-exe build to inject distribution source in the bundle

### DIFF
--- a/scripts/installers/make-docker
+++ b/scripts/installers/make-docker
@@ -8,7 +8,8 @@ from distutils.dir_util import copy_tree
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-from utils import run, tmp_dir, cd, BadRCError
+from utils import run, tmp_dir, cd, BadRCError, \
+                  extract_zip, update_metadata, save_to_zip
 
 
 ROOT = os.path.dirname(os.path.dirname(os.path.dirname(
@@ -18,6 +19,7 @@ DIST_DIR = os.path.join(ROOT, 'dist')
 DEFAULT_EXE_ZIP = os.path.join(DIST_DIR, 'awscli-exe.zip')
 DEFAULT_DOCKER_OUTPUT = os.path.join(DIST_DIR, 'aws-cli-docker.tar')
 DEFAULT_TAGS = ['amazon/aws-cli']
+DISTRIBUTION_SOURCE = 'docker'
 
 
 def make_docker(exe, output, tags):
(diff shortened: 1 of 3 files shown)


scripts/installers/make-exe
```diff
@@ -77,6 +80,16 @@ def copy_file(src, dst):
...
+def update_metadata(dirname):
+    print('Update metadata distribution source to %s' % DISTRIBUTION_SOURCE)
+    metadata_file = os.path.join(dirname, 'awscli', 'data', 'metadata.json')
+    with open(metadata_file) as f:
+        metadata = json.load(f)
+    metadata['distribution_source'] = DISTRIBUTION_SOURCE
+    with open(metadata_file, 'w') as f:
+        json.dump(metadata, f)
+
+
```
kyleknap, line 92: So for this change, I'm thinking we do the following:

1. Move and generalize this `update_metadata` logic  to `awscli/scripts/utils.py` and import it from there.
2. Remove the `sed` logic in the `Dockerfile` and use this shared utility in the `make-docker` by unzipping the exe, updating the metadata file in the exe, and then zipping it back up when used in the Docker build context.

I think this makes it a bit cleaner as it will allow us to isolate this metadata update logic in one place for the various installers.

scripts/installers/make-exe
```diff
@@ -23,6 +24,7 @@ PYINSTALLER_DIR = os.path.join(EXE_DIR, 'pyinstaller')
 ASSETS_DIR = os.path.join(EXE_DIR, 'assets')
 DEFAULT_OUTPUT_ZIP = 'awscli-exe.zip'
 AC_INDEX_INTERNAL_PATH = os.path.join('awscli', 'data', 'ac.index')
+DISTRIBUTION_SOURCE = 'bundle'
```
kyleknap, line 27: Instead of `bundle`, let's call this `exe` . Mainly `bundle` is a little confusing with v1 bundled installer.

scripts/installers/make-docker
```diff
@@ -40,10 +42,19 @@ def _ensure_docker_is_available():
 
 
 def _make_build_context(build_context_dir, exe):
+    _update_exe_metadata(exe)
```
kyleknap, line 45: We should probably update the metadata of the copied exe instead of the source exe to make sure we do not accidentally inject the metadata in an artifact we plan to use later on.

vz10: Yeah, it was temporary commit to run it through the pipeline

PR #5431, opened 2020-07-29
Add message about cli v2 to the bundle install and usage message

vz10:
*Issue #, if available:*

*Description of changes:*
Add message about cli v2 to the bundle install and usage message

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -16,6 +16,14 @@ from awscli.compat import six
 from difflib import get_close_matches
 
 
+AWS_CLI_V2_MESSAGE = (
+    'Note: AWS CLI version 2, the latest major version '
+    'of the AWS CLI, is now stable and recommended for general '
+    'use. For more information, see the AWS CLI version 2 '
+    'installation instructions at: https://docs.aws.amazon.com/cli/'
+    'latest/userguide/install-cliv2.html'
+)
+
 HELP_BLURB = (
     "To see help text, you can run:\n"
     "\n"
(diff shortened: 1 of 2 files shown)


awscli/argparser.py
```diff
@@ -16,6 +16,14 @@
 from difflib import get_close_matches
 
 
+AWS_CLI_V2_MESSAGE = (
+    '\033[93mNote: AWS CLI version 2, the latest major version '
```
stealthycoin, line 20: How portable is this escape code? (And one in the install script as well)

I would prefer we just display it plain, without color and have it show up the same everywhere. Rather than have yellow on some platforms, and a weird escape character on others.

awscli/argparser.py
```diff
@@ -25,7 +33,7 @@
 )
 USAGE = (
     "aws [options] <command> <subcommand> [<subcommand> ...] [parameters]\n"
-    "%s" % HELP_BLURB
+    "%s\n%s" % (HELP_BLURB, AWS_CLI_V2_MESSAGE)
```
stealthycoin, line 36: What do you think about moving this to the beginning of the error output? That way its the first thing they see, and then its not taking up space between their error message and the usage guidance.

awscli/argparser.py
vz10: I've tried it on Mac, Linux and windows and it worked fine. But I agree there for sure will be some platform where it won't work.

scripts/install
```diff
@@ -204,6 +204,11 @@ def main():
             print("You can now run: %s --version" % opts.bin_location)
         else:
             print("You can now run: %s --version" % real_location)
+        print('Note: AWS CLI version 2, the latest major version '
```
kyleknap, line 207: It might be nice to begin this with a newline. Otherwise, the usage and the note are too close together and a bit hard to read especially for smaller terminals.

awscli/argparser.py
```diff
@@ -24,8 +32,9 @@
     "  aws <command> <subcommand> help\n"
 )
 USAGE = (
+    "%s\n"
```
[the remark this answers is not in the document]
stealthycoin: I like the look better, I don't think a carriage return is that hacky, that's what its there for.

Issue #2828, opened 2017-09-13
Allow cloudformation deploy to accept a paramater file


akashisama:
When running the ```cloudformation deploy``` command it would be useful to be able to pass the parameters in as a file (to the ```--parameter-override``` parameter), as can be done with create-stack and update-stack.

Also requested here: https://github.com/awslabs/serverless-application-model/issues/111

cwgem:
Looking over the code it seems we'd just need something around here

https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/deploy.py#L178

that reads in a JSON file, maybe a quick sanity check, then passes it along to `deploy()`.

stealthycoin:
Marking as a feature request for cloduformation deploy.

@sanathkr thoughts?

philipakash:
waiting for this feature 👍

larryboymi:
I would also like the feature for `package` actions. Thanks! 👍

dumrauf:
Passing a parameters file via something like `--parameters` as in the `create-stack` and `update-stack` CF commands (and ideally using the same file contents syntax) would make developing CF templates a tad easier for me. Would love to see that feature.

anthonynovatsis:
+1 for this feature. Thanks!

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

cuifugang:
+1 for this feature. Thanks!

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

foobert:
[...]

PR #5443, opened 2020-08-04
Add support for cf and codepipeline parameters file to cf deploy

vz10:
Add support for cf and codepipeline parameters file to cf deploy command

*Issue #, if available:*  #2828

*Description of changes:*

### Current state:

The current format for “*—parameter-overrides*“ option is

`ParameterKey1=ParameterValue1 ParameterKey2=ParameterKey2 `

or JSON format

```
  [
    "ParameterKey1=ParameterValue1",
    "ParameterKey2=ParameterKey2"
  ]
```

### Proposed improvements:

Keep the shorthand format without changing but update the file parser to accept data in “CloudFromation format” and “CodePipeline format”, it means such JSON formats will be valid as well

* **“CloudFromation format”**

```
[
   {
        "ParameterKey": "ParameterKey1",
        "ParameterValue": "ParameterValue1",
    },
    {
        "ParameterKey": "ParameterKey2",
        "ParameterValue": "ParameterValue2",
    },
...
]
```

* **“CodePipeline format”**

```
[
    "Parameters": {
        "ParameterKey1": "ParameterValue1",
        "ParameterKey2": "ParameterValue2"`
    }

...
]
```

***Note: ***the logic of the “*—parameter-overrides*“ option won’t be changed, it means that it will expect only parameter’s keys and values and throw ValidationError if get unexpected keys (e. g. ***UsePreviousValue*** or ***ResolvedValue **which exist in CloudFormation format)*.. If parameter not set will be used previous value if stack is updated or default value for stack creation.

## Examples

Following command deploys template named ``template.json`` to a stack named ``my-new-stack``:

```
aws cloudformation deploy --template-file /path_to_template/template.json --stack-name my-new-stack --parameter-overrides BucketName=MyBucket Tag1value=awesomeBucket
```

```
template.json{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Description": "AWS CloudFormation Sample S3 bucket Template.",
  "Parameters": {
    "BucketName": {
      "Description": "Name of s3 Bucket",
      "Type": "String"
    },
    "Tag1value": {
      "Description": "Tag1 value",
      "Default": "value1",
      "Type": "String"
    },
    "Tag2value": {
      "Description": "Tag2 value",
      "Default": "value2",
      "Type": "String"
    }
  },
  "Resources": {
    "s3bucket": {
      "Type": "AWS::S3::Bucket",
      "Properties": {
        "BucketName": {"Ref": "BucketName"},
        "Tags": [
          {
            "Key": "Tag1",
            "Value": {"Ref": "Tag1value"}
          },
          {
            "Key": "Tag2",
            "Value": {"Ref": "Tag2value"}
          }
        ]
      }
    }
  }
}
```

You can do the same with inline json 

```
aws cloudformation deploy --template-file /path_to_template/template.json --stack-name my-new-stack --parameter-overrides '["BucketName=MyBucket", "Tag1value=awesomeBucket"]'
```

Or you can do the same using JSON file

```
aws cloudformation deploy --template-file /path_to_template/template.json --stack-name my-new-stack --parameter-overrides file://parameter.json 

parameter.json in CloudFormation format
[
   {
        "ParameterKey": "BucketName",
        "ParameterValue": "MyBucket",
    },
    {
        "ParameterKey": "Tag1value",
        "ParameterValue": "awesomeBucket",
    },
]

parameter.json in CodePipeline format
[
    "Parameters": {
        "BucketName": "MyBucket",
        "Tag1value": "awesomeBucket"
    }
]

```

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -33,6 +33,70 @@ from awscli.utils import write_exception
 LOG = logging.getLogger(__name__)
 
 
+class BaseParameterOverrideParser:
+    def can_parse(self, data):
+        # Returns true/false if it can parse
+        raise NotImplementedError('can_parse')
+
+    def parse(self, data):
+        # Return the properly formatted parameter dictionary
+        raise NotImplementedError('parse')
+
+
+class CodePipelineLikeParameterOverrideParser(BaseParameterOverrideParser):
+    def can_parse(self, data):
+        return isinstance(data, dict) and 'Parameters' in data
+
+    def parse(self, data):
+        # Parse parameter_overrides if they were given in
+        # CodePipeline params file format
+        # {
+        #     "Parameters": {
+        #         "ParameterKey": "ParameterValue"
+        #     }
+        # }
+        return data['Parameters']
+
+
+class CloudFormationLikeParameterOverrideParser(BaseParameterOverrideParser):
+    def can_parse(self, data):
+        for param_pair in data:
+            if ('ParameterKey' not in param_pair or
+                    'ParameterValue' not in param_pair):
+                return False
+            if len(param_pair.keys()) > 2:
+                return False
+        return True
+
+    def parse(self, data):
+        # Parse parameter_overrides if they were given in
+        # CloudFormation params file format
+        # [{
+        #   "ParameterKey": "string",
+        #   "ParameterValue": "string",
+        # }]
+        return {
+            param['ParameterKey']: param['ParameterValue']
+            for param in data
+        }
+
+
+class StringEqualsParameterOverrideParser(BaseParameterOverrideParser):
+    def can_parse(self, data):
+        return all(
+            isinstance(param, str) and len(param.split("=", 1)) == 2
+            for param in data
+        )
+
+    def parse(self, data):
+        result = {}
+        for param in data:
+            # Split at first '=' from left
+            key_value_pair = param.split("=", 1)
+            result[key_value_pair[0]] = key_value_pair[1]
+        return result
+
+
 class DeployCommand(BasicCommand):
 
     MSG_NO_EXECUTE_CHANGESET = \
@@ -356,42 +420,9 @@ class DeployCommand(BasicCommand):
 
         return parameter_values
 
-    def _validate_cf_params(self, param):
-        if 'ParameterKey' in param and 'ParameterValue' in param:
-            if len(param.keys()) > 2:
-                raise ParamValidationError(
-                    'CloudFormation like parameter JSON should have format '
-                    '[{"ParameterKey": "string", "ParameterValue": "string"}]')
-        return True
-
-    def _cf_param_parser(self, data):
-        # Parse parameter_overrides if they were given in
-        # CloudFormation params file format
-        # [{
-        #   "ParameterKey": "string",
-        #   "ParameterValue": "string",
-        # }]
-        try:
-            return {
-                param['ParameterKey']: param['ParameterValue']
-                for param in data if self._validate_cf_params(param)
-            }
-        except (KeyError, TypeError):
-            return None
-
-    def _codepipeline_param_parser(self, data):
-        # Parse parameter_overrides if they were given in
-        # CodePipeline params file format
-        # {
-        #     "Parameters": {
-        #         "ParameterKey": "ParameterValue"
-        #     }
-        # }
-        if isinstance(data, dict):
-            return data.get('Parameters', None)
-
     def _parse_input_as_json(self, arg_value):
-        # In case of inline json input it'll be list where json string
+        # In case of reading from file it'll be string and in case
+        # of inline json input it'll be list where json string
         # will be the first element
         if arg_value:
             if isinstance(arg_value, str):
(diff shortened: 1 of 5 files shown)


awscli/customizations/cloudformation/deploy.py
```diff
@@ -359,6 +356,68 @@ def merge_parameters(self, template_dict, parameter_overrides):
...
+        # Parse parameter_overrides if they were given in
+        # CodePipeline params file format
+        # {
+        #     "Parameters": {
+        #         "ParameterKey": "ParameterValue"
+        #     }
+        # }
+        if isinstance(data, dict):
+            return data.get('Parameters', None)
+
```
kyleknap, line 392: Left this off of my last review... Should we validate that it does not have any of the top-level keys like `Tags` and `StackPolicy`? I think it makes sense because the CLI is generally strict on making sure all parameters provided will be used.

awscli/customizations/cloudformation/deploy.py
```diff
@@ -359,6 +356,68 @@ def merge_parameters(self, template_dict, parameter_overrides):
...
+            )
+            # In case it was in deploy command format
+            # but was in file
+            if result is None:
+                return self.parse_key_value_arg(
+                    data,
+                    self.PARAMETER_OVERRIDE_CMD
+                )
+            return result
+        else:
```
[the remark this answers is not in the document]
vz10: I think it should work inline by default because I don't read from file directly it's done by parser.

awscli/customizations/cloudformation/deploy.py
vz10: It's discussible, if we take only Parameters key and just ignore the rest it'll allow customers to reuse the same files without charging

awscli/customizations/cloudformation/deploy.py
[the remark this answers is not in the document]
vz10: I'm not sure anyone will do it but to be consistent we need to allow it, I think

awscli/customizations/cloudformation/deploy.py
```diff
@@ -359,6 +356,79 @@ def merge_parameters(self, template_dict, parameter_overrides):
...
+                parsers, None
+            )
+            # In case it was in deploy command format
+            # but was in file
+            if result is None:
+                return self.parse_key_value_arg(
+                    data,
+                    self.PARAMETER_OVERRIDE_CMD
+                )
+            return result
```
kyleknap, line 423: So I'm not entire sure we should be relying on the old parsing behavior as the final attempt and not provide any helpful error message after it. Specifically playing around with this I got some really weird looking errors.

In this example, I accidentally misspelled `ParameterValue` as `ParameterVlue`
```
$ aws cloudformation deploy --stack-name test-cfn --template-file template.json --parameter-overrides "[{\"ParameterKey\":\"BucketName\",\"ParameterVlue\":\"kyleknap-test-cfn-bucket\"},{\"ParameterKey\":\"Tag1value\",\"ParameterValue\":\"awesomeBucket\"}]"

'dict' object has no attribute 'split'

``` 
Then even in the test case for just a generally bad structure, this message is confusing:
```
$ aws cloudformation deploy --stack-name test-cfn --template-file template.json --parameter-overrides '{"SomeKey":[{"RedundantKey": "RedundantValue"}]}'

['SomeKey'] value passed to --parameter-overrides must be of format Key=Value
```
In general, it might make sense to abstract this logic a bit into classes/this format:
```python
class BaseParameterOverrideParser:
     def can_parse(self, data):
           # Returns true/false if it can parse

    def parse(self, data):
         # Return the properly formatted parameter dictionary

# Example parser
class CodePipelineParameterOverrideParser(BaseParameterOverrideParser):
     def can_parse(self, data):
         return isinstance(data, dict) and 'Parameters' in data

     def parse(self, data):
         return data['Parameters']


# In the DeployCommand class
def parse_parameter_overrides(self, arg_value):
     # After checking if it is json:
     parsers = [
          CfnParameterOverrideParser(),  # maybe find better names here
          CodePipelineParameterOverrideParser(),
          StringEqualsParameterOverrideParser()  
     ]
     for parser in parsers:
         if parser.can_parse(data):
            return parser.parse(data)
     raise InvalidParameterOverrideFormat('Show helpful information about the accepted formats')
```
The advantages with this approach is that:

* We can avoid weird/unexpected errors from trying to parse a format that will not work.
* We can throw a more helpful error about generally incorrect parameter formats instead of relying on any one specific parser errors.

awscli/examples/cloudformation/deploy.rst
```diff
@@ -4,3 +4,40 @@ Following command deploys template named ``template.json`` to a stack named
...
+            "ParameterKey": "Key1",
+            "ParameterValue": "Value1",
+        },
+        {
+            "ParameterKey": "Key2",
+            "ParameterValue": "Value2",
+        },
+    ]
+
+    Note: Only ParameterKey and ParameterValue are expected keys, command will through an exception if receives unexpected keys (e.g. UsePreviousValue or ResolvedValue).
```
kyleknap, line 34: Another thing I noticed when I generated the docs we should use the reStructuredText note e.g. (`.. note::`) so that this wording does not get captured in the codeblock.

tests/functional/cloudformation/test_deploy.py
```diff
@@ -113,33 +118,15 @@ def test_parameter_overrides_from_inline_cf_like_json(self):
...
+        cf_like_json = [
+            {'ParameterKey': 'Key1', 'ParameterValue': 'Value1'},
+            {'ParameterKey': 'Key2', 'ParameterValue': 'Value2'}
+        ]
+        path = self.create_json_file('param.json', cf_like_json)
         self.command += ' --parameter-overrides file://%s' % path
         self.run_cmd(self.command)
         self._assert_parameters_parsed()
 
-    def test_parameter_overrides_from_invalid_cf_like_json_file(self):
```
kyleknap: We should probably keep this test but just make sure we assert that the new error message is coming back.

awscli/customizations/cloudformation/deploy.py
```diff
@@ -405,22 +435,18 @@ def parse_parameter_overrides(self, arg_value):
...
-                return self.parse_key_value_arg(
-                    data,
-                    self.PARAMETER_OVERRIDE_CMD
-                )
-            return result
+            for parser in parsers:
+                if parser.can_parse(data):
+                    return parser.parse(data)
+            raise ParamValidationError(
+                'JSON passed to --parameter-overrides must be in of '
```
kyleknap, line 446: Instead of `must be in of` let's change it to `must be one of`

awscli/customizations/cloudformation/deploy.py
```diff
@@ -359,6 +420,41 @@ def merge_parameters(self, template_dict, parameter_overrides):
 
         return parameter_values
 
+    def _parse_input_as_json(self, arg_value):
+        # In case of inline json input it'll be list where json string
+        # will be the first element
+        if arg_value:
+            if isinstance(arg_value, str):
+                return json.loads(arg_value)
```
kyleknap, line 428: Then is this the case where it got loaded from a file? If so, it might be good to include that in the comment.

awscli/examples/cloudformation/deploy.rst
```diff
@@ -4,3 +4,42 @@ Following command deploys template named ``template.json`` to a stack named
...
+        },
+        {
+            "ParameterKey": "Key2",
+            "ParameterValue": "Value2"
+        }
+    ]
+
+.. note::
+
+ Only ParameterKey and ParameterValue are expected keys, command will through an exception if receives unexpected keys (e.g. UsePreviousValue or ResolvedValue).
```
kyleknap, line 36: Typo: `through` -> `throw`

PR #5472, opened 2020-08-12
Add bulk profiles update to configure set command

vz10:
*Issue #, if available:*

*Description of changes:*
Add additional option to `configure set` command that accepts regexp pattern and updates all the profiles that match the pattern


vz10:
In the current implementation if we do something like 

`aws configure set plugins.foo.bar value`

it will add to the default profile section

`plugins.foo.bar=value`

I preserved this logic but I'm not sure it's expected behaviour

KapitalEarth:
Value and command configured Apple gum end. <סּםיי-Good.>

PR #5477, opened 2020-08-13
Add no-cli-pager option to disable output to pager

vz10:
*Issue #, if available:*

*Description of changes:*
Add no-cli-pager option to disable output to pager

### DIFF
--- a/.changes/next-release/enhancement-pager-93694.json
+++ b/.changes/next-release/enhancement-pager-93694.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "pager",
+  "description": "Add ``--no-cli-pager`` command line option to disable output ot pager"
+}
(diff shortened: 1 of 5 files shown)


awscli/argparser.py
```diff
@@ -59,6 +60,11 @@ def choices(self, val):
         pass
 
 
+class DisablePagerAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        os.environ['AWS_PAGER'] = ''
```
kyleknap, line 65: Ah so I'd recommend following this PR for implementing this flag: https://github.com/aws/aws-cli/pull/4748

Specifically we should define it in `awscli/data/cli.json` so the argument gets auto generated and we can then register a handler that set the session variable for the pager.

awscli/argparser.py
```diff
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import argparse
+import os
```
kyleknap, line 14: We can probably remove this line.

PR #5520, opened 2020-09-04
Fix sts parameter example

my3:
duration-seconds is the correct option to use as per https://docs.aws.amazon.com/cli/latest/reference/sts/get-session-token.html
Tested with aws-cli 2.0.0

*Issue #, if available:*

*Description of changes:*


kdaily (approved):
Thanks, @my3!

kyleknap (approved):

PR #5538, opened 2020-09-14
Move bcdoc to awscli

nateprewitt:
This PR will migrate the vendored copy of bcdoc from botocore into awscli. This is the first step to removing docutils as a direct dependency of botocore going forward.

The primary purpose of bcdoc in the CLI is to generate help documentation in the terminal with the `aws help` command. Without docutil, we're unable to render that properly on Windows.


kyleknap (commented):
Looks good. The only suggestion that I had is let's hoist `bcdoc` out of its own `docs` directory. Specifically, it is a little weird have the `docs` package when none of the other doc related logic like examples, topics, and help logic do not live in it. In the future, we may consolidate this into a single directory but want to hold off doing that for this change.

nateprewitt:
@kyleknap yep, that seems reasonable. I've pulled bcdoc up to the top level for `awscli`.

kyleknap (approved):
Looks good! :ship:

PR #5543, opened 2020-09-16
Fix README rendering on PyPI.

methane:
*Description of changes:*

README markdown is not rendered now on PyPI: https://pypi.org/project/awscli/
Add long_description_content_type to fix the rendering issue.

See https://packaging.python.org/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata

---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setup_options = dict(
     version=find_version("awscli", "__init__.py"),
     description='Universal Command Line Environment for AWS.',
     long_description=read('README.md'),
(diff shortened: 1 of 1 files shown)


kdaily (changes requested):
Hi @methane, 

Thanks for catching this! However you opened this against `master`, hence why the Travis tests are failing. Can you please open it against `develop`? We tried to change the base here in GitHub but it introduced too many other commits. Thanks!

methane:
OK, I created #5551 for develop branch.

PR #5609, opened 2020-10-06
Prompt history mode

vz10:
*Issue #, if available:*

*Description of changes:*
Add history to the autoprompt mode

### DIFF
--- a/awscli/customizations/autoprompt/history.py
+++ b/awscli/customizations/autoprompt/history.py
@@ -1,0 +1,80 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import json
+import logging
+import os
+
+from prompt_toolkit.completion import Completion, Completer
+from prompt_toolkit.history import FileHistory
+
+from awscli.autocomplete.completer import CompletionResult
+from awscli.autocomplete.filters import fuzzy_filter
+
+
+LOG = logging.getLogger(__name__)
+
+
+class HistoryDriver(FileHistory):
+    HISTORY_VERSION = 1
+
+    def __init__(self, filename, max_commands=500):
+        super().__init__(filename)
+        self._max_commands = max_commands
+
+    def load_history_strings(self):
+        try:
+            with open(self.filename, 'r') as f:
+                commands = json.load(f).get('commands', [])
+            return reversed(commands)
+        except Exception:
+            LOG.debug('Exception on loading prompt history:', exc_info=True)
+            return []
+
+    def store_string(self, string):
+        history = {'version': self.HISTORY_VERSION, 'commands': []}
+        try:
+            if os.path.exists(self.filename):
+                with open(self.filename, 'r') as f:
+                    history = json.load(f)
+            history['commands'].append(string)
+            history['commands'] = history['commands'][-self._max_commands:]
+            with open(self.filename, 'w') as f:
+                json.dump(history, f)
+        except Exception:
+            LOG.debug('Exception on storing prompt history:', exc_info=True)
+
+
+class HistoryCompleter(Completer):
+
+    def __init__(self, buffer):
+        self.buffer = buffer
+        self.working_lines = buffer.history.get_strings()
+
+    def get_completions(self, document, *args):
+        found_completions = set()
+        completions = []
+        current_line = document.current_line_before_cursor.lstrip()
+
+        for line in self.working_lines:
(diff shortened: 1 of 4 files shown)


awscli/customizations/autoprompt/history.py
```diff
@@ -0,0 +1,71 @@
...
+import os
+
+from prompt_toolkit.completion import Completion, Completer
+from prompt_toolkit.history import FileHistory
+
+from awscli.autocomplete.completer import CompletionResult
+from awscli.autocomplete.filters import fuzzy_filter
+
+
+class HistoryDriver(FileHistory):
```
kyleknap, line 22: So I dived into this class a bit more. I'm thinking we just do our own implementation on top of the base `History` class. Mainly, I think I'd like to see us control our format/schema for recording commands. Specifically, I was thinking we just make this history file a json document (I can go either way on the order of commands):
```
{
    "version": "1",
    "commands": [
        "<oldest-command>"
        ...
        "<2nd-most-recent-command>"
        "<most-recent-command>",
     ]
}
```
I like this approach because:
1. We are in complete control of how completions are saved in the file. Right now our heuristic for saving max lines is based on what we expect prompt toolkit to be saving, but I'm not sure if there is any guarantee on how that is saved.
2. We can add a version string to the file so that we can disambiguate if we ever make a breaking change to the history format and need to account for different format versions.
3. We can load the contents of the history as a list making it a bit easier to interpret/work with.

Thoughts?

awscli/customizations/autoprompt/history.py
```diff
@@ -0,0 +1,71 @@
...
+            super().store_string(string)
+        except Exception:
+            pass
+
+
+class HistoryCompleter(Completer):
+
+    def __init__(self, buffer):
+        self.buffer = buffer
+        self.working_lines = buffer.history.get_strings()
```
kyleknap, line 52: We should probably just make this a lazy load as we are loading it from a file:
```python
    def __init__(self, buffer):
        self.buffer = buffer
        self._working_lines = None

   @property
    def working_lines(self):
        if self._working_lines is None:
            self._working_lines = self.buffer.history.get_strings()
        return self._working_lines
```

tests/unit/customizations/autoprompt/test_history.py
```diff
@@ -0,0 +1,110 @@
...
+    def setUp(self):
+        self.strings = [
+            'chime associate-phone-number-with-user',
+            'opsworks-cm associate-node',
+            'deploy add-tags-to-on-premises-instances',
+            's3 ls',
+        ]
+        self.buffer = mock.Mock(spec=Buffer)
+        history = mock.Mock(spec=History)
+        history.get_strings = mock.Mock()
```
kyleknap, line 39: Do we need to assign `mock.Mock()` here? The method should already be mocked out and can just immediately assign a `return_value` to it.

kyleknap (commented):
Looks good. It works well. I also like that now I can just use the up and down arrows to give me history as well. My main comment was about the schema of the history file. Let me know what you think

awscli/customizations/autoprompt/history.py
vz10: Prompt toolkit starts loading strings in background on buffer creation so this laziness won't change anything

awscli/customizations/autoprompt/history.py
```diff
@@ -0,0 +1,77 @@
...
+        try:
+            if os.path.exists(self.filename):
+                with open(self.filename, 'r') as f:
+                    history = json.load(f)
+            history['commands'].append(string)
+            history['commands'] = history['commands'][-self._max_commands:]
+            with open(self.filename, 'w') as f:
+                json.dump(history, f)
+        except Exception:
+            pass
```
kyleknap, line 51: So one thing that I think that would be valuable here is to log what the exception was here in the case we are trying to trouble shoot why history is not showing up from a bug report. However, I also noticed that we do not enable debug logs before we enter in auto-prompt mode so for this case, it would not output the message if there was an issue.

I think for this PR we should instantaite a logger similar to what we do in other modules: https://github.com/aws/aws-cli/blob/63c1e9182c3411a1b1b02baff80b33bc9ecdd8de/awscli/customizations/commands.py#L19 and log the exception. Then in a follow up PR we need to figure out the story around debugging this because this will be important before we release this feature.

kyleknap (commented):
Looks good. I just had two more follow up comments.

awscli/customizations/autoprompt/history.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            if os.path.exists(self.filename):
+                with open(self.filename, 'r') as f:
+                    history = json.load(f)
+            history['commands'].append(string)
+            history['commands'] = history['commands'][-self._max_commands:]
+            with open(self.filename, 'w') as f:
+                json.dump(history, f)
+        except Exception as e:
+            LOG.debug(f'Exception on storing prompt history - {e}')
```
kyleknap, line 54: Oh instead we should just use `exc_info` as it will log the Traceback to. So something like this: https://github.com/aws/aws-cli/blob/ab41dc3429137dd964670e9bc342590ca4b7bf2b/awscli/customizations/sso/logout.py#L92

kyleknap (commented):
Just a small comment about logging exceptions. Otherwise, let's rebase this off the tracking branch to get CI to pass and then we can merge it.

kyleknap (approved):
Awesome! :ship:

PR #5613, opened 2020-10-07
[v2] Initial implementation for revamped autoprompt mode

vz10:
*Issue #, if available:*

*Description of changes:*


tests/functional/autoprompt/test_doc.py
```diff
@@ -0,0 +1,178 @@
...
+    def setUp(self):
+        self.driver = create_clidriver()
+        self.docs_getter = DocsGetter(self.driver)
+
+    def test_get_service_command_docs(self):
+        command_text = 'ec2'
+        actual_docs = self.docs_getter.get_docs(command_text)
+        expected_docs = (
+            '\n\nAmazon Elastic Compute Cloud (Amazon EC2) provides secure and'
+            '\nresizable computing capacity in the AWS cloud. Using '
```
kyleknap, line 31: Ok so for all of these tests where we assert the exact sentences is really fragile because the documentation is liable to changing and breaking our tests suddenly. I'd recommend updating these to either:

1. Assert something very minimal but identifiable for that command like `Amazon EC2` and `Available Commands` are in the `expected_docs`
2. Create our own service model and keep our really specific assertions on that since we can guarantee that the service model will not change.

kyleknap (commented):
Reading through the code, I just had a few small suggestions that I think we can make now before we merged it. For the one larger one related to fuzzy searching for server side completions, we can address that in a follow up PR. 

When I tried it out, it seemed like it is working fine, but I do think I found a regression where if you first set:
```
aws configure set cli_auto_prompt on
```
and then type:
```
aws
```
and enter in auto-prompt mode:
```
dynamodb describe-table --table-name DemoTable
```
We get the error message as if we just typed `aws`:
```

  aws help
  aws <command> help
aws: error: the following arguments are required: command
```
Specifically, it is running the command that you initially inputted instead of the one you build up in the auto-prompt mode.

kyleknap (approved):
Looks good and is working well! :ship:

Issue #484, opened 2013-11-13
Support simple specification of value and key=value parameters


ehammond:
1. Instead of:
   
   ```
   --instance-initiated-shutdown-behavior '{"Value":"terminate"}'
   ```
   
   let me alternatively say:
   
   ```
   --instance-initiated-shutdown-behavior 'terminate'
   ```
2. Instead of:
   
   ```
   --tags '[{"Key":"project","Value":"fun"}]'
   ```
   
   or:
   
   ```
   --tags Key=project,Value=fun
   ```
   
   let me alternatively say:
   
   ```
   --tags 'project=fun'
   ```
3. And so on, for all simple value and key/value parameters...

jlordiales:
For what is worth, I've created this simple Ruby script (https://github.com/jlordiales/easy_cf_parameters) to read a YAML file with the easy to understand "key:value" syntax and generate the format expected by CF.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

twitu:
I am using cli version `aws-cli/1.16.243 Python/3.5.2 Linux/4.15.0-64-generic botocore/1.12.233` and `--parameters` does not support the simpler method for specifying key value pairs. Is there any timeline for this?

kdaily:
For 2 and 3, the CLI supports shorthand syntax:

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-shorthand.html

[...]

PR #5615, opened 2020-10-08
Use shorthand syntax for instance type

Fydon:
Possibly related to #484 

Simplify the [documentation for modify-instance-attribute](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/modify-instance-attribute.html) by using shorthand syntax.


Fydon:
I'm not sure what the next step is here. It's a one line change to documentation so I don't get why this change is blocked. Do I just need to rebase?

kdaily:
Hi @Fydon, thanks for the contribution. We try to maintain consistency for the example syntax. The more explicit example is easier to compare to the API. Given how short this example is (there is only one parameter) the change to using shorthand does not provide much more clarity.

I do realize that the specifications for how to write examples are not available in our contributing guide, and I am going to work with our doc writing team to make that available.

PR #5622, opened 2020-10-09
Move error handling from clidriver

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -59,13 +59,8 @@ from awscli.autoprompt.core import (
     AutoPrompter, AutoPromptDriver,
     validate_auto_prompt_args_are_mutually_exclusive
 )
-from awscli.errorhandler import (
-    ParamValidationErrorsHandler, ClientErrorHandler,
-    ConfigurationErrorHandler, NoRegionErrorHandler,
-    NoCredentialsErrorHandler, InterruptExceptionHandler,
-    UnknownArgumentErrorHandler, GeneralExceptionHandler,
-    ChainedExceptionHandler
- )
+from awscli.errorhandler import construct_cli_error_handlers_chain
+
 
 LOG = logging.getLogger('awscli.clidriver')
 LOG_FORMAT = (
@@ -90,21 +85,6 @@ def main():
     HISTORY_RECORDER.record('CLI_RC', rc, 'CLI')
     return rc
 
-
-def construct_cli_error_handlers_chain():
-    handlers = [
-        ParamValidationErrorsHandler(),
-        UnknownArgumentErrorHandler(),
-        ClientErrorHandler(),
-        ConfigurationErrorHandler(),
-        NoRegionErrorHandler(),
-        NoCredentialsErrorHandler(),
-        InterruptExceptionHandler(),
-        GeneralExceptionHandler()
-    ]
-    return ChainedExceptionHandler(exception_handlers=handlers)
-
-
 def create_clidriver():
     session = botocore.session.Session()
     _set_user_agent_for_session(session)
@@ -112,7 +92,7 @@ def create_clidriver():
                  event_hooks=session.get_component('event_emitter'))
     error_handlers_chain = construct_cli_error_handlers_chain()
     driver = CLIDriver(session=session,
-                       error_handlers_chain=error_handlers_chain)
+                       error_handler=error_handlers_chain)
     return driver
 
 
@@ -171,15 +151,15 @@ def no_pager_handler(session, parsed_args, **kwargs):
 
 class CLIDriver(object):
 
-    def __init__(self, session=None, error_handlers_chain=None):
+    def __init__(self, session=None, error_handler=None):
         if session is None:
             self.session = botocore.session.get_session()
             _set_user_agent_for_session(self.session)
         else:
             self.session = session
-        self._error_handlers_chain = error_handlers_chain
-        if self._error_handlers_chain is None:
-            self._error_handlers_chain = ChainedExceptionHandler()
+        self._error_handler = error_handler
+        if self._error_handler is None:
+            self._error_handler = construct_cli_error_handlers_chain()
         self._update_config_chain()
         self._cli_data = None
         self._command_table = None
@@ -444,8 +424,9 @@ class CLIDriver(object):
                 'CLI_VERSION', self.session.user_agent(), 'CLI')
             HISTORY_RECORDER.record('CLI_ARGUMENTS', args, 'CLI')
             return command_table[parsed_args.command](remaining, parsed_args)
-        except Exception as e:
-            return self._error_handlers_chain.handle_exception(
+        except BaseException as e:
+            LOG.debug("Exception caught in main()", exc_info=True)
+            return self._error_handler.handle_exception(
                 e,
                 stdout=get_stdout_text_writer(),
                 stderr=get_stderr_text_writer()
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -14,7 +14,6 @@ import json
 import os
 import platform
 import sys
-import signal
 import logging
 
 import distro
@@ -22,12 +21,6 @@ import distro
 import botocore.session
 from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
-from botocore.exceptions import NoCredentialsError
-from botocore.exceptions import NoRegionError
-from botocore.exceptions import ClientError
-from botocore.exceptions import (
-    ParamValidationError as BotocoreParamValidationError
-)
 from botocore.history import get_global_history_recorder
 from botocore.configprovider import InstanceVarProvider
 from botocore.configprovider import EnvironmentProvider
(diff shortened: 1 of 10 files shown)


kyleknap (commented):
I liked the direction in trying to peel out of this logic. I jotted the main feedback that I had in one of the comments. While the comment is sort of long, I suspect that it will not be too difficult to do if we decide to go forward with it. It should just entail reorganizing logic that you already had.

awscli/errorhandler.py
```diff
@@ -11,68 +11,110 @@
...
+        stderr.write("\n")
+        stderr.write(f'usage: {USAGE}\n{exception}\n')
+        stderr.write("\n")
+        return self.RC
+
+
+class InterruptExceptionHandler(FilteredExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = KeyboardInterrupt
+    RC = 128 + signal.SIGINT
+
```
kyleknap, line 103: It would be good to double check if it is needed, but looking over the code, it looks like it prints a newline to stdout when a interrupt was encountered.

awscli/errorhandler.py
```diff
@@ -11,68 +11,110 @@
...
+class GeneralExceptionHandler(FilteredExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = Exception
+    RC = GENERAL_ERROR_RC
+
+
+class ChainedExceptionHandler(BaseExceptionHandler):
+    def __init__(self, exception_handlers=None):
+        self._exception_handlers = exception_handlers
+        if self._exception_handlers is None:
+            self._exception_handlers = [GeneralExceptionHandler]
```
kyleknap, line 114: Let's just make `exception_handlers` a required parameter here and use the list provided instead of have a default handler. If we want to have the `GeneralExceptionHandler` be the only exception handler, I'd imagine we just use that directly instead of going through a chained handler.

awscli/errorhandler.py
```diff
@@ -11,68 +11,110 @@
...
+    def handle_exception(self, exception, stdout, stderr):
+        raise NotImplementedError('handle_exception')
+
+
+class FilteredExceptionHandler(BaseExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = ()
+    MESSAGE = '%s'
+
+    def handle_exception(self, exception, stdout, stderr):
+        if issubclass(exception.__class__, self.EXCEPTIONS_TO_HANDLE):
```
kyleknap, line 46: I wonder if instead we should just use an `isinstance` check here to simplify this check since we are provided error instances?

awscli/clidriver.py
```diff
@@ -103,12 +91,28 @@ def main():
     return rc
 
 
+def construct_cli_error_handlers_chain():
+    handlers = [
+        ParamValidationErrorsHandler(),
+        UnknownArgumentErrorHandler(),
+        ClientErrorHandler(),
```
kyleknap, line 98: Going through the original try/except branching logic, it seems like `ClientError` is handled later. I think it may make sense to lower this to after `InterruptExceptionHandler` to limit the blast radius if the order does actually matter and keep parity.

awscli/errorhandler.py
```diff
@@ -11,68 +11,110 @@
...
+        stderr.write(self.MESSAGE % exception)
+        stderr.write("\n")
+        return self.RC
+
+
+class ParamValidationErrorsHandler(FilteredExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = (
+        ParamError, ParamSyntaxError, ArgParseException,
+        ParamValidationError, BotocoreParamValidationError,
+        UnknownArgumentError
```
kyleknap, line 62: Should we be trying to handle `UnknownArgumentError` here? It seems like it gets caught by `UnknownArgumentErrorHandler` already.

awscli/errorhandler.py
```diff
@@ -11,68 +11,110 @@
...
+
+
+class NoCredentialsErrorHandler(FilteredExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = NoCredentialsError
+    RC = CONFIGURATION_ERROR_RC
+    MESSAGE = '%s. You can configure credentials by running "aws configure".'
+
+
+class UnknownArgumentErrorHandler(FilteredExceptionHandler):
+    EXCEPTIONS_TO_HANDLE = UnknownArgumentError
```
kyleknap, line 90: Is there a reason the new `ArgParseException` is not handled here as well? It seems like we can consolidate this class into `ParamValidationErrorsHandler` or delegate some of the error formatting logic in the updates for the `ArgParseException` class to this handler.

tests/unit/test_errorhandler.py
```diff
@@ -10,101 +10,74 @@
...
-class TestErrorHandler(unittest.TestCase):
+class TestChainedExceptionHandler(unittest.TestCase):
 
-    def create_http_response(self, **kwargs):
-        response = mock.Mock()
-        for key, value in kwargs.items():
-            setattr(response, key, value)
-        return response
+    @classmethod
+    def setUpClass(cls) -> None:
```
kyleknap, line 34: I'm guessing the type check is coming from the IDE here. I'd say remove the type check if we are not being consistent about it.

tests/unit/test_errorhandler.py
```diff
@@ -10,101 +10,74 @@
...
+            Case(UnknownArgumentError(), 252, ''),
+            Case(ArgParseException(), 252, '\n\n'),
+            Case(ParamSyntaxError(), 252, '\n\n'),
+            Case(ParamError(cli_name='cli', message='message'), 252,
+                 "'cli': message"),
+            Case(ParamValidationError(), 252, '\n\n'),
+            Case(ConfigurationError(), 253, '\n\n'),
+        ]
+        for case in cases:
+            self._assert_rc_and_error_message(case)
```
kyleknap, line 83: So it may make sense to use `yield` here. Similar to this test case: https://github.com/aws/aws-cli/blob/345da7177703ba297fd4b192bd8e9ad04898a592/tests/functional/autoprompt/test_autoprompt.py#L66. Mainly yield allows us to get through all of the test cases and not have it short circuit on us if one fails and potentially hide other failures.

tests/unit/test_errorhandler.py
```diff
@@ -10,101 +10,74 @@
...
-if __name__ == '__main__':
-    unittest.main()
+    def test_error_handling(self):
+        Case = namedtuple('Case', [
+            'exception',
+            'rc',
+            'message',
+        ])
+        cases = [
+            Case(Exception(), 255, '\n\n'),
```
kyleknap, line 66: Instead of asserting the newlines, let's adding a message to the exception and assert that the exception message is in the stderr. So to update this example:
```
Case(Exception('exception msg'), 255, 'exception msg')
```
I like this more because it is not really clear for any of the assertions that only use newlines or even empty strings why we would be asserting it. By asserting that the error message is in stderr is a bit more clear.

tests/unit/test_errorhandler.py
```diff
@@ -10,101 +10,74 @@
...
+    ParamValidationError, ConfigurationError
+)
 
-import mock
+from awscli.testutils import unittest
 from awscli import errorhandler
 
 
-class TestErrorHandler(unittest.TestCase):
+class TestChainedExceptionHandler(unittest.TestCase):
```
kyleknap, line 31: Then if we move to using `yield` we can just get rid of this class and replace the `setUpClass` method with calling `construct_default_cli_error_handlers_chain` in the `_assert_rc_and_error_message` function.

kyleknap (commented):
Nice. This is looking good. Just had some more suggestions and questions.

awscli/errorhandler.py
vz10: No idea why did we do it but ok.

awscli/errorhandler.py
vz10: Parsers use `format_usage` method to create usage message for `ArgParseException` and  different parsers can use different formats or pass there more than just a `USAGE`

awscli/clidriver.py
```diff
@@ -444,8 +424,8 @@ def main(self, args=None):
                 'CLI_VERSION', self.session.user_agent(), 'CLI')
             HISTORY_RECORDER.record('CLI_ARGUMENTS', args, 'CLI')
             return command_table[parsed_args.command](remaining, parsed_args)
-        except Exception as e:
-            return self._error_handlers_chain.handle_exception(
+        except BaseException as e:
```
kyleknap, line 427: Is there a particular reason/edge case we are catching `BaseException`? In general, `Exception` should be the base class for all potential errors that are thrown and is typically what we always catch if we want to catch all exceptions.

awscli/clidriver.py
```diff
@@ -434,68 +411,25 @@ def main(self, args=None):
...
-            # by the service. Generally, this will indicate incorrect API
-            # usage and is likely not an issue with CLI.
-            LOG.debug("Service returned an exception", exc_info=True)
-            write_exception(e, outfile=get_stderr_text_writer())
-            return CLIENT_ERROR_RC
-        except Exception as e:
-            # RC 255 is the catch-all. 255 specifically should not be relied
-            # on as exceptions can move from this catch-all classification
-            # to a more specific RC such as one of the above.
-            LOG.debug("Exception caught in main()", exc_info=True)
```
kyleknap, line 495: One thing that I missed from the previous review... We should definitely log the exception caught in `main()` and include the Traceback so that we can guarantee that we have a traceback for any exception in the logs. We can probably just take this line and paste it right before calling the `exception_handler.handle_exception` method

kyleknap (commented):
Looks good. Just had two more comments, and I think we are set to get this merged.

awscli/clidriver.py
vz10: KeyboardInterrupt inherited from BaseException

kyleknap: Gotcha. Let's keep it at `BaseException` then

kyleknap (approved):
Nice! I really like how this shaped up. 🚢

PR #5626, opened 2020-10-11
Fix bug when shorthand prompting not autocomplete in maps

vz10:
*Issue #, if available:*

*Description of changes:*
Fix shorthand prompting bug when autoprompt doesn't suggest anything  for map shape
`string -> structure` on such input

`key={`

### DIFF
--- a/awscli/autocomplete/local/basic.py
+++ b/awscli/autocomplete/local/basic.py
@@ -446,7 +446,7 @@ class ShorthandCompleter(BaseCompleter):
 
     def _get_prompt_for_map(self, arg_model, parsed_input):
         last_key, last_value = self._get_last_key_value(parsed_input)
-        if last_value:
+        if last_value is not None:
             return self._get_completion(arg_model.value, last_value)
         return self._get_completion(arg_model.key, last_key)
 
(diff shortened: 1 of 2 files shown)


kyleknap:
Can you provide an example command that demonstrates the bug that this fixes? I was trying to run some commands to test it out and I was not following what it fixed. Specifically in auto-prompt mode, I added:
```
> aws s3api put-object --metadata
```
and tried to tab complete the key value for `--metadata` and it was not giving me any suggestions.

vz10:
`> aws dynamodb put-item --item foo={` <tab>  

Now - nothing
Expected 
```
 S=                  [string] An attribute of type String. For example:  "S": "Hello"                                                                                                                                        
 N=                  [string] An attribute of type Number. For example:  "N": "123.45"  Numbers are sent across the network to DynamoDB as strings, to maximize compatibility across languages and libraries. However, D...  
 B=                  [blob] An attribute of type Binary. For example:  "B": "dGhpcyB0ZXh0IGlzIGJhc2U2NC1lbmNvZGVk"                                                                                                           
 SS=[                [list] An attribute of type String Set. For example:  "SS": ["Giraffe", "Hippo" ,"Zebra"]                                                                                                               
 NS=[                [list] An attribute of type Number Set. For example:  "NS": ["42.2", "-19", "7.5", "3.14"]  Numbers are sent across the network to DynamoDB as strings, to maximize compatibility across languages ...  
 BS=[                [list] An attribute of type Binary Set. For example:  "BS": ["U3Vubnk=", "UmFpbnk=", "U25vd3k="]                                                                                                        
 M={                 [map] An attribute of type Map. For example:
......
```

kyleknap (approved):
:ship:

PR #5635, opened 2020-10-14
Add auroprompt on-partial mode

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/core.py
+++ b/awscli/autoprompt/core.py
@@ -11,65 +11,66 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+from awscli.argparser import ArgTableArgParser
 from awscli.customizations.exceptions import ParamValidationError
 from awscli.autoprompt.prompttoolkit import PromptToolkitPrompter
-
-
-def validate_auto_prompt_args_are_mutually_exclusive(parsed_args, **kwargs):
-    cli_auto_prompt = getattr(parsed_args, 'cli_auto_prompt', False)
-    no_cli_auto_prompt = getattr(parsed_args, 'no_cli_auto_prompt', False)
-    if cli_auto_prompt and no_cli_auto_prompt:
-        raise ParamValidationError(
-            'Both --cli-auto-prompt and --no-cli-auto-prompt cannot be '
-            'specified at the same time.'
-        )
+from awscli.autocomplete.main import create_autocompleter
+from awscli.errorhandler import SilenceParamValidationMsgErrorHandler
 
 
 class AutoPromptDriver:
 
-    NO_PROMPT_ARGS = ['help', '--version']
+    _NO_PROMPT_ARGS = ['help', '--version']
+    _CLI_AUTO_PROMPT_OPTION = '--cli-auto-prompt'
+    _NO_CLI_AUTO_PROMPT_OPTION = '--no-cli-auto-prompt'
 
-    def __init__(self, session, prompter=None):
+    def __init__(self, driver, completion_source=None, prompter=None):
+        self._completion_source = completion_source
         self._prompter = prompter
-        self._session = session
+        self._session = driver.session
+        self._driver = driver
+        if self._completion_source is None:
+            self._completion_source = create_autocompleter(driver=self._driver)
 
-    def _should_autoprompt(self, parsed_args, args):
+    @property
+    def prompter(self):
+        if self._prompter is None:
+            self._prompter = AutoPrompter(self._completion_source,
+                                          self._driver)
+        return self._prompter
+
+    def validate_auto_prompt_args_are_mutually_exclusive(self, args):
+        no_cli_auto_prompt = self._NO_CLI_AUTO_PROMPT_OPTION in args
+        cli_auto_prompt = self._CLI_AUTO_PROMPT_OPTION in args
+        if cli_auto_prompt and no_cli_auto_prompt:
+            raise ParamValidationError(
+                'Both --cli-auto-prompt and --no-cli-auto-prompt cannot be '
+                'specified at the same time.'
+            )
+
+    def resolve_mode(self, args):
         # Order of precedence to check:
         # - check if any arg rom NO_PROMPT_ARGS in args
         # - check if '--no-cli-auto-prompt' was specified
         # - check if '--cli-auto-prompt' was specified
         # - check configuration chain
-        if any(arg in args for arg in self.NO_PROMPT_ARGS):
-            return False
-        if getattr(parsed_args, 'no_cli_auto_prompt', False):
-            return False
-        if getattr(parsed_args, 'cli_auto_prompt', False):
-            return True
+        self.validate_auto_prompt_args_are_mutually_exclusive(args)
+        if any(arg in args for arg in self._NO_PROMPT_ARGS):
+            return 'off'
+        if self._NO_CLI_AUTO_PROMPT_OPTION in args:
+            return 'off'
+        if self._CLI_AUTO_PROMPT_OPTION in args:
+            return 'on'
         config = self._session.get_config_variable('cli_auto_prompt')
-        return config.lower() == 'on'
+        return config.lower()
 
-    def auto_prompt_arguments(self, args, parsed_args, **kwargs):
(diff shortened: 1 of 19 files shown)


awscli/autoprompt/core.py
```diff
@@ -10,9 +10,20 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import sys
```
kyleknap, line 13: Alright so I have been doing some thinking on how to organize these changes. I think most of the logic needed is here and most of my feedback is about rearranging the interfaces. Specifically, it seems like this change was optimized toward reducing the change in the top-level `awscli.clidriver.main()` function. I think the problem with taking this approach is that as a result we have to push a lot of the logic into the `AutopromptDriver` and ends up seeming like it does too much, is not too reusable (e.g. similar to where we are at with the CLIDriver), and we also have to introduce new non-integer return codes to get it all working.

I think instead what we should do is create a new top-level class the wires together the logic of these two major classes (`CLIDriver` and `AutopromptDriver`) to make both of these classes less bloated and more reusable. Let's call this class for now `AWSCLIEntryPoint` and it has this sample structure:
```python
class AWSCLIEntryPoint:  # Maybe we can come with a better class name
    def main(self, args):
        try:
            rc = self._do_main(args)
        except Exception as e:
            # log exception
            # Set RC based on whatever handlers we want here
        HISTORY_RECORDER.record('CLI_RC', rc, 'CLI')
        return rc

    def _do_main(self, args):
        driver = create_clidriver()
        autoprompt_driver = AutoPromptDriver(driver)
        # This will return the mode and throw any validation errors if needed
        auto_prompt_mode = autoprompt_driver.resolve_mode(args)
        # We can also probably isolate each if branch to separate methods
        # and delegate to them using a gettattr()
        if auto_prompt_mode == 'on':
            args = autoprompt_driver.prompt_for_args(args)
            rc = driver.main(args)
        elif auto_prompt_mode == 'on-partial':
            autoprompt_driver.inject_requires_prompting_error_handler(driver)
            try:
                rc = driver.main(args)
            except autoprompt_driver.RequiresPromptingException:
                args = autoprompt_driver.prompt_for_args(args)
                driver = create_clidriver()
                rc = driver.main(args)
        else:
            rc = driver.main(args)
        return rc
```
and then make `main()` simply:
```python
def main():
    return AWSCLIEntryPoint().main(sys.argv[1:])
```
I sort of like this approach because:
1. It makes it easier to follow what the flow of a CLI command is from a single class. As it is right now, it sort of tricky to follow how the behavior works because this logic is all encompassed in the `AutoPromptDriver.prompt_for_args()` and that method is a little misleading itself in that the CLIDriver may be ran on your behalf (and succeed) or the user may never be auto-prompted for.
2. It simplifies the role of the `AutoPromptDriver` and its interfaces to help narrow down its role: as a bridge between CLI parameters/configuration and the `AutoPrompt` class to control when to use it. In terms of the interfaces that I had in the code sample, I just added two more methods: 1) `resolve_mode` which just does the validation of arguments and returns what mode to use. 2) `inject_requires_prompting_error_handler` which just should prepend a handler that raises a `RequiresPromptingException` in the chain handler if there is a parameter validation error. Then the `prompt_for_args` method should be as simple as delegating to the `AutoPrompter`.
3. The class starts laying out the framework for making the `CLIDriver` more pluggable and reusable. For example, on top of these changes I would like to see in the future we introduce a factory/builder for setting up a session based on configuration/global arguments that can be plumbed into the `CLIDriver`.  It also allows for better testing in that we can also add a `cli_driver_factory` and `auto_prompt_factory` to the tantalizer so we do not necessarily have to rely on patching. 
4. We keep `rc` reserved to integers/values that users will receive in their shell instead of introducing other potential type of rc's that we would need to account for and trip us up.

That was a bit long, but I don't think it is too big of a change of direction from where we are at. Let me know what you think.

kyleknap (commented):
Neat! This mode is going to be really helpful. See the comment I had below on organization of all of this

awscli/autoprompt/core.py
stealthycoin: 👍

awscli/autoprompt/core.py
```diff
@@ -29,47 +32,54 @@ class AutoPromptDriver:
...
+    def resolve_mode(self, args):
         # Order of precedence to check:
         # - check if any arg rom NO_PROMPT_ARGS in args
         # - check if '--no-cli-auto-prompt' was specified
         # - check if '--cli-auto-prompt' was specified
         # - check configuration chain
+        option_parser = ArgTableArgParser(self._driver.arg_table)
+        filtered_args = self._filter_out_options(args)
+        parsed_options, _ = option_parser.parse_known_args(filtered_args)
+        validate_auto_prompt_args_are_mutually_exclusive(parsed_options)
```
kyleknap, line 66: So one thing that I'm wonder if we should just roll out our own basic argument parsing for `--cli-auto-prompt` and `--no-cli-auto-prompt`? Specifically we would just validate both arguments are not in the args provided and just return `on` or `off` based on the argument provided. Looking over the driver code: https://github.com/aws/aws-cli/blob/ab41dc3429137dd964670e9bc342590ca4b7bf2b/awscli/clidriver.py#L319-L331 I'm a little concerned that we would emitting `building-top-level-params` out of order of the typical order it gets invoked and I'm not sure if there will be unknown side effects. Like I think this event might be typically emitted before it: https://github.com/aws/aws-cli/blob/ab41dc3429137dd964670e9bc342590ca4b7bf2b/awscli/clidriver.py#L298

awscli/autoprompt/core.py
```diff
@@ -29,47 +32,54 @@ class AutoPromptDriver:
...
-        if getattr(parsed_args, 'no_cli_auto_prompt', False):
-            return False
-        if getattr(parsed_args, 'cli_auto_prompt', False):
-            return True
+            return 'off'
+        if getattr(parsed_options, 'no_cli_auto_prompt', False):
+            return 'off'
+        if getattr(parsed_options, 'cli_auto_prompt', False):
+            return 'on'
         config = self._session.get_config_variable('cli_auto_prompt')
```
kyleknap, line 73: One bug that I think we will need to address is when we resolve `cli_auto_prompt` we do not take into account `--profile`. For example, if I'm using my `default` profile and and specify a `--profile other-profile` that does not have auto-prompt configured, I will still get auto-prompt mode. This may be a little tricky to fix and may need to make a larger change. So I'm happy with following up on this in another pull request if it requires larger changes

awscli/clidriver.py
```diff
@@ -145,10 +140,47 @@ def no_pager_handler(session, parsed_args, **kwargs):
...
+
+    def _do_main(self, args):
+        driver = create_clidriver()
+        autoprompt_driver = AutoPromptDriver(driver)
+        auto_prompt_mode = autoprompt_driver.resolve_mode(args)
+        if auto_prompt_mode == 'on':
+            args = autoprompt_driver.prompt_for_args(args)
+            rc = driver.main(args)
+        elif auto_prompt_mode == 'on-partial':
+            autoprompt_driver.inject_silent_param_error_handler(driver)
```
kyleknap, line 173: Let's rename this method to: `inject_silence_param_error_msg_handler(driver)` and the corresponding handler to `SilenceParamValidationMsgErrorHandler`. Mainly when I read silence, I was expecting it to swallow the entire exception and return a rc of 0. Adding the `Msg` part and switching to `Silence` better indicates that we are just not printing out the message.

kyleknap (commented):
Looks good. I really like the updates. I just had a few more comments and I think I ran into a bug around profiles.

awscli/autoprompt/core.py
kyleknap: On a similar note, it may make sense to update the functional tests to start using the new `AWSCLIEntryPoint` to see if points out any issues related to my concern.

vz10: Hmmmm. good point, no idea why we did it this way

kyleknap (approved):
Nice! :ship:

PR #5637, opened 2020-10-14
Add CLIRunner testing abstraction

kyleknap:
This class allows us to run a CLI command from a testing context without needing to subclass from any particular `TestCase` class from the `testutils` module. In addition, this class gives us the following benefits over the previous `TestCase` abstractions:

* All of the environment/system patching is built into the the `run` method of the abstraction, which makes it more convenient to use and reduces the chance of a bad patch.
* We exercise the full CLI stack down to when `botocore` sends a request to `urllib3`. This allows us to have more confidence that functionality is behaving appropriately.
* We validate parsed responses provided to the runner against its corresponding model.
* There are abstractions for both AWS responses and HTTP responses that can be interleaved during testing.

I also migrated the `codeartifact login` test suite over to use this abstraction to both show how other tests would adopt it and add some additional validation safety that would have prevented issues for the `login` command like this one: https://github.com/aws/aws-cli/pull/5382

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,5 @@
 import collections
+import copy
 import os
 
 import awscli
@@ -63,8 +64,9 @@ class CLIRunner(object):
         rc = driver.main(cmdline)
         self._session_stubber.assert_no_remaining_responses()
         runner_result = CLIRunnerResult(rc)
-        aws_requests = self._session_stubber.received_aws_requests.copy()
-        runner_result.aws_requests = aws_requests
+        runner_result.aws_requests = copy.copy(
+            self._session_stubber.received_aws_requests
+        )
         return runner_result
 
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -23,11 +23,13 @@ _LOADER = botocore.loaders.Loader()
 
 class CLIRunner(object):
     """Runs CLI commands in a stubbed environment"""
-    def __init__(self, env=None):
+    def __init__(self, env=None, session_stubber=None):
         if env is None:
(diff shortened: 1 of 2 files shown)


tests/__init__.py
```diff
@@ -1,4 +1,261 @@
...
+
+class AWSResponse(BaseResponse):
+    def __init__(self, service_name, operation_name, parsed_response):
+        self._service_name = service_name
+        self._operation_name = operation_name
+        self._parsed_response = parsed_response
+        self._service_model = self._get_service_model()
+        self._operation_model = self._service_model.operation_model(
+            self._operation_name)
+        self._validate_parsed_response()
```
stealthycoin, line 137: Doing validation logic in `__init__` is a bit weird, I can see the reason, being defensive against test-writers. Let's add a `validate=True` kwarg here so it can be disabled if needed.

tests/__init__.py
```diff
@@ -1,4 +1,261 @@
...
+_LOADER = botocore.loaders.Loader()
+
+
+class CLIRunner(object):
+    """Runs CLI commands in a stubbed environment"""
+    def __init__(self, env=None):
+        if env is None:
+            env = self._get_default_env()
+        self.env = env
+        self._session_stubber = SessionStubber()
```
stealthycoin, line 30: Can we inject this? Right now this is hardcoded into CLIRunner, and the response classes are hardcoded into this stubber. So we could make it easier to inject your own abstractions if we make all of this injectable.

stealthycoin (commented):
Looks good, just a few comments.

stealthycoin (approved):

PR #5638, opened 2020-10-14
Autoprompt functional tests scaffolding and functional tests for doc …

vz10:
…panel

*Issue #, if available:*

*Description of changes:*
Autoprompt functional tests scaffolding and functional tests for doc panel

### DIFF
--- a/tests/functional/autoprompt/test_prompttoolkit.py
+++ b/tests/functional/autoprompt/test_prompttoolkit.py
@@ -11,14 +11,13 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import json
-import os
-import shutil
-import tempfile
+import io
 
 from prompt_toolkit import Application
 from prompt_toolkit.key_binding.key_processor import KeyProcessor, KeyPress
 from prompt_toolkit.key_binding import merge_key_bindings
 from prompt_toolkit.key_binding.defaults import load_key_bindings
+from prompt_toolkit.input import DummyInput
 
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.utils import Event
@@ -32,7 +31,7 @@ from awscli.autoprompt.prompttoolkit import (
     PromptToolkitCompleter, PromptToolkitPrompter
 )
 from awscli.autoprompt.history import HistoryDriver
-from awscli.testutils import unittest, random_chars
+from awscli.testutils import unittest, random_chars, FileCreator
 
 
 def _ec2_only_command_table(command_table, **kwargs):
@@ -59,12 +58,23 @@ class FakeApplication:
         pre_run()
 
 
+class FakeInput(DummyInput):
+
+    def fileno(self):
+        return 0
+
+    @property
+    def closed(self):
+        return False
+
+
 class ApplicationStubber:
     _KEYPRESS = 'KEYPRESS'
     _ASSERTION = 'ASSERTION'
 
     def __init__(self, app):
         self._app = app
+        self._app.input = FakeInput()
         self._queue = []
         self._assertion_callback = lambda x: True
         self._default_keybindings = load_key_bindings()
@@ -84,7 +94,7 @@ class ApplicationStubber:
         # After each rendering application will run this callback
         # it takes the next action from the queue and performs it
         # some key_presses can lead to rerender, after which this callback
-        # will be run again before re-rendering app.invalidation property
+        # will be run again before re-rendering app.invalidated property
         # set to True.
         # On exit this callback also run so we need to remove it before exit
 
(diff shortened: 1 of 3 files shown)


tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -146,3 +234,273 @@ def test_handle_args_with_spaces(self):
...
+class TestPromptToolkitDocBuffer(BasicPromptToolkitTest):
+    def get_updated_doc_buffer_text(self, original_args):
+        self.prompter.args = original_args
+        self.prompter.pre_run()
+        return self.prompter.doc_buffer.document.text
+
+    def test_doc_buffer_not_shown_on_start_and_not_focusable(self):
+        stubber = ApplicationStubber(self.prompter.create_application())
+        stubber.add_key_assert(None, lambda app: app.show_doc is False)
+        stubber.add_key_assert(
```
kyleknap, line 248: It would be nice if we can then add helper methods to any of these common types of assertions so we do not have to build up the lambda each time. For example:
```python
def add_current_buffer_assertion(self, expected_buffer_name)
```
And we can then add assertions in these test cases such as:
```python
self.add_current_buffer_assertion('input_buffer')
```

tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -146,3 +234,273 @@ def test_handle_args_with_spaces(self):
...
+        stubber.add_key_assert('j')
+        stubber.add_key_assert('j')
+        stubber.add_key_assert('j', lambda app: app.layout.get_buffer_by_name(
+                    'doc_buffer').document.cursor_position_row == 3)
+        # get back to input
+        stubber.add_key_assert('q')
+        # go on enter the rest of the command and  check that cursor in doc
+        # panel remains on the same place
+        for k in ' --instances':
+            stubber.add_key_assert(k, lambda app: app.layout.get_buffer_by_name(
```
kyleknap, line 293: Is it possible to add assertions to the body for these? For example, we'd be able to define this as:
```python
lambda app: assert app.layout.get_buffer_by_name('doc_buffer').document.cursor_position_row == 3)
```
This would be nice because we would be able to make writing assertions a bit more intuitive and more powerful (e.g. have multiple assertions in a single function, asserting any caught errors, etc.)

kyleknap (commented):
Looks good. The interface is shaping up well. I just had some comments.

tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -146,3 +237,166 @@ def test_handle_args_with_spaces(self):
...
+            not app.current_buffer.history_mode and
+            app.current_buffer.name == 'input_buffer' and
+            app.current_buffer.document.text == 's3 ls '
+        )
+        stubber.run(self.prompter.pre_run)
+
+
+class TestCompletions(BasicPromptToolkitTest):
+    def test_service_full_name_shown(self):
+        self.prompter.args = ['e']
```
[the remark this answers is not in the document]
vz10: My model index for these tests has only `ec2` service data.

tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -51,45 +58,129 @@ def run(self, pre_run):
...
+    _ASSERTION = 'ASSERTION'
+
+    def __init__(self, app):
+        self._app = app
+        self._app.input = FakeInput()
+        self._queue = []
+        self._assertion_callback = lambda x: True
+        self._default_keybindings = load_key_bindings()
+
+    def add_key_assert(self, key=None, assertion=None):
```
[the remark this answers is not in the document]
vz10: Usually key press and assertion come together if I split them I'll have to pass key that was pressed on the previous step with assertion to be able to give meaningful error message

tests/functional/autoprompt/test_prompttoolkit.py
vz10: Yeah, it would be cleaner but unfortunately it's not a valid Python code

tests/functional/autoprompt/test_prompttoolkit.py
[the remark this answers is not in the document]
kyleknap: That's fair. Let's instead do the following:
1. Add a `add_keypress(self, key)` method just for adding a key press action
2. Rename this method to `add_keypress_with_assertion` and make the `key` and `assertion` required.

I'm mainly suggesting this because the method interface is a little hard to follow given there are two ways to use it (listed above) and it seems like it would make a little easier to follow if we just broke it up into two methods.

tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -51,45 +57,136 @@ def run(self, pre_run):
...
+    def add_current_buffer_assertion(self, buffer_name):
+        return lambda app: app.current_buffer.name == buffer_name
+
+    def add_buffer_is_visible_assertion(self, buffer_name):
+        return lambda app: buffer_name in [window.content.buffer.name
+                    for window in app.layout.get_visible_focusable_windows()]
+
+    def add_buffer_not_visible_assertion(self, buffer_name):
+        return lambda app: buffer_name not in [window.content.buffer.name
+                    for window in app.layout.get_visible_focusable_windows()]
```
kyleknap, line 175: These look good, but let's rename them to use `get` instead of `add`. I'm mainly suggesting this because when comparing it to the `add_keypress` and `add_keypress_with_assertion` these methods are only generate the assertions but do not affect test results because they are not added to the action queue.

kyleknap (commented):
Looks good. I just had a couple naming/refactoring suggestions.

kyleknap (approved):
Thanks! :ship:

PR #5645, opened 2020-10-16
Use startswith filter for completer and fuzzy filter for prompter

vz10:
*Issue #, if available:*

*Description of changes:*


kyleknap (commented):
Looks good. One thing I do wonder is if there is any functional tests we can put in place so that we do not regress on the default filter?

kyleknap (approved):
Thanks! I think we will probably need to figure out a more generalized/expansive approach to functional testing the auto-rompleter but this suffices for this PR. :ship:

PR #5650, opened 2020-10-16
Improve shorthand completer to handle nested empty lists

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autocomplete/local/basic.py
+++ b/awscli/autocomplete/local/basic.py
@@ -439,10 +439,15 @@ class ShorthandCompleter(BaseCompleter):
         # we have two way we can enter lists:
         # - if it's a top level structure it will be space separated and
         # parsed_input will contain the last element of the list
-        # - if it's a nested structure we'll get the whole list as an input
-        if not isinstance(parsed_input, list):
-            parsed_input = [parsed_input]
-        return self._get_completion(arg_model.member, parsed_input[-1])
+        # - if it's a nested structure we'll get the whole list as an input 
+        # and take the last item
+        # - if list is empty we'll pass None to the next completer
+        if isinstance(parsed_input, list):
+            if parsed_input:
+                parsed_input = parsed_input[-1]
+            else:
+                parsed_input = None
+        return self._get_completion(arg_model.member, parsed_input)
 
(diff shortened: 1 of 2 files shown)


awscli/autocomplete/local/basic.py
```diff
@@ -442,6 +442,8 @@ def _get_prompt_for_list(self, arg_model, parsed_input):
         # - if it's a nested structure we'll get the whole list as an input
         if not isinstance(parsed_input, list):
             parsed_input = [parsed_input]
+        if not len(parsed_input):
+            parsed_input = ['']
```
kyleknap, line 446: A couple questions here:

1) Could we just do `if not parsed_input` instead? It seems a little strange that we would be basing it off the result of a `len` when that should not be needed.
2) Can we maybe add a comment on how this fixes the issue? It is not really clear what the purpose of these lines are.

kyleknap (commented):
Cool. Just a couple of small questions.

awscli/autocomplete/local/basic.py
```diff
@@ -439,10 +439,12 @@ def _get_prompt_for_list(self, arg_model, parsed_input):
...
         # parsed_input will contain the last element of the list
-        # - if it's a nested structure we'll get the whole list as an input
-        if not isinstance(parsed_input, list):
-            parsed_input = [parsed_input]
-        return self._get_completion(arg_model.member, parsed_input[-1])
+        # - if it's a nested structure we'll get the whole list as an input 
+        # and take the last item
+        # - if list is empty we'll pass None to the next completer
+        if isinstance(parsed_input, list):
+            parsed_input = ([None] + parsed_input)[-1]
```
kyleknap, line 446: This is still hard to follow even with the comments. Could we organize it to follow the items in the comment? So something like this:
```python
if isinstance(parsed_input, list):
     if parsed_input:
        parsed_input = parsed_input[-1]
     else:
        parsed_input = None
```

kyleknap (commented):
Looks fine. Just commented on how the logic is structured.

kyleknap (approved):
Thanks! :ship:

PR #5655, opened 2020-10-19
Remove newlines from stripped help_text

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autocomplete/local/basic.py
+++ b/awscli/autocomplete/local/basic.py
@@ -19,9 +19,9 @@ from awscli.autocomplete.filters import startswith_filter
 from awscli.autocomplete import LazyClientCreator
 
 
-def strip_html_tags(text):
+def strip_html_tags_and_newlines(text):
     clean = re.compile('<.*?>')
-    return re.sub(clean, '', text)
+    return re.sub(clean, '', text).replace('\n', '')
 
 
 class ProfileCompleter(BaseCompleter):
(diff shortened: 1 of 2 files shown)


awscli/autocomplete/local/basic.py
```diff
@@ -19,9 +19,9 @@
 from awscli.autocomplete import LazyClientCreator
 
 
-def strip_html_tags(text):
+def strip_html_tags_and_newlines(text):
     clean = re.compile('<.*?>')
-    return re.sub(clean, '', text)
+    return re.sub(clean, '', text).replace('\n', '')
```
kyleknap, line 24: I wonder if we should also try to remove other potentially problematic characters such as carriage returns, tabs, etc? We do similar logic in another customization: https://github.com/aws/aws-cli/blob/v2/awscli/customizations/utils.py#L208 where we also remove the line separator character. It might be interesting to dive more into this to see what other characters result in weird rendering.

kyleknap (commented):
Looks fine. Just had a comment about proactively catching potential other characters.

awscli/autocomplete/local/basic.py
vz10: Basically the only help text that consists newlines is help text for CLI custom arguments, all the models have "clean" help text. Probably if we have help text with tabs and carriage returns in it we need to fix text and not just trim it.

kyleknap: Sounds good. We can just go with newline for now.

kyleknap (approved):

PR #5659, opened 2020-10-20
Add help panel

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/factory.py
+++ b/awscli/autoprompt/factory.py
@@ -12,44 +12,24 @@
 # language governing permissions and limitations under the License.
 import os
 
-from prompt_toolkit.application import get_app
 from prompt_toolkit.buffer import Buffer
 from prompt_toolkit.document import Document
-from prompt_toolkit.filters import Condition
-from prompt_toolkit.formatted_text import HTML, to_formatted_text
-from prompt_toolkit.formatted_text.utils import fragment_list_to_text
 from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.keys import Keys
-from prompt_toolkit.layout import Float, FloatContainer, HSplit, Window
+from prompt_toolkit.layout import Float, FloatContainer, HSplit, Window, VSplit
 from prompt_toolkit.layout.controls import BufferControl
 from prompt_toolkit.layout.dimension import Dimension
 from prompt_toolkit.layout.layout import Layout, ConditionalContainer
 from prompt_toolkit.layout.menus import CompletionsMenu, MultiColumnCompletionsMenu
-from prompt_toolkit.layout.processors import (
-    BeforeInput, Processor, Transformation)
+from prompt_toolkit.layout.processors import BeforeInput
 from prompt_toolkit.widgets import HorizontalLine, SearchToolbar
 
-from awscli.autoprompt.history import (
-    HistoryDriver, HistoryCompleter
+from awscli.autoprompt.history import HistoryDriver, HistoryCompleter
+from awscli.autoprompt.widgets import HelpPanelWidget, ToolbarWidget
+from awscli.autoprompt.filters import (
+    is_one_column, is_multi_column, doc_section_visible,
+    input_buffer_has_focus, doc_window_has_focus, is_history_mode
 )
-
-
-@Condition
-def doc_section_visible():
-    app = get_app()
-    return getattr(app, 'show_doc')
-
-
-@Condition
-def is_multi_column():
-    app = get_app()
-    return getattr(app, 'multi_column', False)
-
-
-@Condition
-def is_one_column():
-    app = get_app()
-    return not getattr(app, 'multi_column', False)
 
 
 class PrompterKeyboardInterrupt(KeyboardInterrupt):
@@ -80,7 +60,6 @@ class PromptToolkitFactory:
         'input_buffer_height_min': 10,
         'menu_height_max': 16,
         'menu_scroll_offset': 2,
-        'toolbar_height_max': 2
     }
 
     def __init__(self, completer):
@@ -101,9 +80,6 @@ class PromptToolkitFactory:
 
     def create_doc_buffer(self):
         return Buffer(name='doc_buffer', read_only=True)
-
-    def create_bottom_toolbar_buffer(self):
-        return Buffer(name='bottom_toolbar_buffer', read_only=True)
 
     def create_input_buffer_container(self, input_buffer):
         return FloatContainer(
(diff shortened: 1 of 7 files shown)


awscli/autoprompt/factory.py
```diff
@@ -191,26 +243,51 @@ def create_layout(self, on_input_buffer_text_changed=None,
...
             bottom_toolbar_container = \
                 self.create_bottom_toolbar_container(bottom_toolbar_buffer)
+        help_doc_buffer = self.create_help_buffer(
+            self._help_text.full_doc_window_key_binding_text
+        )
+        help_doc_window = self.create_help_window(
+            help_doc_buffer, title='Doc panel help')
+        help_input_buffer = self.create_help_buffer(
+            self._help_text.full_input_buffer_key_binding_text
+        )
```
kyleknap, line 262: Is there a particular reason we are introducing a new pattern for a creating a container for every new type of help panel? From the looks of it the toolbar just uses a single container and we update the input buffer based on the keypresses. I sort of like that approach because it makes following the layout of the app easier especially if we add more help panels for more views.

awscli/autoprompt/factory.py
```diff
@@ -191,26 +243,51 @@ def create_layout(self, on_input_buffer_text_changed=None,
...
+                    ConditionalContainer(
+                        help_input_window,
+                        help_section_visible & input_buffer_has_focus),
+                ]),
+                search_field,
+                ConditionalContainer(HorizontalLine(),
+                                     ~help_section_visible),
+                ConditionalContainer(bottom_toolbar_container,
+                                     ~help_section_visible),
+            ])
```
kyleknap, line 290: It would be nice if we can reduce the complexity of this top-level view of the app back to what we had before. Specifically, there are too many containers from this view and too much nesting. I think to address it, we can just make more helper functions that return top-level containers and hide the nesting in these helper functions and still have a nice higher-level view of the setup like we had before.

awscli/autoprompt/factory.py
```diff
@@ -430,39 +499,55 @@ def apply_transformation(self, text_input):
         return Transformation(fragments)
 
 
-class ToolbarHelpText:
-    def __init__(self, style=None, spacing=None):
+class HelpText:
```
kyleknap, line 502: Ok I think it makes sense if we start refactoring this class and try to make it more generalized/reusable. Specifically, I'm thinking we make this class more of a factory class that returns help text based on some identifier for the component we are getting help text for. For example:

```python
class BaseHelpFormatter:
     def format(self):
          raise NotImplementedError('format')


class ToolbarHelpFormatter(BaseHelpFormatter):
    KEYS_TO_DESCRIPTION = {}
    def format(self):
         # Format string using spaces


class InputBufferToolbarHelpFormatter(ToolbarHelpFormatter):
    KEYS_TO_DESCRIPTION = {
         # Put the keys we care about here
    }


 # Add similar classes for the full help windows but make them delineated by newlines


class HelpTextGetter:
     def __init__(self, help_texts=None):
           if not help_texts:
                help_texts = self._get_default_help_texts()  # This just uses the formatter classes to build up the classes
           self._help_texts = help_texts  # type: Dict[str, str] where the key is the id and the value is the help text

     def get_help_text(self, help_id):
            # Returns the help text based on the id provided
```
Then getting help text to populate the buffers would be a matter of calling the `get_help_text()` function with the correct id. For example:
```
help_text_getter.get_help_text('input_buffer_tool_bar')
```

kyleknap (commented):
Works well. I mainly had refactoring suggestions.

awscli/autoprompt/factory.py
vz10: Yes. Because title in frame is immutable.

awscli/autoprompt/factory.py
vz10: Maybe we can move to another direction. It's just a bunch of strings so we can  just get rid of class and create a bunch of constants.

kyleknap: Yeah the works too. The only reason I added the classes so we can reuse logic in building up these big menus.

awscli/autoprompt/factory.py
kyleknap: That's fair. I think my biggest feedback then is let's make a helper method, factory, or even a new prompt-toolkit class (e.g. implement the `__pt_container__` method) to encompass all of these containers as it makes it hard to understand this top-level view of the app structure. Ideally this utility would just return a single container with all of the full help windows and we can just add that container to the top-level view.

awscli/autoprompt/factory.py
kyleknap: Looking over more of this conversation, another idea is that we could just implement our own prompt-toolkit container class representing a particular help window/buffer and just make the content constants of the class. I think that can help consolidate all of this logic as a well to a single class.

awscli/autoprompt/widgets.py
```diff
@@ -0,0 +1,208 @@
...
+            help_input_buffer, title='Prompt panel help')
+        self.container = HSplit([
+            ConditionalContainer(
+                self.help_doc_window,
+                help_section_visible & (doc_window_has_focus
+                                        | search_input_has_focus)),
+            ConditionalContainer(
+                self.help_input_window,
+                help_section_visible & input_buffer_has_focus)
+        ])
```
kyleknap, line 69: Instead of chaining the `help_section_visible` filter for both of these ConditionalContainer, could we just encapsulate the `HSplit` with a `ConditionalContainer` and hoist the `help_section_visible` to that?

awscli/autoprompt/widgets.py
```diff
@@ -0,0 +1,208 @@
...
+    def input_buffer_key_binding_text(self):
+        return (
+            f'{self.style}[ENTER]</style> Autocomplete '
+            f'Choice/Execute Command{self.eol}'
+            f'{self.style}[F1]</style> Hide/Show Help{self.eol}'
+            f'{self.style}[F2]</style> Focus on Docs{self.eol}'
+            f'{self.style}[F3]</style> Hide/Show Docs{self.eol}'
+            f'{self.style}[F4]</style> One/Multi column prompt{self.eol}'
+            f'{self.style}[CONTROL+R]</style> On/Off bck-i-search'
+        )
```
kyleknap, line 80: So instead of having a bunch of these helper methods, what are your thoughts about generalizing these methods into their own classes? So for example, we would have a base class that is a prompt toolkit container and try to parameterize/generalize option in class variables as much as we can:
```python
class BaseHelpView:
    TITLE = None
    STYLE = '<style fg="darkturquoise">'
    KEYS_TO_DESCRIPTION_MAPPING = {}
    CONDITION = None

def __pt_container__(self):
    pass
```
and then we would have specific classes for different views:
```
class InputBufferHelpView:
    ...

class DocWindowHelpView:
   ...
```
I like this approach for the following reasons:
1. It makes it really easy to add new views in the future as we would just need to fill out a new class
2. We make the `HelpPanelWidget` class even more slimmed down and general. For example, with this approach we could then allow `help_views` to be provided to its tantalizer and the `HelpPanelWidget` would not necessarily need to know anything about the different views and have specific methods for them.

awscli/autoprompt/widgets.py
```diff
@@ -0,0 +1,208 @@
...
+            ConditionalContainer(
+                self.help_input_window,
+                help_section_visible & input_buffer_has_focus)
+        ])
+
+    def input_buffer_key_binding_text(self):
+        return (
+            f'{self.style}[ENTER]</style> Autocomplete '
+            f'Choice/Execute Command{self.eol}'
+            f'{self.style}[F1]</style> Hide/Show Help{self.eol}'
```
kyleknap, line 75: So one thing that I found a little jarring is that in the toolbar, I interpreted `Show Help` as show the doc panel because the content in the doc panel comes from the CLI's `Help` command. I wonder if we should just be a little more specific about the type of help we would be seeing and describe it as `Hide/Show Shortkey Help`

awscli/autoprompt/widgets.py
```diff
@@ -0,0 +1,208 @@
...
+                wrap_lines=True,
+            ),
+            title=title
+        )
+
+    def __pt_container__(self):
+        return self.container
+
+
+class ToolbarWidget:
```
kyleknap, line 124: If we go with the feedback that I suggested with the `HelpPanelWidget`, I think we should do the same with the `ToolbarWidget` as well.

tests/unit/autoprompt/test_widgets.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_container_has_correct_children(self):
+        children = self.widget.container.get_children()
+        self.assertTrue(isinstance(children[0], ConditionalContainer))
+        self.assertTrue(isinstance(children[1], VSplit))
+
+    def test_toolbar_history_wording(self):
+        buffer = self.widget.container.get_children()[1]\
+                            .get_children()[0].content.content.buffer
+        self.assertEqual(buffer.document.text, self.widget.history_mode_sign())
```
kyleknap, line 40: I think if we break the methods out into separate classes, I think that will improve the tests as well. Right now it is a little hard to understand with the children traversal and assertions of general classes what we are trying to test.

kyleknap (commented):
Nice. I like how it starting to look. Just had some more suggestions.

kyleknap (approved):
Nice! This looks a lot cleaner. Thanks for making all of these updates. :ship: Feel free to merge it assuming Travis passes.

PR #5663, opened 2020-10-22
Add paramfile exception for sesv2 put-account-details website-uri.

### LINKED ISSUE #5641: `aws sesv2 put-account-details --website-url https://example.com` shouldn't download the contents of example.com
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) ~and didn't find any solution~ and found a similar problem from 2015.
  
**Describe the bug**

I'm trying to use the CLI to move my SES region out of the sandbox.

```sh
aws --debug sesv2 put-account-details --production-access-enabled --mail-type TRANSACTIONAL --website-url "https://example.com" --use-case-description "..." --contact-language EN
```

This produces a very odd message:
~~~~
An error occurred (BadRequestException) when calling the PutAccountDetails operation: 2 validation errors detected: Value at 'websiteURL' failed to satisfy constraint: Member must satisfy regular expression pattern: ^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?; Value at 'websiteURL' failed to satisfy constraint: Member must have length less than or equal to 1000
~~~~

Using `--debug` shows that it actually queryied my URL:
~~~~
2020-10-15 15:51:55,756 - MainThread - botocore.hooks - DEBUG - Changing event name from load-cli-arg.email.put-account-details.website-url to load-cli-arg.ses.put-account-details.website-url
2020-10-15 15:51:55,756 - MainThread - botocore.hooks - DEBUG - Event load-cli-arg.ses.put-account-details.website-url: calling handler <awscli.paramfile.URIArgumentHandler object at 0x7f06e95e3940>
2020-10-15 15:51:55,757 - MainThread - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): example.com:443
2020-10-15 15:51:55,836 - MainThread - urllib3.connectionpool - DEBUG - https://example.com:443 "GET / HTTP/1.1" 200 1454225
[...]
~~~~

**SDK version number**
aws-cli/1.18.157 Python/3.8.6 Linux/5.8.11-200.fc32.x86_64 botocore/1.18.16

**Platform/OS/Hardware/Device**
Fedora 32, amd64

**Additional context**

This looks similar to https://github.com/aws/aws-cli/issues/1130, so I suspect the correct fix is to modify https://github.com/aws/aws-cli/blob/develop/awscli/paramfile.py accordingly for this parameter.

A work-around: I couldn't find a way to quote this in a way that would disable the downloading, BUT the regexp points to the solution: Remove the `https://` and use `--website-url example.com`.

kdaily:
*Issue #, if available:*

Fixes #5641.

*Description of changes:*

Add paramfile exception for sesv2 put-account-details website-uri

### DIFF
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -107,7 +107,7 @@ PARAMFILE_DISABLED = set([
     'service-catalog.create-product.support-url',
     'service-catalog.update-product.support-url',
 
-    'sesv2.put-account-details.website-url',
+    'ses.put-account-details.website-url',
 
     'sqs.add-permission.queue-url',
     'sqs.change-message-visibility.queue-url',
--- a/awscli/paramfile.py
+++ b/awscli/paramfile.py
@@ -107,6 +107,8 @@ PARAMFILE_DISABLED = set([
     'service-catalog.create-product.support-url',
     'service-catalog.update-product.support-url',
 
+    'sesv2.put-account-details.website-url',
+
     'sqs.add-permission.queue-url',
     'sqs.change-message-visibility.queue-url',
     'sqs.change-message-visibility-batch.queue-url',
--- a/tests/functional/sesv2/__init__.py
+++ b/tests/functional/sesv2/__init__.py
@@ -1,0 +1,12 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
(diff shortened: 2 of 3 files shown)


awscli/paramfile.py
```diff
@@ -107,6 +107,8 @@
     'service-catalog.create-product.support-url',
     'service-catalog.update-product.support-url',
 
+    'sesv2.put-account-details.website-url',
```
kyleknap, line 110: So one not so great gotcha (that we should change) is that for the prefix of these event names, we use the endpoint prefix. For `sesv2` the endpoint prefix is `ses`, so the event should be `ses.put-account-details.website-url`.

I think for this one because of this gotcha, we should add a test for it. It would be great if we can add a test like this one: https://github.com/aws/aws-cli/blob/develop/tests/functional/iam/test_create_open_id_connect_provider.py where we make sure the value used is the URL instead of its contents.

kyleknap (commented):
Just had that comment after testing out the update. Should be an easy fix.

kyleknap (approved):
:ship:

Issue #5670, opened 2020-10-23
[v2] CLI auto-prompt mode documentation shows ^M characters on windows


ppittle:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ ] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**

**SDK version number**
aws-cli/2.0.58 Python/3.7.7 Windows/10 exe/AMD64

**Platform/OS/Hardware/Device**
What are you running the cli on? Windows 10

**To Reproduce (observed behavior)**
Steps to reproduce the behavior
1. On a Windows 10 machine open Powershell (pwsh)
2. Type `aws`
3. Press `F3`

**Expected behavior**

Don't show `^M` in place of whitespace:

![image](https://user-images.githubusercontent.com/1190907/97046094-b8045200-152b-11eb-95b1-d90ebbda5cc4.png)

**Logs/output**
Get full traceback and error logs by adding `--debug` to the command.

**Additional context**

kdaily:
Thanks @ppittle!

stobrien89:
Closing, as fix has been merged.

PR #5671, opened 2020-10-23
Remove unprintable character from docs. Fix typo

vz10:
*Issue #, if available:* #5670

*Description of changes:*

### DIFF
--- a/tests/unit/autoprompt/test_doc.py
+++ b/tests/unit/autoprompt/test_doc.py
@@ -123,6 +123,19 @@ class TestBaseDocsGetter(unittest.TestCase):
         response = self.base_docs_getter.get_doc_content(self.help_command)
         self.assertEqual(response, expected_response)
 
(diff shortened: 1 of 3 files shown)


tests/unit/autoprompt/test_doc.py
```diff
@@ -96,16 +96,16 @@ def test_get_doc_content(self):
 
     def test_get_rst_doc_in_txt(self):
         content = textwrap.dedent("""\
-            MySection
```
kyleknap, line 99: It would be nice if we break this test into its own separate test case (e.g. `get_doc_content_strips_carriage_returns`). My main concern in reusing the existing test case is that it does not indicate that it also ensures  carriage returns are stripped out. So if we ever update the content, it may be easy to overlook that the carriage returns need to be retained for the test.

kyleknap (commented):
Looks fine. Just had a small comment about the test.

kyleknap (approved):
Nice! :ship:

PR #5673, opened 2020-10-26
Add debug panel to automprompt UI

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/widgets.py
+++ b/awscli/autoprompt/widgets.py
@@ -10,20 +10,29 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import os
+
+from prompt_toolkit.application import get_app
+from prompt_toolkit.filters import has_focus
 from prompt_toolkit.formatted_text import HTML, to_formatted_text
 from prompt_toolkit.formatted_text.utils import fragment_list_to_text
 from prompt_toolkit.buffer import Buffer
 from prompt_toolkit.document import Document
-from prompt_toolkit.layout import HSplit, Window, VSplit
+from prompt_toolkit.key_binding import KeyBindings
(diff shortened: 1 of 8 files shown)


awscli/autoprompt/widgets.py
```diff
@@ -195,3 +213,97 @@ def __init__(self):
...
+            app = get_app()
+            textfield.text = default_filename
+            app.layout.focus(textfield)
+            self.float_container.floats.pop()
+            app.layout.focus(self._dialog_prev_focus)
+
+        def ok_handler():
+            app = get_app()
+            buffer = app.layout.get_buffer_by_name('debug_buffer')
+            with open(textfield.text, 'w') as f:
```
kyleknap, line 287: We should make sure to expand both the absolute path and user for this file provided. I ran into an error when I typed `~/debug.log` in the dialogue.

awscli/autoprompt/widgets.py
```diff
@@ -195,3 +213,97 @@ def __init__(self):
...
+
+        textfield = TextArea(text=default_filename, multiline=False)
+
+        return Dialog(
+            title='Save logs to file',
+            body=HSplit([
+                Label(text='Log file name', dont_extend_height=True),
+                textfield,
+            ], padding=Dimension(preferred=1, max=1)),
+            buttons=[ok_button, cancel_button],
```
kyleknap, line 305: Not sure how easy it is, but it would be nice I could hit `<ENTER>` when filename field is focused and have that trigger the saving of the file. Mainly it took me a little time to figure out that I needed to tab to the save button. Then maybe instead we could get rid of the save button and just let users hit `<ENTER>` on the text field to simplify the dialogue? Thoughts?

kyleknap (commented):
This looks good to me. I just had a few questions/suggestions. I'm excited to see what it looks like with the logging wired up to the frame.

kyleknap (commented):
Looks good. Just one more logging name to update and we should be good to merge this.

kyleknap (approved):
Thanks! :ship:

PR #5674, opened 2020-10-27
Update cloudformation describe changeset test to not be as strict

stealthycoin:
Test is now no longer dependent on the exact output from describe
changeset, rather it checks for the fields that are set in the parsed
response.


tests/unit/output/test_text_output.py
```diff
@@ -131,14 +131,16 @@ def test_non_aggregate_keys(self):
...
-        self.assertEqual(
-            output,
-            ("arn:aws:cloudformation:us-west-2:12345:changeSet/mychangeset/"
-             "12345\tmychangeset\t2019-04-08T14:21:53.765Z\tNone\tAVAILABLE"
-             "\tNone\tarn:aws:cloudformation:us-west-2:12345:stack/MyStack"
-             "/12345\tMyStack\tCREATE_COMPLETE\tNone\tNone\n"
-             "CAPABILITIES\tCAPABILITY_IAM\nCHANGES\t1\nCHANGES\t2\n")
-        )
+        fields = output.split()
+        self.assertIn((
```
kyleknap, line 135: It would be nice to assert that we only include non aggregate keys once in the output in order to make sure we do not treat them like result keys (e.g. printing out multiple times). That is one thing that we lose out from in not asserting the whole output.

kyleknap (commented):
Looks fine. Just had a small comment.

PR #5681, opened 2020-10-29
Add GitHub action for CI

stealthycoin:
Add CI GitHub Action.


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,47 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
```
kyleknap, line 4: Let's name this `AWS CLI tests`

.travis.yml
```diff
@@ -1,30 +0,0 @@
...
-      dist: trusty
-    - python: 3.7
-      dist: xenial
-    - python: 3.8
-      dist: xenial
-
-before_install:
-  - if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then
-      echo "No pull requests can be sent to the master branch" 1>&2;
-      exit 1;
```
kyleknap, line 20: Can we convert this into a GitHub action? This was a helpful check to have as it is really easy to miss if a pull request is targeting the master branch.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
```
kyleknap, line 10: I'm assuming we would just add `v2` to this list when we port it to that branch?

tests/functional/s3/test_cp_command.py
```diff
@@ -667,22 +668,50 @@ def test_upload_unicode_path(self):
...
+    #             'BucketName': 'bucket'
+    #         }
+    #     }]
+    #     self.http_response.status_code = 404
+
+    #     full_path = self.files.create_file('foo.txt', 'bar')
+    #     set_invalid_utime(full_path)
+    #     _, stderr, rc = self.run_cmd(command % full_path, expected_rc=1)
+    #     self.assertIn('upload failed', stderr)
+    #     self.assertIn('warning: File has an invalid timestamp.', stderr)
```
kyleknap, line 686: We should delete this

tests/functional/s3/test_cp_command.py
```diff
@@ -14,6 +14,7 @@
 import mock
 import os
 
+from awscli.testutils import skip_if_windows
```
kyleknap, line 17: Is this extra import needed?

tests/functional/s3/test_sync_command.py
```diff
@@ -132,25 +132,6 @@ def test_turn_off_glacier_warnings(self):
         self.assertEqual(self.operations_called[0][0].name, 'ListObjectsV2')
         self.assertEqual('', stderr)
 
-    def test_warning_on_invalid_timestamp(self):
```
kyleknap, line 135: Looking over the commit it got added in: https://github.com/aws/aws-cli/commit/630ae1d9c83744f4df980b94d799d490dc9ead97#, I don't feel comfortable removing this test as there is not really anything to replace it. If there is not a great way to reproduce this at a system level anymore, I'd even be fine with just patching the `get_file_stat()` utility method for this test and adding a todo to find a better way to test it in the future with out patching.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+        pip install codecov
+        python scripts/ci/install
+        python scripts/ci/install-check
+    - name: Run tests and checks
+      run: |
+        python scripts/ci/run-tests
+        python scripts/ci/run-check
+
+    - name: codecov
```
kyleknap, line 43: Codecov does not seem to be appearing on this PR. I wonder if we are missing anything here? It might be worth trying the codecov action: https://github.com/codecov/codecov-action

kyleknap (commented):
Nice. Looks good. It's awesome that we have all three platforms that we support now covered in our CI for PR's. Otherwise, I just had some comments.

.travis.yml
stealthycoin: Currently it wont even run tests if its made against non-develop branch. Do we want to run them and then error out?

.github/workflows/python-package.yml
stealthycoin: We would change it to just v2 for the v2 branch.

.github/workflows/python-package.yml
stealthycoin: Its in the list of checks. The checks list is large now due to windows/mac. The comment did appear in my fork, don't know why its not here.

.travis.yml
kyleknap: I was thinking we just have a separate action that just fails out on submitters to tell them we do not accept PR's sent to master. The main reason to have this is so we can signal back to the submitter to switch the target branch themselves. Otherwise, it is not obvious that you messed up the PR by sending it to master.

kyleknap: Offline conversation... We are just going to send this update to fail on PRs to master as a separate PR.

kyleknap (approved):
Looks great. :ship: Let's merge it once the checks pass.

PR #5687, opened 2020-10-29
Replace tests that rely on unreadable files with mocked timestamp

stealthycoin:
The timestamp tests do not work on all platforms, but the replacement
tests that rely on an unreadable file also do not work in some
environments. This commit reverts back to a timestamp based test, but
rather than relying on the acutal file timestamp we just mock out the
helper method that fetches it. We should replace these in the future
with something more robust that does not rely on a mock.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -91,18 +91,6 @@ def skip_if_windows(reason):
     return decorator
 
 
-def set_invalid_utime(path):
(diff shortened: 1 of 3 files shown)


kyleknap (approved):
:ship: assuming checks pass

PR #5688, opened 2020-10-30
Fix v2 functional tests

stealthycoin:


tests/functional/s3/test_sync_command.py
```diff
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from mock import patch
```
kyleknap, line 13: This line does not look like it is used.

awscli/testutils.py
```diff
@@ -81,6 +81,7 @@ def decorator(func):
     return decorator
 
 
+
```
kyleknap, line 84: Extra newline here we can probably remove.

kyleknap (commented):
Looks fine. There were just a couple of lines we can remove

kyleknap (approved):
:ship:

PR #5696, opened 2020-11-03
Update logger to support autoprompt UI

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/core.py
+++ b/awscli/autoprompt/core.py
@@ -10,8 +10,8 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from botocore.exceptions import ProfileNotFound
 
-from awscli.argparser import ArgTableArgParser
 from awscli.customizations.exceptions import ParamValidationError
 from awscli.autoprompt.prompttoolkit import PromptToolkitPrompter
 from awscli.autocomplete.main import create_autocompleter
@@ -63,8 +63,11 @@ class AutoPromptDriver:
             return 'off'
         if self._CLI_AUTO_PROMPT_OPTION in args:
             return 'on'
-        config = self._session.get_config_variable('cli_auto_prompt')
-        return config.lower()
+        try:
+            config = self._session.get_config_variable('cli_auto_prompt')
+            return config.lower()
+        except ProfileNotFound:
+            return 'off'
 
     def inject_silence_param_error_msg_handler(self, driver):
         driver.error_handler.inject_handler(
--- a/awscli/autoprompt/logger.py
+++ b/awscli/autoprompt/logger.py
@@ -1,0 +1,36 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import logging
+
+from prompt_toolkit.document import Document
+from prompt_toolkit.application import get_app
+
+
+class PromptToolkitHandler(logging.StreamHandler):
+
+    def emit(self, record):
+        try:
+            app = get_app()
+            if app.is_running and getattr(app, 'debug', False):
+                msg = self.format(record)
+                debug_buffer = app.layout.get_buffer_by_name('debug_buffer')
+                current_document = debug_buffer.document.text
+                if current_document:
+                    msg = '\n'.join([current_document, msg])
+                debug_buffer.set_document(
+                    Document(text=msg), bypass_readonly=True
+                )
+            else:
+                super().emit(record)
+        except:
+            self.handleError(record)
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -36,6 +36,7 @@ from awscli.formatter import get_formatter
 from awscli.plugin import load_plugins
 from awscli.commands import CLICommand
 from awscli.argparser import MainArgParser
+from awscli.argparser import FirstPassGlobalArgParser
 from awscli.argparser import ServiceArgParser
 from awscli.argparser import ArgTableArgParser
 from awscli.help import ProviderHelpCommand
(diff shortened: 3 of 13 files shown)


tests/unit/test_clidriver.py
```diff
@@ -1077,5 +1093,20 @@ def test_handle_exception_in_main(self):
...
+        with contextlib.redirect_stderr(stderr):
+            driver = create_clidriver(['--profile', 'foo', '--debug'])
+        profile = driver.session.get_config_variable('profile')
+        self.assertEqual(profile, 'foo')
+        self.assertIn('CLI version', stderr.getvalue())
+
+    def test_create_cli_driver_wo_args(self):
+        driver = create_clidriver([])
+        profile = driver.session.get_config_variable('profile')
+        self.assertIsNone(profile)
```
kyleknap, line 1108: I'd be hesitant to asserting `None` here without patching out the environment. Specifically, the profile can also be set through environment variables and would probably cause this test to fail if they got set.

awscli/clidriver.py
```diff
@@ -166,7 +172,7 @@ def main(self, args):
     def _do_main(self, args):
         driver = self._driver
         if driver is None:
-            driver = create_clidriver()
+            driver = create_clidriver(args)
```
kyleknap, line 175: So one thing, I think this change breaks is any command that can accept a non-existent profile, no longer works. For example, I now get this error:
```
$ aws configure set foo bar --profile not-here

The config profile (not-here) could not be found
```
To get this working with profile, I think we will have to rethink how we do auto-prompt resolution.

awscli/logger.py
```diff
@@ -0,0 +1,103 @@
...
+
+    """
+    log = logging.getLogger(logger_name)
+    handler_name = f'{logger_name}_stream_handler'
+    for handler in log.handlers:
+        if handler.name == handler_name:
+            log.removeHandler(handler)
+
+
+class PromptToolkitHandler(logging.StreamHandler):
```
kyleknap, line 81: It feels a bit strange that we have promptoolkit/autoprompt logic in this `logger` module. I wonder if we should figure out how to move this class to the `autoprompt` module and maybe figure out how we can apply the handler in a context manager in the `AutoPromptDriver.prompt_for_args` method? Another thing that seemed a little off was that we are applying this prompt toolkit handler even if we are not configured to use auto-prompt at all.

kyleknap (commented):
It was working well for me. I just mainly had some suggestions/questions and found a one issue we will need to fix.

awscli/clidriver.py
```diff
@@ -464,30 +474,28 @@ def _show_error(self, msg):
...
             LOG.debug("CLI version: %s", self.session.user_agent())
             LOG.debug("Arguments entered to CLI: %s", sys.argv[1:])
-
         else:
-            self.session.set_stream_logger(logger_name='awscli',
-                                           log_level=logging.ERROR)
+            remove_stream_logger('botocore')
+            remove_stream_logger('awscli')
+            remove_stream_logger('s3transfer')
+            remove_stream_logger('urllib3')
```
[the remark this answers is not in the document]
vz10: In case we added `--debug` before we get into autoprompt mode and removed it during editing and before actual command run

awscli/autoprompt/core.py
```diff
@@ -63,8 +63,11 @@ def resolve_mode(self, args):
             return 'off'
         if self._CLI_AUTO_PROMPT_OPTION in args:
             return 'on'
-        config = self._session.get_config_variable('cli_auto_prompt')
-        return config.lower()
+        try:
+            config = self._session.get_config_variable('cli_auto_prompt')
```
kyleknap, line 67: We should add a test for if the profile is not found.

awscli/autoprompt/logger.py
```diff
@@ -0,0 +1,64 @@
...
+                    Document(text=msg), bypass_readonly=True
+                )
+            else:
+                super().emit(record)
+        except:
+            self.handleError(record)
+
+
+class LoggersHandlerSwitcher:
+    def __init__(self, handler=PromptToolkitHandler(),
```
kyleknap, line 41: We should just make the handler required. Otherwise, we have a mutable default object in the initializer and this class is not as reusable.

awscli/autoprompt/logger.py
```diff
@@ -0,0 +1,64 @@
...
+                if current_document:
+                    msg = '\n'.join([current_document, msg])
+                debug_buffer.set_document(
+                    Document(text=msg), bypass_readonly=True
+                )
+            else:
+                super().emit(record)
+        except:
+            self.handleError(record)
+
```
kyleknap, line 38: I'm wondering if instead of making a generalized class we just make a new method on this class that is a context manager (e.g. `capture_all_logging()`) and does the logic from the `LoggersHandlersSwitcher` class? I think that makes the logic in the `prompttoolkit.py` logic a bit easier to follow and I'm not entirely sure if we need something generalized right now.

awscli/autoprompt/logger.py
```diff
@@ -0,0 +1,64 @@
...
+
+class LoggersHandlerSwitcher:
+    def __init__(self, handler=PromptToolkitHandler(),
+                 log_level=logging.DEBUG,
+                 format_string=LOG_FORMAT):
+        self._handler = handler
+        self._handler.setLevel(log_level)
+        self._handler.setFormatter(logging.Formatter(format_string))
+        self._old_handlers = {}
```
kyleknap, line 47: So I'm not sure how reusable this class is because it looks like we do not clear the `old_handlers` cache when we exit the context manager. If we want to continue making this a generalized abstraction, I think we should clear this dictionary when we exit.

awscli/clidriver.py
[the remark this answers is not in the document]
kyleknap: That's fair. I don't think I have strong opinions on what the behavior, but if I was to take a side, I think that makes sense.

awscli/clidriver.py
```diff
@@ -465,29 +476,31 @@ def _show_error(self, msg):
...
+                              format_string=LOG_FORMAT)
+            set_stream_logger('urllib3', logging.DEBUG,
+                              format_string=LOG_FORMAT)
             LOG.debug("CLI version: %s", self.session.user_agent())
             LOG.debug("Arguments entered to CLI: %s", sys.argv[1:])
-
         else:
-            self.session.set_stream_logger(logger_name='awscli',
-                                           log_level=logging.ERROR)
+            remove_stream_logger('botocore')
```
kyleknap, line 498: To make this a bit easier to manage, let's:
1) Keep a list of packages we need to add handlers for and programmatically add and remove handlers using this list
2) Add a comment in the code on why we need to remove the stream handlers. It definitely was not obvious why it was needed.

kyleknap (commented):
It's looking good. I just had some more minor comments.

PR #5708, opened 2020-11-10
Prompt output panel

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autocomplete/completer.py
+++ b/awscli/autocomplete/completer.py
@@ -29,6 +29,9 @@ class AutoCompleter(object):
         self._parser = parser
         self._completers = completers
 
+    def parse(self, command_line, index=None):
+        return self._parser.parse(command_line, index)
+
     def autocomplete(self, command_line, index=None):
         """Attempt to find completion suggestions.
 
@@ -39,7 +42,7 @@ class AutoCompleter(object):
         :return: A list of ``CompletionResult`` objects.
 
         """
-        parsed = self._parser.parse(command_line, index)
+        parsed = self.parse(command_line, index)
         for completer in self._completers:
             result = completer.complete(parsed)
             if result is not None:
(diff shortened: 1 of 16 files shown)


kyleknap (commented):
Just going to put some of the higher level notes from playing around from this. It is looking pretty neat. I broke down the feedback in terms of behavior suggestions and edge cases that I found. Note I never really got to try the query behavior because of the first point in the edge cases list where it kept freezing up on me.

**Behavior suggestions**
* YAML output looks a little off. Specifically in dictionary is being represented as an `!!omap`. We should make that look like a normal yaml dictionary. You can run the following command to see it: `aws dynamodb describe-table --output yaml`
* It would be great if we can use the search bar from the docs to maneuver through the output panel. This is helpful for outputs that are really large.
* It would be nice to be able to use the `q` button to exit the output panel. It does not seem that is implemented and instead I need to use `F2` to toggle back to the prompt. 


**Edge cases**
* Whenever I type a command out, the prompt freezes on me and one point I ran into a segfault? Any idea on what is happening here? I've never ran into this before when using any of the releases auto-prompt mode.
  ```
  $ aws
  > aws ec2 dSegmentation fault: 11
  ```
* Auto-prompt mode crashes for custom commands like if you type `s3 ls`
* The `s3api` commands do not work because it seems like it is trying to search for `s3api` in the available service commands. This causes auto-prompt to crash as well.

kyleknap (commented):
It's looking better. I'm still hitting the issue where it freezes on me when I type instead of use the tab completion, and would be nice to fix the ordered map representation for yaml. Otherwise, I think one more thing that would be nice is if we introduced an error bar that we can toggle off an on that details any issues with the query expression.

I'm mainly suggesting this because it is sort of difficult to tell if and how you are messing up your JMESPath expression as if you update the expression to be invalid it retains the last valid expression. So for example I updated my prompt from a valid query to be:
```
aws s3api list-buckets --output json --query Buckets['Name']
```
And the output panel showed a valid output but when I ran it, it failed with:
```
Bad value for --query Buckets[Name]: Expecting: star, got: unquoted_identifier: Parse error at column 8, token "Name" (UNQUOTED_IDENTIFIER), for expression:
"Buckets[Name]"
```
In general, it would have been nicer if I told me that the query I had is just invalid and not going to work.

PR #5711, opened 2020-11-11
Update add-tags-to-resource.rst

aidenvaines-bjss:
*Issue #, if available:*
N/A

*Description of changes:*
The example command was badly formatted, the tagging structure didn't work and the output example was wrong with current AWS CLI. I believe this should be correct now.


kdaily:
Thanks for the change, @aidenvaines-bjss. I'm going to check with the doc writers about the key/value choices in this example to see which are the ones that should be used.

PR #5714, opened 2020-11-11
 Fix bug when configure sso command does not appear in docs

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/commands.py
+++ b/awscli/customizations/commands.py
@@ -223,7 +223,8 @@ class BasicCommand(CLICommand):
             subcommand_name = subcommand['name']
             subcommand_class = subcommand['command_class']
             subcommand_table[subcommand_name] = subcommand_class(self._session)
-        self._session.emit('building-command-table.%s' % self.NAME,
+        name = '_'.join([c.name for c in self.lineage])
+        self._session.emit('building-command-table.%s' % name,
                            command_table=subcommand_table,
                            session=self._session,
                            command_object=self)
@@ -254,7 +255,8 @@ class BasicCommand(CLICommand):
 
     def _build_arg_table(self):
         arg_table = OrderedDict()
-        self._session.emit('building-arg-table.%s' % self.NAME,
+        name = '_'.join([c.name for c in self.lineage])
+        self._session.emit('building-arg-table.%s' % name,
                            arg_table=self.ARG_TABLE)
         for arg_data in self.ARG_TABLE:
 
--- a/tests/functional/configure/test_addmodel.py
+++ b/tests/functional/configure/test_addmodel.py
@@ -82,12 +82,3 @@ class TestAddModel(BaseAWSCommandParamsTest):
             os.path.exists(os.path.join(
                 self.customer_data_root, 'override-name', '2015-12-02',
                 'service-2.json')))
-
-
-class TestConfigureSSO(BaseAWSCommandParamsTest):
-    prefix = 'configure sso'
-
-    def test_sso_logout_do_not_exist(self):
-        cmdline = self.prefix + ' logout'
-        _, err, _ = self.run_cmd(cmdline, 252)
-        self.assertIn('Unknown options', err)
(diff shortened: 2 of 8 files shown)


awscli/customizations/commands.py
```diff
@@ -223,7 +228,8 @@ def _build_subcommand_table(self):
             subcommand_name = subcommand['name']
             subcommand_class = subcommand['command_class']
             subcommand_table[subcommand_name] = subcommand_class(self._session)
-        self._session.emit('building-command-table.%s' % self.NAME,
+        name = self.EVENT_NAME or self.NAME
```
kyleknap, line 231: This is probably the most minimal change to not impact, but I'm wondering if we should just make it even more general and not even require the an override. I'm mainly asking this because if we do solve this without needing an override, we may be able to fix other issues. For example, the `cloudformation deploy` command collides with the `deploy` service command, so you get weird behavior such as this:
```
$ aws cloudformation deploy dsfd

  aws help
  aws <command> help

aws: error: argument subcommand: Invalid choice, valid choices are:

push                                     | register                                
deregister                               | install                                 
uninstall                               
```
where because the event collides with custom code deploy commands the code deploy commands are injected into `cloudformation deploy`'s command table and causes the command to try to parse it as a command.

As a solution, I'm wondering if we can join the command lineage together on a character that is not used in commands such as underscore to use as the name for the event? So for example we would just make the line:
```python
self._session.emit('building-command-table.%s' % '_'.join([c.name for c in self.lineage])
```
This should avoid any potential future collisions. The one downside is that we may have to update some preexisting handler registrations but hopefully should not be too many.

Also, note I'm not suggesting to use `.` as what to join on because that could introduced undesired behavior where trigger handlers that are intended to be used only for the service command are also triggered for operation commands due to how the event hierarchy works.

Thoughts?

tests/unit/customizations/test_commands.py
```diff
@@ -136,6 +142,25 @@ def test_expected_events_are_emitted_in_order(self):
...
+
+        self.assert_events_fired_in_order([
+            'building-command-table.main',
+            'building-top-level-params',
+            'top-level-args-parsed',
+            'session-initialized',
+            'building-command-table.s3',
+            # here should be use event_name
+            'building-command-table.foo-command',
+            'building-arg-table.foo',
```
kyleknap, line 160: Whatever we decide for the `building-command-table` event we should probably apply to it to this event as well because it may exhibit similar issues with collisions in the future.

tests/unit/customizations/test_commands.py
```diff
@@ -136,6 +142,25 @@ def test_expected_events_are_emitted_in_order(self):
             'before-building-argument-table-parser.s3.foo'
         ])
 
+    def test_expected_events_are_emitted_in_order_with_event_name(self):
```
kyleknap, line 145: We should also add a functional test specficially for sso to make sure we are not adding to the `configure sso` command table. Oddly enough a command like this:
```
aws configure sso logout help
```
currently works in the CLI and it should not. Maybe we can add a functional test to make sure `aws configure sso logout` is an invalid command?

kyleknap (commented):
Looks fine. I just had a few suggestions.

tests/functional/configure/test_addmodel.py
```diff
@@ -82,3 +82,12 @@ def test_add_model_with_service_name(self):
             os.path.exists(os.path.join(
                 self.customer_data_root, 'override-name', '2015-12-02',
                 'service-2.json')))
+
+
+class TestConfigureSSO(BaseAWSCommandParamsTest):
```
kyleknap, line 87: Let's add this test in a new `test_sso.py` file in the `configure` functional test suite instead of using the `test_addmodel.py`. The tests are typically organized by which command they apply to so it is weird to be adding a test case for `configure sso` to a test file meant only for `configure add-model`

kyleknap (commented):
Looks good. I'm glad that we did not have to switch too many handlers around. I just had one more comment about the new functional test.

kyleknap (approved):
:ship:

PR #5719, opened 2020-11-12
Add output panel to autoprompt mode

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/factory.py
+++ b/awscli/autoprompt/factory.py
@@ -22,14 +22,15 @@ from prompt_toolkit.layout.dimension import Dimension
 from prompt_toolkit.layout.layout import Layout, ConditionalContainer
 from prompt_toolkit.layout.menus import CompletionsMenu, MultiColumnCompletionsMenu
 from prompt_toolkit.layout.processors import BeforeInput
-from prompt_toolkit.widgets import HorizontalLine, SearchToolbar
+from prompt_toolkit.widgets import SearchToolbar, VerticalLine
+from prompt_toolkit.key_binding.bindings.focus import focus_next
 
 from awscli.autoprompt.history import HistoryDriver, HistoryCompleter
 from awscli.autoprompt.widgets import (
-    HelpPanelWidget, ToolbarWidget, DebugPanelWidget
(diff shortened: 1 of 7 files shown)


awscli/autoprompt/factory.py
```diff
@@ -147,30 +151,36 @@ def create_layout(self, on_input_buffer_text_changed=None,
...
-                                             doc_section_visible),
+                        VSplit([
+                            ConditionalContainer(doc_window,
+                                                 doc_section_visible),
+                            ConditionalContainer(
+                                VerticalLine(),
+                                output_section_visible & doc_section_visible),
+                            ConditionalContainer(output_window,
+                                                 output_section_visible),
+                        ]),
```
kyleknap, line 179: To clean up the top level view of the layout, it would be nice if we just represented this `VSplit` as a single variable. Maybe call it `bottom_window` and have it be returned from the `create_bottom_window` method? Then we can rename the current `create_bottom_window` method that creates the different types of windows to `create_searchable_window()`?

awscli/autoprompt/widgets.py
```diff
@@ -48,11 +50,32 @@ def apply_transformation(self, text_input):
...
+    def __init__(self, title):
+        fill = partial(Window, style='class:frame.border')
+        self.container =  VSplit([
+            fill(char=Border.HORIZONTAL),
+            fill(width=1, height=1, char='|'),
+            # Notice: we use `Template` here, because `self.title` can be an
+            # `HTML` object for instance.
+            Label(lambda: Template(' {} ').format(title),
+                  style='class:frame.label',
+                  dont_extend_width=True),
```
kyleknap, line 64: Do we actually care about it being an `HTML` object? Could we just specify `title` here instead of the lambda?

kyleknap (commented):
Looks fine. Just had a couple comments. I think this should be set once we update it as well to not share the auto-complete index connection which was what was causing the freezing issues on my side.

kyleknap (approved):
Looks good. It would be great before merging this to get the additional commits into the upstream `prompt-output-panel` so that this PR only has the single commit about adding the UI. It is mainly useful for tracking what exactly the intended change was if we ever need to refer back to this PR. Otherwise, :ship:

PR #5722, opened 2020-11-14
Generate indexes for ac index tables

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/unit/autocomplete/local/test_indexer.py
+++ b/tests/unit/autocomplete/local/test_indexer.py
@@ -67,7 +67,7 @@ class DummyArg(object):
         self.help_text = help_text
 
 
-class TestCanRetrieveCommands(unittest.TestCase):
+class BaseIndexerTest(unittest.TestCase):
     def setUp(self):
         self.session = mock.Mock(spec=Session)
         self.aws_command = DummyCommand(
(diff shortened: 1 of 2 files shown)


tests/unit/autocomplete/local/test_indexer.py
```diff
@@ -215,3 +215,18 @@ class TestCanCreateModelIndexer(unittest.TestCase):
     def test_can_create_model_indexer(self):
         index = indexer.create_model_indexer('/tmp/a/b/c/d')
         self.assertIsInstance(index, indexer.ModelIndexer)
+
+
+class TestGeneratesIndex(TestCanRetrieveCommands):
```
kyleknap, line 220: Is there a particular reason we are subclassing from the `TestCanRetrieveCommands`? One unintentional side affect of doing this is that it will cause us to rerun all of the test cases from `TestCanRetrieveCommands` as well as the new one we added. We probably should either just:

1. Rename `TestCanRetrieveCommands` to something more general and add the `test_generates_indexes_for_tables` to that command class.

Or

2. Create a base class that does the `setUp` and `tearDown` from `TestCanRetrieveCommands` and have this class subclass from this new base class so we do not rerun the tests from `TestCanRetrieveCommands` again.

kyleknap (commented):
Looks fine. I just had a suggestion about the test.

kyleknap (approved):
Thanks! Feel free to merge it once CI passes :ship:

PR #5723, opened 2020-11-15
Move getting index file name to DatabaseConnection class

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autoprompt/prompttoolkit.py
+++ b/awscli/autoprompt/prompttoolkit.py
@@ -21,6 +21,8 @@ from prompt_toolkit.completion import Completion
 from prompt_toolkit.document import Document
 
 from awscli.logger import LOG_FORMAT
+from awscli.autocomplete import parser
+from awscli.autocomplete.local import model
 from awscli.autoprompt.doc import DocsGetter
 from awscli.autoprompt.factory import PromptToolkitFactory
 from awscli.autoprompt.logger import PromptToolkitHandler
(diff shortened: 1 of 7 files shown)


awscli/autoprompt/prompttoolkit.py
```diff
@@ -131,7 +134,7 @@ def _set_app_defaults(self, app):
         app.debug = False
 
     def _update_doc_window_contents(self, *args):
-        parsed_args = self._completion_source.parse(
+        parsed_args = self._parser.parse(
```
kyleknap, line 137: We should probably remove the public `parse()` method on the `AutoCompleter` class since that is not really used anymore so that we do not have any unnecessary interfaces.

awscli/autoprompt/prompttoolkit.py
```diff
@@ -55,7 +57,7 @@ class PromptToolkitPrompter:
 
     """
     def __init__(self, completion_source, driver, completer=None,
-                 factory=None, app=None):
+                 factory=None, app=None, index_filename=None):
```
kyleknap, line 60: Instead of accepting the `index_filename`, it probably makes a bit more sense to accept a `cli_parser` and then create one if `None` is specified. Mainly, this class should not really know that a model index is used rather it only really cares about having a parser that it can use to parse the results from the input buffer and that could use whatever implementation.

kyleknap (commented):
Looks fine. Just had a couple of suggestions.

kyleknap (approved):
Looks good. :ship: Feel free to merge it once CI passes

PR #5724, opened 2020-11-15
 Add output getter and query prompter

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/autocomplete/local/basic.py
+++ b/awscli/autocomplete/local/basic.py
@@ -496,3 +496,84 @@ class ShorthandCompleter(BaseCompleter):
                     )
                 )
         return self._filter(last_key, results)
+
+
+class QueryCompleter(BaseCompleter):
+
+    def __init__(self, cli_driver_fetcher=None,
+                 response_filter=startswith_filter):
+        self._filter = response_filter
+        self._cli_driver_fetcher = cli_driver_fetcher
+        self._argument_generator = None
+        self._jmespath = None
+
+    @property
+    def jmespath(self):
+        if self._jmespath is None:
+            import jmespath
+            self._jmespath = jmespath
+        return self._jmespath
+
+    @property
+    def argument_generator(self):
+        if self._argument_generator is None:
+            from botocore.utils import ArgumentGenerator
+            self._argument_generator = ArgumentGenerator
+        return self._argument_generator
+
+    def complete(self, parsed):
+        if self._cli_driver_fetcher is None:
+            return
+        if parsed.current_param == 'query' and \
+                parsed.current_fragment is not None:
+            operation_model = self._cli_driver_fetcher.get_operation_model(
+                parsed.lineage, parsed.current_command)
+            if operation_model:
+                return self._get_completions(parsed.current_fragment,
+                                             operation_model)
+
+    def _get_query_and_last_key(self, query):
+        # Because output example has only 1 element in any list if
+        # user enters query like Groups[2] jmespath will return "null"
+        # and we won't be able to prompting so we change all the numbers
+        # in brackets to 0 to be able to get the shape of the nested structure
+        query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>0\g<2>', query)
+        return query.rsplit('.', 1)
+
+    def _create_completions(self, results, last_key, fragment):
+        completions = self._filter(
+            last_key,
+            [CompletionResult(last_key, display_text=result)
+             for result in results]
+        )
+        for completion in completions:
+            name_part_len = len(fragment) - len(completion.name)
+            completion.name = "%s%s" % (
+                fragment[:name_part_len],
+                completion.display_text
+            )
+        return completions
+
+    def _get_completions(self, fragment, operation_model):
+        results = []
+        last_key = fragment
+        if operation_model.output_shape:
+            argument_generator = self.argument_generator(
+                use_member_names=True)
+            response = argument_generator.generate_skeleton(
+                operation_model.output_shape)
+            if '.' not in fragment:
+                if isinstance(response, dict):
+                    results = response.keys()
+            else:
+                try:
+                    query, last_key = self._get_query_and_last_key(fragment)
+                    expression = self.jmespath.compile(query)
+                    parsed_response = expression.search(response)
+                    if parsed_response and isinstance(parsed_response, list):
+                        parsed_response = parsed_response[0]
+                    if isinstance(parsed_response, dict):
+                        results = parsed_response.keys()
+                except self.jmespath.exceptions.JMESPathError:
+                    pass
+        return self._create_completions(results, last_key, fragment)
--- a/awscli/autocomplete/local/fetcher.py
+++ b/awscli/autocomplete/local/fetcher.py
@@ -13,17 +13,26 @@
 
 
 class CliDriverFetcher:
+
     def __init__(self, cli_driver):
         self._cli_driver = cli_driver
 
-    def _get_argument(self, lineage, current_command, arg_name):
+    def _get_command(self, lineage, current_command):
         subcommand_table = self._cli_driver.subcommand_table
         for arg in lineage[1:]:
             subcommand = subcommand_table[arg]
             subcommand_table = subcommand.subcommand_table
-        command = subcommand_table.get(current_command)
+        return subcommand_table.get(current_command)
+
+    def _get_argument(self, lineage, current_command, arg_name):
+        command = self._get_command(lineage, current_command)
         if command:
             return command.arg_table.get(arg_name)
+
+    def get_operation_model(self, lineage, current_command):
+        if len(lineage) > 1 and current_command:
+            command = self._get_command(lineage, current_command)
+            return command.create_help_command().obj
 
     def get_argument_model(self, lineage, current_command, arg_name):
         return getattr(self._get_argument(
--- a/awscli/autocomplete/parser.py
+++ b/awscli/autocomplete/parser.py
@@ -126,8 +126,9 @@ class CLIParser(object):
     not a general purpose AWS CLI parser.
 
     """
-    def __init__(self, index):
+    def __init__(self, index, return_first_command_match=False):
         self._index = index
+        self._return_first_command_match = return_first_command_match
 
     def parse(self, command_line, location=None):
         """Parses as much of the command line input as possible.
@@ -286,16 +287,15 @@ class CLIParser(object):
         return any(command.startswith(current) and command != current
                    for command in command_names)
 
-    def _handle_positional(self, current, state, remaining_parts, parsed):
-        # This is can either be a subcommand or a positional argument
+    def _is_command_name(self, current, remaining_parts, command_names):
+        # If _return_first_command_match is True
         #
-        # First we can check if this is a valid subcommand given our lineage.
-        # We're one off here, we need to compute a new *potential*
-        # lineage.
-        command_names = self._index.command_names(state.full_lineage)
-        positional_argname = None
-        is_command_name = current in command_names
-        is_part_of_command = self._is_part_of_command(current, command_names)
+        # We just check if the 'current' is in 'command_names'
+        # and don't care if there any other commands that could have 'current'
+        # as a part
+        #
+        # If _return_first_command_match is False
+        #
         # To decide if 'current' part is a command or not we consider such cases
         # - if 'current' is in 'command_names' and we already moved forward (at
         # least entered ' ' after) then it's a command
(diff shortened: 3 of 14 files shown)


awscli/autocomplete/local/fetcher.py
```diff
@@ -25,6 +27,18 @@ def _get_argument(self, lineage, current_command, arg_name):
...
 
+    def get_operation_model(self, lineage, current_command, operation):
+        if current_command and len(lineage) > 1:
+            try:
+                service_name = self.SERVICE_MAPPER.get(lineage[1], lineage[1])
+                service_model = self._cli_driver.session.get_service_model(
+                    service_name)
+                operation = ''.join([part.capitalize()
+                                     for part in current_command.split('-')])
+                return service_model.operation_model(operation)
```
kyleknap, line 38: There are other services that are renamed and therefore do not match what the client name is in botocore. I wonder instead if we can just grab the `service_model` or `operation_model` attached to the command object to make this handle renames more generically?

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,75 @@
...
+        if parsed.current_param == 'query':
+            query = parsed.current_fragment
+        else:
+            query = parsed.global_params.get('query')
+        if query:
+            query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>\g<2>', query)
+        try:
+            self._current_expression = jmespath.compile(query)
+        except Exception:
+            pass
```
kyleknap, line 65: Instead of swallowing the exception. I wonder if it makes sense to actually display the parsing error in the output panel? Mainly there is a couple of times where I provided an invalid jmespath expression in auto-prompt mode but I did not know until I actually ran the command. It would have been nice if there was some output to tell me that I formatted it wrong.

kyleknap (commented):
It's looking good. I just had some suggestions on the functionality after trying it out a bit.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,75 @@
...
+            output = self._session.get_config_variable('output')
+        return output
+
+    def _get_query(self, parsed):
+        if parsed.current_param == 'query':
+            query = parsed.current_fragment
+        else:
+            query = parsed.global_params.get('query')
+        if query:
+            query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>\g<2>', query)
```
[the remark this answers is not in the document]
vz10: What about to change any number in brackets to 0

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,75 @@
...
+            self._current_expression = jmespath.compile(query)
+        except Exception:
+            pass
+        return self._current_expression
+
+    def _get_display(self, operation, output_shape, output, query):
+        args = argparse.Namespace(query=query, color='off')
+        argument_generator = ArgumentGenerator(use_member_names=True)
+        response = argument_generator.generate_skeleton(output_shape)
+        formatter = get_formatter(output, args)
```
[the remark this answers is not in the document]
vz10: I remember. I still can't understand why they appear in this output but don't appear in the normal one.

awscli/autoprompt/output.py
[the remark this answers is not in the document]
kyleknap: That is fine too.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,89 @@
...
+        self._cli_driver_fetcher = CliDriverFetcher(driver)
+        self._output_formats = driver.arg_table['output'].choices
+        self._session = driver.session
+        self._cache = {}
+        self._current_expression = None
+
+    def get_output(self, parsed):
+        operation_model = self._cli_driver_fetcher.get_operation_model(
+            parsed.lineage, parsed.current_command)
+        if operation_model:
```
kyleknap, line 35: Not sure where this is happening now but for some command, it shows up as `No output` but after I add a space after the command the output shows up. Once command was: `aws deploy get-deployment` If I do not include a space, it suggests `No output`, but if I add a space `aws deploy get-deployment ` then it shows the normal output.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,89 @@
...
+            query = parsed.current_fragment
+        else:
+            query = parsed.global_params.get('query')
+        if query:
+            query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>0\g<2>', query)
+            query = re.sub(r'([\{\[])[^\]^\}]*$', '', query).strip('.')
+        try:
+            self._current_expression = jmespath.compile(query)
+        except jmespath.exceptions.ParseError as e:
+            error_message = e.msg
```
kyleknap, line 75: I like where this is going. It would be nice if this error was a bit more formatted instead of just the literal error message. Maybe something like:
```
ERROR: Unable to display output.

Bad value for --query: <query expression>

<Add helper string for where the error is>

<Insert error message here>
```
In the end, I'd imagine it would behave similarly to what the CLI does with its error messages when it encounters invalid JMESPath expressions. So for example:
```
aws ec2 describe-regions --query Regions[0]..Endpoint

Bad value for --query Regions[0]..Endpoint: Expecting: ['quoted_identifier', 'unquoted_identifier', 'lbracket', 'lbrace'], got: dot: Parse error at column 11, token "." (DOT), for expression:
"Regions[0]..Endpoint"
            ^

```

awscli/autoprompt/output.py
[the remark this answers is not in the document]
kyleknap: We may have to update the `ArgumentGenerator` in `botocore`. I know it uses ordered dictionaries when really do not have to do that anymore because dictionaries are now ordered for the minimum python version that we support.

awscli/autocomplete/local/basic.py
```diff
@@ -496,3 +496,78 @@ def _get_struct_keys_completions(self, arg_model, entered_keys,
...
+        if parsed.current_param == 'query' and \
+                parsed.current_fragment is not None:
+            operation_model = self._cli_driver_fetcher.get_operation_model(
+                parsed.lineage, parsed.current_command)
+            if operation_model:
+                return self._get_completions(parsed.current_fragment,
+                                             operation_model)
+
+    def _get_query_and_last_key(self, query):
+        query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>0\g<2>', query)
```
kyleknap, line 534: A comment here explaining what the query does would be nice.

awscli/autocomplete/local/basic.py
```diff
@@ -496,3 +496,78 @@ def _get_struct_keys_completions(self, arg_model, entered_keys,
...
                 )
         return self._filter(last_key, results)
+
+
+class QueryCompleter(BaseCompleter):
+
+    def __init__(self, cli_driver_fetcher=None,
+                 response_filter=startswith_filter):
+        self._filter = response_filter
+        self._cli_driver_fetcher = cli_driver_fetcher
```
kyleknap, line 506: One thing that I ran into was that I tried this with the tab completer (e.g. `aws_completer). Specifically, I tried:
```
aws ec2 describe-regions --query <TAB>
```
And it printed out the following Traceback:
```
  File "/Users/kyleknap/.pyenv/versions/pyinstaller-3.7.2/bin/aws_completer", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/kyleknap/GitHub/aws-cli/bin/aws_completer", line 36, in <module>
    main()
  File "/Users/kyleknap/GitHub/aws-cli/bin/aws_completer", line 28, in main
    autocomplete(command_line, command_index)
  File "/Users/kyleknap/GitHub/aws-cli/awscli/autocomplete/main.py", line 55, in autocomplete
    results = completer.autocomplete(command_line, position)
  File "/Users/kyleknap/GitHub/aws-cli/awscli/autocomplete/completer.py", line 44, in autocomplete
    result = completer.complete(parsed)
  File "/Users/kyleknap/GitHub/aws-cli/awscli/autocomplete/local/basic.py", line 527, in complete
    operation_model = self._cli_driver_fetcher.get_operation_model(
AttributeError: 'NoneType' object has no attribute 'get_operation_model'
```
I think at a minimum it should be not allow an error to propagate like this.

awscli/autocomplete/local/basic.py
```diff
@@ -496,3 +496,78 @@ def _get_struct_keys_completions(self, arg_model, entered_keys,
...
+            else:
+                try:
+                    query, last_key = self._get_query_and_last_key(fragment)
+                    expression = self.jmespath.compile(query)
+                    parsed_response = expression.search(response)
+                    if parsed_response and isinstance(parsed_response, list):
+                        parsed_response = parsed_response[0]
+                    if isinstance(parsed_response, dict):
+                        results = parsed_response.keys()
+                except Exception:
```
kyleknap, line 571: What is the exception we are expecting to catch here? We should probably be specific here so we do not accidentally swallow any unintended issues.

awscli/autoprompt/output.py
kyleknap: Following up here... It looks like if the command is a substring of another command (e.g. `get-deployment` is a substring of `get-deployment-config`) that is when it shows up of having `No output` when toggling to the command that is a substring in the operation list. 

It looks like this is an issue with the doc panel as well. Is this something easy to fix?

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def _get_query(self, parsed):
+        error_message = None
+        if parsed.current_param == 'query':
+            query = parsed.current_fragment
+        else:
+            query = parsed.global_params.get('query')
+        if query:
+            query = re.sub(r'([\{\[])\d+?([\}\]])', '\g<1>0\g<2>', query)
+            query = re.sub(r'([\{\[])[^\]^\}]*$', '', query).strip('.')
```
kyleknap, line 70: Comments on what these regex's do would be nice.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,88 @@
...
+        try:
+            self._current_expression = jmespath.compile(query)
+        except jmespath.exceptions.ParseError as e:
+            error_message = "Bad query %s: %s" % (query, str(e))
```
kyleknap, line 74: For the error message, Let's make it instead:
```python
error_message = "Bad value for --query: %s\n\n%s" % (query, str(e))
```
I think the main points is that we should use `--` to indicate it is the `--query` parameter and it will add newlines to the error to help make it a little more readable.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap, line 1: The year should be 2020 here.

kyleknap (commented):
Looks good. I just had a bit more suggestions/questions, but I think we are getting close. I think one thing I notice was there were no tests for `autoprompt/output.py`. Was that intentional? I think there is a fair amount of behavior in that module we would want to have tests for.

awscli/autocomplete/local/basic.py
vz10: Actually I don't expect any other exceptions here added it just in case.

kyleknap: Does it makes sense to log it then? I just want to have just some mechanism to figure out what is going wrong if we do a catch an unexpected exception.

tests/unit/autoprompt/test_output.py
```diff
@@ -0,0 +1,99 @@
...
+                                False),
+                        }
+                    },
+                    'aws.s3api': {
+                        'get-object': {}
+                    }
+                }
+            })
+        self.parser = parser.CLIParser(index)
+        self.driver = create_clidriver()
```
kyleknap, line 56: We should make sure that for these tests we are using a stubbed environment. It is possible to set in your `~/.aws/config` file the output format. So we do not want to unintentionally be using that when some of these tests assume that the fallback is configured to json.

awscli/autoprompt/output.py
```diff
@@ -0,0 +1,104 @@
...
+        return shape and (getattr(shape, 'members', False) or
+                          getattr(shape, 'member', False))
+
+    def _get_output(self, parsed):
+        if parsed.current_param == 'output':
+            output = parsed.current_fragment
+        else:
+            output = parsed.global_params.get('output')
+        if output not in self._output_formats:
+            output = self._session.get_config_variable('output')
```
kyleknap, line 63: One subtle thing that I missed in previous reviews that I don't think we discussed. Do we want to be reverting back to the configured variable for invalid output formats, or should we print out an error message saying that the output format is invalid (similar to what we do with the `--query` expression)? The main benefit of what we have now is that while someone is typing the value to  `--output` won't show an error message, but I wonder if that is a strong enough of a reason to warrant this behavior? Like if I had a typo in the output value, I would prefer that the output panel tell me about it before I try to execute the command. Thoughts?

awscli/autocomplete/parser.py
```diff
@@ -126,8 +126,9 @@ class CLIParser(object):
     not a general purpose AWS CLI parser.
 
     """
-    def __init__(self, index):
+    def __init__(self, index, first_wins=False):
```
kyleknap, line 129: I feel like we can make a better name here. Mainly it is a little ambiguous about what is first and how does it win. How about `return_first_command_match`? It's a little lengthy but I think it represents what we are going for. I also think it might be worth adding docstrings to describe the parameter because it is a pretty specific parameter.

kyleknap (commented):
Looks good. Just had some more suggestions based on the updates, but it is looking good.

kyleknap (approved):
Looks great! :ship:

PR #5728, opened 2020-11-17
[v2] Initial groundwork for wizard improvements

kyleknap:
This pull request adds the initial groundwork for wizard improvements. This change allows the prompts to be displayed as a basic form-based UI that you can type input into. To test it, I updated the `cli-dev wizard-dev` to start using this new scaffolding. So you can try it out by inputting:
```
aws cli-dev wizard-dev --run-wizard file://path/to/wizard.yml
```
I tried to break this as best as I could into separate commits to accurately represent what changes I made but to summarize:
* Added a new `ui` package to host anything related to prompt toolkit UI.
* Added a new wizard app that uses the UI components to make a basic form-based wizard UI
* Updated the `cli-dev wizard-dev` command to use this UI
* Generalized the prompt toolkit application stubber to use for wizards
* Added basic behavior tests for the wizard application as a starting point for adding future tests. (I still want to add more once the UI starts becoming finalized)
* Updated the auto-prompt tests to use the updated prompt toolkit application stubber.

**Future work:**
There is still a bit of work that is needed for the UI. In terms of immediate plans, I will next be adding:
* Section pages to show prompts only for a specific section
* Tighter control over which prompt to move next to based on wizard definition
* Visibility for certain prompts based on a previous selection.


tests/__init__.py
```diff
@@ -320,3 +328,102 @@ def stream(self, **kwargs):
...
+        self._app = app
+        self._app.input = FakeApplicationInput()
+        self._queue = []
+
+    def add_keypress(self, key, app_assertion=None):
+        self._queue.append(AppKeyPressAction(key))
+        if app_assertion:
+            failure_message_format = (
+                f'Incorrect action on key press "{key}": '
+                '{message}'
```
vz10, line 369: Need to add `f` on this iine as well

awscli/customizations/wizard/ui/layout.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+
+    def _create_title(self, definition):
+        title = Label(f'{definition["title"]}', style='class:wizard.title')
+        title.window.align = WindowAlign.CENTER
+        return title
+
+    def _create_all_prompts(self, definition):
+        prompts = []
+        for _, step_definition in definition['plan'].items():
```
vz10, line 41: Probably using `.values()` would be more clear

awscli/customizations/wizard/ui/prompt.py
```diff
@@ -0,0 +1,120 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from prompt_toolkit.application import get_app
+from prompt_toolkit.buffer import Buffer
+from prompt_toolkit.document import Document
+from prompt_toolkit.key_binding import KeyBindings
+from prompt_toolkit.layout.containers import (
+    Window, VSplit, Dimension
```
vz10, line 18: [NIT] I believe it fits in one line

awscli/customizations/wizard/ui/prompt.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class WizardPromptAnswer:
+    def __init__(self, value_name):
+        self._value_name = value_name
+        self.buffer = self._get_answer_buffer()
+        self.container = self._get_answer_container()
+
+    def _get_answer_buffer(self):
+        return Buffer(name=f'{self._value_name}_answer_buffer')
```
vz10, line 104: What the reason of giving name to this Buffer if we never use this name?

I was wondering if it makes sense to name the buffer just `f'{self._value_name}` and make keybindings global instead of injecting them into each `WizardPrompt` 

Something like

```python

def submit_current_answer(event):
    current_buffer = event.app.current_buffer.name
    event.app.values[current_buffer.name] = .current_buffer.text

@kb.add('tab')
@kb.add('enter')
def next_prompt(event):
    submit_current_answer(event)
    event.app.layout.focus_next()

@kb.add('s-tab')
def last_prompt(event):
    submit_current_answer(event)
    event.app.layout.focus_last()
```

awscli/customizations/wizard/ui/prompt.py
```diff
@@ -0,0 +1,120 @@
...
+        self._value_definition = value_definition
+        self.container = self._get_container()
+
+    def _get_container(self):
+        answer_container = WizardPromptAnswer(self._value_name)
+        answer_buffer = answer_container.buffer
+        return VSplit(
+            [
+                WizardPromptDescription(
+                    answer_buffer,
```
vz10, line 35: I'm a bit concern about passing the reference to the buffer between containers, maybe it would be safer just to path its name

vz10 (commented):
I think it looks great. 

Just have a couple of NIT-like suggestions/comments

tests/__init__.py
kyleknap: Oh that omission was on purpose. Mainly the `{message}` is filled out later when the `AssertionError` is propagated.

awscli/customizations/wizard/ui/prompt.py
kyleknap: I think I originally wanted to use the name but it was something that got missed after trying a couple of different approaches. Honestly the key tabbing logic is going to be going away in the next PR, specifically I'm going to introduce an abstraction that walks though the wizard prompts telling us what the current prompt is and what the next prompt will be. I do like the idea of making the keybindings global now though so I'll introduce a `keybindings` module to encompass the current logic and build off of it.

awscli/customizations/wizard/ui/prompt.py
kyleknap: Yeah I'm just going to switch this to use the value name.

vz10 (approved):
Perfect. I like it 🏄‍♀️

PR #5735, opened 2020-11-18
Get output format from current_fragment, improve yaml formatter

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/functional/autocomplete/test_completer.py
+++ b/tests/functional/autocomplete/test_completer.py
@@ -28,9 +28,11 @@ class TestShorthandCompleter(unittest.TestCase):
                 '': [('aws', None)],
                 'aws': [('codebuild', None),
                         ('dynamodb', None),
-                        ('s3', None)],
+                        ('s3', None),
+                        ('ec2', None)],
                 'aws.codebuild': [('create-project', None)],
                 'aws.dynamodb': [('put-item', None)],
+                'aws.ec2': [('bundle-instance', None)]
             },
             'arg_names': {
                 '': {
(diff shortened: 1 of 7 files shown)


tests/functional/autocomplete/test_completer.py
```diff
@@ -236,6 +236,11 @@ def test_autoclose_brackets(self):
         self.assertEqual(suggestions[-1].name,
                          'key={M=[key1={M=[key2={BS=[1]}]}]}')
 
+    def test_not_start_autocompletion_wo_trailing_space(self):
+        parsed = self.parser.parse('aws dynamodb put-item --item')
```
kyleknap, line 240: I don't think the `--items` parameter even has auto-completions? Maybe we should use the `codebuild create-project` command from before and just assert it does not have auto-completions without the whitespace.

kyleknap (commented):
Looks good. Just one more comment.

tests/functional/autocomplete/test_completer.py
kyleknap: Or I guess initially it will auto-complete eventually but not on the first input character to the value because this value is a map.

vz10: it has, basically all the other testcases use it

kyleknap: Or to rephrase it, the string `aws dynamodb put-item --item` will not trigger the bug. We should use a string that does trigger the bug to make sure we have a case that catches it in the future.

vz10: Ah, yeah, makes sense

kyleknap (approved):
Looks good. :ship:

PR #5745, opened 2020-11-24
Serverside completions for codebuild commands

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
+++ b/awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
@@ -1,0 +1,113 @@
+{
+  "merge": {
+    "resources": {
+      "Project": {
+        "operation": "ListProjects",
+        "resourceIdentifier": {
+          "name": "projects[]"
+        }
+      }
+    },
+    "operations": {
+      "BatchDeleteBuilds": {
+        "ids": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Build",
+              "resourceIdentifier": "id"
+            }
+          ]
+        }
+      },
+      "BatchGetBuilds": {
+        "ids": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Build",
+              "resourceIdentifier": "id"
+            }
+          ]
+        }
+      },
+      "StartBuild": {
+        "projectName": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Project",
+              "resourceIdentifier": "name"
+            }
+          ]
+        }
+      },
+      "BatchGetProjects": {
+        "names": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Project",
+              "resourceIdentifier": "name"
+            }
+          ]
+        }
+      },
+      "DeleteProject": {
+        "name": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Project",
+              "resourceIdentifier": "name"
+            }
+          ]
+        }
+      },
+      "UpdateProject": {
+        "name": {
+          "completions": [
+            {
+              "parameters": {},
+              "resourceName": "Project",
+              "resourceIdentifier": "name"
+            }
+          ]
+        }
+      },
+      "InvalidateProjectCache": {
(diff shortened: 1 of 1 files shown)


awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
```diff
@@ -0,0 +1,102 @@
+{
+  "merge": {
+    "resources": {
+      "Projects": {
+        "operation": "ListProjects",
+        "resourceIdentifier": {
+          "project": "projects[]"
```
kyleknap, line 7: Let's switch this from `project` to `name` as the values returned from this response are project names.

awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
```diff
@@ -0,0 +1,102 @@
+{
+  "merge": {
+    "resources": {
+      "Projects": {
```
kyleknap, line 4: Let's make this singular (e.g. `Project`). Typically, the resource definition are singular.

awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
```diff
@@ -0,0 +1,102 @@
...
+        "operation": "ListProjects",
+        "resourceIdentifier": {
+          "project": "projects[]"
+        }
+      }
+    },
+    "operations": {
+      "BatchDeleteBuilds": {
```
kyleknap, line 12: If we are going to touch builds, it probably be helpful to add `BatchGetBuilds` and `RetryBuild` as well.

kyleknap (commented):
Looks good. Just had a few suggestions.

awscli/data/codebuild/2016-10-06/completions-1.sdk-extras.json
vz10: `RetryBuild` is in script generated completions

kyleknap: I thought it was only `StopBuild` that in the completion file: https://github.com/aws/aws-cli/blob/2cb3e286ab4b1abbf0496f5c80cc0dc47e3e8dc9/awscli/data/codebuild/2016-10-06/completions-1.json#L12 and not `RetryBuild`.

vz10: In such a case we need to run `gen-server-completions` and push updated completion files, I thought we did in CI

kyleknap: Gotcha. Yeah let's then just run the `gen-server-completions` script to update this completions model for `codebuild` and then fill the missing using the extra's model. In the long run, it would be nice to see the `gen-server-completions` script catch the project related completions.

awscli/data/codebuild/2016-10-06/completions-1.json
```diff
@@ -1,14 +1,114 @@
 {
   "version": "1.0",
   "resources": {
+    "BuildBatche": {
```
kyleknap, line 4: This seems like the script is not able to handle plurals that end in `es`. I wonder if this is something easy to fix?

awscli/data/codebuild/2016-10-06/completions-1.json
```diff
@@ -1,14 +1,114 @@
...
+          }
+        ]
+      }
+    },
+    "DeleteReport": {
+      "arn": {
+        "completions": [
+          {
+            "parameters": {},
+            "resourceName": "SourceCredential",
```
kyleknap, line 42: That's definitely not the correct resource we want to be using. I'd assume that we would want to generate a `Report` resource and use its arn?

kyleknap (commented):
So most of my comments were around the `gen-server-completion` script. I don't think it is doing what we want it to do. I think there are two options:

1. Update the `gen-server-completion` script in this PR to fix all of the issues.
2. Do not regen the completions model and regen the model in another PR once the script is fixed.

kyleknap (approved):
Thanks! :ship: Feel free to merge it once CI passes.

PR #5746, opened 2020-11-24
[v2] Integrate select menu with choices in wizard ui

kyleknap:
This pull request makes it so that any value that requires choices becomes a drop down select menu. I'd recommend reviewing it commit by commit to understand the changes as I tried to logically group each of the different changes I made. To try out the updated, just run the same command from the previous PR:
```
~/GitHub/aws-cli/awscli/customizations/wizard/wizards/iam$ aws cli-dev wizard-dev --run-wizard file://_new-role-wip.yml
```

**Note:** The WIP new role wizard had some of its choices commented out because it requires computing values to obtain the choices. This will be implemented in a subsequent PR.


vz10:
Looks good. But I've got an exception each time I tried to traverse the wizard without choosing anything (just TAB through it). Is it something that will be fixed in future commits?

kyleknap:
I'll fix it in this PR. Seems like something that should work now. I think I did not run into this because I would always toggle the select menu when testing it out.

kyleknap:
@vz10 I fixed the issue you were running into. Should be good to look at again.

awscli/customizations/wizard/ui/keybindings.py
```diff
@@ -22,20 +25,26 @@ def exit_(event):
...
     @kb.add('enter')
     def next_prompt(event):
         submit_current_answer(event)
         next_prompt = event.app.traverser.next_prompt()
-        event.app.layout.focus(next_prompt)
+        layout = event.app.layout
+        next_control = get_ui_control_by_buffer_name(layout, next_prompt)
+        layout.focus(next_control)
 
     @kb.add('s-tab')
```
vz10, line 41: Why don't use constants from `prompt_toolkit.keys.Keys` instead

vz10 (commented):
No exceptions any more 👍

awscli/customizations/wizard/ui/keybindings.py
kyleknap: I did it mainly because that is what the prompt toolkit documentation advertises, but I'm fine changing it to make it consistent with what auto-prompt mode does.

kyleknap:
Ok. I updated it to use the `Keys` constants. It should be good to look at again.

vz10 (approved):
Thanks ⛴

PR #5750, opened 2020-11-24
Add autoprompt mode to user agent

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -171,6 +171,10 @@ class AWSCLIEntryPoint:
         HISTORY_RECORDER.record('CLI_RC', rc, 'CLI')
         return rc
 
+    def _run_driver(self, driver, args, prompt_mode):
+        driver.session.user_agent_extra += " prompt/%s" % prompt_mode
+        return driver.main(args)
+
     def _do_main(self, args):
         driver = self._driver
         if driver is None:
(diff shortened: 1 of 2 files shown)


stealthycoin (approved):

awscli/clidriver.py
```diff
@@ -171,6 +171,10 @@ def main(self, args):
         HISTORY_RECORDER.record('CLI_RC', rc, 'CLI')
         return rc
 
+    def _run_driver(self, driver, args, prompt_mode):
+        driver.session.user_agent_extra += " prompt/%s" % prompt_mode
```
kyleknap, line 175: So I was thinking for the user agent updates we do the following:
1. Not include `prompt/<val>` if the CLI never touches auto-prompt mode.  So that would entail removing the `off` value too.
2. Differentiate between `on` and `on-partial` by using the value `on` for `on` mode (e.g. final value is `prompt/on`) and `partial` for `on-partial` mode (e.g. final value is `prompt/partial`).

Thoughts?

kyleknap (commented):
Looks good. Just had a comment about the values we are setting.

awscli/clidriver.py
vz10: That was my initial idea to put only positive results but  I realised that it'd make query to get amount of requests with disabled autoprompt much more complicated.

kyleknap (approved):
Looks good. :ship:

PR #5764, opened 2020-11-29
[v2] Allow referencing of wizard choices from a variable

kyleknap:
Values in the wizard planning stage can now be retrieved outside of a prompting stage such as making API calls and listing profiles. In addition, this change allows choices to reference both a variable and the value of choices can be a list of strings. To try out the updated, just run the same command from the previous PR's:
```
~/GitHub/aws-cli/awscli/customizations/wizard/wizards/iam$ aws cli-dev wizard-dev --run-wizard file://_new-role-wip.yml
```


vz10 (approved):
Thank you 🛳

PR #5772, opened 2020-12-01
Add details panel to wizard UI

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from collections.abc import MutableMapping
+import json
 
 from prompt_toolkit.application import Application
 
@@ -19,6 +20,7 @@ from awscli.customizations.wizard import core
 from awscli.customizations.wizard.ui.keybindings import get_default_keybindings
 from awscli.customizations.wizard.ui.layout import WizardLayoutFactory
 from awscli.customizations.wizard.ui.style import get_default_style
+from awscli.utils import json_encoder
 
 
 def create_wizard_app(definition, session):
@@ -41,6 +43,7 @@ def create_wizard_app(definition, session):
         }
     )
     app.traverser = WizardTraverser(definition, app.values)
+    app.details_visible = False
     return app
 
 
@@ -52,7 +55,7 @@ class WizardTraverser:
     def __init__(self, definition, values):
         self._definition = definition
         self._values = values
-        self._prompt_definitions = self._collect_all_prompt_definitions()
+        self._prompt_definitions = self._collect_prompt_definitions()
         self._prompt_to_sections = self._map_prompts_to_sections()
         self._current_prompt = self._get_starting_prompt()
         self._previous_prompts = []
@@ -70,6 +73,9 @@ class WizardTraverser:
             return [choice['display'] for choice in choices]
         return None
 
+    def current_prompt_has_details(self):
+        return 'details' in self._prompt_definitions[self._current_prompt]
+    
     def submit_prompt_answer(self, answer):
         if 'choices' in self._prompt_definitions[self._current_prompt]:
             answer = self._convert_display_value_to_actual_value(
@@ -101,7 +107,7 @@ class WizardTraverser:
     def has_visited_section(self, section_name):
         return section_name in self._visited_sections
 
-    def _collect_all_prompt_definitions(self):
+    def _collect_prompt_definitions(self):
         value_prompt_definitions = {}
         for _, section_definition in self._definition['plan'].items():
             for name, value_definition in section_definition['values'].items():
@@ -130,6 +136,18 @@ class WizardTraverser:
                 choices = self._values[value_definition['choices']]
             return self._get_normalized_choice_values(choices)
         return None
+
+    def get_details_for_choice(self, choice):
+        step_definition = self._prompt_definitions[self._current_prompt]
+        if 'details' in step_definition:
+            value = self._convert_display_value_to_actual_value(
+                self._get_choices(self._current_prompt), choice
+            )
+            temp_values = self._values.copy()
+            temp_values[self._current_prompt] = value
+            details = temp_values[step_definition['details']['value']]
+            return json.dumps(details, indent=2,
+                              default=json_encoder)
 
     def _get_normalized_choice_values(self, choices):
         normalized_choices = []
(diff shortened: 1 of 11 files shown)


awscli/customizations/wizard/ui/selectmenu.py
```diff
@@ -242,4 +260,15 @@ def move_up(event):
         def move_down(event):
             self._move_cursor(1)
 
+        @kb.add('f3', filter=Condition(self._prompt_has_details), is_global=True)
```
kyleknap, line 263: So I was trying to keep the select menu classes fairly general without too much of the logic of how wizards work leak into it. To keep it relatively general, what are your thoughts on theses changes?

1. Switch the `details` parameter to be an `on_toggle` callback that can be provided and is called on any up or down key press. This callback would just accept a single value representing what the new selection is. We can then hoist this classes `_show_details()` method to the `WizardPromptSelectionAnswer` class. Another option I'm fine if we need to share this logic in multiple places is to just move the logic in the `_show_details()` method to a function in the`utils.py` module.

2. Add the keybinding for `f3` to the keybindings module. I'm also fine hoisting this to the `WizardPromptSelectionAnswer` if moving it to the keybindings module is not feasible.

awscli/customizations/wizard/ui/section.py
```diff
@@ -74,6 +80,15 @@ def _get_container(self):
...
 
+    def _create_details_panel(self, step_name):
+        buffer_name = f'{step_name}-details'
+        return Window(
+            content=BufferControl(
+                buffer=Buffer(name=buffer_name, read_only=True),
+            ),
+            wrap_lines=True
+        )
+
```
kyleknap, line 91: I don't think this method is used anywhere? It seems like we can delete it.

awscli/customizations/wizard/ui/section.py
```diff
@@ -86,3 +101,55 @@ def _create_prompts_from_section_definition(self):
 
     def __pt_container__(self):
         return self.container
+
+
+class WizardDetailsPanel:
```
kyleknap, line 106: For now, it might make sense to move the detail panels to the `layout` module as it is created once for the layout and is then shared across all of the sections/prompts. I'm also fine if we created a `details.py` module if we have more details related logic we want to consolidate. Mainly it is a little weird that is being grouped with the sections right now.

awscli/customizations/wizard/app.py
```diff
@@ -70,6 +75,9 @@ def get_current_prompt_choices(self):
             return [choice['display'] for choice in choices]
         return None
 
+    def is_current_prompt_has_details(self):
```
kyleknap, line 78: Let's just remove the `is` on this method name. It is a little hard to read when there is both an `is` and `has` in the method.

kyleknap (commented):
It looked good! It was really neat seeing the functionality in action. My main comments were around refactoring, generalizing, and cleaning the implementation up a little bit.

awscli/customizations/wizard/app.py
```diff
@@ -131,6 +142,22 @@ def _get_choices(self, value_name):
...
+        if 'details' in step_definition:
+            value = self._convert_display_value_to_actual_value(
+                self._get_choices(self._current_prompt), choice
+            )
+            parameters = {self._current_prompt: value}
+            for detail in step_definition['details'].split('.'):
+                step_definition = self._apicall_definitions.get(detail)
+                if not step_definition:
+                    return detail
+                parameters[detail] = self._api_call_step.run_step(
```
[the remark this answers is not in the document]
vz10: Nice! I like it

awscli/customizations/wizard/ui/selectmenu.py
```diff
@@ -230,6 +231,11 @@ def _get_items(self):
     def _move_cursor(self, delta):
         super()._move_cursor(delta)
         self.buffer.text = self._get_items()[self._selection]
+        if callable(self.on_toggle):
+            self.on_toggle(self.buffer.text)
+
+    def _prompt_has_details(self):
+        return get_app().traverser.current_prompt_has_details()
```
kyleknap, line 238: This looks like this can be deleted.

awscli/customizations/wizard/app.py
```diff
@@ -49,14 +52,15 @@ class InvalidChoiceException(Exception):
...
+    def __init__(self, definition, values, api_invoker):
         self._definition = definition
         self._values = values
-        self._prompt_definitions = self._collect_all_prompt_definitions()
+        self._prompt_definitions = self._collect_prompt_definitions()
         self._prompt_to_sections = self._map_prompts_to_sections()
         self._current_prompt = self._get_starting_prompt()
         self._previous_prompts = []
         self._visited_sections = [self.get_current_section()]
+        self._api_call_step = core.APICallStep(api_invoker=api_invoker)
```
kyleknap, line 63: I don't think we need to include the `api_call_step` in the initializer any more. We'll probably have to clean this up in the tests as well.

awscli/customizations/wizard/app.py
```diff
@@ -131,6 +138,18 @@ def _get_choices(self, value_name):
...
+    def get_details_for_choice(self, choice):
+        step_definition = self._prompt_definitions[self._current_prompt]
+        if 'details' in step_definition:
+            value = self._convert_display_value_to_actual_value(
+                self._get_choices(self._current_prompt), choice
+            )
+            temp_values = self._values.copy()
+            temp_values[self._current_prompt] = value
+            details = temp_values[step_definition['details']['value']]
+            return json.dumps(details, indent=2,
```
kyleknap, line 150: Were we planning to add the `format_type` to the `details` schema in this PR or in a follow up PR? I think it could be useful now so we are not always json dumping for this method by default.

awscli/customizations/wizard/app.py
```diff
@@ -211,3 +230,14 @@ def _collect_all_value_definitions(self):
             for name, value_definition in section_definition['values'].items():
                 value_definitions[name] = value_definition
         return value_definitions
+
+    def __copy__(self):
```
kyleknap, line 234: Let's add a test for this new copy interface. Should be easy to add.

awscli/customizations/wizard/ui/layout.py
```diff
@@ -64,3 +73,52 @@ def _create_section_tab(self, section_name, section_definition):
...
+        'details_window_height_pref': 30,
+    }
+
+    def __init__(self):
+        self.container = self._get_container()
+
+    def _get_container(self):
+        return ConditionalContainer(
+            HSplit([
+                TitleLine('Details panel'),
```
kyleknap, line 90: Is the plan to put the dynamic title line (based on the details description) in a subsequent PR?

kyleknap (commented):
Looks good. Just some more smaller clean up comments and questions.

awscli/customizations/wizard/ui/layout.py
vz10: Yes

kyleknap (approved):
Thanks! Looks good.

PR #5782, opened 2020-12-04
[v2] Implement run wizard dialog

kyleknap:
The dialog appears at the end when all prompts have been answered. Currently, when you select "Yes", it will just exit the app. In a subsequent commit, it will run the actual wizard execution definition.

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -97,11 +97,9 @@ class WizardTraverser:
             return self._current_prompt
         new_prompt = self._get_next_prompt()
         self._previous_prompts.append(self._current_prompt)
-        if new_prompt != self.DONE:
-            section_of_new_prompt = self._prompt_to_sections.get(new_prompt)
-            if section_of_new_prompt != self.get_current_section():
-                self._visited_sections.append(
-                    self._prompt_to_sections[new_prompt])
+        section_of_new_prompt = self._prompt_to_sections.get(new_prompt)
+        if section_of_new_prompt != self.get_current_section():
+            self._visited_sections.append(self._prompt_to_sections[new_prompt])
         self._current_prompt = new_prompt
         return new_prompt
 
@@ -142,6 +140,7 @@ class WizardTraverser:
         sections = self._definition['plan']
         for section_name, section_definition in sections.items():
             if section_name == self.DONE:
+                prompts_to_sections[section_name] = self.DONE
                 continue
             for name, value_definition in section_definition['values'].items():
                 prompts_to_sections[name] = section_name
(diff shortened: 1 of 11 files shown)


awscli/customizations/wizard/ui/utils.py
```diff
@@ -31,6 +31,14 @@ def get_ui_control_by_buffer_name(layout, buffer_name):
         f"Couldn't find buffer in the current layout: {buffer_name}")
 
 
+def move_to_previous_prompt(traverser, layout):
```
vz10, line 34: Since we have only one `traverser` and only one `layout` probably makes sense to get them from the app inside function instead of passing them in as arguments every time.

awscli/customizations/wizard/app.py
```diff
@@ -52,6 +52,8 @@ class InvalidChoiceException(Exception):
 
 
 class WizardTraverser:
+    DONE = '__DONE__'
```
vz10, line 55: I'm wondering if it'd be better to make `Done` a real step in definition and let customers set their own dialog and button titles, also if we decide to add additional functionality to this dialog (eg. third button) customer will be able to switch it on/off

vz10 (commented):
Sweet. I played with it and it worked really good. Just have a couple suggestions.

awscli/customizations/wizard/app.py
kyleknap: I could see it making sense in the future. I'm just not sure how we would want to approach it in the definition. Mainly it is different than a normal prompt where it saves a value and seems like it would need something more similar to the execute step logic where it defines actions/side effects to do. I think I'd prefer deferring coming up with an interface for this until we have more examples on how we want to customize it. Or were you thinking just adding like an empty `DONE` section in the definition as a sentinel for now? An example of what you had in mind would be helpful.

awscli/customizations/wizard/ui/utils.py
kyleknap: I'm fine with making that change.

awscli/customizations/wizard/app.py
kyleknap: Talked offline. Going to add a reservered `__DONE__` section where all wizards must define:
```
plan:
  first_section:
    shortname: First
    description: First section stuff
    values: ...
  ...  # Do more sections
  __DONE__:
```
Defining this will give you some default dialog but if you wanted to customize it you could. So for example:
```
__DONE__:
  description: Run wizard?  # This is the default dialog title
  options: # This specifies which buttons to include `yes` and `button` are by default
    - yes
    - back
```
Then you could override the built-in options (here it will override the button text):
```
options:
  - name: yes
    description: Yep!
  - name: back
    description: Go back
```

vz10:
Great! ⛴

PR #5786, opened 2020-12-04
Add WizardRunnerV2 to the wizard command

vz10:
*Issue #, if available:*

*Description of changes:*


awscli/customizations/wizard/devcommands.py
```diff
@@ -1,75 +0,0 @@
-# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap, line 1: Instead of deleting this command, could we just adapt the `wizard-dev` command to accept both v1 and v2 wizard files? I probably should have adapted this command in the first place instead of make it support only v2 commands, but this command is useful if you want to load it from outside the built-in wizard directory for the CLI.

tests/unit/customizations/wizard/test_command.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap, line 1: Copyright here should be 2020.

awscli/customizations/wizard/factory.py
```diff
@@ -31,3 +32,18 @@ def create_default_wizard_runner(session):
...
+
+
+class WizardRunnerV2(object):
+    def __init__(self, session):
+        self._session = session
+
+    def run(self, loaded):
+        """Run a single wizard given the contents as a string."""
+        app = create_wizard_app(loaded, self._session)
+        app.run()
```
kyleknap, line 48: So in terms of refactoring the current v2 wizards, I was thinking, we:
1. Move the `create_wizard_app` to this module in order to maintain consistency with how things are declared in the `create_default_wizard_v1_runner`
2. Rename `WizardRunnerV2` to `WizardAppRunner` and move that class to `app.py`. Also update its `__init__()` to just accept a session and wizard app factory (e.g. the new signature would be something like:
```python
class WizardAppRunner:
   def __init__(self, session, app_factory):
        ...
```
3. In the `create_default_wizard_v2_runner`, pass the `create_wizard_app` function as the `app_factory` to the `WizardAppRunner` 
Thoughts? I think the main thing that I'm trying to get us to fix is to get all of the wiring logic that I put in `app.py` module to this factory class and do it in a way where the interface makes sense.

kyleknap (commented):
Looks fine. I just had a few comments.

kyleknap (approved):
Cool. Looks good. :ship:

PR #5792, opened 2020-12-08
[v2] Add events new-rule wizard

kyleknap:
This PR is based off of this PR: https://github.com/aws/aws-cli/pull/5791. So it would make sense to review that one first.

This PR adds a new `events wizard new-rule` command for helping set up an EventBridge rule. I tried to break the commits out so refer to the commit message for more details on each particular change. In general, along with the `new-rule` wizard, I needed to:
* Add support for not equals in condition templating
* Add support for loading and dumping json as a step.
* Fix an issue where key values were getting swallowed masking issues in the output template


awscli/customizations/wizard/core.py
```diff
@@ -204,22 +204,37 @@ class TemplateStep(BaseStep):
...
+                    if left != right:
+                        return group_dict['body']
+                return ''
+        if parameters.get(condition, False) in ['False', 'no', False]:
             return ''
         return group_dict['body']
 
+    def _resolve_value_in_condition(self, value, parameters):
+        if value in parameters:
+            return parameters[value]
```
vz10, line 235: Here can be an ambiguity because we don't distinguish string and key, if I want to compare with string but actually I have the same key in the parameters.

awscli/customizations/wizard/core.py
```diff
@@ -548,3 +571,31 @@ def _deep_merge(self, original, newvalue):
...
+
+    def run_step(self, step_definition, parameters):
+        var_resolver = VariableResolver()
+        value = var_resolver.resolve_variables(
+            parameters, step_definition['value'],
+        )
+        format_type = step_definition['load_type']
+        if format_type == 'json':
+            loaded_value = json.loads(value)
+            parameters[step_definition['output_var']] = loaded_value
```
vz10, line 587: Probably it makes sense to raise an exception if format isn't supported instead of just exit

vz10 (changes requested):
Looks and works good, just a couple suggestions.

vz10:
Perfect 🛳

PR #5801, opened 2020-12-09
[v2] Small wizard refactorings

kyleknap:
Specifically includes:
* Create explicit class for the `Application` and `Layout` classes. This better defines the public interfaces for those objects.
* Override the built-in exception handler to just exit the application with the captured error on any unhandled errors.

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -13,7 +13,11 @@
 import json
 from collections.abc import MutableMapping
 
+from prompt_toolkit.application import Application
+
 from awscli.customizations.wizard import core
+from awscli.customizations.wizard.ui.style import get_default_style
+from awscli.customizations.wizard.ui.keybindings import get_default_keybindings
 from awscli.utils import json_encoder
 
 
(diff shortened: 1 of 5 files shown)


awscli/customizations/wizard/app.py
```diff
@@ -39,6 +63,40 @@ def run(self, loaded):
...
+        self.error_bar_visible = None
+        super().__init__(
+            layout=layout, style=style, key_bindings=key_bindings,
+            full_screen=full_screen
+        )
+
+    def run(self, pre_run=None, **kwargs):
+        loop = get_event_loop()
+        previous_exc_handler = loop.get_exception_handler()
+        loop.set_exception_handler(self._handle_exception)
```
vz10, line 86: Probably makes sense to add test to check that we exit on exception.

vz10 (changes requested):
I really like this refactoring, just one small comment

kyleknap:
Ok finally got a test case working for the exception handler. Should be good to look at now.

vz10 (approved):
Thank you. 🚀

PR #5807, opened 2020-12-10
[v2] Some fixes after testing various wizards

kyleknap:
Specifically:
* Allow variables to not exist in determining template conditions. This is needed because templates are used to form values that are in turn used as conditions, which do not guarantee a value being set.
* Update CloudFormation and CLI syntax in samples
* Encapsulate cron expression with cron()


vz10 (approved):
👍 🚢

PR #5812, opened 2020-12-11
Add data types casting and default values to prompts

vz10:
Add data types casting and default values to prompts
Update new-table wizard to support new wizard UI

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -12,8 +12,6 @@
 # language governing permissions and limitations under the License.
 import json
 from collections.abc import MutableMapping
-from traceback import format_tb
-
 
 from prompt_toolkit.application import Application
 from prompt_toolkit.eventloop import get_event_loop, run_until_complete
@@ -21,32 +19,10 @@ from prompt_toolkit.eventloop import get_event_loop, run_until_complete
 from awscli.customizations.wizard import core
 from awscli.customizations.wizard.ui.style import get_default_style
 from awscli.customizations.wizard.ui.keybindings import get_default_keybindings
+from awscli.customizations.wizard.exceptions import (
+    UnexpectedWizardException, UnableToRunWizardError, InvalidChoiceException
+)
 from awscli.utils import json_encoder
-
-
-class UnexpectedWizardException(Exception):
-    MSG_FORMAT = (
-        'Encountered unexpected exception inside of wizard:\n\n'
-        'Traceback:\n{original_tb}'
-        '{original_exception_cls}: {original_exception}'
-    )
-
-    def __init__(self, original_exception):
-        self.original_exception = original_exception
-        message = self.MSG_FORMAT.format(
-            original_tb=''.join(format_tb(original_exception.__traceback__)),
-            original_exception_cls=self.original_exception.__class__.__name__,
-            original_exception=self.original_exception
-        )
-        super().__init__(message)
-
-
-class InvalidChoiceException(Exception):
-    pass
-
-
-class UnableToRunWizardError(Exception):
-    pass
 
 
 class WizardAppRunner(object):
@@ -130,11 +106,16 @@ class WizardTraverser:
             self._current_prompt, {})
     
     def submit_prompt_answer(self, answer):
-        if 'choices' in self._prompt_definitions[self._current_prompt]:
+        definition = self._prompt_definitions[self._current_prompt]
+        if 'choices' in definition:
             answer = self._convert_display_value_to_actual_value(
                 self._get_choices(self._current_prompt),
                 answer
             )
+        if 'datatype' in definition:
+            answer = core.DataTypeConverter.convert(
+                definition['datatype'], answer)
+
         self._values[self._current_prompt] = answer
 
     def next_prompt(self):
--- a/awscli/customizations/wizard/ui/keybindings.py
+++ b/awscli/customizations/wizard/ui/keybindings.py
@@ -19,6 +19,7 @@ from awscli.customizations.wizard.ui.utils import (
     get_ui_control_by_buffer_name, move_to_previous_prompt,
     show_details_if_visible_by_default, refresh_details_view
 )
+from awscli.customizations.wizard.exceptions import BaseWizardException
 
 
 @Condition
@@ -47,12 +48,21 @@ def get_default_keybindings():
         current_prompt = event.app.traverser.get_current_prompt()
         prompt_buffer = get_ui_control_by_buffer_name(
             event.app.layout, current_prompt).buffer
-        event.app.traverser.submit_prompt_answer(prompt_buffer.text)
+        try:
+            event.app.traverser.submit_prompt_answer(prompt_buffer.text)
+            if isinstance(event.app.layout.error_bar.current_error,
+                          BaseWizardException):
+                event.app.layout.error_bar.clear()
+        except BaseWizardException as e:
+            event.app.layout.error_bar.display_error(e)
+            return False
+        return True
 
     @kb.add(Keys.Tab)
     @kb.add(Keys.Enter)
     def next_prompt(event):
-        submit_current_answer(event)
+        if not submit_current_answer(event):
+            return
         traverser = event.app.traverser
         layout = event.app.layout
         event.app.details_visible = False
@@ -67,7 +77,8 @@ def get_default_keybindings():
 
     @kb.add(Keys.BackTab)
     def previous_prompt(event):
-        submit_current_answer(event)
+        if not submit_current_answer(event):
+            return
         event.app.details_visible = False
         move_to_previous_prompt(event.app)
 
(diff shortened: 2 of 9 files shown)


awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -95,6 +100,101 @@ plan:
...
+          variable: encryption_settings
+          equals: customer_managed_cmk
+  preview:
+    shortname: Preview
+    description: Preview results
+    values:
+      preview_cli_command_value:
+        type: template
+        value: |
+          aws dynamodb create-table \
```
kyleknap, line 113: Running through the template, I ran into a condition when a backslash is ended to the last line:
```
aws dynamodb create-table \ 
  --table-name 'MyTable' \ 
  --attribute-definitions \ 
    'AttributeName=ds,AttributeType=S' \ 
  --key-schema \ 
    'AttributeName=ds,KeyType=HASH' \ 
```
It seems like it happens when you do not specify a KMS key.

kyleknap (commented):
Looks fine. I just had a few comments. I think my biggest one was around handling of problematic input. I still want to test it out a bit more by deploying some of the templates and use the CLI commands. So this is just some of the initial stuff that I saw

awscli/customizations/wizard/app.py
```diff
@@ -130,11 +137,23 @@ def current_prompt_has_details(self):
...
+        definition = self._prompt_definitions[self._current_prompt]
+        if 'choices' in definition:
             answer = self._convert_display_value_to_actual_value(
                 self._get_choices(self._current_prompt),
                 answer
             )
+        if 'datatype' in definition:
+            try:
+                answer = self.CONVERSION_FUNCS[definition['datatype']](answer)
+            except ValueError:
```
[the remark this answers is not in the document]
vz10: Yeah, such flow sounds reasonable

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -95,6 +100,106 @@ plan:
...
+          equals: customer_managed_cmk
+  preview:
+    shortname: Preview
+    description: Preview results
+    values:
+      line_brake_if_KMS_set:
+        type: template
+        value: |
+          {% if {encryption_settings} != default %}
+          \ {% endif %}
```
kyleknap, line 114: Instead of adding this variable, I think we should make the last line a parameter that we know will have to be included and therefore should not include the backslash. For example, we can just make the last command related to determining the capacity provision:
```
             {% if {capacity_mode} == PROVISIONED %}
            --provisioned-throughput \
              'ReadCapacityUnits={read_capacity},WriteCapacityUnits={write_capacity}'
            {% endif %}
            {% if {capacity_mode} == PAY_PER_REQUEST %}
            --billing-mode PAY_PER_REQUEST
            {% endif %}
```
In general, the ordering of parameters is not too important and I think this makes the templating a logic a bit simpler.

awscli/customizations/wizard/ui/keybindings.py
```diff
@@ -47,12 +47,22 @@ def submit_current_answer(event):
         current_prompt = event.app.traverser.get_current_prompt()
         prompt_buffer = get_ui_control_by_buffer_name(
             event.app.layout, current_prompt).buffer
-        event.app.traverser.submit_prompt_answer(prompt_buffer.text)
+        try:
+            event.app.traverser.submit_prompt_answer(prompt_buffer.text)
+            event.app.traverser.exception_handler.clear_if_exception(ValueError)
```
kyleknap, line 52: Then for this logic, I'm thinking we just expose a `current_error` property and compare the type against that. It seems like that would be a little more general purpose and useful instead of a new clear method. So this logic wold now become something like:
```python
event.app.traverser.submit_prompt_answer(prompt_buffer.text)
if isinstance(event.app.layout.error_bar.current_error, BaseWizardException):
     event.app.layout.error_bar.clear()
```
Then if we find the `BaseWizardException` becomes too general, we can then create a new subclass `BaseWizardValueException` and update other classes appropriately so it reflects any exception that may be raised because of bad prompt input inside of the wizard.

kyleknap (commented):
It's looking fine. I like the functional update regards to not moving onto the next prompt on bad input. My main comments were related to refactoring and some more edge cases that I ran into playing with the wizard. I still need to take a deeper look at the CloudFormation template and CLI commands being generated for the different branches, but I'll follow up with any more feedback if there other cases we missed.

awscli/customizations/wizard/exceptinos.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap, line 1: Typo in the filename: `exceptinos.py` -> `exceptions.py`

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -95,6 +100,107 @@ plan:
...
+                  - AttributeName: {sort_key_name}
+                    KeyType: RANGE
+                  {% endif %}
+                {% if {capacity_mode} == PROVISIONED %}
+                ProvisionedThroughput:
+                  ReadCapacityUnits: {read_capacity}
+                  WriteCapacityUnits: {write_capacity}
+                {% endif %}
+                {% if {capacity_mode} == PAY_PER_REQUEST %}
+                  BillingMode: PAY_PER_REQUEST
```
kyleknap, line 167: It seems the `BillingMode` has one level of indentation too much. I was getting the following CloudFormation template:
```
Resources:
  MyTable:
    Type: "AWS::DynamoDB::Table"
    Properties:
      TableName: MyTable
      AttributeDefinitions:
        - AttributeName: Foo
          AttributeType: S
      KeySchema:
        - AttributeName: Foo
          KeyType: HASH
        BillingMode: PAY_PER_REQUEST
```
which was not valid yaml. It instead needs to be:
```
Resources:
  MyTable:
    Type: "AWS::DynamoDB::Table"
    Properties:
      TableName: MyTable
      AttributeDefinitions:
        - AttributeName: Foo
          AttributeType: S
      KeySchema:
        - AttributeName: Foo
          KeyType: HASH
      BillingMode: PAY_PER_REQUEST
```

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -152,7 +258,24 @@ execute:
...
+__OUTPUT__:
+  value: |
+    Wizard successfully created DynamoDB Table:
+      Table name: {table_name}
+      Primary partition key: {primary_key_name} ({primary_key_type})
+      {% if wants_sort_key == yes %}
+      Primary sort key: {sort_key_name} ({sort_key_type})
+      {% endif %}
+    {% if {preview_type} == preview_cli_command %}
+    Steps to create role is equivalent to running the following sample AWS CLI commands:
```
kyleknap, line 273: Should change this from `role` to `table` here.

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -152,7 +258,24 @@ execute:
...
+
+    {preview_cli_command_value}
+    {% endif %}
+    {% if {preview_type} == preview_cfn_template %}
+    Steps to create role is equivalent to deploying the following sample AWS CloudFormation template:
```
kyleknap, line 278: Same thing. We should change this from `role` to `table` here.

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
```diff
@@ -152,7 +258,24 @@ execute:
...
       operation: dynamodb.CreateTable
       params: "{create_table_params}"
+__OUTPUT__:
+  value: |
+    Wizard successfully created DynamoDB Table:
+      Table name: {table_name}
+      Primary partition key: {primary_key_name} ({primary_key_type})
+      {% if wants_sort_key == yes %}
+      Primary sort key: {sort_key_name} ({sort_key_type})
+      {% endif %}
```
kyleknap, line 271: It would be nice to add a newline after this in order to break up the summary section from the sample CLI commands and CloudFormation template sections.

kyleknap (commented):
Looks good. I just had some smaller feedback after testing it out more. Once that gets addressed, I think we should be good to merge it.

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
vz10: 🤦

awscli/customizations/wizard/wizards/dynamodb/new-table.yml
vz10: 🤦

kyleknap (approved):
Nice! Looks good to me :ship:

PR #5818, opened 2020-12-16
Migrate new-function wizard to new UI

vz10:
*Issue #, if available:*

*Description of changes:*
- Migrate new-function wizard to new UI
- Update prompt section to support completers

### DIFF
--- a/awscli/customizations/wizard/ui/prompt.py
+++ b/awscli/customizations/wizard/ui/prompt.py
@@ -14,12 +14,16 @@ from prompt_toolkit.application import get_app
 from prompt_toolkit.buffer import Buffer
 from prompt_toolkit.document import Document
 from prompt_toolkit.filters import Condition
+from prompt_toolkit.key_binding import KeyBindings
+from prompt_toolkit.keys import Keys
 from prompt_toolkit.layout.containers import (
-    Window, VSplit, Dimension, ConditionalContainer
+    Window, VSplit, Dimension, ConditionalContainer, FloatContainer, Float
 )
 from prompt_toolkit.layout.controls import BufferControl
 from prompt_toolkit.layout.margins import ScrollbarMargin
 from prompt_toolkit.layout.containers import ScrollOffsets
+from prompt_toolkit.layout.menus import CompletionsMenu
+from prompt_toolkit.completion import PathCompleter
 
 from awscli.customizations.wizard.ui.selectmenu import (
     CollapsableSelectionMenuControl
@@ -35,11 +39,19 @@ class WizardPrompt:
 
     def _get_container(self):
         uses_choices = 'choices' in self._value_definition
+        uses_completer = 'completer' in self._value_definition
         if uses_choices:
             answer = WizardPromptSelectionAnswer(self._value_name)
+        elif uses_completer:
+            answer = WizardPromptCompletionAnswer(
+                self._value_name,
+                default_value=self._value_definition.get('default_value'),
+                completer=self._value_definition['completer']
+            )
         else:
             answer = WizardPromptAnswer(
-                self._value_name, self._value_definition.get('default_value')
+                self._value_name,
+                default_value=self._value_definition.get('default_value'),
             )
         return ConditionalContainer(
             VSplit(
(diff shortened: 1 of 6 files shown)


awscli/customizations/wizard/wizards/lambda/new-function.yml
```diff
@@ -1,12 +1,14 @@
-version: "0.1"
+version: "0.2"
```
kyleknap, line 1: We should add a changelog entry for this similar to what we did for the new table wizard for DynamoDB.

awscli/customizations/wizard/ui/prompt.py
```diff
@@ -124,6 +136,60 @@ def __pt_container__(self):
...
+                wrap_lines=True
+            ),
+            floats=[
+                Float(
+                    xcursor=True, ycursor=True, top=1,
+                    height=self._get_menu_height,
+                    content=CompletionsMenu(),
+                )
+            ],
+            key_bindings=self._get_key_bindings()
```
kyleknap, line 177: One thing that tripped me up using it was that if you hit `<TAB>` while completing it will not toggle to the suggested completion and instead move onto the next prompt. I'm wondering if we should override the tab keybindings for anything that has a completer? I'm open to ideas on this.

awscli/customizations/wizard/wizards/lambda/new-function.yml
```diff
@@ -42,16 +48,55 @@ plan:
...
+          {preview_cli_command_value}
+          {% endif %}
+      preview_type:
+        type: prompt
+        description: Select an preview format
+        choices:
+          - display: None
+            actual_value: preview_none
+          - display: AWS CLI command
+            actual_value: preview_cli_command
```
kyleknap, line 88: I'm wondering, if we should also generate CloudFormation snippets for this. My guess is that we did not do it because the zip file with the code would need to be uploaded to s3. But I'm wondering if we just generate one anyway and just add placeholders for the source code part informing the user that they need to update the values to whatever S3 bucket they want to use? The nice part is that because we are using yaml, we can embed comments to elaborate more on the placeholders if we want.

kyleknap (commented):
It's looking good! I'm still trying to take a deeper look at it but figured to pass some initial suggestions to start a discussion around.

tests/unit/customizations/wizard/test_app.py
```diff
@@ -1523,3 +1538,105 @@ def test_exception_handler(self):
...
+
+        yield from [
+            Completion('file1'[prefix_len:], 0, display='file1'),
+            Completion('file2'[prefix_len:], 0, display='file2'),
+        ]
+
+
+class TestPromptCompletionWizardApplication(BaseWizardApplicationTest):
+    def setUp(self):
+        self.patch = mock.patch(
```
kyleknap, line 1555: Instead of patching, what is your thoughts on just using a temporary/sandboxed directory to work with? Specifically, we have the `FileCreator` abstraction that makes it easy to create temporary files: https://github.com/aws/aws-cli/blob/4c71115ef0538004ab0df93f55b92b5399667a94/awscli/testutils.py#L501 and we often use it in dealing with the filesystem layer.

kyleknap (commented):
Looks good! It worked well when I was playing around with it. I just had some more comments. Otherwise, we should be good to merge it after.

tests/unit/customizations/wizard/test_app.py
vz10: I had an issue with `FileCreator` clean up on windows containers so I'd prefer not to use it if it's not necessary and in this case by adding real files we will test not our code but `PromptToolkit` `PathCompleter`

kyleknap: Yeah I think ideally the completer would be injectable, but that would require quite a bit of refactoring. I think the main reason that I brought up the suggestion was that I really do not like when we use patching because it has caused issues for us in the past related to patches not being properly cleaned up and also covering up bugs.

It's a little concerning that the `FileCreator` abstraction is not working properly, but I'm assuming this will get addressed in the other GitHub action PR you have. Otherwise, I'm fine keeping the test as is.

kyleknap (approved):
Looks good. :ship:

PR #5826, opened 2020-12-18
Adding custom sso timestamp parser

nateprewitt:
Adding custom json loader to the SSO login workflow to create a correctly formatted timestamp in the token cache. This is dependent upon boto/botocore#2254.


joguSD (approved):
Looks good assuming the tests past after the botocore changes are merged.

PR #5842, opened 2020-12-24
Fix windows-latest GitHub Actions

vz10:
*Issue #, if available:*

*Description of changes:*
- Split `run-tests` and `run-checks` in two different steps because before tests failed but checks were successful and actions assumed it as a positive result
- Add `--traverse-namespace` to nosetests runner, otherwise on Python 3.8 it didn't run tests except ones in upper level folder
- Add `FakeOutput` to prompt toolkit Application in tests, otherwise it failed on Application creation in tests
- Add `encoding` argument to `create_file` method of `FileCreator` to set it explicitly to `UTF-8` in cli_encoding tests
- ~Swallow `PermissionError` on tempfile clean up in PromptToolkit tests~
- Use abs path instead of relative in `CommandArchitectureTest` in case there is no relative path (tests and file are on different volumes)

### DIFF
--- a/awscli/autocomplete/db.py
+++ b/awscli/autocomplete/db.py
@@ -17,6 +17,7 @@ BUILTIN_INDEX_FILE = os.path.join(
     'data', 'ac.index'
 )
 
+
 # This is similar to DBConnection in awscli.customization.history.
 # I'd like to reuse code, but we also have the contraint that we don't
 # want to import anything outside of awscli.autocomplete to ensure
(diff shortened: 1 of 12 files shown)


awscli/autocomplete/db.py
```diff
@@ -33,9 +34,16 @@ def __init__(self, db_filename=None):
...
         if self._db_conn is None:
-            self._db_conn = sqlite3.connect(
-                self._db_filename, check_same_thread=False,
-                isolation_level=None)
+            kwargs = {
+                'check_same_thread': False,
+                'isolation_level': None
+            }
+            if self._db_filename.startswith('file::memory:'):
+                # This statement was added because old versions of sqlite don't support 'uri' but we use it for tests
```
kyleknap, line 42: The comments look longer than 79 characters. Can we break these lines up a little bit to make them shorter?

kyleknap (approved):
Looks good! Just a really small comment. Otherwise :ship:

kyleknap (approved):
Thanks!

Issue #2402, opened 2017-01-21
aws_completer doesn't complete filenames


Malvineous:
If you use a command like this:

    aws s3 cp a<tab>

Then normally the `<tab>` will result in autocompleting a file in the current directory starting with `a`.  However this does not happen, and pressing tab has no effect.

Can the autocompleter be updated to allow completing of filenames at the appropriate times?

kyleknap:
Marking as a feature request. I think the difficult part about this is determining if a user wants to autocomplete a parameter/command or autocomplete a local file.

ahuwaa:
@Malvineous what OS and terminal are you using? I'm able to autocomplete local filenames using <tab> on Mac OSX in iTerm2 and regular Terminal

Malvineous:
@ahuwaa I'm using bash under Linux with [aws-cli/bin/aws_bash_completer](https://github.com/aws/aws-cli/blob/master/bin/aws_bash_completer).  Have you enabled an autocomplete script for your shell like this one?  You might be using your shell's default autocomplete process which will only complete filenames but not aws-cli command parameters.

ahuwaa:
Sorry, I misunderstood. I did not know there was an autocomplete feature but have now reproduced the issue.

I'd like to take a shot at this, if someone hasn't already started. Any advice/tips appreciated.

ahuwaa:
Going to need help with this due to time constraints. So far I thought it was simplest to just default to the shell's autocomplete e.g. this works in bash

```python
import subprocess
import pipes
# ...
def complete(cmdline, point):
    choices = Completer().complete(cmdline, point)
    if len(choices) > 0:
        print(' \n'.join(choices))
    else:
        cmd = cmdline.split(" ").pop()
        try:
            out = subprocess.check_output("compgen -o default " + pipes.quote(cmd), shell=True)
        except subprocess.CalledProcessError:
            out = ""
        print out
```
but of course this relies on something like compgen being available. If anyone has a better suggestion or would like to take over, feel free

nhed:
@kyleknap remote filename completion is the primary thing I hoped for when I learned that there is a complete module, calling it a feature-request is somewhat of an understatement.  Even `scp` in bash completion goes to the remote server over ssh to obtain list of files/paths.  Having the path start with `s3://` should give you the hint (as its the hint that `aws cp` uses to figure out which side of the copy is remote which is local)

CBribiescas:
Is there any ETA on this feature?

[...]

PR #5849, opened 2020-12-29
completer improvements

erankor:
*Issue #, if available:*
#2402

*Description of changes:*
add support for s3 and local path completion in s3 commands.
the s3 path completion includes both bucket names and prefixes.
in case --profile/--region are passed on the command line, they are used
when performing the s3 queries.
the creation of the 'help command' was moved away from the constructor
since it takes a considerable amount of time, and is not required for path
completion.
also, support completion of --profile p<TAB> (partial profile name)


erankor:
Ping

erankor:
Ping

erankor:
@garnaat / @JordonPhillips I see you are the authors of this file, can one of you help?

catskul:
@kdaily any chance you have time to review this? If no do you know another who can review this?

scottedwards2000:
> I reviewed this PR with the team. Unfortunately, it can't be merged, mainly for the following: it is opened against the develop branch and hence for the AWS CLI v1 only. For us to accept this feature, it would need to be done against the v2 branch and would be a v2-only feature.
> 
> The implementation for the v2 branch will be more straightforward and less invasive as a customization instead of being integrated directly into the completer class. You can see examples of this kind of thing as a customization here that other services are using:
> 
> https://github.com/aws/aws-cli/tree/v2/awscli/autocomplete/serverside/custom_completers
> 
> I understand that this doesn't help the work that has already been done and the delay to get this reviewed, and I apologize for that. We are working on reviewing backlog items and improving the user contribution process, including documentation of the process and contributor expectations. As part of that, we need to get more information on a GitHub issue for a feature on what the scope of change, use cases, proposed implementation, and alternatives considered are, and ideally this should occur before an implementation is submitted for review. One of the big things with this specific feature request is the addition of API calls that would be necessary to perform the server-side completion, and users of this feature would need to be made aware of how this would work.

Are these custom completers something we could add easily to our own CLI, or do they need to be integrated by you guys after a pull request into the whole codebase?  thx

PR #5850, opened 2020-12-29
Migrate configure wizard to new UI

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.changes/next-release/enhancement-Wizards-81054.json
+++ b/.changes/next-release/enhancement-Wizards-81054.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Wizards",
+  "description": "Migrate configure wizard to new UI"
+}
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -202,6 +202,11 @@ class WizardTraverser:
             return step_definition['details'].get('description')
         return ''
 
+    def get_details_for_prompt(self):
+        step_definition = self._prompt_definitions[self._current_prompt]
+        if 'details' in step_definition:
+            return self._values[step_definition['details']['value']]
+
     def get_details_for_choice(self, choice):
         if not choice:
             return ''
--- a/awscli/customizations/wizard/core.py
+++ b/awscli/customizations/wizard/core.py
@@ -14,7 +14,7 @@
 import re
 import json
 import os
-from functools import partial
+from functools import partial, reduce
 
 from botocore import xform_name
 import jmespath
@@ -445,6 +445,9 @@ class ConditionEvaluator:
         if 'equals' in single:
             expected = single['equals']
             return parameters.get(varname) == expected
+        if 'exists' in single:
+            expected = single['exists']
+            return bool(parameters.get(varname)) == expected
         return False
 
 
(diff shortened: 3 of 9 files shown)


awscli/customizations/wizard/wizards/configure/main.yml
```diff
@@ -0,0 +1,361 @@
...
+          variable: add_config
+          equals: yes
+        details:
+          value: region_details
+          visible: True
+          description: "Docs"
+      prompt_mode_details:
+        type: template
+        value: |
+          Enables the auto-prompt for the AWS CLI. There are two settings that can be used:
```
kyleknap, line 172: Similar to what we did with the `iam wizard new-role`, it may be interesting if we labeled the details with `output` and make it `rst` (or any other format) so that:

1. We can right these doc strings in rst or even load it from some other rst file that we have (we may want to consider doing that now depending how much documentation we want embedded into the wizard. The wizard definition is getting quite lengthy already...)
2. Render the values similar to what we do for the man pages that we do in auto-prompt mode

I think that will make the docs panel look cleaner in general.

awscli/customizations/wizard/wizards/configure/main.yml
```diff
@@ -0,0 +1,361 @@
...
+        type: template
+        value: |
+          Specifies the AWS Region to send requests to for commands requested using this profile.
+
+            You can specify any of the Region codes available for the chosen service as listed in AWS Regions and Endpoints in the Amazon Web Services General Reference.
+
+            "aws_global" enables you to specify the global endpoint for services that support a global endpoint in addition to regional endpoints, such as AWS Security Token Service (AWS STS) and Amazon Simple Storage Service (Amazon S3).
+      region:
+        type: prompt
+        description: "Enter the region name"
```
kyleknap, line 161: It would be interesting if we used a `completer` here so that we can get region completions similar to auto-prompt mode. I think that would definitely be helpful.

awscli/customizations/wizard/wizards/configure/main.yml
```diff
@@ -0,0 +1,361 @@
+version: "0.2"
```
kyleknap, line 1: We should retain the file name `_main.yml`. Mainly, `_main.yml` is a reserved value for wizard command generator so that it allows you to jump into the wizard at the `wizard` command (e.g. `aws configure wizard`). Right now to enter the wizard, we need to specify `aws configure wizard main` which is not what we want.

awscli/customizations/wizard/wizards/configure/main.yml
```diff
@@ -0,0 +1,361 @@
...
+      config_type:
+        type: prompt
+        description: "Choose credentials type"
+        choices:
+          - display: Static Credentials
+            actual_value: static_creds
+          - display: Assume Role
+            actual_value: assume_role
+          - display: Process Provider
+            actual_value: proc_provider
```
kyleknap, line 59: It would be really neat if there was a doc panel explaining what each of these credential providers do similar to the additional configuration prompts.

kyleknap (commented):
I like the flow of the wizard. I'm especially excited about the doc panel to help with understanding configurations. I mainly left higher-level suggestions on the functionality. Really did not dive too much into the implementation quite yet. Just wanted to get discussions around these suggestions started.

awscli/customizations/wizard/wizards/configure/main.yml
vz10: sounds reasonable but I'd prefer to use HTML or Markdown instead and get rid of `rst`

awscli/customizations/wizard/wizards/configure/main.yml
vz10: I thought about it but it's not clear what profile to use for the completer client

tim-finnigan:
Closing as not planned at this time, but the team may revisit the proposed changes at some point in the future. There hasn't been any activity here in several years but there still an internal backlog item tracking this, so we may reopen this PR in the future to pick up where it left off.

PR #5884, opened 2021-01-20
[v2] Introduce pytest into s3 integration test suite 

kyleknap:
This introduces pytest to the v2 integration test suite. To start, I ported over a test class from the s3 integration tests to start a precedence on how pytest style test look in the codebase and updated the `run-integ-tests` script to run tests using pytest. After this PR, I plan to:

* Port the rest of the s3 `test_plugin.py` tests over to pytest style of testing
* Look into porting the other two integration test that are using nose abstractions and thus can't be run on pytest. This seems helpful as it will simplify the `run-integ-tests` script as we will only need one test runner.

I just wanted to get initial feedback on this work before I make too much progress on it. The end goal is to build up the `pytest-integ-v2` branch and then send out a PR that merges that into the `v2` branch once the remaining tasks are done.

### DIFF
--- a/scripts/ci/run-integ-tests
+++ b/scripts/ci/run-integ-tests
@@ -4,17 +4,66 @@
 # binary package not from the CWD.
 
 import os
-from subprocess import check_call
+from subprocess import run, CalledProcessError
 
 _dname = os.path.dirname
 
 REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
 os.chdir(os.path.join(REPO_ROOT, 'tests'))
+NOSE_ONLY_TESTS = [
+    'test_smoke.py',
+    os.path.join('customizations', 'test_generatecliskeleton.py'),
+]
 
 
-def run(command):
-    return check_call(command, shell=True)
+class TestFailureException(Exception):
+    pass
 
 
-run('nosetests --with-xunit --cover-erase --with-coverage '
-    '--cover-package awscli --cover-xml -v integration')
+def get_nose_only_integ_tests():
+    return [
+        os.path.join('integration', integ_test)
+        for integ_test in NOSE_ONLY_TESTS
+    ]
+
+
+def run_pytest():
+    tests_to_ignore = get_nose_only_integ_tests()
+    ignore_params = [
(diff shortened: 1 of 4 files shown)


scripts/ci/run-integ-tests
```diff
@@ -4,17 +4,66 @@
...
+    cmd = ['py.test', '-v', 'integration/'] + ignore_params
+    print(f'Running cmd: {cmd}')
+    return run(cmd)
+
+
+def run_nose():
+    test_to_run = get_nose_only_integ_tests()
+    cmd = ['nosetests', '-v'] + test_to_run
+    print(f'Running cmd: {cmd}')
+    return run(cmd)
```
vz10, line 44: Is there any particular reason why we use subprocess instead of running tests using subprocess instead of invoke test runners directly via `pytest.main()` and `nose.run()`

vz10 (commented):
Looks good. Just a small comment/suggestion regarding the test run.

scripts/ci/run-integ-tests
kyleknap: Honestly, I did not really consider it. One thing that I would be concerned about in running `main()`/`run()` directly is that any modifications from a test (e.g. a bad patch) may be carried over to next test run because it is all in the same Python process. So subprocess provides a bit more sandboxing for us. Is there a particular reason, you would want to see the test runners being invoked directly?

vz10: The main reason is that it can give us more flexibly in test running and test error handling  during the nose -> pytest migration period.

kyleknap: Yeah I don't really have too strong of a preference here. I also do not want to spend too much time focusing on specifics of this script in its current state since after this PR, I'm going to be sending another PR that ports the entire suite to pytest and simplify this script to this: https://github.com/kyleknap/aws-cli/commit/cd9bda021cec3bc49608cdbef9791436e33ba193. In the commit, I'm using `pytest.main()` and seems to be working fine. What's your thoughts on deferring making this change to the PR that includes that commit?

vz10: Sounds good for me 👍

vz10 (approved):
🛳

PR #5891, opened 2021-01-21
[v2] Port more remaining non-s3 integ test that need nose to pytest

kyleknap:
Specifically, I ported both the `test_smoke.py` and `customizatons/test_generateskeleton.py` to pytest. And because we no longer use nose for that test suite, I removed the logic for running tests both under nose and pytest in the `run-integ-tests` script.

### DIFF
--- a/tests/integration/customizations/test_generatecliskeleton.py
+++ b/tests/integration/customizations/test_generatecliskeleton.py
@@ -15,14 +15,14 @@ import json
 import logging
 
 import mock
-from nose.tools import assert_equal
+import pytest
 import ruamel.yaml as yaml
 
-from awscli.testutils import capture_output
 from awscli.clidriver import create_clidriver
 
 
-def test_can_generate_skeletons_for_all_service_comands():
+def get_all_cli_skeleton_commands():
+    skeleton_commands = []
     environ = {
         'AWS_DATA_PATH': os.environ['AWS_DATA_PATH'],
         'AWS_DEFAULT_REGION': 'us-east-1',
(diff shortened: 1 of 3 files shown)


tests/integration/customizations/test_generatecliskeleton.py
```diff
@@ -15,14 +15,14 @@
...
 import ruamel.yaml as yaml
 
-from awscli.testutils import capture_output
 from awscli.clidriver import create_clidriver
 
 
-def test_can_generate_skeletons_for_all_service_comands():
+def get_all_cli_skeleton_commands():
+    skeleton_commands = []
     environ = {
```
vz10, line 26: What do you think about moving environment patching to a separate fixture.

vz10 (commented):
I like how it looks now, much cleaner and concise! Just a small comment/suggestion.

tests/integration/customizations/test_generatecliskeleton.py
kyleknap: So I looked into making this a separate fixture, but one thing I could not determine is if you can apply a fixture to something that is not declared as a fixture or being ran as part of the test suite. Do you know if it is possible to call the fixture inside of this function (which is neither a fixture or test case itself)?

Looking over the diff again, I many need to create a fixture anyway because I'm no longer patching the environment when I'm running the generate skeleton command, but unfortunately seems that I will have to duplicate the logic in two places and cannot just reuse it as a fixture.

vz10 (approved):
🚢

PR #5905, opened 2021-01-29
[customizations/eks]: Make kubeconfig not world readable

micahhausler:
*Issue #, if available:*
N/A

*Description of changes:*

EKS never writes credentials to $KUBECONFIG files the aws eks update-kubeconfig command creates, but its possible that a customer could use the same $KUBECONFIG with another provider that does later write credentials into the file. This is a defense-in-depth change that is consistent with the file mode Kubernetes uses https://github.com/kubernetes/kubernetes/blob/v1.20.2/staging/src/k8s.io/client-go/tools/clientcmd/loader.go#L437


vz10 (approved):
Looks good to me 🚢

PR #5928, opened 2021-02-08
Add permissions to compat_open function

vz10:
*Issue #, if available:*
Add optional `access_permissions` argument to `compat_open` function to let create files with more strict permissions.

*Description of changes:*

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -21,6 +21,7 @@ import signal
 import contextlib
 import collections.abc as collections_abc
 import locale
+from functools import partial
 import urllib.parse as urlparse
 from urllib.error import URLError
 
@@ -128,7 +129,7 @@ def getpreferredencoding(*args, **kwargs):
     )
 
 
-def compat_open(filename, mode='r', encoding=None):
+def compat_open(filename, mode='r', encoding=None, access_permissions=None):
     """Back-port open() that accepts an encoding argument.
 
     In python3 this uses the built in open() and in python2 this
@@ -139,9 +140,13 @@ def compat_open(filename, mode='r', encoding=None):
     encoding.
 
     """
+    opener = os.open
+    if access_permissions is not None:
+        opener = partial(os.open, mode=access_permissions)
     if 'b' not in mode:
         encoding = getpreferredencoding()
-    return open(filename, mode, encoding=encoding)
+    return open(filename, mode, encoding=encoding, opener=opener)
+
 
 def bytes_print(statement, stdout=None):
     """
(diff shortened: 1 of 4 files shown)


awscli/compat.py
```diff
@@ -128,7 +129,7 @@ def getpreferredencoding(*args, **kwargs):
     )
 
 
-def compat_open(filename, mode='r', encoding=None):
+def compat_open(filename, mode='r', encoding=None, access_permissions=0o755):
```
kyleknap, line 132: Instead of exposing our new abstraction, what's your thoughts on just allowing an `opener` here? So specifically, we would:

1. Update this signature to be: `def compat_open(filename, mode='r', encoding=None, opener=None)` and then provide `opener` directly  to the `open` call

2. Update the kubeconfig logic to be:
```python
      opener = partial.partial(os.open, flags=os.O_CREAT | os.O_RDWR, mode=0o600)     
      with compat_open(config.path, opener=opener) as 
```
I like this strategy a bit more for the following reasons:

1. We are limiting the blast radius in which this can affect other usages of `compat_open` as I'm not sure what will happen if we defaulted to always using an opener with those permissions set.
2. We would be using the exact flags and mode used in the PR to v1. Specifically, I noticed we were not setting the `os.O_CREAT | os.O_RDWR` that we were used in the v1 PR.
3. It gives us more flexibility in the future if we want to use something other than `os.open` to open the underlying file descriptor in `compat_open`.

Let me know what you think.

kyleknap (commented):
Looks fine. I just had one suggestion that I had about the `compat_open` interface update.

awscli/compat.py
vz10: I don't think we need to expose `opener` outside of the `compat_open` because:
1. We make people to implement redundant interface each time they need just set permissions
2. We'll always have to test provided callable because it can do whatever (and I don't think we'll ever use anything other than `os.open`)
3. These flags coming from `open` function when it calls `opener`

micahhausler:
LGTM from EKS side

kyleknap (approved):
Looks good! :ship:

awscli/compat.py
kyleknap: Capturing discussion we had.
* For the example, I gave would not work because the `open` built-in would provide flags itself resulting in duplicate `flags` kwarg. 
* In this PR, the use of the `opener` here results in the same behavior as in the PR sent to v1. The use of `opener` just makes it simpler and works with our `compat_open` abstraction.
* We opted to only use the opener if an access mode is set in order to reduce blast radius in affecting other logic that may use `compat_open`. We can adjust the default value if there is an issue we can point to and we end up needing the permission explicitly set for all opens.

PR #5937, opened 2021-02-11
Add `--proxy-url` flag to `eks update-kubeconfig`

aThorp96:
*Issue #, if available:*

*Description of changes:*
This PR adds a flag to `aws eks update-kubeconfig` in order to support proxy URLs. When supplied, the flag will add the url string, unmodified, as the `proxy-url` attribute of the `cluster` object. The attribute is not added if the flag is not provided. 

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -97,6 +97,12 @@ class UpdateKubeconfigCommand(BasicCommand):
             'help_text': ("Alias for the cluster context name. "
                           "Defaults to match cluster ARN."),
             'required': False
+        },
+        {
+            'name': 'proxy-url',
+            'help_text': ("Proxy URL to connect to the cluster endpoint. "
+                          "Defaults to None."),
+            'required': False
         }
     ]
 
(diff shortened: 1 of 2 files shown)


awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -230,12 +237,13 @@ def _expand_path(self, path):
 
 
 class EKSClient(object):
-    def __init__(self, session, cluster_name, role_arn, parsed_globals=None):
+    def __init__(self, session, cluster_name, role_arn, proxy_url, parsed_globals=None):
```
justindho, line 240: To maintain backwards compatibility for any CLI customers that have custom code utilizing this class, let's make `proxy_url` an optional argument.
```suggestion
    def __init__(self, session, cluster_name, role_arn, proxy_url=None, parsed_globals=None):
```

tests/unit/customizations/eks/test_update_kubeconfig.py
```diff
@@ -194,7 +194,7 @@ def setUp(self):
         self._session.create_client.return_value = self._mock_client
         self._session.profile = None
 
-        self._client = EKSClient(self._session, "ExampleCluster", None)
+        self._client = EKSClient(self._session, "ExampleCluster", None, None)
```
justindho, line 198: Can we also add a test for when a `--proxy-url` is specified? I imagine it would just take the returned `cluster_entry` and verify that the `proxy_url` set on it is not `None`.

justindho (changes requested):
@aThorp96 Thanks for the PR! I just had a couple comments/questions.

yuxiang-zhang:
#7623 implements the same feature and has addressed the review comments.

aThorp96:
Closing in favor of #7623

PR #5944, opened 2021-02-15
Fix s3 cp example for grants to specific users

jcragg:
*Issue #, if available:*

*Description of changes:* The example for `aws s3 cp --grants` is incorrectly specifying `uri` as the Grantee_Type for the "specific user" part of the example, it should be `id` instead as the Grantee_Id in that part of the example is a Canonical ID.


justindho:
Thanks for the PR, @jcragg!

Issue #5352, opened 2020-07-08
Update rsa requirements


limburgher:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x ] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
requires rsa <=3.5.0, and 4.6 is now released.

Initial testing with awscli 1.18.95 and rsa 4.6 shows no regressions.

jekriske-lilly:
+1 on this since  https://github.com/sybrenstuvel/python-rsa/blob/master/CHANGELOG.md

Regarding 4.6

> No functional changes compared to version 4.2.

Whereas 4.5 is virtually retagged 4.0

kdaily:
Hi @limburgher and @jekriske-lilly ,

See this PR for a recent change to the RSA requirements: https://github.com/aws/aws-cli/pull/5355

Now at a max of `<=4.5.0` depending on the Python version.

limburgher:
Thank you. I'm still modifying this on Fedora to work with 4.7.

jekriske-lilly:
@kdaily Thanks, looking for 4.6 to align with the currently released version on PyPi for which the author claims would still be compatible.

surfaceowl:
additionally, there are [multiple known security issues](https://snyk.io/vuln/search?q=rsa&type=pip) with older versions of rsa 

These have been fixed in rsa versions >=4.2.0 ...and as noted above current version is 4.6.0 (released in June 2020) -- would be great if aws would upgrade this rsa dependency soon because we have pinned the rsa version==4.6.0

Importantly `aws-cli==1.18.115` **fails to pip install using the new pip dependency resolver** currently in beta in pip==20.2.1 (python -m pip install -r requirements.txt --use-feature=2020-resolver).  **This will fail all the time when pip releases 20.3**

robeden:
Due to the security vulnerabilities (specifically [this one](https://snyk.io/vuln/SNYK-PYTHON-RSA-1038401)), it might be good to think of this as a bug (higher priority).

nateprewitt:
[...]

PR #5954, opened 2021-02-19
Update rsa pin to < 4.8

nateprewitt:
This addresses #5352 and ups our allowed versions for rsa to current release 4.7.1 for Python 3.6 and later. Python 2.7 is no longer supported by RSA and will remain pinned at 4.5.0. The CLI is not impacted by the recent [CVE-2020-25658](https://nvd.nist.gov/vuln/detail/CVE-2020-25658) but any other usage of rsa-4.5.0 in an environment with Python 2.7 may be subject to exploit.

We _strongly_ encourage anyone still using the CLI on Python 2.7 to update promptly. We'd recommend [upgrading to the CLI v2](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) which will provide isolated copies of all of it's dependencies ensuring better system security going forward.

### DIFF
--- a/scripts/assets/constraints-bundled.txt
+++ b/scripts/assets/constraints-bundled.txt
@@ -1,2 +1,1 @@
-python-dateutil==2.8.0
 rsa==3.4.2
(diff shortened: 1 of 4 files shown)


kyleknap (approved):
Looks good. :ship:

Issue #5963, opened 2021-02-21
aws ecs deploy should incorporate time to shift traffic for wait time calculation


oliver-schoenherr:
**Is your feature request related to a problem? Please describe.**

The `aws ecs deploy` command calculate the expected duration for a deployment based on the configured wait and termination time only, but in case of a canary deployment additional time is required to shift the traffic to the new version step by step. This leads to failed execution of the command, because the command gives up way to early - while the deployment is still not finished yet.

**Describe the solution you'd like**

Take the values of the traffic rerouting in consideration if the expected duration of the deployment is calculated.

**Describe alternatives you've considered**

Allow to override the expect deployment duration via some command line switch.

**Additional context**

https://aws.amazon.com/about-aws/whats-new/2020/02/amazon-elastic-container-service-support-canary-deployments/

kdaily:
Thanks for the feature request and associated PR, @oliver-schoenherr.

kdaily:
V323668604

PR #5964, opened 2021-02-21
ECS CodeDeploy: Include traffic rerouting time in deployment duration

oliver-schoenherr:
*Issue #, if available:*

#5963

*Description of changes:*

Retrieve the configured ECS CodeDeploy rerouting config and calculated the expected time base on the configuration. This time is added to the already previously calculated deployment time based on wait and termination time.

### DIFF
--- a/awscli/customizations/ecs/deploy.py
+++ b/awscli/customizations/ecs/deploy.py
@@ -39,7 +39,8 @@ class ECSDeploy(BasicCommand):
         "complete. This command will exit with a return code of 255 if the "
         "deployment does not succeed within 30 minutes by default or "
         "up to 10 minutes more than your deployment group's configured wait "
-        "time (max of 6 hours)."
+        "time + the time required for shifting the traffic in the case of canary "
+        "releases (max of 6 hours)."
     )
 
     ARG_TABLE = [
@@ -131,7 +132,7 @@ class ECSDeploy(BasicCommand):
 
         self._validate_code_deploy_resources(codedeploy_client)
 
-        self.wait_time = self._cd_validator.get_deployment_wait_time()
+        self.wait_time = self._cd_validator.get_deployment_duration()
 
         self.task_def_arn = self._register_task_def(
             register_task_def_kwargs, ecs_client_wrapper)
@@ -341,6 +342,24 @@ class CodeDeployValidator():
             raise exceptions.ServiceClientError(
                 action='describe Code Deploy deployment group', error=e)
 
+        try:
+            dgp_info = self.deployment_group_details['deploymentGroupInfo']
+            deployment_config_name = dgp_info['deploymentConfigName']
+            self.deployment_config = self._client.get_deployment_config(
+                deploymentConfigName=deployment_config_name)
+        except ClientError as e:
+            raise exceptions.ServiceClientError(
+                action='describe Code Deploy deployment config', error=e)
+
+    def get_deployment_duration(self):
+        wait_time = self.get_deployment_wait_time()
+        rerouting_time = self.get_traffic_rerouting_time()
+
+        if wait_time is None or rerouting_time is None:
+            return None
+
+        return wait_time + rerouting_time + TIMEOUT_BUFFER_MIN
+
     def get_deployment_wait_time(self):
 
         if (not hasattr(self, 'deployment_group_details') or
@@ -359,7 +378,26 @@ class CodeDeployValidator():
 
             configured_wait = deploy_ready_wait_min + termination_wait_min
 
-            return configured_wait + TIMEOUT_BUFFER_MIN
+            return configured_wait
+
+    def get_traffic_rerouting_time(self):
+        if (not hasattr(self, 'deployment_config') or
+                self.deployment_config is None):
+            return None
+        else:
+            config_info = self.deployment_config['deploymentConfigInfo']
+            routing_config = config_info['trafficRoutingConfig']
+            routing_type = routing_config['type']
+
+            if routing_type == 'AllAtOnce':
+                return 0
+            elif routing_type == 'TimeBasedCanary':
+                return routing_config['timeBasedCanary']['canaryInterval']
+            elif routing_type == 'TimeBasedLinear':
+                time_based_linear = routing_config['timeBasedLinear']
+                return int((100 / time_based_linear['linearPercentage']) * time_based_linear['linearInterval'])
+            else:
+                return None
 
     def validate_all(self):
         self.validate_application()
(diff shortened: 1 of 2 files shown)


awscli/customizations/ecs/deploy.py
```diff
@@ -341,6 +342,24 @@ def describe_cd_resources(self):
             raise exceptions.ServiceClientError(
                 action='describe Code Deploy deployment group', error=e)
 
+        try:
+            dgp_info = self.deployment_group_details['deploymentGroupInfo']
+            deployment_config_name = dgp_info['deploymentConfigName']
+            self.deployment_config = self._client.get_deployment_config(
```
allisaurus, line 348: is `deploymentConfigName` a resource which will always be present in the `get_deployment_group` response, or only when specific types of deployments are configured?

allisaurus (commented):
Thanks for cutting https://github.com/aws/aws-cli/issues/5963 and submitting this PR, @oliver-schoenherr! I have a couple initial questions, added below.

awscli/customizations/ecs/deploy.py
oliver-schoenherr: `deployemntConfigName` is kind of mandatory. From the API docs:

> If specified, the deployment configuration name can be either one of the predefined configurations provided with AWS CodeDeploy or a custom deployment configuration that you create by calling the create deployment configuration operation.
CodeDeployDefault.OneAtATime is the default deployment configuration. It is used if a configuration isn't specified for the deployment or deployment group.

If no `deployemntConfigName` is provided the system use `CodeDeployDefault.OneAtATime` as default, which leads to an error, because this config is not valid for ECS.

But to be safe, I can for sure here check the existence of the key first and skip the whole rerouting time calculation for this case.

awscli/customizations/ecs/deploy.py
```diff
@@ -341,6 +342,24 @@ def describe_cd_resources(self):
...
+                deploymentConfigName=deployment_config_name)
+        except ClientError as e:
+            raise exceptions.ServiceClientError(
+                action='describe Code Deploy deployment config', error=e)
+
+    def get_deployment_duration(self):
+        wait_time = self.get_deployment_wait_time()
+        rerouting_time = self.get_traffic_rerouting_time()
+
+        if wait_time is None or rerouting_time is None:
```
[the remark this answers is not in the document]
oliver-schoenherr: To be honest I do not understand why `get_deployment_wait_time` returns `None` in the case of no information instead of just 0. I would propose to return 0 from `get_deployment_wait_time` and `get_traffic_rerouting_time` instead of the `None` special use case. WDYT?

awscli/customizations/ecs/deploy.py
allisaurus: Thanks for the context!  Curious, this key isn't currently provided in any of our [functional tests](https://github.com/aws/aws-cli/blob/develop/tests/functional/ecs/test_deploy.py#L223)... though [they're supposed to be run](https://github.com/aws/aws-cli/blob/develop/scripts/ci/run-tests#L19-L20) as an automated check on PRs. If this code is always making a `GetDeplotmentConfig` call with it's value now, how are these passing?? 

Can you try to run the functional test locally and see what happens? Since this call will (nearly) always be made now, I also expect we'll need to update these to include it.

oliver-schoenherr:
> Can you try to run the functional test locally and see what happens? Since this call will (nearly) always be made now, I also expect we'll need to update these to include it.

Looks like the functional test are broken.

```
  File "/Users/oliverschonherr/IdeaProjects/aws-cli/awscli/customizations/ecs/deploy.py", line 347, in describe_cd_resources
    deployment_config_name = dgp_info['deploymentConfigName']
KeyError: 'deploymentConfigName'
awscli.clidriver: DEBUG: Exiting with rc 255
```

Will try to fix the issue.

oliver-schoenherr:
functional tests are now updated as well.

allisaurus:
@oliver-schoenherr  thank you so much for your tests and updates! I'll follow up on why these test failures weren't apparent via the automated check - would have preferred that info be available to you from the get-go, vs. a later ask. 
Will review the recent changes soon and follow up with any lingering questions/feedback.

kdaily:
Hi @oliver-schoenherr - I determined that the tests for this module weren't running due to a missing `__init__.py` in the test directory. 😱 

I opened a PR to your branch to add this!

https://github.com/oliver-schoenherr/aws-cli/pull/1

oliver-schoenherr:
Looks like the python tests are broken even in the develop branch without my changes. I will try my best to fix the issue.

allisaurus:
@oliver-schoenherr would it be possible for you to submit a separate PR for the `__init__` addition and Python 2.7 functional tests fix (one that does not include your `ecs deploy` duration changes)? That way we can review and test those changes in isolation from your feature contribution, and merge it in first (while we continue to review this one in parallel). What do you think of this plan?

My apologies this has turned into a bit of a production. The code base will be in a much better state on account of your contribution, so thank you in advance for the work you've already done and any additional effort you're able to put in to vault it over the finish line!

oliver-schoenherr:
@allisaurus no worries - all good. Will split the PR and create a dedicated PR just to enable and fix the functional tests.

allisaurus (commented):
awesome work, @oliver-schoenherr! I have only a few nitpicky comments re: naming/readability, but otherwise LGTM. 

re: the `Codecov` report, I'm not sure why it shows reduced coverage on several files you didn't modify - coverage of `ecs/deploy.py` has gone up by ~2%, so I'm satisfied, but would like @kdaily to confirm whether the other misses are of any concern.

Similarly: I suspect the Python 2.7 functional test failure's aren't due to anything within the `ecs deploy` code, so I'm not inclined to block approval on those.

kdaily:
Hi @oliver-schoenherr, I went ahead and opened a PR #5994 to test out the changes to adding the `__init__.py`. The tests are failing there, so the errors seen here are not related to your code. I'll work with @allisaurus and the CLI team to determine what's going on.

allisaurus (approved):
Approved! Pending we can merge/rebase on top of https://github.com/aws/aws-cli/pull/5995 and the functional tests pass :D

kdaily:
Hi @oliver-schoenherr,

Sorry for the delays. I think we've resolved the testing issues. Can you rebase your branch and we can continue getting this merged in? After @allisaurus is OK with all the changes, I'll flag it for review by the CLI team. Thanks again for your patience.

justindho:
#6179 picks up where this PR left off, so I'll be closing this PR in favor of that one.

Issue #5566, opened 2020-09-23
Wrong help text for `emr list-clusters --created-before` option


ldormoy:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
The help text for the `emr list-clusters --created-before` option is:

> List only those clusters created after the date and time specified in the format yyyy-mm-ddThh:mm:ss. For example, --created-after 2017-07-04T00:01:30.

It is obviously a copy-paste from the `--created-after` option.

**SDK version number**
CLI: 2.0.43
Online HTML Reference doc: latest

**Platform/OS/Hardware/Device**
x86_64/Linux/Dell XPS 13

**To Reproduce (observed behavior)**
aws emr list-clusters help

**Expected behavior**
The help text should be:

> List only those clusters created before the date and time specified in the format yyyy-mm-ddThh:mm:ss. For example, --created-before 2017-07-04T00:01:30.

kdaily:
Thanks for the report. @ldormoy! That documentation is part of the API model, so it will be present across all SDKs. I'll take care of getting it fixed, but can't provide an ETA as to when it will be addressed.

stobrien89:
Closing, as @ldormoy changes have been merged!

PR #5967, opened 2021-02-22
Fix emr list-clusters documentation

mprencipe:
The documentation for emr list-clusters --created-before had the same help text as --created-after. Changed references from after to before. Fixes issue #5566.


mprencipe:
Duplicate pull request.

PR #5980, opened 2021-02-25
Return UTF-8 as preferred encoding when using the "POSIX" locale

vz10:
*Issue #, if available:*

*Description of changes:*
All the logic was replicated from Python `sys.flags.utf8_mode` setting implementation 

https://github.com/python/cpython/blob/4b9aad49992a825d8c76e428ed1aca81dd3878b2/Python/preconfig.c#L648

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -123,10 +123,16 @@ def getpreferredencoding(*args, **kwargs):
     if it's not set use locale.getpreferredencoding()
     to find the preferred encoding.
     """
-    return os.environ.get(
-        'AWS_CLI_FILE_ENCODING',
-        locale.getpreferredencoding(*args, **kwargs)
-    )
+    if 'AWS_CLI_FILE_ENCODING' in os.environ:
+        return os.environ['AWS_CLI_FILE_ENCODING']
+    # in Python 3.8 PyConfig API has been changed but pyInstaller only
+    # partially supports these changes, one thing it doesn't support is
+    # auto-conversion POSIX locale to UTF-8 which is part of PEP-540
+    # so we have to implement this conversion on our side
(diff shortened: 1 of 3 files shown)


awscli/compat.py
```diff
@@ -123,10 +123,16 @@ def getpreferredencoding(*args, **kwargs):
...
-        locale.getpreferredencoding(*args, **kwargs)
-    )
+    if 'AWS_CLI_FILE_ENCODING' in os.environ:
+        return os.environ['AWS_CLI_FILE_ENCODING']
+    # in Python 3.8 PyConfig API has been changed but pyInstaller only
+    # partially supports these changes, one thing it doesn't support is
+    # auto-conversion POSIX locale to UTF-8 which is part of PEP-540
+    # so we have to implement this conversion on our side
+    lc_type = locale.setlocale(locale.LC_CTYPE)
+    if lc_type is not None and lc_type in ('C', 'POSIX'):
```
kyleknap, line 133: Do we need to do the `is None` check here? Could we just keep it as `lc_ctype in 'C', 'POSIX')`?

awscli/compat.py
```diff
@@ -123,10 +123,16 @@ def getpreferredencoding(*args, **kwargs):
     if it's not set use locale.getpreferredencoding()
     to find the preferred encoding.
     """
-    return os.environ.get(
-        'AWS_CLI_FILE_ENCODING',
-        locale.getpreferredencoding(*args, **kwargs)
-    )
+    if 'AWS_CLI_FILE_ENCODING' in os.environ:
```
kyleknap, line 126: We should also add a changelog entry noting it as a bug fix.

kyleknap (commented):
Looks good. Just two really small comments.

kyleknap (approved):
:ship:

Issue #5854, opened 2021-01-07
"aws eks update-kubeconfig" runs into "YamlError while loading kubeconfig: while scanning a simple key"


Mierpo:
I'm trying to setup AWS commandline tool to access an eks cluster from a local ubuntu installation. The following command runs into an error:

```
$ aws eks update-kubeconfig --name "[REDACTED]" --region=[REDACTED]

YamlError while loading kubeconfig: while scanning a simple key
  in "/home/[REDACTED]/.kube/config", line 30, column 1
could not find expected ':'
  in "/home/[REDACTED]/.kube/config", line 32, column 1
```

The following command also runs into the same issue:

```
$ kubectl get all

error: error loading config file "/home/[REDACTED]/.kube/config": yaml: line 32: could not find expected ':'
```

Version information:

```
$ aws --version

aws-cli/2.1.16 Python/3.7.3 Linux/4.4.0-19041-Microsoft exe/x86_64.ubuntu.20 prompt/off

$ kubectl version --client

Client Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:21:03Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

KubeConfig (redacted contents):

```
kubectl_config = apiVersion: v1
preferences: {}
kind: Config

clusters:
- cluster:
    server: [REDACTED]
    certificate-authority-data: [REDACTED]
  name: [REDACTED]

contexts:
- context:
    cluster: [REDACTED]
    user: [REDACTED]
  name: [REDACTED]

current-context: [REDACTED]

users:
- name: [REDACTED]
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: aws-iam-authenticator
      args:
        - "token"
        - "-i"
        - "[REDACTED]"

region = [REDACTED]
```

The problem does not occur on my windows machine with the following versions:

```
$ aws --version

aws-cli/2.0.26 Python/3.7.7 Windows/10 botocore/2.0.0dev30

$ kubectl version --client

[...]

PR #5981, opened 2021-02-26
Pass O_TRUNC when rewriting kubeconfig

jcaesar:
*Issue #, if available:*
The symptoms I'm seeing are the same as in #5854, but I think it was introduced by #5905, so... probably none.
[Edit:] Added after the PR: #5986 

*Description of changes:*
Updating `~/.kube/config` with commands like `aws eks --region ap-northeast-1 update-kubeconfig --name foobar` may produce invalid YAML due to an improperly truncated config file, e.g.:
```yaml
- name: zing
  user:
    exec:
      args: [] # zing
      env:
      - name: AWS_STS_REGIONAL_ENDPOINTS
        value: regional
      provideClusterInfo: false
e: AWS_STS_REGIONAL_ENDPOINTS
        value: regional
      provideClusterInfo: false
```
or
```yaml
- name: zing
  user:
    exec:
      args: [] # zing
      command: aws
      env:
      - name: AWS_STS_REGIONAL_ENDPOINTS
        value: regional
ional
```
This PR fixes that.

Note: It is possible that another instance of this behavior is at [`configure/writer.py`](https://github.com/aws/aws-cli/blob/19acfed0a6978d717845a0e134e080997bffaa1d/awscli/customizations/configure/writer.py#L76), but I have little idea what the intention of that file is and leave investigating it to someone else.

### DIFF
--- a/awscli/customizations/eks/kubeconfig.py
+++ b/awscli/customizations/eks/kubeconfig.py
@@ -193,7 +193,7 @@ class KubeconfigWriter(object):
             with os.fdopen(
                     os.open(
                         config.path,
-                        os.O_CREAT | os.O_RDWR,
+                        os.O_CREAT | os.O_RDWR | os.O_TRUNC,
                         0o600),
                     "w+") as stream:
                 ordered_yaml_dump(config.content, stream)
--- a/tests/unit/customizations/eks/test_kubeconfig.py
+++ b/tests/unit/customizations/eks/test_kubeconfig.py
@@ -79,6 +79,16 @@ class TestKubeconfigWriter(unittest.TestCase):
         stat = os.stat(config_path)
         self.assertEqual(stat.st_mode & 0o777, 0o600)
 
+    def test_truncates(self):
+        tmpdir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, tmpdir)
+        config_path = os.path.join(tmpdir, "config")
+        with open(config_path, "w+") as f:
+            f.write("#" * 100)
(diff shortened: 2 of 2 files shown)


kdaily:
V326754868

micahhausler:
@jcaesar Thank you for the fix, this looks good! Can you add the following to the existing test so we're sure we don't regress?
```diff
diff --git a/tests/unit/customizations/eks/test_kubeconfig.py b/tests/unit/customizations/eks/test_kubeconfig.py
index 977bbe446..470e89708 100644
--- a/tests/unit/customizations/eks/test_kubeconfig.py
+++ b/tests/unit/customizations/eks/test_kubeconfig.py
@@ -79,6 +79,11 @@ class TestKubeconfigWriter(unittest.TestCase):
         stat = os.stat(config_path)
         self.assertEqual(stat.st_mode & 0o777, 0o600)
 
+        config = Kubeconfig(config_path, {})
+        KubeconfigWriter().write_kubeconfig(config)
+        empty_stat = os.stat(config_path)
+        self.assertEqual(empty_stat.st_size, 3, "file should be '{}\\n', 3 bytes long ")
+
 
 class TestKubeconfigValidator(unittest.TestCase):
     def setUp(self):
```

jcaesar:
Felt like a separate test case would be proper, hope you're fine with that.
(CRLF made me [roundtrip](https://github.com/aws/aws-cli/actions/runs/619477208) once...)

micahhausler (approved):
LGTM

vz10 (approved):
🚢

Issue #4947, opened 2020-02-11
publish v2 to PyPI


mattsb42-aws:
It's great that there's a first-class supported installation mechanism that does not rely on environment specifics, but this is still a normal Python package, and for those customers who *do* already have an existing compatible environment the installation mechanism is a lot heavier-weight than using the existing Python distribution mechanisms.

Could we get v2 publishing to PyPI so that those customers who have those prerequisites can have a simpler and easier to maintain installation/updating/etc story?

joguSD:
Not publishing CLI v2 to PyPI has been a conscious decision and for the time being we have no plans to publish to PyPI. In the long term we want the fact that the AWS CLI is written in Python to be more of an implementation detail than a feature.

sbbowers:
@joguSD Could you please elaborate on the reasoning behind this conscious decision? Perhaps link us to some dialog around this discussion?

ssbarnea:
I think that time may prove if that decision was more towards conscious or in-conscious one. I am  really curious to see how this will evolve, I only hope I have enough popcorn 🍿.

On the other hand, other wheels coming from the same source did not get their notoriety without any reasons. boto, boto3 anyone?

At least pypi was 3rd party that guaranteed that a specific version is not altered once uploaded. I guess now we will have to keep saving checksums .

mattsb42-aws:
So `pip install git+git://github.com/aws/aws-cli.git#2.0.0` then.

ssbarnea:
Sure, we all know that is impossible to force push code to github, not.

adamchainz:
Or use a tarballs link for greater speed https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/

joguSD:
Just as a reminder, the AWS CLI V1 will continue to receive regular updates
that are published to PyPI and `pip` will continue to be a supported
installation mechanism for AWS CLI V1.

Here are some reasons why we dropped publishing to PyPI for AWS CLI V2, and are
not supporting `pip` as an official installation mechanism.

1) Installing with `pip` requires Python knowledge. The AWS CLI isn't a tool
exclusive to Python developers and it's very easy to end up with an invalid
installation. Often, reported issues end up being due to a misconfigured `pip`
install that's getting dependencies from the wrong place, etc. These issues
happen unexpectedly and are hard to debug which is frustrating for customers.
It's also possible to inadvertently break other packages in the same
environment.

[...]

PR #5987, opened 2021-03-04
Wheels are only universal if they support Python 2 & 3

michael-k:
*Issue #, if available:* n/a

*Description of changes:*

Remove the `universal` flag from section `wheel` in `setup.cfg`, because wheels are only universal if they support both Python 2 and 3.

----

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,3 @@
-[wheel]
-universal = 1
-
-
 [metadata]
 requires-dist =
         botocore==2.0.0dev97


kdaily:
Hi @michael-k, thanks for the PR. What's the impact of having this configured as it is now?

michael-k:
> What's the impact of having this configured as it is now?

I'm not familiar with all the details of Python's packaging.  Therefore my answer is probably incomplete.

In general when supporting Python 2 and 3 one wanted non-universal wheels when using (optional) C extensions.  Since Python 2 is less and less common and the introduction of `python_requires` (which is used by aws-cli v1, but not v2?), the impact for Python-3-only packages is probably low.

I don't know if there's any impact at all for aws-cli v2 since it's sadly not on PyPI and people installing directly from source (see eg. https://github.com/aws/aws-cli/issues/4947#issuecomment-586250169) are creating the appropriate wheel (if at all) for their version of Python.

When running the following you can see that it creates wheels `botocore-2.0.0.dev97-py2.py3-none-any.whl` (botocore v2 also has `universal` set to `1` and doesn't use `python_requires`) and `awscli-2.1.29-py2.py3-none-any.whl`.

```sh
python -m venv .venv
source .venv/bin/activate
python -m pip install wheel
python -m pip install  https://github.com/boto/botocore/archive/v2.tar.gz
python -m pip install  https://github.com/aws/aws-cli/archive/v2.tar.gz
deactivate
```

When running `python setup.py bdist_wheel` in the v2 branch it creates `dist/awscli-2.1.29-py2.py3-none-any.whl` when `universal` is set to `1` and `dist/awscli-2.1.29-py3-none-any.whl` with the change proposed in this PR.

----

TL;DR There's probably no practical impact at the moment.

michael-k:
Here's some documentation for universal wheels: https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

justindho:
It looks like the latest `setup.cfg` on the `v2` branch has this PR's proposed changes included already, so I'll be closing this PR.

PR #5988, opened 2021-03-04
Omit ResponseMetadata when Errors is empty

mojodna:
*Issue #, if available:* n/a

*Description of changes:* Previously, `ResponseMetadata` would be surfaced for responses containing an empty list of `Errors`. Since that state indicates an absence of problems, it's equivalent to Errors being fully absent.

### DIFF
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -47,7 +47,8 @@ class Formatter(object):
         # the request id) if there is an error in the response.
         # Since all errors have been unified under the Errors key,
         # this should be a reasonable way to filter.
-        if 'Errors' not in response_data:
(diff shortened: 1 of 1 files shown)


justindho:
Hi @mojodna, thanks for the PR. Unfortunately, this doesn't seem like it is backwards compatible. If existing clients expect a response to have an `Errors` key and they have a conditional that checks that the list of errors is empty before further processing, with the removal of the `Errors` key for responses with an empty list of errors, existing clients will hit a `KeyError: 'Errors'` error.

So something like this would result in a `KeyError` if this change is made.
```
if len(response['Errors']) == 0:  # this line now throws a KeyError
    # continue processing
```

PR #5994, opened 2021-03-05
add init so tests run

kdaily:
*Issue #, if available:*

Related to PR #5964.

*Description of changes:*

Add a `__init__.py` so that the `nose` test suite in the GitHub action runs them.


nateprewitt:
It looks like the tests may be broken for Python 2.7? A quick check is also showing we're missing an __init__.py in:
* `tests/functional/datapipeline`
* `tests/functional/ecs`
* `tests/functional/lex`
* `tests/functional/sagemaker`
* `tests/integration/history`

We should probably try to get these added in one go if possible. It may also be worth adding something to deployments to check for directories under tests that are missing an `__init__.py`

kdaily:
Reopened this PR at #5995. Agree @nateprewitt, testing for this would be prudent!

PR #5995, opened 2021-03-05
add init to ecs customization tests

kdaily:
*Issue #, if available:*

Related to PR #5964.

*Description of changes:*

Add a `__init__.py` so that the `nose` test suite in the GitHub action runs them.

### DIFF
--- a/tests/functional/ecs/__init__.py
+++ b/tests/functional/ecs/__init__.py
@@ -1,0 +1,12 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License'). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
--- a/tests/functional/ecs/test_execute_command.py
+++ b/tests/functional/ecs/test_execute_command.py
@@ -25,7 +25,7 @@ class TestExecuteCommand(BaseAWSCommandParamsTest):
         cmdline = 'ecs execute-command --cluster someCluster ' \
                   '--task someTaskId ' \
                   '--interactive --command ls ' \
-                  '--region use-west-2'
+                  '--region us-west-2'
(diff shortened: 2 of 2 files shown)


allisaurus (commented):
YAY! So glad these are running! 
For my own understanding, do we know why the lack of  `sort_keys=True` in the `ecs deploy` code causing Python 2.7 test to fail? Was there an implicit ordering happening somewhere else that was missing? Wondering how we'll know when to enforce similar ordering in the future

kdaily:
For Python, dictionary ordering was determined by how the keys were stored in the hash table, not the order they were inserted in. So in the conversion to a string via `json.dumps`, a different string was produced, thus a different hash. 

I think the safest thing to do would be to use `OrderedDict` - but given the pending deprecation of Python 2 in July 2021 this might not be necessary to change.

allisaurus (approved):

tests/unit/customizations/ecs/test_codedeployer.py
```diff
@@ -89,7 +89,7 @@ def test_get_create_deploy_request(self):
         self.assertEqual(actual_hash, self.deployer._get_appspec_hash())
 
     def test_get_appspec_hash(self):
-        appspec_str = json.dumps(self.deployer._appspec_dict)
+        appspec_str = json.dumps(self.deployer._appspec_dict, sort_keys=True)
```
kyleknap, line 92: So instead of updating the actual deploy logic, let's just update the `TEST_APPSPEC` to use `OrderedDict`. You should be able import the `OrderedDict` class from `awscli.compat`. The reason I'm suggesting this instead is that if we start off with an ordered dict it should guarantee order when it gets dumped meaning we do not need to update the actual source code for the command  and make this contribution be strictly a test update.

kyleknap (commented):
Looks good. I just had a comment on how to limit the change to just being a test update. Also if you can squash the commits that would be great.

kdaily:
The tests changed on this address the original `test_deploy` tests only. The `test_deploy` tests are affected by the following CloudFormation customization test:

https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/cloudformation/test_yamlhelper.py

This test relies on a YAML loader:

https://github.com/aws/aws-cli/blob/develop/awscli/customizations/cloudformation/yamlhelper.py#L91-L93

The `yaml_parse` function changes global PyYAML configuration. Any test that runs after a test that loads this will possibly fail.

The CI tests are failing due to a new test added for the execute command which needs to be addressed separately.

A possible permanent solution to the underlying ordering is proposed in #6032.

nateprewitt (approved):
Looking good, let's ship it!

PR #5999, opened 2021-03-08
Updating aws configure import description

elysahall:
Updating aws configure import description to include what headers are needed to import.


kyleknap (approved):
Thanks!

PR #6006, opened 2021-03-10
Log stream name clarity details added

aviboy2006:
*Issue #, if available:*

*Description of changes:*
Added clarity details related to log stream name. How its look a like in some of AWS account and how to specify.


kdaily:
Hi @aviboy2006, thanks for contributing.

I'm a little unclear where you're suggesting that different log stream name formats come from. A log stream name is user-defined when creating a log stream, so it could be many other things besides the formats you suggested. Depending on the operating system or name, single quotes may or may not be required. We have a documentation page on quoting string parameters:

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-quoting-strings.html

If there are still specific things about log streams that are unclear, or missing examples, I would encourage you to open up an issue here for them to be addressed. Thanks!

aviboy2006:
It is mentioned as format in string but not mentioned in example. That confused me also while copy pasting example given document. For making changes required to create issue ?

kdaily:
Hi @aviboy2006,

In general, we do not duplicate parameter documentation from the Options section into the examples to reduce the number of places to make changes. The Options now shows:

```
--log-stream-name (string)
```

This indicates that the log stream name needs to be a string. This is autogenerated from the API model. The example states the name of the log string in the paragraph and uses it in the example. One known limitation is that we do not display string limits for parameters like this. For example, the API documentation states that colons and asterisks are not allowed in names, and gives the following regex:

> Pattern: `[^:*]*`

Unfortunately I don't see any change for this specific example, since we won't duplicate documentation.

PR #6007, opened 2021-03-10
Make test_show.py compatible with py27

kyleknap:
There were two issues with the test:

* Unicode `u` values were being embedded into the show
* The order of the elements in the xml body is not stable

So updated the assertion value to be able to handle those possibilities. I also confirmed it runs correctly on Python3.


vz10 (approved):
Looks good 🚢

PR #6008, opened 2021-03-10
Force regional endpoint for test_show.py

kyleknap:
This environment variable was not previously patched so if an environment opted into using the regional sts endpoint, the test would fail.


nateprewitt (approved):
Looks good to me :shipit:. We can merge after tests pass.

vz10 (approved):
🚢

PR #6009, opened 2021-03-10
Remove unused imports, add Flake8 check for unused imports 

vz10:
Remove unused imports, add Flake8 check for unused imports into run-check action

*Issue #, if available:*

*Description of changes:*


RyanFitzSimmonsAK:
Hi @vz10, thanks for this feature request and for your patience.  We're going to close the pull request at this time.  Adding a new dependency is something we'd like to avoid in this case, and if we choose to pursue this eventually, it's something we'd prefer to handle internally.  Thanks again for contributing.

PR #6032, opened 2021-03-19
use safe loader wrapper to isolate use.

kdaily:
*Issue #, if available:*

*Description of changes:*

This addresses the underlying cause of tests for the ECS customization failing. The cloudformation customization is changing global options for PyYAML's safe loader. A solution for this is provided here:

https://github.com/yaml/pyyaml/issues/221

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -80,6 +80,12 @@ def _dict_constructor(loader, node):
     return OrderedDict(loader.construct_pairs(node))
 
 
+class SafeLoaderWrapper(yaml.SafeLoader):
+    """Isolated safe loader to allow for customizations without global changes.
+    """
+
+    pass
+
 def yaml_parse(yamlstr):
     """Parse a yaml string"""
     try:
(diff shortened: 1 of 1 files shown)


awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -80,6 +80,9 @@ def _dict_constructor(loader, node):
     return OrderedDict(loader.construct_pairs(node))
 
 
+class SafeLoaderWrapper(yaml.SafeLoader):  # type: ignore
```
vz10, line 83: Let's add some comment do we need this because it's not clear without full context.

vz10 (changes requested):
Looks good. Just a small comment

awscli/customizations/cloudformation/yamlhelper.py
vz10: And let's remove "# type: ignore" because we don't check typing anyway

kdaily: Done!

vz10 (approved):
Thank you 🚢

mgrandis (approved):

PR #6050, opened 2021-03-29
Fix metadata for install_requires specifiers

nateprewitt:
Fixing an issue with how we're specifying requirements inside of our setup.py. This format creates issues when wheels are built on Python 2.7 and then installed on Python 3, unnecessarily enforcing version pins.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,13 +29,9 @@ install_requires = [
     's3transfer>=0.3.0,<0.4.0',
     'PyYAML>=3.10,<5.5',
     'colorama>=0.2.5,<0.4.4',
+    'rsa>=3.1.2,<=4.5.0; python_version=="2.7"',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -29,14 +29,10 @@ def find_version(*file_paths):
     's3transfer>=0.3.0,<0.4.0',
     'PyYAML>=3.10,<5.5',
     'colorama>=0.2.5,<0.4.4',
+    'rsa>=3.1.2,<4.5.0; python_version=="2.7"',
```
kyleknap, line 32: This should probably be `<=` so we pull in `4.5` instead of `4.4` and are consistent with the `setup.cfg`

kyleknap (commented):
Looks fine. Just had one thing that I noticed.

kyleknap (approved):
Thanks! :ship:

Issue #5546, opened 2020-09-16
[v2] `aws configure list-profiles` errors when an invalid profile is configured, despite not relying on that information


benkehoe:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
`aws configure list-profiles` should not be affected by invalid session configuration, like an invalid profile, because it doesn't actually use the profile information. However, because it creates a session to read the config, an invalid profile causes errors.

**SDK version number**
aws-cli/2.0.48 Python/3.7.3 Linux/4.4.0-18362-Microsoft exe/x86_64.ubuntu.18

**To Reproduce (observed behavior)**
```bash
$ export AWS_PROFILE=invalid_profile
$ aws configure list-profiles
The config profile (invalid_profile) could not be found
```
No additional information is printed with `--debug`

This can't be worked around by explicitly providing `--profile` because of https://github.com/aws/aws-cli/issues/5016

**Expected behavior**
It should not be affected by invalid session configuration.

kdaily:
Thanks @benkehoe and for linking to #5016.

kdaily:
Hi @benkehoe, I confirmed that this has been fixed by #6066, available in AWS CLI version 2.1.36. Thanks again!

Issue #5016, opened 2020-02-27
Precedence of --profile CLI argument vs AWS_PROFILE env var is broken


mvallaly-rally:
According to the docs (and in accordance with most unix tools) command line arguments generally override environment variables. https://amzn.to/2tglZeS

This seems to be broken or the docs are wrong.

aws --version
aws-cli/1.17.9 Python/3.8.1 Darwin/19.3.0 botocore/1.14.9

You can easily confirm this behavior via:

AWS_PROFILE=foo aws --profile bar --version

```
$ AWS_PROFILE=foo aws --profile bar --version

Traceback (most recent call last):
  File "/usr/local/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/local/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 69, in main
    rc = driver.main()
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 203, in main
    command_table = self._get_command_table()
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 112, in _get_command_table
    self._command_table = self._build_command_table()
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 129, in _build_command_table
    self.session.emit('building-command-table.main',
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/botocore/session.py", line 674, in emit
    return self._events.emit(event_name, **kwargs)
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/botocore/hooks.py", line 228, in emit
    return self._emit(event_name, kwargs)
  File "/usr/local/Cellar/awscli/1.17.9/libexec/lib/python3.8/site-packages/botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
[...]

PR #6066, opened 2021-04-06
Fix profile resolution order

vz10:
*Issue #, if available:* #5016 #5546

*Description of changes:*

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -21,6 +21,7 @@ from botocore import xform_name
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
+from botocore.exceptions import ProfileNotFound
 from botocore.history import get_global_history_recorder
 
 from awscli import EnvironmentVariables, __version__
@@ -342,8 +343,11 @@ class ServiceCommand(CLICommand):
 
     def _get_service_model(self):
         if self._service_model is None:
-            api_version = self.session.get_config_variable('api_versions').get(
-                self._service_name, None)
+            try:
+                api_version = self.session.get_config_variable(
+                    'api_versions').get(self._service_name, None)
+            except ProfileNotFound:
+                api_version = None
             self._service_model = self.session.get_service_model(
                 self._service_name, api_version=api_version)
         return self._service_model
--- a/.changes/next-release/bugfix-profile-29812.json
+++ b/.changes/next-release/bugfix-profile-29812.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``profile``",
+  "description": "Fix bug in profile resolution order when AWS_PROFILE environment variable contains non-existing profile but `--profile` command line argument contains correct profile name"
+}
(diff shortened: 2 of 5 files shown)


tests/functional/test_args_resolution.py
```diff
@@ -0,0 +1,22 @@
...
+# language governing permissions and limitations under the License.
+from awscli.testutils import create_clidriver
+from awscli.testutils import BaseAWSCommandParamsTest
+
+
+class TestArgsResolution(BaseAWSCommandParamsTest):
+
+    def test_profile_resolution_order(self):
+        self.environ['AWS_PROFILE'] = 'foo'
+        self.run_cmd('--profile bar ec2 describe-instances', 255)
```
kyleknap, line 21: It would be nice to have the expected rc be zero by creating a profile that exists and mocking out the response. For v1, I think a non-existent profile results in a rc of `1` , but it would be a much stronger assertion/safety net to have a test that asserts we can successfully make a CLI call where the profile overrides the environment variable that has a non-existent profile.

tests/functional/test_api_versions.py
```diff
@@ -49,3 +49,8 @@ def test_command_interface_reflects_api_version(self):
         cmdline = 'ec2 describe-nat-gateways'
         _, stderr, _ = self.run_cmd(cmdline, expected_rc=2)
         self.assertIn("Invalid choice: 'describe-nat-gateways'", stderr)
+
+    def test_handles_ProfileNotFound_exception(self):
+        self.environ['AWS_PROFILE'] = 'foo'
+        # ProfileNotFound exception shouldn't be raised
+        self.run_cmd('--version')
```
kyleknap, line 56: It would be nice to explicitly add `expected_rc=0` to this command to point out the assertion is that the command succeeds.

tests/functional/test_api_versions.py
```diff
@@ -49,3 +49,8 @@ def test_command_interface_reflects_api_version(self):
         cmdline = 'ec2 describe-nat-gateways'
         _, stderr, _ = self.run_cmd(cmdline, expected_rc=2)
         self.assertIn("Invalid choice: 'describe-nat-gateways'", stderr)
+
+    def test_handles_ProfileNotFound_exception(self):
```
kyleknap, line 53: I'm thinking we do the following with this test:
1. We rename this to `test_can_get_version_with_non_existent_profile` to better capture the issues we are trying to prevent.
2. We rename the `test_args_resolution.py` to `test_args.py` and move that test there. In v2, the api version functionality all goes away so we need to find a home for this test. Also with the test name change, the test case is something a bit more general then api versions.

awscli/clidriver.py
```diff
@@ -21,6 +21,7 @@
 from botocore.compat import copy_kwargs, OrderedDict
 from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import NoRegionError
+from botocore.exceptions import ProfileNotFound
```
kyleknap, line 24: We should also a changelog `bugfix` entry for this.

kyleknap (commented):
Looks fine. I mainly had some suggestions on the tests, but nothing significant.

kyleknap (approved):
Looks good to me! :ship: It would be great if you could also send the PR to the `v2` branch as well because:
1. This will probably not merge cleanly because we had to update the api version logic which no longer exists in v2
2. It would be great if we can add a functional test to make sure we never regress on: https://github.com/aws/aws-cli/issues/5546 since that is a v2 only feature and therefore could not be added a test as part of this PR

kyleknap (approved):
Thanks! :ship:

PR #6072, opened 2021-04-07
Fix "profile" argument resolution order

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/functional/test_args.py
+++ b/tests/functional/test_args.py
@@ -1,0 +1,51 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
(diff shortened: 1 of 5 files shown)


tests/functional/test_args.py
```diff
@@ -0,0 +1,51 @@
...
+        self.entry_point = AWSCLIEntryPoint(self.driver)
+
+    def tearDown(self):
+        super(TestArgsResolution, self).tearDown()
+        self.files.remove_all()
+
+    def test_profile_resolution_order(self):
+        self.environ['AWS_PROFILE'] = 'foo'
+        self.parsed_responses = [{"Reservations": []}]
+        self.run_cmd('--profile bar ec2 describe-instances --debug', expected_rc=0)
```
kyleknap, line 39: Was there a particular reason `--debug` was being added to these commands? I also noticed that it was done in the previous PR but did not notice it till now when I saw it in this case and the `list-profiles` case.

kyleknap (commented):
Looks good. I just had a question about a small thing in the test.

kyleknap (approved):
Thanks! :ship:

PR #6098, opened 2021-04-15
Update installation recommendations

nateprewitt:
Updating the CLI recommendations for minimum versions of `pip` and `setuptools`. This patch helps clarify requirements to avoid installation complications with early releases of Python's packaging tools released prior to 2014.


kyleknap (approved):
:ship:

PR #6104, opened 2021-04-19
Use repr to format s3 exceptions

vin01:
Use of repr would ensure that exceptions without any error message (like `MemoryError`) are displayed properly to support debugging.

I recently debugged a `MemoryError` where error message would just say:

```
download failed: s3://remote to local-path
```

Having exception information in the log would be very helpful, it is not even displayed in the `DEBUG` logs currently.

I am not sure if using `repr` is the best approach here though, wdyt?


stealthycoin:
This needs an issue outlining the problem, if it gets sufficient traction we could move forward to the implementation phase.
https://github.com/aws/aws-cli/pull/6828 details our new contribution process.

PR #6107, opened 2021-04-20
Add acl to `copy-props` option for `s3 ` commands

### LINKED ISSUE #901: s3 sync does not preserve permissions across buckets

vz10:
*Issue #, if available:* 

Fixes #901

*Description of changes:*

### DIFF
--- a/.changes/next-release/enhancement-s3-50152.json
+++ b/.changes/next-release/enhancement-s3-50152.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``s3``",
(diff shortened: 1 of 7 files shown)


tim-finnigan:
Brought this PR up for discussion with the team, and the decision was to close as not planned at this time due to higher-priority ongoing issues involving S3.

PR #6109, opened 2021-04-21
[docs] fix example for "update apikey name"

benjaminxscott:
*Description of changes:*
This fixes an example for "Update apikey name" to refer to the correct jsonpatch path of `name`

Previously the example changed the apikey description, which doesn't seem to match the intent of the example

I tested the new example with aws-cli and it seems to work as expected
> aws --version
aws-cli/2.1.38 Python/3.9.4 Darwin/20.3.0 source/x86_64 prompt/off


benjaminxscott:
i'll rebase against `develop` so there's not a bunch of merge commits

Issue #5144, opened 2020-04-22
pager invoked even when there is no output


stickystyle:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
PR https://github.com/aws/aws-cli/pull/4702 introduced a default pager for output, however this pager gets invoked even when there is no output, such as writing to a file. This results in the pager getting opened with empty content.

**SDK version number**
aws-cli/2.0.8 Python/3.8.2 Darwin/18.7.0 botocore/2.0.0dev12

**Platform/OS/Hardware/Device**
OS X 10.14.6

**To Reproduce (observed behavior)**
Run a command that supports writing to a file.
`$ aws appsync get-introspection-schema --api-id $API_ID --include-directives --format SDL .api.graphql`

**Expected behavior**
The pager should not start

**Additional context**
The pager can of course be disabled with `AWS_PAGER=""`, but it doesn't make sense to even call the pager when there is nothing to display.

KaibaLopez:
Hi @stickystyle ,
Thanks for pointing this out!

kdaily:
Hi @stickystyle,

I think this has been fixed. It can be an issue on Mac where the default `LESS` environment variable is not set correctly too. The documentation has been updated to describe this and how to remedy it, under the section "How to use pager flags":

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html#cli-usage-pagination-clientside

kdaily:
@stickystyle, this should go out as part of a release next week with #6117.

kdaily:
This is now available, closing!

PR #6117, opened 2021-04-23
Not create pager process if there is no output

vz10:
*Issue #, if available:* #5144

*Description of changes:*

### DIFF
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -39,6 +39,12 @@ class LazyStdin:
             self._stream = self._process.initialize().stdin
         return getattr(self._stream, item)
 
+    def flush(self):
+        # if stream has not been created yet there is no reason to create it
+        # just to call `flush`
+        if self._stream is not None:
+            return self._stream.flush()
+
 
 class LazyPager:
     # Spin up a new process only in case it has been called or its stdin
@@ -56,6 +62,14 @@ class LazyPager:
 
     def __getattr__(self, item):
         return getattr(self.initialize(), item)
+
+    def communicate(self, *args, **kwargs):
+        # if pager process has not been created yet it means we didn't
+        # write to its stdin and there is no reason to create it just
+        # to call `communicate` so we can ignore this call
+        if self._process is not None or args or kwargs:
+            return getattr(self.initialize(), 'communicate')(*args, **kwargs)
+        return None, None
 
 
 class IMDSRegionProvider(BaseProvider):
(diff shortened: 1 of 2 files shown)


awscli/utils.py
```diff
@@ -57,6 +61,10 @@ def initialize(self):
     def __getattr__(self, item):
         return getattr(self.initialize(), item)
 
+    def communicate(self, *args, **kwargs):
+        if self._process is not None or args or kwargs:
```
stealthycoin, line 65: So if process is `None`, and `args` and `kwargs` are not provided this does nothing? Is that what we want? There is a lot of code that does this to wait for the process to terminate.

tests/unit/test_utils.py
```diff
@@ -227,16 +227,17 @@ def test_explicit_pager(self):
 
     def test_exit_of_context_manager_for_pager(self):
         self.set_session_pager('mypager --option')
-        with self.stream_factory.get_pager_stream():
+        with self.stream_factory.get_pager_stream() as stream:
+            stream.write()
             pass
```
stealthycoin, line 232: Can remove this.

tests/unit/test_utils.py
```diff
@@ -227,16 +227,17 @@ def test_explicit_pager(self):
 
     def test_exit_of_context_manager_for_pager(self):
         self.set_session_pager('mypager --option')
-        with self.stream_factory.get_pager_stream():
+        with self.stream_factory.get_pager_stream() as stream:
+            stream.write()
```
stealthycoin, line 231: Why do we need an empty write now?

tests/unit/test_utils.py
```diff
@@ -321,18 +322,13 @@ def test_create_process_on_stdin_method_call(self):
                 env={'LESS': 'ABC'}
             )
 
-    def test_create_process_on_process_method_call(self):
```
stealthycoin, line 324: The diff is confusing me here, is this test no longer needed?

stealthycoin (commented):
Can you put a short description of the reasoning behind these changes? I don't understand from just the context of the changes. It seems like you added some methods but I don't see why, or how the tests changes follow from them.

tests/unit/test_utils.py
vz10: To create a real stream  since it's lazy now.

awscli/utils.py
vz10: Because it's a pager process we spawn it, write to its stdin and communicate at the end. Due to it's lazy nature process will be created only in case we actually wrote something to the stream and if by the moment we get to the communicate step process hasn't been created yet it means nothing was written and we can just omit this step and finish.

tests/unit/test_utils.py
vz10: yes because just a `communicate` call doesn't create pager object any more

stealthycoin (approved):

Issue #2028, opened 2016-06-17
Configure List command shows wrong source from profile from env


jamesls:
I noticed in later versions of the CLI, if I have `export AWS_DEFAULT_PROFILE=name` set, then `aws configure list` will show the profile coming from the `--profile` arg.  I tracked this down to a change in the 1.7.x versions but haven't dug into what changed:

```
~ $ env | grep AWS_
AWS_DEFAULT_PROFILE=profile-from-env


~ $ aws-1.7.36 configure list
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile         profile-from-env           manual    --profile
access_key      ****************foo      config-file
secret_key      ****************bar      config-file
    region                <not set>             None    None
~ $ aws-1.7.35 configure list
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile         profile-from-env              env    ['AWS_DEFAULT_PROFILE', 'AWS_PROFILE']
access_key      ****************foo      config-file
secret_key      ****************bar      config-file
    region                <not set>             None    None
~ $ aws-1.7.34 configure list
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile         profile-from-env              env    AWS_DEFAULT_PROFILE
access_key      ****************foo      config-file
secret_key      ****************bar      config-file
    region                <not set>             None    None
```

kyleknap:
Interesting it changes so quickly across versions. We should fix this.

kdaily:
Still exists in AWS CLI v2 as well.

PR #6119, opened 2021-04-24
Fix `configure list` profiles resolution when AWS_DEFAULT_PROFILE set

vz10:
Fix `configure list` profiles resolution when AWS_DEFAULT_PROFILE env variable set

*Issue #, if available:* #2028

*Description of changes:*


stealthycoin (approved):
Bugfix changelog but other wise looks good.

PR #6124, opened 2021-04-27
Remove redundat '-' from two character pluralized acronyms in argument names

vz10:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/argrename.py
+++ b/awscli/customizations/argrename.py
@@ -132,6 +132,13 @@ HIDDEN_ALIASES = {
     'lambda.publish-version.code-sha256': 'code-sha-256',
     'lightsail.import-key-pair.public-key-base64': 'public-key-base-64',
     'opsworks.register-volume.ec2-volume-id': 'ec-2-volume-id',
+    'mgn.*.replication-servers-security-groups-ids':
+        'replication-servers-security-groups-i-ds',
+    'mgn.*.source-server-ids': 'source-server-i-ds',
+    'mgn.*.replication-configuration-template-ids':
+        'replication-configuration-template-i-ds',
+    'elasticache.create-replication-group.preferred-cache-cluster-azs':
+        'preferred-cache-cluster-a-zs'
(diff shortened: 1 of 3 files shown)


awscli/customizations/argrename.py
```diff
@@ -101,6 +101,11 @@
     'codepipeline.get-action-type.version': 'action-version',
     'ecs.*.no-enable-execute-command': 'disable-execute-command',
     'ecs.execute-command.no-interactive': 'non-interactive',
+    'mgn.*.replication-servers-security-groups-i-ds':
```
kyleknap, line 104: It is a bit concerning that there are multiple ID-related parameters that have that casing as in the future it is likely the casing will stay the same for any new resource for `mgn`. I'm wondering if there is a more general approach so we do not have to continually update this casing of IDs again. Specifically, we addressed these types of acronyms way back but it looks it only handles acronyms of length 3+ and not 2+: https://github.com/boto/botocore/pull/703. I'm wondering if we can just update this heuristic and then add hidden aliases for these arguments for backwards compatibility?

awscli/customizations/argrename.py
```diff
@@ -101,6 +101,11 @@
...
+        'replication-servers-security-groups-ids',
+    'mgn.*.source-server-i-ds': 'source-server-ids',
+    'mgn.*.replication-configuration-template-i-ds':
+        'replication-configuration-template-ids',
```
kyleknap, line 108: If we cannot generalize this approach, we should move these to the `HIDDEN_ALIASES` dictionary as if we do not we would be introducing a breaking change since the poorly named id parameters would be removed and users could be writing scripts that use these poorly named parameters because they are the only option.

kyleknap (commented):
Glad we are getting this fixed. I mainly wanted to see if we can come up with a more general solution to this casing of IDs.

kyleknap (approved):
Looks good! :ship:

PR #6158, opened 2021-05-11
Fix the help string for specifying multiple To, CC, BCC addresses

chan-vince:
*Issue #, if available:*

*Description of changes:*

**No code changes - only modification of the help strings that appear in the docs pages.**

The To, CC, and BCC fields do support multiple addresses, but not as space separated values. The API would respond with an error:

`An error occurred (InvalidParameterValue) when calling the SendEmail operation: Domain contains control or whitespace`

Instead, multiple address have to be specified with multiple `--to` fields, as alluded to in the Synopsis section further up the page.

Example - multiple To addresses:

The previous sentence suggests this would work:

```bash
$ aws ses send-email --from "sender@example.com" --to "recipient1@example.com recipient2@example.com" --subject "foo" --text "bar"
```

Whereas this is how multiple recipients should be specified:

```bash
$ aws ses send-email --from "sender@example.com" --to "recipient1@example.com" --to "recipient2@example.com" --subject "foo" --text "bar"
```

### DIFF
--- a/awscli/customizations/sessendemail.py
+++ b/awscli/customizations/sessendemail.py
@@ -28,11 +28,11 @@ from awscli.customizations.utils import validate_mutually_exclusive_handler
 
 
 TO_HELP = ('The email addresses of the primary recipients.  '
-           'You can specify multiple recipients as space-separated values')
+           'You can specify multiple recipients by specifying this option mulitple times.')
 CC_HELP = ('The email addresses of copy recipients (Cc).  '
-           'You can specify multiple recipients as space-separated values')
+           'You can specify multiple recipients by specifying this option mulitple times.')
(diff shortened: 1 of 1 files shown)


stealthycoin:
Hi @chan-vince, quotes delineate one value in bash. Space separated values implies: 
`--to value1 value2`
or 
`--to "value1" "value2"`.

`--to "a b c"` is all one value of `a b c` in bash.

The fact that you can use the same argument name multiple times and it gets synthesized to a list is an artifact of argparse, not  something we document or intend to document. I would prefer to keep this consistent with the rest of the documentation for specifying lists: https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-shorthand.html

Issue #6155, opened 2021-05-10
Typo in awscli/customizations/emr/emrfsutils.py


kokkonisd:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
There is a typo in `awscli/customizations/emr/emrfsutils.py`:
```python
def build_emrfs_confiuration(emrfs_args):
```
IIUC that should be _configuration_, not _confiuration_.
The same typo is propagated in `awscli/customizations/emr/createcluster.py`:
```python
emrfs_configuration = emrfsutils.build_emrfs_confiuration(
```

**SDK version number**

**Platform/OS/Hardware/Device**
What are you running the cli on?

**To Reproduce (observed behavior)**
Steps to reproduce the behavior

**Expected behavior**

**Logs/output**
Get full traceback and error logs by adding `--debug` to the command.

**Additional context**

kdaily:
Noted, @kokkonisd, thanks for catching that! We'll see if we can correct and alias the existing function name to keep backwards compatibility.

PR #6163, opened 2021-05-12
Fix typo in createcluster.py and emrfsutils.py

kokkonisd:
A typo is fixed: confiurations -> configurations

#6155 


stealthycoin:
We cannot change the name due to backwards compatibility. We will need to make an alias with the correct spelling, and preserve the incorrect spelling.

tim-finnigan:
Thanks for your patience here — as mentioned above, we cannot change the name due to backwards compatibility and an alias would need to be created with the correct spelling. Closing PR but continuing to track the issue in https://github.com/aws/aws-cli/issues/6155

PR #6167, opened 2021-05-13
Update wait service-stable example

godd9170:
*Issue #, if available:*
#6959 

*Description of changes:*

When trying the second example, I get the error:

```sh
aws: error: the following arguments are required: --services
```

it seems `--services` is a mandatory flag, so the example should either be `tasks-running`, or just removed altogether.


kdaily (dismissed):
Thanks @godd9170!

It looks like this should be an example for the `aws ecs wait task-running` example. Could you create a new file called `tasks-running.rst` and move it there? Thanks!

godd9170:
Hey @kdaily better late than never!

tim-finnigan (approved):

PR #6170, opened 2021-05-14
Add quotes to the last parameter in the command

akshayhiremath:
*Issue #, if available:*

*Description of changes:*

Add quotes to the last parameter in the command. Without these quotes command will return following error,
Parameter validation failed:
Invalid type for parameter IpPermissions[0].IpRanges[0], value: CidrIp=XX.XX.XX.XX/32, type: <class 'str'>, valid types: <class 'dict'>
Invalid type for parameter IpPermissions[1].IpRanges[0], value: Description=Access from Xyz, type: <class 'str'>, valid types: <class 'dict'>


akshayhiremath:
When is this going to be merged?

stealthycoin:
Thanks for the fix.

PR #6179, opened 2021-05-21
ECS CodeDeploy: Include traffic rerouting time in deployment duration

### LINKED ISSUE #5963: aws ecs deploy should incorporate time to shift traffic for wait time calculation

kdaily:
*Issue #, if available:*

Fixes #5963

Picks up on #5964 from @oliver-schoenherr to rebase develop after fixing other tests.

*Description of changes:*

Retrieve the configured ECS CodeDeploy rerouting config and calculated the expected time base on the configuration. This time is added to the already previously calculated deployment time based on wait and termination time.

### DIFF
--- a/awscli/customizations/ecs/deploy.py
+++ b/awscli/customizations/ecs/deploy.py
@@ -39,7 +39,8 @@ class ECSDeploy(BasicCommand):
         "complete. This command will exit with a return code of 255 if the "
         "deployment does not succeed within 30 minutes by default or "
         "up to 10 minutes more than your deployment group's configured wait "
-        "time (max of 6 hours)."
+        "time + the time required for shifting the traffic in the case of canary "
+        "releases (max of 6 hours)."
     )
 
     ARG_TABLE = [
@@ -131,7 +132,7 @@ class ECSDeploy(BasicCommand):
 
         self._validate_code_deploy_resources(codedeploy_client)
 
-        self.wait_time = self._cd_validator.get_deployment_wait_time()
+        self.wait_time = self._cd_validator.get_deployment_duration()
 
         self.task_def_arn = self._register_task_def(
             register_task_def_kwargs, ecs_client_wrapper)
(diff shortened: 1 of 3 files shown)


allisaurus (approved):
LGTM! Thank you :100: for all your effort and detective work @kdaily !

awscli/customizations/ecs/deploy.py
```diff
@@ -341,11 +342,26 @@ def describe_cd_resources(self):
             raise exceptions.ServiceClientError(
                 action='describe Code Deploy deployment group', error=e)
 
+        try:
+            dgp_info = self.deployment_group_details['deploymentGroupInfo']
+            deployment_config_name = dgp_info['deploymentConfigName']
+            self.deployment_config_details = self._client.get_deployment_config(
```
kyleknap, line 348: @kdaily @allisaurus I'm concerned that it appears that we are adding an new API call `GetDeploymentConfig` to the default behavior of the command? If this is the case, it seems that it would cause issues where a user has a minimally scoped IAM policy to use this command (which would not include `GetDeploymentConfig` permissions) and if they were to upgrade their version of the CLI, they may start getting permissions issues. As an alternative, we could trigger this logic behind a flag (e.g. `--wait-for-traffic-reroute`) so that users have to opt into using the API call/wait logic. Thoughts?

allisaurus: @kyleknap thanks for pointing this out, it's a valid concern. My one counter argument is that incorporating the required time for the different deployment steps was something this cmd was already doing previously, and customers could reasonably expect this to happen for free (vs. only w/ the use of a flag). Additionally, the [AWSCodeDeployDeployerAccess](https://docs.aws.amazon.com/codedeploy/latest/userguide/security_iam_id-based-policy-examples.html#managed-policies) managed policy includes `"codedeploy:Get*"`, so this population of users would not be effected by the addition of this call.

That said, I defer to your recommendation based on convention / previously seen issues elsewhere in `aws cli`. Are there other avenues (e.g., version release notes) where we could advertise the need for an additional read permission, to mitigate the minimally-scoped user concern in the absence of a flag?

tim-finnigan:
Closed as not planned at this time, but the team may revisit the proposed changes with ECS customization owners at a later date.

Issue #6120, opened 2021-04-24
Uploading a ~90GB file from stdin fails


leezu:
When uploading a ~90GB file that is streamed from stdin, awscli fails as it selects a too small parts size and runs into a part number limit: `UploadPart operation: Part number must be an integer between 1 and 10000, inclusive` The issue occurs after a few hours, supposedly once 10000 parts have been uploaded already.

For example, the following can trigger the issue:

```
aws s3 cp s3://bucket/some-large-file.xz - | xz -d | zstd -T0 -19 | aws s3 cp - s3://bucket/some-large-file.zst
```

Please increase the part size chosen by default when copying files from stdin.

stobrien89:
Hi @leezu,

Thanks for reaching out! Posting some of the points from our conversation here for visibility: 

Appending `--expected-size <file_size_in_bytes>` should resolve this.  The explanation of the behavior is found in the option description, but `cp` has quite a few options and you have to do a bit of digging to find this.  

> --expected-size (string) This argument specifies the expected size of a stream in terms of bytes. Note that this argument is needed only when a stream is being uploaded to s3 and the size is larger than 50GB. Failure to include this argument under these conditions may result in a failed upload due to too many parts in upload

I agree that an error like `Upload failed. When uploading files larger than 50GB, you must specify --expected-size to cp`  would be more helpful than the one currently provided and will discuss this with the team tomorrow.

stobrien89:
Hi @leezu,

Thanks again for bringing this to our attention.  I discussed with the team and we agreed that adding an example (#6180) that shows usage of the `--expected-size` option with an explanation of how an upload may fail would bring better visibility to this and be more helpful than changing the error message, as the error still only comes after the upload has reached 50GB in size.  Closing, but let us know if you have any additional questions!

PR #6180, opened 2021-05-22
S3 cp large file example

stobrien89:
Addresses #6120.

Added an example that shows usage of the `--expected-size` option and fixed two warning directives that weren't rendering correctly.

### DIFF
--- a/awscli/examples/s3/cp.rst
+++ b/awscli/examples/s3/cp.rst
@@ -151,7 +151,7 @@ Output::
 
 **Uploading a local file stream to S3**
 
-WARNING:: PowerShell may alter the encoding of or add a CRLF to piped input.
+.. WARNING:: PowerShell may alter the encoding of or add a CRLF to piped input.
 
 The following ``cp`` command uploads a local file stream from standard input to a specified bucket and key::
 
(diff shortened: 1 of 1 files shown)


awscli/examples/s3/cp.rst
```diff
@@ -151,16 +151,21 @@ Output::
...
 **Uploading a local file stream to S3**
 
-WARNING:: PowerShell may alter the encoding of or add a CRLF to piped input.
+.. WARNING:: PowerShell may alter the encoding of or add a CRLF to piped input.
 
 The following ``cp`` command uploads a local file stream from standard input to a specified bucket and key::
 
     aws s3 cp - s3://mybucket/stream.txt
 
+**Uploading a local file stream that is larger than 50gb to S3**
```
kyleknap, line 160: Let's make the casing here `50GB` to indicate gigabytes. If it is lowercase, users may think it is in gigabits.

awscli/examples/s3/cp.rst
```diff
@@ -151,16 +151,21 @@ Output::
...
+
+The following ``cp`` command uploads a local file stream from standard input to a specified bucket and key.  The ``--expected-size`` option must be provided, or the upload may fail when it reaches the default part limit of 10,000::
+
+    aws s3 cp - s3://mybucket/stream.txt --expected-size <expected_size_in_bytes>
```
kyleknap, line 164: It may be worth adding an actual value here instead of the placeholder `<expected_size_in_bytes>` to give an example of how the value should look, and in the prose we can mention the local file stream is the same size as that value.

kyleknap (commented):
Looks good. Just had a couple of small comments.

stobrien89:
Updated.  Let me know what you think!

kyleknap (approved):
Thanks :ship:

PR #6249, opened 2021-06-26
Changed the cutoff to 2/3 instead of 0.8

YoniMelki:
*Description of changes:*
I changed the cutoff of the `argparser` to 2/3 instead of 0.8 in order to cover more services (e.g: ec2)

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -89,7 +89,7 @@ class CLIArgParser(argparse.ArgumentParser):
                 for choice in action.choices[i:i+self.ChoicesPerLine]:
                     current.append('%-40s' % choice)
                 msg.append(' | '.join(current))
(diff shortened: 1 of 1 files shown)


stealthycoin:
Hi @YoniMelki thanks for the PR. Our team just put out a recent proposal in https://github.com/aws/aws-cli/pull/6828 detailing improvements to the contribution process. 

In order to have a PR like this looked at and reviewed we need an issue to be created that details what the change is and why. Before it can be advanced to the pull request stage it needs to have at least 10 👍s. Until then I am closing this PR.

Specifically for this PR, we would need a good reason to make this change since the impact is somewhat unknown. Examples of what the effects of this change are would help, and tests that demonstrate that nothing unexpected is also changing.

PR #6260, opened 2021-07-04
Mention to redact sensitive information

mehagar:
*Issue #, if available:*

*Description of changes:*

The Create Issue template should mention that you should redact sensitive information from the `--debug` output. I accidentally leaked a secret earlier because I didn't realize this option would produce sensitive information.

This updates it to match the equivalent line in the `---signaturedoesnotmatch-error.md` file.

### DIFF
--- a/.github/ISSUE_TEMPLATE/---bug-report.md
+++ b/.github/ISSUE_TEMPLATE/---bug-report.md
@@ -28,5 +28,7 @@ A clear and concise description of what you expected to happen.
 **Logs/output**
 Get full traceback and error logs by adding `--debug` to the command.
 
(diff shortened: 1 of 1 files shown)


kdaily:
Thanks for the suggestion, @mehagar! You're correct in that being a good idea, and we usually ask when we specifically request information. I think this would be a useful addition.

We should probably add it here, too:

https://github.com/aws/aws-cli/blob/develop/.github/ISSUE_TEMPLATE/---questions---help.md

kdaily (commented):
Add it to the question/guidance template too, please!

stealthycoin:
Closing from lack of activity.

PR #6268, opened 2021-07-09
raise SessionManagerPlugin is not found even if terminate session failed

krrrr38:
## *Issue*

When not installing `session-manager-plugin` and failed to call ssm:TerminateSession, users see only following logs.

```
botocore.exceptions.ClientError: An error occurred (AccessDeniedExceptio) when calling the TerminateSession operation:
```

## *Description of changes:*

Users want to know `SessionManagerPlugin is not found` error. So change the raising error when failing terminate-session.


stealthycoin:
I'm not sure I understand what is missing from the logs. There is a log statement https://github.com/aws/aws-cli/blob/develop/awscli/customizations/sessionmanager.py#L93 that has that information, unless I am misunderstanding.

krrrr38:
This is too old change, so I forgot the motivation, too. close it.

PR #6269, opened 2021-07-10
Fixing broken aws-cli commands in examples/

vjeffz:

### DIFF
--- a/awscli/examples/athena/update-data-catalog.rst
+++ b/awscli/examples/athena/update-data-catalog.rst
@@ -6,7 +6,7 @@ The following ``update-data-catalog`` example updates the Lambda function and de
         --name cw_logs_catalog \
         --type LAMBDA \
         --description "New CloudWatch Logs Catalog" \
-        --function=arn:aws:lambda:us-west-2:111122223333:function:new_cw_logs_lambda
+        --parameters function=arn:aws:lambda:us-west-2:111122223333:function:new_cw_logs_lambda
 
 This command produces no output. To see the result, use ``aws athena get-data-catalog --name cw_logs_catalog``.
 
--- a/awscli/examples/chime/put-voice-connector-origination.rst
+++ b/awscli/examples/chime/put-voice-connector-origination.rst
@@ -4,7 +4,7 @@ The following ``put-voice-connector-origination`` example sets up the originatio
 
     aws chime put-voice-connector-origination \
         --voice-connector-id abcdef1ghij2klmno3pqr4 \
-        --origination Routes=[{Host="10.24.34.0",Port=1234,Protocol="TCP",Priority=1,Weight=5}],Disabled=false
+        --origination Routes=[{Host="10.24.34.0"},{Port=1234},{Protocol="TCP"},{Priority=1},{Weight=5}],Disabled=false
 
 Output::
 
--- a/awscli/examples/codeartifact/get-package-version-asset.rst
+++ b/awscli/examples/codeartifact/get-package-version-asset.rst
@@ -8,6 +8,7 @@ The following ``get-package-version-asset`` example retrieves the package.tgz as
         --format npm \
         --package test-package \
         --package-version 4.0.0 \
+        --asset test-asset
         outfileName
 
 Output::
--- a/awscli/examples/codebuild/batch-get-report-groups.rst
+++ b/awscli/examples/codebuild/batch-get-report-groups.rst
@@ -3,7 +3,7 @@
 The following ``batch-get-report-groups`` example retrieves information about the report group with the specified ARN. ::
 
     aws codebuild batch-get-report-groups \
-        --report-group-arns arn:aws:codebuild:<region-ID>:<user-ID>:report-group/<report-group-name>
+        --report-group-arns "arn:aws:codebuild:<region-ID>:<user-ID>:report-group/<report-group-name>"
 
 Output::
 
--- a/awscli/examples/codebuild/batch-get-reports.rst
+++ b/awscli/examples/codebuild/batch-get-reports.rst
@@ -3,7 +3,7 @@
 The following ``batch-get-reports`` example retrieves information about the reports with the specified ARNs. ::
 
     aws codebuild batch-get-reports \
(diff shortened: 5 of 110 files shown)


kdaily:
Hi @vjeffz,

Thanks for this contribution. There are some valid fixes here, but some changes are not correct. Just to call out a few:

- The example for `awscli/examples/codestar-connections/create-host.rst` is explicitly for a third party enterprise server.
- `awscli/examples/ec2/describe-byoip-cidrs.rst` does not need `max-results`.
- Some changes to parameter length (like` awscli/examples/ssm/register-task-with-maintenance-window.rst` don't change the length of the parameter, they were already the correct length).

Please verify that all of the changes are valid within their context. Feel free to open smaller targeted pull requests, as we'll be able to merge those in faster.

Thanks!

kdaily (changes requested):
There are many great changes here, thanks! But there are too many that are not valid to pull this in all together. As it is, it will be difficult to pull this in due to the size. It would help if for some services (like `ec2`, among others) to pull out those into a separate PR to make it easier to review.

vjeffz:
> * The example for `awscli/examples/codestar-connections/create-host.rst` is explicitly for a third party enterprise server.
> * `awscli/examples/ec2/describe-byoip-cidrs.rst` does not need `max-results`.
> * Some changes to parameter length (like` awscli/examples/ssm/register-task-with-maintenance-window.rst` don't change the length of the parameter, they were already the correct length).


Hi @kdaily,

Thanks. A bit of context , I'm creating a test harness for a piece of software I'm writing. It requires valid AWS commands to be run so I turned to the AWS CLI documentation to extract every AWS command for Linux, there's roughly ~5200 commands. You can see the full list here (https://gist.github.com/vjeffz/4a644973763574695b73d4b839a244e7). During this process a lot of them failed, the fixes to get them to run are what you see in this PR.

As a user, I would expect the examples provided by AWS should pass the syntax described in the synopsis of each command without producing the validation errors.

* `awscli/examples/codestar-connections/create-host.rst`  Correct. I'll revert this change.

* `awscli/examples/ec2/describe-byoip-cidrs.rst` . If I run the example provided it fails with:
```
$ aws ec2 describe-byoip-cidrs

  aws help
  aws <command> help
aws: error: the following arguments are required: --max-results
```

* ` awscli/examples/ssm/register-task-with-maintenance-window.rst`. While some are correct, there are some that are incorrect. For e.g:
```
aws ssm register-task-with-maintenance-window \
    --targets "Key=WindowTargetIds,Values=350d44e6-28cc-44e2-951f-4b2c9EXAMPLE" \
    --task-arn "AWS-RunShellScript" \
    --service-role-arn "arn:aws:iam::111222333444:role/MaintenanceWindowsRole" \
    --window-id "mw-ab12cd34eEXAMPLE" \
    --task-type "RUN_COMMAND" \
    --task-parameters  "{\"commands\":{\"Values\":[\"df\"]}}" \
    --max-concurrency 1 \
    --max-errors 1 \
    --priority 10

Parameter validation failed:
Invalid length for parameter WindowId, value: 19, valid min length: 20
```

Throughout this particular command the examples which use a window id of `"mw-ab12cd34eEXAMPLE"` are incorrect length. The working length with a window of id `"mw-ab12cd34ef56gh78"` is correct. I thought making them all the same would make it easier for a user  while making it clear they're example window ids so I used "`"mw-06cf17cbefEXAMPLE"` (Adding an extra `f` to the original example)

kdaily:
Hi @vjeffz,

Thanks for your response! You've hit on at least one known issue (the CIDRS `max-results` should not be required) so glad to know that it's working!

> I'm creating a test harness for a piece of software I'm writing.

I'm really interested to know more about your approach here and if you could contribute it to the AWS CLI to run on our examples. The automated validation of examples is currently limited and it would be great to expand it:

https://github.com/aws/aws-cli/blob/develop/tests/functional/docs/test_examples.py

I think that starting with some easy cases to see how it works (e.g., multi-line commands should end with `\` not `/`, parameters with regular expression or length constraints should be enforced) would be overall more useful than a one-off change. Having a scoped set of things to check would help us reduce a bunch of false positives that need to be manually excluded somehow.

The big one that would need tweaking I see right now is how quoting is handled, but we can save that for another iteration.

Let me know what you think, and anything you've thought about the design for enabling/disabling specific validators.

vjeffz:
Hi @kdaily 

The approach is a bit different to what you have in your example. We're building an inspection proxy ([kivera.io](kivera.io)) and the purpose of the test harness is to check our AWS engine is working correctly which uses http properties such as method, path, headers to work out the intended AWS action of the user. 

In our CI process we have a Github Actions workflow which uses the file I linked in the previous comment to generate [BATS](https://github.com/bats-core/bats-core) tests for each command in the examples folder. All 5200+ commands then run through the proxy (obviously we block all calls at the proxy so not to create a tonne of resources in AWS). While our test scenario is a bit more complex than what you require, it's also been able to validate if the command itself is correct.  An example test:

```
@test "AWS-EC2-620 Test proxy output matches Kivera.Error and contains action: UnassignPrivateIpAddresses " {
    result=$(aws ec2 unassign-private-ip-addresses --network-interface-id eni-e5aa89a3 --private-ip-addresses 10.0.0.82  --debug 2>&1 | grep "Kivera.Error" | grep "b'")
    action="UnassignPrivateIpAddresses"
    if ! [[ "${result,,}" =~ "${action,,}" ]]; then
        echo "# error: $result" >&3
        echo '# cmd: aws ec2 unassign-private-ip-addresses --network-interface-id eni-e5aa89a3 --private-ip-addresses 10.0.0.82 ' >&3
    fi
    [[ "${result,,}" =~ "${action,,}" ]]
}
```

I'm not sure how much help this is, we can explore integrating kivera (free license) into your ci process to run the tests or give you a demo on how it works if you feel it would be valuable. Alternatively, our approach could be tweaked to run through a regular squid proxy. Happy to contribute either way.

RyanFitzSimmonsAK:
Hi @vjeffz, thanks for this pull request and for your patience.  This pull request is a little bit too broad in order to be merged.  I'm going to close it as not planned.  However, I do encourage you to open smaller, more targeted pull requests as mentioned earlier, because improving our examples is certainly something worth pursuing.

PR #6271, opened 2021-07-12
Remove Python 2.7 references/usage

nateprewitt:
As announced in our [blog post](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-python-2-7-in-aws-sdk-for-python-and-aws-cli-v1/) in January, we will be formally dropping support for Python 2.7 in the AWS CLI on July 15, 2021. This PR will make final changes required to remove references to Python 2.7, update testing infrastructure, and wheel generation settings.

For those looking to continue receiving updates after the deprecation date, we recommend following the instructions in our [blog post](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-python-2-7-in-aws-sdk-for-python-and-aws-cli-v1/) to upgrade to the AWS CLI v2 or Python 3.6+.


joguSD (approved):

scripts/make-bundle
```diff
@@ -30,13 +30,6 @@ BUILDTIME_DEPS = [
     ('wheel', '0.33.6'),
 ]
 PIP_DOWNLOAD_ARGS = '--no-binary :all:'
-# The constraints file is used to lock the version of rsa needed
-# to be 3.4.2 until we can drop py27 support. This lets us ensure
-# we're distributing a copy that works on all supported platforms.
-CONSTRAINTS_FILE = os.path.join(
```
kyleknap, line 36: I would assume we would have to remove logic wherever this is referenced? It looks like it is used in the `pip download` command? https://github.com/aws/aws-cli/blob/132c9524ba675b08c19c6073e4f3821475f302ed/scripts/make-bundle#L84-L85

kyleknap (commented):
Looks fine. I just had a comment about `make-bundle` script.

scripts/install
```diff
@@ -211,29 +197,17 @@ def main():
...
+            "or later.\n"
+            "For more information see the following blog post: "
+            "https://aws.amazon.com/blogs/developer/announcing-end-"
+            "of-support-for-python-2-7-in-aws-sdk-for-python-and-"
+            "aws-cli-v1/\n"
+        ).format(py_version[0], py_version[1])
+        print(unsupported_python_msg, file=sys.stderr)
         sys.exit(1)
 
-    if py_version in DEPRECATED_PYTHON:
```
kyleknap, line 224: I'm also wondering if we just keep it around similar to what you did for the boto3 PR: https://github.com/boto/boto3/pull/2922/files#diff-44c325661f26048dc7f7108da58a07c0bb25e4f016a6dfc37bfbcab91ceec4beR66? It is sort of annoying that we have to go back and remember/find/reintroduce this logic when a new version of Python is deprecated. I could go either way on this.

stealthycoin: Yea I thought the same thing.

nateprewitt: I waffled on this too. I think I'm fine keeping it as a no-op for now. I think we'd just need to put in place holders to make it a useful template for next time.

scripts/make-bundle
nateprewitt: Yep, good call out. I'm looking at why this wasn't caught in testing but will get it updated.

scripts/install
kyleknap: Yeah I'd even be fine with just commenting out what we have right now for Python2.7 as an example to follow in the future.

kyleknap (approved):
Awesome! :ship:

stealthycoin (approved):

Issue #6277, opened 2021-07-14
What are the default values for `--cli-connect-timeout` and `--cli-read-timeout`?


My6UoT9:
Confirm by changing [ ] to [x] below:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution

Issue is about usage on:
- [ ] Service API : I want to do X using Y service, what should I do?
- [x] CLI : passing arguments or cli configurations.
- [ ] Other/Not sure.

**Platform/OS/Hardware/Device**
EC2 Amazon2 Linux 

**Describe the question**
What are the default values for `--cli-connect-timeout` and `--cli-read-timeout`?

Please add to the [documentation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/index.html) the information about default values. Its not only those 2 which are missing, but this 2 are very important at least.

Many of the defaults can be inferred from the documentation - although I could not find information about timeouts, it would be very nice to have them in the reference.

**Logs/output**

My6UoT9:
The defaults come from botocore https://github.com/boto/botocore/commit/cc51d6b6e9df64983f0493315771f3678ab8de6c

kdaily:
Hi @My6UoT9, We'll check on this. I can see that it's documented in `botocore` here (they both default to 60 seconds):

https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html

My6UoT9:
Thanks you, excellent!

PR #6279, opened 2021-07-14
Add documentation for the default timeout values.

kdaily:
*Issue #, if available:*

#6277 

*Description of changes:*

Update docstring for timeout values. This puts it in line with the documentation for `botocore`:

https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html#botocore-config


nateprewitt (approved):
It would be nice if we could keep the value in one place but it's difficult cross-package. It's unlikely this will change outside of a major version though, so this seems fine.

PR #6281, opened 2021-07-15
Elaborated on the --additional-info parameter of create-cluster CLI c…

chatnish:
…ommand in helptext.py

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/emr/helptext.py
+++ b/awscli/customizations/emr/helptext.py
@@ -206,7 +206,7 @@ INSTANCE_COUNT = (
 
 ADDITIONAL_INFO = (
     '<p>Specifies additional information during cluster creation. To set development mode when starting your EMR cluster,'
-    ' set this parameter to <code>{"clusterType":"development"}</code>.</p>'
+    ' set this parameter to <code>{"clusterType":"development"}</code>.</p>')
 
 EC2_ATTRIBUTES = (
     '<p>Configures cluster and Amazon EC2 instance configurations. Accepts'
--- a/awscli/customizations/emr/helptext.py
+++ b/awscli/customizations/emr/helptext.py
@@ -206,7 +206,7 @@ INSTANCE_COUNT = (
 
 ADDITIONAL_INFO = (
     '<p>Specifies additional information during cluster creation. To set development mode when starting your EMR cluster,'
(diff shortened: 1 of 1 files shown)


awscli/customizations/emr/helptext.py
```diff
@@ -205,7 +205,8 @@
     ' are used for the core node type.</p>')
 
 ADDITIONAL_INFO = (
-    '<p>Specifies additional information during cluster creation.</p>')
+    '<p>Specifies additional information during cluster creation. To set development mode when starting your EMR cluster,'
+    ' set this parameter to <code>'{"clusterType":"development"}'</code>.</p>')
```
kyleknap, line 209: My guess is that the tests are not passing because there are single quotes around the parameter value. I think we will need to escape these quotes (e.g. `\'`). So update the value in the string to:
```
' set this parameter to <code>\'{"clusterType":"development"}\'</code>.</p>'
or we can just remove the single quote altogether:
```
' set this parameter to <code>{"clusterType":"development"}</code>.</p>'
```
```

kyleknap (commented):
Looks fine. Just had a suggestion which I think should get the tests passing.

awscli/customizations/emr/helptext.py
chatnish: Thanks, I made a second commit hopefully resolving the issue.

awscli/customizations/emr/helptext.py
```diff
@@ -205,7 +205,8 @@
     ' are used for the core node type.</p>')
 
 ADDITIONAL_INFO = (
-    '<p>Specifies additional information during cluster creation.</p>')
+    '<p>Specifies additional information during cluster creation. To set development mode when starting your EMR cluster,'
+    ' set this parameter to <code>{"clusterType":"development"}</code>.</p>'
```
kyleknap, line 209: Oh I think it is missing the ending parenthesis (e.g. `)`) now, which is causing the tests to fail. So line 209 should be just:
```
' set this parameter to <code>{"clusterType":"development"}</code>.</p>')
```

chatnish: My bad - wasn't careful in my last commit.

joguSD (approved):

kyleknap (approved):
Thanks!

PR #6282, opened 2021-07-15
[v2] Merge develop into v2 branch

kyleknap:
Pulled in changes from 1.19.112 to 1.20.0. I reverted the Python2 removal PR because this branch already removed support for Python 2.

### DIFF
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -23,6 +23,7 @@ import logging
 from awscli.compat import six
 from awscli.compat import get_stdout_text_writer
 from awscli.compat import get_popen_kwargs_for_pager_cmd
+from botocore.utils import resolve_imds_endpoint_mode
 from botocore.utils import IMDSFetcher
 from botocore.configprovider import BaseProvider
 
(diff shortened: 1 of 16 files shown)


stealthycoin (approved):

awscli/utils.py
```diff
@@ -129,7 +130,9 @@ def _create_fetcher(self):
         metadata_num_attempts = self._session.get_config_variable(
             'metadata_service_num_attempts')
         imds_config = {
```
zdutta, line 132: I think we're missing [`ec2_metadata_service_endpoint`](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L70) in the `imds_config`. This allows users to set a custom IMDS endpoint directly which maintains precedence over setting the endpoint mode. For more background, it is resolved [here](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L381).

zdutta (changes requested):

awscli/utils.py
kyleknap: It looks like this was never added as part of this PR: https://github.com/aws/aws-cli/pull/5844. Given v2 has never used this variable for IMDS region resolution, I'd prefer to just send this update as a separate PR since this commit is only needed because it unbreaks the merge from the develop branch.

stealthycoin (approved):

awscli/utils.py
kyleknap: Talked offline. We are going to send this update as a separate PR in order to unblock this one.

PR #6297, opened 2021-07-23
[v2] Add support for IMDS endpoint configuration

kyleknap:
This was previously only supported for retrieving IMDS credentials and now is extended for region resolution via IMDS. This is this PR where it was added to botocore's credential resolver: https://github.com/boto/botocore/pull/2179 but it never got plumbed into the IMDS region resolver.


nateprewitt (approved):
Looks good. It would be cool if we could back port these tests to v1 as well.

kyleknap:
@nateprewitt This is only needed for v2 because v1 does not do IMDS region resolution.

PR #6329, opened 2021-08-12
New CLI examples 8-12-20

elysahall:

### DIFF
--- a/awscli/examples/codeartifact/get-package-version-asset.rst
+++ b/awscli/examples/codeartifact/get-package-version-asset.rst
@@ -4,11 +4,11 @@ The following ``get-package-version-asset`` example retrieves the package.tgz as
 
     aws codeartifact get-package-version-asset \
         --domain test-domain \
-        --repo test-repo \
+        --repository test-repo \
         --format npm \
         --package test-package \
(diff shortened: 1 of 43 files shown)


awscli/examples/codeartifact/get-package-version-asset.rst
```diff
@@ -4,11 +4,11 @@ The following ``get-package-version-asset`` example retrieves the package.tgz as
 
     aws codeartifact get-package-version-asset \
         --domain test-domain \
-        --repo test-repo \
+        --repository test-repo \
         --format npm \
         --package test-package \
         --package-version 4.0.0 \
-        outfileName
+        --asset outfileName
```
kyleknap, line 11: I think we need to keep the `outfileName` positional arg per the docs: https://docs.aws.amazon.com/cli/latest/reference/codeartifact/get-package-version-asset.html as that is filename for downloading the asset. We also still need the `--asset` parameter but that would be the name asset hosted in CodeArtifact.

kyleknap (commented):
Looks fine. I just had a small comment on one of the examples.

elijahbatkoski:
What's the status on this?

PR #6340, opened 2021-08-19
Add support for Python 3.9

hugovk:
*Description of changes:*

* Add support for Python 3.9, released in October 2020.

* Replace alias `base64.encodestring` (deprecated in Python 3.1, removed in 3.1) with `base64.encodebytes` (added in 3.9)
  * https://docs.python.org/3.8/library/base64.html#base64.encodestring
  * https://docs.python.org/3.9/whatsnew/3.9.html#removed

---

And a heads-up, nose is unmaintained (last release 1.3.7 was in 2015) and will stop working in Python 3.10 (to be released October 2021 and RC now available on GHA as `3.10-dev`) due to:

> Remove deprecated aliases to Collections Abstract Base Classes from the `collections` module. 

https://docs.python.org/3.10/whatsnew/3.10.html#removed

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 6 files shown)


nateprewitt (approved):
Alright, I think we're ready to merge this now. Thanks @hugovk!

PR #6352, opened 2021-08-24
[v2][proposal] Installing AWS CLI v2 from source

kyleknap:
The proposal specifies an official mechanism to install the AWS CLI v2 from source along with an official source distribution artifact. For MarkDown rendered version of this proposal, view it [here](https://github.com/kyleknap/aws-cli/blob/source-proposal/proposals/source-install.md).

### ⚠️  Important note for community members

The AWS CLI development team is trying out a new process where proposals for new features are opened and reviewed as pull requests on GitHub, and this is the first proposal we are trying with this new process. This allows us to be more open on our thought process and solicit design feedback from community members. Currently, we do not recommend community members from opening pull requests with new proposals until we have fully solidified and documented the proposal process. However, we do recommend and encourage community engagement by:

* Reading the proposal and providing feedback. If you are in general agreement with the proposal, please add the :+1: reaction to this top-level comment. If you have any questions or suggestions, feel free to leave a comment in the pull request.

* Trying out the sample implementation (if provided) and leaving any feedback with regards to the approach. In general, the sample implementation is a proof of concept so there will be some refactoring and polishing prior to merging the implementation.

### Sample implementation

I also have a sample implementation of this proposal hosted at the [`poc-sdist`](https://github.com/kyleknap/aws-cli/tree/poc-sdist) branch on my fork. While reviewing this proposal, I'd recommend checking out this branch and trying to install the AWS CLI v2 from source yourself. After checking out the branch, the simplest set of commands to get an up and running installation from source (without clobbering your existing v2 installation) is to run the following checked out branch:
```bash
$ ./configure --prefix=/tmp/install --with-download-deps
$ make
$ make install
$ /tmp/install/bin/aws --version  # verify the CLI is installed correctly
```
If you run into any functional issues with the proof of concept feel free to comment on this pull request.

### DIFF
--- a/proposals/source-install.md
+++ b/proposals/source-install.md
@@ -15,7 +15,7 @@ source along with an official source distribution artifact.
 ## Motivation
 
 The AWS CLI v2 is available as pre-built executables for macOS,
-Linux x86_64, Linux aarch64, and Windows (64-bit). The AWS CLI v2 is also
+glibc-based Linux x86_64, glibc-based Linux aarch64, and Windows (64-bit). The AWS CLI v2 is also
 available as a Docker image. Generally, these artifacts provide coverage for
 most platforms and environments, but they do not satisfy all use cases:
 
@@ -42,6 +42,12 @@ most platforms and environments, but they do not satisfy all use cases:
   maintain full control over the code and  packages being imported into their
   distribution. Therefore, importing a pre-built executable is generally a
   [non-starter for distro maintainers](https://github.com/aws/aws-cli/issues/4842).
+
+* Users may want to patch AWS CLI v2 functionality, but in order to use the
+  patched functionality, it requires building and installing the AWS CLI v2 from
+  source. This is especially important for community members that want to test
+  changes they've made to the source prior to proposing/contributing the change in
+  the upstream GitHub repository.
 
 Full support for installing the AWS CLI v2 from source will unblock these use
 cases because users will be able to build and install the AWS CLI v2 for their
@@ -130,7 +136,12 @@ These are the requirements to build the AWS CLI v2 from source:
   to run Autotools generated files. For more information, see the
   [Appendix on Windows usage](#windows).
 
-* Python 3.8+ interpreter
+* Python 3.8+ interpreter. The minimum Python version required will increase
+  over time and follow the same timelines as the [official Python support policy
+  for AWS SDKs and Tools](https://aws.amazon.com/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/)
+  where an interpreter will only continue to be supported 6 months after its end-of-support date. For example,
+  Python 3.8 reaches end of support in October 2024, and thus in April 2025, Python 3.8 will no longer be supported
+  for installing the AWS CLI v2 from source and instead require a Python 3.9+ interpreter.
 
 * (Optional) All build and runtime Python library dependencies
   of the AWS CLI v2. This can be opted out of through configuration that
@@ -173,7 +184,7 @@ Optional Packages:
   --with-install-type=[portable-exe|system-sandbox]
                           Specify type of AWS CLI installation. Options are:
                           "portable-exe", "system-sandbox" (default is
-                          "frozen-exe")
+                          "system-sandbox")
   --with-download-deps    Download all dependencies and use those when
                           building the AWS CLI. If not specified, the
                           dependencies (including all python packages) must be
@@ -200,7 +211,7 @@ install the AWS CLI v2:
   (e.g., `aws`, `aws_completer`) will be installed. The default location is
   `/usr/local/bin`.
 
-The following options are offered to control the directories used:
+The following `configure` options are offered to control the directories used:
 
 * `--prefix` - Sets the directory prefix to use for the installation. The
   default value is `/usr/local`.
@@ -228,6 +239,22 @@ install the AWS CLI v2 as an add-on application in the `/opt` directory:
 ```
 This command results in the AWS CLI v2 being installed at `/opt/aws-cli` and
 the executables being installed at their default location of `/usr/local/bin`.
+
+In addition, the `make install` rule supports:
+
+* [`DESTDIR`](https://www.gnu.org/software/make/manual/html_node/DESTDIR.html#DESTDIR) variable - The path
+  to prepend to the configured installation prefix path when installing the AWS CLI. By default, no value is set
+  for this variable.
+
+The `DESTDIR` variable allows users to install the AWS CLI to an alternative location to the finalized installation
+location. For example, users can use the variable to install to a temporary location:
+```
+./configure --prefix=/usr/local
+make
+make DESTDIR=/tmp/stage install
+```
+These commands result in the AWS CLI v2 being installed at `/tmp/stage/usr/local/lib/aws-cli` and its
+executables located in `/tmp/stage/usr/local/bin`.
 
 
 ##### Python interpreter
@@ -405,7 +432,7 @@ These are the aspects that have **no** backwards compatibility guarantees:
   * Pulling in a new Python library dependency
   * Requiring a new system dependency
   
-  Furthermore, the `--with-downloaded-deps` option does not guarantee that all
+  Furthermore, the `--with-download-deps` option does not guarantee that all
   possible new dependencies in the future will be accounted for by the flag. For
   example, a new programming language may  be required in the future
   (e.g., Go, Rust) and the `--with-download-deps` option would likely not account
--- a/proposals/source-install.md
+++ b/proposals/source-install.md
@@ -17,7 +17,7 @@ source along with an official source distribution artifact.
 The AWS CLI v2 is available as pre-built executables for macOS,
 Linux x86_64, Linux aarch64, and Windows (64-bit). The AWS CLI v2 is also
 available as a Docker image. Generally, these artifacts provide coverage for
-most platforms and environments, but they do not satisfy all uses cases:
+most platforms and environments, but they do not satisfy all use cases:
 
 * The desired platform (e.g, [ARM 32-bit](https://github.com/aws/aws-cli/issues/5426))
   is not supported by any of the pre-built executables.
@@ -412,7 +412,7 @@ These are the aspects that have **no** backwards compatibility guarantees:
   for that new requirement.
 
 In general, users cannot make the assumption that when upgrading versions of
-the AWS  CLI v2 that using the same environment and same build steps will
+the AWS CLI v2, using the same environment and same build steps will
 always result in a successful install. If building from source in a CI/CD
 setting, users should pin to a specific version of the AWS CLI v2 code base
 using repository tags or a versioned source distribution
@@ -640,8 +640,7 @@ tools' core functionality around building C/C++ libraries is actually
 leveraged. For the purpose of being a thin, familiar interface, Autotools is
 advantageous over CMake because:
 
-* It requires little to no dependencies in order to use the build/install
-  interface.
+* Minimal dependencies required to use the build/install interface.
   
 * The end user interface is more minimal as it is scoped to running a single
   `configure` script followed by `make` commands. Users do not have to
@@ -715,10 +714,10 @@ the three goals from the motivation section.
 
 #### Q. Why are there two different installation types?
 
-By having the two different installations types, it provides more flexibility
-in how to build and install the AWS CLI v2.
+By having the two different installations types, we're able to provide
+more flexibility in how to build and install the AWS CLI v2.
 
-The `system-sandbox` provides a light-weight install mechanism to get the
+The `system-sandbox` provides a lightweight install mechanism to get the
 AWS CLI  v2 installed on a system, while following best Python practices by
 sandboxing the installation in a virtual environment. This installation is
 intended for users that want to install the CLI from source in the
--- a/proposals/source-install.md
+++ b/proposals/source-install.md
@@ -1,0 +1,1134 @@
+# Installing the AWS CLI v2 from source
+
+Proposal         | Metadata
+---------------- | -------------
+**Author**       | Kyle Knapp
+**Status**       | Proposed
+**Created**      | 24-August-2021
+
+## Abstract
+
+This document proposes an official mechanism to install the AWS CLI v2 from
+source along with an official source distribution artifact.
+
+
+## Motivation
+
+The AWS CLI v2 is available as pre-built executables for macOS,
+Linux x86_64, Linux aarch64, and Windows (64-bit). The AWS CLI v2 is also
+available as a Docker image. Generally, these artifacts provide coverage for
+most platforms and environments, but they do not satisfy all uses cases:
+
+* The desired platform (e.g, [ARM 32-bit](https://github.com/aws/aws-cli/issues/5426))
+  is not supported by any of the pre-built executables.
+  
+* The environment lacks system dependencies that are required for the pre-built
+  executable. For example, Alpine Linux uses [musl](https://musl.libc.org/),
+  but the current executables require glibc. This causes the pre-built
+  executables to not work out of the box for
+  [Alpine Linux](https://github.com/aws/aws-cli/issues/4685).
+  
+* The environment restricts access to resources only needed by the
+  pre-built executable. For example, on security hardened systems, it does not
+  give permission to shared memory
+  [which is needed](https://github.com/aws/aws-cli/issues/5769)
+  during the bootloading process of the frozen `aws` executable.
+
+* Users want to be able to install the AWS CLI using the package manager
+  standard to their environment (e.g., `brew`, `yum`, `apt`) instead of having
+  to download and use a ZIP, PKG,  MSI, etc. The software available through
+  these package managers are typically maintained by upstream distribution
+  maintainers. When maintainers import software, they  build from source to
+  maintain full control over the code and  packages being imported into their
+  distribution. Therefore, importing a pre-built executable is generally a
+  [non-starter for distro maintainers](https://github.com/aws/aws-cli/issues/4842).
+
+Full support for installing the AWS CLI v2 from source will unblock these use
+cases because users will be able to build and install the AWS CLI v2 for their
+particular environment.
+
+
+### Goals
+
+The approach for installing the AWS CLI v2 from source should satisfy the
+following goals:
+
+1. The source installation process maximizes the number of environments the
+   AWS CLI v2 can be installed on.
+   
+2. The source installation process is straightforward and intuitive. It minimizes
+   the number of cycles required to figure out how to install the AWS CLI from
+   source.
+
+3. The source installation process is language agnostic. While the AWS CLI v2
+   requires Python, users should not require experience in Python and Python
+   tooling to build from source. Furthermore, if the AWS CLI was to be
+   rewritten in a different programming language, users installing
+   from source should not have to learn a new set of build and install steps
+   and should be able to stay largely unaware of the change other
+   than possibly needing to install a new dependency.
+
+
+### Non-goals
+
+While this proposal helps distribution maintainers make the
+AWS CLI v2 available through standard package managers, it does not address
+adding official support for installing the AWS CLI v2 through them.
+It is focused on providing a mechanism for users that want or need
+to install the AWS CLI v2 from source.
+
+
+## Specification
+
+This section defines:
+
+* [The interface for installing from source](#source-buildinstall-interface)
+* [The build and install mechanics when installing from source](#buildinstall-mechanics)
+* [The details of a hosted source distribution artifact](#hosted-source-distribution)
+* [Provisional support for packaging AWS CLI v2 according to PEP517](#provisional-support-for-pep-517)
+
+### Source build/install interface
+
+The AWS CLI v2 leverages [GNU Autotools](https://www.gnu.org/software/automake/faq/autotools-faq.html)
+to install from source. In the simplest case, the AWS CLI v2 can be installed
+from source by running the following commands from the root of the repository:
+```bash
+./configure
+make
+make install
+```
+These commands are specific to Autotools where:
+
+* `./configure` -  Checks the system for all required dependencies and
+  generates a `Makefile` for building and installing the AWS CLI v2 based on
+  detected and explicitly specified configurations. See the
+  [Configuration section](#configuration) for the available configuration
+  options.
+  
+* `make` - Builds the AWS CLI v2. For details on the build mechanics,
+   see the [Build/install mechanics section](#buildinstall-mechanics).
+  
+* `make install` - Installs the built AWS CLI v2 to the configured location
+   on the system. For details on the install mechanics, see the
+   [Build/install mechanics section](#buildinstall-mechanics).
+
+After the above commands complete, the AWS CLI v2 is installed at the default
+location of `/usr/local/lib/aws-cli` and creates symlinks for the `aws` and
+`aws_completer` executables in the `/usr/local/bin` directory. These
+locations  are configurable through Autotools. See the
+[Install location configuration section](#install-location) for more
+information.
+
+#### Requirements
+
+These are the requirements to build the AWS CLI v2 from source:
+
+* An environment that can run Autotools generated files (e.g.,
+  `configure`, `Makefile`). These files are widely portable across POSIX
+  platforms, and for systems that do not come with POSIX-compliant shells like
+  Windows, there is additional software available (e.g., MSYS2) that allow users
+  to run Autotools generated files. For more information, see the
+  [Appendix on Windows usage](#windows).
+
+* Python 3.8+ interpreter
+
+* (Optional) All build and runtime Python library dependencies
+  of the AWS CLI v2. This can be opted out of through configuration that
+  downloads and uses the dependencies as part of the build step.
+  See the [Downloading dependencies configuration section](#downloading-dependencies)
+  for more information.
+
+#### Configuration
+
+Configuration for the build and install of the AWS CLI v2 is specified using
+the `configure` script. For the documentation of all configuration options, run
+the `configure` script with the `--help` option:
+```bash
+./configure --help
+```
+This is a sample output from the help page:
+```
+Configures builds and installs of the AWS CLI
+
+Usage: ./configure [OPTION]... [ENV_VAR=VALUE]...
+
+Help options:
+  -h, --help              Display help
+  -V, --version           Display version
+
+Installation directories:
+  --prefix=PREFIX         Set installation prefix. By default, this value is
+                          "/usr/local".
+  --libdir=LIBDIR         Set parent directory for AWS CLI installation. The
+                          full path to the AWS CLI installation is "LIBDIR/aws-cli".
+                          The default value for "LIBDIR" is "PREFIX/lib"
+                          (i.e., "/usr/local/lib" if "--prefix" is not set).
+  --bindir=BINDIR         Set install directory for AWS CLI executables. The
+                          default value for "BINDIR" is "PREFIX/bin"
+                          (i.e., "/usr/local/bin" if "--prefix" is not set).
+
+Optional Packages:
+  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
+  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
+  --with-install-type=[portable-exe|system-sandbox]
+                          Specify type of AWS CLI installation. Options are:
+                          "portable-exe", "system-sandbox" (default is
+                          "frozen-exe")
+  --with-download-deps    Download all dependencies and use those when
+                          building the AWS CLI. If not specified, the
+                          dependencies (including all python packages) must be
+                          installed on your system
+
+Some influential environment variables:
+  PYTHON      the Python interpreter
+
+```
+The sections below describes the most pertinent options. 
+
+
+##### Install location
+
+Source installation of the AWS CLI v2 uses two configurable directories to
+install the AWS CLI v2:
+
+* `libdir` - Parent directory where the AWS CLI v2 will be installed. The
+  path to the AWS CLI v2 installation is `<libdir-value>/aws-cli`. The default
+  `libdir` value is `/usr/local/lib` making the default installation directory
+  `/usr/local/lib/aws-cli`
+  
+* `bindir` - Directory where the AWS CLI v2 executables
+  (e.g., `aws`, `aws_completer`) will be installed. The default location is
+  `/usr/local/bin`.
+
+The following options are offered to control the directories used:
+
+* `--prefix` - Sets the directory prefix to use for the installation. The
+  default value is `/usr/local`.
+
+* `--libdir` - Sets the `libdir` to use for installing the AWS CLI v2. The
+  default value is `<prefix-value>/lib` (i.e., `/usr/local/lib/` if
+  `--prefix` is not specified).
+
+* `--bindir` - Sets the `bindir` to use for installing the AWS CLI v2
+  executables. The default value is `<prefix-value>/bin`
+  (i.e., `/usr/local/bin/` if both `--prefix` is not specified).
+
+For example, users can use the `--prefix` option to do a user install of the
+AWS CLI:
+```
+./configure --prefix=$HOME/.local
+```
+This command results in the AWS CLI v2 being installed at
+`$HOME/.local/lib/aws-cli` and its executables located in `$HOME/.local/bin`
+
+Users can also use more granular options such as the `--libdir` option to
+install the AWS CLI v2 as an add-on application in the `/opt` directory:
+```
+./configure --libdir=/opt
+```
+This command results in the AWS CLI v2 being installed at `/opt/aws-cli` and
+the executables being installed at their default location of `/usr/local/bin`.
+
+
+##### Python interpreter
+
+Through the [`AM_PATH_PYTHON`](https://www.gnu.org/software/automake/manual/html_node/Python.html)
+Autoconf macro, the `./configure` script automatically selects a Python
+interpreter installed on the system that is of version 3.8 or higher to use
+in building and running the AWS CLI v2. The Python interpreter to use can
+be explicitly set using the `PYTHON` environment variable when running the
+`configure` script:
+```bash
+PYTHON=/path/to/python ./configure
+```
+
+
+##### Downloading dependencies
+
+By default, it is required that all build and runtime dependencies of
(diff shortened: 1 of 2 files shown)


kellertk:
I was able to get this to successfully build on Windows 10 using [git-bash](https://git-scm.com/download/win), [MinGW](http://mingw-w64.org/doku.php), and a manually installed [CMake](https://cmake.org/download/). The configure script didn't check for the presence of CMake, which is needed to build `awscrt`, but manually installing it was simple enough. The only thing I'd change is to somehow get the `make install` script to add the installed artifacts to PATH.

Still, great success!

```sh
C:\Users\theto>"c:\Program Files\AWSCLI\bin\aws.cmd" --version
aws-cli/2.2.1 Python/3.9.6 Windows/10 source/AMD64 prompt/off
```

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+## Abstract
+
+This document proposes an official mechanism to install the AWS CLI v2 from
+source along with an official source distribution artifact.
+
+
+## Motivation
+
+The AWS CLI v2 is available as pre-built executables for macOS,
+Linux x86_64, Linux aarch64, and Windows (64-bit). The AWS CLI v2 is also
```
graywolf-at-work, line 18: The linux should probably also specify glibc.

graywolf-at-work:
I can confirm the poc-sdist branch works on alpine 3.14:

```
ea1831aaab68:~/aws-cli$ /tmp/aws/bin/aws --version
aws-cli/2.2.1 Python/3.9.5 Linux/5.10.60-1-lts source/x86_64.alpine.3 prompt/off
```

It takes quite long time but I guess that is expected.

pditommaso:
> I can confirm the poc-sdist branch works on alpine 3.14

Sorry for jumping in this thread, but this is very good news 👍

KTamas:
so happy this works on alpine now, can't wait for it to be merged

rvandegrift:
Hi @kdaily - I've played with this branch a bit from the distro packaging point of view.  Thanks very much for your effort here!  The debian cloud team has had requests for packages for awscli v2, and we'd like to be able to provide them.

I haven't gotten very far, but already have some questions.  Some of these things are probably bugs on other repos, and I'm sure others are misunderstandings.  Please pardon my ignorance, I'm new to the awslabs tools used in this repo.

# Build order
Is there a reasonable build order known?  I started to recurse through git modules, but it was hard to be confidant I had things correct.

# s2n and libcrypto
Would the team react negatively to awscliv2 packages that used s2n's openssl libcrypto support instead of aws-lc?  In this config, s2n was easy to get going on, while aws-lc would be more work.

# static vs dynamic linking
So far, most of the modules I've looked at default to building statically.  Some (e.g. `aws-c-common`) appear to not have any support for building dynamically.  Others (e.g., `s2n`) provide an option.

Generally, dynamic libs are preferred in distro packages.  But if that's not really/well/fully supported by the upstream projects, that'd be really useful to know.

# litani & aws-templates-for-cbmc-proofs
Neither one defines the usual kind of python package.  Both look like they are just designed to be included in the source of a new project.  Am I understanding their intended usage correctly?

Debian's packages are only built from components in debian, not gitmodules.  So we'll need to work a bit differently with these tools.  Here's one way this could work.  During the build of e.g. `aws-c-common`, we'll install a package for `aws-templates-for-cbmc-proofs` and run the required setup in the unpacked source.  I can think of two issues here:
1. The interactive repo setup script won't work during automated builds.  That looks easy enough to fix.
2. There will be one version of these tools shared by all packages in a given release.  Are we likely to run into incompatibilities, like module A requires rev X, but module B requires rev Y?

# Build instructions/docs
No question here, just some feedback - some modules do not have build instructions.  Some assume you're building on Ubuntu in an environment emulating CodeBuild.  Generally, distro folks will have an easier time with direct build instructions.

Thanks again for this effort!

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+
+* Dependencies. The AWS CLI v2 will add new dependencies in the future. This
+  means users must install any new dependencies to their system in order to
+  install the AWS CLI v2 from source. This includes anything from:
+  
+  * Increasing minimum required Python version
+  * Pulling in a new Python library dependency
+  * Requiring a new system dependency
+  
+  Furthermore, the `--with-downloaded-deps` option does not guarantee that all
```
lorengordon, line 408:
```suggestion
  Furthermore, the `--with-download-deps` option does not guarantee that all
```

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+
+* Port all information from the `setup.py` to the `setup.cfg` that can be
+  programmatically parsed for runtime dependencies.
+  
+* Introduce a `pyproject.toml` file (see [PEP 518](https://www.python.org/dev/peps/pep-0518/))
+  and in-tree build backend that builds the autocomplete index when building
+  both the sdist and wheel
+  (see [PEP 517](https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends)).
+  
+* Pull in and maintain the unreleased `botocore` `v2` branch along with
```
lorengordon, line 508: This makes me so sad. Just release `botocore` v2 and `s3transfer`! So much easier for the entire community.

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+following goals:
+
+1. The source installation process maximizes the number of environments the
+   AWS CLI v2 can be installed on.
+   
+2. The source installation process is straightforward and intuitive. It minimizes
+   the number of cycles required to figure out how to install the AWS CLI from
+   source.
+
+3. The source installation process is language agnostic. While the AWS CLI v2
```
lorengordon, line 63: This feels like such an odd goal and the result of overthinking and overcomplicating things. The installation process can be language agnostic, but the build process _never_ is. It is the responsibility of users _choosing_ to build from source or install using language-specific tooling to understand how that works. You don't have to *promote* language-specific tooling as an official method and you can guide users to your own packaged and hosted installers, but language-specific methods should be supported and maintained (not "provisional"). As a python project implementing a cli utility, users should be able to simply `pipx install awscliv2`.

If you switch to Go, for example, users will want to be able to use `go install ...` and if this were a Go project that _should_ just work! Switching languages like this would almost certainly qualify as a major release anyway, and so have no backwards combability guarantees, not of the user interface and not of the build process!

proposals/source-install.md
kyleknap: > So much easier for the entire community.

Could you elaborate on what specifically will make things easier? If we were to release a new major version of botocore using what is currently on the v2 branch, it would have the following implications:

* Right now the changes on the v2 branch is only scoped to [changing defaults and removing functionality](https://github.com/boto/botocore/pulls?q=is%3Apr+is%3Aclosed+base%3Av2+NOT+develop+in%3Atitle+NOT+bump+in%3Atitle+). The major version release would not address many of the popular feature requests that would only be feasible in a major version bump (e.g., [Async support](https://github.com/boto/botocore/issues/458), [modularization](https://github.com/boto/botocore/issues/1543), [out of the box typing support](https://github.com/boto/boto3/issues/1055)) and likely require larger rewrites of botocore (or even replace botocore outright). Currently, the maintainers of the Python SDK are exploring how to go about making these changes such as in their work writing the [Amazon Transcribe Streaming SDK](https://github.com/awslabs/amazon-transcribe-streaming-sdk).

* It would require a major version bump of `boto3` and force many downstream Python packages to account for new major versions of `botocore`/`boto3` (e.g., by updating their version range on `botocore`/`boto3` or making major version bumps themselves). If appropriate adjustments aren't made, then these downstream packages would not be usable alongside the latest version of `botocore`.

In the end, given the current state of changes in the `v2` branch, a major version bump does not seem worth the churn it would cause to the community, especially because it would be missing key features that would require another major version bump down the road.

proposals/source-install.md
kyleknap: > It is the responsibility of users *choosing* to build from source or install using language-specific tooling to understand how that works.

What if building from source is their only choice? A large number of the users that will be leveraging this proposal will be the ones that have no other option to install the AWS CLI v2 (i.e., they are running on platforms where there is no official installer). Furthermore, these users are not all going to be experienced in any particular language, and nor should they need to be experienced in order to get a language-agnostic tool installed on their system. For example with v1, using Python tooling was often the only way to get the CLI installed on their system and this often bit non-Python developers whether they had to spend an significant amount of cycles trying to install the CLI or broke other tools installed on their system. I think if there were other valid alternatives to install the AWS CLI v2 for this group of users, the goal could be relaxed/removed.

> language-specific methods should be supported and maintained (not "provisional"). As a python project implementing a cli utility, users should be able to simply `pipx install awscliv2`.

Maybe provisional has the wrong connotations and there's a better word or something needs to be more explicit in the definition? But the idea is that as long as the AWS CLI is a Python project, we will comply to Python standards (e.g., PEP517) to enable building/installing of the AWS CLI using Python tooling. For example with `pipx` (assuming it complies with PEP517 in building a wheel), you should be able to just run `pipx install .` on the repository. In the end, if a specific tool supports PEP517 and only requires support for that standard, we are not going to prevent users from using that specific tool on the CLI codebase; we'd likely just not document/promote using the tool on the codebase.

The idea with labeling it as provisional is that allows the project to adapt to whatever language-tooling is appropriate based on the repository and language community. For example, it does not make sense to support PEP517 if the codebase was entirely written in Go. And yes while if we were to rewrite it in Go, it would likely warrant a major version bump, breaking PEP517 compliance would not necessarily require a complete rewrite. For example:
* We could introduce a new separate executable to the AWS CLI package (similar to the separate  `aws_completer` we ship in the package today) that could be written in a different language.

* A new PEP in the future could replace PEP517 (i.e., come out with a new format to replace wheels). Ideally the CLI project would just support this new PEP while dropping support for PEP517 in order to follow what is recommended for a Python project.

kyleknap:
> Can "community contributions to the AWS CLI" be included in the motivations? Without the ability to build from source, it's impossible for a potential contributor to test their changes. I know that's not an immediate need but this is a building block for it.

@benkehoe Yep I can add that. As of right now there are [dev instructions](https://github.com/aws/aws-cli/tree/v2#cli-dev-version) in order to be able to work on feature development, but this proposal will make it easier to get  up and running in order to make contributions.

proposals/source-install.md
kyleknap: I have a bullet point for it later on, but I'll see if I can include a note earlier.

proposals/source-install.md
lorengordon: That makes sense. I didn't realize the botocore v2 dependency of the released v2 awscli was so incomplete! That's a bit of an ugly dependency tree for the v2 cli. Sorry I'm so bitter. Appreciate the detailed reply. I'll try to be more kind.

So, would it not work to publish the dev releases of botocore v2? There's no compatibility promises that way (except to yourselves). And you could still include "normal" python dependency references and installation that way. Pip won't install dev releases unless you tell it to explicitly.

* https://www.python.org/dev/peps/pep-0440/#developmental-releases
* https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions

kyleknap:
> It takes quite long time but I guess that is expected.

@graywolf-at-work Yeah depending on the environment it is being built on, most of the time it is going to be spent building `awscrt`dependency and/or building the auto-complete SQLite index.

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+                          default value for "BINDIR" is "PREFIX/bin"
+                          (i.e., "/usr/local/bin" if "--prefix" is not set).
+
+Optional Packages:
+  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
+  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
+  --with-install-type=[portable-exe|system-sandbox]
+                          Specify type of AWS CLI installation. Options are:
+                          "portable-exe", "system-sandbox" (default is
+                          "frozen-exe")
```
lorengordon, line 176: If the options are `portable-exe` or `system-sandbox`, what is the default of `frozen-exe`?

kyleknap:
@rvandegrift all of these questions seem specific to the [`awscrt`](https://pypi.org/project/awscrt/) dependency. I talked to the maintainers of the [`aws-crt-python`](https://github.com/awslabs/aws-crt-python) and [`aws-c-common`](https://github.com/awslabs/aws-c-common) libraries and got some answers to help with building it:

> Is there a reasonable build order known? I started to recurse through git modules, but it was hard to be confidant I had things correct.

There is no public documentation on this, but they have a visualization on how they all depend on each other in order to dictate build order:
image

> Would the team react negatively to awscliv2 packages that used s2n's openssl libcrypto support instead of aws-lc? 

It's fine to use openssl with the caveat about dynamic linking in the next reply

> So far, most of the modules I've looked at default to building statically. Some (e.g. aws-c-common) appear to not have any support for building dynamically. Others (e.g., s2n) provide an option. Generally, dynamic libs are preferred in distro packages. But if that's not really/well/fully supported by the upstream projects, that'd be really useful to know.


All of the C libraries can be built as shared, but the AWS Common Runtime (CRT) itself assumes (as in has never been tested otherwise) that the native part is a single embedded shared object.  They don't see why it wouldn't work (installing all the C libraries as .sos and then the crt just having the tiny bindings in its embedded .so) but there are no guarantees on binary compatibility between releases of the C libraries - they're intended to be a point-in-time snapshot.

> litani & aws-templates-for-cbmc-proofs: Neither one defines the usual kind of python package. Both look like they are just designed to be included in the source of a new project. Am I understanding their intended usage correctly?

These support CI-time correctness proofs that do not need to be bundled


I hope this helps! Let us know if you have any follow up questions.

proposals/source-install.md
lorengordon: >What if building from source is their only choice? A large number of the users that will be leveraging this proposal will be the ones that have no other option to install the AWS CLI v2 (i.e., they are running on platforms where there is no official installer).

This really seems to be taking on an awful lot yourselves and introducing a lot of complexity as a result. At some point, there has to be a boundary between this project and the user's environment. I might suggest focusing on creating official installers for those platforms?

And, the approach detailed here is still building from source and requires a healthy python install to make that work! So I don't really see much difference, I guess. Users that, before, had trouble managing python runtime isolation and installing packages into a specific python runtime will still have trouble, won't they? It is certainly nice to provide the logic for creating the portable and sandbox packages, though. Certainly do appreciate that!

>this often bit non-Python developers whether they had to spend an significant amount of cycles trying to install the CLI or broke other tools installed on their system.

Publishing to PyPi and constraining the recommendation for a python-based cli installation to use `pipx` instead of `pip` might go a long way to reducing these problems, I think.

proposals/source-install.md
kyleknap: Oh that is a typo in the POC. The default should be `portable-exe`. The `frozen-exe` option does not exist and I believe that was just the name that I originally gave `portable-exe`.

proposals/source-install.md
kyleknap: > Sorry I'm so bitter. Appreciate the detailed reply. I'll try to be more kind.

No worries. The frustration is understandable given our inconsistent communication with the community about the conversations/decisions happening around issue prioritization and project direction (which we are trying to improve).

> I didn't realize the botocore v2 dependency of the released v2 awscli was so incomplete! That's a bit of an ugly dependency tree for the v2 cli.

I completely agree. Originally the intention was to fast-follow with a major version bump of botocore/boto3 so we would not be sitting on this unreleased branch of botocore for long, but the direction has changed as we realized it was going to require much larger breaking changes to implement the larger features that the community wants.

> So, would it not work to publish the dev releases of botocore v2?

These are the main concerns:

* If you are going to try to install the AWS CLI v2 side-by-side with `boto3` as a Python package installation, it would not be feasible because `boto3` relies on v1 of `botocore` and the CLI v2 would require this dev version. For Python developers, they should be savvy enough to install the CLI into a separate `virtualenv` or use `pipx` . However, this becomes really problematic for Linux distro maintainers where they only package Python packages at the system level without a virtual environment (because virtualenvs work against their packaging principles). As alternative, distro maintainers should be able package using the `portable-exe` install type, but if we were to rely on a public dev version for `botocore` v2, it would block their ability to treat packaging the AWS CLI v2 as a typical Python package. And we want to make sure we are minimizing roadblocks for distro maintainers to pull in v2.

* Because a larger rewrite will be needed for the larger changes that the Python SDK maintainers want to make, there is a reasonable chance it warrants creating a new package name (e.g. `botocore2` or even depart from the `botocore` name entirely). If we start releasing public dev versions of `botocore` v2 and decide to use a new package name for v2, it creates an awkward story where we've been releasing dev versions of `botocore` v2 but the actual GA version is a completely separate package on PyPI.

In the end, I do not like that we'd be pulling in the unreleased botocore branch into the codebase, but at the current moment, seems like the most attainable option that unblocks issues installing the AWS CLI v2 from source and is also a two-way door. Long term, I would like to see us move a way from this approach through any of these options below:

* Integrate with whatever the new major version of the Python SDK would be.
* Figure out if there is a path forward to re-integrate with `botocore` v1 if the new major version of the Python SDK is not a viable path forward. Theoretically it seems feasible? But there may be some issues that make it not straightforward. 
* Rely on another language SDK or more directly rely on the [Python bindings for the AWS Common Runtime](https://github.com/awslabs/aws-crt-python) as that project expands.

Thoughts?

lorengordon: Ahh, so if a botocore v2.dev version is installed in the `botocore` namespace, I can see how that could conflict with other python development. Practices in this area of python have [come a long way the last few years](https://ben11kehoe.medium.com/my-python-setup-77c57a2fc4b6), and not using a virtualenv for that python development, and not using `pipx` for cli installs...yeah that's going to end in pain.

In that case, a new namespace, e.g. `botocore2`, and publishing _only_ dev versions for now makes a lot of sense to me. It would also allow other python libraries to start integrating with it without jumping through hoops, to explore new features or any interface changes they may need to make.

I don't think I have enough experience with the distro packaging world to understand this piece:

>if we were to rely on a public dev version for botocore v2, it would block [distro maintainers] ability to treat packaging the AWS CLI v2 as a typical Python package

Why wouldn't they be able to use the public dev version? I'm probably off base, but I'd think they would also package "botocore2" and establish the dependency the same way they would for other packages?

rvandegrift:
> @rvandegrift all of these questions seem specific to the [`awscrt`](https://pypi.org/project/awscrt/) dependency. I talked to the maintainers of the [`aws-crt-python`](https://github.com/awslabs/aws-crt-python) and [`aws-c-common`](https://github.com/awslabs/aws-c-common) libraries and got some answers to help with building it:

Yes, you're right- sorry about that, but many thanks for the response.

> > Is there a reasonable build order known? I started to recurse through git modules, but it was hard to be confidant I had things correct.
> 
> There is no public documentation on this, but they have a visualization on how they all depend on each other in order to dictate build order:

That's great, thanks!
> > Would the team react negatively to awscliv2 packages that used s2n's openssl libcrypto support instead of aws-lc?
> 
> It's fine to use openssl with the caveat about dynamic linking in the next reply

As I've worked through more of the C SDK dependencies, I've run into some which don't actually support this.  In particular, the docs for `aws-c-cal` say that `aws-lc` is required, and https://github.com/awslabs/aws-c-cal/issues/70 says the openssl path isn't ready in 0.5.  So I'm giving it a go with `aws-lc`.

> All of the C libraries can be built as shared, but the AWS Common Runtime (CRT) itself assumes (as in has never been tested otherwise) that the native part is a single embedded shared object. They don't see why it wouldn't work (installing all the C libraries as .sos and then the crt just having the tiny bindings in its embedded .so) but there are no guarantees on binary compatibility between releases of the C libraries - they're intended to be a point-in-time snapshot.

Thanks, that's really good to know.  Since they don't intend on ensuring binary compatibility, I think that undermines all of the value we'd get from shared builds.  So I'll continue as if this wasn't an option.

> > litani & aws-templates-for-cbmc-proofs: Neither one defines the usual kind of python package. Both look like they are just designed to be included in the source of a new project. Am I understanding their intended usage correctly?
> 
> These support CI-time correctness proofs that do not need to be bundled

Ah great - that'll substantially simplifies my first pass.

proposals/source-install.md
rvandegrift: > I don't think I have enough experience with the distro packaging world to understand this piece:

@lorengordon I haven't followed the `botocore` situation that closely, but here's what's going on in general terms.

If `botocore` v2 were packaged in a separate namespace, then packaging a public dev version would be easy - you just install both next to each other.  But that's not what's happening: at least for now, they're continuing to use the `botocore` namespace.

If packagers ship it as-is, as `botocore`, the backwards incompatibilities risk impacting a huge number of apps.  If packagers patch it to `botocore2`, that risk is resolved.  But then every app that wants v2 needs to be patched as well.  So it's not very attractive.

Debian has a transition process to support incompatible upgrades - new version is uploaded into experimental, apps that depend on it are tested, bugs are filed for upgrades, when they're all fixed the library can be upgraded in unstable.  But until v2 has a stable api and version, this is unattractive: if another update causes more compatibility breaks, another transition would be required for the update.

rvandegrift: > Thoughts?

@kyleknap I agree that bundling the `botocore` source isn't ideal, but I agree that it sounds like a reasonable trade-off.  Hopefully the situation eventually gets simpler, and the embedded copy can go away.

Also, thanks for thinking about the challenges facing distros, it's definitely appreciated.

kyleknap: > Also, thanks for thinking about the challenges facing distros, it's definitely appreciated.

@rvandegrift Yep! Removing these challenges is important to us. While our official installers (PKG, MSI, Linux installer) are not going away, long term it is going to be a better experience for everyone if they can install the AWS CLI v2 through their standard package manager.

kyleknap: > In that case, a new namespace, e.g. `botocore2`, and publishing only dev versions for now makes a lot of sense to me. It would also allow other python libraries to start integrating with it without jumping through hoops, to explore new features or any interface changes they may need to make.

@lorengordon Yeah I think is the strategy that we would like to see if we went down the route of integrating with the new major version of the Python SDK. Specifically, the maintainers of the AWS Python SDK would publish `0.x.y` versions under the new namespace and the AWS CLI would try to keep up with any interface changes they made while progress was still being made. This approach would also allow other Python developers to give feedback on the changes being made.

If we were to try to integrate with the package before it was GA'd, we'd definitely need to be mindful of a couple things as a GA'd project relying on a non-GA'd project:

* We should make sure to integrate once we are confident that there would be no significant changes to the client/session interfaces. If there were significant changes to those interfaces, the hope would be that there is some back compat interface to ease the migration as it could be potentially an enormous effort to update all client usage and we would only have to make the change once.
 
* The moment we do integrate it makes it more difficult the new project to make larger changes quickly as a large GA'd project is relying on it. So if we just put whatever was on the `v2` branch published under a`botocore2` namespace and immediately started relying on it, it will be impede/slow down progress on many of the broader, more invasive changes the maintainers of the Python SDK team do want to make. This was similar to the situation when AWS CLI v1 GA'd but `botocore` was still a non-GA'd project. While it was still feasible to make larger changes in the `botocore` codebase, making changes was not straightforward, and at times, design choices were made with emphasis on the AWS CLI usage, resulting in tight coupling and interfaces that could be better generalized/defined. With this new major version, it would be ideal to see the larger changes have a stronger foundation before on-boarding the AWS CLI to not too strongly influence the approach.

kyleknap:
> As I've worked through more of the C SDK dependencies, I've run into some which don't actually support this. In particular, the docs for `aws-c-cal` say that aws-lc is required, and awslabs/aws-c-cal#70 says the openssl path isn't ready in 0.5. So I'm giving it a go with `aws-lc`.

@rvandegrift Sounds good. It may be worth pinging the GitHub issue about it to see if it is addressed. I think another thing worth noting is that in the POC branch, I'm relying on an older version of AWS CRT so maybe in a newer version it is fixed (i.e., the POC is on version [0.11.13](https://pypi.org/project/awscrt/0.11.13/) but the latest version is [0.12.2](https://pypi.org/project/awscrt/0.12.2/))?

Also out of curiosity, out of the options (e.g. `system-sandbox`, `portable-exe`, PEP517 + system install) in the proposal what approach were you planning to try to package the AWS CLI v2?

proposals/source-install.md
kyleknap: > This really seems to be taking on an awful lot yourselves and introducing a lot of complexity as a result. At some point, there has to be a boundary between this project and the user's environment.

Yeah it's definitely not an insignificant amount of effort to maintain this. The end approach is actually fairly similar to the [bundled installer](https://docs.aws.amazon.com/cli/latest/userguide/install-linux.html#install-linux-bundled) for v1. So we have experience with the approach/boundary and this new approach tries to incorporate lessons learned with that installer. For example, the main differences between the bundled installer and the build process in this proposal are:

* The bundled installer is a zip containing the sdists for all of the AWS CLI v1's dependencies. This was fairly limiting because users could not control what version of a dependency they wanted to install and could not leverage already built wheels (and would have to rebuild the wheel as part of installing the bundled installer). With this build process it just reuses the Python packages already on the system or just downloads them into a virtualenv using pip (which has more flexibility in that it can better choose what artifact to download and install depending on the system).

* The bundled installer had backwards compatibility guarantees that all you need was a Python interpreter and any subsequent upgrade should continue to pass without any new software being installed on the system. This has prevented us from pulling in dependencies that require a compiler or even upgrade some dependencies. With the proposal, there is no backwards compatibility guarantee on needing additional system dependencies.

> I might suggest focusing on creating official installers for those platforms?

Yep we should or even better yet make the AWS CLI v2 installable through standard package managers for those platforms. The primary purpose for this approach is for the development team to stop being the bottleneck for all potential users that currently have no path forward to installing the AWS CLI v2 as it would take a considerable amount of time to reach full coverage for all of the platforms people want to run the AWS CLI. Ideally, in the future, less users would need to use the source distribution, but there is no timetable for when that future would take hold.

> And, the approach detailed here is still building from source and requires a healthy python install to make that work! So I don't really see much difference, I guess. Users that, before, had trouble managing python runtime isolation and installing packages into a specific python runtime will still have trouble, won't they? 

Not necessarily. The idea is that for the user that has no experience in Python nor cares about reusing existing Python dependencies (i.e., just get the AWS CLI v2 installed in the quickest way possible) they should be able to blindly just run the following commands and the build/install just works (because it is abstracting over the common issues such as installing/isolating the packages to a virtualenv):
```
./configure --with-download-deps
make
make install
```
Especially, with v1, non-Python users would just blindly run `pip install awscli` because they would not know any better and not realize there is best practices to installing Python packages.

I do think the Python community has improved in the area of tools enforcing best practices and making them more accessible (e.g., `pipx`), but the concern is these tools are still Python specific and would force non-Python users to install and learn a new tool/terminology outside of their standard package manager or `configure`/`make` workflow.

dan-petty:
This is not a universal package distribution solution like we should already have using pip. 👎

proposals/source-install.md
lorengordon: >The moment we do integrate it makes it more difficult the new project to make larger changes quickly as a large GA'd project is relying on it. So if we just put whatever was on the v2 branch published under abotocore2 namespace and immediately started relying on it, it will be impede/slow down progress on many of the broader, more invasive changes the maintainers of the Python SDK team do want to make.

I'm not sure I understand how this is true. It is already the case that you [pin to a specific botocore v2 tag](https://github.com/aws/aws-cli/blob/258dc7c50243488ba7ee0d8f8487c5bcaecc9ff1/setup.py#L27), and increment that for each release, testing as you go, right? So, already a large GA'd project is relying on the botocore v2 implementation. Are you saying the botocore project _wants_ to start v2 over again, basically? Or that you want to relax that strict pin in awscli v2?

proposals/source-install.md
lorengordon: >Not necessarily. The idea is that for the user that has no experience in Python nor cares about reusing existing Python dependencies (i.e., just get the AWS CLI v2 installed in the quickest way possible) they should be able to blindly just run the following commands and the build/install just works (because it is abstracting over the common issues such as installing/isolating the packages to a virtualenv)

Well ok. It seems like a nice idea. But isn't it just trading expertise with python for expertise with make? And probably particular versions of make and the <shell> used (MacOS problems). And some number of other system packages, and the inscrutable errors from make when an assumption in the script/Makefile gets violated. 

There are totally valid use cases where users need to build from source, and providing utilities like this does make it easier (which is awesome!). But it doesn't really seem like this is making anything any easier for the type of user that is going to be having problems managing their system and Python runtimes. This is still a very advanced type of install requiring specialized knowledge and tools. These are orthogonal use cases.

>these tools are still Python specific and would force non-Python users to install and learn a new tool/terminology

When there is no official installer for their platform. Or when their environment restrictions are such that they are prevented from using an official installer (e.g. non-admin user on Windows with awscli v2).

But even then, it feels much lower effort to lean into the packaging tools and practices and expectations for the programming language, and to actively support that as a lowest common denominator (see above, trading one bit of expertise for another). Publishing to PyPi is trivially easy, and addresses a [large swath of the user and developer community](https://pypistats.org/packages/awscli)(~100mil/month!) . Why put so much thought and effort into making things harder for those users?

lorengordon: Oh wow, over the last month, botocore and s3transfer are in the top 5, and awscli v1 is #10! People really love these on pypi! https://pypistats.org/top

rvandegrift:
> > As I've worked through more of the C SDK dependencies, I've run into some which don't actually support this. In particular, the docs for `aws-c-cal` say that aws-lc is required, and [awslabs/aws-c-cal#70](https://github.com/awslabs/aws-c-cal/issues/70) says the openssl path isn't ready in 0.5. So I'm giving it a go with `aws-lc`.
> 
> @rvandegrift Sounds good. It may be worth pinging the GitHub issue about it to see if it is addressed. I think another thing worth noting is that in the POC branch, I'm relying on an older version of AWS CRT so maybe in a newer version it is fixed (i.e., the POC is on version [0.11.13](https://pypi.org/project/awscrt/0.11.13/) but the latest version is [0.12.2](https://pypi.org/project/awscrt/0.12.2/))?

@kyleknap thanks, I'll keep this in mind.  Once I didn't need to figure out how to handle litani & the cmbc proof repos, it's been fairly straightforward to get `aws-lc` built.  Currently I'm working on `aws-crt-python`, hoping to get it far enough to test my CRT packages, but I haven't had much time this week.

> Also out of curiosity, out of the options (e.g. `system-sandbox`, `portable-exe`, PEP517 + system install) in the proposal what approach were you planning to try to package the AWS CLI v2?

PEP517 + system install.  The other options don't really make sense for distros, and probably couldn't be made compliant with Debian's packaging policy.

proposals/source-install.md
benkehoe: I think this is actually evidence the AWS CLI has a significantly wider userbase than Python users.

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+  PYTHON      the Python interpreter
+
+```
+The sections below describes the most pertinent options. 
+
+
+##### Install location
+
+Source installation of the AWS CLI v2 uses two configurable directories to
+install the AWS CLI v2:
```
skyzyx, line 192: I'd like to request the addition of support for `make install DESTDIR=/path/to/dest`. ([source](https://www.gnu.org/software/make/manual/html_node/DESTDIR.html#DESTDIR))

In short, it prepends `$DESTDIR` to the `--prefix` option (and, in turn, `--libdir` and `--bindir`) exclusively for the purpose of staging files in preparation for packaging. The difference between `$DESTDIR` and simply updating `$PREFIX` is subtle, but important. I believe that this can meaningfully help with the packaging process.

manics:
@kyleknap I increased the upper version bounds for:
- colorama: `<0.4.4` ➡️  `<0.5.0`
- docutils: `<0.16` ➡️  `<0.18`
- cryptography: `<3.4.0` ➡️  `<3.5.0`
- ruamel.yaml: `0.16.0` ➡️  `<0.17.0`

philippsimon:
Does building the AWS Cli v2 for Alpine according to this proposal will still enforce TLS 1.2+ as it's written in the AWS documentation? See https://docs.aws.amazon.com/cli/latest/userguide/cli-security-enforcing-tls.html

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+  [Alpine Linux](https://github.com/aws/aws-cli/issues/4685).
+  
+* The environment restricts access to resources only needed by the
+  pre-built executable. For example, on security hardened systems, it does not
+  give permission to shared memory
+  [which is needed](https://github.com/aws/aws-cli/issues/5769)
+  during the bootloading process of the frozen `aws` executable.
+
+* Users want to be able to install the AWS CLI using the package manager
+  standard to their environment (e.g., `brew`, `yum`, `apt`) instead of having
```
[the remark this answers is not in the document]
kyleknap: Chocolatey is already using our MSIs and stays up to date on `v2`: https://community.chocolatey.org/packages/awscli. I'm not planning to add it to this list.

proposals/source-install.md
```diff
@@ -0,0 +1,1133 @@
...
+These are the requirements to build the AWS CLI v2 from source:
+
+* An environment that can run Autotools generated files (e.g.,
+  `configure`, `Makefile`). These files are widely portable across POSIX
+  platforms, and for systems that do not come with POSIX-compliant shells like
+  Windows, there is additional software available (e.g., MSYS2) that allow users
+  to run Autotools generated files. For more information, see the
+  [Appendix on Windows usage](#windows).
+
+* Python 3.8+ interpreter
```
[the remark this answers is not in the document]
kyleknap: Our expected version support is going to follow suit with what it is planned for v1, boto3/botocore: https://aws.amazon.com/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/. Basically, we will support the Python runtime up to 6 months after the PSF ends support. I'll update this bullet.

stealthycoin (approved):

kyleknap:
I moved the proposal to accepted and merged the PR. For those following the PR, the accepted state indicates that the initial proposal has been approved and progress will be started toward releasing an official implementation. As progress is made on the official implementation, amendments may be added to the accepted proposal.

manics:
@kyleknap That's great news! Is there a single GitHub issue we can subscribe to track the implementation?

kyleknap:
@manics Yep! I went ahead and created a GitHub issue to track progress: https://github.com/aws/aws-cli/issues/6785. Feel free to watch that issue for status updates.

PR #6379, opened 2021-09-02
Update regex for MRAP ARNs

kyleknap:
For MRAP arns there is no region field so we needed to relax the regex to allow for empty region values in the arn.


nateprewitt (approved):
Looks good :shipit:

PR #6413, opened 2021-09-21
Migrate to separate development dependency management

nateprewitt:
This PR will introduce a few changes that are cascading requirements of the original goal. As part of building a stable test space for the boto/cli projects, we're migrating to a consistent set of scripts that will manage all parts of the process. One requirement we now have is the ability to separate installation of testing dependencies from installing the github branches of our dependencies.

This patch will:
* Allow us to install development dependencies separately
* Move us to a static lock file for testing dependencies
* Move testing scripts to a context manager for directory changes
* Move us off the Python 2 `mock` import syntax that required an additional dependency


joguSD (approved):

tests/unit/customizations/test_waiters.py
```diff
@@ -10,13 +10,13 @@
...
 # language governing permissions and limitations under the License.
-import mock
-
 from botocore.waiter import WaiterModel
 from botocore.exceptions import DataNotFoundError
 
-from awscli.testutils import unittest, BaseAWSHelpOutputTest, \
-    BaseAWSCommandParamsTest
+from awscli.testutils import (
+        mock, unittest,
```
zdutta, line 17: nit: We should get this line aligned with the line below. It also looks like there is extra white space at the EOL.

tests/unit/customizations/test_waiters.py
```diff
@@ -10,13 +10,13 @@
...
-
 from botocore.waiter import WaiterModel
 from botocore.exceptions import DataNotFoundError
 
-from awscli.testutils import unittest, BaseAWSHelpOutputTest, \
-    BaseAWSCommandParamsTest
+from awscli.testutils import (
+        mock, unittest, 
+     BaseAWSHelpOutputTest,BaseAWSCommandParamsTest
+     )
```
zdutta, line 19: nit: There is extra white space at the EOL.

zdutta (changes requested):

zdutta (approved):

PR #6423, opened 2021-09-24
[v2] Convert project to PEP517-based setuptools layout

kyleknap:
This is one step in implementing this proposal: https://github.com/aws/aws-cli/pull/6352 in that it will allow us to leverage PEP 517 to build sdists and wheels. Note the build process still does not automatically build the auto-complete index. See commit message for more details and rationale for each change.


nateprewitt (approved):
🤌💋

kdaily:
This should close out https://github.com/aws/aws-cli/pull/5987 too for removing the universal wheel flag.

PR #6438, opened 2021-09-29
[v2] Migrate to pytest

kyleknap:
This corresponds to this PR for the develop branch: https://github.com/aws/aws-cli/pull/6427. Note that v2 already was relying on `pytest` for the `integration` suite so minimal work was needed there.

With regards to comparing tests between `nosetests` and `pytest`, these were the numbers:

| Test directory | `nosetests` results | `pytest` results | Net difference |
| --- | --- | --- | --- |
| `unit/` | 3275 total with 2 skip | 3273 total with 1 skip | -2 total, -1 skip with `pytest` |
| `functional/` | 25119 total with 1 skip | 10926 total with 1 skip | -14193 total with `pytest` |

The specific break down for the net difference is as follows:

* `/tests/unit/test_clidriver.py` - Removed test that had a decorated `unittest.skip` . CLI v2 did not support that functionality anymore (**Net: -1 total and -1 skip**)
* `/tests/unit/customizations/s3/test_utils.py` - `test_human_readable_size` had an extra `1.0 MiB` case that was removed in the port (**Net: -1 total**)

* `/tests/functional/wizards/` - The tests in this directory were not being ran before because it was missing a `__init__.py`. They are now being ran with `pytest`. (**Net: +10 total**)
* `tests/functional/docs/test_examples.py` - Same calculation as in [original PR](https://github.com/aws/aws-cli/pull/6427) but v2 has one less RST example (**Net: -14203 total**)

Based on that break down, everything should line up.

In addition, I pulled in test script updates from this PR to be consistent between the branches: https://github.com/aws/aws-cli/pull/6413

For any additional details, see the commit messages. I tried to add a good amount of context there.

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -41,7 +41,8 @@ class WizardAppRunner(object):
 
 class WizardApp(Application):
     def __init__(self, layout, values, traverser, executor, style=None,
-                 key_bindings=None, full_screen=True, output=None):
+                 key_bindings=None, full_screen=True, output=None,
+                 app_input=None):
         self.values = values
         self.traverser = traverser
         self.executor = executor
@@ -53,7 +54,7 @@ class WizardApp(Application):
         self.error_bar_visible = None
         super().__init__(
             layout=layout, style=style, key_bindings=key_bindings,
-            full_screen=full_screen, output=output
+            full_screen=full_screen, output=output, input=app_input,
         )
 
     def run(self, pre_run=None, **kwargs):
--- a/tests/functional/autoprompt/test_prompttoolkit.py
+++ b/tests/functional/autoprompt/test_prompttoolkit.py
@@ -30,7 +30,7 @@ from awscli.autoprompt.prompttoolkit import (
 from awscli.autoprompt.history import HistoryDriver
 from awscli.testutils import unittest, mock, FileCreator, cd
 from tests import PromptToolkitApplicationStubber as ApplicationStubber
-from tests import FakeApplicationOutput
+from tests import FakeApplicationOutput, FakeApplicationInput
 
 
 def _ec2_only_command_table(command_table, **kwargs):
@@ -106,7 +106,8 @@ class BasicPromptToolkitTest(unittest.TestCase):
             driver=self.driver,
             factory=self.factory,
             cli_parser=self.cli_parser,
-            output=FakeApplicationOutput()
+            output=FakeApplicationOutput(),
+            app_input=FakeApplicationInput(),
         )
         self.prompter.args = []
         self.prompter.input_buffer = self.factory.create_input_buffer()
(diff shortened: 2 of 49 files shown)


awscli/customizations/wizard/app.py
```diff
@@ -41,7 +41,8 @@ def run(self, loaded):
 
 class WizardApp(Application):
     def __init__(self, layout, values, traverser, executor, style=None,
-                 key_bindings=None, full_screen=True, output=None):
+                 key_bindings=None, full_screen=True, output=None,
+                 app_input=None):
```
nateprewitt, line 45: What were these `app_input` additions for?

tests/functional/autoprompt/test_prompttoolkit.py
```diff
@@ -106,7 +106,8 @@ def setUp(self):
             driver=self.driver,
             factory=self.factory,
             cli_parser=self.cli_parser,
-            output=FakeApplicationOutput()
+            output=FakeApplicationOutput(),
+            app_input=FakeApplicationInput(),
```
nateprewitt, line 110: Not recognizing this change either, is this testing infra specific to v2?

tests/unit/test_errorhandler.py
```diff
@@ -43,13 +44,14 @@ def _assert_rc_and_error_message(case, error_handler):
         raise case.exception
     except BaseException as e:
         cr = error_handler.handle_exception(e, stdout, stderr)
-        nose.tools.eq_(cr, case.rc, case.exception.__class__)
-        nose.tools.assert_in(case.stderr, stderr.getvalue())
-        nose.tools.eq_(case.stdout, stdout.getvalue())
+        assert cr == case.rc, case.exception.__class__
+        assert case.stderr in stderr.getvalue()
+        assert case.stdout in stdout.getvalue()
```
nateprewitt, line 49: It looks like this is a v2 specific change. We moved the last assertion from `equals` to `in`. That actually looks like the correct fix, but wanted to double check this was intentional.

nateprewitt (commented):
Some minor comments, but fine approving once those are answered. It looks like they're all around v2 specific changes.

awscli/customizations/wizard/app.py
kyleknap: I had the rationale in the second commit message:
```
Add the option to prompt toolkit abstractions to inject an input. This was needed because the stdin
provided by pytest does not  have a fileno and as a result causes prompt toolkit to fail when instantiating
its default wrapper around stdin.
```
Basically we cannot instantiate a normal prompt toolkit `Application` class (which is the main class for prompt toolkit usage) when running tests so we need to inject a fake input in order to be able instantiate it.

tests/functional/autoprompt/test_prompttoolkit.py
kyleknap: Yeah these are v2 changes. I put all of the v2 changes in a separate commit (the second one).

tests/unit/test_errorhandler.py
kyleknap: Good catch. This should be equals.

nateprewitt (approved):
Cool, looks good 🚢

Issue #6424, opened 2021-09-25
aws ecr batch-delete-image returns failure but 0 return code


matteoolivi:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
I have an empty and private ECR repo, if I run `aws ecr batch-delete-image` to remove the image tag `latest`, the command output returns an `ImageNotFound` failure (since the repo is empty), but `echo $?` returns `0`

**SDK version number**
```shell
$ aws --version
aws-cli/2.2.32 Python/3.8.8 Darwin/20.6.0 exe/x86_64 prompt/off
```

**Platform/OS/Hardware/Device**
```shell
$ uname -a
Darwin <name of my laptop> 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:31 PDT 2021; root:xnu-7195.141.2~5/RELEASE_X86_64 x86_64
```

**To Reproduce (observed behavior)**
Create a private, empty ECR repository, and configure the aws cli with a profile that can pull/push that ECR repo. Then:

```shell
$ aws ecr batch-delete-image --repository-name <repo name> --image-ids imageTag="latest" --no-cli-pager
{
    "imageIds": [],
    "failures": [
        {
            "imageId": {
                "imageTag": "latest"
            },
            "failureCode": "ImageNotFound",
            "failureReason": "Requested image not found"
        }
    ]
}

$ echo $?
0
```

**Expected behavior**
Create a private, empty ECR repository, and configure the aws cli with a profile that can pull/push that ECR repo. Then:

```shell
$ aws ecr batch-delete-image --repository-name <repo name> --image-ids imageTag="latest" --no-cli-pager
{
    "imageIds": [],
    "failures": [
        {
            "imageId": {
                "imageTag": "latest"
            },
            "failureCode": "ImageNotFound",
            "failureReason": "Requested image not found"
        }
    ]
}

$ echo $?
<non zero code> (probably 255 by looking at https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-returncodes.html)
```

**Logs/output**
```shell
[...]

PR #6444, opened 2021-10-01
Update return code topic page

stobrien89:
Addresses #6424

Description of changes:
There are some services that respond with a 200 when an an operation is what some would intuitively consider unsuccessful. S3 does this in certain cases, but see the issue for a concrete example. Updating the topic page to clarify that the CLI considers a response code of 200 an indicator of a successful request/operation.

Also removed usage of the "error/errors thrown" phrase from the page. This seems somewhat language and SDK-specific, so that was simplified to reduce any potential confusion.

### DIFF
--- a/awscli/topics/return-codes.rst
+++ b/awscli/topics/return-codes.rst
@@ -6,8 +6,8 @@
 These are the following return codes returned at the end of execution
 of a CLI command:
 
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):
Thanks, @stobrien89!

PR #6449, opened 2021-10-05
Suppress unnecessary error output in aws.cmd

c22:
*Description of changes:*

The call to "assoc .py" produces unnecessary stderr output every time aws.cmd is invoked if there are no system-level associations with .py files, eg.

`File association not found for extension .py`

We can suppress the error output by redirecting the stderr output to the NUL device (similar to what you would do in POSIX environments).

This has no effect on the "happy case" where file associations do exist.

*Submitter notes:*

It might be worth pointing out that relying on the file association to find the Python interpreter is probably not a good idea. Users may have `.py` extensions set to open with their code editor of choice, but this doesn't mean it should be used as the Python interpreter.

The `aws.cmd` script itself is quite a kludge and it doesn't support other use-cases such as running under pipx (see: pypa/pipx#719).

This small QoL change should not be too controversial, but I would be willing to make further improvements to the aws.cmd script if they would be welcome.


aaoswal:
Hi @c22,
Thanks for bringing this to our attention. We're making some changes to our contribution process, which if you're interested you can read more about [here](https://github.com/kdaily/aws-cli/blob/contrib-guide-proposal/proposals/contribution-guide.md).

I think some more investigation needs to be made before we would accept this change, so we're going to close this pull request for now and open up an issue for our team to investigate, reproduce, and discuss a solution. We'll link the issue we create here for future reference.
#6853 

 Once we've determined what the correct solution is, we can open this up for a contribution that we could help you (or another community member) with.
 
Thanks!

PR #6474, opened 2021-10-13
[Fix issue #6012] Prevent sending HEAD request with sse params

Da-1kun:
*Issue #, if available:*
[aws s3 cp bucket to bucket copy fails when sse-c is set #6012](https://github.com/aws/aws-cli/issues/6012)

*Description of changes:*
I fixed the codes not to send HEAD request with sse params.

Test for this change
https://github.com/aws/aws-cli/blob/47ccc20ae74fce359adb7a661aacf033b9d85cbb/tests/functional/s3/test_cp_command.py#L440

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -352,6 +352,8 @@ class FileGenerator(object):
         try:
             params = {'Bucket': bucket, 'Key': key}
             params.update(self.request_parameters.get('HeadObject', {}))
+            del params['SSECustomerAlgorithm']
+            del params['SSECustomerKey']
             response = self._client.head_object(**params)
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @Da-1kun,

Sorry for the delay in reviewing this.  I confirmed with the team that we won't be able to accept this change, as there are scenarios where we do actually need to provide these headers in a head request (if the object is already encrypted).  One of the possible solutions to this would be creating a new flag which ensures these headers are not added to the request, or modifying an existing flag (like `--sse-c-copy-source`) to accept an option of `none` or something to that effect, to prevent the headers from being sent in this scenario.  This will likely need to be done by the development team, but you're welcome to give it a shot if you'd like!

Issue #6308, opened 2021-07-30
CLI versions 1.20.9 and 2.2.24 shipped a breaking change in a dot release


dylanlingelbach:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
Versions 1.20.9 and 2.2.24 shipped a breaking change to EKS support by changing the supported k8s API value from `v1alpha1` to `v1beta1`. This breaks any existing kubeconfigs that have `v1alpha1` specified in their kubeconfig.

The [PR that changed this behavior](https://github.com/aws/aws-cli/pull/6289) clearly specified this is a breaking change, however the changed shipped in a point release, with no clear communication about it being a breaking change.

The error message you see when attempting to use a kubeconfig with an incorrect version is:
```
unable to use kubectl with the EKS cluster (check 'kubectl version'): Unable to connect to the server: getting credentials: exec plugin is configured to use API version client.authentication.k8s.io/v1alpha1, plugin returned version client.authentication.k8s.io/v1beta1
```

**SDK version number**
1.20.9
2.2.24

**Platform/OS/Hardware/Device**
What are you running the cli on?

Linux/Mac OS X

**To Reproduce (observed behavior)**
1. Have a kubeconfig that uses `eks get-token` for authentication
2. Upgrade CLI to 1.20.9 or 2.2.24
3. Run a `kubectl` command

**Expected behavior**
Point releases of CLIs to not include breaking changes

**Logs/output**
```
unable to use kubectl with the EKS cluster (check 'kubectl version'): Unable to connect to the server: getting credentials: exec plugin is configured to use API version client.authentication.k8s.io/v1alpha1, plugin returned version client.authentication.k8s.io/v1beta1
```

**Additional context**
Related to https://github.com/aws-samples/eks-workshop/issues/1232

kdaily:
Hi @dylanlingelbach,

Thanks for opening this issue. This change breaks `kubectl` when upgrading the AWS CLI but not running `aws eks update-kubeconfig`, and this was unintentional.

[...]

PR #6309, opened 2021-07-30
Revert "[customizations/eks]: Update credential API version"

nateprewitt:
Reverts aws/aws-cli#6289 in response to #6308. This will migrate back to the previous usage of the v1alpha1 endpoint

PR #6289, opened 2021-07-21
[customizations/eks]: Update credential API version

micahhausler:
Kubernetes has deprecated v1alpha1, v1beta1 has been available since Kubernetes
v1.11 (kubernetes/kubernetes#64482), and EKS currently supports Kubernetes
versions v1.16 through v1.21. This is a breaking change for clients running
versions v1.10 and older, which haven't been supported by EKS since September
2019.

Signed-off-by: Micah Hausler <mhausler@amazon.com>

*Issue #, if available:*
N/A

*Description of changes:*

Update Kubernetes client authentication API version.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #6476, opened 2021-10-13
Deprecate Kubernetes client API version v1alpha1

micahhausler:
Kubernetes has deprecated v1alpha1, v1beta1 has been available since Kubernetes
v1.11 (kubernetes/kubernetes#64482), and EKS currently supports Kubernetes
versions v1.16 through v1.21. This is a breaking change for clients running
versions v1.10 and older, which haven't been supported by EKS since September
2019.

"aws eks get-token" now respects the KUBERNETES_EXEC_INFO environment
variable and conservatively falls back to v1alpha1, which is supported
by Kubernetes versions 1.10 through 1.22 (released upstream August 2021, to be
released by EKS in Q4 2021). It also now supports clients requesting credentials 
using the "v1beta1" and "v1" versions.

"aws eks update-kubeconfig" now writes "v1beta1" in the kubeconfig which
will be supported by Kubernetes until 1.29 (aproximately December 2023).
At or around that date, we can change the default version written to
kubeconfigs to "v1"

Signed-off-by: Micah Hausler <mhausler@amazon.com>

*Issue #, if available:*
- #6289 
- #6308
- #6309

*Description of changes:*

See above

cc: @nckturner @sgundapu 


stealthycoin:
fixes #6308

tests/functional/eks/test_get_token.py
```diff
@@ -143,3 +144,105 @@ def test_url_different_partition(self):
             expected_endpoint='sts.cn-north-1.amazonaws.com.cn',
             expected_signing_region='cn-north-1'
         )
+
+    def test_api_version_discovery_deprecated(self):
+        os.environ["KUBERNETES_EXEC_INFO"] = '{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{"interactive":true}}'
```
kyleknap, line 149: Instead of mutating the global environment variables and cleaning them up after the assertions for each case, this test class already patches the `os.environ`: https://github.com/aws/aws-cli/blob/7a1da123217518faa1d850f23d0f819db2a04c84/awscli/testutils.py#L323-L332

So for all of these tests, we should be able to set the environment variable as follows:
```python
self.environ['KUBERNETES_EXEC_INFO'] = ...  # JSON document
```
This allows us to set the environment variables without risk of mutating the global state and also removes the need to delete it from the global env vars at the end.

tests/functional/eks/test_get_token.py
```diff
@@ -143,3 +144,105 @@ def test_url_different_partition(self):
...
+           "client.authentication.k8s.io/v1alpha1",
+        )
+
+        self.assertEqual(
+            stderr,
+           ("Empty KUBERNETES_EXEC_INFO, defaulting to client.authentication.k8s.io/v1alpha1. "
+            "This is likely a bug in your Kubernetes client. Please update your Kubernetes client.\n")
+        )
+    def test_api_version_discovery_v1(self):
+        os.environ["KUBERNETES_EXEC_INFO"] = '{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1","spec":{"interactive":true}}'
```
kyleknap, line 202: It would be great if we had a helper function that sets the exec info based on the api version instead of having to duplicate the JSON document for each version. Specifically something like:
```python
def set_kubernetes_exec_info(self, api_version):
     # Set KUBERNETES_EXEC_INFO env var with default payload and provided api version
```
This will also make it easier to do bulk updates of the document used if needed in the future.

kyleknap (commented):
Looks fine. All of my feedback was related to Python-specifics and codebase usage/organization.

awscli/customizations/eks/get_token.py
```diff
@@ -13,6 +13,8 @@
 import base64
```
kyleknap, line 13: It would also be great if we can get a changelog for this feature? You can use the [`scripts/new-change`](https://github.com/aws/aws-cli/blob/develop/scripts/new-change) script to generate a changelog entry. For the type field, I'd recommend setting the type to `enhancement` and the category to `eks get-token`

tests/functional/eks/test_get_token.py
```diff
@@ -143,3 +144,105 @@ def test_url_different_partition(self):
...
+        self.assertEqual(
+            stderr,
+           ("Kubeconfig user entry is using deprecated API "
+            "version client.authentication.k8s.io/v1alpha1. Run 'aws eks update-kubeconfig' to update\n")
+        )
+
+        del os.environ["KUBERNETES_EXEC_INFO"]
+
+    def test_api_version_discovery_malformed(self):
+        os.environ["KUBERNETES_EXEC_INFO"] = '{{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{"interactive":true}}'
```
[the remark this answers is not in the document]
kyleknap: We've used this in the past for setting env vars: https://github.com/aws/aws-cli/pull/6476#discussion_r866181788

awscli/customizations/eks/get_token.py
```diff
@@ -86,6 +88,72 @@ def _run_main(self, parsed_args, parsed_globals):
...
+        exec_info_raw = os.environ.get("KUBERNETES_EXEC_INFO", "")
+        if len(exec_info_raw) == 0:
+            # All kube clients should be setting this, we'll return the fallback and write an error
+            sys.stderr.write(error_msg_tpl.format(error_prefixes["empty"], fallback_api_version))
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+            return fallback_api_version
+        try:
+            exec_info = json.loads(exec_info_raw)
+        except json.JSONDecodeError as e:
```
nateprewitt, line 135: We aren't using the error so allocating it `as e` shouldn't be necessary.

tests/functional/eks/test_get_token.py
nateprewitt: We should consider just making it a context manager to do something like this. That also handles cleanup for us instead of using `del` everywhere.

```python
@contextlib.contextmanager
def set_environ(name, value):
    if value is not None:
        original_value = os.environ.get(name)
        os.environ[name] = value
    try:
        yield
    finally:
        if value is not None:
            if original_value is None:
                del os.environ[name]
            else:
                os.environ[name] = original_value
```

awscli/customizations/eks/get_token.py
```diff
@@ -86,6 +88,72 @@ def _run_main(self, parsed_args, parsed_globals):
         uni_print('\n')
         return 0
 
+    def discover_api_version(self):
+        """
+        Parses the KUBERNETES_EXEC_INFO environment variable and returns the API
+        version. If the environment variable is empty, malformed, or invalid,
+        return the v1alpha1 response and print an message to stderr.
```
nateprewitt, line 95:
```suggestion
        return the v1alpha1 response and print a message to stderr.
```

awscli/customizations/eks/get_token.py
```diff
@@ -86,6 +88,72 @@ def _run_main(self, parsed_args, parsed_globals):
...
+        If the v1alpha1 API is specified explicitly, a message is printed to
+        stderr with instructions to update.
+
+        :return: The client authentication API version
+        :rtype: string
+        """
+        alpha_api = "client.authentication.k8s.io/v1alpha1"
+        beta_api = "client.authentication.k8s.io/v1beta1"
+        v1_api = "client.authentication.k8s.io/v1"
+        # At the time Kubernetes v1.29 is released upstream (aprox Dec 2023),
```
nateprewitt, line 106:
```suggestion
        # At the time Kubernetes v1.29 is released upstream (approx Dec 2023),
```

.changes/next-release/enhancement-eksgettoken-91863.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "eks get-token",
+  "description": "Add support to respect env var KUBERNETES_EXEC_INFO in eks get-token"
```
kyleknap, line 4: For the description, let's update it to include API version somewhere in it like:

> Add support for respecting API version found in KUBERNETES_EXEC_INFO environment variable

.changes/next-release/enhancement-eksupdatekubeconfig-63017.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "eks update-kubeconfig",
+  "description": "Update default API version for eks update-kubeconfig"
```
kyleknap, line 4: We should also note what the default API version is so something like:
> Update default API version to v1beta1

tests/functional/eks/test_get_token.py
```diff
@@ -57,12 +62,24 @@ def assert_url_correct(self, response,
...
 
+    def set_kubernetes_exec_info(self, api_version):
+        # Set KUBERNETES_EXEC_INFO env var with default payload and provided
+        # api version
+        kubernetes_exec_info_tpl = (
+            '{{"kind":"ExecCredential",'
+            '"apiVersion":"client.authentication.k8s.io/{0}",'
+            '"spec":{{"interactive":true}}}}'
+        )
+        # kubernetes_exec_info_tpl = '{{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/{}","spec":{{"interactive":true}}}}'
```
kyleknap, line 79: Looks like a left over comment that we can remove.

tests/functional/eks/test_get_token.py
```diff
@@ -57,12 +62,24 @@ def assert_url_correct(self, response,
...
         if missing_padding:
             b64_token += b'=' * (4 - missing_padding)
         return base64.urlsafe_b64decode(b64_token).decode()
 
+    def set_kubernetes_exec_info(self, api_version):
+        # Set KUBERNETES_EXEC_INFO env var with default payload and provided
+        # api version
+        kubernetes_exec_info_tpl = (
+            '{{"kind":"ExecCredential",'
+            '"apiVersion":"client.authentication.k8s.io/{0}",'
```
kyleknap, line 76: Can we use an f-string here?

awscli/customizations/eks/get_token.py
```diff
@@ -90,80 +110,93 @@ def _run_main(self, parsed_args, parsed_globals):
...
 
-        error_msg_tpl = ("{0} KUBERNETES_EXEC_INFO, defaulting "
-                    "to {1}. This is likely a bug in your Kubernetes "
-                    "client. Please update your Kubernetes client.")
-        unrecognized_msg = ("Unrecognized API version in KUBERNETES_EXEC_INFO, defaulting "
-                    "to {0}. This is likely due to an outdated AWS CLI."
-                    " Please update your AWS CLI.".format(fallback_api_version))
-        deprecation_msg_tpl = ("Kubeconfig user entry is using deprecated API "
-                    "version {0}. Run 'aws eks update-kubeconfig' to update")
+        error_msg_tpl = (
```
kyleknap, line 134: It would be great if we can move the templates up as module level constants as well to make it easier to find the wording and reduce the size of the core logic for this function.

kyleknap (commented):
Looks good. Just had some more suggestions.

kyleknap (approved):
Looks good! :ship:

nckturner:
LGTM

awscli/customizations/eks/get_token.py
```diff
@@ -26,6 +28,33 @@
...
+FULLY_SUPPORTED_API_VERSIONS = [
+    V1_API,
+    BETA_API,
+]
+DEPRECATED_API_VERSIONS = [
+    ALPHA_API,
+]
+
+ERROR_MSG_TPL = (
+    "{0} KUBERNETES_EXEC_INFO, defaulting to {1}. This is likely a "
```
nckturner, line 44: Should it say "Error reading KUBERNETES_EXEC_INFO" instead of just "KUBERNETES_EXEC_INFO"

kyleknap: It will say `Error parsing` to start the sentence as the variable is being used as a template. Tests also confirm that too: https://github.com/aws/aws-cli/pull/6476/files#diff-ee0c03e30582b485e61dba3f70f956a6024118bdffe8657c476fee7f9ac655afR194.

tests/functional/eks/test_get_token.py
micahhausler: The context manager seems at odds with the `self.environ` usage, I'll go for the helper function

micahhausler:
@zawachte-c thank you! Take a look at #6940, as this should resolve the <= 1.19 problems

PR #6482, opened 2021-10-14
Improve incorrect access policy in s3api example

bdavs3:
*Description of changes:* Fix the "Resource" key in an SNS Access Policy so that it correctly reflects an SNS topic ARN, not an S3 bucket.

To confirm that this is correct, see the relevant [AWS docs](https://docs.aws.amazon.com/sns/latest/dg/sns-access-policy-use-cases.html#sns-grant-aws-account-access-to-topic).

### DIFF
--- a/awscli/examples/s3api/put-bucket-notification-configuration.rst
+++ b/awscli/examples/s3api/put-bucket-notification-configuration.rst
@@ -30,7 +30,7 @@ The SNS topic must have an IAM policy attached to it that allows Amazon S3 to pu
      "Action": [
       "SNS:Publish"
      ],
-     "Resource": "arn:aws:sns:us-west-2:123456789012:my-bucket",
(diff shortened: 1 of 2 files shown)


kdaily:
Thanks @bdavs3!

In this case, wouldn't the SNS topic need to be the same as in the example command (`arn:aws:sns:us-west-2:123456789012:s3-notification-topic`)?

kdaily (changes requested):
Thanks @bdavs3!

In this case, wouldn't the SNS topic need to be the same as in the example command (`arn:aws:sns:us-west-2:123456789012:s3-notification-topic`)?

bdavs3:
@kdaily Yep, I agree. I amended my change.

justindho:
Hi @bdavs3, thanks for pointing out this issue initially. It looks like this was fixed in #6866, which was merged a few weeks ago. As a result, I will be closing this PR.

PR #6494, opened 2021-10-21
[v2] Absorb implementations of botocore and s3transfer

kyleknap:
Motivated by this [proposal](https://github.com/aws/aws-cli/pull/6352), this pull request pulls in AWS CLI v2 implementation of botocore and the s3transfer library into the code base. See the [rationale section of the proposal](https://github.com/kyleknap/aws-cli/blob/source-proposal/proposals/source-install.md#q-why-are-copies-of-botocore-and-s3transfer-being-maintained-in-the-aws-cli-source) for the advantages in doing this.

These packages are now directly located underneath the main `awscli` package (e.g. `awscli.botocore` and `awscli.s3transfer`). I also added a [`MetaPathFinder`](https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder) so that you can still import `botocore` and `s3transfer` as if they were part of the top level namespace if the `awscli` package is ever imported:
```python
import awscli
import botocore  # Finds code under awscli.botocore
import s3transfer  # Finds code under awscli.s3transfer
```
This is advantageous because:
1. Reduces the amount of churn in changing import statements. This also makes it easier to import/backport future changes if logic is shareable between our implementation and the official version available through PyPI.
2. It preserves import statements for anyone using [legacy plugins](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/config-vars.html#plugins).

For the tests, they were split so that each library's test suite was added to the respective top-level test suite of the codebase under the library name. For example, tests for botocore are split up into: `tests/unit/botocore`, `tests/functional/botocore`, and `tests/integration/botocore`).

## Reviewing
I'd recommend reviewing this commit by commit. For the two large commits where we pull in the two projects, I have validation below to make sure I did it correctly:

### Source packages
Because of the aliases introduced with the meta path finder, the snapshots of the core package should be the same. This was validated using `diff` against the checked out repos/branches:
```bash
$ diff -r botocore/botocore/ aws-cli/awscli/botocore/  # comparing this PR with botocore v2 branch
$ diff -r s3transfer/s3transfer/ aws-cli/awscli/s3transfer/  # comparing this PR with s3transfer develop branch
Only in s3transfer/s3transfer/: processpool.py
```
**Note:** The `s3transfer/processpool.py` module is not used by the CLI and was causing issues in the tests where it was not able to find the aliased modules with the custom finder. Instead of trying to maintain support for it, I decided to remove it because it was not used at all for the purposes of v2.

### Test audit

To make sure I pulled in all of the tests, I compared the following the number of unit/functional and integration tests ran from this PR to the number of tests ran when the projects were separate:

| Project | Number of `unit`/`functional` tests | Number of `integration` tests  |
| --- | --- | --- |
| This PR | 40832 passed, 4 skipped | 31588 passed, 10 skipped |
| AWS CLI `v2` branch | 14239 passed, 2 skipped | 21030 passed, 6 skipped |
| botocore `v2` branch | 26031 passed, 2 skipped | 10504 passed, 4 skipped |
| s3transfer `develop` branch | 617 passed | 58 passed |
| **Net difference** [e.g. This PR - (AWS CLI `v2` branch + botocore `v2` branch + s3transfer)] | **-55 passed, 0 skipped** | **-4 passed, 0 skipped**

**Analysis for `unit`/`functional` test difference**
* This PR removed the `processpool.py` module and its tests impacting the net number of tests by: **-64**
* This PR added for tests for new meta path finder: **+9**

This difference adds up to a net difference of: **-55**

**Analysis for `integration` test difference**
* This PR removed the `processpool.py` module and its tests impacting the net number of tests by: **-4**

### DIFF
--- a/.changes/next-release/enhancement-Source-90857.json
+++ b/.changes/next-release/enhancement-Source-90857.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Source",
+  "description": "Absorb implementations of botocore and s3transfer libraries to improve building of CLI codebase (`#6494 <https://github.com/aws/aws-cli/pull/6494>`__)."
+}


glaubitz:
I don't understand the rationale for this. This is against the policies of most Linux distributions which want to avoid bundling as it creates duplicate copies of libraries that have to be maintained.

I just wanted to switch `aws-cli` in openSUSE/SLE from the v1 series to v2 only to run into this particular change which undermines the maintenance policies that we have in openSUSE/SLE.

kyleknap:
@glaubitz The rationale for this decision is here: https://github.com/kyleknap/aws-cli/blob/source-proposal/proposals/source-install.md#q-why-are-copies-of-botocore-and-s3transfer-being-maintained-in-the-aws-cli-source. There was also quite a long discussion around it in this conversation for the proposal: https://github.com/aws/aws-cli/pull/6352#discussion_r706186719

Specifically, the AWS CLI v2 does not rely on the publicly released `botocore` library and instead it relied on an unreleased, AWS CLI-maintained fork of `botocore` consisting of backwards incompatible/fundamental changes that are required for AWS CLI v2.

Bundling of the fork better reflects that the fork is implementation detail of AWS CLI v2 and is not intended for public consumption. Furthermore, if we did not bundle this fork into AWS CLI v2 and because of Python package namespacing mechanics, it would force Linux distro maintainers to essentially have to choose between having their users/downstream packages depend either on the publicly available `botocore` package or the backwards incompatible, unreleased, unstable version of `botocore` that is only usable by/required for AWS CLI v2. Prior to the bundling, maintainers reasonably chose the former, which meant they could not pull in AWS CLI v2 into their distros. Bundling the AWS CLI v2's version of `botocore` removes the need to make this choice and thus help unblock being able to package v2.

Hope this helps!

kyleknap:
@glaubitz can you link what exact policy that would prevent v2 from being packaged into openSUSE/SLE and Debian?

The `botocore` that is maintained in AWS CLI v2 has many fundamental differences (and will continue to grow in fundamental differences) than what is in the official `botocore` library to the point that I'd argue they are farther from being duplicates and closer to being separate packages that just share the same `botocore` name. Or is more of the problem around v2 retaining the name `botocore` for its implementation (as opposed to renaming its implementation to something different e.g. `awsclicore`)?

glaubitz:
See:

> https://wiki.debian.org/EmbeddedCopies
> https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries

Couldn't find the openSUSE policy documents quickly but we have the same policy.

PR #6527, opened 2021-11-02
Move to git+https:// for requirements

nateprewitt:
https://github.blog/2021-09-01-improving-git-protocol-security-github/

Github started a brownout today for users still on git:// for their repository links. We're moving to HTTPS to make sure we're using proper cipher suites when cloning.


justindho (approved):
:ship:

Issue #6545, opened 2021-11-15
IAM new-role wizard doesn't handle nullable description


kiiadi:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
While running the iam `new-role` wizard (e.g. `aws iam wizard new-role`) the description attribute is not required, if a value isn't added, an 'empty' key is added to the generated CloudFormation template - which is invalid - resulting in a failed deployment via cloudformation.

**SDK version number**
aws-cli/2.3.2 Python/3.8.5 Linux/4.14.252-195.483.amzn2.x86_64 source/x86_64.amzn.2 prompt/off

**Platform/OS/Hardware/Device**
Cloud 9

**To Reproduce (observed behavior)**
Steps to reproduce the behavior

**Expected behavior**
The Description property should just be dropped from the YAML if it's not specified.

**Logs/output**
Get full traceback and error logs by adding `--debug` to the command.

**Additional context**

tim-finnigan:
Thanks @kiiadi for reaching out and bringing this to our attention. I could reproduce the behavior you described. I created the PR linked above to address this.

stealthycoin:
Merged the linked PR.

PR #6548, opened 2021-11-16
Validate description for iam new-role wizard

tim-finnigan:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/6545

*Description of changes:*  Only show "Description: ..." in Cloudformation template preview if it exists.

### DIFF
--- a/awscli/customizations/wizard/wizards/iam/new-role.yml
+++ b/awscli/customizations/wizard/wizards/iam/new-role.yml
@@ -174,7 +174,7 @@ plan:
               Type: "AWS::IAM::Role"
               Properties:
                 RoleName: {role_name}
-                {% if {role_description} !=   %}
+                {% if '{role_description}' != '' %}
                 Description: {role_description}
                 {% endif %}
                 AssumeRolePolicyDocument:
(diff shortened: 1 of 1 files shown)


awscli/customizations/wizard/wizards/iam/new-role.yml
```diff
@@ -174,7 +174,9 @@ plan:
               Type: "AWS::IAM::Role"
               Properties:
                 RoleName: {role_name}
+                {% if {role_description} !=   %}
```
nateprewitt, line 177: I'm curious if there's a more idiomatic way to specify `role_description` being null. Does `!= ''` satisfy the conditional? That may look more complete if it also applies to `null` values correctly.

tim-finnigan: Hi @nateprewitt I agree that looks off but neither `%if {role_description} != '' %}` or `{% if {role_description} %}` worked. And it looks like either a `==` or `!=` operator is required [here](https://github.com/aws/aws-cli/blob/v2/awscli/customizations/wizard/core.py#L204-L232).

But adding quotes around role_description like this worked:  `{%if '{role_description}' != '' %}`, so I updated the PR. Thanks!

Issue #5001, opened 2020-02-25
v2 aws kafka create-configuration behavior is different than v1


herbertgoto:
With v1 of the awscli I'm able to make a command call like:

`aws kafka create-configuration --name "Test2" --kafka-versions "2.3.1" --server-properties file://configuration.txt`

However, with v2, running the same command the awscli returns an error of:

```
An error occurred (BadRequestException) when calling the CreateConfiguration operation: [ {
  "name" : "j�hr����h�',zv���k��^��^��bq��nW",
  "value" : "",
  "errorMessages" : [ "Unsupported config j�hr����h�',zv���k��^��^��bq��nW for Kafka Version 2.3.1" ]
} ]
```

aws-cli/2.0.0 Python/3.7.4 Darwin/18.7.0 botocore/2.0.0dev4

herbertgoto:
BTW, my configuration file has following content:

```
auto.create.topics.enable = true
delete.topic.enable = true
log.retention.hours = 8
```

KaibaLopez:
Hi @herbertgoto ,
Yea, I get the same error, thanks for pointing it out!

stealthycoin:
This is a blob binary type input structure, which means it is affected by our change to how binary input and output work in the CLI. Upgrading guide is here: https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-binaryparam

Basically you have two options, revert to previous behavior with a config option(outlined in the link). Or use `fileb://` instead of `file://`.

herbertgoto:
You are right. Nevertheless, the documentation does not state it like that. 
![Screen Shot 2020-02-28 at 4 30 14 PM](https://user-images.githubusercontent.com/8704017/75589274-577a4f00-5a48-11ea-9934-94551a069df2.png)

stealthycoin:
Merged #6551

PR #6551, opened 2021-11-17
Replace file:// with fileb:// in kafka create-configuration documentation

tim-finnigan:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/5001

*Description of changes:* Replace file:// with fileb:// in kafka create-configuration documentation for `--server-properties` example.

### DIFF
--- a/awscli/examples/kafka/create-configuration.rst
+++ b/awscli/examples/kafka/create-configuration.rst
@@ -6,7 +6,7 @@ The following ``create-configuration`` example creates a custom MSK configuratio
         --name "CustomConfiguration" \
         --description "Topic autocreation enabled; Apache ZooKeeper timeout 2000 ms; Log rolling 604800000 ms." \
         --kafka-versions "2.2.1" \
-        --server-properties file://configuration.txt
+        --server-properties fileb://configuration.txt
 
 Contents of ``configuration.txt``::
 


awscli/examples/kafka/create-configuration.rst
```diff
@@ -6,7 +6,7 @@ The following ``create-configuration`` example creates a custom MSK configuratio
         --name "CustomConfiguration" \
         --description "Topic autocreation enabled; Apache ZooKeeper timeout 2000 ms; Log rolling 604800000 ms." \
         --kafka-versions "2.2.1" \
-        --server-properties file://configuration.txt
+        --server-properties fileb://configuration.txt
```
nateprewitt, line 9: This change should be specific to the CLI v2 right? Do we need it on that branch specifically? We may want to reach out to @elysahall to see if there's a strategy for managing doc variations between the two branches. Otherwise, this may regress with future updates.

elysahall: This topic has come up before, last I knew of there is no strategy for having separate V1 and V2 examples. I'll touch base with the team to see if this is still the case.

stealthycoin: I think we should just merge it into the v2 branch. Potentially this will cause issues when merging the v1 branch back into v2 during updates. Someone would need to deal with that merge conflict manually.

tim-finnigan: Ok I just changed the branch to v2.

Issue #5279, opened 2020-06-09
AWS CLI CloudTrail Events Results -> EventTime timestamp format changed from Unix/ epoch timestamp to other timetamp format


benishue:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [ x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [ x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
The timestamp format for the field (EventTime) in the CloudTrail lookup-events result in the documentation is Unix epoch format 

[https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudtrail/lookup-events.html#examples](url)

> aws cloudtrail lookup-events --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin
> Output:
> 
> {
>   "Events": [
>      {
>          "EventId": "654ccbc0-ba0d-486a-9076-dbf7274677a7",
>          "Username": "my-session-name",
>          **"EventTime": 1453844532.0,**

_See image below_
While running the same command using the AWS CLI, the format of this field is different

> [bla@bla bla ~]# aws cloudtrail lookup-events --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin
> {
>     "Events": [
>         {
>             "EventId": "bla-bla-bla",
>             "EventName": "ConsoleLogin",
>             "ReadOnly": "false",
>             "**EventTime": "2020-06-09T12:57:37+02:00**",
>             "EventSource": "bla.amazonaws.com",
>             "Username": "bla@bla.com",
>             "Resources": [], ...
> 
> 
> 
_See image below_
**SDK version number**
aws-cli/2.0.17 Python/3.7.3 Linux/3.10.0-1062.4.1.el7.x86_64 botocore/2.0.0dev21

**Platform/OS/Hardware/Device**
What are you running the cli on?
CentOs 7

**To Reproduce (observed behavior)**
Steps to reproduce the behavior
Run the following command from the AWS CLI [documentation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudtrail/lookup-events.html);
`aws cloudtrail lookup-events --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin`

**Expected behavior**
[...]

PR #6555, opened 2021-11-18
Fix CloudTrail response syntax example

tim-finnigan:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/5279

*Description of changes:* EventTime response syntax should show a date string rather than UNIX timestamp.


nateprewitt (approved):
Looks good 🚀

PR #6565, opened 2021-11-23
Add support for message-only and json log formatting

jamesls:


awscli/customizations/logs/tail.py
```diff
@@ -128,7 +152,7 @@ class TailCommand(BasicCommand):
         {
             'name': 'format',
             'default': 'detailed',
-            'choices': ['detailed', 'short'],
+            'choices': ['detailed', 'short', 'message-only', 'json'],
```
kyleknap, line 155: I think we need to remove the `message-only` option from the choices. Looks like an update left over from the previous version

stealthycoin (approved):

kyleknap (approved):
Looks good :ship:

ebi-yade:
Hello, maintainer-san 👋🏽 

I would like to ask if this feature is already available.
I found that version 2.4.1 was released after this PR merged with the v2 branch.  Then I tried it and got the result below: an invalid format.
The same command with the latest version ended up with the same result.

I would appreciate it if you give some reaction.  And if not, also I am planning to create an issue after some weeks.

Thanks,

image

PR #6566, opened 2021-11-23
Allow for `~` for the specified cfn template path

jamesls:
This allows a user to specify paths such as `~/foo/mytemplate.yml`.  I ended up refactoring the code a bit to create a separate `load_template_file` so I could test it directly.


kyleknap (approved):
Looks good :ship:

PR #6575, opened 2021-11-29
[v2] Update new rule wizard 

jamesls:
* Add CloudWatch Logs group as a target
* Add CodeCommit as a service to match events for


kyleknap (approved):
Looks good!

Issue #6582, opened 2021-12-01
s3 ls --short


pixarbuff:
Currently the command `aws s3 ls --recursive mybucket` gives long-form results with timestamp, filesize, and key:
```
2021-11-04 14:29:56    3714945 sub/foo.csv
2021-11-19 22:58:25    3706448 bar.csv
2021-11-30 17:01:54        485 baz.csv
```

It would be nice to specify an option to show short-form results that give the key only, like
```
aws s3 ls --short --recursive mybucket
sub/foo.csv
bar.csv
baz.csv
```

without resorting to external commands to parse out the key, as described in these [answers on stackoverflow](https://stackoverflow.com/questions/36813327/how-to-display-only-files-from-aws-s3-ls-command)

stobrien89:
Hi @pixarbuff,

Thanks for the feature request!  We'll leave this open for tracking purposes for right now.  For anyone who is interested in this feature, please leave a reaction on the original post.

sudarshanvm:
@stobrien89 @pixarbuff I am working on adding this feature in aws list commands

stobrien89:
Hi all,

Just wanted to post a workaround for the time being (also mentioned in the [linked stackoverflow article](https://stackoverflow.com/questions/36813327/how-to-display-only-files-from-aws-s3-ls-command)) for better visibility.

It is possible to filter output using `jq` or `awk`, but not all customers will have access to these utilities.  The CLI has built-in client-side filtering via the `--query` option, which utilizes [JMESpath ](https://jmespath.org/) to filter JSON output.  

The desired result can be achieved by using the [S3Api `list-objects-v2` operation](https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html) (which `ls` uses under the hood for listing objects):

```
aws s3api list-objects-v2 --bucket "bucket-name" --query "Contents[].Key[]"
```

Note: The query option will not work for `ls` as the output is not serialized into JSON.

PR #6605, opened 2021-12-09
Add s3 ls --short to display the files or objects names without times…

sudarshanvm:
…tamp or size

*Issue #, if    ###available:* #6582 

*Description of changes:*
Currently the command aws s3 ls --recursive mybucket gives long-form results with timestamp, filesize, and key, so added "aws s3 ls --short" to display only the key without displaying the timestamp or file size.


stobrien89:
Hi @sudarshanvm,

Thanks for the pull request.  I'll have the team take a look and see if they'd accept this change.  If so, I know for sure we'll need some accompanying tests, but I'll see if I can get some specifics as to what those might be.

stobrien89:
Hi @sudarshanvm,

Thanks for the pull request!  I discussed this with the team today— we're not able to accept this because the subcommands as a whole are due for a refactoring and we'd like to see the feature request get a bit more traction before we look at implementing this.  Sorry for the inconvenience!

PR #6613, opened 2021-12-12
Adding Upgrade pip-based install syntax

Rakesh9100:
*Issue #, if available:*
# Added certain syntax for Upgrade pip-based install aws

*Change Description*
Added Upgrade pip-based install syntax for ease of users


kdaily:
Hi @Rakesh9100,

Thanks for the proposed change. Unfortunately it doesn't fit in this section, which is specifically about if you cannot upgrade to Python 3, in which case you would not be able to use `pip3` either. If there are other specifics about the install process, let us know - the most up-to-date installation guide is at:

-  https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-install.html for v1, and
- https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html for v2.

PR #6645, opened 2022-01-07
Add double quote mark of query value in Example 9

lyuin:
Add double quote mark of query value in Example 9

*Issue #, if available:*

*Description of changes:*


kdaily:
Thanks for the change! I'm going to check with our docs writer on one thing. It seems all of the other examples except this one have a Windows-specific and Mac/Linux-specific command.

tim-finnigan:
It looks like this example has since been updated to include double quotes: https://github.com/aws/aws-cli/blob/9113e3c8fa97d3760f69320bacd63ec9063ae312/awscli/examples/ec2/describe-instances.rst?plain=1#L274

PR #6653, opened 2022-01-13
Silence ecs execute-command session manager plugin check

Stretch96:
When running ` aws ecs execute-command`, it  unnecessarily outputs:

```
The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.
```

This is because within the `ExecuteCommandCaller` Class, `check_call(["session-manager-plugin"])` is used

`check_call` in this case is only used to ensure that the command exists

Changing `stddout` to devnull, will still exit if the command errors (eg. if `session-manager-plugin` doesn't exist)

### DIFF
--- a/awscli/customizations/ecs/executecommand.py
+++ b/awscli/customizations/ecs/executecommand.py
@@ -13,6 +13,7 @@
 import logging
 import json
 import errno
+import os
 
 from subprocess import check_call
 from awscli.compat import ignore_user_entered_signals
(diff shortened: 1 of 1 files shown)


awscli/customizations/ecs/executecommand.py
```diff
@@ -83,7 +84,7 @@ def invoke(self, service_name, operation_name, parameters, parsed_globals):
             # before calling execute-command to ensure that
             # session-manager-plugin is installed
             # before execute-command-command is made
-            check_call(["session-manager-plugin"])
+            check_call(["session-manager-plugin"], stdout=open(os.devnull, 'wb'))
```
justindho, line 87: What do you think about simplifying this to just use `subprocess.DEVNULL` (and importing `DEVNULL` from `subprocess` up at the top)?

```suggestion
            check_call(["session-manager-plugin"], stdout=DEVNULL)
```

justindho (commented):
@Stretch96 Thanks for the contribution! This seems reasonable to me. Just had a small comment.

tim-finnigan:
Checking in - it looks like we haven't heard back in a couple of years here. I'm going to close this PR as inactive, if still interested in these changes please create a new issue/PR for continuing the discussion.

PR #6668, opened 2022-01-20
Update EMR visible-to-all-users description

ellkend-aws:
*Issue #, if available:*
N/A

*Description of changes:*
Updated the EMR help text to note that setting the properties 'visible-to-all-users' and 'no-visible-to-all-users' for an EMR cluster is no longer supported. 


awscli/customizations/emr/helptext.py
```diff
@@ -262,6 +262,9 @@
     ' the request to terminate was submitted.</li>'
 )
 VISIBILITY = (
+    '<p>Using this option to set cluster visiblity is no longer supported.'
+    ' The value is set to <code>true</code> by default, and setting it to'
+    ' <code>false</code> now has no effect.</p>'
     '<p>Specifies whether the cluster is visible to all IAM users of'
```
nateprewitt, line 268: Should we consider cleaning the remaining help string up a bit. Perhaps only including what `true` implies since `false`/`--no-visible-to-all-users` is no longer relevant?

ellkend-aws: @nateprewitt Open to more input on the wording, but we could change it to something like "Specifies whether the cluster is visible to all IAM users of the AWS account associated with the cluster. If a user has the proper policy permissions set, they can also manage the cluster. Visibility is on by default. The --no-visible-to-all-users option is no longer supported. To restrict cluster visibility, use an IAM policy."

We can optionally remove mention of the --no-visible-to-all-users altogether.

nateprewitt: I think that new provided wording is fine, it gets the same meaning across more succinctly.

nateprewitt (approved):
LGTM, thanks @ellkend-aws

PR #6675, opened 2022-01-26
Add CLI parameters for `aws configure sso` to skip prompted value 

dradetsky:
*Description of changes:*

Adds some flags to allow the use of `aws configure sso` non-interactively.

NOTE: This is a work in progress. Hoping I can get some preliminary approval before I actually clean it up.

This has been marked TODO since at least 2019-11-06, which is part of why I'm hesitant to submit a 100% cleaned-up PR; maybe there's some reason why previous PRs were rejected.

Anyway, my use-case is: In my org, we want to write some scripts that automate things across a range of aws accounts (e.g. the "prod" account, the "dev" account, the "dmr-fucking-around-with-k8s" account). However, we've never been consistent about how people name profiles, and as the number of accounts gets bigger we're never going to fix this. So we'd rather have a second script to create any script-required profiles so we don't have to worry about whether the user has done `aws configure sso` already or what profile naming convention they have.

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Sure

### DIFF
--- a/awscli/customizations/configure/sso.py
+++ b/awscli/customizations/configure/sso.py
@@ -125,7 +125,55 @@ class ConfigureSSOCommand(BasicCommand):
         '\n\nNote: The configuration is saved in the shared configuration '
         'file. By default, ``~/.aws/config``.'
     )
-    # TODO: Add CLI parameters to skip prompted values, --start-url, etc.
+    ARG_TABLE = [
+        {
+            'name': 'sso-start-url',
+            'required': False,
+            'help_text': (
+                'help get fucked'
+            )
+        },
+        {
+            'name': 'sso-role-name',
+            'required': False,
+            'help_text': (
+                'help get fucked'
+            )
+        },
+        {
+            'name': 'sso-account-id',
+            'required': False,
+            'help_text': (
+                'help get fucked'
+            )
+        },
+        {
+            'name': 'sso-region',
+            'required': False,
+            'default': 'us-east-1'
+        },
+        {
+            'name': 'default-region',
+            'required': False,
+            'help_text': (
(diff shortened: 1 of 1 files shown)


dradetsky:
speaking of confirming you can use this, I also swear on..._checks desk_...my lucky pen that I didn't steal this from some proprietary codebase (i.e. dev cert of origin)

joguSD:
Hey @dradetsky, thanks for working on this! I think this looks mostly correct as far as implementation goes but we'll need some docs, a changelog entry, as well as tests. I have another PR that's likely to conflict with this: #6743 and will likely be merged before this. I'll try and get some time to rebase and finish up the tests as well. 

Perhaps the only thing that will need some more consideration is the `--dry-run` flag. I'm not opposed to the idea of having a mode where we dump the updated profile to `stdout` instead of writing back to the config file, but it might make more sense to actually print the formatted profile instead of the raw dictionary.

dradetsky:
@joguSD yeah `--dry-run` was something I wasn't really sure about. I think it should probably just be taken out.

tim-finnigan:
Checking in — the PR referenced above (https://github.com/aws/aws-cli/pull/6743) has since been merged. Also as mentioned above, further investigation are changes here regarding the [TODO](https://github.com/aws/aws-cli/blob/03be35e71e2e523421b3d809b83769b3876b6808/awscli/customizations/configure/sso.py#L368) ("...docs, a changelog entry, as well as tests"). I'm going to close this PR as not planned but created https://github.com/aws/aws-cli/issues/8769 to track further discussion going forward.

PR #6678, opened 2022-01-27
Deprecate datapipeline create-default-roles

joguSD:
Support for the underlying managed policies will be deprecated with no replacement meaning this custom command will no longer function properly if the roles do not already exist. This removes the command from the documentation and prints a warning notice that the command has been deprecated.

### DIFF
--- a/.changes/next-release/enhancement-datapipeline-86652.json
+++ b/.changes/next-release/enhancement-datapipeline-86652.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "datapipeline",
+  "description": "Deprecated support for the datapipeline create-default-roles command"
+}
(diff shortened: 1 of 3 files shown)


nateprewitt (changes requested):
We may want to also update the tests to capture the warning with `pytest.warns` or `@pytest.mark.filterwarnings`.

```
tests/unit/customizations/datapipeline/test_create_default_role.py::TestCreateDefaultRole::test_constructed_result
tests/unit/customizations/datapipeline/test_create_default_role.py::TestCreateDefaultRole::test_default_roles_exist
tests/unit/customizations/datapipeline/test_create_default_role.py::TestCreateDefaultRole::test_default_roles_not_exist
  /home/runner/work/aws-cli/aws-cli/awscli/customizations/datapipeline/createdefaultroles.py:75: UserWarning: 
  Support for this command has been deprecated and may fail to create these roles
  if they do not already exist.
  
    warnings.warn(_DEPRECATION_NOTICE)
```

stealthycoin (approved):

nateprewitt (approved):

Issue #6507, opened 2021-10-26
[v2] Increase prompt-toolkit version range to support 3.x


kyleknap:
Brought up in this [comment](https://github.com/aws/aws-cli/pull/6352#issuecomment-950389706), the request is to increase the version range of `prompt-toolkit` to use major version 3 as we are still on version 2. According to the [CHANGELOG](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/df8246cf03deec681a521c03a12a68d712158e24/CHANGELOG#L334), it does not seem it would be too difficult for us to migrate to version 3 of prompt-toolkit.

kyleknap:
Yep! It now supports 3.x of `prompt-toolkit`.

PR #6703, opened 2022-02-09
[v2] Port logic to prompt-toolkit 3

kyleknap:
Fixes https://github.com/aws/aws-cli/issues/6507

The main difference is that PTK3 uses the built-in Python asyncio event loop. This required a large refactor of the test scaffolding as some of the synchronous assumptions were no longer valid. With regards to refactoring, I pivoted it to spinning up the application in a thread and then feed input keys to the app and make assertions on it while the app was running. This was also cleaner in that we did not have to add actions and assertions to a queue to be processed and instead can just make the assertion/action directly.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -432,7 +432,7 @@ class PromptToolkitAppRunner:
             self._wait_until_app_is_done_updating()
 
     def wait_for_completions_on_current_buffer(self):
-        if self.app.current_buffer.complete_state:
+        if self._current_buffer_has_completions():
             return
         self.app.current_buffer.on_completions_changed.add_handler(
             self._notify_done_completing
@@ -460,6 +460,12 @@ class PromptToolkitAppRunner:
 
     def _notify_done_completing(self, app):
         self._done_completing_event.set()
+
+    def _current_buffer_has_completions(self):
+        return (
+            self.app.current_buffer.complete_state and
+            self.app.current_buffer.complete_state.completions
+        )
 
     def _convert_key_to_vt100_data(self, key):
         return REVERSE_ANSI_SEQUENCES.get(key, key)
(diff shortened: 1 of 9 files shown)


tests/unit/customizations/wizard/test_app.py
```diff
@@ -36,12 +31,482 @@
 from awscli.customizations.wizard.core import BaseStep, Executor
 
 
+@pytest.fixture
```
stealthycoin, line 34: How slow is all this stuff? I really don't want the unit tests to be significantly slowed down by these tests. It seems more reasonable for them to be functional tests.

tests/__init__.py
```diff
@@ -370,117 +374,82 @@ def stream(self, **kwargs):
...
-                app.exit()
-
-        self._app.input = FakeApplicationInput()
-        self._app.after_render = prompt_toolkit.utils.Event(
-            self._app, callback)
-
-        _stdout = TextIOWrapper(BytesIO(), encoding="utf-8")
-        self._app.output.stdout = _stdout
-        self._app.run(pre_run=pre_run)
+            self._wait_until_app_ui_is_no_longer_redrawing()
```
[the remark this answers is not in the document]
kyleknap: I ended up going back to that in the next commit

tests/unit/customizations/wizard/test_app.py
kyleknap: In the next commit, it should be much faster as there are not blanket sleep statements

kyleknap:
@stealthycoin I made some adjustments to how we wait for the app updates to propagate and also incorporated your other feedback. I ended up using the `after_render` property to help drive when to stop waiting. The tests run much faster now and I believe I removed any of the flakiness from them.

stealthycoin (approved):

stealthycoin (approved):

PR #6706, opened 2022-02-10
Fix typo in ec2 example

ductnn:
*Issue #, if available:*

*Description of changes:*
This PR fixes small typo in example.


kdaily:
HI @ductnn,

Thanks for your proposed change! In reviewing it, I'm wondering if this example is not correct at all. Looking at the API docs for a `CidrBlockAssociationSet`, I don't think they can have a `NetworkBorderGroup` property at all! If you want, can you review the whole example structure to be sure that it's correct?

PR #6730, opened 2022-02-16
Loosen SSO login requirements

### LINKED ISSUE #5727: [v2] `aws sso login` should not require a fully-configured profile
Currently, `aws sso login` operates on a particular profile, even [requiring that `sso_account_id` and `sso_role_name` be present in the profile](https://github.com/aws/aws-cli/blob/f2788558422dac42a5ebe37c7e5a3d24b19dee9f/awscli/customizations/sso/login.py#L32) even though it does not use them, [only fetching the token](https://github.com/aws/aws-cli/blob/f2788558422dac42a5ebe37c7e5a3d24b19dee9f/awscli/customizations/sso/utils.py#L45) (as it should, because AWS SSO-capable SDKs can use the token to get credentials for the appropriate account and role).

At the very least, `sso_account_id` and `sso_role_name` should be removed from the list of required config variables, which would allow a profile like:
```ini
[profile login]
sso_start_url = https://d-2e69cb2b10.awsapps.com/start
sso_region = us-east-2
```
and then `aws sso login --profile login` would just work without requiring a specific account and role that won't be used anyway.

This matters because not all users in an organization have the same permissions, so there's not a good way to provide them all with a single working config file to start from.

A better alternative would be to have AWS SSO configuration be explicit in the config file, perhaps with a new section type:
```ini
[sso default]
sso_start_url = https://d-2e69cb2b10.awsapps.com/start
sso_region = us-east-2
```

Or, `aws sso login` should check the configured profiles and if there's only one AWS SSO configuration (i.e., they all use the same start URL and region), it should just use that.

I've implemented the latter in [`aws-sso-util login`](https://github.com/benkehoe/aws-sso-util#logging-in-and-out).

joguSD:
Fixes #5727

Fundamentally, the `aws sso login` command derives an access token for the given `sso_start_url` and is cached globally for that `sso_start_url`. The particular `sso_account_id` / `sso_role_name` are irrelevant to actually performing the OIDC device authorization flow and aren't true requirements to perform the `aws sso login`.

### DIFF
--- a/.changes/next-release/enhancement-SSO-16533.json
+++ b/.changes/next-release/enhancement-SSO-16533.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "SSO",
+  "description": "Fixes `#5727 <https://github.com/aws/aws-cli/issues/5727>`__. Loosens the requirements to perform an ``aws sso login`` command to just the ``sso_start_url`` and ``sso_region``."
+}
--- a/awscli/customizations/sso/login.py
+++ b/awscli/customizations/sso/login.py
@@ -26,14 +26,15 @@ class LoginCommand(BasicCommand):
         'Retrieves and caches an AWS SSO access token to exchange for AWS '
         'credentials. To login, the requested profile must have first been '
         'setup using ``aws configure sso``. Each time the ``login`` command '
-        'is called, a new SSO access token will be retrieved.'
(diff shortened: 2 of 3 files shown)


lorengordon:
Oh this is fantastic. I was just documenting this for our users today, and kept wondering why these arguments were necessary! Thanks!

kyleknap (approved):
Looks good! :ship:

PR #6735, opened 2022-02-18
[docs] Fixing broken literal code block in DynamoDb create table example

t5kaji:
*Description of changes:*

It seems that the doc doesn't introduce literal code block well for DynamoDb create-table command example.
image
I just added "::" to make it work 👍


nateprewitt (approved):
Thanks, @t5kaji!

PR #6743, opened 2022-02-24
Support the --no-browser flag on sso login commands

### LINKED ISSUE #5301: [v2] aws sso commands should allow opting out of automatically opening a browser
**Is your feature request related to a problem? Please describe.**
I'm using a remote shell via SSH to a machine that has a GUI (macOS). When performing `aws configure sso` or `aws sso login` it dumps out the url and code to enter into a browser myself. However, the browser has already opened on the remote machine even though I cannot easily access it (nor want to). AWS does not allow me to use the url and code from the console in this scenario to login from my host machine because it was already used by the remote, unseen, browser.

**Describe the solution you'd like**
A CLI flag like `--no-browser` to avoid launching the default browser on the system, allowing me to use the url and code myself manually.

**Describe alternatives you've considered**
A CLI interactive prompt asking whether you'd like to open the default browser. I like this too, and think it should exist before automatically opening another window on my system, but I feel like CLI flags should always exist as well.

joguSD:
Added support for the `--no-browser` flag on the `aws confgure sso` and `aws sso` login commands.

Closes #5301.

### DIFF
--- a/awscli/customizations/sso/utils.py
+++ b/awscli/customizations/sso/utils.py
@@ -29,6 +29,18 @@ SSO_TOKEN_DIR = os.path.expanduser(
     os.path.join('~', '.aws', 'sso', 'cache')
 )
 
+LOGIN_ARGS = [
+    {
+        'name': 'no-browser',
+        'action': 'store_true',
+        'default': False,
+        'help_text': (
+            'Disables automatically opening the verfication URL in the '
+            'default browser.'
+        )
+    }
+]
+
 
 def _serialize_utc_timestamp(obj):
     if isinstance(obj, datetime.datetime):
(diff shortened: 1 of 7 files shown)


awscli/customizations/sso/utils.py
```diff
@@ -65,6 +65,25 @@ def open_browser_with_original_ld_path(url):
         webbrowser.open_new_tab(url)
 
 
+class PrintOnlyHandler(object):
```
nateprewitt, line 68: We probably don’t need the object inherentence anymore.

nateprewitt (approved):
Minor note, otherwise looks good.

awscli/customizations/sso/utils.py
```diff
@@ -65,6 +65,25 @@ def open_browser_with_original_ld_path(url):
...
+                 verificationUriComplete, **kwargs):
+        opening_msg = (
+            f'Browser will not be automatically opened.\n'
+            f'Please visit the following URL:\n'
+            f'\n{verificationUri}\n'
+            f'\nThen enter the code:\n'
+            f'\n{userCode}\n'
+            f'\nAlternatively, you may visit the following URL:'
+            f'\n{verificationUriComplete}\n'
+            f'\nWhich will autofill the code upon loading.\n'
```
kyleknap, line 82: It may be nice to move this line up to with the line `Alternatively, you may visit the following URL:`. When trying out the parameter, it took me a few looks to realize that the autofill was referring to the URL and not something that was about to be autofilled in the terminal as login progress was halted at that line.

tests/unit/customizations/configure/test_sso.py
```diff
@@ -256,6 +257,19 @@ def test_basic_configure_sso_flow(self):
...
         self.assert_config_updates()
 
+    def test_single_account_single_role_flow_no_browser(self):
+        self._add_prompt_responses()
+        self._add_simple_single_item_responses()
+        with self.sso_stub:
+            self.configure_sso(
+                args=['--no-browser'],
+                parsed_globals=self.global_args,
+            )
```
kyleknap, line 267: I wonder if there is any additional assertion we can make here to make sure that the browser does not open or we use the expected handler?

awscli/customizations/sso/utils.py
kyleknap: Or even better, it would be great if we can make a base class that both the `PrintOnlyHandler` and `OpenBrowserHandler` subclass from. Specifically, that will make it clear what the interface that is needed as a `on_pending_authorization` handler.

kyleknap (commented):
Looks good to me. I just had some small feedback.

kyleknap (approved):
Looks good! :ship:

PR #6780, opened 2022-03-11
Fix URL, causes an error when rendering help.

kdaily:
*Issue #, if available:*

NA

*Description of changes:*

The command `aws cloudformation update-stack help` shows the following error:

```
<string>:600: (ERROR/3) Unknown target name: "updating a stack".
```

The reference being made is not correct.


nateprewitt (approved):
⛵

Issue #6387, opened 2021-09-09
global output flag is not respected for aws ddb commands


patrickgreenwell:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
When using the `aws ddb select` and `aws ddb put` operations the global --output flag is not respected 
**SDK version number**
aws-cli/2.2.5 Python/3.8.8 Linux/5.10.16.3-microsoft-standard-WSL2 exe/x86_64.ubuntu.20 prompt/off

**Platform/OS/Hardware/Device**
aws-cli/2.2.5 Python/3.8.8 Linux/5.10.16.3-microsoft-standard-WSL2 exe/x86_64.ubuntu.20 prompt/off

**To Reproduce (observed behavior)**
1. Run a aws ddb select <tableName> with output flag set to text, json, yaml, or yaml-stream. 
2. Note the output format does not change except in the cast of yaml-stream to say it is unsupported

**Expected behavior**
Output is formatted in the respective formats using the global formatters 

OR

CLI returns an error message stating format is not accepted.

**Logs/output**
Get full traceback and error logs by adding `--debug` to the command.
```console
:~$ aws ddb select SampleTable --output yaml
Count: 0
Items: []
ScannedCount: 0
```

```console
:~$ aws ddb select SampleTable --output json
Count: 0
Items: []
ScannedCount: 0
```

```console
:~$ aws ddb select SampleTable --output text
Count: 0
Items: []
ScannedCount: 0
```

```console
:~$ aws ddb select OrganizationTable --output yaml-stream

yaml-stream output format is not supported for ddb commands
```

**Additional context**

[...]

PR #6783, opened 2022-03-14
ddb `select` output error handling

stobrien89:
Fixes https://github.com/aws/aws-cli/issues/6387

Description of changes:

Refactored `select` to throw exception if an unsupported output type is provided as a command line argument.  Previously, a request was made regardless of whether or not the output type was supported and only a YAML formatted response output was displayed. Added a note to the command reference page that only YAML output is supported. Also fixed some typos in the return code documentation.

### DIFF
--- a/awscli/customizations/dynamodb/subcommands.py
+++ b/awscli/customizations/dynamodb/subcommands.py
@@ -166,6 +166,7 @@ class SelectCommand(PaginatedDDBCommand):
         parameters.RETURN_CONSUMED_CAPACITY,
         parameters.NO_RETURN_CONSUMED_CAPACITY,
     ]
+    _SUPPORTED_OUTPUT_TYPES = ('yaml',)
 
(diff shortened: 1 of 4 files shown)


awscli/customizations/dynamodb/subcommands.py
```diff
@@ -176,6 +173,12 @@ def _run_main(self, parsed_args, parsed_globals):
         return 0
 
     def _select(self, parsed_args, parsed_globals):
+        output_type = parsed_globals.output
+        if output_type is not None and output_type != 'yaml':
```
stealthycoin, line 177: Lets hoist the constant `yaml` up to a `_SUPPORTED_OUTPUT_TYPES` and do the comparison against that collection.

stobrien89: Makes sense.  Just pushed up the change

PR #6791, opened 2022-03-15
Update copy-image.rst

mcalmari:
Corrections: `name` is mandatory in both examples and `source-image-id` was incorrectly using a snapshot ID in the second example. Once these are changed the commands work as intended.

*Issue #, if available:*

*Description of changes:*


kdaily (approved):
Thanks @mcalmari!

nateprewitt (approved):

PR #6795, opened 2022-03-16
Fix http links

elysahall:


kdaily:
I added these to #6792 to be sure they pass the new tests that @stealthycoin added.

Issue #5261, opened 2020-06-04
Provide way to get current credentials (AWS SDKs do not support SSO)


chrsmith:
My team has set up AWS SSO and is starting to use `aws sso login` for most of their needs. Everything is working smoothly. However, it appears that AWS SDKs (e.g. Golang apps that call AWS APIs) do not support _reading_ the temporary SSO credentials stored in `~/.aws/cli/cache`/`~/.aws/sso/cache`. (See https://github.com/aws/aws-sdk-go/issues/3186) 

And as a result, my team is blocked from adopting AWS SSO because it only works with the AWS CLI but none one our existing tools.

Ideally, the various language-specific AWS SDKs would be able to pick up on the current AWS SSO credentials seamlessly. Honoring `AWS_PROFILE` or `AWS_DEFAULT_PROFILE` environment variables, and using the same credential lookup algorithm as the CLI. (I assume that's the on the roadmap eventually.)

However, getting that change fixed across all of the AWS SDKs, as well as them upstreamed into tools that rely on them, will take a long time. And it would be nice to unblock my team until then so they can just rely on `aws sso login`.

I can see the credential files on my local disk, e.g. `~/.aws/sso/cache/61368d38a2497e42a24a243072108001849d0b07.json`. But it isn't clear how to map the current set of environment variables to which JSON file to load.

Could the CLI support some way of _returning_ whatever the credentials it is using? e.g.

```
# Return the SSO credentials file for the provided profile, or just use AWS_PROFILE, etc.
aws --profile staging-environment \
    sts get-caller-credentials

# Or even better: write the SSO credentials to ~/.aws/credentials. So other tools
# could just read from it.
aws sso write-credentials-to-disk --profile staging-environment
```

I don't know if there is a better approach here, as I'm do not know the specific differences between credential resolution in the AWS CLI vs. AWS SDKs. But hopefully there is some sort of workaround to make this scenario work?

aleksandrzhiliaev:
Are there any updates on this feature request?

ldormoy:
In the meantime, aws-vault v6+ is a nice solution to circumvent this issue in your local environment.

See https://github.com/99designs/aws-vault/pull/549

ryansonshine:
I've created an npm package for updating the credentials from the command line that has served me fairly well https://github.com/ryansonshine/aws-sso-creds-helper

ryansonshine:
[...]

PR #6808, opened 2022-03-23
Adding the option to export AWS credentials with a command

HugoCL:
*Issue #, if available:*

This PR aims to solve #5261, which is adding the capability to the AWS CLI to export the credentials of the current session.

*Description of changes:*
- Created a new command `aws configure export-credentials` for the purpose of the solution
- Using the current session data, I use the "frozen credentials" to get the data needed to be outputted.
- After getting the data, it's printed to STDOUT in JSON format using uni_print for better compatibility.


awscli/customizations/configure/exporter.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        Exports the AWS credentials used with the command 'aws configure' to a file.
+        """
+        credentials = session.get_credentials()
+        if credentials is None:
+            raise RuntimeError(
+                'No credentials available. Try running "aws configure" first.')
+        try:
+            credentials_data = credentials.get_frozen_credentials()
+            dump = {'aws_access_key_id': credentials_data.access_key,
```
mdaniel, line 29: Since this is a new command, without backward compat concerns, may I suggest using the same shape as is emitted by `aws sts assume-role`? it'd be one less special case for those wishing to consume the output
```json
{
    "Credentials": {
        "AccessKeyId": "ASIA...",
        "SecretAccessKey": "sekrit...",
        "SessionToken": "bAsE64=",
        "Expiration": "2022-04-13T17:24:10.189000+00:00"
    }
}
```
where obviously `SessionToken` and `Expiration` would be elided just like they are in your current output

stealthycoin:
Tests need to pass before it can be reviewed.

awscli/customizations/configure/exporter.py
HugoCL: Hi @mdaniel, I'll try to make the changes as suggested

HugoCL:
Thanks for the suggestions @benkehoe! I'll try to work on that

HugoCL:
Thanks for the suggestions @benkehoe! I'll try to work on that

tim-finnigan:
Thanks for working on this PR Hugo. https://github.com/aws/aws-cli/pull/7398 has now been merged, which mentioned building off of the interface proposed here. Now that this command is available I think the PR can be closed: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/configure/export-credentials.html.

Issue #440, opened 2013-10-25
S3 Sync with "--delete" option is deleting files that have folders with the same name!


ddarbyson:
If you have a folder in your S3 bucket and a file name that contains the same starting characters as the folder name, the target files will be intermittently be deleted upon each run.

For example, create a folder in your S3 bucket:

s3://my-bucket/test/

Now create some txt files in the bucket with 'test' as the first 4 characters of the file name.. ie:

s3://my-bucket/test-123.txt
s3://my-bucket/test-321.txt
s3://my-bucket/test.txt

Run `aws s3 sync --delete s3://my-bucket /my/local/folder`.. Do this 3 or 4 times. You will see each file 'test-123.txt', 'test-321.txt', 'test.txt' will get intermittently be deleted and downloaded with each run. Each run produces different results.

Having files unexpectedly being deleting is a big concern as we use 's3 sync --delete' for daily backup's.

Please see AWS post for originating report https://forums.aws.amazon.com/message.jspa?messageID=497335#497335

ClementWalter:
I had a similar issue with `aws-cli/1.15.10 Python/3.6.5 Darwin/15.6.0 botocore/1.10.45` but because of CAPITAL folders name. Basically I had data in my local computer within the `folder/` directory but they were stored as `s3://bucket-name/FOLDER/`.

Each call to `aws s3 sync s3://bucket-name/FOLDER folder --delete` would remove them and them again in the same `folder/` directory

beckindesigns:
I know this is an older post but the resolution to perfectly sync what you have on your local machine to your bucket all you need to do is add the --delete option to the original sync from your local (source) to your bucket (destination).

`aws s3 sync /localfolder s3://bucket-name/localfolder --delete`

If the file/folder exists on local it will create it in your bucket and if it no longer exists on local then it will delete it in your bucket. This is the proper way to fully sync (upload) your local to the bucket.

PR #6845, opened 2022-04-04
Adding a sorting process for s3 objects when "aws s3 sync --delete" is executed.

takaakibaba2021:
*Issue #, if available:*

(A similar issue: https://github.com/aws/aws-cli/issues/440 )

When "aws s3 sync <local_directory> <s3_url> --delete" command is executed, under certain conditions, both upload and delete requests are executed for a same file in local_directory.
That's because the result of list_files() for local_directory is sorted but the result of list_objects() for s3_url is not sorted.
Therefore, some files in s3_url synced from local_directory is deleted after uploaded sometimes.
```
$ aws --version
aws-cli/2.5.1 Python/3.9.11 Linux/3.10.0-1127.19.1.el7.x86_64 exe/x86_64.centos.7 prompt/off


// case 1: problem is reproduced
$ ls data/ -1
dir/
dir-2/

$ aws <endpoint_option> s3 ls \
    s3://bucket/data/
                           PRE dir-2/
                           PRE dir/
=> all files are already synced

$ aws <endpoint_option> s3 sync --delete \
    data \
    s3://bucket/data
upload: data/dir-2/file2.txt to s3://bucket/data/dir-2/file2.txt
delete: s3://bucket/data/dir-2/file2.txt
=> files in dir-2/ are always uploaded and deleted parallelly
   (when files are deleted after uploaded, corresponding objects in s3 disappear)


// case 2: problem is not reproduced
$ ls data/ -1
dir/
dir_2/

$ aws <endpoint_option> s3 ls \
    s3://bucket/data/
                           PRE dir/
                           PRE dir_2/
=> all files are already synced

$ aws <endpoint_option> s3 sync --delete \
    data \
    s3://bucket/data
=> no output, no problem
```

*Description of changes:*

In order to solve this problem, I added a sort process for the result of list_objects() for s3_url.
I checked this problem does not happen when I use fixed AWS CLI.

### DIFF
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -319,9 +319,11 @@ class FileGenerator(object):
         else:
             lister = BucketLister(self._client)
             extra_args = self.request_parameters.get('ListObjectsV2', {})
-            for key in lister.list_objects(bucket=bucket, prefix=prefix,
-                                           page_size=self.page_size,
-                                           extra_args=extra_args):
+            objects = lister.list_objects(bucket=bucket, prefix=prefix,
+                                          page_size=self.page_size,
+                                          extra_args=extra_args)
+            for key in sorted(
+                    objects, key=lambda item: item[0].replace(os.sep, '/')):
                 source_path, response_data = key
                 if response_data['Size'] == 0 and source_path.endswith('/'):
                     if self.operation_name == 'delete':


takaakibaba2021:
@virgofx 
> Does this fix https://github.com/aws/aws-cli/issues/3273 ?

I'll check it.

And I am sorry I didn't explain this clearly enough.
This problem happens when the target object storage is Quobyte S3 ( https://www.quobyte.com/product/native-s3 ).
When I use OpenStack Swift (with S3 API) as the target object storage, this problem does not happen.
Maybe Quobyte S3 should be fixed, but this pull request may be useful if you want AWS CLI to be compatible with a wide range of object storage.
I check if there is a problem with the object listing process of Quobyte S3 too.

RyanFitzSimmonsAK:
Hi, thanks for this pull request and for your patience.  This is something Quobyte would need to fix.  The `ListObjectsV2` response from S3 actually *is* sorted, and adding client side sorting on top of that would introduce some issues.  Given that this is only an issue for third party object storage, and the CLI is intended for use with AWS, I'm going to close this pull request.

PR #6850, opened 2022-04-05
Update IoT Data doc string to account for all endpoints

zleidall:
*Description of changes:*
Refactored doc string to point to endpoint documentation.


joguSD (approved):

joguSD (changes requested):
Looks like we need to update one of the tests. It's asserting that the old note was being inserted.

joguSD (approved):

PR #6871, opened 2022-04-12
[doc]Remove OperationId in output

umihico:
Currently update-service command does not return OperationId.

### DIFF
--- a/awscli/examples/apprunner/update-service.rst
+++ b/awscli/examples/apprunner/update-service.rst
@@ -20,7 +20,6 @@ Contents of ``input.json``::
 Output::
 
     {
-        "OperationId": "17fe9f55-7e91-4097-b243-fcabbb69a4cf",
         "Service": {
             "CreatedAt": "2020-11-20T19:05:25Z",
             "UpdatedAt": "2020-11-23T12:41:37Z",


kdaily:
Hi @umihico,

Thanks for your contribution. According to the output specification, `OperationId` is returned as part of the response. See the very bottom of the section here:

https://docs.aws.amazon.com/cli/latest/reference/apprunner/update-service.html#output

The description section also refers to this parameter:

> This is an asynchronous operation. On a successful call, you can use the returned `OperationId` and the `ListOperations` call to track the operation's progress.

These docs are generated from the API model, as described here:

https://docs.aws.amazon.com/apprunner/latest/api/API_UpdateService.html#apprunner-UpdateService-response-OperationId

umihico:
@kdaily 
Thank you for the detail and asking. This happens when I actually execute the command. I agree that it seems a bug and showing OperationId is helpful.

Reproduce command is like this.
```bash
echo '{
  "ServiceArn": "YOUR_ARN_HERE,
  "InstanceConfiguration": {
    "Memory": "4 GB"
  }
}' | jq '.' > arg.json
aws apprunner update-service --cli-input-json fileb://arg.json
```

kdaily:
Hi @umihico,

Thanks! We've reproduced it here too. I've let the service team know about this - it will need to be fixed in the API model. I'm going to close this PR as we cannot make the change here, and open up a separate issue to track the resolution. Thanks again for bringing it to our attention!

kdaily:
I opened up https://github.com/aws/aws-sdk/issues/226 to track this. Thanks!

PR #6912, opened 2022-05-03
EKS: Update API version of ExecCredential to v1

jstastny:
This is because the kubeconfig generated by `kubectl` 1.24 (released 3rd May) no longer supports `client.authentication.k8s.io/v1alpha1` version and fails with:
```
error: exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"
```
This commit is more or less a naive search&replace. Some tests may be failing so more changes may be needed.

## *Description of changes:*
Change the `apiVersion` used in kube config's `exec` yaml from `client.authentication.k8s.io/v1alpha1` to `client.authentication.k8s.io/v1`.

Also add the newly required `interactiveMode` to the kube config.

### DIFF
--- a/awscli/customizations/eks/get_token.py
+++ b/awscli/customizations/eks/get_token.py
@@ -74,7 +74,7 @@ class GetTokenCommand(BasicCommand):
 
         full_object = {
             "kind": "ExecCredential",
-            "apiVersion": "client.authentication.k8s.io/v1alpha1",
+            "apiVersion": "client.authentication.k8s.io/v1",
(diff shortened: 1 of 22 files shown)


jstastny:
https://github.com/aws/aws-cli/pull/6476 seems to be more sophisticated implementation of the same thing.

kdaily:
Thanks @jstastny! I think as you note that #6476 fixes this. It was just merged and should be released soon. Closing this out!

PR #6952, opened 2022-05-15
Make regex more re-usable

evanmccarter:
In most situations a dash denotes a range
anywhere but the first and last character in a regex

The dash is intended to match literally in this pattern
and is output verbatim due to line 256
which has caused confusion when re-using the pattern with `grep`

*Description of changes:*

Move dash to end of range for better consistency when re-using pattern with other tools. 

When naively copying the current pattern to use with `grep` it unexpectedly did not match "-" but did match "]" and "^" which confused me the first time I saw this pattern. 


tim-finnigan:
Thanks for creating a PR and reaching out here. If still interested in this, please create a feature request for further discussion. We need more information here regarding specific examples of use cases that you are trying to address. We also have to be very careful about not introducing any breaking changes, and also remain consistent with S3 specifications.

PR #6963, opened 2022-05-17
Remap Eventbridge events for Sigv4a signing

nateprewitt:
This PR will port forward some of the changes from the CLI v1 that are currently missing and remaps the `EventbridgeSignerSetter` to the correct service name in the CLI v2. There's a variance between `v1` where the name is `events` and v2 where it's `eventbridge` in some parts of the code base.

### DIFF
--- a/tests/functional/eventbridge/test_put_events.py
+++ b/tests/functional/eventbridge/test_put_events.py
@@ -1,0 +1,41 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from awscli.testutils import BaseAWSCommandParamsTest
+
+
+class TestPutEventsCommand(BaseAWSCommandParamsTest):
+
+    def run_put_events(self, cmd):
+        response, _, _ = self.run_cmd(cmd)
+        return response
+
+    def test_put_events_with_endpoint_id(self):
+        """Verify SigV4a signer has been loaded for put-events with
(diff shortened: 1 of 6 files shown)


tests/functional/eventbridge/test_put_events.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestPutEventsCommand(BaseAWSCommandParamsTest):
+    def setUp(self):
+        super().setUp()
+        self.access_key = "ABCDEFGHIJKLMNOPQRST"
+        self.secret_key = "TSRQPONMLKJUHGFEDCBA"
+        self.session_token = "TOKENTOKENTOKENTOKEN"
+        self.environ["AWS_ACCESS_KEY_ID"] = self.access_key
+        self.environ["AWS_SECRET_ACCESS_KEY"] = self.secret_key
```
kyleknap, line 32: We should not be needing to define a `setUp` as the [`BaseAWSCommandParamsTest`](https://github.com/aws/aws-cli/blob/5fb687de952f312f2b9ee19f226afc0975b56e94/awscli/testutils.py#L343) should be already patching the environment vars out for us.

tests/utils/botocore/__init__.py
```diff
@@ -32,6 +31,7 @@
 import datetime
 from io import BytesIO
 from subprocess import Popen, PIPE
+from unittest import mock
```
kyleknap, line 34: Probably a change we should do in the future, but most of the usage for `mock` is from the backport `mock` library. I'd prefer not updating it now to keep this change minimal. I think there is a larger effort to switch to using unittest's mock.

kyleknap (commented):
Looks fine. It would also be great to see if we get this file added back to the functional test suite: https://github.com/aws/aws-cli/blob/v2/tests/botocore/functional/test_eventbridge.py. Right now its path is `tests/botocore/functional/test_eventbridge.py` and instead we should make it `tests/functional/botocore/test_eventbridge.py`. This just looks like a miss and unfortunately because we only run over tests in the `unit` and `functional` we missed running it.

tests/utils/botocore/__init__.py
nateprewitt: Yep, this was a miss from testing. I think this work was already done for v1 last fall so it should fairly straight forward to port.

kyleknap (approved):
Looks good! :ship:

PR #6982, opened 2022-05-25
Move install script to use python -m pip

nateprewitt:
Moving the install script to use `python -m pip` instead of a hardcoded `pip` path because that's no longer a given in all environments. Since we only support Python 3.6+ (soon 3.7+), we shouldn't have any concerns that pip isn't bundled with the base Python version.


justindho (approved):
🚢

PR #6999, opened 2022-05-31
Remove f-strings to preserve error message on Python 2.7

nateprewitt:
The error message for unsupported versions broke for Python 2.7 in #6982 due to the inclusion of f-strings. We'll keep the deprecation notice for the time being and revert back to basic `format` calls for the install commands.


kyleknap (approved):
:ship:

PR #7003, opened 2022-06-01
Use Sphinx to author contribution guide

kdaily:
*Issue #, if available:*

NA

*Description of changes:*

Use Sphinx to build contributing guide. To test locally (assuming Python 3 is installed), run:

```
make html
```

Check contents in the `_build/html` directory.

This also adds a `.nojekyll` file so that GitHub does not run Jekyll on this branch.


requirements.txt
```diff
@@ -0,0 +1,4 @@
+Sphinx<5.0.0
+furo<2022.4.8
+myst-parser<0.18
+sphinx-lint<1
```
kyleknap, line 4: For these dependencies, let's do an exact pin (e.g. `sphinx-lint==0.6.0`). This will ensure that updates to these dependencies do not break our builds from underneath of us. Furthermore, we do not really need to use ranges as generally these builds will be in their own separate virtualenv and not necessarily need to be used with another project. So we can be strict on the version.

_config.yml
```diff
@@ -1,13 +0,0 @@
-remote_theme: just-the-docs/just-the-docs
```
kyleknap, line 1: Were planning to include the GitHub action update as part of this PR as well? I don't think I'm seeing it anywhere in the PR and would assume it would not be buildable by GitHub actions if we merged this as is.

Makefile
```diff
@@ -0,0 +1,137 @@
...
+	$(SPHINXBUILD) -b doctest $(ALLSPHINXOPTS) $(BUILDDIR)/doctest
+	@echo "Testing of doctests in the sources finished, look at the " \
+	      "results in $(BUILDDIR)/doctest/output.txt."
+
+htmlview: html
+	$(PYTHON) -c "import os, webbrowser; webbrowser.open('file://' + os.path.realpath('_build/html/index.html'))"
+
+check: ensure-venv
+	# Ignore the tools and venv dirs and check that the default role is not used.
+	$(SPHINXLINT) -i tools -i $(VENVDIR) --enable default-role
```
kyleknap, line 137: Let's do this in a follow up PR, but we should add a `make prcheck` rule that runs all of the rules for checking/linting the content of our docs. Here's an example of another project that does it: https://github.com/aws/chalice/blob/9de5fe604b4f0784d469de3052e256ac0b989d40/Makefile#L58. I think right now we would have it run the `check` and `linkcheck` rules? From there we could add it as part of GitHub workflow and our README so we can have quality checks running on this from the get-go.

kyleknap (commented):
Looks good! Just had some comments.

justindho (approved):
Looks good to me! Just had a small comment, but not a blocker.

.gitignore
```diff
@@ -0,0 +1,160 @@
...
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
```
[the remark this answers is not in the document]
kdaily: Yes, it was from GitHub's suggested Python `.gitignore` file, which it looked like the other AWS Python projects were using an older version of:

https://github.com/github/gitignore/blob/main/Python.gitignore

Can replace it with the one from `v2` branch though, or update the `v2` ones to use this one.

Makefile
```diff
@@ -0,0 +1,137 @@
...
+		$(MAKE) ensure-venv; \
+	fi
+
+ensure-venv:
+	@if [ ! -d $(VENVDIR) ] ; then \
+		$(PYTHON) -m venv $(VENVDIR); \
+		$(VENVDIR)/bin/python3 -m pip install --upgrade pip; \
+		$(VENVDIR)/bin/python3 -m pip install -r requirements.txt; \
+		echo "The venv has been created in the $(VENVDIR) directory"; \
+	fi
```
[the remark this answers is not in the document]
kdaily: Sure, I'll update README to reflect that workflow.

.gitignore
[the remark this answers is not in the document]
kyleknap: I'd prefer replacing this one with what is on the `v2` branch. Mainly it is misleading on whether the project actually uses some of those libraries/tools by including it in the gitignore and creates unnecessary noise if you go to update the gitignore

_config.yml
kdaily: Sure!

Makefile
kdaily: Sounds good. I updated this command to not reference the venvdir.

.gitignore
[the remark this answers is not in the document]
kdaily: Replaced with CLI v2 gitignore.

requirements.txt
```diff
@@ -1,4 +1,3 @@
-Sphinx<5.0.0
-furo<2022.4.8
-myst-parser<0.18
-sphinx-lint<1
```
kyleknap: It looks like we are not longer including `sphinx-lint` in the new requirements file? I think we should keep it if we plan to keep the `make check` rule.

Makefile
```diff
@@ -0,0 +1,116 @@
...
+	@echo "  linkcheck  to check all external links for integrity"
+	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
+
+clean:
+	-rm -rf $(BUILDDIR)/*
+
+install:
+		$(PYTHON) -m pip install -r requirements.txt; \
+		echo "Requirements installed"; \
+	fi
```
kyleknap, line 45: It looks like there is a stray `fi` statement here from the previous `if` statement that we should probably remove.

kyleknap (commented):
Looks good! I just had a few smaller follow up comments

requirements.txt
kdaily: Oops! Yep it needs to be there. Thanks!

Makefile
kdaily: Fixed, also accidentally removed the `html` make target.

kyleknap (approved):
Looks good! Let's make sure to squash all of the commits before merging too. :ship:

PR #7009, opened 2022-06-03
Add workflow to check PRs to gh-pages

kdaily:
*Issue #, if available:*

NA

*Description of changes:*

Before allowing a merge, run `sphinx-build linkcheck` and `sphinx-lint`.

### DIFF
--- a/.github/workflows/sphinx-check.yml
+++ b/.github/workflows/sphinx-check.yml
@@ -1,0 +1,22 @@
+name: Sphinx check
+
+on:
+  pull_request:
+    branches:
+    - gh-pages
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps: 
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+        cache: pip
+    - name: Install dependencies
+      run: make install
+    - name: Run checks
+      run: make check
+    - name: Run link check
+      run: make linkcheck


.github/workflows/sphinx-check.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+        cache: pip
+    - name: Install dependencies
+      run: make install
+    - name: Run checks
+      run: make check
+    - name: Run link check
+      run: make linkcheck
```
kyleknap, line 22: I mentioned this in a previous comment, but it would be nice if we had a single `make prcheck` rule that we could run locally and in CI so that we would not have to enumerate through the different make rule checks when making sure a PR will pass CI.

.github/workflows/sphinx-check.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps: 
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+        cache: pip
+    - name: Install dependencies
+      run: make install
+    - name: Run checks
```
kyleknap, line 19: In addition to running the checks, we should also have a separate step that just builds the docs (i.e. `make html`) to make sure the documentation can be generated.

kyleknap (commented):
Looks good. Just had a couple of suggestions.

.github/workflows/sphinx-check.yml
kdaily: 👍🏻 agree for the convenience on local side!

The idea for doing the steps separately in the CI was that I wanted to eventually have them show up on separate lines so it was more clear to the user which part the error was coming from, but I didn't finish investigating how to share steps between jobs so install and setup doesn't need to happen more than once.

Added it as a single step for now and can revisit.

kdaily:
Confirmed it works with changes to workflow:

https://github.com/aws/aws-cli/actions/runs/2435117577

.github/workflows/sphinx-check.yml
kdaily: I guess it's not straightforward/worth the effort to reuse the install parts. Here's an example using a matrix strategy to run each make target, which I think makes it more clear what is being run. The `make prcheck` could still be used locally.

- https://github.com/kdaily/aws-cli/pull/28/files
- https://github.com/kdaily/aws-cli/pull/28

kyleknap: With regards to command output, I agree the output is not necessarily easy to scan through. However, with regards to these checks, the user should not have to worry about the output in the first place; it should be something that they quickly run or have run in CI and only have to check the return code is zero and be able to move on quickly.

With regards to errors, is there are a specific error message that is hard to follow? With the way it is formatted now, if any of the checkers fail, it should fast fail meaning the failed command + traceback should just be at the bottom of the logs and you should not have to look through the logs of other commands for other failures.

kyleknap (approved):
Looks good! :ship:

PR #7023, opened 2022-06-08
Update helptext.py to remove visible markup tags

chatnish:
Customer called out that the documentation looked poor because of the markup tags were appearing and causing this to not display as intended.

*Issue #, if available:*

*Description of changes:*


nateprewitt (approved):
Thanks @chatnish!

PR #7028, opened 2022-06-10
make aws logs tail --follow ~2x faster

timotheecour4:
Reduce sleep time which makes aws logs tail --follow ~2x faster (for lambdas that emit frequently enough, eg every second); before this PR, you often had a ~10s latency between the time a log is written to some lambda's stdout and the time it appears in terminal (5s buffer time for cloudwatch logs + _TIME_TO_SLEEP); after this PR this becomes ~6 seconds.

### DIFF
--- a/awscli/customizations/logs/tail.py
+++ b/awscli/customizations/logs/tail.py
@@ -333,7 +333,7 @@ class NoFollowLogEventsGenerator(BaseLogEventsGenerator):
 
 
 class FollowLogEventsGenerator(BaseLogEventsGenerator):
-    _TIME_TO_SLEEP = 5
+    _TIME_TO_SLEEP = 1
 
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @timotheecour4,

Thanks! This was originally 1 second and bumped to 5:

https://github.com/aws/aws-cli/commit/7d0981dc65959a04421d6fe5efbb9dd249aef198

We'd need to investigate why that change was made and if it's still relevant here before we can change it back.

timotheecour4:
ping on this

timotheecour4:
ping, is there anything else needed on my end?

orlovw:
@timotheecour4 I think this could be a parameter or environment variable, because this can result in API DDOS

PR #7040, opened 2022-06-15
[v2] fix protocol tls deprecation

kdaily:
*Issue #, if available:*

Pulls in the same change made here: https://github.com/boto/botocore/pull/2557

> This will update our vendored copy of urllib3's `create_urllib3_context` function. This updates our use of deprecated aliases for Python 3.10 and provides some security improvements in TLS 1.3 for Python versions 3.7.4+.

*Description of changes:*

### DIFF
--- a/awscli/botocore/httpsession.py
+++ b/awscli/botocore/httpsession.py
@@ -229,20 +229,18 @@ class URLLib3Session(object):
     v2.7.0 implemented this themselves, later version urllib3 support this
     directly via a flag to urlopen so enabling it if needed should be trivial.
     """
-    def __init__(
-        self,
-        verify=True,
-        proxies=None,
-        timeout=None,
-        max_pool_connections=MAX_POOL_CONNECTIONS,
-        socket_options=None,
-        client_cert=None,
-        proxies_config=None,
+    def __init__(self,
+                 verify=True,
+                 proxies=None,
+                 timeout=None,
+                 max_pool_connections=MAX_POOL_CONNECTIONS,
+                 socket_options=None,
+                 client_cert=None,
+                 proxies_config=None,
     ):
         self._verify = verify
-        self._proxy_config = ProxyConfiguration(
-            proxies=proxies, proxies_settings=proxies_config
-        )
+        self._proxy_config = ProxyConfiguration(proxies=proxies,
+                                                proxies_settings=proxies_config)
         self._pool_classes_by_scheme = {
             'http': botocore.awsrequest.AWSHTTPConnectionPool,
             'https': botocore.awsrequest.AWSHTTPSConnectionPool,
@@ -375,7 +373,9 @@ class URLLib3Session(object):
             return self._path_url(url)
 
     def _chunked(self, headers):
-        return headers.get('Transfer-Encoding', '') == 'chunked'
+        transfer_encoding = headers.get('Transfer-Encoding', b'')
+        transfer_encoding = ensure_bytes(transfer_encoding)
+        return transfer_encoding.lower() == b'chunked'
 
     def send(self, request):
         try:
--- a/awscli/botocore/httpsession.py
+++ b/awscli/botocore/httpsession.py
@@ -1,26 +1,20 @@
-import logging
-import os
 import os.path
+import os
+import logging
 import socket
-import sys
 from base64 import b64encode
+import sys
 
-from urllib3 import PoolManager, ProxyManager, Timeout, proxy_from_url
-from urllib3.exceptions import (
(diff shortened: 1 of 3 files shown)


kdaily:
Confirmed in test output that warning is no longer present.

awscli/botocore/httpsession.py
```diff
@@ -1,46 +1,34 @@
...
-    OP_NO_SSLv3,
-    ssl,
+    ssl, OP_NO_SSLv2, OP_NO_SSLv3, OP_NO_COMPRESSION, OP_NO_TICKET,
+    PROTOCOL_TLS, PROTOCOL_TLS_CLIENT, DEFAULT_CIPHERS,
 )
-
+from urllib3.exceptions import SSLError as URLLib3SSLError
+from urllib3.exceptions import ReadTimeoutError as URLLib3ReadTimeoutError
+from urllib3.exceptions import ConnectTimeoutError as URLLib3ConnectTimeoutError
+from urllib3.exceptions import NewConnectionError, ProtocolError, ProxyError
```
nateprewitt, line 17: Is there a reason we're unordering the imports? Happy to help get them realigned after the old commit if needed.

awscli/botocore/httpsession.py
```diff
@@ -72,7 +60,13 @@ def create_urllib3_context(ssl_version=None, cert_reqs=None,
         We vendor this function to ensure that the SSL contexts we construct
         always use the std lib SSLContext instead of pyopenssl.
     """
-    context = SSLContext(ssl_version or ssl.PROTOCOL_SSLv23)
+    # PROTOCOL_TLS is deprecated in Python 3.10
+    if not ssl_version or ssl_version == PROTOCOL_TLS:
+        ssl_version = PROTOCOL_TLS_CLIENT
```
nateprewitt, line 65: The only concern here is `PROTOCOL_TLS_CLIENT` doesn't necessarily exist for older versions of urllib3 (<1.25.9). While those are mostly unsupported, we ended up adding a [special import dance](https://github.com/boto/botocore/commit/ddce27a9f4d5e788e6e38cb626b06150aa7cb295) in botocore to avoid failing.

I think this shouldn't be relevant for CLI v2 since everything is distributed, but you might consider the fallback if users are building from source with old urllib3 versions.

nateprewitt (changes requested):
Couple of quick comments, otherwise looks good!

awscli/botocore/httpsession.py
kdaily: That was a blanket copy from `botocore`s current `httpsession.py`, I can adjust it!

kdaily: Reverted these changes! Should be good now.

awscli/botocore/httpsession.py
kdaily: Thanks. Will check on this since we do technically support `urllib3>=1.25.4`.

awscli/botocore/httpsession.py
```diff
@@ -334,9 +375,7 @@ def _get_request_target(self, url, proxy_url):
             return self._path_url(url)
 
     def _chunked(self, headers):
-        transfer_encoding = headers.get('Transfer-Encoding', b'')
-        transfer_encoding = ensure_bytes(transfer_encoding)
-        return transfer_encoding.lower() == b'chunked'
+        return headers.get('Transfer-Encoding', '') == 'chunked'
```
nateprewitt, line 378: What prompted this change, it doesn't like this exists upstream in Botocore 1.x? Are we certain this isn't going to be breaking for any headers that are byte-encoded because of a different codec?

nateprewitt (commented):
One question, otherwise looks good.

awscli/botocore/httpsession.py
kdaily: Hmm. I was comparing against the v2 branch of `botocore` and took it from there, maybe I misunderstood how the vendored copy of `botocore` is there now!

https://github.com/boto/botocore/blob/v2/botocore/httpsession.py#L324-L325

kdaily: I'll undo it since it's also not in the vended copy.

awscli/botocore/httpsession.py
kdaily: Added the check here:

https://github.com/aws/aws-cli/pull/7040/files#diff-ac7672d263d603094ead793a19df43bc2a33cdbfe0f5b730489836bd82816950R26

Will confirm next week this is the best approach.

kyleknap:
Addressed as part of this PR: https://github.com/aws/aws-cli/pull/7107. Closing.

Issue #5097, opened 2020-03-31
Add man page for V2


ceuk:
`man aws` should be equivalent to `aws help`, `aws x help` etc

KaibaLopez:
Hi @ceuk,
I'm not sure if "should" or "fail" is the correct wording here, man is a linux command that helps describe other linux commands not a general application descriptor, so maybe this would be a feature request we could consider but not really a bug.

Can you describe a situation where having this would allow you to do something that can't be done with the `aws help` command is it just command preference?
I think it would be nice to add this to the cli so I'll mark it as a feature request for now, it just might be low priority if it's just a preference thing.

ceuk:
Hi @KaibaLopez  you're absolutely right, apologies for my wording, I was on my mobile on desktop mode because I had just had to jump up to deal with a crying baby (the joys of quarantine). So I was a bit briefer than I might have been and I was bit stressed at the time. Thanks for the gracious reply to a rather terse message. 

In answer to your question, from a design perspective (and admittedly, a Linux perspective),  `man aws` is more predictable than `aws help`. I actually tried: `man aws` and `aws --help` before finding `aws help`. The fact that it's `aws --version` made it slightly more confusing for me to reason out which of the possible syntaxes would be correct. 

Man pages also tend to be fairly standardised which makes them easy for those used to them to parse.

Finally (and I feel like I'm stretching here) they conform to the "everything is a file" Unix philosophy. Not sure what practical benefit that has in this instance though :)

Perhaps these might come across as fairly flimsy justifications, but in terms of cost-benefit, it seems like it would maybe be a low bar.

Thanks for the consideration

salewski:
Adding a comment to show interest in seeing this done. A quality manpage (or set of manpages) is invaluable, and is the first place people look on *nix hosts. I would volunteer to take this on, but I'm very short on time at the moment.

PR #7041, opened 2022-06-15
Prefer man directly over groff

kevin-sellers:
`groff` is great, as it easily generates ASCII text for help pages, but
lacks the escape sequences of `man` pages (bold, italic, etc.). Since
it's evident that `aws` help pages are formatted in the same way as
manual pages, I would expect it to look and feel the same way. For this
reason, instead of piping output through `groff` and then the user's
pager of choice, we now favor `man` (when available) to format these
pages in the standard way (which automatically uses the correct pager).

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

### DIFF
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -106,16 +106,23 @@ class PosixHelpRenderer(PagingHelpRenderer):
 
     def _convert_doc_content(self, contents):
         man_contents = publish_string(contents, writer=manpage.Writer())
-        if not self._exists_on_path('groff'):
-            raise ExecutableNotFoundError('groff')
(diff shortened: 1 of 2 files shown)


tim-finnigan:
Linking tracking issue: https://github.com/aws/aws-cli/issues/5097

tim-finnigan:
Thanks for creating this PR and your patience here. It looks like the changes here are now out of date. There also hasn't been any activity here or on the related issue (https://github.com/aws/aws-cli/issues/5097) for a long time. If this is still something people are interested in, please create a new feature request for further tracking and discussion.

PR #7049, opened 2022-06-17
Tweak install_requires, now compatible with AL2022

stewartsmith:
This is how we are building awscli-2 on Amazon Linux 2022 today.


tim-finnigan:
Closing as not planned — Dependabot is since being used to update dependencies.

PR #7055, opened 2022-06-22
Update example output for S3 presigned URL max expiration

AetherUnbound:
*Description of changes:*

This PR modifies the output URL for a presigned S3 object when the max expiration date is provided. Admittedly I haven't used this API before but the output URLs to both example commands are identical and I assume that the `X-Amz-Expires` query parameter is supposed to reflect the provided expiration date, so I altered the latter output to reflect this. Apologies if the outputs are indeed identical for both command!


justindho (approved):
Thanks, @AetherUnbound!

PR #7064, opened 2022-06-24
[v2] Fix deprecation warnings for cloudformation yamlhelper

kdaily:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -12,13 +12,14 @@
 # language governing permissions and limitations under the License.
 import re
 
-import ruamel.yaml as yaml
+import ruamel.yaml
 from ruamel.yaml.resolver import ScalarNode, SequenceNode
 from botocore.compat import json
 from botocore.compat import OrderedDict
 
 
 from awscli.compat import six
+from awscli.utils import dump_yaml_to_str
 
 
 def intrinsics_multi_constructor(loader, tag_prefix, node):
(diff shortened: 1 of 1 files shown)


awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -62,21 +63,19 @@ def _dict_representer(dumper, data):
...
-    # these values as strings and not booleans.
-    boolean_regex = re.compile(
-        '^(?:yes|Yes|YES|no|No|NO'
-        '|true|True|TRUE|false|False|FALSE'
-        '|on|On|ON|off|Off|OFF)$'
-    )
-    boolean_first_chars = list(u'yYnNtTfFoO')
-    resolver_cls.add_implicit_resolver(
-        'tag:yaml.org,2002:bool', boolean_regex, boolean_first_chars)
+# CloudFormation treats unquoted values that are YAML 1.1 native
```
kyleknap, line 66: Is there a particular reason we removed this function? I'd assume we would have kept it and instead just pass the `Resolver` class that is dangled as a property to the `YAML` instance. If we switch it back, it would help reduce the number of lines we are changing and repeating the logic.

awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -85,13 +84,16 @@ def yaml_dump(dict_to_dump):
...
-    _add_yaml_1_1_boolean_resolvers(FlattenAliasDumper)
-    return yaml.dump(
-        dict_to_dump,
-        default_flow_style=False,
-        Dumper=FlattenAliasDumper,
-    )
+
+    yaml = ruamel.yaml.YAML(typ="safe", pure=True)
+    yaml.default_flow_style = False
+    yaml.representer.ignore_aliases = lambda *data: True
```
kyleknap, line 90: I think I prefer if we just update the `FlattenAliasDumper` class to be a `FlattenAliasRepresenter` class instead of monkey patching the instance method with a lambda. It is a bit more explicit and easier to debug when there is a concrete class and method. We can then just update this line to be:
```python
yaml.Representer = FlattenAliasRepresenter
```

kyleknap (commented):
Looks fine. I just had a couple of suggestions. Feel free to promote this from a draft.

awscli/customizations/cloudformation/yamlhelper.py
kdaily: I put it back, thanks.

awscli/customizations/cloudformation/yamlhelper.py
kdaily: Done

PR #7070, opened 2022-06-28
[v2] Remove unused models

kyleknap:
There were a couple of categories of models from v1 that existed in the v2 repository but were no longer used and thus were just taking up install space for v2:

* `examples-1.json` - These are only used by the Python SDK as the AWS CLI has its own hand curated set of examples.
* Old API versions - In v1, API versions were configurable. In v2, this was removed and it always used the latest API version. Specifically, we had an old API version of `cloudfront` that was not used that I removed.

I also added functional tests to make sure we do not regress and accidentally pull in unused models in the future.


nateprewitt (approved):
👋

stealthycoin (approved):

PR #7081, opened 2022-07-01
[v2] Change event loop init to fix deprecation warning

kdaily:
*Issue #, if available:*

*Description of changes:*

Fixes deprecation warning in wizard tests.

```
tests/unit/customizations/wizard/test_app.py: 50 warnings
  /opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/awscli/customizations/wizard/app.py:64: DeprecationWarning: There is no current event loop
    loop = get_event_loop()
```

### DIFF
--- a/awscli/customizations/wizard/app.py
+++ b/awscli/customizations/wizard/app.py
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 import json
 import os
-from asyncio import get_event_loop
+from asyncio import new_event_loop, set_event_loop
 from collections.abc import MutableMapping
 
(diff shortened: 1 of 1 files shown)


kdaily:
Tested with:

- `aws configure wizard`
- `aws events wizard new-rule`

Both were successful.

kyleknap:
I updated the logic in this one to close the event loop at the end and remove the logic for restoring the error handler as it does not need to be restored since we are the ones to create the event loop in the first place.

stealthycoin (approved):

PR #7114, opened 2022-07-15
Deprecate ssl common name

dlm6693:
This is an attempt to maintain the usage of `sslCommonName` in the AWS CLI.  A handler is added that adds the `endpoint-url` arg to any given command if the user has not provided it themselves and the service and region are a match from a provided list.

It will be deprecated in botocore but cannot in the CLI as it causes breaking changes. This PR will counteract that. All SSL common name endpoints are hardcoded

### DIFF
--- a/.changes/next-release/enhancement-Endpoints-23278.json
+++ b/.changes/next-release/enhancement-Endpoints-23278.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Endpoints",
+  "description": "Enforce SSL common name as default endpoint url"
+}
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -484,7 +484,7 @@ class ServiceOperation(object):
         event = 'before-building-argument-table-parser.%s.%s' % \
             (self._parent_name, self._name)
         self._emit(event, argument_table=self.arg_table, args=args,
-                   session=self._session)
+                   session=self._session, parsed_globals=parsed_globals)
         operation_parser = self._create_operation_parser(self.arg_table)
         self._add_help(operation_parser)
         parsed_args, remaining = operation_parser.parse_known_args(args)
--- a/awscli/customizations/commands.py
+++ b/awscli/customizations/commands.py
@@ -134,7 +134,7 @@ class BasicCommand(CLICommand):
         event = 'before-building-argument-table-parser.%s' % \
             ".".join(self.lineage_names)
         self._session.emit(event, argument_table=self._arg_table, args=args,
-                           session=self._session)
+                           session=self._session, parsed_globals=parsed_globals)
         parser = ArgTableArgParser(self.arg_table, self.subcommand_table)
         parsed_args, remaining = parser.parse_known_args(args)
 
--- a/awscli/customizations/overridesslcommonname.py
+++ b/awscli/customizations/overridesslcommonname.py
@@ -1,0 +1,138 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+SSL_COMMON_NAMES = {
+    "sqs": {
+        "af-south-1": "af-south-1.queue.amazonaws.com",
+        "ap-east-1": "ap-east-1.queue.amazonaws.com",
+        "ap-northeast-1": "ap-northeast-1.queue.amazonaws.com",
+        "ap-northeast-2": "ap-northeast-2.queue.amazonaws.com",
+        "ap-northeast-3": "ap-northeast-3.queue.amazonaws.com",
+        "ap-south-1": "ap-south-1.queue.amazonaws.com",
+        "ap-southeast-1": "ap-southeast-1.queue.amazonaws.com",
+        "ap-southeast-2": "ap-southeast-2.queue.amazonaws.com",
+        "ap-southeast-3": "ap-southeast-3.queue.amazonaws.com",
+        "ca-central-1": "ca-central-1.queue.amazonaws.com",
+        "eu-central-1": "eu-central-1.queue.amazonaws.com",
+        "eu-north-1": "eu-north-1.queue.amazonaws.com",
+        "eu-south-1": "eu-south-1.queue.amazonaws.com",
+        "eu-west-1": "eu-west-1.queue.amazonaws.com",
+        "eu-west-2": "eu-west-2.queue.amazonaws.com",
+        "eu-west-3": "eu-west-3.queue.amazonaws.com",
+        "me-south-1": "me-south-1.queue.amazonaws.com",
+        "sa-east-1": "sa-east-1.queue.amazonaws.com",
+        "us-east-1": "queue.amazonaws.com",
+        "us-east-2": "us-east-2.queue.amazonaws.com",
+        "us-west-1": "us-west-1.queue.amazonaws.com",
+        "us-west-2": "us-west-2.queue.amazonaws.com",
+        "cn-north-1": "cn-north-1.queue.amazonaws.com.cn",
+        "cn-northwest-1": "cn-northwest-1.queue.amazonaws.com.cn",
+        "us-gov-west-1": "us-gov-west-1.queue.amazonaws.com",
+        "us-isob-east-1": "us-isob-east-1.queue.sc2s.sgov.gov",
+    },
+    "emr": {
+        "af-south-1": "af-south-1.elasticmapreduce.amazonaws.com",
+        "ap-east-1": "ap-east-1.elasticmapreduce.amazonaws.com",
+        "ap-northeast-1": "ap-northeast-1.elasticmapreduce.amazonaws.com",
+        "ap-northeast-2": "ap-northeast-2.elasticmapreduce.amazonaws.com",
+        "ap-northeast-3": "ap-northeast-3.elasticmapreduce.amazonaws.com",
+        "ap-south-1": "ap-south-1.elasticmapreduce.amazonaws.com",
+        "ap-southeast-1": "ap-southeast-1.elasticmapreduce.amazonaws.com",
+        "ap-southeast-2": "ap-southeast-2.elasticmapreduce.amazonaws.com",
+        "ap-southeast-3": "ap-southeast-3.elasticmapreduce.amazonaws.com",
+        "ca-central-1": "ca-central-1.elasticmapreduce.amazonaws.com",
+        "eu-north-1": "eu-north-1.elasticmapreduce.amazonaws.com",
+        "eu-south-1": "eu-south-1.elasticmapreduce.amazonaws.com",
+        "eu-west-1": "eu-west-1.elasticmapreduce.amazonaws.com",
+        "eu-west-2": "eu-west-2.elasticmapreduce.amazonaws.com",
+        "eu-west-3": "eu-west-3.elasticmapreduce.amazonaws.com",
+        "me-south-1": "me-south-1.elasticmapreduce.amazonaws.com",
+        "sa-east-1": "sa-east-1.elasticmapreduce.amazonaws.com",
+        "us-east-2": "us-east-2.elasticmapreduce.amazonaws.com",
+        "us-west-1": "us-west-1.elasticmapreduce.amazonaws.com",
+        "us-west-2": "us-west-2.elasticmapreduce.amazonaws.com",
+    },
+    "rds": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "docdb": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "neptune": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "health": {
+        "aws-global": "health.us-east-1.amazonaws.com",
+        "af-south-1": "health.us-east-1.amazonaws.com",
+        "ap-east-1": "health.us-east-1.amazonaws.com",
+        "ap-northeast-1": "health.us-east-1.amazonaws.com",
+        "ap-northeast-2": "health.us-east-1.amazonaws.com",
+        "ap-northeast-3": "health.us-east-1.amazonaws.com",
+        "ap-south-1": "health.us-east-1.amazonaws.com",
+        "ap-southeast-1": "health.us-east-1.amazonaws.com",
+        "ap-southeast-2": "health.us-east-1.amazonaws.com",
+        "ap-southeast-3": "health.us-east-1.amazonaws.com",
+        "ca-central-1": "health.us-east-1.amazonaws.com",
+        "eu-central-1": "health.us-east-1.amazonaws.com",
+        "eu-north-1": "health.us-east-1.amazonaws.com",
+        "eu-south-1": "health.us-east-1.amazonaws.com",
+        "eu-west-1": "health.us-east-1.amazonaws.com",
+        "eu-west-2": "health.us-east-1.amazonaws.com",
+        "eu-west-3": "health.us-east-1.amazonaws.com",
+        "me-south-1": "health.us-east-1.amazonaws.com",
+        "sa-east-1": "health.us-east-1.amazonaws.com",
+        "us-east-1": "health.us-east-1.amazonaws.com",
+        "us-east-2": "health.us-east-1.amazonaws.com",
+        "us-west-1": "health.us-east-1.amazonaws.com",
+        "us-west-2": "health.us-east-1.amazonaws.com",
+        "cn-north-1": "health.cn-northwest-1.amazonaws.com.cn",
+        "cn-northwest-1": "health.cn-northwest-1.amazonaws.com.cn",
+        "aws-cn-global": "health.cn-northwest-1.amazonaws.com.cn",
+    },
+}
+
+REGION_TO_PARTITION_OVERRIDE = {
+    "aws-global": "aws",
+    "aws-cn-global": "aws-cn",
+}
+
+
+def register_override_ssl_common_name(cli):
+    cli.register_last(
+        "before-building-argument-table-parser", update_endpoint_url
+    )
+
+
+def update_endpoint_url(session, parsed_globals, **kwargs):
+    service = parsed_globals.command
+    endpoints = SSL_COMMON_NAMES.get(service)
+    region = session.get_config_variable("region")
+    # only change url if user has not overridden already themselves
+    if endpoints is not None and parsed_globals.endpoint_url is None:
+        endpoint_url = endpoints.get(region)
+        if endpoint_url is not None:
+            parsed_globals.endpoint_url = f"https://{endpoint_url}"
+            if service == "health":
+                _override_health_region(region, session, parsed_globals)
+
+
+def _override_health_region(region, session, parsed_globals):
+    if region in REGION_TO_PARTITION_OVERRIDE:
+        partition = REGION_TO_PARTITION_OVERRIDE[region]
+    else:
+        partition = session.get_partition_for_region(region)
+    if partition == "aws-cn":
+        parsed_globals.region = "cn-northwest-1"
+    else:
+        parsed_globals.region = "us-east-1"
--- a/tests/functional/test_override_ssl_common_name.py
+++ b/tests/functional/test_override_ssl_common_name.py
@@ -1,0 +1,64 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from awscli.testutils import BaseAWSCommandParamsTest
+from awscli.customizations.overridesslcommonname import SSL_COMMON_NAMES
+
+
+class OverrideSSLCommonNameTestCase(BaseAWSCommandParamsTest):
+    def _common_name_test_cases(self):
+
+        service_ops = {
+            "sqs": "list-queues",
+            "emr": "list-clusters",
+            "rds": "describe-db-clusters",
+            "neptune": "describe-db-clusters",
+            "docdb": "describe-db-clusters",
+        }
+        for service, operation in service_ops.items():
+            for region in SSL_COMMON_NAMES[service]:
+                yield (service, operation, region)
+
+    def _assert_region_endpoint_used(
+        self, expected_region, expected_endpoint_url
+    ):
+        self.assertEqual(
+            expected_region, self.last_request_dict["context"]["client_region"]
+        )
+        self.assertEqual(
+            expected_endpoint_url,
+            self.last_request_dict["url"],
+        )
+
+    def test_set_endpoint_url_arg(self):
+        for service, operation, region in self._common_name_test_cases():
+            self.run_cmd(f"{service} {operation} --region {region}".split())
+            expected_endpoint_url = (
+                f"https://{SSL_COMMON_NAMES[service][region]}/"
+            )
+            self._assert_region_endpoint_used(region, expected_endpoint_url)
+
+    def test_override_health_endpoint_and_region(self):
+        expected_override_region = "us-east-1"
+        health_regions = SSL_COMMON_NAMES["health"]
+        for region in health_regions:
+            if region in ["cn-north-1", "cn-northwest-1", "aws-cn-global"]:
+                expected_override_region = "cn-northwest-1"
+            expected_endpoint_url = (
+                f"https://{SSL_COMMON_NAMES['health'][region]}/"
+            )
+            self.run_cmd(f"health describe-events --region {region}".split())
+            self._assert_region_endpoint_used(
+                expected_override_region,
+                expected_endpoint_url,
+            )
--- a/tests/unit/customizations/test_overridesslcommonname.py
+++ b/tests/unit/customizations/test_overridesslcommonname.py
@@ -1,0 +1,75 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from awscli.testutils import create_clidriver
+from awscli.customizations.overridesslcommonname import (
+    update_endpoint_url,
+    SSL_COMMON_NAMES,
+)
+
+import pytest
+import argparse
+
+
+def parameters():
+    for service, regions in SSL_COMMON_NAMES.items():
+        for region in regions:
+            yield (service, region)
+
+
+@pytest.fixture
+def parsed_globals():
+    pg = argparse.Namespace()
+    pg.endpoint_url = None
+    pg.region = None
+    return pg
+
+
+@pytest.fixture
+def session():
+    driver = create_clidriver()
+    return driver.session
+
+
+@pytest.mark.parametrize("service,region", parameters())
+def test_update_endpoint_url(parsed_globals, session, service, region):
+    parsed_globals.command = service
+    session.set_config_variable("region", region)
+    update_endpoint_url(session, parsed_globals)
+    assert parsed_globals.endpoint_url == (
+        f"https://{SSL_COMMON_NAMES[service][region]}"
+    )
+
+
+@pytest.mark.parametrize("service,region", parameters())
+def test_url_modified_from_event(parsed_globals, session, service, region):
+    assert parsed_globals.endpoint_url is None
+    parsed_globals.command = service
+    session.set_config_variable("region", region)
+    session.emit(
+        f"before-building-argument-table-parser.{service}",
+        args=[],
+        session=session,
+        argument_table={},
+        parsed_globals=parsed_globals,
+    )
+    assert parsed_globals.endpoint_url == (
+        f"https://{SSL_COMMON_NAMES[service][region]}"
+    )
+
+
+def test_dont_modify_provided_url(parsed_globals, session):
+    parsed_globals.endpoint_url = "http://test.com"
+    parsed_globals.command = "sqs"
+    update_endpoint_url(session, parsed_globals)
+    assert parsed_globals.endpoint_url == "http://test.com"
--- a/awscli/customizations/rds.py
+++ b/awscli/customizations/rds.py
@@ -96,8 +96,10 @@ class GenerateDBAuthTokenCommand(BasicCommand):
 
     def _run_main(self, parsed_args, parsed_globals):
         rds = self._session.create_client(
-            'rds', parsed_globals.region, parsed_globals.endpoint_url,
-            parsed_globals.verify_ssl
+            'rds',
+            region_name=parsed_globals.region,
+            endpoint_url=parsed_globals.endpoint_url,
+            verify=parsed_globals.verify_ssl
         )
         token = rds.generate_db_auth_token(
             DBHostname=parsed_args.hostname,
--- a/awscli/handlers.py
+++ b/awscli/handlers.py
@@ -91,6 +91,7 @@ from awscli.customizations.s3events import register_event_stream_arg
 from awscli.customizations.sessionmanager import register_ssm_session
 from awscli.customizations.sms_voice import register_sms_voice_hide
 from awscli.customizations.dynamodb import register_dynamodb_paginator_fix
+from awscli.customizations.overridesslcommonname import register_override_ssl_common_name
 
 
 def awscli_initialize(event_handlers):
@@ -183,3 +184,4 @@ def awscli_initialize(event_handlers):
     register_ssm_session(event_handlers)
     register_sms_voice_hide(event_handlers)
     register_dynamodb_paginator_fix(event_handlers)
+    register_override_ssl_common_name(event_handlers)


awscli/clidriver.py
```diff
@@ -484,7 +484,7 @@ def __call__(self, args, parsed_globals):
         event = 'before-building-argument-table-parser.%s.%s' % \
             (self._parent_name, self._name)
         self._emit(event, argument_table=self.arg_table, args=args,
-                   session=self._session)
+                   session=self._session, parsed_globals=parsed_globals)
```
dlm6693, line 487: after digging into the code, it made more sense to add the handler during the `before-building-argument-table-parser` event over `building-command-table` because of the availability of `parsed_globals` in the function call. It could be done but would require more changes to the code by passing around the variable to a couple of different functions or setting it as an instance variable.

awscli/customizations/commands.py
```diff
@@ -134,7 +134,7 @@ def __call__(self, args, parsed_globals):
         event = 'before-building-argument-table-parser.%s' % \
             ".".join(self.lineage_names)
         self._session.emit(event, argument_table=self._arg_table, args=args,
-                           session=self._session)
+                           session=self._session, parsed_globals=parsed_globals)
```
dlm6693, line 137: since `update_endpoint_url` runs every time this event is emitted, `parsed_globals` needs to be passed here as well for the command not to fail/return a 255

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,93 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+import os
+
+DEFAULT_DNS_SUFFIX = 'amazonaws.com'
+CHINA_DNS_SUFFIX = 'amazonaws.com.cn'
+GOV_DNS_SUFFIX = 'sc2s.sgov.gov'
+
+common_regions = [
```
stealthycoin, line 20: Where did this list come from?

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,93 @@
...
+        and parsed_globals.endpoint_url is None
+    ):
+        region = parsed_globals.region
+        # only resolve region if user has not provided it in the command
+        if region is None:
+            region = (
+                session.get_config_variable('region')
+                or os.environ.get('AWS_DEFAULT_REGION')
+            )
+        dnsSuffix = service_regions.get(region)
```
stealthycoin, line 89: same here `dns_suffix`

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,74 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from awscli.testutils import mock, unittest
+from awscli.testutils import BaseAWSCommandParamsTest
+from awscli.customizations.overridesslcommonname import update_endpoint_url
+
+
+class TestOverrideSslCommonName(BaseAWSCommandParamsTest):
```
stealthycoin, line 18: Since it is a known finite set of test cases, we should make a parametrized functional tests that goes through every case using: https://docs.pytest.org/en/6.2.x/parametrize.html

stealthycoin (changes requested):

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @mock.patch('awscli.clidriver.CLIOperationCaller.invoke', return_value=0)
+    def test_set_endpoint_url_arg(self, invoke):
+        self.run_cmd(self.prefix.split())
+        call_args = invoke.call_args[0]
+        self.assertEqual(call_args[0], 'sqs')
+        self.assertEqual(call_args[1], 'ListQueues')
+        self.assertEqual(call_args[3].endpoint_url, 'https://sqs.us-east-1.amazonaws.com')
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
dlm6693: ah bad copy paste on my part from `test_cloudsearchdomain.py`. Looks like its an artifact from 7 years ago or so. I'll remove from this file, but guessing it should be removed from others as well?

awscli/customizations/overridesslcommonname.py
dlm6693: i had to manually put it together by combing through the `endpoints.json` file. Obviously far from ideal and mistake prone, but I'm not sure there's a better way. I suppose I could put a script together that runs through every region/service combo and compile a list of ones that use SSL common name. I can work on that to validate.

dlm6693: ok so I compiled a list programmatically (can share code if desired) to come up with a full list of services and regions that have an SSL common name. This doesn't account for ones that do have the correct format `{service}.{region}.{dnsSuffix}` which I tried to exclude.  Do we want to just override them all? List below with `service` being the key and `(region, endpoint_url)` being the value:

```txt
rds: (us-east-1, rds.amazonaws.com)
elasticmapreduce: (af-south-1, af-south-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-east-1, ap-east-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-northeast-1, ap-northeast-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-northeast-2, ap-northeast-2.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-northeast-3, ap-northeast-3.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-south-1, ap-south-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-southeast-1, ap-southeast-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-southeast-2, ap-southeast-2.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ap-southeast-3, ap-southeast-3.elasticmapreduce.amazonaws.com)
elasticmapreduce: (ca-central-1, ca-central-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (eu-central-1, elasticmapreduce.eu-central-1.amazonaws.com)
elasticmapreduce: (eu-north-1, eu-north-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (eu-south-1, eu-south-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (eu-west-1, eu-west-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (eu-west-2, eu-west-2.elasticmapreduce.amazonaws.com)
elasticmapreduce: (eu-west-3, eu-west-3.elasticmapreduce.amazonaws.com)
elasticmapreduce: (me-south-1, me-south-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (sa-east-1, sa-east-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (us-east-1, elasticmapreduce.us-east-1.amazonaws.com)
elasticmapreduce: (us-east-2, us-east-2.elasticmapreduce.amazonaws.com)
elasticmapreduce: (us-west-1, us-west-1.elasticmapreduce.amazonaws.com)
elasticmapreduce: (us-west-2, us-west-2.elasticmapreduce.amazonaws.com)
rds: (us-east-1, rds.amazonaws.com)
rds: (us-east-1, rds.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
shield: (aws-global, shield.us-east-1.amazonaws.com)
sqs: (af-south-1, af-south-1.queue.amazonaws.com)
sqs: (ap-east-1, ap-east-1.queue.amazonaws.com)
sqs: (ap-northeast-1, ap-northeast-1.queue.amazonaws.com)
sqs: (ap-northeast-2, ap-northeast-2.queue.amazonaws.com)
sqs: (ap-northeast-3, ap-northeast-3.queue.amazonaws.com)
sqs: (ap-south-1, ap-south-1.queue.amazonaws.com)
sqs: (ap-southeast-1, ap-southeast-1.queue.amazonaws.com)
sqs: (ap-southeast-2, ap-southeast-2.queue.amazonaws.com)
sqs: (ap-southeast-3, ap-southeast-3.queue.amazonaws.com)
sqs: (ca-central-1, ca-central-1.queue.amazonaws.com)
sqs: (eu-central-1, eu-central-1.queue.amazonaws.com)
sqs: (eu-north-1, eu-north-1.queue.amazonaws.com)
sqs: (eu-south-1, eu-south-1.queue.amazonaws.com)
sqs: (eu-west-1, eu-west-1.queue.amazonaws.com)
sqs: (eu-west-2, eu-west-2.queue.amazonaws.com)
sqs: (eu-west-3, eu-west-3.queue.amazonaws.com)
sqs: (me-south-1, me-south-1.queue.amazonaws.com)
sqs: (sa-east-1, sa-east-1.queue.amazonaws.com)
sqs: (us-east-1, queue.amazonaws.com)
sqs: (us-east-2, us-east-2.queue.amazonaws.com)
sqs: (us-west-1, us-west-1.queue.amazonaws.com)
sqs: (us-west-2, us-west-2.queue.amazonaws.com)
sqs: (cn-north-1, cn-north-1.queue.amazonaws.com.cn)
sqs: (cn-northwest-1, cn-northwest-1.queue.amazonaws.com.cn)
sqs: (us-gov-west-1, us-gov-west-1.queue.amazonaws.com)
sqs: (us-isob-east-1, us-isob-east-1.queue.sc2s.sgov.gov)

File: tests/unit/customizations/test_overridesslcommonname.py (hunk 5 above)
Comment (dlm6693) [reply tests/unit/customizations/test_overridesslcommonname.py:18]: sure I'll update

File: awscli/customizations/overridesslcommonname.py (hunk 3 above)
Comment (nateprewitt) [reply awscli/customizations/overridesslcommonname.py:20]: Yeah, we definitely need to make sure the list is comprehensive if we're going with this approach. Can we be explicit and just generate all possible combinations for the 3 services. That can either live here or in a separate lookup json file.

i.e.
```json
{
   "elasticmapreduce": {
     "af-south-1": "af-south-1.elasticmapreduce.amazonaws.com",
     ...
   },{
   "sqs": {
     ...
   },
   ...
}
```

We also need to consider that not all regions are in a services endpoints.json entry, so let's make sure the region list is complete.

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,93 @@
...
+        if dnsSuffix is not None:
+            service_data = session.get_service_data(service)
+            endpointPrefix = service_data['metadata']['endpointPrefix']
+            parsed_globals.endpoint_url = f"https://{endpointPrefix}.{region}.{dnsSuffix}"
```
nateprewitt, line 93: Is this right? Aren't almost all of the entries supposed to be region.endpointPrefix.dnsSuffix?

I think the idea here was to hardcode most of this so we're not relying on the session anymore. If we went with this approach we could just pull the endpoint resolver off the session and all this code would go away.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_other_dns_suffixes(self):
+        self.parsed_globals.region = 'cn-north-1'
+        update_endpoint_url(self.driver.session, self.parsed_globals)
+        self.assertEqual(self.parsed_globals.endpoint_url, 'https://sqs.cn-north-1.amazonaws.com.cn')
+
+        self.parsed_globals.endpoint_url = None
+        self.parsed_globals.region = 'us-isob-east-1'
+        update_endpoint_url(self.driver.session, self.parsed_globals)
+        self.assertEqual(self.parsed_globals.endpoint_url, 'https://sqs.us-isob-east-1.sc2s.sgov.gov')
```
nateprewitt, line 50: This also isn't a valid endpoint. Can we take a careful look at these tests to make sure they're validating the expected behavior not the implemented behavior.

nateprewitt (changes requested):
Can we take a closer look at the endpoints entries? It looks like we're creating unroutable URIs with this change.

awscli/customizations/overridesslcommonname.py
dlm6693: no that's what `sslCommonName` is doing for the most part currently. I was under the impression we wanted to update all services/regions using `{region}.{endpointPrefix/service}.{dnsSuffix}` to `{endpointPrefix/service}.{region}.{dnsSuffix}`. 

And sure I can remove the call to grab the service data and hard code it in the config mentioned above.

nateprewitt: I think the point of this PR was to preserve all the sslCommonName entries for the CLI. That allows us to remove the code path in Botocore.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,74 @@
...
+        self.parsed_globals.endpoint_url = 'https://test.com'
+        self.parsed_globals.region = 'us-west-1'
+        update_endpoint_url(self.driver.session, self.parsed_globals)
+        self.assertEqual(self.parsed_globals.endpoint_url, 'https://test.com')
+        self.assertEqual(self.parsed_globals.region, 'us-west-1')
+
+    def test_other_dns_suffixes(self):
+        self.parsed_globals.region = 'cn-north-1'
+        update_endpoint_url(self.driver.session, self.parsed_globals)
+        self.assertEqual(self.parsed_globals.endpoint_url, 'https://sqs.cn-north-1.amazonaws.com.cn')
```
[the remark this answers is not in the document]
dlm6693: This is following the `{service}.{region}.{dnsSuffix}` convention that I thought was supposed to be applied to all endpoints. Is that not the case? As for `sqs` vs. `queue`, I was just grabbing the endpoint prefix from the service model, but since it seems arbitrary for this service, I'll hardcode for each region.

tests/unit/customizations/test_overridesslcommonname.py
dlm6693: I'm not sure I understand. This is testing expected behavior. All we're doing is updating the endpoint url so I'm not sure what else I should be validating here. I could write integration tests that makes sure every call with a generated url is successful.

awscli/customizations/overridesslcommonname.py
dlm6693: just pushed up some updates. every service/region combination listed above is now included and hardcoded as a dictionary in the file

awscli/testutils.py
```diff
@@ -99,7 +99,7 @@ def get_aws_cmd():
     import awscli
     if AWS_CMD is None:
         # Try <repo>/bin/aws
-        repo_root = os.path.dirname(os.path.abspath(awscli.__file__))
+        repo_root = os.path.dirname(os.path.dirname(os.path.abspath(awscli.__file__)))
```
nateprewitt, line 102: Why did this need to change?

tests/functional/test_override_ssl_common_name.py
```diff
@@ -0,0 +1,57 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from awscli.testutils import mock, create_clidriver
+import pytest
+
+from awscli.customizations.overridesslcommonname import ssl_common_names
+
+regions_to_exclude = [
```
nateprewitt, line 19: What was the reason for excluding some of the regions?

Another approach may be to map services to the operation and grab everything.

````python
def common_name_test_cases():

    service_ops = {
        "sqs": "list-queues",
        "emr": "list-clusters" ,
        "shield": "list-protections" ,
        "rds": "describe-db-clusters",
    }
    for service in ssl_common_names.keys():
        for region in ssl_common_names[service]:
            yield (service, service_ops[service], region)

File: tests/functional/test_override_ssl_common_name.py [hunk 12]
```diff
@@ -0,0 +1,57 @@
...
+    for region in ssl_common_names["emr"]
+    if region not in regions_to_exclude
+]
+shield = [
+    ("shield", "list-protections", region) for region in ssl_common_names["shield"]
+]
+rds = [("rds", "describe-db-clusters", region) for region in ssl_common_names["rds"]]
+test_cases = sqs + emr + shield + rds
+
+driver = create_clidriver()
```
nateprewitt, line 46: Can we make this a [fixture](https://docs.pytest.org/en/6.2.x/fixture.html) so it can be shared among tests without any concerns for parallelization?

tests/functional/test_override_ssl_common_name.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@mock.patch("awscli.clidriver.CLIOperationCaller.invoke", return_value=0)
```
nateprewitt, line 49: Let's move this up to the fixture so it's localized.

tests/integration/customizations/test_override_ssl_common_name.py
```diff
@@ -0,0 +1,58 @@
...
+def test_update_endpoint_url(service, operation, region):
+
+    p = aws(f"{service} {operation} --region {region}")
+    assert p.rc == 0
+    if service == 'emr':
+        assert p.json == {'Clusters': []}
+    if service == 'shield':
+        assert p.json == {'Protections': []}
+    if service == 'rds':
+        assert p.json == {'DBClusters': []}
```
nateprewitt, line 58: This has a success case where we end up with an unknown service and don't make any assertions. Those kind of edge cases can lead to the test subtly changing, resulting in nothing being checked. We're also not asserting anything about SQS.

I also don't know if this will pass in an actual account. As soon as any resources are added, these tests will fail. I'd probably bias against doing integration tests for these, but if we decide to, let's make a more general assertion that we got a valid JSON response and it was not empty.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,67 @@
...
+    for region in regions
+]
+
+
+def _setup_test():
+    parsed_globals = mock.Mock()
+    parsed_globals.endpoint_url = None
+    parsed_globals.region = None
+    driver = create_clidriver()
+    return parsed_globals, driver.session
```
nateprewitt, line 35: We can use a fixture for this as well.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,67 @@
...
+    parsed_globals.command = service
+    session.set_config_variable("region", region)
+    session.emit(
+        "before-building-argument-table-parser.sqs",
+        args=[],
+        session=session,
+        argument_table=mock.Mock(),
+        parsed_globals=parsed_globals,
+    )
+    assert parsed_globals.endpoint_url == f"https://{ssl_common_names[service][region]}"
```
nateprewitt, line 59: Lets assert the endpoints_url is either empty or doesn't exist before emitting.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_dont_modify_provided_url():
+    parsed_globals, session = _setup_test()
+    parsed_globals.endpoint_url = "https://test.com"
+    parsed_globals.command = "sqs"
+    update_endpoint_url(session, parsed_globals)
+    assert parsed_globals.endpoint_url == "https://test.com"
```
nateprewitt, line 67: Let's us a non-https url to ensure we're not changing the scheme.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,67 @@
...
+        argument_table=mock.Mock(),
+        parsed_globals=parsed_globals,
+    )
+    assert parsed_globals.endpoint_url == f"https://{ssl_common_names[service][region]}"
+
+
+def test_dont_modify_provided_url():
+    parsed_globals, session = _setup_test()
+    parsed_globals.endpoint_url = "https://test.com"
+    parsed_globals.command = "sqs"
```
nateprewitt, line 65: Why are we setting the command for this one but not the others?

nateprewitt (changes requested):

awscli/testutils.py
dlm6693: don't need it! was messing around with a few things and left it in accidentally. will remove

tests/functional/test_override_ssl_common_name.py
dlm6693: will do!

tests/integration/customizations/test_override_ssl_common_name.py
dlm6693: I omitted SQS because I was getting back an empty string in the `stdout` which won't translate to valid json. i'll see if i can find another op that returns valid json even if there are no results.

Maybe I can add the keys to the test case tuples and check that its in the json?

tests/unit/customizations/test_overridesslcommonname.py
dlm6693: the command is set for the other tests. the tests wouldn't pass otherwise

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,89 @@
...
+        "eu-west-3": "eu-west-3.elasticmapreduce.amazonaws.com",
+        "me-south-1": "me-south-1.elasticmapreduce.amazonaws.com",
+        "sa-east-1": "sa-east-1.elasticmapreduce.amazonaws.com",
+        "us-east-1": "elasticmapreduce.us-east-1.amazonaws.com",
+        "us-east-2": "us-east-2.elasticmapreduce.amazonaws.com",
+        "us-west-1": "us-west-1.elasticmapreduce.amazonaws.com",
+        "us-west-2": "us-west-2.elasticmapreduce.amazonaws.com",
+    },
+    "shield": {
+        "us-east-1": "shield.us-east-1.amazonaws.com",
```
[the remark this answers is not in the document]
dlm6693: I checked out my `deprecate-ssl-common-name` feature branch in botocore and created a shield client. I'm getting the same url:
```python
from botocore.session import get_session
session = get_session()
shield = session.create_client('shield', 'us-east-1')
client._endpoint
shield('https://shield.us-east-1.amazonaws.com')

File: tests/integration/customizations/test_override_ssl_common_name.py (hunk 14 above)
Comment (dlm6693) [reply tests/integration/customizations/test_override_ssl_common_name.py:58]: there also don't seem to be any other operations for sqs that don't require an existing resources besides ones that create them. i could use one of those and add a cleanup function at the end

File: tests/functional/test_override_ssl_common_name.py (hunk 13 above)
Comment (dlm6693) [reply tests/functional/test_override_ssl_common_name.py]: it looks like moving the mock to the fixture actually de-localized it. i'm going to move back to the decorator so i don't have to call `start` and `stop`

File: awscli/customizations/overridesslcommonname.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (nateprewitt) [reply awscli/customizations/overridesslcommonname.py]: Yeah, that's what I'd assumed. We won't need the shield entry then. Thanks for checking!

File: tests/functional/test_override_ssl_common_name.py (hunk 13 above)
Comment (nateprewitt) [reply tests/functional/test_override_ssl_common_name.py:57]: On first read, this had looked like we were mocking out components of the driver. If this is a separate class, it's fine to keep it as a decorator.

**Review by nateprewitt — APPROVED** [review]
Looks good to me ⛵

File: tests/unit/customizations/test_overridesslcommonname.py [hunk 20]
```diff
@@ -0,0 +1,71 @@
...
+test_cases = [
+    (service, region)
+    for service, regions in ssl_common_names.items()
+    for region in regions
+]
+
+
+@pytest.fixture
+def parsed_globals():
+    pg = mock.Mock()
```
kyleknap, line 32: Instead of using `Mock` here, let's use `argparse.Namespace()`. The main concern in using `Mock()` without the spec argument is that it can swallow potential issues in the code as it happily accepts any property/method calls and this can lead to issues in updating the code down the road.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.mark.parametrize("service,region", test_cases)
+def test_url_modified_from_event(parsed_globals, session, service, region):
+    assert parsed_globals.endpoint_url is None
+    parsed_globals.command = service
+    session.set_config_variable("region", region)
+    session.emit(
+        "before-building-argument-table-parser.sqs",
+        args=[],
+        session=session,
+        argument_table=mock.Mock(),
```
kyleknap, line 61: For `argument_table` let's just set it to an empty dictionary.

kyleknap (commented):
Looks good. I just had some of suggestions and questions

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,84 @@
...
+        "us-west-2": "us-west-2.elasticmapreduce.amazonaws.com",
+    },
+    "rds": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+}
+
+
+def register_override_ssl_common_name(cli):
+    cli.register_last("before-building-argument-table-parser", update_endpoint_url)
```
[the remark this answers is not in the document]
dlm6693: sure thing. is `flake8` or `black` preferable? I have `pre-commit` on my machine but just for the python sdk repos

tests/unit/customizations/test_overridesslcommonname.py
dlm6693: yeah makes sense! i think i did this based off of other tests i saw in the code base, but agreed it would be good to use a more realistic that isn't magic and can accept anything as a property/attribute/method

awscli/customizations/commands.py
```diff
@@ -134,7 +134,7 @@ def __call__(self, args, parsed_globals):
         event = 'before-building-argument-table-parser.%s' % \
             ".".join(self.lineage_names)
         self._session.emit(event, argument_table=self._arg_table, args=args,
-                           session=self._session)
```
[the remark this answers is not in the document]
dlm6693: I know we had originally agreed upon `building-command-table`, but upon diving into the code, this seemed by far to be the less invasive approach. I do see now that it runs twice, once for the service and then for the operation, which is definitely behavior that should be avoided even if the handler is idempotent.

The trouble is figuring out a clean implementation of passing `parsed_globals` to that event. The `ServiceOperation` object only gets that variable when its invoked in a `ServiceCommand` object's `__call__` method, which is after `building-command-table` has run. I think its definitely doable, but would require a larger refactor, which I wanted to avoid for something that is pretty core behavior to the CLI.

The easiest way that is occurring to me now would be to pass `parsed_globals` to `_get_command_table`, which would pass it in turn to `_create_command_table`. Let me know what you think of that approach and if you have any concerns with it. Happy to investigate further, but if you do come up with any suggestions let me know.

tests/functional/test_override_ssl_common_name.py
```diff
@@ -0,0 +1,46 @@
...
+        "emr": "list-clusters",
+        "rds": "describe-db-clusters",
+    }
+    for service in ssl_common_names:
+        for region in ssl_common_names[service]:
+            yield (service, service_ops[service], region)
+
+
+@pytest.mark.parametrize("service,operation,region", common_name_test_cases())
+@mock.patch("awscli.clidriver.CLIOperationCaller.invoke", return_value=0)
```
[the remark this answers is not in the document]
dlm6693: Ok sounds good thanks Kyle. I'll take a look and follow up if I have any questions.

awscli/customizations/overridesslcommonname.py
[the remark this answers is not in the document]
kyleknap: Let's use `flake8`. The repository is not compatible with `black` as of right now and that will change a lot of lines.

tests/functional/test_override_ssl_common_name.py
```diff
@@ -0,0 +1,44 @@
...
+    def _common_name_test_cases(self):
+
+        service_ops = {
+            "sqs": "list-queues",
+            "emr": "list-clusters",
+            "rds": "describe-db-clusters",
+        }
+        for service, regions in SSL_COMMON_NAMES.items():
+            for region in regions:
+                yield (service, service_ops[service], region)
```
dlm6693, line 28: probably doesn't matter for this. but i got a warning from `pytest`:
```bash
PytestCollectionWarning: yield tests were removed in pytest 4.0 - test_cases will be ignored
```

awscli/customizations/commands.py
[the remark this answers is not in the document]
kyleknap: I think originally I had two different potential events in mind:

* `top-level-args-parsed` - This is the event for when the global args get parsed. It seemed suiting as we would just immediately set the global argument of `--endpoint-url` at that stage and we can keep endpoint url mutations isolated to that stage. The only issue with that approach is that the session had technically not been initialized yet, so there may be unintended consequences/configuration resolution if we try to grab the region before that. This event also happens before even logging is enabled so that is a concern as well if we are trying to do any tracing of it.
* `session-initialized` - This seemed like the next best event to use as the session was initialized there. The only problem I saw in using this event is that we could not 100% capture correctly the service being used. Specifically, for [CLI aliases](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-alias.html) the command that gets parsed is represented as the alias and not the underlying service command at that point. So we would be using the wrong endpoint if the command was under an alias.

This all being said. I don't know if I have a clear preference between `building-command-table` and `before-building-argument-table-parser`. They both share similar concerns, but it seems like we need an event lower into the stack to get this to work appropriately and I do not want to be adding a new event for this. I think for now let's just stick with `before-building-argument-table-parser` unless we get a better idea.

tests/functional/test_override_ssl_common_name.py
kyleknap: I wonder if it is the name `_common_name_test_cases`  that is triggering it? Like if we remove `test` from the method. It would be nice to not have the warning to make the py.test output cleaner.

dlm6693: yup that did it! good call

awscli/customizations/commands.py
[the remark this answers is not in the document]
dlm6693: Ok cool sounds good Kyle. I'll keep the behavior as is. I'll be adding the additional services that were missed in my script that you caught in yours shortly.

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,122 @@
...
+        "eu-south-1": "health.us-east-1.amazonaws.com",
+        "eu-west-1": "health.us-east-1.amazonaws.com",
+        "eu-west-2": "health.us-east-1.amazonaws.com",
+        "eu-west-3": "health.us-east-1.amazonaws.com",
+        "me-south-1": "health.us-east-1.amazonaws.com",
+        "sa-east-1": "health.us-east-1.amazonaws.com",
+        "us-east-1": "health.us-east-1.amazonaws.com",
+        "us-east-2": "health.us-east-1.amazonaws.com",
+        "us-west-1": "health.us-east-1.amazonaws.com",
+        "us-west-2": "health.us-east-1.amazonaws.com",
```
kyleknap, line 100: So when running the CLI manually with these hard coded endpoints, I'm getting the following error for non-us-east-1 regions when using `health`:
```
$ aws health describe-events --region us-west-2

An error occurred (InvalidSignatureException) when calling the DescribeEvents operation: Credential should be scoped to a valid region. 
```

It looks like the reason this is happening is that `health` is defined as a non-regionalized service and thus has a default credentialScope of `us-east-1` for the `aws` partition. We are going to have to adjust the logic here so that we also set the region to the default credential scope for `health` as well as the endpoint url. I would not worry about generalizing it too much. It should be just:

If we find we need to set the endpoint url for health:

1. Call the `session.get_partition_for_region()` for the provided region to determine the partition we are working in.
2. Map the partition to the value representing the region of the default credential scope for the partition (i.e. determine if the region we need to set should be `us-east-1`, `cn-northwest-1`)
3. Set the region either on the `parsed_globals` or via the `session.set_config_variable()` method, whatever does the job of making sure we use the correct credential scope for the region.

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,122 @@
...
+        "cn-north-1": "health.cn-northwest-1.amazonaws.com.cn",
+        "cn-northwest-1": "health.cn-northwest-1.amazonaws.com.cn",
```
kyleknap, line 102: We should also make sure to add `aws-cn-global` to this list. It appears to be a valid pseudo-region value respected for `health` that I did not include in the script for the initial audit. I don't think in of the the other services need it though.

kyleknap (commented):
It's looking good. I mainly focused on looking through the endpoint mapping to make sure they were valid.

tests/unit/customizations/test_overridesslcommonname.py
```diff
@@ -0,0 +1,71 @@
...
+    assert parsed_globals.endpoint_url == f"https://{SSL_COMMON_NAMES[service][region]}"
+
+
+@pytest.mark.parametrize("service,region", parameters())
+def test_url_modified_from_event(parsed_globals, session, service, region):
+    assert parsed_globals.endpoint_url is None
+    parsed_globals.command = service
+    session.set_config_variable("region", region)
+    session.emit(
+        "before-building-argument-table-parser.sqs",
```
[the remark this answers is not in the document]
dlm6693: definitely not a nit! thanks for catching

awscli/customizations/overridesslcommonname.py
```diff
@@ -0,0 +1,122 @@
...
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "docdb": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "neptune": {
+        "us-east-1": "rds.amazonaws.com",
+    },
+    "health": {
+        "aws-global": "health-us-east-1.amazonaws.com",
```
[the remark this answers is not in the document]
dlm6693: oof this was super careless. sorry about that :/

awscli/customizations/overridesslcommonname.py
dlm6693: alright i shot up a fix. i was getting the same error before making the fix, but a different one. not sure how to handle it since it was a subscription error but as far as i can tell there are no subscriptions in aws health.

kyleknap (commented):
Looks great! Can you squash all of these commits into a single commit? After, I'll go ahead and approve and merge the PR.

kyleknap (approved):
Thanks! :ship:

stealthycoin (changes requested):
The code looks a lot better now, I think the one main thing I want to se updated before we merge is the commit title/message. This is a fairly involved change and we should document some of the conversations we had in here in the commit.

Think of someone trying to figure out why the code does what it does, and historically why we chose to do what we did. For example if someone was looking at the session.emit line from here and wondering why we added `parsed_globals`  as an argument to that event and they check the git blame it just says "deprecate ssl common name" which is not helpful. You don't need to copy and paste all of the conversations into the commit, but I would like to see a few paragraphs outlining some of the important design considerations and consequences of this change.

stealthycoin (commented):
A changelog entry would be nice as well. I think this change is a little too involved to be omitted even though it shouldn't have any consequences.

dlm6693:
@stealthycoin sure thing. i'll have something ready in a bit.

auditendpoints.py
```diff
@@ -0,0 +1,37 @@
+import json
```
kyleknap, line 1: I'm not sure if this file should have been added to the PR.

.changes/next-release/enhancement-Endpoints-23278.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Endpoints",
+  "description": "Allow usage of SSL common name with deprecation"
```
kyleknap, line 4: For the description, let's update it to: "Enforce SSL common name as default for endpoint url". Mainly the wording is a bit more consistent with the change we are making as "Allow usage" is more indicative of adding an opt-in option or loosening a constraint.

kyleknap (commented):
Just had a couple more comments after the recent updates.

auditendpoints.py
dlm6693: Oops! I thought this was in my global git ignore but I guess it snuck through. Will remove

.changes/next-release/enhancement-Endpoints-23278.json
dlm6693: Got it! Will update.

stealthycoin (approved):

kyleknap (approved):

PR #7118, opened 2022-07-19
Add Examples note

hssyoo:
Generating a note under the `Examples` sections for CLI commands will help users troubleshoot basic problems they come across.

The note:
- Reminds users to have the AWS CLI installed and configured.
- Clarifies that the examples use unix-like quotation rules.

The URLs included in the added note link to v1 docs and when merging this change into the `v2` branch the following links should be used instead:
https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html
https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-quoting-strings.html

### DIFF
--- a/awscli/customizations/addexamples.py
+++ b/awscli/customizations/addexamples.py
@@ -44,6 +44,20 @@ def add_examples(help_command, **kwargs):
     LOG.debug("Looking for example file at: %s", doc_path)
     if os.path.isfile(doc_path):
         help_command.doc.style.h2('Examples')
+        help_command.doc.style.start_note()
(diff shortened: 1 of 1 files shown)


stealthycoin:
Looks good, can I get a screenshot of the before and after to get an idea of what it looks like.

hssyoo:
> Looks good, can I get a screenshot of the before and after to get an idea of what it looks like.

Before:
Screen Shot 2022-07-21 at 10 07 28 AM

After:
Screen Shot 2022-07-21 at 10 09 38 AM

stealthycoin (approved):

PR #7134, opened 2022-07-26
Include global options in command help pages

hssyoo:
We get a lot of requests from users to include global options in a command's help page. The changes in this PR sets a global variable `GLOBAL_ARG_TABLE` once the global argument table is built and passes it to help objects that build their respective help commands. The specific help command's argument table is then added onto the global argument table before the `Options` docs are generated.
The changes are applied to both standard and custom commands.

`s3` Before:
image

`s3` After:
image

`s3 cp` Before:
image

`s3 cp` After:
image
. . .
image


hssyoo:
Closing this after chatting with @kyleknap and going with a different approach of passing the global args table as a parameter.

Issue #4374, opened 2019-08-05
awscli assume role creates tmpfile in $HOME/.aws, this breaks if $HOME/.aws is readonly mounted


pieterza:
If $HOME/.aws is read only mounted:

When using a profile that does an assume role like so:
```
[master]
....

[eu-staging]
role_arn= arn:aws:iam::REDACTED
source_profile=master
region = eu-west-1
```

The command:

`
aws --profile eu-staging ec2 describe-instances
`

Error:

[Errno 30] Read-only file system: '/home/spinnaker/.aws/cli'

Please address the tempfile requirement, as in kubernetes land we mount secrets as folders, and can't make them read-write

stealthycoin:
Can you give me some more details of what your setup is? I am not familiar with kubernetes at all. My concern is will be really inefficient as you'll need to do the assume role call before every single CLI operation. Would an environment variable/config option that allows you to disable caching be sufficient for your needs?

jerrawa:
I have a similar issue. The home directory of a service account does not get mounted, so when the awscli tries to write the $HOME/.aws/cli/cache files I get the following error.
[Errno 13] Permission denied: '/home/username'

I am using the environment variables to redirect the credentials and config files.

For me, an Environment variable to redirect the whole .aws folder to a new location would be a good solution.

pieterza:
@stealthycoin I've dug into the code, and the issue is caused by the caching mechanism of awscli/boto3.
When we mount secrets/configmaps into kubernetes containers, such as the credentials/config to ~/.aws it is done as read-only (by k8s, and not configurable.).

running awscli sometimes creates a ~/.aws/.cache which fails due to read-only filesystem.

pieterza:
Sorry @stealthycoin , to answer your question:  A flag that disables caching would solve this problem, yes.

jhindulak:
@pieterza I ran into the same problem trying to use an AWS Config file with Spinnaker's Clouddriver deployed in K8s. I mounted my config file to `$HOME/.aws` as you would and noticed the secret mounts as root, so the `spinnaker` user doesn't have rights to do anything with the `.aws` directory resulting in the error you mentioned. 

[...]

PR #7135, opened 2022-07-26
Add environment variable to configure assume-role-with-web-identity credential cache directory

jeremydonahue:
*Issue #, if available:*
A description of the problem was discussed in #4374, which was closed by the author because they had a workaround. This addresses the problem directly.

*Description of changes:*
This makes the credential cache directory (default: `~/.aws/cli/cache`) used to store temporary credentials obtained using `AssumeRoleWithWebIdentity` configurable using the environment variable `AWS_CREDENTIAL_CACHE_DIR`. The use case is environments where the home directory is mounted as read-only. The cache directory being configurable means it can be pointed at a writable volume.

### DIFF
--- a/awscli/customizations/assumerole.py
+++ b/awscli/customizations/assumerole.py
@@ -5,7 +5,9 @@ from botocore.exceptions import ProfileNotFound
 from botocore.credentials import JSONFileCache
 
 LOG = logging.getLogger(__name__)
-CACHE_DIR = os.path.expanduser(os.path.join('~', '.aws', 'cli', 'cache'))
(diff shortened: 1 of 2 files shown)


talregev:
Build and run on vm snap package for aws v1.25.47

image

kdaily:
Hi @jeremydonahue,

I added a comment at https://github.com/aws/aws-cli/issues/4374#issuecomment-1251306126. The main concern is that the AWS CLI can write to other directories as well, for example, the AWS SSO cache at `~/.aws/sso/cache`. So, read only mounts would still affect this functionality.

jeremydonahue:
Thanks, I responded there.

PR #7153, opened 2022-08-01
Add a Global Options section to help docs

hssyoo:
We get a lot of requests from users to include global options in a command's help page. The changes in this PR passes the global args table created once by the `CLIDriver` to the help commands that generate help docs. The global args table is used to generate a new `Global Options` section in the help docs.
The changes are currently only applied to standard commands, and does not include customizations.

image

### DIFF
--- a/tests/unit/customizations/test_commands.py
+++ b/tests/unit/customizations/test_commands.py
@@ -121,8 +121,8 @@ class TestBasicCommandHooks(unittest.TestCase):
         driver.main('s3 foo'.split())
 
         self.assert_events_fired_in_order([
-            'building-command-table.main',
             'building-top-level-params',
+            'building-command-table.main',
             'top-level-args-parsed',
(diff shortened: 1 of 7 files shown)


tests/unit/customizations/test_commands.py
```diff
@@ -121,8 +121,8 @@ def test_expected_events_are_emitted_in_order(self):
         driver.main('s3 foo'.split())
 
         self.assert_events_fired_in_order([
-            'building-command-table.main',
             'building-top-level-params',
+            'building-command-table.main',
```
hssyoo, line 125: I had to change the order of these events because the global arg table must be built as part of the `_build_builtin_commands` method which gets run before the session emits a `building-command-table.main` event.

hssyoo:
Closing this in favor of https://github.com/aws/aws-cli/pull/7157

PR #7157, opened 2022-08-02
Include global options using pre-generated rst files

hssyoo:
We get a lot of requests from users to include global options in a command's help page. The changes in this PR include:
- A script used to pre-generate `.rst` files for global options
- The inclusion of the global options in help docs
- A pair of tests to ensure that the generated `.rst` files are up to date

Preview of doc changes
image

image


awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
 
     def doc_subitems_start(self, help_command, **kwargs):
         pass
+
+
+class GlobalsRenderer:
```
kyleknap, line 735: For the name of this class, let's see if we can figure out a different name than `Render`. Mainly that name is generally preserved for rendering generated docs to the shell that the user is using (e.g. the renderer classes in the `help.py` module). How about `GlobalOptionsDocumenter`? I'm open for ideas

awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
...
+    """
```
kyleknap, line 736: For docstrings, we generally try to follow PEP257 specifically for [multiline doc strings](https://peps.python.org/pep-0257/#multi-line-docstrings) where the first line should be a summary of the utility and then any other elaboration can be done in a separate summary below. If I need to double check my docstring, I've had success with [`pydocstyle`](http://www.pydocstyle.org/en/stable/) in the past for quick checks.

awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
...
+    Synopsis and Global Options sections.
+    """
+
+    def __init__(self, driver):
+        self._driver = driver
+        self._global_options = None
+        self._global_synopsis = None
+
+    @property
+    def global_options(self):
```
kyleknap, line 747: Then for the public methods, I think we can stick to two public methods and get rid of the `write_global_options` and `write_global_synopsis` methods:
```python
def doc_global_options():
    ...
def doc_global_synopsis():
    ...
```
These methods will just output the documentation string for both the global options and its synopsis. Then the script can take the output and write it to the desired file. The main reason, I'm suggesting it is:
1. It narrows the public interface and makes this abstraction strictly around generating what the content of these sections should be.
2. It makes it more straightforward to test. We can just call these methods and compare it to the content in the `examples` directory instead of having to publicly expose the properties and compare against those.

awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
...
+    
+    @property
+    def global_synopsis(self):
+        if not self._global_synopsis:
+            self._generate_global_synopsis()
+        return self._global_synopsis
+    
+    def _generate_global_options(self):
+        doc = ReSTDocument()
+        arg_table = self._driver._get_argument_table()
```
kyleknap, line 760: In general, we should avoid accessing private methods (i.e. ones that start with underscore) as there is no compatibility contract on these. Instead, let's try to access the argument table from public interfaces. I think we may be able to use `self._driver.create_help_command().arg_table`?

awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
...
+                    doc.style.li(choice)
+                doc.style.end_ul()
+        self._global_options = doc.getvalue()
+    
+    def _generate_global_synopsis(self):
+        doc = ReSTDocument()
+        arg_table = self._driver._get_argument_table()
+        for arg in arg_table:
+            argument = arg_table.get(arg)
+            option_str = '[%s]' % argument.cli_name
```
kyleknap, line 777: It would be interesting to see if we can reuse this logic from other places where generate the non-global synopsis arguments. Mainly if we ever update that logic, we will want to make sure we are consistent across global and non-global options.

kyleknap (commented):
It's looking good! I definitely like this approach over all the other ones we considered. This first round of feedback was more around usage/style patterns and interfaces. I did not look at the tests/script or look to deeply through the content generated for the various command types, but I'll look at that in the next pass.

awscli/clidocs.py
```diff
@@ -178,6 +188,17 @@ def doc_option(self, arg_name, help_command, **kwargs):
...
+    def doc_global_option(self, help_command, **kwargs):
+        doc = help_command.doc
+        doc.style.h2('Global Options')
+        global_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)),
+            'examples/global/global_options.rst'
+        )
+        with open(global_path, 'r') as global_doc:
+            for line in global_doc:
+                doc.writeln(line)
```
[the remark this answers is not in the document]
hssyoo: 1. I tried this but it ends up writing something like `option1option2option3`. I'm not sure where that extra new line is coming from, my current workaround is to remove `\n` at a later point. `doc.write(f"{option}\n"` would be functionally the same as `doc.writeln`. I'll play around with it and see what makes sense.
2. I agree this would be a useful method for `ReSTDocument`. Will implement.

awscli/clidocs.py
```diff
@@ -716,3 +730,62 @@ def _line_has_tag(self, line):
...
+        for arg in arg_table:
+            argument = arg_table.get(arg)
+            option_str = '[%s]' % argument.cli_name
+            doc.write('%s\n' % option_str)
+        self._global_synopsis = doc.getvalue()
+    
+    def write_global_options(self, dir):
+        filepath = os.path.join(dir, 'global_options.rst')
+        with open(filepath, 'w') as global_doc:
+            for line in self.global_options.decode('utf-8'):
```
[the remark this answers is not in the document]
hssyoo: `ReSTDocument`'s `getvalue()` method encodes the doc in `utf-8`. I could get the unencoded contents from the document object's `self._writes` property, but I believe that would be inconsistent with what you said elsewhere in this PR about not accessing private methods (I assume same applies to private properties).

awscli/clidocs.py
hssyoo: I figured this was an imprecise name. I'm not too opinionated on this, just want to make sure it's consistent with the rest of the codebase. I was also thinking `GlobalOptionsGenerator`. Is `Generator` reserved for anything?

awscli/clidocs.py
hssyoo: I decided not to reuse logic for 2 reasons:
1. For global options, the logic was just copied over from `ProviderDocumentEventHandler`. The top-level global options had its own logic before this PR and that hasn't changed.
2. For global synopsis, the logic in `CLIDocumentEventHandler` is incompatible with global options. `elif argument.cli_name.startswith('--'):` would target all global options (they all start with `--`) but we definitely don't want all global options being processed under this conditional

awscli/bcdoc/restdoc.py
```diff
@@ -102,6 +103,13 @@ def remove_last_doc_string(self):
             start, end = self._last_doc_string
             del self._writes[start:end]
 
+    def write_from_file(self, filename, remove_newlines=False):
+        with open(filename, 'r') as f:
+            for line in f.readlines():
+                if remove_newlines:
+                    line = re.sub('\n', '', line)
```
hssyoo, line 110: Not ecstatic about this, but global options and synopsis were being reacting differently to being written, so I needed to add this `remove_newlines` flag to create outputs without extra newlines.

scripts/document-globals
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
kyleknap, line 1: For the script, let's:

1. Rename the script to `make-global-opts-documentation` to be consistent with the `make` prefix usage in the `scripts` directory.
2. We should make sure the file is executable (e.g. `chmod +x scripts/make-global-opts-documentation`) to be able to execute the file without needing to specify python.

awscli/examples/global_options.rst
```diff
@@ -0,0 +1,34 @@
+``--debug`` (boolean)
+Turn on debug logging.
```
kyleknap, line 2: A bit of a nitpick but it would be nice if there were two newlines between the description and the next global option:
```
``--debug`` (boolean)
Turn on debug logging.

``--endpoint-url`` (string)
Override command's default URL with the given URL.
```
Mainly with all of the newlines reduced to one, it is a little difficult to read because everything is combined together.

awscli/clidocs.py
```diff
@@ -716,3 +721,48 @@ def _line_has_tag(self, line):
...
+class GlobalOptionsDocumenter:
+    """Documenter used to pre-generate global options docs."""
+
+    # Maps a condition to a string format.
+    SYNOPSIS_MAP = {
+        lambda x: x.cli_type_name != 'boolean':
+            lambda x: f"[{x.cli_name} <value>]",
+        lambda x: bool(x):
+            lambda x: f"[{x.cli_name}]"
+    }
```
kyleknap, line 735: As a rule of thumb, we try to avoid using lambdas especially if it is anything complicated mainly because they are difficult to troubleshoot. If we were to go forward with this approach I'd recommend just simplifying it down to:
```python
_SYNOPSIS_TYPE_TO_FORMAT = {
     "boolean": "[{cli_name}]",
     # Add formats for other types
}
```
However, I'd recommend we do not need to generalize the logic in a map for this as we really only need to special case one type and we should just have an if statement for the boolean type.  So something like this later down in the logic:
```python
        for arg in help_command.arg_table.values():
            if arg.cli_type_name == 'boolean':
                arg_synopsis = f"[{x.cli_name}]"
            else:
                arg_synopsis = f"[{arg.cli_name} <value>]"
            help_command.doc.writeln(arg_synopsis)
```

tests/functional/test_globals.py
```diff
@@ -0,0 +1,21 @@
...
+
+class TestGlobalOptionsDocumenter(unittest.TestCase):
+    def setUp(self):
+        self.driver = create_clidriver()
+        self.globals = GlobalOptionsDocumenter(self.driver)
+
+    def test_global_options(self):
+        filepath = os.path.join(EXAMPLES_DIR, 'global_options.rst')
+        self.assertEqual(self.globals.doc_global_options(),
+                         open(filepath).read())
```
kyleknap, line 16: For all `open`s we should make sure to close the file handle to leak file descriptors. So instead we should let's just wrap these with a context manager e.g.:
```python
with open(filepath, 'r') as f:
     self.assertEqual(self.globals.doc_global_synopsis(), f.read())
```

tests/functional/test_globals.py
```diff
@@ -0,0 +1,21 @@
...
+from awscli.testutils import create_clidriver
+from awscli.clidocs import EXAMPLES_DIR, GlobalOptionsDocumenter
+
+
+class TestGlobalOptionsDocumenter(unittest.TestCase):
+    def setUp(self):
+        self.driver = create_clidriver()
+        self.globals = GlobalOptionsDocumenter(self.driver)
+
+    def test_global_options(self):
```
kyleknap, line 13: For the test case name, let's make it a bit more explicit about what property we are asserting to be more explicit what the purpose of the test is. For example, let's name this to something like: `test_doc_global_options_match_saved_content()`. I'm up for suggestions, but mainly want to make sure we are capturing the purpose of this test which is what we generate is the same as what is stored on disk/in the repository.

tests/functional/test_globals.py
```diff
@@ -0,0 +1,21 @@
+import os
```
kyleknap, line 1: In addition to a functional test, it may be worth adding a unit test for the `GlobalOptionsDocumenter` class that would:

* Use the test's own controlled arg table (e.g. not sourced from the `cli.json` file). 
* Assert the content for both the global options and synopsis are correct for arguments that:
   * Are of type string
   * Are of type int
   * Are of type boolean
   * Have enums

The main reason that I'm suggesting adding these tests is that if we ever decide to refactor this logic, we have some basic tests in place that we can update or port refactor to make sure our abstraction is working correctly. For example, if we decided to do away with storing the content to a file and we remove those functional tests, we would not have any (direct or indirect) tests capturing the behavior of what are abstraction should be behaving.

awscli/bcdoc/restdoc.py
kyleknap: Yeah I'm not sure if I'm following why we need the `remove_newlines` logic. I would have expected the logic to be just:
```python
def write_from_file(self, filename):
     with open(filename, 'r') as f:
         self.write(f.read())
```

awscli/clidocs.py
kyleknap: That's fine. I was more referring to point 2 here. Looking through the logic on `CLIDocumentEventHandler` I'm fine with having our own logic. Mainly it would take larger rework to reuse the logic in `CLIDocumentEventHandler`  to work with the top level arguments we generate and would prefer not to touch that logic for now.

kyleknap (commented):
It's looking good. I just had some more comments as we polish this.

awscli/bcdoc/restdoc.py
hssyoo: Yeah one would definitely think that haha. I tried quite hard to avoid doing this but the reason is because after making the rst files for options and synopsis:
1. `write()` documents them without _any_ newlines eg option1option2option3.
2. `writeln()` documents the two files differently. Global options look fine (single newlines), but global synopsis has double newlines for each option.

I think my problem is that the generated rst files weren't originally intended to be human readable in raw format, so a lot of the code that documents things erred on the side of caution and liberally applied newlines (to avoid problem 1). I couldn't find out where the extra newline was coming from for the global synopsis because these files are pre-made and should just work with `write` but it just doesn't

kyleknap: I poked around with this and that rationale seems on track specifically. It looks like `writeln()` is doing a couple of things for us:
1. Adding an extra newline, which guarantees the lines are in a new paragraph when it gets generated into rst and html. 
2. The `writeln` also makes sure we do the appropriate indenting for content. So specifically we need the arguments indented in the synopsis otherwise we generate invalid rst.

I'd say to make this work nicely, let's do the following:

1. Update the helper function to be this:
    ```python
    def write_from_file(self, filename):
        with open(filename, 'r') as f:
            for line in f.readlines():
                self.writeln(line.strip())
    ```
    This will ensure the same number of newlines get added and also guarantee that we have the proper indentation. 

2. Let's go ahead and add the extra newlines in the `global_options.rst` generation logic to add the extra newlines where needed to make sure they are appropriately split up in the help page. So I think we will need to add an extra newline after the `--arg <value>` lines and before the choice enumeration. It makes it less nice to read but this is just intermediary content and the important part is the end content (e.g. html and `help` output).

awscli/clidocs.py
```diff
@@ -179,6 +187,11 @@ def doc_option(self, arg_name, help_command, **kwargs):
         doc.style.dedent()
         doc.style.new_paragraph()
 
+    def doc_global_option(self, help_command, **kwargs):
```
kyleknap, line 190: One thing that I was noticing was that for some of the custom commands where it was intermediary command to another was displaying the global arguments. For example, the CLI was waiter commands where you can specify `aws ec2 wait <name-of-waiter>`. These intermediary commands are more for providing organization/hierarchy and cannot be successfully invoked directly (e.g. `aws ec2 wait`) and so it does not make sense to include the global options as part of the documentation (e.g. running `aws ec2 wait help`) similar to what we do for service commands.

To address this, we should probably override the `doc_global_option` for the base custom command doc handler: https://github.com/aws/aws-cli/blob/f39908a51c2843e146c56b2c2625b63f15312ad0/awscli/customizations/commands.py#L363  In overriding it, we should have the logic check to see if there is anything in the command table and only document global arguments if there are no commands present.

kyleknap (commented):
Looks good. I think we are getting close. I just had some more comments.

awscli/clidocs.py
```diff
@@ -335,6 +338,9 @@ def doc_option_example(self, arg_name, help_command, **kwargs):
     def doc_options_end(self, help_command, **kwargs):
         pass
 
+    def doc_global_option(self, help_command, **kwargs):
```
kyleknap, line 341: Another thing I noticed is that with this change we should remove this logic as well: https://github.com/aws/aws-cli/blob/f39908a51c2843e146c56b2c2625b63f15312ad0/awscli/clidocs.py#L377-L384 that adds the `See 'aws help' for descriptions of global parameters.` since we are now documenting them.

awscli/clidocs.py
```diff
@@ -716,3 +710,42 @@ def _line_has_tag(self, line):
...
+            help_command.doc.write((f"``{argument.cli_name}`` "
+                                    f"({argument.cli_type_name})"))
+            help_command.doc.include_doc_string(argument.documentation)
+            if argument.choices:
+                help_command.doc.writeln(" ")
+                help_command.doc.style.start_ul()
+                for choice in argument.choices:
+                    help_command.doc.style.li(choice)
+                help_command.doc.style.end_ul()
+            help_command.doc.writeln(" ")
```
kyleknap, line 737: Another thing that I noticed while taking another pass at it is that the option docs are slightly different than the operation options:

1. The docstrings are on a new line
2. The docstrings are indented

I realize this is mainly ported from the original implementation, but at this point it probably makes sense to make them consistent in format between the two different types of options. Playing around with this a bit more. I was able to update it to this to do the job:
```python
            help_command.doc.writeln(
                f"``{argument.cli_name}`` ({argument.cli_type_name})")
            help_command.doc.style.indent()
            help_command.doc.style.new_paragraph()
            help_command.doc.include_doc_string(argument.documentation)
            if argument.choices:
                help_command.doc.style.start_ul()
                for choice in argument.choices:
                    help_command.doc.style.li(choice)
                help_command.doc.style.end_ul()
            help_command.doc.style.dedent()
            help_command.doc.style.new_paragraph()
```
I think there may be a couple places where there are more newlines than needed but that's fine.

kyleknap (commented):
Looks fine. I just had some more comments based on the updates.

tests/unit/test_clidocs.py
```diff
@@ -613,3 +573,43 @@ def test_description_tags_in_body(self):
...
+        arg_table = {}
+        for t in types:
+            name = f'{t}_type'
+            help_text = f'This arg type is {t}'
+            choices = ['A', 'B', 'C'] if t == 'string' else []
+            arg_table[name] = CustomArgument(name=name,
+                                             cli_type_name=t,
+                                             help_text=help_text,
+                                             choices=choices)
+        help_command = mock.Mock()
```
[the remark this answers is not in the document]
hssyoo: I mocked the `help_command` because in order to initialize an actual help command object I would need to mock out a `session` anyways. I added a `spec`.

awscli/customizations/commands.py
```diff
@@ -411,12 +410,16 @@ def doc_synopsis_option(self, arg_name, help_command, **kwargs):
             pass
 
     def doc_synopsis_end(self, help_command, **kwargs):
-        if not help_command.synopsis:
+        if not help_command.synopsis and help_command.arg_table:
```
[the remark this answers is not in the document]
hssyoo: This was based on observed behavior. Using `command_table` failed to generate global options when expected whereas `arg_table` succeeded. `command_table` also generates global options for `wait` commands

awscli/clidocs.py
```diff
@@ -179,6 +187,11 @@ def doc_option(self, arg_name, help_command, **kwargs):
         doc.style.dedent()
         doc.style.new_paragraph()
 
+    def doc_global_option(self, help_command, **kwargs):
+        doc = help_command.doc
+        doc.style.h2('Global Options')
+        doc.write_from_file(GLOBAL_OPTIONS_FILE)
+
```
kyleknap, line 194: We should also make sure to include a changelog entry for this using the `scripts/new-change` script.

kyleknap (commented):
Looks good. Just some more more smaller comments mostly around the tests.

tests/unit/customizations/test_commands.py
```diff
@@ -144,45 +144,101 @@ def create_help_command(self):
...
+        operation_handler.doc_global_option(help_command=help_command)
+        rendered = help_command.doc.getvalue()
+        rendered = rendered.decode('utf-8')
+        for arg in help_command.arg_table:
+            arg_name = help_command.arg_table.get(arg).cli_name
+            type_name = help_command.arg_table.get(arg).cli_type_name
+            target = f'``{arg_name}`` ({type_name})'
+            self.assertIn(target, rendered)
+
+    def test_not_includes_global_options_html(self):
```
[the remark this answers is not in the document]
hssyoo: * I saw the same thing happening elsewhere, but upon a closer look I realize that the tests were separate because the texts were slightly different between `man` and `html`. I've made them into single tests.
* That's a good point about the exact conditions that trigger them. The logic is based on whether or not the help command has an empty `command_table`. Changing the tests to reflect that

awscli/customizations/commands.py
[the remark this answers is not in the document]
hssyoo: Ohh I was wrongfully checking `help_command.command_table` without the `not`. I see now that it makes sense to only show the global options if it's reached the end of the command chain.

kyleknap (commented):
Looks good! Let's go ahead squash the commit and reformat the commit message and we should be set to merge this.

kyleknap (approved):
Looks great! :ship: Merging

PR #7185, opened 2022-08-10
Document differences between regular and streaming blobs

hssyoo:
The lack of a distinction between regular and streaming blobs in our documentation is causing some confusion among our users. This PR addresses the issue by:
-  Explicitly labeling streaming blobs as `streaming blob`.
- Generating a note that clarifies the usage distinction between regular and streaming blobs if any blob type argument is detected.

Preview of changes:
image

image

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -22,7 +22,7 @@ from awscli.bcdoc.docevents import DOC_EVENTS
 from awscli.topictags import TopicTagDB
 from awscli.utils import (
     find_service_and_method_in_event_name, is_document_type,
-    operation_uses_document_types
+    operation_uses_document_types, is_streaming_blob_type
 )
 
 LOG = logging.getLogger(__name__)
@@ -48,6 +48,8 @@ class CLIDocumentEventHandler(object):
             return 'JSON'
         if is_document_type(shape):
             return 'document'
+        if is_streaming_blob_type(shape):
+            return 'streaming blob'
         return default
 
     def _map_handlers(self, session, event_class, mapfn):
@@ -173,6 +175,8 @@ class CLIDocumentEventHandler(object):
             argument.argument_model, argument.cli_type_name)))
         doc.style.indent()
         doc.include_doc_string(argument.documentation)
+        if is_streaming_blob_type(argument.argument_model):
+            self._add_streaming_blob_note(doc)
         if hasattr(argument, 'argument_model'):
             self._document_enums(argument.argument_model, doc)
             self._document_nested_structure(argument.argument_model, doc)
@@ -263,6 +267,15 @@ class CLIDocumentEventHandler(object):
             self._doc_member(doc, '', member_shape.member, stack)
         doc.style.dedent()
         doc.style.new_paragraph()
+
+    def _add_streaming_blob_note(self, doc):
+        doc.style.start_note()
+        msg = ("This argument is of type: streaming blob. "
+               "Its value must be the path to a file "
+               "(e.g. ``path/to/file``) and must **not** "
+               "be prefixed with ``file://`` or ``fileb://``")
+        doc.writeln(msg)
+        doc.style.end_note()
 
 
 class ProviderDocumentEventHandler(CLIDocumentEventHandler):
--- a/tests/unit/test_clidocs.py
+++ b/tests/unit/test_clidocs.py
@@ -13,7 +13,7 @@
 import json
 
 from botocore.model import ShapeResolver, StructureShape, StringShape, \
-    ListShape, MapShape
+    ListShape, MapShape, Shape
 
 from awscli.testutils import mock, unittest, FileCreator
 from awscli.clidocs import OperationDocumentEventHandler, \
@@ -22,6 +22,7 @@ from awscli.clidocs import OperationDocumentEventHandler, \
 from awscli.bcdoc.restdoc import ReSTDocument
 from awscli.help import ServiceHelpCommand, TopicListerCommand, \
     TopicHelpCommand
+from awscli.arguments import CustomArgument
 
 
 class TestRecursiveShapes(unittest.TestCase):
@@ -405,6 +406,32 @@ class TestCLIDocumentEventHandler(unittest.TestCase):
             "See :doc:`'aws help' </reference/index>` for descriptions of "
             "global parameters", rendered
         )
+
+    def test_includes_streaming_blob_options(self):
+        help_command = self.create_help_command()
+        blob_shape = Shape('blob_shape', {'type': 'blob'})
+        blob_shape.serialization = {'streaming': True}
+        blob_arg = CustomArgument('blob_arg', argument_model=blob_shape)
+        help_command.arg_table = {'blob_arg': blob_arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='blob_arg',
+                                     help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertIn('streaming blob', rendered)
+
+    def test_streaming_blob_comes_after_docstring(self):
+        help_command = self.create_help_command()
+        blob_shape = Shape('blob_shape', {'type': 'blob'})
+        blob_shape.serialization = {'streaming': True}
+        blob_arg = CustomArgument(name='blob_arg',
+                                  argument_model=blob_shape,
+                                  help_text='FooBar')
+        help_command.arg_table = {'blob_arg': blob_arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='blob_arg',
+                                     help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertRegex(rendered, r'FooBar[\s\S]*streaming blob')
 
 
 class TestTopicDocumentEventHandlerBase(unittest.TestCase):
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 import signal
 import platform
+import pytest
 import subprocess
 import os
 
@@ -20,10 +21,15 @@ import botocore.model
 from awscli.testutils import unittest, skip_if_windows, mock
 from awscli.utils import (
     split_on_commas, ignore_ctrl_c, find_service_and_method_in_event_name,
-    is_document_type, is_document_type_container,
+    is_document_type, is_document_type_container, is_streaming_blob_type,
     operation_uses_document_types, ShapeWalker, ShapeRecordingVisitor,
     OutputStreamFactory
 )
+
+
+@pytest.fixture()
+def argument_model():
+    return botocore.model.Shape('argument', {'type': 'string'})
 
 
 class TestCSVSplit(unittest.TestCase):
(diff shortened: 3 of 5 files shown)


tests/unit/test_utils.py
```diff
@@ -409,3 +409,20 @@ def test_can_escape_recursive_shapes(self):
...
+        self.streaming_blob = mock.Mock()
+        self.streaming_blob.type_name = 'blob'
+        self.streaming_blob.serialization = {'streaming': True}
+        self.non_blob = mock.Mock()
+        self.non_blob.type_name = 'string'
+    
+    def test_is_streaming_blob_type(self):
+        self.assertEqual(is_streaming_blob_type(self.blob), False)
+        self.assertEqual(is_streaming_blob_type(self.streaming_blob), True)
+        self.assertEqual(is_streaming_blob_type(self.non_blob), False)
```
kyleknap, line 428: For these test assertions, let's break them out into separate test methods. The reason for doing it in separate methods is that if any of the assertions fail, any assertions made past that point of failure will not be run and you will not be aware if an assertion fails until you fix the failing assertion. By spitting it into multiple test cases, you get a better representation of all assertions that are failing.

tests/unit/test_utils.py
```diff
@@ -409,3 +409,20 @@ def test_can_escape_recursive_shapes(self):
         }
         self.walker.walk(self.get_shape_model('Recursive'), self.visitor)
         self.assert_visited_shapes(['Recursive'])
+
+
+class TestStreamingBlob(unittest.TestCase):
```
kyleknap, line 414: Instead of relying on `unittest` for test cases declaration, let's convert this to `pytest` style meaning:

1. No subclassing from `unittest.TestCase`
2. Using fixtures instead of the `setUp` method: https://docs.pytest.org/en/6.2.x/fixture.html. We'd likely would want just a fixture for a fake `argument_model` and then we can set the necessary properties (e.g. `type_name` and `serilization`) in each respective test case.
3. Using `assert` instead of methods from `unittest.TestCase`

Most of the codebase is still relying on `unittest` style but in the long run we want test to be following more closely to pytest style of writing test, as it is the standard now days. So for any new tests, we've been trying to use pytest style of tests assuming it would be straightforward/possible. I would recommend checking out the botocore codebase for  examples of how tests are written using pytest style.

awscli/clidocs.py
```diff
@@ -153,6 +155,18 @@ def doc_synopsis_end(self, help_command, **kwargs):
     def doc_options_start(self, help_command, **kwargs):
         doc = help_command.doc
         doc.style.h2('Options')
+        for arg in help_command.arg_table:
+            shape = help_command.arg_table.get(arg).argument_model
+            if is_streaming_blob_type(shape):
```
kyleknap, line 160: Instead of adding this snippet to the top of the `Options` section, would it be possible to add it as part of the description to the argument? I think that would make it easier to notice.

awscli/clidocs.py
```diff
@@ -153,6 +155,18 @@ def doc_synopsis_end(self, help_command, **kwargs):
...
+                doc.style.start_note()
+                msg = ("<p>``blob`` type options must be prefixed by "
+                       "``file://`` or ``fileb://`` "
+                       "(``file://path/name.json``) and "
+                       "``streaming blob`` type options must be the path "
+                       "to a file (``path/name.json``).</p>")
```
kyleknap, line 166: Then for the message we include, let's just keep it to:
```
This argument is of type: streaming blob. It's value must be the path to a file (e.g. ``path/to/file``) and must **not** be prefixed with ``file://`` or ``fileb://``
```

awscli/clidocs.py
```diff
@@ -48,6 +48,8 @@ def _get_argument_type_name(self, shape, default):
             return 'JSON'
         if is_document_type(shape):
             return 'document'
+        if is_streaming_blob_type(shape):
```
kyleknap, line 51: As a rule of thumb, for any new logic we add there should be a test for it. So we should make sure we have unit tests for adding this new type to documentation and the note on the value to specify for a streaming blob. We probably can add it somewhere in `tests/unit/test_clidocs.py`.

kyleknap (commented):
Looks good. Just had some comments on the wording and the tests.

awscli/clidocs.py
```diff
@@ -159,6 +161,14 @@ def doc_options_start(self, help_command, **kwargs):
...
         doc = help_command.doc
         argument = help_command.arg_table[arg_name]
+        if is_streaming_blob_type(argument.argument_model):
+            doc.style.start_note()
+            msg = ("This argument is of type: streaming blob. "
+                   "Its value must be the path to a file "
+                   "(e.g. ``path/to/file``) and must **not** "
+                   "be prefixed with ``file://`` or ``fileb://``")
+            doc.include_doc_string(msg)
+            doc.style.end_note()
```
kyleknap, line 171: Let's encapsulate this logic as a `_add_streaming_blob_note()` method. The `doc_option()` method is already too long and could use some refactoring. So this will help ensure it does not make readability any worse.

awscli/clidocs.py
```diff
@@ -159,6 +161,14 @@ def doc_options_start(self, help_command, **kwargs):
     def doc_option(self, arg_name, help_command, **kwargs):
         doc = help_command.doc
         argument = help_command.arg_table[arg_name]
+        if is_streaming_blob_type(argument.argument_model):
```
kyleknap, line 164: Let's move this logic down to after the line `doc.include_doc_string(argument.documentation)`. Mainly when the documentation gets generated, the note shows up above the argument's documentation and it can be interpreted as a note for the previous argument, especially since the note does not explicitly use the argument name.

awscli/clidocs.py
```diff
@@ -159,6 +161,14 @@ def doc_options_start(self, help_command, **kwargs):
...
+            doc.style.start_note()
+            msg = ("This argument is of type: streaming blob. "
+                   "Its value must be the path to a file "
+                   "(e.g. ``path/to/file``) and must **not** "
+                   "be prefixed with ``file://`` or ``fileb://``")
+            doc.include_doc_string(msg)
```
kyleknap, line 170: Was there a particular reason `include_doc_string()` was used instead of `writeln()`? The main reason I ask is the `include_doc_string()` method is generally only used when we are dealing with XML/HTML based docstrings that we need to convert to RSt.

tests/unit/test_utils.py
```diff
@@ -409,3 +410,25 @@ def test_can_escape_recursive_shapes(self):
         }
         self.walker.walk(self.get_shape_model('Recursive'), self.visitor)
         self.assert_visited_shapes(['Recursive'])
+
+
+@pytest.fixture(scope='module')
+def argument_model():
+    return botocore.model.Shape('', {'type': ''})
+
```
kyleknap, line 418: It might be worth running `flake8` over the PR. Noticed that there is only one new line between this method and class and there should be two.

tests/unit/test_utils.py
```diff
@@ -409,3 +410,25 @@ def test_can_escape_recursive_shapes(self):
         }
         self.walker.walk(self.get_shape_model('Recursive'), self.visitor)
         self.assert_visited_shapes(['Recursive'])
+
+
+@pytest.fixture(scope='module')
+def argument_model():
```
kyleknap, line 416: For pytest fixtures, let's keep it at the top of the file in between the imports and test cases. The main reason I'm suggesting it is that when the fixtures are mixed between test cases it can be hard to realize that there is a fixture available and we may want to use it in the other test cases in the future. So keeping the fixtures all near the top, makes it easier to find them.

tests/unit/test_utils.py
```diff
@@ -409,3 +410,25 @@ def test_can_escape_recursive_shapes(self):
         }
         self.walker.walk(self.get_shape_model('Recursive'), self.visitor)
         self.assert_visited_shapes(['Recursive'])
+
+
+@pytest.fixture(scope='module')
```
kyleknap, line 415: Let's remove the `scope` argument for this fixture. Specifically, I don't think we want the `scope` to be `module` because it will mean this instance will only be instantiated once, which means a test case can change the state of the argument model and have it affect the state in a subsequent test case. By removing the `scope`, it will make sure the fixture is instantiated brand new for each test case and thus have a clean state.

tests/unit/test_utils.py
```diff
@@ -409,3 +410,25 @@ def test_can_escape_recursive_shapes(self):
...
+def argument_model():
+    return botocore.model.Shape('', {'type': ''})
```
kyleknap, line 417: For the `Shape` instantiation, let's use realistic values. I'm fine giving it the name `argument` and its type `string`.

kyleknap (commented):
Looks good! I just had some smaller comments.

awscli/clidocs.py
hssyoo: Ah I see. No reason other than it works. Changing to `writeln()`.

tests/unit/test_clidocs.py
```diff
@@ -406,6 +407,29 @@ def test_includes_global_args_ref_in_html_options(self):
...
+        blob_shape = Shape('blob_shape', {'type': 'blob'})
+        blob_shape.serialization = {'streaming': True}
+        blob_arg = CustomArgument('blob_arg', argument_model=blob_shape)
+        help_command.arg_table = {'blob_arg': blob_arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='blob_arg', help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertIn('streaming blob', rendered)
+
+    def test_includes_streaming_blob_in_html_options(self):
```
kyleknap, line 422: Similar to another PR, I'm not sure if we need to add another test for when html is the target especially since there is no branching logic. I'd suggest just removing this test and rename the other one to `test_includes_streaming_blob_options()`.

tests/unit/test_clidocs.py
```diff
@@ -406,6 +407,29 @@ def test_includes_global_args_ref_in_html_options(self):
...
+    def test_includes_streaming_blob_in_man_options(self):
+        help_command = self.create_help_command()
+        blob_shape = Shape('blob_shape', {'type': 'blob'})
+        blob_shape.serialization = {'streaming': True}
+        blob_arg = CustomArgument('blob_arg', argument_model=blob_shape)
+        help_command.arg_table = {'blob_arg': blob_arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='blob_arg', help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertIn('streaming blob', rendered)
```
kyleknap, line 420: It would also be nice if we assert that the streaming blob note gets added after the docstring.

kyleknap (commented):
Looks good. I think we are close to merging it. Just had a couple comments on the tests. Also let's make sure to:

1. Add a changelog for this. We can mark it as an `enhancement` for the category of `Documentation`.
2. Let's squash all of these commits into a single commit.

tests/unit/test_clidocs.py
```diff
@@ -406,6 +407,32 @@ def test_includes_global_args_ref_in_html_options(self):
...
+        blob_shape.serialization = {'streaming': True}
+        blob_arg = CustomArgument(name='blob_arg',
+                                  argument_model=blob_shape,
+                                  help_text='FooBar')
+        help_command.arg_table = {'blob_arg': blob_arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='blob_arg',
+                                     help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertRegex(rendered, 'FooBar[\s\S]*streaming blob')
```
kyleknap, line 434: Let's make sure to prefix the regex string with an `r` (e.g. `r'FooBar[\s\S]*streaming blob'`) to make it a raw string. The reason I'm suggesting it is that if it the string is not prefixed with `r`, Python will throw deprecation warnings about invalid escape sequences being used in the literal string and this can make the pytest output noisy.

kyleknap (commented):
Looks good. Just had a really small comment. Let's also make sure to format the commit message in this style: https://github.com/aws/aws-cli/blob/develop/CONTRIBUTING.md#git-commits-and-workflow specifically the summary sentence is a bit long.

kyleknap (approved):
Looks good! :ship: Merging

Issue #6695, opened 2022-02-03
Improve functionality/documentation for `--generate-cli-skeleton`


tim-finnigan:
The current behavior of `--generate-cli-skeleton` has caused some confusion (see https://github.com/aws/aws-cli/issues/5266 and https://github.com/aws/aws-cli/issues/6684). 

Here in the code shows where a random value is chosen for strings: https://github.com/boto/botocore/blob/c3731390f7a98c35d13f766ba99e465c2ede9c78/botocore/utils.py#L942

I think that value should be set to whatever the default is, or the docs/user guide should clarify the current behavior: https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-skeleton.html

PR #7190, opened 2022-08-12
Remove random behavior when using cli-generate-skeleton

hssyoo:
Related issue: https://github.com/aws/aws-cli/issues/6695

Currently, if a shape object has the enum attribute, then it'll return a randomly sampled element from the list when using the `generate-cli-argument`. This is confusing users who expect consistent outputs from their inputs. This PR takes the first element of the list instead in order to provide consistent reproducibility. It also auto-generates a note in docs whenever `generate-cli-argument` is a valid argument that lets users know that the change in behavior is only in v2 and there's no backwards-compatibility guarantee since v1's behavior is random.

Preview of doc note:
image

### DIFF
--- a/.changes/next-release/enhancement-generatecliskeleton-45373.json
+++ b/.changes/next-release/enhancement-generatecliskeleton-45373.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``--generate-cli-skeleton``",
+  "description": "Argument populates enums in JSON skeleton with stable enum value, fixes (`#6695 <https://github.com/aws/aws-cli/issues/6695>`__)."
+}
--- a/awscli/botocore/utils.py
+++ b/awscli/botocore/utils.py
@@ -970,7 +970,7 @@ class ArgumentGenerator(object):
                 if self._use_member_names:
                     return name
                 if shape.enum:
-                    return random.choice(shape.enum)
+                    return shape.enum[0]
                 return ''
             elif shape.type_name in ['integer', 'long']:
                 return 0
--- a/awscli/customizations/generatecliskeleton.py
+++ b/awscli/customizations/generatecliskeleton.py
@@ -54,7 +54,10 @@ class GenerateCliSkeletonArgument(OverrideRequiredArgsArgument):
             '``yaml-input`` it will print a sample input YAML that can be '
             'used with ``--cli-input-yaml``. If provided with the value '
             '``output``, it validates the command inputs and returns a '
-            'sample output JSON for that command.'
+            'sample output JSON for that command. '
+            'The generated JSON skeleton is not stable between versions '
+            'of the AWS CLI and there are no backwards compatibility '
+            'guarantees in the JSON skeleton generated.'
         ),
         'nargs': '?',
         'const': 'input',
--- a/tests/unit/botocore/test_utils.py
+++ b/tests/unit/botocore/test_utils.py
@@ -546,14 +546,14 @@ class TestArgumentGenerator(unittest.TestCase):
 
     def test_generate_string_enum(self):
         enum_values = ['A', 'B', 'C']
-        model = {
-            'A': {'type': 'string', 'enum': enum_values}
-        }
-        shape = DenormalizedStructureBuilder().with_members(
-            model).build_model()
-        actual = self.arg_generator.generate_skeleton(shape)
-
-        self.assertIn(actual['A'], enum_values)
+        self.assert_skeleton_from_model_is(
+            model={
+                'A': {'type': 'string', 'enum': enum_values}
+            },
+            generated_skeleton={
+                'A': enum_values[0]
+            }
+        )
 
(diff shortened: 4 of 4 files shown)


awscli/clidocs.py
```diff
@@ -159,6 +159,15 @@ def doc_options_start(self, help_command, **kwargs):
     def doc_option(self, arg_name, help_command, **kwargs):
         doc = help_command.doc
         argument = help_command.arg_table[arg_name]
+        if argument.cli_name == '--generate-cli-skeleton':
```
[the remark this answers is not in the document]
hssyoo: Or even just a list of functions
```python
for note_gen in note_gens: # list of functions that write notes if condition
    note_gen(argument)
```

awscli/botocore/utils.py
```diff
@@ -970,7 +970,7 @@ def _generate_skeleton(self, shape, stack, name=''):
                 if self._use_member_names:
```
kyleknap, line 970: Let's make sure to add change log for this. You can run the `scripts/new-change` script to add one. I would recommend labeling it as an enhancement for `--generate-cli-skeleton`.

awscli/clidocs.py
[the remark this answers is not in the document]
kyleknap: Ah. So for the `--generate-cli-skeleton` option it is a customization with its own hand-curated documentation: https://github.com/aws/aws-cli/blob/v2/awscli/customizations/generatecliskeleton.py#L50. I would recommend just updating that docstring directly with the content that we plan to add. In general, we should only be updating `clidocs` module if we are working with model-driven arguments (e.g. derived from the `service-2.json` models).

awscli/clidocs.py
```diff
@@ -159,6 +159,15 @@ def doc_options_start(self, help_command, **kwargs):
...
+            doc.style.start_note()
+            msg = ("AWS CLI version 2's implementation of "
+                   "``generate-cli-skeleton`` selects the first element "
+                   "if an input shape contains an enum. This behavior "
+                   "may not be backwards-compatible with version 1's "
+                   "implementation, which selects a random element.")
```
kyleknap, line 168: For the message that we are adding let's:

1. Not make it a note directive especially if we are adding it to the docstring. Mainly, I want to make sure we do not have too many note blocks in the documentation as it becomes hard to read.
2. For the content, let's just append to the end of the current docstring: `The generated JSON skeleton is not stable between versions of the AWS CLI and there are no backwards compatibility guarantees in the JSON skeleton generated.` I'm mainly suggesting this because the note about the enum is a bit specific and in reality, we may change more than just the enum values (e.g. we may change the placeholder values too in the future) and it gets more to the point that users should be carefully in using the value programatically.

tests/unit/test_clidocs.py
```diff
@@ -407,6 +408,41 @@ def test_includes_global_args_ref_in_html_options(self):
             "global parameters", rendered
         )
 
+    def test_includes_generate_cli_skeleton_note_in_man_options(self):
```
kyleknap, line 411: If we add it directly to the docstring of the argument, we can go ahead and remove these tests. We don't need tests for direct docstrings.

kyleknap (commented):
Looks good! I just had some comments.

.changes/next-release/enhancement-generatecliskeleton-45373.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``--generate-cli-skeleton``",
+  "description": "Fixes `#6695 <https://github.com/aws/aws-cli/issues/6695>`__. Now consistently returns a stable JSON skeleton, rather than a random one each time."
```
kyleknap, line 4: For the changelog entry, let's just keep it the description to:
```
Argument populates enums in JSON skeleton with stable enum value, fixes (`#6695 <https://github.com/aws/aws-cli/issues/6695>`__).
```
Mainly, I'm not sure if we are necessarily committing to the JSON skeleton to remain stable across invocations in the future. If we were, we should make sure to add functional tests (to ensure stability) if this is behavior that we fully supported. I think for now let's just keep it scoped to making enums stable.

awscli/customizations/generatecliskeleton.py
```diff
@@ -55,6 +55,9 @@ class GenerateCliSkeletonArgument(OverrideRequiredArgsArgument):
             'used with ``--cli-input-yaml``. If provided with the value '
             '``output``, it validates the command inputs and returns a '
             'sample output JSON for that command.'
+            'The generated JSON skeleton is not stable between versions '
```
kyleknap, line 58: I think we need to add a space between `command.` and `The`. I think we can just add the space in the string above this line.

kyleknap (commented):
Looks good. Just a couple more comments. Let's also make sure to squash the commits into one and we should be set.

kyleknap (approved):
Looks great! :ship: Merging the pull request.

Issue #7173, opened 2022-08-05
Allow multiple regions to be specified on invocation


ablackrw:
### Describe the feature

For selected subcommands (such as `ses list-identities` or `ec2 describe-vpcs`), allow the user to specify multiple regions for execution using the `--region` command line switch. Rather than specifying a single region name, the user would provide a comma separated list of region names. For this enhancement to be relevant, the subcommand would need to
1. Be region specific
2. Allow execution without a specifying an instance identifier

When multiple regions are specified, the output would be generated in map form, with the map keys being the region name where the command was executed and the map values being the output of the command execution.

### Use Case

From time to time, we need to gather information about our AWS environment across all of our accounts and regions. To gather this information, we resort to using a shell looping construct similar to the following
```bash
echo -n "{" > ses.json
for R in us-east-2 us-east-1 us-west-1 us-west-2 af-south-1 ap-south-1 ap-northeast-3 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-northeast-1 ca-central-1 eu-central-1 eu-west-1 eu-west-2 eu-south-1 eu-west-3 eu-north-1 me-south-1 sa-east-1
do 
  echo -n "${R}:" | tee -a ses.json
  aws --output=json --region=${R} ses list-identities >> ses.json
  echo "," >> ses.json
done
echo "}" >> ses.json
```

A problem with the above construct is that it fails to produce syntactically correct json (though the output is close)

The proposed feature would allow us to make the invocation as follows
`aws --output=json --region=us-east-2,us-east-1,us-west-1,us-west-2,af-south-1,ap-south-1,ap-northeast-3,ap-northeast-2,ap-southeast-1,ap-southeast-2,ap-northeast-1,ca-central-1,eu-central-1,eu-west-1,eu-west-2,eu-south-1,eu-west-3,eu-north-1,me-south-1,sa-east-1 ses list-identities > ses.json`
This would have the side benefit of producing syntactically correct json.

### Proposed Solution

A naive (and untested) implementation of the change is as follows  
```diff
index 304cebbe6..1baef958e 100644
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -645,12 +645,23 @@ class CLIOperationCaller(object):
             value is returned.
 
         """
-        client = self._session.create_client(
-            service_name, region_name=parsed_globals.region,
-            endpoint_url=parsed_globals.endpoint_url,
-            verify=parsed_globals.verify_ssl)
[...]

PR #7193, opened 2022-08-15
Add support for iterating over regions and accounts in CLI

ablackrw:
*Issue #, if available:*
Prototype implementation for #7173

*Description of changes:*
The goal of this change is to provide a simple way to execute a particular
command in multiple regions and accounts. The implemented mechanism is roughtly
an order of magnitude faster in execution time than iterating using a shell
loop.

For account iteration, the organizations:ListAccounts facility is leveraged
to provide a list of accounts to iterate over. The sts:AssumeRole service is
then used to obtain access to the identified accounts using role switching.

For region switching, a client is created for each enumerated region. This
facility uses a new command line switch rather than overloading the
--region switch to allow the account iteration to function correctly.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import copy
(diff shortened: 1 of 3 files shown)


volphy:
It would be a great addition to the AWS CLI.

Currently, I use steampipe.io with its AWS plugin to simulate a similar functionality.

PR #7203, opened 2022-08-18
Allow usage of SSL common name with deprecation.

dlm6693:
In the comming months, the method by which every AWS SDK resolves
their endpoints, including aws-cli v1 and boto3, will be udpated with
completely different behavior. At this time,the use of the
`sslCommonName` paradigm will be fully deprecated. Additionally, it
was mainly used to support python 2.6, which has not been supported by
boto3 or the CLI since January 1, 2020. The boto team wanted to ease the
transition by raising a warning to users and providing an optional
environment variable to fully disable the behavior immediately. However,
this causes breaking changes in v1 of aws-cli so additional changes were
required in it to release the deprecation.

This commit attempts to formally maintain the `sslCommonName` paradigm
in aws-cli v1. This is accomplished by providing hardcoded, manual
overrides for each service and region combination where this parameter
is used over the standard `hostname`. The list of these service/region
combinations was compiled by creating a client for every single
combination and emitting the `endpoint_url` to a file twice. The first
using a version of botocore that uses `sslCommonName` and one that
doesn't, then diffing the two files to see which URLs changed.

While behavior for the end user will largely be the same, if they do not
disable the common name behavior by setting the environment variable
`BOTO_DISABLE_COMMONNAME` to `true`, a warning will be omitted in the
console once per session. Additional details can be found in
boto/botocore#2705.

The override occurs when the `before-building-argument-table-parser` is
omitted by an event handler. At this point, if the user has not set
`endpoint_url` in `parsed_globals` (`--endpoint-url` flag from the command
line) the endpoint that the CLI uses will be overridden to continue to
use the SSL common name. There were a few different options of which
event to run the override in, but ultimately this one was chosen because
it was the least invasive to the existing code base. As this is very low
level behavior for the CLI, minimizing the amount of changes made was
prioritized over one that might seemt to make more logical sense like
`top-level-args-parsed`, which is when the arguments passed to the
command line are actually parsed through. However, the usage of a
session instance was required  in order to get config variables like
region and partition. There is a strong possiblity that this
customization override will run twice per call, but that is a design
flaw the maintainers are aware of and have accpeted.

Additionally for the `health` service, the region parameter needed to be
overwritten because of it's use of pseudo regions like `aws-global`. In
the other cases, the region within the URL essentially always matched
the region, but since health uses the SSL common name in global pseudo
regions, this change was required as well or an exception would be
raised.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.changes/next-release/enhancement-Endpoints-36706.json
+++ b/.changes/next-release/enhancement-Endpoints-36706.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Endpoints",
+  "description": "Enforce SSL common name as default endpoint url"
+}
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -484,7 +484,7 @@ class ServiceOperation(object):
         event = 'before-building-argument-table-parser.%s.%s' % \
             (self._parent_name, self._name)
         self._emit(event, argument_table=self.arg_table, args=args,
-                   session=self._session)
+                   session=self._session, parsed_globals=parsed_globals)
         operation_parser = self._create_operation_parser(self.arg_table)
         self._add_help(operation_parser)
         parsed_args, remaining = operation_parser.parse_known_args(args)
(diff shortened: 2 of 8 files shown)


kyleknap (approved):
:ship:

PR #7205, opened 2022-08-18
Add note for Tagged Unions

hssyoo:
Services and `botocore` support Tagged Union structures. The purpose of this PR is to enhance the CLI docs by generating a usage note for Tagged Union structures. Because Tagged Union structures can be nested under other complex types, the change needed to be made for nested options, in addition to top level options.

Preview of changes:
image

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -23,7 +23,8 @@ from awscli.bcdoc.docevents import DOC_EVENTS
 from awscli.topictags import TopicTagDB
 from awscli.utils import (
     find_service_and_method_in_event_name, is_document_type,
-    operation_uses_document_types, is_streaming_blob_type
+    operation_uses_document_types, is_streaming_blob_type,
+    is_tagged_union_type
 )
 
 LOG = logging.getLogger(__name__)
@@ -56,6 +57,8 @@ class CLIDocumentEventHandler(object):
             return 'document'
         if is_streaming_blob_type(shape):
             return 'streaming blob'
+        if is_tagged_union_type(shape):
+            return 'tagged union structure'
         return default
 
     def _map_handlers(self, session, event_class, mapfn):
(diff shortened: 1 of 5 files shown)


awscli/clidocs.py
```diff
@@ -173,6 +173,8 @@ def doc_option(self, arg_name, help_command, **kwargs):
             argument.argument_model, argument.cli_type_name)))
         doc.style.indent()
         doc.include_doc_string(argument.documentation)
+        if getattr(argument.argument_model, 'is_tagged_union', False):
```
kyleknap, line 176: One thing that I noticed is that it would be great if we can update the type in the docs similar to what we did for streaming blobs (e.g. make the type `(tagged union structure)`) next to the field name. I'm mainly suggesting this because it will help reconfirm what field the note is for.

tests/unit/test_clidocs.py
```diff
@@ -406,6 +407,41 @@ def test_includes_global_args_ref_in_html_options(self):
...
+        arg = CustomArgument(name='tagged_union',
+                             argument_model=tagged_union,
+                             help_text='FooBar')
+        help_command.arg_table = {'tagged_union': arg}
+        operation_handler = OperationDocumentEventHandler(help_command)
+        operation_handler.doc_option(arg_name='tagged_union',
+                                     help_command=help_command)
+        rendered = help_command.doc.getvalue().decode('utf-8')
+        self.assertRegex(rendered, r'FooBar[\s\S]*Tagged Union')
+
```
kyleknap, line 444: It would be great if we can add a test as well for tagged unions being documented in the output structure of the command. It looks like the tests are only covering input currently.

kyleknap (commented):
Looks good! I just had a couple of suggestions.

tests/unit/test_clidocs.py
hssyoo: Agreed. I ended up mocking out a `Shape` object for this test because `OperationDocumentEventHandler.doc_output()` calls a surprising number of methods, `_resolve_shape_ref()` being particularly pesky. Open to suggestions though

kyleknap: Ah. We have a class for helping with that. I'd recommend trying to use `botocore.models.DenormalizedStructureBuilder` and just set the union property on the shape it generates. We use it here for example: https://github.com/aws/aws-cli/blob/develop/tests/unit/test_shorthand.py#L166 and helps with all of the scaffolding needed.

kyleknap (commented):
Looks good! Other than the comment on using the `DenormalizedStructureBuilder`, just had one more comment. It also may be worth rebasing this off of develop. It looks like there may be some merge conflicts.

kyleknap (approved):
Looks great! :ship:

PR #7211, opened 2022-08-22
refactor config tests to be in CI and use `run_cmd` instead of `aws`

dlm6693:
This PR moves the tests for the `configure` command to the functional suite. This was done for a few reasons. First and foremost, these tests silently failed or were not run at all in a recent release of the CLI v1 that allowed a bug to be introduced to customers. Additionally, this command doesn't actually make any HTTP requests over the wire so they can be run in the GitHub CI.

These tests were also refactored to use the `run_cmd` test function instead of `aws`, which provides a more robust mocking of CLI commands.

Some light amount of changes were made to custom `configure` subcommands to comply with the new test suite.


awscli/customizations/configure/get.py
```diff
@@ -34,8 +34,12 @@ class ConfigureGetCommand(BasicCommand):
          'cli_type_name': 'string', 'positional_arg': True},
     ]
 
-    def __init__(self, session, stream=sys.stdout, error_stream=sys.stderr):
+    def __init__(self, session, stream=None, error_stream=None):
         super(ConfigureGetCommand, self).__init__(session)
+        if stream is None:
+            stream = sys.stdout
+        if error_stream is None:
+            error_stream = sys.stderr
```
dlm6693, line 42: This was updated because of the method default variable bug that every python programmer encounters from time to time. Because these defaults are evaluated before runtime, they are not properly mocked out by the test runner. Same issue for `list.py`.

awscli/customizations/configure/set.py
```diff
@@ -105,3 +105,4 @@ def _run_main(self, args, parsed_globals):
             section = profile_to_section(profile)
         updated_config = {'__section__': section, varname: value}
         self._config_writer.update_config(updated_config, config_filename)
+        return 0
```
dlm6693, line 108: The new test class `test_configure` inherits from expects a return code for ALL commands.

kyleknap (commented):
Looks great! Thanks for doing this. I realize this was likely tedious to port. I only had small comments and suggestions on improvements on the previous test content.

tests/functional/test_configure.py
```diff
@@ -0,0 +1,391 @@
...
+            "region = us-west-1\n",
+            self.get_config_file_contents()
+        )
+
+    def test_set_with_a_url(self):
+        self.run_cmd(
+            "configure set endpoint http://www.example.com",
+        )
+        self.assertEqual(
+            "[default]\n" "endpoint = http://www.example.com\n",
```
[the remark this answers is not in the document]
dlm6693: ah my apologies for missing this one

kyleknap (approved):
Looks good! :ship:

PR #7233, opened 2022-08-30
remove call to print() from the complete() call in completer

evilcrash:
completer.py contains a function known as complete(). on the last line of completer.py, complete() is printed to the console, but it returns no value. this pull request removes the call to print(), leaving just complete().

alternatively, complete() could be removed from this file and its purpose can be merged into the main function.

### DIFF
--- a/awscli/completer.py
+++ b/awscli/completer.py
@@ -143,4 +143,4 @@ if __name__ == '__main__':
     else:
         print('usage: %s <cmdline> <point>' % sys.argv[0])
         sys.exit(1)
-    print(complete(cmdline, point))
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating a PR and for your patience here. Regarding the proposed change to remove `print()` or `complete()` here, I think we would want more input from the team/community. Please [create an issue](https://github.com/aws/aws-cli/issues/new?assignees=&labels=feature-request%2Cneeds-triage&projects=&template=feature-request.yml&title=%28short+issue+description%29) outlining your reasoning if still interested in this, and we can track the discussion there before considering potential changes.

PR #7258, opened 2022-09-08
[v2] Add support for bearer auth tokens

kyleknap:
This PR ports botocore bearer auth support to v2: https://github.com/boto/botocore/pull/2743. It also enables providing configured scopes when doing SSO OIDC registration.


stealthycoin (approved):

tdaniely-dn:
I might be missing something, but seems that actually using sso creds this way doesn't work, because nothing plugs the sso-session keys in SSOProvider. It only works for login, but any following cli command will fail because it still needs the keys in the profile.

kyleknap:
@tdaniely-dn yes that is correct that the changes in this PR does not affect the `SSOProvider` class. There is some follow up work required from our end to enable using the `sso-session` configuration section with the SSO AWS credential provider.

PR #7262, opened 2022-09-09
Source dist

stealthycoin:


backends/build_system/awscli_venv.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _install_requirements(self, requirements_file, cwd=None):
+        self._pip_install(["-r", requirements_file], cwd=cwd)
+
+    def _install_awscli(self):
+        self._pip_install(
+            [
+                ROOT_DIR,
+                "--no-build-isolation",
+                "--no-cache-dir",
```
kyleknap, line 60: Was `--use-feature=in-tree-build` removed? I'm assuming it is because it is the default now? It might be worth updating the proposal to remove that flag as well to reflect the implementation now.

requirements/bootstrap.txt
```diff
@@ -0,0 +1,3 @@
+pip>=22.0.0
```
kyleknap, line 1: What are your thoughts on adding an upper bound to these dependencies? Mainly I want to make sure that we are explicit in which dependencies the source build supports as it is not uncommon for them to break backwards compatibility.

bin/aws_bash_completer
```diff
@@ -1,6 +0,0 @@
-# Typically that would be added under one of the following paths:
```
kyleknap, line 1: For the bash and zsh completers, I was thinking it may be worth keeping it in the repository and labeled as a script for now in the `setup.cfg`. For Linux distros/Homebrew, they automatically take these files and configure as part of the install to automatically get command completion. For the purpose of our source distribution though, we can just not copy it into the bin directory.

backends/build_system/exe.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _ensure_no_existing_build_dir(self):
+        if self._utils.isdir(self._dist_dir):
+            self._utils.rmtree(self._dist_dir)
+
+    def _build_aws(self):
+        aws_exe_build_dir = self._run_pyinstaller("aws.spec")
+        self._utils.update_metadata(
+            aws_exe_build_dir,
+            distribution_source=DISTRIBUTION_SOURCE,
```
kyleknap, line 46: Not needed for this PR, but something we should consider... For installations from the source distribution, we should probably update the distribution metadata to be unique for both `portable-exe` and `system-sandbox` source installs. Right now it seems if you choose `portable-exe`, the metadata will be `exe`, and if you choose `system-sandbox`, it will be `source`. However, this makes it difficult to differentiate between whether users are using the source dist or our public exes (for `portable-exe`) or direct `pip install` (for `system-sandbox`)

kyleknap (commented):
Looks like a good start. I did a preliminary pass through the implementation. I did not get to dive too much into the tests nor trying out the implementation yet. I'll try to get through that in the next pass.

configure.ac
```diff
@@ -0,0 +1,41 @@
+AC_CONFIG_MACRO_DIRS([m4])
+AC_INIT([awscli], [2.2.1])
```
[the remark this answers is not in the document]
stealthycoin: Ah yea I missed this.

backends/build_system/awscli_venv.py
stealthycoin: Yep this is the default now, the flag was removed from pip.

requirements/bootstrap.txt
stealthycoin: I think setuptools and pip are probably the two candidates for that as they break fairly regularly. Wheel is probably safe.

kyleknap (commented):
I just did a follow up pass in trying it all out locally. It was looking good. Just left some comments after trying it out a bit.

requirements/bootstrap-lock.txt
```diff
@@ -0,0 +1,60 @@
+#
```
kyleknap, line 1: I'm wondering if it makes sense to rework the naming conventions of these lock files. Mainly it is sort of hard to follow which is used where. What are your thoughts on this (I'm up for a different naming)?

* `requirements/pip.txt` - A ranged dependency on versions of pip that are required to build/install the CLI.
* `requirements/build-exe.txt` - A ranged dependency on versions of `pyinstaller` allowed
* `requirements/download-deps/bootstrap-extras.txt`- Exact pinned versions of extra dependencies that may be used as part of the `--with-downloaded-deps` install process for building/installing any of our dependencies.
* `requirements/download-deps/bootstrap-lock.txt` - Lock consisting of requirements from `pip.txt`, `bootstrap-extras.txt`, and build reqs from `pyproject.toml`
* `requirements/download-deps/system-sandbox-lock.txt` - Lock consisting of runtime dependencies from `pyproject.toml`
* `requirements/download-deps/portable-exe-lock.txt` - Lock consisting of runtime dependencies from `pyproject.toml` and `build-exe.txt`.

This feels a little more easy to understand which files are used in which portions of the build/install process.

kyleknap (commented):
It's looking good. Just had some more comments after the latest rebase and looking more into the tests.

tests/backends/build_system/unit/test_install.py
```diff
@@ -0,0 +1,201 @@
...
+            ("makedirs", "bin_dir"),
+            ("islink", "bin_dir\\aws.cmd"),
+            ("symlink", "lib_dir\\Scripts\\aws.cmd", "bin_dir\\aws.cmd"),
+        ]
+
+
+class TestUninstaller:
+    @pytest.mark.skipif(
+        platform.system() == "Windows", reason="Non-windows test"
+    )
```
kyleknap, line 173: Is there a particular reason we are switching the style for skip ifs here? Before we were using the decorator `if_windows` and `skip_if_windows`

kyleknap (commented):
Looks good. I mainly focused on the tests this round. I liked the environment abstraction you made for the integration tests.

requirements/bootstrap-lock.txt
```diff
@@ -0,0 +1,16 @@
...
+flit-core==3.7.1 \
+    --hash=sha256:14955af340c43035dbfa96b5ee47407e377ee337f69e70f73064940d27d0a44f \
+    --hash=sha256:e454fdbf68c7036e1c7435ec7479383f9d9a1650ca5b304feb184eba1efcdcef
+    # via -r requirements/bootstrap.txt
+
+# The following packages are considered to be unsafe in a requirements file:
+pip==22.2.2 \
+    --hash=sha256:3fd1929db052f056d7a998439176d3333fa1b3f6c1ad881de1885c0717608a4b \
+    --hash=sha256:b61a374b5bc40a6e982426aede40c9b5a08ff20e640f5b56977f4f91fed1e39a
+    # via -r requirements/bootstrap.txt
```
kyleknap, line 16: Is there a particular reason we have these top-level `bootstrap-lock.txt` and `bootstrap-win-lock.txt` lock files? I would have assumed we would removed them as the ones that would be used are the ones in the `download-deps` directory.

requirements/download-deps/bootstrap.txt
```diff
@@ -0,0 +1,3 @@
+setuptools==65.3.0
+wheel>=0.37.1
```
kyleknap, line 2: Is there a reason we are not pinning the dependency on `wheel` here? I'd assume we would as it would guarantee the exact version that will end up in the lock file.

requirements/portable-exe-extras-win-lock.txt
```diff
@@ -0,0 +1,43 @@
+#
```
kyleknap, line 1: Do we use this lock file anywhere? I would assume we should be able to remove this and rely solely on the `portable-exe-win-lock.txt` file for download deps.

requirements/runtime-lock.txt
```diff
@@ -0,0 +1,209 @@
+#
```
kyleknap, line 1: I think we should be able to remove this lock file as well? I would not expect it to be used directly for the non-download-deps option.

scripts/ci/install
```diff
@@ -14,10 +14,14 @@ def run(command):
...
 
 
-run('pip install --no-build-isolation -r requirements-base.txt')
-run('pip install --no-build-isolation -r requirements.txt')
-if os.path.isdir('dist') and os.listdir('dist'):
-    shutil.rmtree('dist')
-run('python -m build')
-wheel_dist = glob.glob(os.path.join('dist', '*.whl'))[0]
-run('pip install %s' % wheel_dist)
+run("pip install setuptools==57.5.0")
```
kyleknap, line 17: Let's make sure to remove this direct install and defer to our requirements files for dependencies.

scripts/ci/install-build-system
```diff
@@ -0,0 +1,23 @@
...
+_dname = os.path.dirname
+
+REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
+os.chdir(REPO_ROOT)
+
+
+def run(command):
+    return check_call(command, shell=True)
+
+run('pip install --no-build-isolation build==0.7.0')
```
kyleknap, line 16: It would be nice if we can store this dependency pin in a requirements file to make it easy to locate and update

kyleknap (commented):
It's looking good. I just posted some more smaller comments to help get this polished up so we can merge it to the feature branch.

kyleknap (commented):
Looks good. Just had a really small clean up comment on print statements, but otherwise let's go ahead and squash this into a single commit.

kyleknap (approved):
Looks good! :ship:

PR #7279, opened 2022-09-16
Pull Amazon Linux image from ECR

hssyoo:
This PR modifies the Dockerfile to pull its base image from ECR Public rather than Docker Hub, since both Amazon Linux and ECR Public are Amazon-owned. We still continue to pull the tag that tracks the major version (2), so the Dockerfile effectively does not change behavior.


kyleknap (commented):
Looks good! Let's make sure to add an `enhancement` changelog entry noting that the base image was changed to the ECR public Amazon Linux image and we should be set to get this merged.

kyleknap (approved):
Looks good! :ship:

Issue #3637, opened 2018-10-06
SQS Message Attributes Shorthand error


ldadams:
I am getting the below error trying to send SQS message attributes via shorthand syntax.  I have tried a number of different ways with same result.

`aws sqs send-message --queue-url https://sqs.us-east-1.amazonaws.com/... --message-body TestBody --message-attributes TenantID=StringValue=TestTenant,DataType="string",FileID=StringValue=352b58c522fd9ec4c764ce7a72c93cef,DataType=string`

Parameter validation failed:
Invalid type for parameter MessageAttributes.DataType, value: string, type: <type 'unicode'>, valid types: <type 'dict'>
Invalid type for parameter MessageAttributes.FileID, value: StringValue=352b58c522fd9ec4c764ce7a72c93cef, type: <type 'unicode'>, valid types: <type 'dict'>

Thanks!

justnance:
@ldadams - Thank you for reaching out.  It appears our documentation needs some work.  Please try this command as workaround while I investigate this issue further:

```
aws sqs send-message --queue-url https://sqs.us-east-1.amazonaws.com/310949364363/etl-upload-tus --message-body TestBody --message-attributes "TenantID={StringValue=TestTenant,DataType=String}"
```

tillkuhn:
Thanks @ldadams, I just ran into the same issue and your hint worked for me. Note that the [AWS CLI docs on sns publish](https://docs.aws.amazon.com/cli/latest/reference/sns/publish.html ) still don't mention that you have to use curly brackets in the Shorthand Syntax.

alam0rt:
+1 @tillkuhn , the documentation is still lacking.

kdaily:
Checking to see if this is still an issue.

joodlehammond:
For anyone ending up here from debugging SNS message attributes, the above fix works for SNS too, yet [the documentation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sns/publish.html) still states the old syntax without the curly braces.

wangshu3000:
Looks like it's a common problem for structure filed in a map. Structure filed should be wrapped with curly brace in the map.
Looking into it.

aemous:
Fix released to versions `1.44.25` and `2.33.7`. I confirmed the SNS publish docs have been updated with the fix. The shorthand syntax examples are now correct.

Issue #5129, opened 2020-04-16
SNS publiish documentation incorrect


alam0rt:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) and the [API reference](https://docs.aws.amazon.com/cli/latest/reference/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-cli/issues) and didn't find any solution
  
**Describe the bug**
The documentation [here](https://docs.aws.amazon.com/cli/latest/reference/sns/publish.html) fails to mention the correct short-hand syntax for the `aws sns publish` subcommand.

**SDK version number**
aws-cli/1.16.300 Python/2.7.16 Linux/4.14.171-136.231.amzn2.x86_64 botocore/1.13.36

**To Reproduce (observed behavior)**
incorrect (as per documentation):
aws sns publish ... --message-attributes 'KeyName1=DataType=string,StringValue=string'

correc (notice DataType needs to be 'String' rather than 'string'):
aws sns publish ... --message-attributes 'KeyName1={DataType=String,StringValue=string}'

**Expected behavior**
Expect the message to be published with the correct message attributes

**Logs/output**
Invalid type for parameter MessageAttributes.InstanceId, value: DataType=string, type: <type 'unicode'>, valid types: <type 'dict'>

nateprewitt:
Hi @alam0rt,

Thanks for opening an issue on this! We're currently tracking this for a couple spots in the docs where this has cropped up. Unfortunately, these shorthand examples are autogenerated so it's not as straight forward as updating a line in the docs.

Once we're able to get that resolved, the docs should start displaying these correctly. Let's go ahead and use this as the tracking issue for now.

wangshu3000:
Tried to fix this in this PR. Can anyone please have a review?
https://github.com/aws/aws-cli/pull/7286/files

erpel:
5 years later and I still need to spend a lot of time googling to find this report with the correct examples.
If updating the autogenerated examples is not easy, at least put a note into the freeform text, hinting at the mistake in the autogenerated example.

aemous:
Fix released to versions `1.44.25` and `2.33.7`. I confirmed the SNS publish docs have been updated with the fix. The shorthand syntax examples are now correct.

PR #7286, opened 2022-09-21
Fix map shorthand docs issue, struct in a map should be wrapped in cu…

wangshu3000:
*Issue #:* #3637

*Description of changes:*

This PR is to fix a shorthand doc issue. The struct type in the map should be wrapped in curly braces {}. 

Doc url: https://docs.aws.amazon.com/cli/latest/reference/sqs/send-message.html

Before the change: 
`KeyName1=DataType=string,StringValue=string,BinaryValue=blob`

After the change:
`KeyName1={DataType=string,StringValue=string,BinaryValue=blob}`

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -503,7 +503,11 @@ class ParamShorthandDocGen(ParamShorthand):
 
     def _map_docs(self, argument_model, stack):
         k = argument_model.key
-        value_docs = self._shorthand_docs(argument_model.value, stack)
+        stack.append(argument_model.key)
+        try:
+            value_docs = self._shorthand_docs(argument_model.value, stack)
+        finally:
+            stack.pop()
         start = 'KeyName1=%s,KeyName2=%s' % (value_docs, value_docs)
         if k.enum and not stack:
             start += '\n\nWhere valid key names are:\n'
--- a/tests/unit/test_argprocess.py
+++ b/tests/unit/test_argprocess.py
@@ -556,6 +556,14 @@ class TestDocGen(BaseArgProcessTest):
         )
         self.assert_generated_example_contains(argument, expected_example_str)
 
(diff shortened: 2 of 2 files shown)


wangshu3000:
Can anybody please have a review? Is this repo still open to external contribution? Thanks.

brianwarner:
Hi @kdaily and @dlm6693, would it be possible to have a look at this?

Me042589:
Yes. I will do the math and get back to you

tim-finnigan:
Thanks for creating this PR and for your patience here. Linking related issues/documentation:
* https://github.com/aws/aws-cli/issues/3637 - [sqs send-message]()
* https://github.com/aws/aws-cli/issues/5129 - [sns publish]()
* https://github.com/aws/aws-cli/issues/7433 - [iot update-event-configurations]()

So taking `sqs send-message` for example, the Shorthand Syntax in the documentation currently looks like:

image

And with your changes:

image

And that looks like the right format to use, for example this works for me:
```
aws sqs send-message --queue-url https://sqs.us-west-2.amazonaws.com/account-id/test --message-body TestBody --message-attributes "TenantID={StringValue=TestTenant,DataType=String},TenantName={StringValue=Bob,DataType=String}"
````

---

I brought this up for discussion with the team, and there were a couple more things that we wanted to look into prior to submitting a fix. For example, we want to test that the text still renders correctly when running `aws sqs send-message help` in different shells for example, which it looks like it does at least in macOS Terminal:

image

So from what I can tell these changes look good, but we may just need to verify in a couple more places prior to approving the PR.

aemous:
Thank you for your submission. This PR has been superseded by https://github.com/aws/aws-cli/pull/9982. 

The issue is now already fixed as of versions `1.44.25` and `2.33.7`. Closing PR.

PR #7320, opened 2022-10-02
Spell macOS consistently

rjc:


g23polar (commented):
Looks good, ready for approval

tim-finnigan:
Thanks for creating this PR and your patience here. However, I think a more comprehensive approach is needed to address this and similar issues. I think we would want to break this down to a more granular level, by misspellings and filetype, for example: inconsistencies in the code (like the [help](https://github.com/aws/aws-cli/blob/a2c46ba4056956cabf70b40ad8256b756cecc1ba/awscli/help.py#L112) command) and CLI examples:

image

image

The CHANGELOG files would likely not be changed, because those are release notes from service teams that are used across SDKs.

If still interested in this, please create an issue for further discussion on possible paths forward here.

PR #7358, opened 2022-10-24
Register disable_signing first to ensure proper usage

nateprewitt:
This PR will ensure we register the `disable_signing` event handler first when the flag `--no-sign-request` global arg is passed. This is to avoid race conditions with other handlers for a response when choosing an appropriate signer.


kyleknap (approved):

jonemo (approved):

jonemo (approved):

St0nersdash (commented):
.

PR #7397, opened 2022-11-02
Update wizard role workflow with additional prompts

jamesls:
* Adds a "this account" as a trusted entity to match the behavior of the AWS console's create role workflow.
* Update the CLI profile creation to support credential sources in addition to just the `source_profile`.
* Unable the use of shell vars, e.g. `$ROLE_NAME`, as prompt values. This is useful when saving the preview output as a shell script. Double quotes are not valid in a role name so we don't need to use single quotes here.

### DIFF
--- a/.changes/next-release/enhancement-Wizards-38242.json
+++ b/.changes/next-release/enhancement-Wizards-38242.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Wizards",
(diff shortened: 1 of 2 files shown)


kyleknap (approved):
Looks good. It worked well when I tried it out. Let's also make sure to add a enhancement changelog entry before merging. :ship:

kyleknap (approved):

PR #7398, opened 2022-11-02
Add export-creds command to the CLI

### LINKED ISSUE #5261: Provide way to get current credentials (AWS SDKs do not support SSO)

### LINKED ISSUE #7388: Add command to export credentials
### Describe the feature

This is attempting to drive closure to existing feature requests/PRs by proposing the exact feature set we should support for exporting credentials.  This proposes an `aws configure export-creds` command that resolves the existing AWS CLI credential chain and formats the credentials in various formats, including as environment variables and in the credential process format, allowing the AWS CLI to be used as a credential process.

### Use Case

For context, original issue:

* [Provide way to get current credentials](https://github.com/aws/aws-cli/issues/5261)
* Current PR with partial implementation: https://github.com/aws/aws-cli/pull/6808

To save you a few clicks, the rationale for this feature:

* While we strive for parity across our tools, there are times when the AWS CLI might support certain credential providers before other tools.  Customers would like the ability to leverage the credential resolution process of the CLI in other SDKs and tools.
* Even with credential provider parity, there may be differences in implementations and configurations across tools.  A common one that comes up, and is referenced in the linked issue, is that the CLI by default uses file based caching for credentials, where often times in SDKs you need to provide your own caching provider and explicitly provide the type of cache you want.   Customers want a way want to reuse the existing caching mechanism/defaults in place in the CLI without having to configure each individual tool.
* Customers may sometimes need to interact with frameworks, tools, or services that only support a subset of credential providers, the most common one being environment variables.  Customers would like the ability to export credentials in these formats, e.g. environment variables, to interop with other tools.
* There are several third party tools that provide this, or similar, functionality, but this is a crucial/core enough feature that we really should be providing a "built-in" mechanism that customers can utilize.
* This also gives you an easy way to consolidate all your credential related configuration in a single profile that you can use across multiple profiles.  Admittedly, you can see this as a workaround for supporting "profile inheritance/mixins" in the config file (which has come up in various issues throughout the years), and I see that we've provided credential provider specific solutions to this (I see botocore has an `sso_session` section it supports), but it still provides a succinct way to reduce duplication in your AWS config file.

### Proposed Solution

I propose we move forward with https://github.com/aws/aws-cli/pull/6808 with a few additional changes.

* We will add an `aws configure export-creds` command that will use the AWS CLI's credential resolution process and output the results in various formats.
* The command will support a `--format` parameter that supports common formats the customers use for exporting credentials.
* The command will support outputting credentials in the output required by the [credential process feature](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html).  This enables the AWS CLI to be used as a credential process.

## Synopsis

```
aws configure export credentials [--format process|env|powershell|windows-cmd]
```

The default value will be `process` which outputs the credentials in the output required by a credential process.
We'll support the following output formats, which are the formats we see with SSO, or other console related tooling that allows you to export creds in various formats:

* `process` - The default output format, enables usage as a `credential_process` source.
* `env` - Exports env vars as in a posix compatible shell (`export AWS_ACCESS_KEY_ID=...`)
* `powershell` - Export env vars in powershell specific format (`$Env:AWS_ACCESS_KEY_ID="..."`)
* `windows-cmd` - Export env vars in a windows cmd console format (`set AWS_ACCESS_KEY_ID=...`)

## Examples

To use this command as a credential process in your config file:

```
[profile my-sso-config]
#  All the relevant sso config

[profile dev]
region = us-west-2
# Other non-cred related configuration
credential_process = aws configure export-creds --profile my-sso-config

[profile staging]
region = us-west-2
# Other non-cred related config for staging
credential_process = aws configure export-creds --profile my-sso-config
```

Then use just like any other profile:

```
$ aws s3 ls --profile dev
$ aws s3 ls --profile staging
```

To use this with env vars in a posix shell (bash/zsh etc):

```
$ eval $(aws configure export-creds --profile dev)
$ aws configure list
access_key     ****************ABCD              env       # <----- Using the sourced env vars
secret_key     ****************EFGH              env       # <----- Using the sourced env vars
    region                us-west-2      config-file    ~/.aws/config
```

I noticed we have a few tools that also export fish shell creds, not sure if it's worth supporting at the moment so I've omitted it for now, but additional formatters would be straightforward to add.  Also not supporting the "CLI config" format because if you're going to write out a new profile/section to the CLI config file, we should be recommending them to use a `credential_process` instead, which can properly handle (non-trivial) refreshes.

### Other Information

One of the implementation details we'll need to figure out is being able to provide the expiry time of credentials, which is required by the credential process (if you want to refresh credentials), and arguably should be supported by other formats as well (e.g. env vars).  For the AWS CLI, and specifically botocore, we do not provide the expiry time as a public attribute.

FWIW, I don't think I had any specific reason for not exposing the expiry time other than it was an implementation detail that wasn't relevant to consumers of credentials (I don't remember the exact timeline but I know we added assume role / refreshable credentials years before I introduced the credential process functionality).  We really only cared about expiry time to handle the bookkeeping for advisory/mandatory refreshes.

I've seen the discussion on https://github.com/boto/botocore/issues/2694, and while I think it is curious about the "not expired but maybe actually it is" expiry time specific to IMDS, I will follow up and see if there's any path forward there.  Maybe we just add it to `ReadOnlyCredentials` with an extra attribute denoting expiry time fuzziness is sufficient.

Other implementation note we'll need to be careful of is detecting recursive invocations when running CLI commands against profiles that also have the AWS CLI configured as a credential process.  We'll probably need to require you specify a different profile than the parent CLI process otherwise we'll error out, but we can explore what options we have available to us.

Last thing, I'm happy to drive this to completion and implement the remaining delta we need to get this feature complete.

### Acknowledgements

- [X] I may be able to implement this feature request

### CLI version used

v2

All of them

jamesls:
This PR builds on the interface proposed in #6808 and implements the additional features proposed in #7388.

From the original PRs, the additional features are:

* Added support for an explicit `--format` args to control the output format.
* Add support for env vars, powershell/windows vars, and a JSON format that's enables this command to be used as a `credential_process`.
* Detect, and prevent infinite recursion when the credential process resolution results in the CLI calling itself with the same command.

Closes #7388
Closes #5261

### DIFF
--- a/awscli/customizations/configure/exportcreds.py
+++ b/awscli/customizations/configure/exportcreds.py
@@ -59,87 +59,64 @@ class BaseCredentialFormatter(object):
         pass
 
 
-class BaseEnvVarFormatter(BaseCredentialFormatter):
+class BasePerLineFormatter(BaseCredentialFormatter):
 
-    _VAR_PREFIX = ''
+    _VAR_FORMAT = 'export {var_name}={var_value}'
 
     def display_credentials(self, credentials):
-        prefix = self._VAR_PREFIX
         output = (
-            f'{prefix}AWS_ACCESS_KEY_ID={credentials.access_key}\n'
-            f'{prefix}AWS_SECRET_ACCESS_KEY={credentials.secret_key}\n'
-        )
+            self._format_line('AWS_ACCESS_KEY_ID', credentials.access_key) +
+            self._format_line('AWS_SECRET_ACCESS_KEY', credentials.secret_key))
         if credentials.token is not None:
-            output += f'{prefix}AWS_SESSION_TOKEN={credentials.token}\n'
+            output += self._format_line('AWS_SESSION_TOKEN', credentials.token)
         if credentials.expiry_time is not None:
-            output += (
-                f'{prefix}AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
-            )
+            output += self._format_line(
+                'AWS_CREDENTIAL_EXPIRATION', credentials.expiry_time)
         self._stream.write(output)
 
+    def _format_line(self, var_name, var_value):
+        return self._VAR_FORMAT.format(
+            var_name=var_name, var_value=var_value) + '\n'
 
-class BashEnvVarFormatter(BaseEnvVarFormatter):
+
+class BashEnvVarFormatter(BasePerLineFormatter):
 
     FORMAT = 'env'
     DOCUMENTATION = (
         "Display credentials as exported shell variables: "
         "``export AWS_ACCESS_KEY_ID=EXAMPLE``"
     )
-    _VAR_PREFIX = 'export '
+    _VAR_FORMAT = 'export {var_name}={var_value}'
 
 
-class BashNoExportEnvFormatter(BaseEnvVarFormatter):
+class BashNoExportEnvFormatter(BasePerLineFormatter):
 
     FORMAT = 'env-no-export'
     DOCUMENTATION = (
         "Display credentials as non-exported shell variables: "
         "``AWS_ACCESS_KEY_ID=EXAMPLE``"
     )
-    _VAR_PREFIX = ''
+    _VAR_FORMAT = '{var_name}={var_value}'
 
 
-class PowershellFormatter(BaseCredentialFormatter):
+class PowershellFormatter(BasePerLineFormatter):
 
     FORMAT = 'powershell'
     DOCUMENTATION = (
         'Display credentials as PowerShell environment variables: '
         '``$Env:AWS_ACCESS_KEY_ID="EXAMPLE"``'
     )
-
-    def display_credentials(self, credentials):
-        output = (
-            f'$Env:AWS_ACCESS_KEY_ID="{credentials.access_key}"\n'
-            f'$Env:AWS_SECRET_ACCESS_KEY="{credentials.secret_key}"\n'
-        )
-        if credentials.token is not None:
-            output += f'$Env:AWS_SESSION_TOKEN="{credentials.token}"\n'
-        if credentials.expiry_time is not None:
-            output += (
-                f'$Env:AWS_CREDENTIAL_EXPIRATION="{credentials.expiry_time}"\n'
-            )
-        self._stream.write(output)
+    _VAR_FORMAT = '$Env:{var_name}="{var_value}"'
 
 
-class WindowsCmdFormatter(BaseCredentialFormatter):
+class WindowsCmdFormatter(BasePerLineFormatter):
 
     FORMAT = 'windows-cmd'
     DOCUMENTATION = (
         'Display credentials as Windows cmd environment variables: '
         '``set AWS_ACCESS_KEY_ID=EXAMPLE``'
     )
-
-    def display_credentials(self, credentials):
-        output = (
-            f'set AWS_ACCESS_KEY_ID={credentials.access_key}\n'
-            f'set AWS_SECRET_ACCESS_KEY={credentials.secret_key}\n'
-        )
-        if credentials.token is not None:
-            output += f'set AWS_SESSION_TOKEN={credentials.token}\n'
-        if credentials.expiry_time is not None:
-            output += (
-                f'set AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
-            )
-        self._stream.write(output)
+    _VAR_FORMAT = 'set {var_name}={var_value}'
 
 
 class CredentialProcessFormatter(BaseCredentialFormatter):
--- a/awscli/customizations/configure/exportcreds.py
+++ b/awscli/customizations/configure/exportcreds.py
@@ -69,6 +69,10 @@ class BashEnvVarFormatter(BaseCredentialFormatter):
         )
         if credentials.token is not None:
             output += f'export AWS_SESSION_TOKEN={credentials.token}\n'
+        if credentials.expiry_time is not None:
+            output += (
+                f'export AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
+            )
         self._stream.write(output)
 
 
@@ -83,6 +87,10 @@ class PowershellFormatter(BaseCredentialFormatter):
         )
         if credentials.token is not None:
             output += f'$Env:AWS_SESSION_TOKEN="{credentials.token}"\n'
+        if credentials.expiry_time is not None:
+            output += (
+                f'$Env:AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
+            )
         self._stream.write(output)
 
 
@@ -97,6 +105,10 @@ class WindowsCmdFormatter(BaseCredentialFormatter):
         )
         if credentials.token is not None:
             output += f'set AWS_SESSION_TOKEN={credentials.token}\n'
+        if credentials.expiry_time is not None:
+            output += (
+                f'set AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
+            )
         self._stream.write(output)
 
 
--- a/awscli/customizations/configure/exportcreds.py
+++ b/awscli/customizations/configure/exportcreds.py
@@ -58,22 +58,35 @@ class BaseCredentialFormatter(object):
         pass
 
 
-class BashEnvVarFormatter(BaseCredentialFormatter):
+class BaseEnvVarFormatter(BaseCredentialFormatter):
 
-    FORMAT = 'env'
+    _VAR_PREFIX = ''
 
     def display_credentials(self, credentials):
+        prefix = self._VAR_PREFIX
         output = (
-            f'export AWS_ACCESS_KEY_ID={credentials.access_key}\n'
-            f'export AWS_SECRET_ACCESS_KEY={credentials.secret_key}\n'
+            f'{prefix}AWS_ACCESS_KEY_ID={credentials.access_key}\n'
+            f'{prefix}AWS_SECRET_ACCESS_KEY={credentials.secret_key}\n'
         )
         if credentials.token is not None:
-            output += f'export AWS_SESSION_TOKEN={credentials.token}\n'
+            output += f'{prefix}AWS_SESSION_TOKEN={credentials.token}\n'
         if credentials.expiry_time is not None:
             output += (
-                f'export AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
+                f'{prefix}AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
             )
         self._stream.write(output)
+
+
+class BashEnvVarFormatter(BaseEnvVarFormatter):
+
+    FORMAT = 'env'
+    _VAR_PREFIX = 'export '
+
+
+class BashNoExportEnvFormatter(BaseEnvVarFormatter):
+
+    FORMAT = 'env-no-export'
+    _VAR_PREFIX = ''
 
 
 class PowershellFormatter(BaseCredentialFormatter):
@@ -134,8 +147,8 @@ class CredentialProcessFormatter(BaseCredentialFormatter):
 
 SUPPORTED_FORMATS = {
     format_cls.FORMAT: format_cls for format_cls in
-    [BashEnvVarFormatter, CredentialProcessFormatter, PowershellFormatter,
-     WindowsCmdFormatter]
+    [BashEnvVarFormatter, BashNoExportEnvFormatter, CredentialProcessFormatter,
+     PowershellFormatter, WindowsCmdFormatter]
 }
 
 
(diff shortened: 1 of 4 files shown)


jamesls:
Pushed an udpate to add more lenient recursion detection

There's a reasonable scenario where at least two levels of recursion
is reasonable.  Specifically if a user explicitly runs the
`aws configure export-creds` command with a profile that uses
a `credential_process` with a value of
`aws configure export-creds --profile other-profile`.

To handle this I've modified the env var to now track the stack of
profiles seen and ensure there's no cycles given the current profile.

I also added a final safeguard of a general recursion limit of 4.
I figured that 2 levels has valid real world scenarios, we'll give one
more level as a buffer, so fail at 4.

I don't feel super strongly about the hard limit, my rationale being
if a user wants to configure some long `export-creds` chain of 10 profiles,
it's their configuration so let them, but I can't come up with a valid
scenario where you'd want more than 4.  However, I'd be cool with
removing the hard limit if we don't think it's necessary.

benkehoe:
A couple of comments:

As mentioned by @kyleknap [over on the issue](https://github.com/aws/aws-cli/issues/7388#issuecomment-1314826370), `--env` should output the basic KEY=VALUE env format, and people can use `export $(aws configure export-creds --format env)` to export them.

The CLI tends towards verbosity, `aws configure export-credentials` feels more consistent than `export-creds`.

I think having `--format json` might be more obvious for people looking for JSON, maybe it could be an option in addition to `--format process` even though they act the same.

I don't fully understand the recursive profile detection; is it specific to this command or should it be in the credential resolver or something instead?

kyleknap:
Building on this comment:

> I think having --format json might be more obvious for people looking for JSON, maybe it could be an option in addition to --format process even though they act the same.

I'm wondering if it makes sense to just have a `--format` default that outputs the credentials as if they were a normal CLI output that can be mutated by the `--output` and `--query` parameters. So if you were to run this command with all of the defaults, it would be a JSON payload:
```
aws configure export-creds
{
    "AccessKeyId": "access_key",
    "SecretAccessKey": "secret_key",
    "SessionToken": "token",
    "Expiration": "2023-01-01T00:00:00Z"
}
```
but then you can then mutate this response to be YAML/text and use `--query` if you want to pull out a specific value.  Then for the `--format` arg we would just ignore applying `--output` and `--query` if the non-default `--format` value was provided. It seems that could be useful and make the command more reusable.

benkehoe:
> I'm wondering if it makes sense to just have a `--format` default that outputs the credentials as if they were a normal CLI output

I think that makes sense.

awscli/customizations/configure/exportcreds.py
```diff
@@ -69,6 +69,10 @@ def display_credentials(self, credentials):
         )
         if credentials.token is not None:
             output += f'export AWS_SESSION_TOKEN={credentials.token}\n'
+        if credentials.expiry_time is not None:
+            output += (
+                f'export AWS_CREDENTIAL_EXPIRATION={credentials.expiry_time}\n'
```
benkehoe, line 74: I called this `AWS_CREDENTIALS_EXPIRATION` with an S in aws-export-credentials, could we make it the same here? https://github.com/benkehoe/aws-export-credentials/blob/0c53fa6afd4123d7a28550919cea85163a067605/aws_export_credentials/aws_export_credentials.py#L228

jamesls:
> I don't fully understand the recursive profile detection; is it specific to this command or should it be in the credential resolver or something instead?

It happens if you're using the AWS CLI as a credential process (this command) and also running AWS CLI commands.  In the base case you could have just this in your `~/.aws/config` file:

```
[default]
credential_process = aws configure export-credentials
```

So running any CLI command, e.g. `aws s3 ls`, would result in asking botocore to resolve credentials, in which it invokes `aws configure export-credentials`, which in turn asks botocore to resolve credentials, which then invokes `aws configure export-credentials`, etc. until you run out of system resources to spawn more processes.

I initially had it so you could never have the `export-credentials` call itself, but there is a valid case if you have the CLI configured as a credential process while also specifically running the `aws configure export-credentials` command, you just can't have any cycles in terms of profiles visited.  There's also a cap to the number of recursive calls we'll allow.  It does mean that there are certain cases like the config below that won't work, but it doesn't seem like a realistic scenario we'd want to encourage in the first place.

Running: `aws configure export-creds --profile cycle-a` with the following config will error out:

```
[profile cycle-base]
aws_access_key_id = access_key
aws_secret_access_key = secret_key

[profile cycle-a]
credential_process = aws configure export-creds --profile cycle-b

[profile cycle-b]
credential_process = aws configure export-creds --profile cycle-c

[profile cycle-c]
credential_process = aws configure export-creds --profile cycle-d

[profile cycle-d]
credential_process = aws configure export-creds --profile cycle-base
```



> As mentioned by @kyleknap https://github.com/aws/aws-cli/issues/7388#issuecomment-1314826370, --env should output the basic KEY=VALUE env format, and people can use export $(aws configure export-creds --format env) to export them. A --env-export could be provided for convenience.

I can see supporting both, the only question I have is if it's `env` and `env-export` or `env` and `env-no-export` (or whatever name).  Things to consider:

1. Matching the expectations in the various consoles that give you temp creds (sso, etc.) their default env vars are exported.
2. We're also adding windows formats such as powershell, and `$Env:AWS_ACCESS_KEY_ID` behaves like the `export AWS_ACCESS_KEY_ID` variant in bash so we might consider consistent behavior/naming across OSes.

> The CLI tends towards verbosity, aws configure export-credentials feels more consistent than export-creds.

Yeah makes sense, I'll update.

awscli/customizations/configure/exportcreds.py
jamesls: I was basing it off of what's already supported in botocore (https://github.com/boto/botocore/pull/1187).  Looks like it's also supported in [javascript](https://github.com/aws/aws-sdk-js-v3/blob/e0697ed82d7eaa98ccc48337b516bf7f3b369e3a/packages/credential-provider-env/src/fromEnv.ts).

benkehoe: Oh dang, I missed that when I was implementing it! I'll change aws-export-credentials to conform to it, then.

benkehoe:
> I can see supporting both, the only question I have is if it's env and env-export or env and env-no-export (or whatever name).

Could go explicit for both, something like `env-raw` and `env-export`

awscli/customizations/configure/exportcreds.py
benkehoe: Might be worth documenting in here as well: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

benkehoe: Is there a commensurate `aws_credential_expiration` value allowed in the profile config? If not, should there be?

kyleknap (commented):
Looks good! Just left some comments. I also tried it out in powershell and Windows CMD and it was working well there too with regards to consuming their respective formats.

awscli/customizations/configure/exportcreds.py
```diff
@@ -0,0 +1,244 @@
...
+        if self._recursion_barrier_detected():
+            self._error_stream.write(
+                "\n\nRecursive credential resolution process detected.\n"
+                "Try setting an explicit '--profile' value in the "
+                "'credential_process' configuration and ensure there "
+                "are no cycles:\n\n"
+                "credential_process = aws configure export-credentials "
+                "--profile other-profile\n"
+            )
+            return 2
```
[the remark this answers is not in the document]
jamesls: Ended up using 253 for both.  If you have no credentials configured at all and try to run a CLI command you get `253`, so it made sense to be consistent that if `get_credentials()` returns `None`, we have the same RC with this command.

awscli/customizations/configure/exportcreds.py
```diff
@@ -0,0 +1,244 @@
...
+        result = list(csv.reader([value]))[0]
+        return result
+
+    def _run_main(self, parsed_args, parsed_globals):
+        if self._recursion_barrier_detected():
+            self._error_stream.write(
+                "\n\nRecursive credential resolution process detected.\n"
+                "Try setting an explicit '--profile' value in the "
+                "'credential_process' configuration and ensure there "
+                "are no cycles:\n\n"
```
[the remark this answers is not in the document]
jamesls: Yeah ended up having separate error messages for each case, now you get:

```
$ aws configure export-credentials --profile cycle-a

Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Maximum recursive credential process resolution reached (4).
Profiles seen: cycle-a -> cycle-b -> cycle-c -> cycle-d
```

tests/unit/customizations/configure/test_exportcreds.py
```diff
@@ -0,0 +1,296 @@
...
+                '"SecretAccessKey": "secret_key"}')
+        )
+
+    def test_detects_comma_char_with_cycle(self):
+        self.session.get_credentials.return_value = self.creds
+        # To test the round tripping of CSV writing/reading, we'll invoke
+        # the command twice to ensure however it records the profiles
+        # that it detects a cycle with names that need to be escaped/quoted.
+        self.os_env['_AWS_CLI_PROFILE_CHAIN'] = 'foo,bar'
+        self.session.get_config_variable.return_value = 'bar,baz'
```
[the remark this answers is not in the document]
jamesls: I ended up having separate error messages for each case so I updated the test to check this this is because of a cycle and not max recursion.  The cycle error message now looks like this:

```
$ aws configure export-credentials --profile cycle-a

Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Unable to retrieve credentials: Error when retrieving credentials from custom-process:
Credential process resolution detected an infinite loop, profile cycle: cycle-a -> cycle-b -> cycle-c -> cycle-a

```

jamesls:
@kyleknap I believe I got everything, ready for another look.

kyleknap (approved):
Looks good! :ship:

awscli/customizations/configure/exportcreds.py
kyleknap: @benkehoe Thanks for pointing out that the environment variable is not documented there. We will follow up on this to get it documented.

For the a profile config version of `AWS_CREDENTIAL_EXPIRATION`, the CLI/botocore does not have one. I don't think the other SDKs have one either. It could potentially make sense. We have had internal discussions about it before I believe and I think the biggest concern is around potentially reloading the shared config file and handling any edge cases in reloading the shared config file.

If this is something that you'd like to see, I'd say recommend opening a feature request in the shared SDK GitHub repository to get a discussion going around it: https://github.com/aws/aws-sdk

kyleknap:
For reference related to this comment: https://github.com/aws/aws-cli/pull/7398#issuecomment-1315630328, I talked to @jamesls about it. We decided to pass on implementing this format for the initial implementation for the following reasons:

1. It adds complexity to the command in that there are three different args to control output for the command (e.g. `--format`, `--output`, and `--query`) and the application of some of the args is dependent on the format value.
2. The default format is already in a JSON format that can be parsed out by other tools (e.g. `jq`, `jp`).
3. We can always add the format in the future in a backwards compatible way.

PR #7399, opened 2022-11-02
Add support for per sub-command aliases

### LINKED ISSUE #7386: Add support for aliases in sub-command namespaces
### Describe the feature

The AWS CLI has support for creating aliases at the "service command" level [(original PR)](https://github.com/aws/aws-cli/pull/2287), [user docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-alias.html).  This is a proposal to add aliases at any level of a command space, e.g. `aws ec2 <myalias>`, `aws ec2 wait <myalias>`, etc.

### Use Case

Initially, we chose not to support aliases at any other level within the command hierarchy, primarily for scope.  However, there are a few issues with aliases as they exist today:

* We have 300+ services which makes the top level namespace incredibly cluttered with increasing likelihood that an alias you define could overwrite an existing service or prevent a newly launched service from being accessible because a pre-existing alias overwrites it.
* It's hard to quickly find which top level commands are service commands vs. your aliases (ignoring the fact that user defined aliases are explicitly excluded from help/documentation).
* It creates an inconsistent usage between user defined aliases and built-in commands, e.g. `aws ec2-regions` could be a valid command that invokes a service specific API, but `aws ec2` or `aws iam`, which are built-in, are not valid (specifically incomplete) commands.  Being able to create aliases as the "operation" level of the command hierarchy makes aliases more consistent with built in commands, e.g. `aws cloudformation list-stacks` vs. `aws cloudformation list`.
* You have to rely on a naming convention to be able to easily group aliases for a service together, i.e. `aws iam-roles`, `aws iam-<value>`.

### Proposed Solution

In addition to the `[toplevel]` key, you can now specify additional sections in the alias file.  The sections are denoted as:

`[command (<name>\s+)+]`

or in other words `command` followed by one or more whitespace separated command names.  For example:

* `[command ec2]`
* `[command iam]`
* `[command cloudformation wait]`

The contents of each section are the same as defined in the existing `[toplevel]` section.

## Replacement rules

For internal aliases (alias values without the `!` prefix), the alias value is replaced relative to the existing command hierarchy.  In other words, the command namespace is prefixed to the alias value.  For example, given:

```
[command ec2]
regions = describe-regions --query Regions[].RegionName
```

Then the following invocations are replaced as follows:

```
# Running:

aws ec2 regions

# Is (conceptually) replaced with:
aws ec2 describe-regions --query Regions[].RegionName
```

If you want to "break out" of the defined command namespace, you can use an external alias that invokes the AWS CLI:

```
[command ec2]
instance-profiles = !aws iam list-instance-profiles
```

Additional constraints/notes:

* You can't create a brand new command namespace, you can only add to existing command namespaces.  In other words creating a section `[command james]` will not create any aliases if the `aws james` namespace does not already exist.   I like the idea, but I'm not sure we should go down that path just yet.
* A namespace (in terms of implementation) specifically refers to having a command table of subcommands.  So, for example, `aws cloudformation list-stacks` is *not* a namespace (there are no subcommands under that command, only arguments), so you couldn't create any aliases under `[command cloudformation list-stacks]`.  FWIW this wouldn't be hard to implement, but I'm not sure it's a good idea.  It would enable you to create pre configured sets of options that you can reference by an alias name, so you could do crazy stuff like:

```
[command cloudformation list-stacks]
# Note the alias value contains only options, no commands.
created = --stack-status-filter CREATE_COMPLETE --query StackSummaries[].[StackName,StackStatus] --output text
deleted = --stack-status-filter DELETE_COMPLETE --query StackSummaries[].[StackName,StackStatus] --output text
```

Invoked as:

```
aws cloudformation list-stacks created
aws cloudformation list-stacks deleted
```

which if you really wanted to customize, you could combine this with the recursive replacement of aliases and add:

```
[toplevel]
cfn = cloudformation

[command cloudformation]
stacks = list-stacks
```

which I *think*, but have not actually tried to implement this, should allow you to run things like:

```
# This is equivalent to:
# aws cloudformation list-stacks --stack-status-filter CREATE_COMPLETE --query StackSummaries[].[StackName,StackStatus] --output text

aws cfn stacks created

# This is equivalent to:
# aws cloudformation list-stacks --stack-status-filter DELETE_COMPLETE --query StackSummaries[].[StackName,StackStatus] --output text

aws cfn stacks deleted
```

On second thought, that actually is kinda awesome and I'm 99% sure this would work.  Should we support this?

### Other Information

I volunteer to implement this feature.

### Acknowledgements

- [X] I may be able to implement this feature request

### CLI version used

v2

All of them

jamesls:
The [CLI alias feature](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-alias.html)
currently supports creating aliases as top level commands, at the
same level that service commands currently reside.  For example,
given this alias file:

```
[toplevel]
iam-roles = iam list-roles
```

you can run `aws iam-roles`, which will be equivalent to
`aws iam list-roles`.  This helps reduce the amount of typing for
commonly used commands.

However, this requires all aliases be in the to level namespace,
which has hundreds of services.  This makes it hard to remember
and look up what aliases you have available.

This PR adds support for creating aliases within namespaces of
existing commands/subcommands.

There is some implementation specific details with respect to
event names that we needed to rely onin order to make aliases of nested
commands work (i.e. `aws foo bar <myalias>`).

The implementation is different from the approach taken by
the `[toplevel]` aliases for several reasons:

* The toplevel aliases make the assumption that a single parser can be
  used for any of the alias values.  This is essentially the "global
  args" parser, so this holds true for top level aliase but isn't
  guaranteed for subcommands, custom commands, etc.
* We need to ensure we scope the alias expansion to the current
  namespace we're traversing so we need an approach that enforces this.
* There may be additional logic that happens when invoking a command,
  even within a given namespace.  The safer option is to re-execute
  all this logic after expanding the alias.

So the high level idea is to take the reference to the parent command
object, and reinvoke it with the expanded alias value, after
picking off, and updating, and global parameters provided in the
alias value.

This approach also has the benefit of not modifying the existing
code (except for extract pieces out into a shared base class) so
it's conservative in terms of the likelihood of
breaking existing aliases.

Closes #7386

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -275,8 +275,6 @@ class SubCommandArgParser(ArgTableArgParser):
         return arg_table_copy
 
 
-
-
 class FirstPassGlobalArgParser(CLIArgParser):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import argparse
+import copy
 import sys
 from awscli.compat import six
 from difflib import get_close_matches
@@ -207,6 +208,73 @@ class ArgTableArgParser(CLIArgParser):
         else:
             return super(ArgTableArgParser, self).parse_known_args(
                 args, namespace)
+
+
+class SubCommandArgParser(ArgTableArgParser):
+    """Parse args for a subcommand but do not consume the arg table.
+
+    This is similar to the ArgTableArgParser, except it doesn't consume
+    any args from the provided arg table, though it will respect the
+    arg table when looking for a subcommand.
+
+    """
+
+    def parse_known_args(self, args, namespace=None):
+        parsed_args, remaining = super(
+            SubCommandArgParser, self).parse_known_args(args, namespace)
+        if getattr(parsed_args, 'subcommand', None) is not None:
+            new_args = self._remove_subcommand(args, parsed_args)
+            return new_args, parsed_args.subcommand
+        return None
+
+    def _remove_subcommand(self, args, parsed_args):
+        # We want to remove only the subcommand from the args list
+        # and keep everything else.  We should be safe to remove
+        # the first argument that matches the subcommand name.
+        #
+        # .parse_known_args() assumes that any args that it doesn't
+        # understand do not consume any values.  For example:
+        #
+        # aws ecs --cluster mycluster describe-tasks --tasks foo
+        #
+        # is not a valid command, because we ignore `--cluster`
+        # and assume that `mycluster` is a subcommand.
+        #
+        # However, this command works:
+        #
+        # aws ecs --cluster=mycluster desribe-tasks --tasks foo
+        #
+        # Therefore we don't have to worry about the case where
+        # a param *value* shadows the parsed subcommand because
+        # the existing parser would have already errored out.
+        new_args = args[:]
+        new_args.remove(parsed_args.subcommand)
+        return new_args
+
+    def _build(self, argument_table, command_table):
+        # In order to check for a subcommand while still respecting
+        # the arg table, we do need to consider the arg table, but not
+        # fail if any of the required args aren't provided.  We don't
+        # want to mutate the arg table that's provided to us, so we
+        # make a copy of it and then set all the required to not required.
+        non_required_arg_table = self._non_required_arg_table(
+            argument_table)
+        for arg_name in non_required_arg_table:
+            argument = non_required_arg_table[arg_name]
+            argument.add_to_parser(self)
+        if command_table:
+            self.add_argument('subcommand', action=CommandAction,
+                              command_table=command_table, nargs='?')
+
+    def _non_required_arg_table(self, argument_table):
+        arg_table_copy = {}
+        for key, value in argument_table.items():
+            copied = copy.copy(value)
+            copied.required = False
+            arg_table_copy[key] = copied
+        return arg_table_copy
+
+
 
 
 class FirstPassGlobalArgParser(CLIArgParser):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -683,7 +683,6 @@ class ServiceOperation(object):
         self._operation_model = operation_model
         self._session = session
         self._subcommand_table = None
-        self._subcommand_delegated = False
         if operation_model.deprecated:
             self._UNDOCUMENTED = True
 
@@ -734,7 +733,7 @@ class ServiceOperation(object):
         return self._arg_table
 
     def _parse_potential_subcommand(self, args, subcommand_table):
-        if subcommand_table and not self._subcommand_delegated:
+        if subcommand_table:
             parser = SubCommandArgParser(self.arg_table, subcommand_table)
             return parser.parse_known_args(args)
         return None
@@ -751,7 +750,6 @@ class ServiceOperation(object):
             args, subcommand_table)
         if maybe_parsed_subcommand is not None:
             new_args, subcommand_name = maybe_parsed_subcommand
-            self._subcommand_delegated = True
             return subcommand_table[subcommand_name](new_args, parsed_globals)
         operation_parser = self._create_operation_parser(
             self.arg_table, subcommand_table)
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -14,6 +14,7 @@ import json
 import os
 import platform
 import sys
+import copy
 import logging
 
 import distro
@@ -38,6 +39,7 @@ from awscli.argparser import MainArgParser
 from awscli.argparser import FirstPassGlobalArgParser
 from awscli.argparser import ServiceArgParser
 from awscli.argparser import ArgTableArgParser
+from awscli.argparser import SubCommandArgParser
 from awscli.help import ProviderHelpCommand
 from awscli.help import ServiceHelpCommand
 from awscli.help import OperationHelpCommand
@@ -681,6 +683,7 @@ class ServiceOperation(object):
         self._operation_model = operation_model
         self._session = session
         self._subcommand_table = None
+        self._subcommand_delegated = False
         if operation_model.deprecated:
             self._UNDOCUMENTED = True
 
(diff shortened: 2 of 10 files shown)


awscli/clidriver.py
```diff
@@ -764,6 +779,8 @@ def __call__(self, args, parsed_globals):
                 # method of the operation caller. It represents the return
                 # code of the operation.
                 return override
+        elif getattr(parsed_args, 'subcommand', None) is not None:
```
jamesls, line 782: There's one implementation quirk that we need to work out here about arguments
and subcommands.  Suppose we have an alias file like this:

```
[command resourcegroupstaggingapi get-resources]
mysvc = --tag-filters Key=foo,Values=bar \
    Key=bar,Values=baz \
    --query ResourceTagMappingList[].[ResourceARN] --output text
```

The way it works now is if you run the command
`aws resourcegroupstaggingapi get-resources mysvc`, we'll see that there's
an alias defined for `mysvc`, do the text replacement, and the full command
expansion is:

```
aws resourcegroupstaggingapi get-resource \
    # This part below is from the "mysvc" alias definition.
    --tag-filters Key=foo,Values=bar Key=bar,Values=baz \
    --query ResourceTagMappingList[].[ResourceARN] --output text
```

So far so good.  The tricky part happens if we run a command like this:

```
aws resourcegroupstaggingapi get-resources mysvc --resource-type-filters ecs
```

The current behavior right now is that the `--resource-type-filters ecs` is
ignored.  No errors or anything, it's as if you didn't specify it at all.
On the surface that seems counterintuitive because I'd expect the final
expansion to merge with the alias definition and be:

```

aws resourcegroupstaggingapi get-resource \
    # This part is from the "mysvc" alias definition.
    --tag-filters Key=foo,Values=bar Key=bar,Values=baz \
    --query ResourceTagMappingList[].[ResourceARN] --output text \
    # This part from user's explicitly provided args
    --resource-type-filters ecs

```

However, this is not the current behavior of our
`awscli.customizations.commands.BasicCommand`, which I mirrored in the
`ServiceOperation` update.  Essentially, our convention has been that commands
will provide a subcommand table, or an arg table, but not both.  I checked that
this is the case for all of our existing customizations that subclass
`BasicCommand`.  This means that our logic when the command invoked is:

https://github.com/aws/aws-cli/blob/f23912d0b4e72c3b847d5fd74d63d101667f38d6/awscli/customizations/commands.py#L180-L190

In other words, if a subcommand was provided, we essentially throw away
any `parsed_args` and pass the subcommand the `remaining` (i.e. unknown)
args.  However if no subcommand is provided then we call `self._run_main`
with the parsed args.

So going back to the original example, when we invoke the command object for the
`mysvc` alias definition, we throw away the `--resource-type-filters ecs` part
because it was "handled" by the `ServiceOperation`.  It was part of the `parsed_args`
for the command object.  This hasn't mattered in the past because the only command
objects that provide an `arg_table` are the "leaf commands".  That is, by convention,
we don't support scoped args such as:

```
aws foo --local-1=a bar --local-2=b baz --local3=c
```

The expectation is that all of those args are handled by the `baz` subcommand.

I have a few ideas on how we can make this work for aliases (I think), but the question is:
what should the ideal behavior be for `aws resourcegroupstaggingapi get-resources mysvc --resource-type-filters ecs`?

kyleknap: For the command:
```
aws resourcegroupstaggingapi get-resources mysvc --resource-type-filters ecs
```
I think we should respect the `--resource-type-filters ecs` arguments and not throw them away and instead merge it with the alias definition. It would definitely be confusing if we did not do this.

jamesls: In that case we have a few decisions to make.

The simplest way to address this is to first get the subcommand table, and if it's not empty, construct a parser for *just* the subcommand.  If you then find that `parsed_args.subcommand is not None`, then delegate to the subcommand and pass through the remaining args.  That way the `ServiceOperation` never consumes any known parsed args.

I can update the `ServiceOperation` to support this behavior because it never supported subcommands to begin with so there's no concern with backwards compatibility.  However, this would make the behavior inconsistent with the `awscli.customizations.BasicCommand`, from the code I linked above we can see that it will consume any parsed args you provide in its `ARG_TABLE` but still delegate to a subcommand if found.

This won't break any of the customizations that ship with the CLI, as they provide a subcommand table OR an arg table but never both, but this could potentially break existing commands that do this (though I can't imagine why they would as the args literally get thrown away if there's a subcommand).

Should I just update this for `ServiceOperation` for now?

kyleknap: Would only supporting `ServiceOperation` mean we would not be supporting merging alias definitions with arguments provided for leaf commands subclassed from `BasicCommand`? If so, I'd definitely would want us to explore getting the behavior working for custom commands too as the custom commands are fairly popular and I'd imagine customers would want to make aliases for them as well (e.g. make an S3 filter alias if you have a long set of `--include`/`--exclude` filters you commonly use).

> this could potentially break existing commands that do this

Are you referring to external plugins here? I think if any of the builtin custom commands are safe from breakage, I'm fine with making the change as plugins are still provisional in v2. This change will likely go into a minor version bump anyway.

jamesls: > Would only supporting ServiceOperation mean we would not be supporting merging alias definitions with arguments provided for leaf commands subclassed from BasicCommand?

Yeah, I was hoping to minimize amount of files changed, but we'd need to update `BasicCommand` to support this which also makes this a riskier change.

> Are you referring to external plugins here? 

Yeah it would only affect external plugins.

I'll play around with different ideas, I'm starting to suspect this implementation might get a bit tricky.

kyleknap (commented):
The approach looks good so far. I wrote a bunch of aliases in testing it out and did not run into any unexpected behavior. Just had some small comments until we get to addressing merging command line provided arguments to leaf custom commands.

awscli/argparser.py
```diff
@@ -209,6 +210,73 @@ def parse_known_args(self, args, namespace=None):
...
+                              command_table=command_table, nargs='?')
+
+    def _non_required_arg_table(self, argument_table):
+        arg_table_copy = {}
+        for key, value in argument_table.items():
+            copied = copy.copy(value)
+            copied.required = False
+            arg_table_copy[key] = copied
+        return arg_table_copy
+
```
kyleknap, line 276: It looks like there is two additional new lines here.

awscli/customizations/commands.py
```diff
@@ -136,6 +143,13 @@ def __call__(self, args, parsed_globals):
             ".".join(self.lineage_names)
         self._session.emit(event, argument_table=self._arg_table, args=args,
                            session=self._session)
+        maybe_parsed_subcommand = self._parse_potential_subcommand(
+            args, self._subcommand_table
+        )
+        if maybe_parsed_subcommand is not None:
```
kyleknap, line 149: With this if statement now added, does this line ever get used? https://github.com/aws/aws-cli/pull/7399/files#diff-ae84646454f40103f091a890d7a1f826eaebc333454d1a5772ce31c721881cd6R211 It seems like it would not (e.g. in the `s3` subcommand).

kyleknap (commented):
Looks fine. Just had a couple more comments and questions

kyleknap (approved):
Looks good :ship:

PR #7418, opened 2022-11-09
Endpoint provider

stealthycoin:
Pulling this PR into cli v2.

https://github.com/boto/botocore/pull/2737
Also pulls in: https://github.com/boto/botocore/pull/2762 and https://github.com/boto/botocore/pull/2767


kyleknap (commented):
Looks fine. For this PR, we should also consider pulling in this PR: https://github.com/boto/botocore/pull/2767 as it looks like it makes some standalone improvements to the tests that appear to be circled back on. We can probably just add that port as a separate commit to this PR.

kyleknap (approved):
Looks good. :ship:

PR #7448, opened 2022-11-17
Updates cli v2 to use new endpoints 2.0 resolution

stealthycoin:
The code in this branch limits the release to the services identifiers s3 and s3control.

High level overview over the code changes:

* Introduces a new class EndpointRulesetResolver in regions.py. This acts as middleware between a botocore Client object and the EndpointProvider which parses endpoint ruleset files. Endpoint ruleset files are in botocore/data/**/endpoint-rule-set-1.json for all services for which a ruleset is available.

* Updates to the Client object functions that make or presign requests.

* Removal of S3EndpointSetter, S3ControlEndpointSetter, S3ControlArnParamHandler (the old code remains in place for third party importers but is unused)

* New versions of S3ControlArnParamHandler and S3RegionRedirector (the old code remains in place for third party importers, the new versions have a 2 suffix in their names)

* Refactoring of S3-related event handlers to account for the EndpointProvider as new source of endpoint and signing information


tests/unit/botocore/test_utils.py
```diff
@@ -1721,15 +1816,6 @@ def test_no_redirect_from_error_for_accesspoint(self):
             request_dict, response, self.operation)
         self.assertEqual(redirect_response, None)
 
-    def test_no_redirect_from_cache_for_accesspoint(self):
```
kyleknap, line 1724: I think this test got repurposed to be this one: https://github.com/boto/botocore/pull/2785/files#diff-a28b281ee0902c932527e269c585ac109bdd98aa5ceebe05ca97a10a5a432651R1888. We should probably update this test case as well instead of delete the test.

tests/unit/botocore/test_endpoint_provider.py
```diff
@@ -343,3 +349,86 @@ def test_ruleset_unknown_parameter_type_raises(partitions):
...
+    at, sc = empty_resolver.auth_schemes_to_signing_ctx(auth_schemes)
+    assert at == 'v4'
+    assert sc == {
+        'region': 'my-region-1',
+        'signing_name': 'dynamodb',
+        'disableDoubleEncoding': True,
+    }
+
+
+@requires_crt()
```
kyleknap, line 391: Let's update these tests to remove any CRT specific variable setting or test skipping since CRT is always available for this environment.

tests/functional/botocore/test_s3.py
```diff
@@ -1811,8 +1794,7 @@ def test_presign_unsigned(self):
         config = Config(signature_version=botocore.UNSIGNED)
         client = self.session.create_client('s3', 'us-east-2', config=config)
         url = client.generate_presigned_url(ClientMethod='list_buckets')
-        self.assertEqual(
-            'https://s3.us-east-2.amazonaws.com/', url)
+        self.assertEqual("https://s3.amazonaws.com/", url)
```
kyleknap, line 1797: This should still have `us-east-2` in the url

kyleknap (commented):
Looks good! I did a side by side comparison and looks like we got all of it. Most of my comments were about addressing v1 and v2 differences in this port.

kyleknap (approved):
Looks good :ship:

PR #7464, opened 2022-11-22
Remove or rename duplicate tests

stealthycoin:
A few tests were shadowing other tests. Some were simple copy pastes
of other tests, which I removed. Others had unique functionality but
their name was being shadowed so they were never getting run, these I
renamed.


kyleknap (approved):
:ship:

Issue #7434, opened 2022-11-14
aws codeartifact login can't set npm registry


barabadzhi:
### Describe the bug

Since last upgrade, I'm no longer able to set npm registry with `aws codeartifact login`.
Command I execute `aws codeartifact login --tool npm --repository XXX --domain XXX`.

### Expected Behavior

Typical `You're logged in` message as it used to be.
Npm pulls codeartifact as usual.

### Current Behavior

Short:
`Command '['npm', 'config', 'set', '//XXX-195852102923.d.codeartifact.eu-west-1.amazonaws.com/npm/XXX/:always-auth', 'true']' returned non-zero exit status 1.`

Long (`--debug`):
<details>
  <summary>Debug output</summary>
  ➜ aws codeartifact login --tool npm --repository XXX --domain XXX --debug
2022-11-14 14:21:34,263 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/2.8.12 Python/3.10.8 Darwin/22.1.0 source/arm64
2022-11-14 14:21:34,263 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['codeartifact', 'login', '--tool', 'npm', '--repository', 'XXX', '--domain', 'XXX', '--debug']
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function add_s3 at 0x104f2cee0>
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function add_ddb at 0x104d6ac20>
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <bound method BasicCommand.add_command of <class 'awscli.customizations.configure.configure.ConfigureCommand'>>
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function change_name at 0x104a90f70>
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function change_name at 0x104a91ea0>
2022-11-14 14:21:34,274 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function alias_opsworks_cm at 0x104f2f7f0>
[...]

PR #7468, opened 2022-11-23
Ignore errors setting always-auth in npm config set

kdaily:
*Issue #, if available:*

#7434 

*Description of changes:*

The release of NPM 9 has dropped the support of deprecated or unsupported parameters in the `npm config set` command. This includes the `always-auth` parameter. Since `always-auth=true` is being set internally via the `aws codeartifact login --tool npm` command, this results in users of NPM 9 getting an exception. This blocks users to use login command entirely with NPM 9. This flag is ignored by NPM 7 and 8. Since the NPM login command is also used for yarn users, it was not previously removed.

This change is to catch and ignore an exception while setting `always-auth` flag for the `npm config set` command. Users with NPM < 9 will still set the parameter in their `.npmrc` file, while users with NPM >= 9 will not.

### DIFF
--- a/awscli/customizations/codeartifact/login.py
+++ b/awscli/customizations/codeartifact/login.py
@@ -78,20 +78,26 @@ class BaseLogin(object):
             return
 
         for command in commands:
-            try:
-                self.subprocess_utils.check_call(
-                    command,
-                    stdout=self.subprocess_utils.PIPE,
-                    stderr=self.subprocess_utils.PIPE,
-                )
-            except OSError as ex:
-                if ex.errno == errno.ENOENT:
-                    raise ValueError(
-                        self._TOOL_NOT_FOUND_MESSAGE % tool
-                    )
-                raise ex
+            self._run_command(tool, command)
 
         self._write_success_message(tool)
+
+    def _run_command(self, tool, command, *, ignore_errors=False):
+        try:
+            self.subprocess_utils.check_call(
+                command,
+                stdout=self.subprocess_utils.PIPE,
+                stderr=self.subprocess_utils.PIPE
+            )
+        except subprocess.CalledProcessError as ex:
+            if not ignore_errors:
+                raise ex
+        except OSError as ex:
+            if ex.errno == errno.ENOENT:
+                raise ValueError(
+                    self._TOOL_NOT_FOUND_MESSAGE % tool
+                )
+            raise ex
 
     @classmethod
     def get_commands(cls, endpoint, auth_token, **kwargs):
@@ -302,6 +308,10 @@ class NpmLogin(BaseLogin):
             self.repository_endpoint, self.auth_token, scope=scope
         )
         self._run_commands('npm', commands, dry_run)
+
+    def _run_command(self, tool, command):
+        ignore_errors = any('always-auth' in arg for arg in command)
+        super()._run_command(tool, command, ignore_errors=ignore_errors)
 
     @classmethod
     def get_scope(cls, namespace):
--- a/tests/functional/codeartifact/test_codeartifact_login.py
+++ b/tests/functional/codeartifact/test_codeartifact_login.py
@@ -587,6 +587,30 @@ password: {auth_token}'''
         self._assert_operations_called(package_format='npm', result=result)
         self._assert_dry_run_execution(self._get_npm_commands(), result.stdout)
 
+    def test_npm_login_always_auth_error_ignored(self):
+        """Test login ignores error for always-auth.
+
+        This test is for NPM version >= 9 where the support of 'always-auth'
+        has been dropped. Running the command to set config gives a non-zero
+        exit code. This is to make sure that login ignores that error and all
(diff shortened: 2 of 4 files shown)


tests/unit/customizations/codeartifact/test_adapter_login.py
```diff
@@ -450,6 +451,32 @@ def test_login(self):
             expected_calls, any_order=True
         )
 
+    def test_login_always_auth_not_allowed(self):
+        def side_effect(command, stdout, stderr):
```
kdaily, line 455: Can you describe what this is testing for? I'm not sure I follow the logic why the mock reuses the logic of the `_run_command` function to raise an error. I think that to improve the readability and understanding of the tests, we should explicitly be testing the scenarios of using an old or new version of NPM, and if `always-auth` is provided or not. You might need to provide a separate class with a new `setUp` to achieve this.

awscli/customizations/codeartifact/login.py
```diff
@@ -303,6 +309,12 @@ def login(self, dry_run=False):
         )
         self._run_commands('npm', commands, dry_run)
 
+    def _run_command(self, tool, command):
```
kdaily, line 312: We can simplify the logic here since `ignore_errors` is a boolean. Then, we do not need an if statement doing the work for us. We can do something like this:

```
ignore_errors = any('always-auth' in arg for arg in command)
super()._run_command(tool, command, ignore_errors=ignore_errors)
```

Separately, is it possible for us to be more specific in checking where `always-auth` occurs? What scenarios could it be in different (or multiple) positions?

awscli/customizations/codeartifact/login.py
```diff
@@ -78,21 +78,27 @@ def _run_commands(self, tool, commands, dry_run=False):
...
-                if ex.errno == errno.ENOENT:
-                    raise ValueError(
-                        self._TOOL_NOT_FOUND_MESSAGE % tool
-                    )
-                raise ex
+            self._run_command(tool, command)
 
         self._write_success_message(tool)
 
+    def _run_command(self, tool, command, *, ignore_errors=False):
```
kdaily, line 85: I don't think the `*` is going to be doing anything here. If we needed to change this function in the future, any new parameters would need to be added to the end of the function for compatibility (even though this is technically a private method).

Is there something that I'm missing that this is needed for?

tests/functional/codeartifact/test_codeartifact_login.py
```diff
@@ -587,6 +587,23 @@ def test_npm_login_without_domain_owner_dry_run(self):
         self._assert_operations_called(package_format='npm', result=result)
         self._assert_dry_run_execution(self._get_npm_commands(), result.stdout)
 
+    def test_npm_login_always_auth_not_allowed(self):
+        def side_effect(command, stdout, stderr):
```
kdaily, line 591: Similar to the comment about the unit test, I'm not exactly clear what scenario this is testing, or the need for the `if` statement in the side effect.

kdaily (commented):
Thanks for this work so far! I think we've agreed that the approach to catch and ignore errors for the commands using `always-auth` is an acceptable solution. You still handle OS errors for missing the `npm` binary, which is great.

There are some concerns around the testing that we can address to make sure that users understand what is being tested, and we're covering all scenarios.

awscli/customizations/codeartifact/login.py
MitaliBo: I can make the first change. For second question, we only have one command in npm which has "always-auth" (particular keyword) setting and we are only checking if the command we are going to run has it, i am not sure if i understand the question "what scenarios could it be in different positions?"

tests/unit/customizations/codeartifact/test_adapter_login.py
MitaliBo: Since we do not want to have any version dependent code, so in order to test the functionality that all our commands for NPM login work fine even if get an exception from the command to set 'always-auth=true' which would be the case in NPM 9 version, this test case is mocking to raise an exception for that particular command by setting in the side-effect and than running login command. Our existing test case 'test_login' covers all NPM versions < 9 while this one 'test_login_always_auth_not_allowed' covers NPM >=9

tests/functional/codeartifact/test_codeartifact_login.py
MitaliBo: the need of `if` statement is to mock the exception which will be raised when running command to set `always-auth` flag for NPM version >=9 and to test that even though exception is raised login command works fine

awscli/customizations/codeartifact/login.py
MitaliBo: `*` makes it keyword parameter rather than positional. So any new positional param would need to add before * and any new required keyword param after it based on requirement. Are you suggesting it to be not keyword param?

stealthycoin (approved):

MitaliBo:
Changes look good.

PR #7540, opened 2022-12-20
New CLI examples for apitgatewayv2, ec2, kms, rds

elysahall:


awscli/examples/ec2/copy-image.rst
```diff
@@ -31,7 +31,27 @@ The following ``copy-image`` command copies the specified AMI from the ``us-west
...
+
+**Example 3: To include your user-defined AMI tags when copying an AMI**
+
+The following ``copy-image`` command uses the ``--copy-image-tags`` parameter to copy your user-defined AMI tags when copying the AMI. ::
+
+    aws ec2 copy-image \
+        --region us-east-1 \
+        --name ami-name \
+        --source-region us-west-2 \
+        --source-image-id ami-066877671789bd71b \
```
kdaily, line 47: Other image IDs were changed to `ami-0123456789abcdefg`, should this one be as well?

julieso: Yes, it would be good to be consistent. Although, "g" isn't a valid character in an AMI ID, so perhaps changing them all to ami-0123456789abcdef1 would be better.

randyurbano:
kdaily@ Can you tell me why this hasn't been merged?

awscli/examples/ec2/copy-image.rst
kdaily: Looks like the name previously used in other docs is `ami-0abcdef1234567890`. Proposing to change it to that.

kdaily: Actually, I think this might not be the correct change as I don't think the AMI ID here is the same as the output one that was changed to `ami-0123456789abcdefg`. I will open a separate issue to address the `g` in the output AMI though.

kdaily (approved):

PR #7063, opened 2022-06-24
[v2] Bump Python version ceiling to 3.10

kdaily:
*Issue #, if available:*

#7004 

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #7557, opened 2022-12-30
Upgrade to Python version 3.11.1

hssyoo:
This commit upgrades the bundled Python interpreter from version 3.9.11 to 3.11.1. To support this upgrade, this commit adds 3.11 to the GitHub actions testing matrix and bumps PyInstaller to version 5.7.0. It also removes the cgi module deprecated in 3.11 and replaces it with email.message.


nateprewitt (approved):

pyproject.toml
```diff
@@ -76,7 +77,10 @@ filterwarnings = [
     # next patch. This filter exception should also be removed once it's
     # confirmed that tests are passing again in the latest Python 3.10.
     # https://github.com/python/cpython/issues/100160
-    'default:There is no current event loop:DeprecationWarning'
+    'default:There is no current event loop:DeprecationWarning',
+
+    # https://github.com/pytest-dev/pytest-xdist/issues/825
```
kyleknap, line 82: It would be great if we can add a comment here on what we can do to remove this exception similar to the other two exceptions that we added to the `filterwarnings` list. Mainly it is not clear from the referenced GitHub issue what we need to do and ideally we remove this in this exception in the future.

awscli/botocore/utils.py
```diff
@@ -2747,10 +2747,12 @@ def get_encoding_from_headers(headers, default='ISO-8859-1'):
...
 
-    content_type, params = cgi.parse_header(content_type)
+    message = email.message.Message()
+    message['content-type'] = content_type
+    charset = message.get_param("charset")
 
-    if 'charset' in params:
-        return params['charset'].strip("'\"")
+    if charset is not None:
+        return charset
```
kyleknap, line 2755: We should make sure to add the corresponding test case from this update as well: https://github.com/boto/botocore/pull/2794

requirements-build.txt
```diff
@@ -1,4 +1,4 @@
 # Requirements we need to run our build jobs for the installers.
 # We create the separation for cases where we're doing installation
 # from a local dependency directory instead of requirements.txt.
-PyInstaller==5.3
+PyInstaller==5.7.0
```
kyleknap, line 4: Let's decouple upgrading PyInstaller from this change. Mainly we should make this change scoped to purely adding source support for Python 3.11 as the PyInstaller update will need additional vetting and potentially additional code changes.

.github/workflows/run-tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
```
kyleknap, line 13: We should also add an `enhancement` changelog entry for this. We can probably copy the wording from this PR: https://github.com/aws/aws-cli/pull/7063

requirements-test.txt
```diff
@@ -1,6 +1,6 @@
 jsonschema==4.7.2
 mock==1.3.0
-pytest==6.2.5
-coverage==5.5
-pytest-cov==2.12.1
-pytest-xdist==2.4.0
+pytest==7.2.0
+coverage==7.0.1
+pytest-cov==4.0.0
+pytest-xdist==3.1.0
```
kyleknap, line 6: We do not need to do this as part of this PR, but we should enable dependabot for these dependencies so that we get auto-updates for them. Here is an example of where we've enabled dependabot for a dependency before: https://github.com/aws/aws-cli/pull/7521

kyleknap (commented):
Looks good. Just had some comments. We should also update the commit to make it specific to adding source support for Python 3.11. We are a few steps away from actually updating the bundled Python interpreter to 3.11

requirements-test.txt
nateprewitt: This will need to be a prerequisite to this PR if not done here. I believe there are some incompatibilities with older versions of pytest and 3.11.

https://github.com/pytest-dev/py/issues/273
https://github.com/pytest-dev/pytest/issues/9181

requirements-build.txt
kyleknap: Hmm it looks like PyInstaller has a ceiling version pin on Python for the version of PyInstaller that we use and it excludes Python 3.11. We will probably need to upgrade PyInstaller first before we can get CI passing here.

kyleknap (approved):
Looks good to me :ship:

PR #7565, opened 2023-01-04
Add test to ensure configure file is up to date

stealthycoin:
- Add test to ensure that the configure file is synced
- Add test to ensure autoreconf is version 2.7.31
- Regenerates the configure file so the test passes


configure
```diff
@@ -1239,6 +1242,11 @@ Installation directories:
 Optional Packages:
   --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
   --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
+  --with-python-sys-prefix
```
kyleknap, line 1245: I'm curious on where these new options came from and whether we want to document them? It is strange that it looks like the previous `configure` file was generated using 2.71 of autotools but it was not included before. I think I'd prefer not to document these variables by using an `m4` override like we already have for the current help documentation.

kyleknap (commented):
Looks fine. I just had a question on the newly generated `configure` file

PR #7571, opened 2023-01-06
Add test to ensure that the lock files are up to date

stealthycoin:


PARIWAT198 (commented):
if according to regulations We can only go to the measure, right? And then, I'm alone doing business in a remote area. Can I expand beyond the remit?

scripts/regenerate-lock-files
```diff
@@ -85,28 +94,35 @@ def delete_file(path: Path):
         pass
 
 
-def main():
+def main(base_directory: str):
     build_lock_file(
         sources=[os.path.join("download-deps", "bootstrap.txt")],
-        output=os.path.join("download-deps", "bootstrap"),
+        output=os.path.join(base_directory, "download-deps", "bootstrap"),
```
kyleknap, line 100: It seems off on how we are using the `output` parameter. Based on the code, it seems it will be always rooted in the source directory because the `_output_path()` function always uses `REQ_DIR` as the root. Instead it seems we should be making the `base_directory` the root. From there we should be able to get around the Windows `relpath` issues by calling `os.path.relpath()` on the full output path and the output dir.

stealthycoin: It's not rooted in the source directory if you want to generate the output elsewhere, like in the test.

scripts/regenerate-configure
```diff
@@ -0,0 +1,11 @@
+import subprocess
```
kyleknap, line 1: I believe this file is for a different PR?

kyleknap (commented):
Looks good. I just had a small comment on an extra file that looks like got added

kyleknap (approved):
Looks good :ship:

kyleknap (approved):

PR #7595, opened 2023-01-17
Update source distribution proposal

stealthycoin:
Update the source distribution proposal.


kyleknap (commented):
Looks fine. Let's make sure to have this PR target the feature branch instead of v2 so the proposal will be marked final in the final PR that merges support for the source distribution

kyleknap (approved):
:ship:

PR #7598, opened 2023-01-18
Update --with-download-deps help

stealthycoin:
Update `--with-download-deps` help.

### DIFF
--- a/m4/awscli_help.m4
+++ b/m4/awscli_help.m4
@@ -34,9 +34,10 @@ Optional arguments:
                           "portable-exe", "system-sandbox" (default is
                           "system-sandbox")
   --with-download-deps    Download all dependencies and use those when
-                          building the AWS CLI. If not specified, the
-                          dependencies (including all python packages) must be
-                          installed on your system
+                          building the AWS CLI. Note that this is best effort
+			  and has no backwards compatibility guarantee. If not
+			  specified, the dependencies (including all python
+			  packages) must be installed on your system
 Some influential environment variables:
   PYTHON      the Python interpreter
 m4_divert_pop([HELP_BEGIN])dnl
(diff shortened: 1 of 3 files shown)


m4/awscli_help.m4
```diff
@@ -33,8 +33,9 @@ Optional arguments:
                           Specify type of AWS CLI installation. Options are:
                           "portable-exe", "system-sandbox" (default is
                           "system-sandbox")
-  --with-download-deps    Download all dependencies and use those when
-                          building the AWS CLI. If not specified, the
+  --with-download-deps    Download all Python dependencies and use those when
+                          building the AWS CLI. Note that this is best effort
```
kyleknap, line 37: It would probably be good if we add the additional note of "nor continue to work in between versions of the AWS CLI" to better indicate that even if you get it working it may not continue to work in future versions.

m4/awscli_help.m4
```diff
@@ -33,8 +33,9 @@ Optional arguments:
                           Specify type of AWS CLI installation. Options are:
                           "portable-exe", "system-sandbox" (default is
                           "system-sandbox")
-  --with-download-deps    Download all dependencies and use those when
-                          building the AWS CLI. If not specified, the
+  --with-download-deps    Download all Python dependencies and use those when
```
kyleknap, line 36: It would be great if we can indicate that currently it only downloads Python dependencies but this option might be expanded in the future to other dependencies. Maybe we can add the disclaimer at the end of the help that the dependencies downloaded could be expanded beyond Python dependencies?

kyleknap (commented):
Looks good. I just had a couple of comments

m4/awscli_help.m4
stealthycoin: I think it makes more sense to just document what it does now. If we change it we change the string.

kyleknap (commented):
I think we need to regenerate the `configure` file but the wording looks good

Issue #4778, opened 2019-12-18
"'charmap' codec can't encode character" fetching CloudWatch logs and redirecting to a file on Windows 10


aschmied:
The AWSCLI aborts and writes invalid JSON when I fetch logs containing non-ASCII characters and redirect the output to a file in Windows 10 cmd.exe. This looks related to #1070, but in this case it is log messages causing the problem, not filenames.

**Steps**

1. Create a test log group and stream:

        aws logs create-log-group --log-group-name anthony-test-group
        aws logs create-log-stream --log-group-name anthony-test-group --log-stream-name anthony-test-stream

1. Create a test message that contains a non-ASCII character:

        aws logs put-log-events --log-group-name anthony-test-group --log-stream-name anthony-test-stream --log-events "timestamp=1576691153000,message=Test ✓ message"

1. Verify the event was created successfully in the CloudWatch UI
1. Fetch JSON and print output to the console:

        aws logs get-log-events --log-group-name anthony-test-group --log-stream-name anthony-test-stream --output json

1. Fetch JSON and redirect output to file:

        aws logs get-log-events --log-group-name anthony-test-group --log-stream-name anthony-test-stream --output json > out.txt

**Result**

When the output is printed to the console the command succeeds. When the output is redirected to a file AWSCLI fails with the following message and exit code ????:

    'charmap' codec can't encode character '\u2713' in position 6: character maps to <undefined>

It aborts processing at the non-ASCII character and writes invalid JSON to out.txt:

    type out.txt
    {
        "events": [
            {
                "timestamp": 1576691153000,
                "message":

The result is similar for "table" and "text" output formats.

**Clean Up**

    aws logs delete-log-stream --log-group-name anthony-test-group --log-stream-name anthony-test-stream
    aws logs delete-log-group --log-group-name anthony-test-group

aschmied:
The linked branch provides a workaround: it adds a new output format `ascii-json` that escapes the non-ASCII UTF-8 in the output. I'm happy to PR this if it is an acceptable change.

kdaily:
Hi @aschmied, the CLI uses whatever the system encoding is set to be. I tested this on Windows 10 by changing the system locale to use UTF-8 (see this [Stack Overflow answer](https://stackoverflow.com/a/57134096/1601139)) and both cases were successful. Does this solution work for you?

kdaily:
If using AWS CLI v1 you can:

```
set PYTHONIOENCODING=UTF-8
```

[...]

PR #7619, opened 2023-01-29
fix unicode escaped JSON log

furyohei:
*Issue #, if available:*

*Description of changes:*

#### Current state:

```aws logs tail``` CLI command with ```--format json``` for tailing the JSON log events containing Japanese characters results in characters encoded in unicode rather than actual characters.

```
$ aws logs tail test --format json
2023-01-29T17:49:40.180000+00:00 test 
{
    "key": "\u3053\u3093\u306b\u3061\u306f\u4e16\u754c"
}
```

#### Proposed improvements:

Display log messages without escaping.

```
$ aws logs tail test --format json
2023-01-29T17:49:40.180000+00:00 test 
{
    "key": "こんにちは世界"
}
```

### DIFF
--- a/awscli/customizations/logs/tail.py
+++ b/awscli/customizations/logs/tail.py
@@ -95,7 +95,7 @@ class PrettyJSONLogEventsFormatter(BaseLogEventsFormatter):
     def _format_pretty_json(self, log_message):
         try:
             loaded_json = json.loads(log_message)
-            return '\n%s' % json.dumps(loaded_json, indent=4)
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks @furyohei for creating this PR. The CLI uses whatever the system encoding is set to be. You should be able to run `set PYTHONUTF8=1` to display the expected output as mentioned [here](https://github.com/aws/aws-cli/issues/4778#issuecomment-1851064063). Can you confirm this resolves it for you? Please open an issue for further discussion on this is.

PR #7620, opened 2023-01-30
Validate that Python was compiled with sqlite3

stealthycoin:
Add check to configure script and validate_env script to ensure that the Python used was compiled with sqlite3.


kyleknap (approved):
Looks good :ship:

PR #7625, opened 2023-01-31
Fix alpine sdist example

stealthycoin:
Alpine example was not naming the builder image, and needed g++ installed to compile awscrt.


kyleknap (approved):
:ship:

PR #7627, opened 2023-02-01
Fix updating distribution metadata

stealthycoin:
Metadata was set, and then overwritten immediately. This fix writes metadata at the end after the rest of the CLI files have been moved into place.


kyleknap (commented):
So looking over the previous PRs, it seems that we already did this: https://github.com/aws/aws-cli/pull/7357. Was there a bad rebase or something? None of the constants that we settled on are even in the `source-distribution` branch and the merged commit is not in that branch. If it was a bad rebase, it might just make sense to cherry pick the commit from the previous PR and send that up.

kyleknap (approved):
:ship:

PR #7631, opened 2023-02-01
Add a pip install command to the error ouput with unmet dependencies

stealthycoin:
This makes it easier to install all of the dependencies if you do not want to --use-download-deps.

```
$ ./configure
checking for a Python interpreter with version >= 3.8... python
checking for python... /Users/jcarlyl/.envs/env3.9.3/bin/python
checking for python version... 3.9
checking for python platform... darwin
checking for GNU default python prefix... ${prefix}
checking for GNU default python exec_prefix... ${exec_prefix}
checking for python script directory (pythondir)... ${PYTHON_PREFIX}/lib/python3.9/site-packages
checking for python extension module directory (pyexecdir)... ${PYTHON_EXEC_PREFIX}/lib/python3.9/site-packages
checking for --with-install-type... system-sandbox
checking for --with-download-deps... Traceback (most recent call last):
...
validate_env.UnmetDependenciesException: Environment requires following Python dependencies:

flit_core (required: ('>=3.7.1', '<3.7.2')) (version installed: 3.8.0)
colorama (required: ('>=0.2.5', '<0.4.4')) (version installed: None)

Run the following pip command to install manually:
/Users/jcarlyl/.envs/env3.9.3/bin/python -m pip install "flit_core>=3.7.1,<3.7.2" "colorama>=0.2.5,<0.4.4"

configure: error: "Python dependencies not met. Consider using the --with-download-deps flag."


$ /Users/jcarlyl/.envs/env3.9.3/bin/python -m pip install "flit_core>=3.7.1,<3.7.2" "colorama>=0.2.5,<0.4.4"
Collecting flit_core<3.7.2,>=3.7.1
  Using cached flit_core-3.7.1-py3-none-any.whl (60 kB)
Collecting colorama<0.4.4,>=0.2.5
  Using cached colorama-0.4.3-py2.py3-none-any.whl (15 kB)
Installing collected packages: flit_core, colorama
  Attempting uninstall: flit_core
    Found existing installation: flit_core 3.8.0
    Uninstalling flit_core-3.8.0:
      Successfully uninstalled flit_core-3.8.0
Successfully installed colorama-0.4.3 flit_core-3.7.1
(env3.9.3) ➜  aws-cli git:(usable-pip-error) ✗ ./configure
checking for a Python interpreter with version >= 3.8... python
checking for python... /Users/jcarlyl/.envs/env3.9.3/bin/python
checking for python version... 3.9
checking for python platform... darwin
checking for GNU default python prefix... ${prefix}
checking for GNU default python exec_prefix... ${exec_prefix}
checking for python script directory (pythondir)... ${PYTHON_PREFIX}/lib/python3.9/site-packages
checking for python extension module directory (pyexecdir)... ${PYTHON_EXEC_PREFIX}/lib/python3.9/site-packages
checking for --with-install-type... system-sandbox
checking for --with-download-deps... no
configure: creating ./config.status
config.status: creating Makefile
```

### DIFF
--- a/backends/build_system/validate_env.py
+++ b/backends/build_system/validate_env.py
@@ -11,6 +11,8 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import re
+import sys
+import shlex
 from pathlib import Path
 import importlib.metadata
 
@@ -29,12 +31,31 @@ EXTRACT_DEPENDENCIES_RE = re.compile(r'"(.+)"')
 
 
 class UnmetDependenciesException(Exception):
-    def __init__(self, unmet_deps):
+    def __init__(self, unmet_deps, in_venv):
+        pip_install_command_args = ["-m", "pip", "install"]
         msg = "Environment requires following Python dependencies:\n\n"
         for package, actual_version, required in unmet_deps:
             msg += (
                 f"{package} (required: {required.constraints}) "
                 f"(version installed: {actual_version})\n"
+            )
+            pip_install_command_args.append(f'{package}{required.string_constraints()}')
+
+        msg += (
+            "\n"
+            "We recommend using --with-download-deps flag to automatically create a "
+            "virtualenv and download the dependencies.\n\n"
+            "If you want to manage the dependencies yourself instead, run the following "
+            "pip command:\n"
+        )
+        msg += f"{sys.executable} {shlex.join(pip_install_command_args)}\n"
+
+        if not in_venv:
+            msg += (
+                "\nWe noticed you are not in a virtualenv.\nIf not using --with-download-deps "
+                "we highly recommend using a virtualenv to prevent dependencies "
+                "from being installed into your global "
+                "Python environment.\n"
             )
         super().__init__(msg)
 
@@ -43,7 +64,8 @@ def validate_env(target_artifact):
     requirements = _get_requires_list(target_artifact)
     unmet_deps = _get_unmet_dependencies(requirements)
     if unmet_deps:
-        raise UnmetDependenciesException(unmet_deps)
+        in_venv = sys.prefix != sys.base_prefix
+        raise UnmetDependenciesException(unmet_deps, in_venv)
 
 
 def _get_requires_list(target_artifact):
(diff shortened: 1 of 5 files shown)


backends/build_system/validate_env.py
```diff
@@ -30,12 +31,16 @@
...
 class UnmetDependenciesException(Exception):
     def __init__(self, unmet_deps):
+        pip_install_command = f"{sys.executable} -m pip install "
         msg = "Environment requires following Python dependencies:\n\n"
         for package, actual_version, required in unmet_deps:
             msg += (
                 f"{package} (required: {required.constraints}) "
                 f"(version installed: {actual_version})\n"
             )
+            pip_install_command += f'"{package}{required.string_constraints()}" '
```
kyleknap, line 41: Instead of manually quoting these values, it may make sense to just do `shlex.join()` so that all arguments are properly escaped such as the python executable.

backends/build_system/validate_env.py
```diff
@@ -30,12 +31,16 @@
...
+        msg += "\nRun the following pip command to install manually:\n"
+        msg += f"{pip_install_command}\n"
```
kyleknap, line 43: Could we also add a test or expand on an existing test case to make sure this pip command is what we expect?

backends/build_system/validate_env.py
```diff
@@ -30,12 +31,16 @@
...
     def __init__(self, unmet_deps):
+        pip_install_command = f"{sys.executable} -m pip install "
         msg = "Environment requires following Python dependencies:\n\n"
         for package, actual_version, required in unmet_deps:
             msg += (
                 f"{package} (required: {required.constraints}) "
                 f"(version installed: {actual_version})\n"
             )
+            pip_install_command += f'"{package}{required.string_constraints()}" '
+        msg += "\nRun the following pip command to install manually:\n"
```
kyleknap, line 42: Looking at the output from the overview from the PR, I'm a little concerned that the error message is confusing in terms of what to do if someone specifically gets that error. Mainly this error message says to install the dependencies manually but we also have the other error message that says to consider using the `--with-download-deps` flag. So I can see someone being confused on which approach to use. In general, it would be great if we can expand this error message to:

1. Recommend `--with-download-deps` as the top option since that is likely the quickest way to get this working and it won't impact any system installed dependencies
2. Recommend the pip command if `--with-download-deps` does not work or they actually want to use the currently installed Python package
3. Bonus idea - The error detects if the Python interpreter is part of a virtaulenv and gives a warning for the pip commands if the global interpreter is being used and potentially recommends to do all of this setup from within a virtualenv.

Thoughts?

kyleknap (commented):
Looks good. I just had a few comments

backends/build_system/validate_env.py
stealthycoin: Its the same net result.

kyleknap (commented):
Looks like Windows tests are failing but the message looks much better now. Did not have any comments about the latest revision.

PR #7664, opened 2023-02-14
Add error message when flit_core is not installed

stealthycoin:


backends/build_system/utils.py
```diff
@@ -121,7 +121,16 @@ def _parse_req_line(line: str):
 
 
 def get_install_requires():
-    import flit_core.buildapi
+    try:
+        import flit_core.buildapi
+    except ImportError:
+        flit_core_range = get_flit_core_range()
+        raise RuntimeError(
+            f'In order to check dependencies the environment must have '
```
kyleknap, line 129: I wonder if it makes sense to make mention of the `--with-download-deps` flag here as well?

kyleknap (commented):
Looks fine. Just had a small comment.

kyleknap (approved):
Looks good :ship:

PR #7665, opened 2023-02-14
docs: use with docker cli in China region

mefengl:
*Issue #, if available:*
should be `--password-stdin <aws_account_id>.dkr.ecr.<region>.amazonaws.com.cn` instead of `--password-stdin <aws_account_id>.dkr.ecr.<region>.amazonaws.com`
the doc in [amazonaws.cn](https://docs.amazonaws.cn/en_us/AmazonECR/latest/userguide/getting-started-cli.html) is wrong, aws-cli's doc also not mention.

It is reasonable to mention the configuration of a specific area.

*Description of changes:*
I add a comment below the `--password-stdin...`

### DIFF
--- a/awscli/examples/ecr/get-login-password.rst
+++ b/awscli/examples/ecr/get-login-password.rst
@@ -15,5 +15,6 @@ To use with the Docker CLI, pipe the output of the ``get-login-password`` comman
     | docker login \
         --username AWS \
         --password-stdin <aws_account_id>.dkr.ecr.<region>.amazonaws.com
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating this PR. However we do not plan to merge this, because it does not follow the approach used across other CLI examples. Including parameters inline as a comment is not valid, and we don't reference `amazonaws.com.cn` endpoints in examples. There is a link to the ECR User Guide at the bottom of this command's documentation page, which does provide the localized URL: https://docs.amazonaws.cn/AmazonECR/latest/userguide/Registries.html#registry_auth.

PR #7668, opened 2023-02-14
Source distribution

stealthycoin:
Final PR to merge the ongoing feature branch source-distribution into v2.


configure.ac
```diff
@@ -0,0 +1,58 @@
+AC_CONFIG_MACRO_DIRS([m4])
+AC_INIT([awscli], [2.7.31])
```
kyleknap, line 2: We should probably update this to the current version of v2 is now. It will automatically get bumped but it would be good to see it be in sync to start.

kyleknap (commented):
Looks good. Just had one small comment.

kyleknap (approved):
Looks good :ship:

PR #7678, opened 2023-02-17
Use kwargs to simplify adding optional flags to `eks update-kubeconfig`

yuxiang-zhang:
**Issue #, if available:**
No issues, general refactoring to prepare for feature requests like #7623 #7845 #7879

**Description of changes:**
Methods are at risk of getting bloated as more and more parameters are added. Use a variable number of keyword-only arguments allow additional parsed args to be passed without changing the method signature every time. 

[x] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -242,7 +242,8 @@ class EKSClient(object):
         self._cluster_description = None
         self._globals = parsed_globals
 
-    def _get_cluster_description(self):
+    @property
+    def cluster_description(self):
         """
         Use an eks describe-cluster call to get the cluster description
         Cache the response in self._cluster_description.
@@ -276,10 +277,9 @@ class EKSClient(object):
         the previously obtained description.
         """
 
-        cert_data = self._get_cluster_description().get("certificateAuthority",
-                                                        {"data": ""})["data"]
-        endpoint = self._get_cluster_description().get("endpoint")
-        arn = self._get_cluster_description().get("arn")
+        cert_data = self.cluster_description.get("certificateAuthority", {}).get("data", "")
+        endpoint = self.cluster_description.get("endpoint")
+        arn = self.cluster_description.get("arn")
 
         return OrderedDict([
             ("cluster", OrderedDict([
@@ -294,9 +294,8 @@ class EKSClient(object):
         Return a user entry generated using
         the previously obtained description.
         """
-        cluster_description = self._get_cluster_description()
-        region = cluster_description.get("arn").split(":")[3]
-        outpost_config = cluster_description.get("outpostConfig")
+        region = self.cluster_description.get("arn").split(":")[3]
+        outpost_config = self.cluster_description.get("outpostConfig")
 
         if outpost_config is None:
             cluster_identification_parameter = "--cluster-name"
@@ -304,10 +303,10 @@ class EKSClient(object):
         else:
             # If cluster contains outpostConfig, use id for identification
             cluster_identification_parameter = "--cluster-id"
-            cluster_identification_value = cluster_description.get("id")
+            cluster_identification_value = self.cluster_description.get("id")
 
         generated_user = OrderedDict([
-            ("name", user_alias or self._get_cluster_description().get("arn", "")),
+            ("name", user_alias or self.cluster_description.get("arn", "")),
             ("user", OrderedDict([
                 ("exec", OrderedDict([
                     ("apiVersion", API_VERSION),
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -47,6 +47,7 @@ class UpdateKubeconfigCommand(BasicCommand):
     ARG_TABLE = [
         {
             'name': 'name',
+            'dest': 'cluster_name',
             'help_text': ("The name of the cluster for which "
                           "to create a kubeconfig entry. "
                           "This cluster must exist in your account and in the "
@@ -119,9 +120,8 @@ class UpdateKubeconfigCommand(BasicCommand):
 
     def _run_main(self, parsed_args, parsed_globals):
         client = EKSClient(self._session,
-                           parsed_args.name,
-                           parsed_args.role_arn,
-                           parsed_globals)
+                           parsed_args=parsed_args,
+                           parsed_globals=parsed_globals)
         new_cluster_dict = client.get_cluster_entry()
         new_user_dict = client.get_user_entry(user_alias=parsed_args.user_alias)
 
@@ -235,12 +235,12 @@ class KubeconfigSelector(object):
 
 
 class EKSClient(object):
-    def __init__(self, session, cluster_name, role_arn, parsed_globals=None):
+    def __init__(self, session, parsed_args, parsed_globals=None):
         self._session = session
-        self._cluster_name = cluster_name
-        self._role_arn = role_arn
+        self._cluster_name = parsed_args.cluster_name
         self._cluster_description = None
-        self._globals = parsed_globals
+        self._parsed_globals = parsed_globals
+        self._parsed_args = parsed_args
 
     @property
     def cluster_description(self):
@@ -250,14 +250,14 @@ class EKSClient(object):
         describe-cluster will only be called once.
         """
         if self._cluster_description is None:
-            if self._globals is None:
+            if self._parsed_globals is None:
                 client = self._session.create_client("eks")
             else:
                 client = self._session.create_client(
                     "eks",
-                    region_name=self._globals.region,
-                    endpoint_url=self._globals.endpoint_url,
-                    verify=self._globals.verify_ssl
+                    region_name=self._parsed_globals.region,
+                    endpoint_url=self._parsed_globals.endpoint_url,
+                    verify=self._parsed_globals.verify_ssl
                 )
             full_description = client.describe_cluster(name=self._cluster_name)
             self._cluster_description = full_description["cluster"]
@@ -326,10 +326,10 @@ class EKSClient(object):
             ]))
         ])
 
-        if self._role_arn is not None:
+        if self._parsed_args.role_arn is not None:
             generated_user["user"]["exec"]["args"].extend([
                 "--role",
-                self._role_arn
+                self._parsed_args.role_arn
             ])
 
         if self._session.profile:
--- a/tests/unit/customizations/eks/test_update_kubeconfig.py
+++ b/tests/unit/customizations/eks/test_update_kubeconfig.py
@@ -220,7 +220,7 @@ class TestEKSClient(unittest.TestCase):
         self._client = EKSClient(self._session, "ExampleCluster", None)
 
     def test_get_cluster_description(self):
-        self.assertEqual(self._client._get_cluster_description(),
+        self.assertEqual(self._client.cluster_description,
                          describe_cluster_response()["cluster"])
         self._mock_client.describe_cluster.assert_called_once_with(
             name="ExampleCluster"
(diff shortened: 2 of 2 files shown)


awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -113,9 +114,9 @@ def _display_entries(self, entries):
 
     def _run_main(self, parsed_args, parsed_globals):
         client = EKSClient(self._session,
-                           parsed_args.name,
-                           parsed_args.role_arn,
-                           parsed_globals)
+                           cluster_name=vars(parsed_args).pop('cluster_name'),
```
RedbackThomson, line 117: You wouldn't just use `parsed_args.cluster_name` ?

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -229,14 +230,24 @@ def _expand_path(self, path):
...
 class EKSClient(object):
-    def __init__(self, session, cluster_name, role_arn, parsed_globals=None):
+    # A set that contains optional flag keys here. 
+    # If you need to specify a default value other than `None` for the parsed_args,
+    # you need to provide a value for the 'default' key in ARG_TABLE. 
+    _OPTIONAL_ATTRS = {
+        'role_arn',
+    }
+
+    def __init__(self, session, *, cluster_name, parsed_globals=None, **parsed_args):
```
RedbackThomson, line 240: This is honestly a prime place for using `dataclasses` - providing optional parameters and defaults. Here are the docs for Python 3.7 `dataclasses` - https://docs.python.org/3.7/library/dataclasses.html

I don't see them being used anywhere else in the AWS CLI v2 package, though, but they should be supported?

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -270,10 +281,9 @@ def get_cluster_entry(self):
         the previously obtained description.
         """
 
-        cert_data = self._get_cluster_description().get("certificateAuthority",
-                                                        {"data": ""})["data"]
-        endpoint = self._get_cluster_description().get("endpoint")
-        arn = self._get_cluster_description().get("arn")
+        cert_data = self.cluster_description.get("certificateAuthority", {}).get("data", "")
+        endpoint = self.cluster_description.get("endpoint")
+        arn = self.cluster_description.get("arn")
```
RedbackThomson, line 286: No defaults for these?

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -229,14 +230,24 @@ def _expand_path(self, path):
...
         self._cluster_description = None
+        self._cluster_name = cluster_name
         self._globals = parsed_globals
+        
+        for name in EKSClient._OPTIONAL_ATTRS:
+            setattr(self, f'_{name}', parsed_args[name])
 
-    def _get_cluster_description(self):
+    @property
+    def cluster_description(self):
```
RedbackThomson, line 250: I would also add a [`@functools.lru_cache`](https://docs.python.org/3.7/library/functools.html#functools.lru_cache) decoration here, otherwise it'll be an expensive amount of EKS client calls for every operation.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Yes I would. However because I'm passing in `parsed_args` as kwargs later, the init function would receive two (albeit same) values for keyword argument `cluster_name` and results in a TypeError.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Thanks for this pointer! Glad I've learned something new. So I've looked into `dataclasses` and apparently it has some limitations, in order of importance: 
1. is equivalent to setting all fields as named arguments
  - this requires setting *each* field explicitly, e.g. `EKSClient(..., role_arn=parsed_args.role_arn)`
  - still changes the init method signature as data fields are added
    - this leads to updating the code at multiple places, therefore not solving the problems targeted initially
  - it cannot take additional arguments and discard them later like in `EKSClient(..., **vars(parsed_args)` (parsed_args contains extra arguments like `kubeconfig` that will error)
2. requires explicit init method if we want to hide the attribute
  - we could do an explicit `self._session = session` with `InitVar`s
    - this seems to defeat the purpose of using `dataclasses`
  - we could name the data field `_session` and call with `EKSClient(_session=xyz)` but this seems to defeat the purpose of hiding it

In contrast, the limitations of my current approach: 
1. explicitly setting non-public attributes with `setattr()` seems to be bad practice
2. passing extra kwargs that `EKSClient` doesn't need
3. usage of this dict is not obvious

Happy to learn anything I've overlooked.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Fixed in the new changes!

kahirokunn:
awesome! LGTM!

reoring:
I need this😂

awscli/customizations/eks/update_kubeconfig.py
kahirokunn: @RedbackThomson Hi!
What is the cause that is currently stopping this? 👀

RedbackThomson: @kahirokunn My comment was purely a code style suggestion. It certainly should not block the merge of this PR

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -235,14 +236,25 @@ def _expand_path(self, path):
...
         self._cluster_description = None
+        self._cluster_name = cluster_name
         self._globals = parsed_globals
+        
+        for name in EKSClient._OPTIONAL_ATTRS:
+            setattr(self, f'_{name}', parsed_args[name])
 
-    def _get_cluster_description(self):
+    @property
+    @lru_cache()
```
kyleknap, line 256: Do we need to use an `lru_cache` here? It looks like the method already has it's own caching implementation.

awscli/customizations/eks/update_kubeconfig.py
kyleknap: `dataclasses` is definitely an option as v1 now only supports Python 3.7+. I think I would have opted with this approach if we were not concerned with the bloat of initializer arguments and plumbing parsed arguments through.

One question that I have is there a particular reason we just accept the `parsed_args` as is (without expanding them as keyword arguments) similar to how we accept `parsed_globals`? I sort of like this approach for a couple of reasons:
1. It makes even less code we need to the `EKSClient` class every time a new argument is added; we would not even need to add to the `_OPTIONAL_ATTRS` class set.
2. It is very clear what arguments are actually accepted to the initializer of the `EKSClient` class. In reading the class as it is now, readers have to be aware of the `_OPTIONAL_ATTRS` class property in trying to determine what keyword arguments are respected by the client.

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -276,10 +288,9 @@ def get_cluster_entry(self):
         the previously obtained description.
         """
 
-        cert_data = self._get_cluster_description().get("certificateAuthority",
-                                                        {"data": ""})["data"]
-        endpoint = self._get_cluster_description().get("endpoint")
-        arn = self._get_cluster_description().get("arn")
+        cert_data = self.cluster_description.get("certificateAuthority", {}).get("data", "")
+        endpoint = self.cluster_description.get("endpoint", "")
+        arn = self.cluster_description.get("arn", "")
```
kyleknap, line 293: I'd prefer avoiding changing the default types here unless there is a strong reason. Before, it looks like if `endpoint` or `arn` was missing, it would return `None` but now it will be an empty string. Mainly I'm not sure how this will affect downstream logic and would opt to minimizing potential changes in values unless there is a strong reason.

tests/unit/customizations/eks/test_update_kubeconfig.py
```diff
@@ -217,10 +217,16 @@ def setUp(self):
         self._session.create_client.return_value = self._mock_client
         self._session.profile = None
 
-        self._client = EKSClient(self._session, "ExampleCluster", None)
+        self._client = EKSClient(self._session, cluster_name="ExampleCluster", role_arn=None)
+
+    def test_init_client_with_non_supported_parsed_args(self):
+        client = EKSClient(self._session, cluster_name="ClusterName", role_arn=None, non_supported_attr="random value")
+        self.assertIn('_cluster_name', vars(client))
```
kyleknap, line 224: If we move to just accepting the `parsed_args` and not expanded out as keyword arguments, we should be able to remove this test. In general, we should avoid asserting against private members (e.g. members that begin with underscore) as there is no guarantee that interface is stable.

kyleknap (commented):
Looks good! I just had some smaller suggestions and comments.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Thank you, makes a ton of sense. I think I spent too much effort trying to work around the existing usage of `cluster_name` and `role_arn`. The `ARG_TABLE` is and should be kept as the single source of truth for the args.

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -119,9 +120,8 @@ def _display_entries(self, entries):
 
     def _run_main(self, parsed_args, parsed_globals):
         client = EKSClient(self._session,
-                           parsed_args.name,
-                           parsed_args.role_arn,
-                           parsed_globals)
+                           parsed_args=vars(parsed_args),
```
kyleknap, line 123: Let's just accept the [`argparse.Namespace`](https://docs.python.org/3/library/argparse.html#argparse.Namespace) type instead of converting it to a dictionary so that is consistent with the type of `parsed_globals` and other places `parsed_args` is referenced. In the tests, we should be able to just instantiate the `Namespace` class directly. Here is an example of where it is done in a test: https://github.com/aws/aws-cli/blob/4776fb25974c0e1a5f3eeb6ff3f7ff3fe5fc1d97/tests/unit/customizations/cloudtrail/test_validation.py#L897

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -235,28 +235,29 @@ def _expand_path(self, path):
 
 
 class EKSClient(object):
-    def __init__(self, session, cluster_name, role_arn, parsed_globals=None):
+    def __init__(self, session, *, parsed_globals=None, parsed_args):
```
kyleknap, line 238: Let's remove the `*` in the constructor and move the `parsed_args` to a required positional arg. I don't think we should be too worried about backwards compatibility here as this is the only module that relies on the `EKSClient` class. So by the end the initializer signature should be:
```python
def __init__(self, session, parsed_args, parsed_globals=None):
```

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -235,28 +235,29 @@ def _expand_path(self, path):
...
-            if self._globals is None:
+            if self._parsed_globals is None:
                 client = self._session.create_client("eks")
             else:
                 client = self._session.create_client(
                     "eks",
-                    region_name=self._globals.region,
-                    endpoint_url=self._globals.endpoint_url,
-                    verify=self._globals.verify_ssl
+                    region_name=self._parsed_globals["region"],
```
kyleknap, line 258: I believe the `parsed_globals` is still a namespace so we cannot access it as if it has keys; we need to keep it as `self._parsed_globals.region`. This goes for the other places `parsed_globals` are referenced as well.

kyleknap (commented):
Cool. This is looking good. I just had a few more small comments.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Okay that's cool. I wasn't sure whether `parsed_globals` was a Namespace, it seemed like it wasn't because when I put `vars(parsed_globals)` it gave an error.

awscli/customizations/eks/update_kubeconfig.py
yuxiang-zhang: Same observation as above. I was trying to make the use of both variables consistent. But this should be okay if we keep `parsed_args` as `argparse.Namespace`.

kyleknap (approved):
Looks good! :ship: Merging

PR #7682, opened 2023-02-20
Update make-topic-index

aria1991:
I made some changes and refactored the script. The changes included:

— Add docstrings to the `main()` function to describe what it does.

— Add comments to the code to explain what is happening at each step, especially for newbies.

— Add a `shebang` line to specify the Python version and location of the interpreter.

— Use `if __name__ == "__main__`": guard to ensure that the `main()` function is only executed when the script is run directly and not when it is imported as a module.

— Use the `argparse` module to add command-line arguments to the script. This would allow users to specify the source directory, output directory, or any other configuration options.

— Handle exceptions and errors gracefully, and print informative error messages for the user.

— Consider logging errors and status messages to a file, rather than just printing them to the console.

### DIFF
--- a/scripts/make-topic-index
+++ b/scripts/make-topic-index
@@ -1,23 +1,61 @@
-#!/usr/bin/env python
-"""This is a script to generate a JSON index for the CLI Topic Guide
+#!/usr/bin/env python3
+
+"""
+This script generates a JSON index for the CLI Topic Guide.
 
 Whenever a new topic is added to the awscli/topics directory or a tag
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating this PR and for your patience here. I'm not sure that I follow the use case behind these changes. Many of the changes here involving comments don't seem necessary, and I don't follow this approach to logging. If still interested in contributing here, could you please first create an issue outlining what you are trying to address?

PR #7683, opened 2023-02-20
Update benchmark-cp

aria1991:
*Description of changes:*
I refactored, the code made some changes to it including:

— Add docstrings: Add docstrings to the functions and modules to provide descriptions of what the functions do.

— Use f-strings: Use f-strings instead of string concatenation to make the code more readable.

— Use pathlib: Use the pathlib module instead of string manipulation to handle file paths.

— Use context managers: Use context managers to handle file operations to ensure that files are closed properly.

— Add error handling: Add error handling to catch and handle exceptions that may occur during the file operations.

— Use argparse subcommands: Use argparse subcommands to organize the different transfer methods instead of using if/else statements to differentiate between them.

— Rename benchmark_cp: Rename the benchmark_cp function to something more descriptive since it is used to benchmark different transfer methods.

*Issue # N/A

### DIFF
--- a/scripts/performance/benchmark-cp
+++ b/scripts/performance/benchmark-cp
@@ -1,14 +1,27 @@
 #!/usr/bin/env python
-from benchmark_utils import summarize, clean
-from benchmark_utils import get_default_argparser, get_transfer_command
-from benchmark_utils import create_random_subfolder, benchmark_command
 
+import argparse
+import shutil
+from pathlib import Path
+from tempfile import TemporaryDirectory
 
-def benchmark_cp(args):
-    destination = args.destination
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating this PR. That particular [code](https://github.com/aws/aws-cli/tree/develop/scripts/performance) was written ~8 years ago and may have opportunities for refactoring, but I think that would need to be part of a broader refactoring effort which is focused on specific areas across the codebase. Also I'm not sure why those transfer methods were added here. But if you're trying to address something please [create a new issue](https://github.com/aws/aws-cli/issues/new/choose) with more details and then we could discuss potential changes further there.

PR #7684, opened 2023-02-20
Update benchmark-rm

aria1991:
I refactored the script and made some changes to it, including:

— Add docstrings and comments: The script could benefit from some documentation to help explain its purpose, arguments, and functions. Docstrings and inline comments can make the code more readable and maintainable.

— Add exception handling: The script should have proper exception handling to prevent crashes and unexpected behaviors. For example, the script should handle errors if the backup path does not exist, if the command fails to run, or if the cleanup fails.

— Use f-strings for string interpolation: Instead of using %s to format strings, the script could use f-strings, which provide a more concise and readable way to interpolate variables into strings.

— Use argparse for parsing arguments: The script could benefit from using argparse to parse the command-line arguments. This would provide a more consistent and standardized way of parsing and validating arguments.

— Add logging: The script could benefit from adding logging to help debug and trace the execution of the script.

### DIFF
--- a/scripts/performance/benchmark-rm
+++ b/scripts/performance/benchmark-rm
@@ -1,22 +1,65 @@
 #!/usr/bin/env python
-from benchmark_utils import benchmark_command, get_transfer_command
-from benchmark_utils import backup, copy, clean, get_default_argparser
+import argparse
+import logging
+
+from benchmark_utils import (
+    benchmark_command,
+    get_transfer_command,
+    backup,
+    copy,
+    clean,
+)
 
 
 def benchmark_rm(args):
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for creating this PR and your patience in hearing back here. Looking at the `benchmark*` files, they have not been updated for ~8 years: https://github.com/aws/aws-cli/tree/develop/scripts/performance. If still interested in changes here, could you open up an issue outlining what you're trying to do? I think if any changes are made here they should be consistent across the benchmark files, and things like formatting improvements may be better addressed in a separate PR.

PR #7690, opened 2023-02-21
Expand tox range to Python 3.11

hssyoo:
Add Python 3.10 and Python 3.11 to tox range and use Python 3.11 as the base version when running tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38,py39
+envlist = py38,py39,py310,py311
 
 skipsdist = True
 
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
```
kyleknap, line 1: As part of this PR, let's add a `feature` changelog entry to reflect that we'll be bumping the embedded Python interpreter to 3.11

kyleknap (commented):
Looks good. Just had a comment about adding a changelog entry since this change will result in us building the official executables on 3.11 instead of 3.9

kyleknap (approved):

PR #7713, opened 2023-03-01
Add 3.11 sdist test support

hssyoo:
Add Python 3.11 to the sdist test matrix. We also need to bump PyInstaller to 5.7.0 and regenerate the lock files for compatibility with Python 3.11.


pyproject.toml
```diff
@@ -32,6 +32,10 @@ dependencies = [
     "docutils>=0.10,<0.20",
     "cryptography>=3.3.2,<38.0.5",
     "ruamel.yaml>=0.15.0,<=0.17.21",
+    # The clib wheel is generally downloaded as part of ruamel.yaml's
+    # dependency, but it's not being picked up so we explicitly add
```
kyleknap, line 36: Or more specifically `ruamel.yaml` only requires `ruamel.yaml.clib` for less than or equal to Python 3.10. Adding `ruamel.yaml.clib` here ensures we have a consistent dependency closure across all Python versions. I think it would be include that so we are more specific on the rationale.

.changes/next-release/enhancement-dependency-35338.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "dependency",
+  "description": "Update backend dependency on PyInstaller to version 5.7.0"
```
kyleknap, line 4: For this description, let's just make it `Add Python 3.11 support for source distribution`

kyleknap (commented):
Looks good. Just had a couple of comments

kyleknap (approved):
Looks good! :ship:

Issue #7724, opened 2023-03-06
aws eks update-kubeconfig doesn't generate config with env var AWS_SHARED_CREDENTIALS_FILE


zhnu:
### Describe the bug

Hi, when running `aws eks update-kubeconfig --kubeconfig example` the environment variable (if set) `AWS_SHARED_CREDENTIALS_FILE` isn't appended to the generated kubeconfig, while the environment variable  `AWS_PROFILE` is. This causes failures in libraries that use this kubeconfig to authenticate (for example the python kubernetes client library).

### Expected Behavior

Expected to either cli set the environment var on the kubeconfig if exists or documentation warning.

### Current Behavior

Currently the environment variable is read during the command but not written to the kubeconfig file output.

### Reproduction Steps

`
export AWS_SHARED_CREDENTIALS_FILE="~/.aws/test.credentials"
aws --region "something" --profile "something" eks update-kubeconfig --name "example-cluster" --kubeconfig kubeconfig
unset AWS_SHARED_CREDENTIALS_FILE
`
test.py

`from kubernetes import client, config

config.load_kube_config(config_file='./kubeconfig')

v1 = client.CoreV1Api()
print("Listing pods with their IPs:")
ret = v1.list_pod_for_all_namespaces(watch=False)
for i in ret.items:
    print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))`

### CLI version used

aws-cli/2.10.1

### Environment details (OS name and version, etc.)

Linux 22.04.1-Ubuntu

tim-finnigan:
This issue should be fixed via the PR linked above (https://github.com/aws/aws-cli/pull/7726). Those changes will be available in the next CLI release.

tim-finnigan:
The fixed version for v1 will be out today and for v2 tomorrow. You will need to regenerate the config with `update-kubeconfig` again.

zhnu:
Thank you for the fast resolution, have a great day.

Issue #7723, opened 2023-03-06
Parameter output = text blocks kubectl from working.


robinpecha:
### Describe the bug

After the last AWS-CLI upgrade, kubectl stopped working when you have parameter output = text in .aws/config

### Expected Behavior

$ kubectl get ns
aws eks list-clusters 
{
    "clusters": [
        "cluster-sandbox"
    ]
}
$ aws eks update-kubeconfig --name cluster-sandbox
Updated context arn:aws:eks:eu-west-2:123456789:cluster/cluster-sandbox in /home/devops/.kube/config
$ kubectl config current-context
arn:aws:eks:eu-west-2:123456789:cluster/cluster-sandbox
$ kubectl get namespace
NAME              STATUS   AGE
default           Active   33d

### Current Behavior

$ kubectl get ns
aws eks list-clusters 
{
    "clusters": [
        "cluster-sandbox"
    ]
}
$ aws eks update-kubeconfig --name cluster-sandbox
Updated context arn:aws:eks:eu-west-2:123456789:cluster/cluster-sandbox in /home/devops/.kube/config
$ kubectl config current-context
arn:aws:eks:eu-west-2:123456789:cluster/cluster-sandbox
$ kubectl get namespace
Unable to connect to the server: getting credentials: decoding stdout: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }

### Reproduction Steps

Put in your .aws/config value output = text , update your kubeconfig and try to connect there:
$ aws eks update-kubeconfig --name cluster-sandbox
$ kubectl config current-context
$ kubectl get namespace

### Possible Solution

Remove any output = text from .aws/config

### Additional Information/Context

Versions:
Ubuntu 22.04.2 LTS
Linux 5.19.0-35-generic #36~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb 17 15:17:25 UTC 2 x86_64 x86_64 x86_64 
aws-cli/2.11.0 Python/3.11.2 Linux/5.19.0-35-generic exe/x86_64.ubuntu.22 prompt/off
kubectl: Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.2", ... GoVersion:"go1.19.6", ...
Kustomize Version: v4.5.7
Python 3.10.6

### CLI version used

2.11

[...]

Issue #7719, opened 2023-03-03
Malformed response when requesting EKS token from gov-cloud (affects `>=2.10.4`)


timfallmk:
### Describe the bug

After updating to cli v `2.10.4` (and continuing in `2.11.0`) we get the following response from our clusters running in Gov-Cloud
```shell
kubectl get pods
Unable to connect to the server: getting credentials: decoding stdout: couldn't get version/kind; json parse error: json: cannot unmarshal array into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
```
We get the same response from every `kubectl` request.

Narrowed the problem down to the authentication step callout to `aws eks get-token`. With debug logs on, error occurs after the following lines:
```shell
host:sts.us-gov-west-1.amazonaws.com
x-k8s-aws-id:farpoint

host;x-k8s-aws-id
<earlier call stack>
<snip>
2023-03-03 13:43:15,726 - MainThread - botocore.auth - DEBUG - StringToSign:
AWS4-HMAC-SHA256
20230303T214315Z
20230303/us-gov-west-1/sts/aws4_request
<snip>
2023-03-03 13:43:15,726 - MainThread - botocore.auth - DEBUG - Signature:
<snip>
```

A few key details:
- Region: `us-gov-west-1`
- kubectl version:
```shell
Client Version: v1.25.6
Kustomize Version: v4.5.7
Server Version: v1.25.6-eks-48e63af
```
- Clusters are EKS running 1.25 and provisioned with `eksctl`
- We use `aws sso` for account auth and token generation

Confirmed a few things:
- Does not happen when using fixed IAM credentials, either via `aws` or `aws-iam-authenticator`
- Does not happen for cli <=`2.10.3`

I can't seem to get any more debug logs so I can't dig further, but feels like a malformed response somewhere? Naked `aws eks get-token` seems to return expected results.

### Expected Behavior

`kubectl` command returns data

### Current Behavior

```shell
kubectl get pods
Unable to connect to the server: getting credentials: decoding stdout: couldn't get version/kind; json parse error: json: cannot unmarshal array into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
```
Happens for all subcommands checked (that interact with the server)

### Reproduction Steps

Needs:
- EKS cluster 1.25 in Gov-Cloud region
- AWS SSO setup and used as login method

Doesn't work:
[...]

PR #7726, opened 2023-03-06
Use JSON only when getting user entries in kubeconfig

yuxiang-zhang:
*Issue #, if available:*
#7719, #7723, #7724

*Description of changes:*
Hard-code `json` as the output format when getting user entry in kubeconfig. 

### DIFF
--- a/.changes/next-release/bugfix-eks-56963.json
+++ b/.changes/next-release/bugfix-eks-56963.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "eks",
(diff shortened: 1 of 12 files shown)


tim-finnigan:
@yuxiang-zhang or @stealthycoin can we get your thoughts on this comment: https://github.com/aws/aws-cli/issues/7719#issuecomment-1458824817 ?

> I'm not sure this is a complete fix then. New (greenfield) users would be covered, but anyone would else would still have this issue. Since the error message doesn't give any hints on cause they would have to search around, find this issue, and then apply fixes themselves. Unless I'm misunderstanding something.
> 
> I would classify this as still open.
> 
> Suggestion: Ignore user-set formatting variables when not in an interactive tty?

Issue #7720, opened 2023-03-03
system-wide awscli config (/etc/aws/config)


pmoody-:
### Describe the feature

I would like to see a system-wide aws config file.

### Use Case

I'm thinking in terms of being able to distribute a single system-wide configuration to a set of employees, while allowing users to maintain their local ~/.aws/config with whatever specific profiles they want to maintain.

### Proposed Solution

something similar to how ssh(1) supports /etc/ssh/config and ~/.ssh/config, in my mind the configuration precedence would be 

  command line option
  environment variable
  ~/.aws/config
  /etc/aws/config

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

aws-cli/2.11.0 Python/3.11.2 Darwin/22.3.0 source/x86_64 prompt/off

### Environment details (OS name and version, etc.)

osx, 13.2.1

tim-finnigan:
Hi @pmoody- thanks for the feature request. I'll share the current credential precedence here for reference: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence.

Can you elaborate on your use case a bit more? Feature requests like this would need to be considered at a cross-SDK level, similar to https://github.com/aws/aws-sdk/issues/215. It seems like there may be some overlap between this issue and https://github.com/aws/aws-sdk/issues/215 in terms of the problem you're trying to solve. You might also consider trying [Identity Center](https://aws.amazon.com/iam/identity-center/) as an alternative for managing different accounts.

pmoody-:
Hey @tim-finnigan, thanks for looking at this.

at $job, we use identity center with our aws accounts. we don't have hundreds, but we have enough that trying to manage the roles by hand is impossible. we manage the accounts and the roles assigned to each account centrally and I would like to be able to push a fully generated config file to each managed machine (employee laptop). since we're using identity center, having the configuration doesn't get you access to a given account/role, it's being a member of the right group. so centrally managing this file makes my life, and the lives of everyone at work, a little easier.

for every account we might have three basic roles that look like

```
[profile prod-readonly]
sso_start_url = https://foo.awsapps.com/start
sso_region = us-west-2
sso_account_id = 111122223333
sso_role_name = readonly

[profile prod-eng]
sso_start_url = https://foo.awsapps.com/start
sso_region = us-west-2
sso_account_id = 111122223333
sso_role_name = eng

[profile prod-admin]
[...]

PR #7728, opened 2023-03-06
Add support for a system-wide aws config file.

pmoody-:
Issue #7720

*Description of changes:*

It's beneficial in an enterprise to have centrally managed settings for certain applications live in a system-wide location. All of the plumbing for this already exists, so just add /etc/aws/config as a default, lower priority config file location.

### DIFF
--- a/awscli/botocore/session.py
+++ b/awscli/botocore/session.py
@@ -367,7 +367,7 @@ class Session(object):
         if self._config is None:
             try:
                 config_file = self.get_config_variable('config_file')
-                self._config = botocore.configloader.load_config(config_file)
+                self._config = botocore.configloader.multi_file_load_config(config_file, "/etc/aws/config")
             except ConfigNotFound:
                 self._config = {'profiles': {}}
             try:


dlm6693:
Hi @pmoody-. Thank you for the pull request. This feature is under consideration by the AWS SDKs organization and if implemented, will be done across all SDKs and both major versions of the CLI. It cannot be implemented for just V2 of the CLI.

Closing this out.

pmoody-:
hi @dlm6693 is there an issue or something where I can follow along?

dlm6693:
@pmoody- yep
https://github.com/aws/aws-sdk/issues/215

pmoody-:
thanks, @dlm6693 

I mentioned in the issue that this is attached to that I think the use-cases are slightly different.

> so I see this as similar but different to https://github.com/aws/aws-sdk/issues/215. where as he has a ridiculously large set of personal settings that he'd like to be able to organize outside of a monolithic ~/.aws/config, I have a group of centrally managed settings I would like to automatically configure for my users.

should I open a separate issue with aws-sdk or am I misreading that issue and the system-wide configuration is actually covered by that too?

dlm6693:
Ah my apologies for missing that!

Yes I'd say that your feature request is distinct enough from the one referenced to warrant a separate feature request. It is possible that they could be combined into one larger update, but is far from a guarantee. Regardless this is something that would have to be supported across all SDKs and the CLI so I would strongly recommend opening an issue in https://github.com/aws/aws-sdk/issues

pmoody-:
https://github.com/aws/aws-sdk/issues/490

(if anyone sees this PR and wants to know where the issue migrated)

Issue #7721, opened 2023-03-04
`aws configure import --csv` Expected header 'User Name' not found yet exists in the CSV


gitaugakwa:
### Describe the bug

Importing a configuration using the cli doesn't work. Producing an error that the expected header is not found yet is in the csv.

### Expected Behavior

Import the configuration and add the profile to the config.

### Current Behavior

PS C:\Users\DOUG> aws configure import --profile-prefix TheUnderground --csv file://"C:\Clients\TheUnderground\Credentials\Setup\Setup.csv" --debug
2023-03-04 11:56:27,304 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/2.11.0 Python/3.11.2 Windows/10 exe/AMD64
2023-03-04 11:56:27,306 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['configure', 'import', '--profile-prefix', 'TheUnderground', '--csv', 'file://C:\\Clients\\TheUnderground\\Credentials\\Setup\\Setup.csv', '--debug']
2023-03-04 11:56:27,352 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function add_s3 at 0x0000021AFB003920>
2023-03-04 11:56:27,352 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function add_ddb at 0x0000021AFAE50EA0>
2023-03-04 11:56:27,353 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <bound method BasicCommand.add_command of <class 'awscli.customizations.configure.configure.ConfigureCommand'>>
2023-03-04 11:56:27,353 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function change_name at 0x0000021AFADAD1C0>
2023-03-04 11:56:27,354 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function change_name at 0x0000021AFADD0680>
2023-03-04 11:56:27,354 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function alias_opsworks_cm at 0x0000021AFB011E40>
2023-03-04 11:56:27,354 - MainThread - botocore.hooks - DEBUG - Event building-command-table.main: calling handler <function add_history_commands at 0x0000021AFAE87880>
[...]

PR #7744, opened 2023-03-10
Remove BOM from import csv file

tim-finnigan:
*Issue #, if available:*
Fixes https://github.com/aws/aws-cli/issues/7721

*Description of changes:*
Several users reported receiving the following error:
 `awscli.customizations.configure.importer.CredentialParserError: Expected header "User Name" not found`

This occurs when attempting the command `aws configure import --csv file://<filename>.csv` (even when User Name exists in the .csv file.)

_(Note: users do need to manually add User Name to the .csv file provided in the IAM console as noted here in the User Guide: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-creds-import)_

After some community feedback and testing, we found that a Byte Order Mark ([BOM](https://unicode.org/faq/utf_bom.html#bom1)) was getting added to the beginning of the .csv file when users edited/saved the file in certain applications (for example, Excel but not TextEdit). The BOM breaks the credential parser, resulting in the previously referenced error. The change in this PR removes the BOM from the CSV file if it exists, prior to processing the file.

---

### DIFF
--- a/awscli/customizations/configure/importer.py
+++ b/awscli/customizations/configure/importer.py
@@ -160,9 +160,6 @@ class CSVCredentialParser(object):
             raise CredentialParserError(self._EMPTY_CSV)
 
         lines = csv.splitlines()
-        # Remove the BOM if it exists
-        if lines[0].startswith(u'\ufeff'):
-            lines[0] = lines[0][1:]
         parsed_headers = self._parse_csv_headers(lines[0])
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Closing in favor of https://github.com/aws/aws-cli/pull/7752 to support the encoding.

PR #7745, opened 2023-03-12
Replace deprecated BadZipfile with BadZipFile

hugovk:
*Issue #, if available:* n/a

*Description of changes:*

`BadZipfile` (with a small `f`) has been deprecated since Python 3.2, use `BadZipFile` (big `F`) instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/issues/86437

### DIFF
--- a/awscli/customizations/awslambda.py
+++ b/awscli/customizations/awslambda.py
@@ -94,7 +94,7 @@ def _should_contain_zip_content(value):
     try:
         with closing(zipfile.ZipFile(fileobj)) as f:
             f.infolist()
-    except zipfile.BadZipfile:
+    except zipfile.BadZipFile:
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):
Thanks for this, @hugovk!

hugovk:
We decided upstream it's not worth removing the aliases, so we could keep using the deprecated name here.

Would you prefer to merge or close this?

nateprewitt:
It looks like this got dropped waiting for CI, apologies 😅 This still seems like reasonable cleanup regardless of whether it's going to be removed. I'll go ahead and merge it now, thanks again @hugovk!

Issue #7753, opened 2023-03-15
User Name not in credentials CSV file by default


tim-finnigan:
### Describe the feature

When downloading CLI credentials from the IAM console, User Name is not an included header. However, User Name is expected when running `aws configure import –csv file://credentials.csv` as it is used for the profile name.

### Use Case

It is a frustrating manual step to have to open the file and add a User Name.

### Proposed Solution

Possible solution is to prompt the user if "User Name" is not found in CSV file to specify what they want their profile to be called.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

latest

### Environment details (OS name and version, etc.)

macOS Ventura

PR #7755, opened 2023-03-16
configure import csv discover username when missing

stealthycoin:
Candidate solution to #7753 

Posting this as a draft PR to solicit feedback. The idea is to use the credentials to make an `sts get-caller-identity` call to discover the username if possible. The downside is of course it adds a conditional API call where previously there was none.


tim-finnigan:
Closing as this PR has conflicts, can continue to track issue/discussion in https://github.com/aws/aws-cli/issues/7753

PR #7762, opened 2023-03-21
Force timezone info to fix test failure

Avimitin:
This commit will affects the test
`test_credentials.py::SSOSessionTest::test_token_chosen_from_provider`. This test will throw a `RuntimeError: Credentials were refreshed, but the refreshed credentials are still expired` because the [timestamp call here](https://github.com/aws/aws-cli/blob/v2/tests/functional/botocore/test_credentials.py#L1055) uses system local time, which is different from UTC time. And this will cause an unexpected failure if the test is not run in a system configured with `TZ=UTC`.

To reproduce the bug, try to input the below code in python REPL:

```python
from datetime import datetime, timezone

datetime.now(tz=timezone.utc).timestamp()

# datetime.utcnow() returned datetime object doesn't have tzinfo
datetime.utcnow().timestamp()
```

*Description of changes:*

This PR force the test to use UTC tz as default timezone.

### DIFF
--- a/tests/functional/botocore/test_credentials.py
+++ b/tests/functional/botocore/test_credentials.py
@@ -19,7 +19,7 @@ import time
 import mock
 import tempfile
 import shutil
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 import sys
 
 import pytest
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for reaching out. Can you create an issue for this outlining what you are trying to address in more detail? I think we need more discussion here regarding the proposed changes. `datetime.utcnow()` is used throughout the test files so I'm not sure about deviating from that here. Also will link the [error message](https://github.com/boto/botocore/blob/55175cafc3fa06329293e1741f50584cf23d3183/botocore/credentials.py#L566) you referenced for context.

PR #7773, opened 2023-03-23
Remove __pycache__ folders from source tree when building

stealthycoin:
This prevents __pycache__ folders from being generated in the source tree. They will still be generated in the build/ directory, however for out-of-tree builds this will prevent files from littering the source directory.


kyleknap (approved):
Looks good. :ship:

PR #7774, opened 2023-03-23
Add the rest of cli v2 dependencies to dependabot config

stealthycoin:
This adds the following dependencies to dependabot
* ruamel.yaml
* ruamel.yaml.clib
* prompt-toolkit
* distro
* python-dateutil
* jmespath
* urllib3


.github/dependabot.yml
```diff
@@ -14,3 +14,10 @@ updates:
       - dependency-name: "docutils"
       - dependency-name: "cryptography"
       - dependency-name: "awscrt"
+      - dependency-name: "ruamel.yaml"
+      - dependency-name: "ruamel.yaml.clib"
+      - dependency-name: "prompt-toolkit"
+      - dependency-name: "distro"
+      - dependency-name: "python-dateutil"
+      - dependency-name: "jmespath"
+      - dependency-name: "urllib3"
```
kyleknap, line 23: I wonder if we should just remove the `allow` key instead of maintaining this list? I think my concern is that there are probably dependencies that we don't necessarily need consistent updates for (e.g. `tox`, Sphinx). Not sure if there is a way to ignore a specific requirements file.

kdaily: You can control which types of updates would apply based on the manifest they would apply to (e.g., restrict to updates to lock files only). There doesn't seem to be a way to restrict a specific requirements file, though, unless they were in different directories.

kdaily: Specifying `package-ecosystem: "pip"` means it looks for manifests for `pip`, `pipenv`, `pip-compile`, and `poetry` ([reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)).

kyleknap: Gotcha. Makes sense. Let's keep it as is then.

kyleknap (approved):
:ship:

PR #7787, opened 2023-03-28
Use YAML 1.1 spec for CFN

### LINKED ISSUE #3991: `aws cloudformation package` modifies certain string values to numbers
#### `template.yml`
```yaml
Parameters:
  Value7:
    Type: String
    Default: "077777777777"
  Value8:
    Type: String
    Default: "088888888888"
  Value9:
    Type: String
    Default: "099999999999"
Resources:
  NestedStack:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: ./template_child.yml
```

#### `template_child.yml`
```yaml
# This template is not so important
Resources:
  Dummy:
    Type: Custom::Whatever
    Properties:
      ServiceToken: whatever
      Value1: whatever
```

### Commands to reproduce
```console
$ uname -a
uname -a
Darwin xxxxxxx 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64

$ aws --version
aws-cli/1.16.110 Python/3.7.2 Darwin/18.2.0 botocore/1.12.100

$ aws cloudformation package --s3-bucket BUCKET_NAME --template-file template.yml                                                           
Uploading to 137ce8b72427772da39a43ddc087908a.template  101 / 101.0  (100.00%)
Parameters:
  Value7:
    Type: String
    Default: '077777777777'
  Value8:
    Type: String
    Default: 088888888888
  Value9:
    Type: String
    Default: 099999999999
Resources:
  NestedStack:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://s3.amazonaws.com/BUCKET_NAME/137ce8b72427772da39a43ddc087908a.template
```

Quotes of `Value8` and `Value9` are removed after `aws cloudformation package`.
This make value into something like `9.9999999999E10` in the stacks.

hssyoo:
Fixes #3991 

This PR changes the YAML spec for CloudFormation templates from 1.2 (`ruamel`'s default) to 1.1. This is to stay consistent with what CloudFormation supports: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-formats.html

It also adds logic to explicitly quote nonoctal values such as `0888888` so they won't be treated as integers when deployed to the stack.

### DIFF
--- a/awscli/customizations/cloudformation/yamlhelper.py
+++ b/awscli/customizations/cloudformation/yamlhelper.py
@@ -63,21 +63,17 @@ def _dict_representer(dumper, data):
     return dumper.represent_dict(data.items())
 
 
-def _add_yaml_1_1_boolean_resolvers(resolver_cls):
-    # CloudFormation treats unquoted values that are YAML 1.1 native
-    # booleans as booleans, rather than strings. In YAML 1.2, the only
-    # boolean values are "true" and "false" so values such as "yes" and "no"
-    # when loaded as strings are not quoted when dumped. This logic ensures
-    # that we dump these values with quotes so that CloudFormation treats
-    # these values as strings and not booleans.
-    boolean_regex = re.compile(
-        '^(?:yes|Yes|YES|no|No|NO'
-        '|true|True|TRUE|false|False|FALSE'
-        '|on|On|ON|off|Off|OFF)$'
-    )
-    boolean_first_chars = list(u'yYnNtTfFoO')
-    resolver_cls.add_implicit_resolver(
-        'tag:yaml.org,2002:bool', boolean_regex, boolean_first_chars)
+def _str_representer(dumper, data):
+    # ruamel removes quotes from values that are unambiguously strings.
+    # Values like 0888888 can only be a string because integers can't
+    # have leading 0s and octals can't have the digits 8 and 9.
+    # However, CloudFormation treats these nonoctal values as integers
+    # and removes the leading 0s. This logic ensures that nonoctal
+    # values are quoted when dumped.
+    style = None
+    if re.match('^0[0-9]*[89][0-9]*$', data):
+        style = "'"
+    return dumper.represent_scalar('tag:yaml.org,2002:str', data, style=style)
 
 
 def yaml_dump(dict_to_dump):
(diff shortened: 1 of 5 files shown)


tests/unit/customizations/cloudformation/test_yamlhelper.py
```diff
@@ -193,4 +194,14 @@ def test_yaml_dump_quotes_boolean_strings(self):
         ]
         for bool_as_string in bools_as_strings:
             self.assertEqual(
-                yaml_dump(bool_as_string), "'%s'\n" % bool_as_string)
+                re.sub('%YAML 1.1\n--- ', '', yaml_dump(bool_as_string)),
+                "'%s'\n" % bool_as_string)
+
+    def test_yaml_dump_quotes_non_octal_integers(self):
```
stealthycoin, line 200: For completeness it would be good to have a non-octal test as well.

awscli/customizations/cloudformation/yamlhelper.py
```diff
@@ -88,10 +101,12 @@ def yaml_dump(dict_to_dump):
     """
 
     yaml = ruamel.yaml.YAML(typ="safe", pure=True)
+    yaml.version = (1, 1)
     yaml.default_flow_style = False
     yaml.Representer = FlattenAliasRepresenter
     _add_yaml_1_1_boolean_resolvers(yaml.Resolver)
```
stealthycoin, line 107: If we are moving back to 1.1 is this still required?

tests/unit/customizations/cloudformation/test_yamlhelper.py
```diff
@@ -182,7 +183,7 @@ def test_unroll_yaml_anchors(self):
             '      Foo: bar\n'
             '      Spam: eggs\n'
         )
-        actual = yaml_dump(template)
+        actual = re.sub('%YAML 1.1\n---\n', '', yaml_dump(template))
```
stealthycoin, line 186: Since our output now intentionally has `%YAML 1.1` at the beginning we should probably add an explicit test to ensure it is there. And if our other tests assertions are not expecting it to be there we should have a utility function that trims it off rather than writing a regex in each test that doesn't want it.

For example you could write a `yaml_dump_without_header()` function that calls `yaml_dump` and trims the header off. Then use that instead of `yaml_dump` in the tests that do not expect the header to be there.

stealthycoin (commented):
Looks good, just a couple minor comments.

awscli/customizations/cloudformation/yamlhelper.py
hssyoo: Nope, I tried removing it and it passed the tests, including `test_yaml_dump_quotes_boolean_strings`. Also tried `package` on a YAML file with various booleans and everything looked compliant with the 1.1 spec.

stealthycoin (approved):
The changes look good to me.

PR #7802, opened 2023-04-04
Add --include-file and --exclude-file options

### LINKED ISSUE #5160: s3 sync: Read include/exclude patterns from file
# Problem
When using `s3 sync`, we can include/exclude objects for the sync using `--exclude <pattern>` and `--include <pattern>`. This can be cumbersome if someone has to deal with many patterns at once.

# Proposed solution
I'd suggest to implement two additional options: `--exclude-file <filename>` and `--include-file <filename>` so we would be able to specify files which contain lists of patterns. See how `rsync` does it [here](https://download.samba.org/pub/rsync/rsync.html). Implementing it in the same way as `rsync` does would have the benefit of maintaining one list for both.

### Now
```bash
aws s3 sync . s3://bucket/html/ --exclude=".*" --exclude "license/*" --exclude "LICENSE.md" --exclude "*.map" --exclude "*.scss" --exclude "customVariables.css" --exclude "css/bootstrap/*"
```

### With proposed feature
```bash
aws s3 sync . s3://bucket/html/ --exclude-file .syncignore
```

# Alternatives I have considered
None really. The only alternative I know of is using many single statements.

tmccombs:
Fixes #5160 

Add two new options to several s3 commands to include or exclude all patterns from a file.


tmccombs:
Still needs tests. I haven't figured out how the tests are organized yet.

amberkushwaha931:
merge the data field and vendor info. and do the selected file ratio and contability graph.

tmccombs:
@amberkushwaha931 I don't understand what you are trying to say. what data field and vendor info? what is the selected file ration and contability graph?

tmccombs:
Any update on this?

RyanFitzSimmonsAK:
Hey, thanks for this PR and for your patience.  I brought this up to the team, and we've decided to close this as not planned.  We'll be leaving the feature request (https://github.com/aws/aws-cli/issues/5160) open for tracking purposes, but we will not be merging this.

tmccombs:
What is the reason for not adding this feature?

Having to invoke the aws cli for each file individually is significantly slower, since you have to pay the cost of starting the python interpretor, loading modules, etc.  Passing all the files on the command line is limited, because there is a maximum size for the command line.

The only viable option if you need to synchronize a large number of files is to write your own script using the SDK. But that involves having to duplicate a lot of what the CLI already does, especially if you want things like parallelization, conditional uploads/downloads, progress reporting, etc.

PR #7823, opened 2023-04-15
Better sso url

johnf:
*Description of changes:*

I have a workflow where I SSH to a remote box and use `aws sso login`.  Currently, it outputs the URL and the code separately,  which means I have to manually copy and paste the code.

I've modified the output to be similar to the ```--no-browser``` output so you can click on the auto code entering URL

### DIFF
--- a/awscli/customizations/sso/utils.py
+++ b/awscli/customizations/sso/utils.py
@@ -137,6 +137,9 @@ class OpenBrowserHandler(BaseAuthorizationhandler):
             f'\n{verificationUri}\n'
             f'\nThen enter the code:\n'
             f'\n{userCode}\n'
+            f'\nAlternatively, you may visit the following URL which will '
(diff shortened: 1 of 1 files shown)


johnf:
Note: I would prefer that the output only has one URL, and mentions the code as a fallback. I don't think there is value in showing both. But wasn't sure if others would agree

johnf:
@nateprewitt @hssyoo @kyleknap Hi! Sorry to hassle you, but this has been open for months without anyone looking at it.

Any chance of getting this merged? It's a pretty simple change.

Let me know if it needs any changes

tim-finnigan:
Thanks for creating this PR. I brought this up for discussion with the team, and the consensus was that we don't plan to merge this because this appears to be a niche use case which you can already accomplish using `--no-browser`.

PR #7831, opened 2023-04-18
Add generate changelog workflow_dispatch github action

stealthycoin:
This action can be invoked manually on any branch by providing the repo/ref from which to pull and push. The changelog file is generated using the new-change script using the provided inputs type/category/description.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,68 @@
+name: Generate changelog
+
+on:
+  workflow_dispatch:
+    inputs:
+
+      ref:
+        description: |
+          Branch or ref to clone. This is where the changelog will be added and
+          ultimately committed.
+        required: true
+
+      type:
+        type: choice
+        options:
+          - enhancement
+          - bugfix
+          - api-change
+        description: |
+          Type should be one of: feature, bugfix, enhancement, api-change
+          feature: A larger feature or change in behavior, usually resulting in a
+                   minor version bump.
+          bugfix: Fixing a bug in an existing code path.
+          enhancement: Small change to an underlying implementation detail.
+          api-change: Changes to a modeled API.
+        required: true
+
+      category:
+        description: |
+          Category is the high level feature area.
+          This can be a service identifier (e.g ``s3``),
+          or something like: Paginator.
+        required: true
+
+      description:
+        description: |
+          A brief description of the change.  You can
+          use github style references to issues such as
+          "fixes #489", "boto/boto3#100", etc.  These
+          will get automatically replaced with the correct
+          link.
+        required: true
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,69 @@
...
+        required: true
+
+jobs:
+  add-changelog:
+    runs-on: Ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        repo: ${{ github.event.inputs.repo }}
```
kyleknap, line 51: This key needs to be `repository` per the usage: https://github.com/actions/checkout#usage. I was seeing warnings as well about this in the sample workflow I ran.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,69 @@
+name: Generate changelog
+
+on:
+  workflow_dispatch:
+    inputs:
+
+      repository:
+        description: "Repository in which the changelog will be added."
+        required: false
```
kyleknap, line 9: We should make this required as the default text box can be emptied. We should always require a repository so there is no ambiguity on where this will be applied.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      type:
+        description: |
+          Type should be one of: feature, bugfix, enhancement, api-change
+          feature: A larger feature or change in behavior, usually resulting in a
+                   minor version bump.
+          bugfix: Fixing a bug in an existing code path.
+          enhancement: Small change to an underlying implementation detail.
+          api-change: Changes to a modeled API.
+        required: true
```
kyleknap, line 26: It would be great if we could do the options here as these are a known set of values: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,69 @@
...
+        repo: ${{ github.event.inputs.repo }}
+        ref: ${{ github.event.inputs.ref }}
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.11"
+    - name: Create changelog
+      run: |
+        python scripts/new-change -t ${{ github.event.inputs.type }} -c ${{ github.event.inputs.category }} -d "${{ github.event.inputs.description }}"
```
kyleknap, line 60: We should add quotes around the category. This can have spaces in the value.

kyleknap (commented):
Looks good. Just had some comments.

kyleknap (approved):
:ship:

kyleknap (approved):
:ship:

PR #7832, opened 2023-04-18
Add --show-files option to regenerate-lock-files scripts

stealthycoin:
When provided the --show-files option will print out the full contents of all the lock files after they have been regenerated.


kyleknap (approved):
:ship:

PR #7833, opened 2023-04-18
Add a update-lockfiles github workflow

stealthycoin:
This workflow can be manually triggered using workflow_dispatch. It will run both a windows and macOS job to regenerate the lockfiles to match the current state of the requirements files. dry-run mode can be enabled to inspect the output that would be commited before hand.

### DIFF
--- a/.github/workflows/update-lockfiles.yml
+++ b/.github/workflows/update-lockfiles.yml
@@ -1,0 +1,85 @@
+name: Update lockfiles
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: |
+          Branch or ref to clone. This is where the lockfiles will be added and
+          ultimately committed.
+        required: true
+
+      dry-run:
+        type: choice
+        options:
+          - "yes"
+          - "no"
+        default: "yes"
+        description: |
+          Dry run mode when enabled with the "yes" option will not commit or push
+          the generated files.
+
+
+jobs:
+
+  update-lockfiles:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"]
+        os: [macOS-latest, windows-latest]
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-lockfiles.yml
```diff
@@ -0,0 +1,75 @@
...
+      # If changes were created when running the regenerate-lock-files script
+      # then the new lock files need to be commited and pushed to the branch.
+      - name: git commit and push updated lock files
+        if: github.event.inputs.dry-run == 'no' && env.CHANGES
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "<>"
+          git fetch
+          git add requirements
+          git commit -m "Regenerate lock files for ${{ runner.os }}"
```
kyleknap, line 73: It's a little weird that for the MacOS/Linux lock files this commit will be: `Regenerate lock file for MacOS`. I'm wondering if we can either include Linux in there or just leave it at `Regenerate lock files` for that commit

.github/workflows/update-lockfiles.yml
```diff
@@ -0,0 +1,75 @@
...
+        if: runner.os == 'Windows'
+        run: |
+          python scripts/regenerate-lock-files --show-files
+          echo "CHANGES=$(git status --porcelain=v1)" >> $env:GITHUB_ENV
+          echo "CHANGES=$(git status --porcelain=v1)" >> $env:GITHUB_OUTPUT
+      - name: Regenerate lock files
+        if: runner.os != 'Windows'
+        run: |
+          python scripts/regenerate-lock-files --show-files
+          echo "CHANGES<<EOF" >> $GITHUB_ENV
```
kyleknap, line 58: It might be helpful to have a comment on why we need the EOF lines here. It's hard to follow otherwise.

kyleknap (commented):
Looks good. Just had a couple of comments.

kyleknap (approved):
:ship:

PR #7847, opened 2023-04-24
Allow non-interactive configuration of sso-session

### LINKED ISSUE #7835: Allow configuring sso-session settings non-interactively
### Describe the feature

As mentioned in [this comment](https://github.com/aws/aws-cli/pull/7364#issuecomment-1329854689), you cannot use `aws configure set` to set `sso-session` parameters in the AWS config file.

The CLI should provide a mechanism to non-interactively set these values

### Use Case

I'm trying to set up command line tools for multiple users in an organization, making it easy to use AWS SSO for their tooling.
I would like to be able to give them a script to run to add SSO config to their config files vs manual instructions.

### Proposed Solution

Update `aws configure` to work with `sso-session` prefixes like this:
```bash
aws configure set sso-session.session-name.sso_region eu-west-1 
```

(As shown by @ezzatron in https://github.com/aws/aws-cli/pull/7364#issuecomment-1329854689).

Alternatively, update `aws configure sso-session` to allow non-interactive operation by allowing flags instead of passing in values interactively (this might make sense as a separate issue)

### Other Information

_No response_

### Acknowledgements

### CLI version used

2.9.6

Mac OS Ventura

jferris:
Resolves #7835.

The client supports configuring reusable SSO sessions using the command:

    aws configure sso-session

However, there is no way to set or get the sso-session configuration
programmatically.

This updates the `configure set` and `configure get` commands to support
specifying an sso-session by using the prefix `sso-session.NAME`.

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/awscli/customizations/configure/get.py
+++ b/awscli/customizations/configure/get.py
@@ -86,6 +86,13 @@ class ConfigureGetCommand(BasicCommand):
                     section, {}).get(config_name)
             return value
 
+        if parts[0] == 'sso-session':
+            session_name = parts[1]
+            config_name = parts[2]
+            value = self._session.full_config['sso_sessions'].get(
+                session_name, {}).get(config_name)
+            return value
+
         if parts[0] == 'profile':
             profile_name = parts[1]
             config_name = parts[2]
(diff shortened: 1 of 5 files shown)


eytanhanig:
Any updates on this?

PR #7931, opened 2023-05-30
Add Y/N boolean YAML test

hssyoo:
Ensure that unquoted `Y` and `N` values in YAML files don't get converted into booleans when dumping.


kyleknap (approved):
:ship:

PR #7971, opened 2023-06-13
[v2] Add support for ssh and open-tunnel commands

kyleknap:
The ssh command allows you to connect to an EC2 instance using an OpenSSH client.

The open-tunnel command allows you to connect to an EC2 instance using a web socket tunnel. Using the web socket tunnel, you can then send SSH traffic through an EC2 Instance Connect Endpoint in the case your instance does not have a public IP.

This change also required bumping the lower bound dependency range on CRT. The open-tunnel command uses websockets under the hood and websockets were added as part of CRT 0.16.1. Version 0.16.4 is the lowest version that passed all of the ec2-instance-connect tests and had no more bug fixes related to the web socket development.

### DIFF
--- a/.changes/next-release/feature-ec2instanceconnect-65857.json
+++ b/.changes/next-release/feature-ec2instanceconnect-65857.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "``ec2-instance-connect``",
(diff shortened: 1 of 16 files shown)


kdaily (approved):
:shipit:

Issue #7970, opened 2023-06-13
aws: error: The step type spark_application is not supported.


Hongbo-Miao:
### Describe the feature

I have an EMR cluster with step "Spark application" which created from EMR UI.

![image](https://github.com/aws/aws-cli/assets/3375461/5db3b033-6a91-4e92-918f-059b3aac0313)

Now I am hoping to create it by CLI.

I got the command by clicking "View command for cloning cluster" button

![image](https://github.com/aws/aws-cli/assets/3375461/c395203f-d407-4d46-becc-09f74a43c1a0)

It gives me command like

```shell
aws emr create-cluster \
    # ...
    --steps '[{"Name":"xxx","ActionOnFailure":"TERMINATE_CLUSTER","Jar":"command-runner.jar","Properties":"","Args":["spark-submit","--deploy-mode","cluster","s3://my-bucket/MySparkApp-assembly-1.0.jar"],"Type":"SPARK_APPLICATION"}]'
```

However, when I run this command, I got error

> aws: error: The step type spark_application is not supported.

Based on [CLI document](https://docs.aws.amazon.com/cli/latest/reference/emr/create-cluster.html), currently `--steps` only supports

> "Type": "CUSTOM_JAR"|"STREAMING"|"HIVE"|"PIG"|"IMPALA"

It would be great to support `SPARK_APPLICATION`, thanks! 😃

### Use Case

It would be convenient to use CLI to create EMR cluster with step "Spark application".

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

aws-cli/2.8.12 Python/3.9.11 Darwin/22.5.0 exe/x86_64 prompt/off

### Environment details (OS name and version, etc.)

macOS

RyanFitzSimmonsAK:
Hi @Hongbo-Miao, thanks for reaching out.  In the [developer guide](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-spark-submit-step.html) for EMR, it looks like the examples use `Type=Spark` rather than `SPARK_APPLICATION`.  Could you try formatting it similarly to the examples, and letting me know if that works?  Thanks!

Hongbo-Miao:
Hi @RyanFitzSimmonsAK hmm the error is something else now

> An error occurred (ValidationException) when calling the RunJobFlow operation: An instance group must have at least one instance.

Although this new error also should not exist as I exported, when I rerun, it should create same EMR cluster.

[...]

PR #7973, opened 2023-06-13
fix(emr): add missing SPARK type

hongbo-miao:
Issue #7970

*Description of changes:*

Add missing "SPARK" type.


tim-finnigan:
It looks like this was addressed: https://docs.aws.amazon.com/cli/latest/reference/emr/create-cluster.html

image


Please create a new issue if you are encountering any other problems related to this.

PR #8006, opened 2023-06-27
 [v2] Implement configured endpoint URL resolution

kdaily:
Adds a config provider to resolve the endpoint URL provided in an environment variable or shared configuration file. It resolves the endpoint in the following manner:

1. The value provided through the `endpoint_url` parameter provided to the client constructor.
2. The value provided by a service-specific environment variable.
3. The value provided by the global endpoint environment variable (`AWS_ENDPOINT_URL`).
4. The value provided by a service-specific parameter from a services definition section in the shared configuration file.
5. The value provided by the global parameter from a services definition section in the shared configuration file.
6. The value resolved when no configured endpoint URL is provided.

The endpoint config provider uses the client name (name used to instantiate a client object) for construction and add to the config value store. This uses multiple lookups to handle service name changes for backwards compatibility.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/botocore/args.py
+++ b/awscli/botocore/args.py
@@ -70,7 +70,7 @@ class ClientArgsCreator(object):
         s3_config = final_args['s3_config']
         partition = endpoint_config['metadata'].get('partition', None)
         socket_options = final_args['socket_options']
-
+        configured_endpoint_url = final_args['configured_endpoint_url']
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
 
@@ -108,7 +108,7 @@ class ClientArgsCreator(object):
             service_model,
             endpoint_region_name,
             region_name,
-            endpoint_url,
+            configured_endpoint_url,
             endpoint,
             is_secure,
             endpoint_bridge,
@@ -151,10 +151,16 @@ class ClientArgsCreator(object):
                 user_agent += ' %s' % client_config.user_agent_extra
 
         s3_config = self.compute_s3_config(client_config)
+
+        configured_endpoint_url = self._compute_configured_endpoint_url(
+            client_config=client_config,
+            endpoint_url=endpoint_url,
+        )
+
         endpoint_config = self._compute_endpoint_config(
             service_name=service_name,
             region_name=region_name,
-            endpoint_url=endpoint_url,
+            endpoint_url=configured_endpoint_url,
             is_secure=is_secure,
             endpoint_bridge=endpoint_bridge,
             s3_config=s3_config,
@@ -196,12 +202,34 @@ class ClientArgsCreator(object):
             'service_name': service_name,
             'parameter_validation': parameter_validation,
             'user_agent': user_agent,
+            'configured_endpoint_url': configured_endpoint_url,
             'endpoint_config': endpoint_config,
             'protocol': protocol,
             'config_kwargs': config_kwargs,
             's3_config': s3_config,
             'socket_options': self._compute_socket_options(scoped_config)
         }
+
+    def _compute_configured_endpoint_url(self, client_config, endpoint_url):
+        if endpoint_url is not None:
+            return endpoint_url
+
+        if self._ignore_configured_endpoint_urls(client_config):
+            logger.debug("Ignoring configured endpoint URLs.")
+            return endpoint_url
+
+        return self._config_store.get_config_variable('endpoint_url')
+
+    def _ignore_configured_endpoint_urls(self, client_config):
+        if (
+            client_config
+            and client_config.ignore_configured_endpoint_urls is not None
+        ):
+            return client_config.ignore_configured_endpoint_urls
+
+        return self._config_store.get_config_variable(
+            'ignore_configured_endpoint_urls'
+        )
 
     def compute_s3_config(self, client_config):
         s3_configuration = self._config_store.get_config_variable('s3')
--- a/awscli/botocore/configprovider.py
+++ b/awscli/botocore/configprovider.py
@@ -13,6 +13,7 @@
 """This module contains the inteface for controlling how configuration
 is loaded.
 """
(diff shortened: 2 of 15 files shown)


awscli/botocore/args.py
```diff
@@ -196,13 +202,33 @@ def compute_client_args(self, service_model, client_config,
...
             'config_kwargs': config_kwargs,
             's3_config': s3_config,
             'socket_options': self._compute_socket_options(scoped_config)
         }
 
+    def _compute_configured_endpoint_url(self, client_config, endpoint_url):
+        if endpoint_url is not None or self._ignore_configured_endpoint_urls(
+            client_config
+        ):
+           return endpoint_url
```
kyleknap, line 217: It looks like we are missing the debug line that got added toward the end of the botocore PR: https://github.com/boto/botocore/pull/2973/files#diff-019afcd91cca6d46dd1046132b61e78198063592925d01e9a4db1fd9f1bf8c3fR281

awscli/botocore/utils.py
```diff
@@ -3302,3 +3302,79 @@ def is_s3_accelerate_url(url):
...
+    'lexv2-models': 'lex-models-v2',
+    'lex-runtime': 'lex-runtime-service',
+    'lexv2-runtime': 'lex-runtime-v2',
+    'logs': 'cloudwatch-logs',
+    'machinelearning': 'machine-learning',
+    'marketplacecommerceanalytics': 'marketplace-commerce-analytics',
+    'marketplace-entitlement': 'marketplace-entitlement-service',
+    'meteringmarketplace': 'marketplace-metering',
+    'mgh': 'migration-hub',
+    'sms-voice': 'pinpoint-sms-voice',
```
kyleknap, line 3368: This dictionary has one more element than the one that is removed from the `test_endpoints.py` file. I believe this is `sms-voice` which we removed the model outright in CLI v2. We should be able to remove the `sms-voice` entry from this dictionary.

awscli/botocore/session.py
```diff
@@ -840,6 +841,12 @@ def create_client(self, service_name, region_name=None,
         endpoint_resolver = self._get_internal_component('endpoint_resolver')
         exceptions_factory = self._get_internal_component('exceptions_factory')
         config_store = self.get_component('config_store')
```
kyleknap, line 843: This needs to be a shallow copy of the config store here

awscli/botocore/configprovider.py
```diff
@@ -561,3 +568,159 @@ def provide(self):
 
     def __repr__(self):
         return 'ConstantProvider(value=%s)' % self._value
+
+
+class ConfiguredEndpointProvider(BaseProvider):
```
kyleknap, line 573: This class does not match up with what we have in the botocore PR: https://github.com/boto/botocore/pull/2973/files#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R862. Specifically I still see the `__repr__` and setting of `self._transformed_service_id`. It might be worth just copying this class over in its entirety to this PR.

awscli/botocore/configprovider.py
```diff
@@ -89,6 +90,12 @@
         'use_fips_endpoint',
         'AWS_USE_FIPS_ENDPOINT',
         None, utils.ensure_boolean),
+    'ignore_configured_endpoint_urls': (
+        'ignore_configured_endpoint_urls',
+        'AWS_IGNORE_CONFIGURED_ENDPOINT_URLS',
+        None,
+        utils.ensure_boolean,
+    ),
```
kyleknap, line 98: We will still need to implement the `__copy__()` method on the `ConfigValueStore` store

tests/functional/botocore/test_endpoints.py
```diff
@@ -13,68 +13,7 @@
 import pytest
```
kyleknap, line 13: I think we are missing some of the functional tests from `tests/functional/configured_endpoint_urls/` directory in the original PR

tests/unit/botocore/test_config_provider.py
```diff
@@ -556,3 +557,206 @@ def test_provide_section_config_with_only_overrides(self):
                 'override_var': 'override',
             }
         )
+
```
kyleknap, line 560: We should make sure to port over some of the shallow copy tests for the `ConfigValueStore` in the original PR: https://github.com/boto/botocore/pull/2973/files#diff-0d043ca2ec3f69519455add1a50265e2a7fa3c0296392ea9b8a4c9578794f6d7R377

kyleknap (commented):
Just leaving some feedback on this initial port

tests/functional/botocore/configured_endpoint_urls/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,249 @@
...
+
+def _known_service_names_and_models():
+    my_session = botocore.session.get_session()
+    loader = my_session.get_component('data_loader')
+    available_services = loader.list_available_services('service-2')
+
+    result = []
+    for service_name in available_services:
+        model = my_session.get_service_model(service_name)
+        result.append((model.service_name, model))
```
kyleknap, line 169: I think we are running into memory issues in the Windows jobs. Specifically, I think it is because we load all of the models as part of parametrization and that lingers in memory until the test execution finish. This could result in quite a bit of memory being used depending on how many copies of all of the models we have in memory from test collection. I think we should change this test to parametrize off of service id's. I tried it on my fork and that seemed to do the trick in getting the Windows CI passing: https://github.com/kyleknap/aws-cli/commit/ee21da40fdb92d22325f0f716e72f36569e77812

kyleknap (commented):
Looks good. Just had one more comment related to getting CI to pass.

kyleknap (approved):
Looks good! :ship:

Issue #8036, opened 2023-07-17
Unable to install aws-cli via pip - package dependency broke


AMO-brandon:
### Describe the bug

$ pip3 install awscli
Collecting awscli
  Obtaining dependency information for awscli from https://files.pythonhosted.org/packages/43/9c/bf16d97f5de8aa4f9171c6c82bad0b4179921ddca5066ba980358da0e9a5/awscli-1.29.3-py3-none-any.whl.metadata
  Downloading awscli-1.29.3-py3-none-any.whl.metadata (11 kB)
Collecting botocore==1.31.3 (from awscli)
  Obtaining dependency information for botocore==1.31.3 from https://files.pythonhosted.org/packages/b0/f0/5755508b3305534cd4cf2a8a82bbbe42ee9d66fd2688be5ff3dfb85e9a99/botocore-1.31.3-py3-none-any.whl.metadata
  Downloading botocore-1.31.3-py3-none-any.whl.metadata (5.9 kB)
Collecting docutils<0.17,>=0.10 (from awscli)
  Downloading docutils-0.16-py2.py3-none-any.whl (548 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 548.2/548.2 kB 34.8 MB/s eta 0:00:00
Collecting s3transfer<0.7.0,>=0.6.0 (from awscli)
  Downloading s3transfer-0.6.1-py3-none-any.whl (79 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 79.8/79.8 kB 12.2 MB/s eta 0:00:00
Collecting PyYAML<5.5,>=3.10 (from awscli)
  Downloading PyYAML-5.4.1.tar.gz (175 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 175.1/175.1 kB 24.9 MB/s eta 0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [62 lines of output]
      running egg_info
      writing lib3/PyYAML.egg-info/PKG-INFO
      writing dependency_links to lib3/PyYAML.egg-info/dependency_links.txt
[...]

PR #8037, opened 2023-07-17
Add support for PyYAML 6.0

kyleknap:
This should partially address this issue: https://github.com/aws/aws-cli/issues/8036. Specifically, PyYAML 6.0 has wheels for Python 3.10 and Python 3.11. However, it won't fix the issue for users that are on platform with no PyYAML wheels (e.g. Alpine). The only way for that to be addressed is through an upstream PyYAML patch (e.g. https://github.com/yaml/pyyaml/pull/702) and release.

Changelog for 6.0: https://github.com/yaml/pyyaml/blob/957ae4d495cf8fcb5475c6c2f1bce801096b68a5/CHANGES#L7-L18

Diff changes between versions: https://github.com/yaml/pyyaml/compare/5.4.1...6.0

### DIFF
--- a/.changes/next-release/enhancement-dependency-21293.json
+++ b/.changes/next-release/enhancement-dependency-21293.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "dependency",
+  "description": "Support PyYAML 6.0"
+}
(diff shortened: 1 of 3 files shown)


nateprewitt (approved):
⛵

hssyoo (approved):

PR #8038, opened 2023-07-17
Add --no-build-isolation to make-bundle script

kyleknap:
This will ensure we strictly pull down the sdist and not try into install any other build dependencies (e.g. Cython) as part of downloading our main runtime dependencies.


nateprewitt (approved):
:shipit:

Issue #8023, opened 2023-07-07
ec2-instance-connect open-tunnel failed when NO_PROXY not set.


sayurin:
### Describe the bug

ec2-instance-connect open-tunnel failed when NO_PROXY not set.

### Expected Behavior

open-tunnel successfully connected.

### Current Behavior

following error message.

```
2023-07-07 17:47:08,068 - MainThread - awscli.clidriver - DEBUG - Exception caught in main()
Traceback (most recent call last):
  File "awscli\clidriver.py", line 460, in main
  File "awscli\clidriver.py", line 595, in __call__
  File "awscli\customizations\commands.py", line 205, in __call__
  File "awscli\customizations\ec2instanceconnect\opentunnel.py", line 145, in _run_main
  File "awscli\customizations\ec2instanceconnect\websocket.py", line 326, in run
  File "awscli\customizations\ec2instanceconnect\websocket.py", line 363, in _open_websocket_connection
  File "awscli\customizations\ec2instanceconnect\websocket.py", line 158, in connect
TypeError: argument of type 'NoneType' is not iterable

argument of type 'NoneType' is not iterable
kex_exchange_identification: Connection closed by remote host
```

### Reproduction Steps

1. Setup EC2 Instance, EC2 Instance Connect and EC2 Instance Connect Endpoint. (Instance ID as i-xxxx).
2. Setup HTTP Proxy (Proxy address as proxy.example.com:8080).
3. Set environment variable `HTTP_PROXY` to use proxy.
    In bash, `export HTTP_PROXY=http://proxy.example.com:8080`.
    In PowerShell `$env:HTTP_PROXY="http://proxy.example.com:8080"`.
4. Unset env `NO_PROXY`.
    In bash, `unset NO_PROXY`.
    In PowerShell, `$env:NO_PROXY=""`. 
5. Run aws cli `aws ec2-instance-connect ssh --connection-type eice --instance-id i-xxxx --debug`.
    This will fail with the error written at Current Behavior.

Workarround,

1. Set environment variable `NO_PROXY` to dummy value.
    In bash, `export NO_PROXY=1`.
    In PowerShell, `$env:NO_PROXY="1"`.
2. Run aws cli `aws ec2-instance-connect ssh --connection-type eice --instance-id i-xxxx --debug`.
    This will success as expected behavior.

### Possible Solution

[...]

PR #8039, opened 2023-07-17
Fix open tunnel bug when NO_PROXY is unset

hsingh6764:
https://github.com/aws/aws-cli/issues/8023

*Description of changes:*
Allow NO_PROXY to be `None`. If it is not defined or is empty or `None` then we don't check.

### DIFF
--- a/awscli/customizations/ec2instanceconnect/websocket.py
+++ b/awscli/customizations/ec2instanceconnect/websocket.py
@@ -154,7 +154,7 @@ class Websocket:
         environment = os.environ.copy()
         proxy_options = None
         proxy_url = environment.get("HTTP_PROXY") or environment.get("HTTPS_PROXY")
-        no_proxy = environment.get("NO_PROXY")
+        no_proxy = environment.get("NO_PROXY", "")
         if proxy_url and parsed_url.hostname not in no_proxy:
             parsed_proxy_url = urlparse(proxy_url)
             logger.debug(f"Using the following proxy: {parsed_proxy_url.hostname}")
--- a/tests/unit/customizations/ec2instanceconnect/test_websocket.py
+++ b/tests/unit/customizations/ec2instanceconnect/test_websocket.py
@@ -294,6 +294,41 @@ class TestWebsocket:
 
     @patch.object(websocket, "connect")
     @patch.object(websocket, "create_handshake_request")
+    @mock.patch.dict(os.environ, {"HTTPS_PROXY": "http://user1:pass1@localhost:8989"})
+    def test_connect_with_proxy_but_no_proxy_env_empty(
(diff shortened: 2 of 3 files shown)


awscli/customizations/ec2instanceconnect/websocket.py
```diff
@@ -155,7 +155,7 @@ def connect(self, url, user_agent=None):
         proxy_options = None
         proxy_url = environment.get("HTTP_PROXY") or environment.get("HTTPS_PROXY")
         no_proxy = environment.get("NO_PROXY")
-        if proxy_url and parsed_url.hostname not in no_proxy:
+        if proxy_url and (no_proxy is None or parsed_url.hostname not in no_proxy):
```
kyleknap, line 158: To simplify this down we should be able to just add a default to the environment `get()`:
```python
no_proxy = environment.get("NO_PROXY", "")
```
and from there we would not need to expand the if statement at all.

tests/unit/customizations/ec2instanceconnect/test_websocket.py
```diff
@@ -292,6 +292,41 @@ def test_connect_with_proxy(
 
         assert mock_on_connection_event.wait.called
 
+    @patch.object(websocket, "connect")
+    @patch.object(websocket, "create_handshake_request")
+    @mock.patch.dict(os.environ, {"HTTPS_PROXY": "http://user1:pass1@localhost:8989", "NO_PROXY": ""})
```
kyleknap, line 297: Let's remove the `NO_PROXY` env var in patched environment variables to make sure the test case represents the bug that this fixes, which is when `HTTPS_PROXY` but `NO_PROXY` is not present as an environment variable.

awscli/customizations/ec2instanceconnect/websocket.py
```diff
@@ -155,7 +155,7 @@ def connect(self, url, user_agent=None):
         proxy_options = None
         proxy_url = environment.get("HTTP_PROXY") or environment.get("HTTPS_PROXY")
         no_proxy = environment.get("NO_PROXY")
```
kyleknap, line 157: Let's make sure to add a changelog entry for this fix. From the root of the repository, you should be able to run the `new-change` script to generate one (feel free to use or repurpose this):
```
~/GitHub/aws-cli$ ./scripts/new-change -t bugfix -c '``ec2-instance-connect ssh``' -d 'Fix runtime error when HTTP_PROXY environment variable is set and NO_PROXY is not set for eice connection type #8023'
```
This will add a new json file for this change at `.changes/next-release` that we can than add as part of the PR.

kyleknap (commented):
Looks good. I just had a few smaller comments and we should be good to merge this.

kyleknap (approved):
Looks good! Merging.

PR #8040, opened 2023-07-17
Update setup.cfg

ishu202:
fixed the pyyaml version to less than 5.4 for it to work in python 3.11 with pip 23.*

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ requires_dist =
         botocore==1.31.3
         docutils>=0.10,<0.17
         s3transfer>=0.6.0,<0.7.0
-        PyYAML>=3.10,<6.1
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @ishu202, this is unfortunately not the correct fix for this issue. You can follow along with the current work in https://github.com/aws/aws-cli/issues/8036#issuecomment-1638981884.

ishu202:
Hey @nateprewitt that's correct it's not the correct fix although based on this comment https://github.com/aws/aws-cli/issues/8036#issuecomment-1638981884 we probably need to wait until pyyaml it self upgrades a version for alpine. in the mean time I fixed my issue by changing the deps. thanks for the feedback.

PR #8053, opened 2023-07-21
preserves timestamps when copying files to temporary directory

jtheuer:
*Issue [#3131](https://github.com/aws/aws-cli/issues/3131)*

*Description of changes:*

Under (some?) circumstances the `cloudformation package` command copies the files to a temporary directory before creating the ZIP file. This updates the timestamps and therefore results in a new s3 resource when upload. This triggers always a redeploy of the affected resource, even when this wasn't required. 

This change at least keeps the original timestamps when copying the file. This limits the cases where the resources are updated unnecessarily.


dmarinuswoodwing:
I just tested this and this seems to work, thanks!

I can't believe this hasn't been merged for 2 years, such an obvious fix with such a low risk.

PR #8100, opened 2023-08-14
Add dependabot for v1

nateprewitt:
This PR will setup dependabot for the v1 (develop) branch to monitor both third party dependencies on a weekly cadence, as well as keep our github actions up to date. This will be followed by a separate PR to move us to strict pins for GHA packages to meet OpenSSF guidance.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,14 @@
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "develop"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+
   - package-ecosystem: "pip"
     directory: "/"
     open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -23,3 +33,20 @@ updates:
...
+    schedule:
+      interval: "weekly"
+    target-branch: "develop"
+    labels:
+      - "dependencies"
+      - "v1"
+    allow:
+      - dependency-name: "colorama"
+      - dependency-name: "docutils"
+      - dependency-name: "awscrt"
```
nateprewitt, line 49: This actually probably doesn't need to be managed here. It would be pulled in through botocore.
```suggestion
```

.github/dependabot.yml
```diff
@@ -23,3 +33,19 @@ updates:
       - dependency-name: "jmespath"
       - dependency-name: "urllib3"
       - dependency-name: "wheel"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    open-pull-requests-limit: 10
+    schedule:
+      interval: "weekly"
```
kdaily, line 41: nit: the v2 schedule is daily, any reason this is weekly?

nateprewitt: Primarily to limit noise. Given we haven't merged any of these same day, it just saves on wasted runners. Happy to move to a higher cadence if we think we'll action on that.

kdaily (approved):
Thanks! 🚢

PR #8106, opened 2023-08-15
Add support for Python 3.12

nateprewitt:
## Python 3.12 
This PR will add support for Python 3.12 to the AWS CLI. This will ensure we're properly testing ahead of the official release on October 2nd, 2023.

The change highlights are:

1. Running tests off the latest release candidate for Python 3.12
2. Updating our testing dependencies to avoid new warnings in the test suite
3. Install setuptools in all CI builds with Python 3.12+ where it's no longer included by default
4. Updating the classifiers and tox file for Python 3.12
4. Fixing EKS' test suite

## EKS
There is a separate small commit here to fix an oddity in the EKS test suite that's surfaced in the Python 3.12 build. When tests were added for Kubeconfig, they had calls to an erroneous `called_with` method on the mock (not to be confused with the correct [`assert_called_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with)). When this is called, the Mock simply creates a new Mock to represent what the imaginary call might return. Because the test suite never actually uses an `assert` directive or a unittest assertion, the test just passes doing nothing.

```python
>>> from unittest import mock
>>> m = mock.Mock()
>>> m.called_with("Nonsense")
<Mock name='mock.called_with()' id='4345973448'>
```

In order to fix this issue, I've moved the `called_with` calls to `assert_called_with`. This introduced another issue though that Kubeconfig has no concept of equivalence beyond a strict object reference match. This means without further mocking or rewriting the tests, we can't use this call.

To resolve this, I've added an `__eq__` method on `Kubeconfig` to perform a straight forward comparison of the type and all members. This will give us the standard behavior of `==` instead of `is` when calling `assert_called_with`.

This issue also exists in codedeploy ([here](https://github.com/aws/aws-cli/blob/e639ac458e3c0e90ff15fd1205adfe5e3582a9ec/tests/unit/customizations/codedeploy/test_install.py#L162) and [here](https://github.com/aws/aws-cli/blob/e639ac458e3c0e90ff15fd1205adfe5e3582a9ec/tests/unit/customizations/codedeploy/test_install.py#L180)), but those tests are legitimately broken. I'm going to save touching those for another PR and provided the minimum here to get us out the door with 3.12.


scripts/ci/install-dev-deps
```diff
@@ -25,4 +26,7 @@ def run(command):
 
 if __name__ == "__main__":
     with cd(REPO_ROOT):
+        if sys.version_info[:2] >= (3, 12):
+            run("pip install setuptools")
```
kyleknap, line 30: Do we want to bound the version of setuptools in this install command? It feels like it would probably be worth it to make sure we are pulling in a consistent version of setuptools for our CI.

kyleknap (commented):
Looks fine. I just had one comment.

scripts/ci/install-dev-deps
dlm6693: I asked a similar question for the equivalent PR in botocore https://github.com/boto/botocore/pull/2967#discussion_r1293967757

Issue #7739, opened 2023-03-09
aws-cliv2 doesn't honor tab completion


ArjunDandagi:
### Describe the bug

i can use --profile and press <tab> to get list of all the profiles i can pass 
only with the aws sts get-caller-identity command 

if i try aws s3 ls --profile <tab> . it doesn't work 

### Expected Behavior

Tab completion for --profile should work for all commands . 

### Current Behavior

tab completion only works for `aws sts` command

### Reproduction Steps

1. have multiple profile configured 
2. ensure you have complete script is loaded to your terminal session 
3. type aws sts get-caller-identity --profile <tab>(press tab button) . you will see list of all the profiles you can pass 
4. now try with any other command for.example 

aws s3 ls --profile <tab> (see that it doesn't work)

### CLI version used

aws-cli/2.9.20 

### Environment details (OS name and version, etc.)

Python/3.11.2 Darwin/22.3.0 source/arm64 prompt/off

tim-finnigan:
Hi @ArjunDandagi thanks for reaching out. I could reproduce the issue with `aws s3` commands but the tab completion for `--profile` worked for other commands, for example:

`aws s3api list-buckets --profile ...`

Did you encounter the issue outside of `s3` commands? This may just be an edge case with `s3` being a customization.

ArjunDandagi:
@tim-finnigan , you are right , its not working for `s3` only. tab completion for profiles works for other commands . i tried , sts , iam , ec2 , lex , and some other command . Thanks .

ArjunDandagi:
I also noticed today that it doesn't work for `accessanalyzer` command as well

PR #8126, opened 2023-08-23
Fix --profile completion for aws s3 ls

REslim30:
Issue #7739

Bug also affects any option completions following commands that expect positional args

*Description of changes:*
- Adds a check in the parser that ensures state isn't currently at a parameter before assuming current fragment is a positional.
- Also fixes some grammatical errors in comments

[Screencast from 23-08-23 16:58:46.webm](https://github.com/aws/aws-cli/assets/43259657/148cb810-827b-4b9e-b716-0045803c44f0)


REslim30:
Some builds are failing:
image

But I can't seem to reproduce it locally (I'm on Ubuntu), are these flaky tests or real failures?

PR #8133, opened 2023-08-25
Add ability to specify a yaml file as parameters

### LINKED ISSUE #2275: Cloudformation external parameter file do not support YAML file

aidando73:
*Issue #2275 

- Adds ability to specify a yaml file as parameters
- Will work for any list yaml files (but not for dictionary yaml files)

[Screencast from 25-08-23 18:45:15.webm](https://github.com/aws/aws-cli/assets/43259657/cc8e320a-13f5-4949-a3a4-d98bec8f843d)

### DIFF
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -13,6 +13,9 @@
 """Module for processing CLI args."""
 import os
 import logging
+
+from ruamel.yaml import YAML, YAMLError
+
 from awscli.compat import six
 
 from botocore.compat import OrderedDict, json
@@ -188,6 +191,16 @@ def _unpack_complex_cli_arg(argument_model, value, cli_name):
         if isinstance(value, six.string_types):
             if value.lstrip()[0] == '[':
                 return _unpack_json_cli_arg(argument_model, value, cli_name)
+            else:
+                # Value could be yaml so try parsing as yaml
+                try:
+                    yaml = YAML(typ='safe')
+                    result = yaml.load(value)
+                    if isinstance(result, list):
+                        return result
+                except YAMLError as e:
+                    # If it's not valid yaml, then we can continue on
+                    pass
         elif isinstance(value, list) and len(value) == 1:
             single_value = value[0].strip()
             if single_value and single_value[0] == '[':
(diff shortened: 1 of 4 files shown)


awscli/argprocess.py
```diff
@@ -406,6 +419,20 @@ def _should_parse_as_shorthand(self, cli_argument, value):
...
                       "param shorthand.", cli_argument.py_name)
             return False
+        # The second short-circuit case is if the argument is yaml
+        if isinstance(check_val, six.string_types):
+            try:
+                yaml = YAML(typ='safe')
+                result = yaml.load(check_val)
+                # Yaml parsing is only supported for yaml list documents as there are currently no
+                # known use cases for yaml map documents
+                if isinstance(result, list):
```
REslim30, line 429: Wasn't sure if there was a use case for dictionary yaml documents. Let me know if you guys want to support it.

REslim30:
I just realized there's another PR that aims to achieves something similar https://github.com/aws/aws-cli/pull/5482/ but with a different approach.

Up to you guys which approach you want to choose.

REslim30:
One drawback of my approach is that it's difficult to provide useful error messages; It's difficult to determine whether the argument passed in is an invalid yaml document or shorthand or JSON in a clean way.

Alternatives might be:
- Parse yaml files based on file extensions .yml or .yaml (not sure if there are use cases where file extensions might not be .yml or .yaml)
- Go with the yaml:// prefix idea 

These would allow better error messages

REslim30:
The more I think about it, the more file extension based matching makes more sense. Let me know if you want me to redo this PR

Issue #7087, opened 2022-07-04
Make env AWS_PROFILE optional on update-kubeconfig


rubensa:
### Describe the feature

Since https://github.com/aws/aws-cli/pull/3683 when `aws eks update-kubeconfig` is run the current active profile (if any) is always included in **kube** **config** **env** section (**AWS_PROFILE**).

Wee should be able to disable this auto inclusion.

### Use Case

We want to create **Kubernetes** config using an active profile but been able to use the same configuration with any other profile (without the need to reconfigure the kubernetes config).

If the **kube** **config** **env** section with the **AWS_PROFILE** is not set, the actual **AWS_PROFILE** is the currently active profile (if any) when a **kubectl** command is issued.

### Proposed Solution

Add an option **--include-profile**

- **true** (default): works like now to keep compatibility (the current active profile, if any, is included in kube config env section)
- **false**: the env section with the profile information is not included (works like pre https://github.com/aws/aws-cli/pull/3683)

### Other Information

For now, every time I want to use a different profile I have to update the Kubernetes config with `aws eks update-kubeconfig` and then use **kubectl**.

Alternatively I can manually edit **.kube/config** and comment the env parts (only once, but remember to do so every time I update the kube config) like so:
```
      #env:
      #- name: AWS_PROFILE
      #  value: profile
```
and I don't have to run `aws eks update-kubeconfig`  every time I change the current profile.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

aws-cli/2.7.4

### Environment details (OS name and version, etc.)

Ubuntu 22.04

tim-finnigan:
Thanks @rubensa for the feature request. The EKS team owns this customization so I reached out to them regarding your request. I will update this issue when I hear back.

tim-finnigan:
P67449827

rubensa:
**As a side note**: I also tried to use `kubectl config unset` command to "manually" remove the generated `env` part after issuing the `aws eks update-kubeconfig` command but couldn't find right way to do it (see: https://discuss.kubernetes.io/t/kubectl-config-unset/20502)

tim-finnigan:
I heard back from a member of the EKS team and they are willing to take this request into consideration during their roadmap planning. In the meantime we can leave this open for tracking, and we encourage others who are interested in this feature to 👍 the issue or leave a comment if they have additional use cases to share.

heesuk-ahn:
[...]

PR #7845, opened 2023-04-22
Add --exclude-profile option to eks update-kubeconfig

soggycactus:
*Issue #, if available:*

This PR resolves https://github.com/aws/aws-cli/issues/7087

*Description of changes:*

This PR adds a new option to `aws eks update-kubeconfig` called `exclude-profile`. Currently, when using `update-kubeconfig`, if an AWS profile session is detected the generated kubeconfig will include an `env` block with `AWS_PROFILE` in the user's exec config. Example:

```
users:
- name: arn:aws:eks:region:111222333444:cluster/ExampleCluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - region
      - eks
      - get-token
      - --cluster-name
      - ExampleCluster
      - --output
      - json
      command: aws
      env:
        - name: AWS_PROFILE
          value: <profile-name>
```
This may be undesirable behavior depending on how you manage your AWS credentials. When specifying the `--exclude-profile` option added in this PR, the `env` block will not be added.

Rather than as described in the issue above, I chose to add `exclude-profile` instead of `include-profile` to keep consistent behavior with previous versions of awscli. 

I added a unit test to ensure the option works properly. The test injects an `AWS_PROFILE` environment variable at runtime to simulate an AWS "session" and passes `--exclude-profile` to the command, and asserts that the resulting kubeconfig does _not_ have an `env` block; if `exclude_profile` is set to `False`, the test fails (as it should). 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. I am available to make any changes needed for this PR to be merged.

PR #8163, opened 2023-09-08
Add --exclude-profile option to eks update-kubeconfig

soggycactus:
*Issue #, if available:*

This PR resolves https://github.com/aws/aws-cli/issues/7087, and replaces https://github.com/aws/aws-cli/pull/7845

*Description of changes:*

This PR adds a new option to `aws eks update-kubeconfig` called `exclude-profile`. Currently, when using `update-kubeconfig`, if an AWS profile session is detected the generated kubeconfig will include an `env` block with `AWS_PROFILE` in the user's exec config. Example:

```
users:
- name: arn:aws:eks:region:111222333444:cluster/ExampleCluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - region
      - eks
      - get-token
      - --cluster-name
      - ExampleCluster
      - --output
      - json
      command: aws
      env:
        - name: AWS_PROFILE
          value: <profile-name>
```
This may be undesirable behavior depending on how you manage your AWS credentials. When specifying the `--exclude-profile` option added in this PR, the `env` block will not be added.

Rather than as described in the issue above, I chose to add `exclude-profile` instead of `include-profile` to keep consistent behavior with previous versions of awscli. 

I added a unit test to ensure the option works properly. The test injects an `AWS_PROFILE` environment variable at runtime to simulate an AWS "session" and passes `--exclude-profile` to the command, and asserts that the resulting kubeconfig does _not_ have an `env` block; if `exclude_profile` is set to `False`, the test fails (as it should). 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. I am available to make any changes needed for this PR to be merged.

### DIFF
--- a/tests/unit/customizations/eks/test_update_kubeconfig.py
+++ b/tests/unit/customizations/eks/test_update_kubeconfig.py
@@ -217,7 +217,7 @@ class TestEKSClient(unittest.TestCase):
         self._session.create_client.return_value = self._mock_client
         self._session.profile = None
 
-        self._client = EKSClient(self._session, parsed_args=Namespace(cluster_name="ExampleCluster", role_arn=None))
+        self._client = EKSClient(self._session, parsed_args=Namespace(cluster_name="ExampleCluster", role_arn=None, exclude_profile=False))
 
     def test_get_cluster_description(self):
         self.assertEqual(self._client.cluster_description,
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -103,7 +103,14 @@ class UpdateKubeconfigCommand(BasicCommand):
             'help_text': ("Alias for the generated user name. "
                           "Defaults to match cluster ARN."),
             'required': False
-        }
+        },
+        {
+            'name': 'exclude-profile',
+            'action': 'store_true',
+            'default': False,
+            'help_text': ("Exclude the AWS_PROFILE environment variable "
+                          "from the generated user's exec config.")
+        },
     ]
 
     def _display_entries(self, entries):
@@ -332,7 +339,7 @@ class EKSClient(object):
                 self._parsed_args.role_arn
             ])
 
-        if self._session.profile:
+        if self._session.profile and not self._parsed_args.exclude_profile:
             generated_user["user"]["exec"]["env"] = [OrderedDict([
                 ("name", "AWS_PROFILE"),
                 ("value", self._session.profile)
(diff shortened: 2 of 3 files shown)


tests/unit/customizations/eks/test_update_kubeconfig.py
```diff
@@ -217,7 +217,7 @@ def setUp(self):
         self._session.create_client.return_value = self._mock_client
         self._session.profile = None
 
-        self._client = EKSClient(self._session, parsed_args=Namespace(cluster_name="ExampleCluster", role_arn=None))
+        self._client = EKSClient(self._session, parsed_args=Namespace(cluster_name="ExampleCluster", role_arn=None, exclude_profile=False))
```
yuxiang-zhang, line 220:
```suggestion
        self._client = EKSClient(self._session, parsed_args=Namespace(cluster_name="ExampleCluster"))
```
Just realized that I set a bad example here. Since these are just default values, they don't need to be specified here.

soggycactus: I thought so too, but the test actually fails without this - I'm not sure why defaults aren't being passed here; the functional tests & running the cli via an egg both work properly

soggycactus: Trying your suggestion I get ```AttributeError: 'Namespace' object has no attribute 'role_arn'``` (I would get the same thing respectively with `exclude_profile`

yuxiang-zhang: Ah understood, this is a Namespace from a clean slate, so it doesn't know about the args we defined

yuxiang-zhang (approved):

soggycactus:
@tim-finnigan Do we need a review from an "official" maintainer here? It looks like @yuxiang-zhang has taken a look and approved the changes

soggycactus:
> @tim-finnigan Do we need a review from an "official" maintainer here? It looks like @yuxiang-zhang has taken a look and approved the changes

CC @kyleknap @RedbackThomson

tim-finnigan:
Yes an official maintainer like @kyleknap will need to review this prior to merging.

soggycactus:
@kyleknap friendly bump!

soggycactus:
Hey @tim-finnigan @kyleknap could we get a quick review here? The change is small and it would be great to finally close the loop on this issue

soggycactus:
@tim-finnigan @kyleknap 

image

soggycactus:
@tim-finnigan @kyleknap Hey any chance I can get a review here? I know this is a relatively unimportant change but it's really small and I'd like to close the loop on this PR considering I went through the effort of implementing this, having to wait on an upstream PR approval, re-base changes based off of that, and am now waiting again. I know you guys are busy but I'd really just appreciate anything other than silence here - even a "hey we're gonna close this"

PR #8167, opened 2023-09-10
Improve error messages for codeartifact login

dafangc-amazon:
*Issue #, if available:*

*Description of changes:*

Currently codeartifact login command would show the error message from the subprocess in case of failures, which doesn't help with debugging because the message would only contain the exit code. This commit includes the error message from the package tool CLI that is run by the login.

### DIFF
--- a/awscli/customizations/codeartifact/login.py
+++ b/awscli/customizations/codeartifact/login.py
@@ -10,7 +10,7 @@ from dateutil.tz import tzutc
 from dateutil.relativedelta import relativedelta
 from botocore.utils import parse_timestamp
 
-from awscli.compat import is_windows, urlparse, RawConfigParser, StringIO
+from awscli.compat import is_windows, urlparse, RawConfigParser, StringIO, get_stderr_encoding
 from awscli.customizations import utils as cli_utils
 from awscli.customizations.commands import BasicCommand
 from awscli.customizations.utils import uni_print
@@ -32,6 +32,17 @@ def get_relative_expiration_time(remaining):
 
     message = " ".join(values)
     return message
+
+
+class CommandFailedError(Exception):
+    def __init__(self, called_process_error):
+        msg = str(called_process_error)
+        if called_process_error.stderr is not None:
+            msg +=(
+                f' Stderr from command:\n'
+                f'{called_process_error.stderr.decode(get_stderr_encoding())}'
+            )
+        Exception.__init__(self, msg)
 
 
 class BaseLogin(object):
@@ -84,14 +95,14 @@ class BaseLogin(object):
 
     def _run_command(self, tool, command, *, ignore_errors=False):
         try:
-            self.subprocess_utils.check_call(
+            self.subprocess_utils.run(
                 command,
-                stdout=self.subprocess_utils.PIPE,
-                stderr=self.subprocess_utils.PIPE
+                capture_output=True,
+                check=True
             )
         except subprocess.CalledProcessError as ex:
             if not ignore_errors:
-                raise ex
+                raise CommandFailedError(ex)
         except OSError as ex:
             if ex.errno == errno.ENOENT:
                 raise ValueError(
@@ -153,13 +164,14 @@ class NuGetBaseLogin(BaseLogin):
             return
 
         try:
-            self.subprocess_utils.check_output(
+            self.subprocess_utils.run(
                 command,
-                stderr=self.subprocess_utils.PIPE
+                capture_output=True,
+                check=True
             )
         except subprocess.CalledProcessError as e:
             uni_print('Failed to update the NuGet.Config\n')
-            raise e
+            raise CommandFailedError(e)
 
         uni_print(source_configured_message % source_name)
         self._write_success_message('nuget')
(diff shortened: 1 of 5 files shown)


awscli/customizations/codeartifact/login.py
```diff
@@ -84,13 +84,16 @@ def _run_commands(self, tool, commands, dry_run=False):
...
                 command,
-                stdout=self.subprocess_utils.PIPE,
-                stderr=self.subprocess_utils.PIPE
+                capture_output=True,
+                check=True
             )
         except subprocess.CalledProcessError as ex:
             if not ignore_errors:
+                sys.stdout.write(
+                    ex.stderr.decode(os.device_encoding(1) or "utf-8")
```
kyleknap, line 95: For determining encoding of stderr, I think it would be worth exposing a `get_stderr_encoding` in the `compat.py` module similar to what we have here: https://github.com/aws/aws-cli/blob/9abc05dc392d5466b0d728006b177db10f4fe4cb/awscli/testutils.py#L654. The advantage in going with this approach is that in accessing `sys.__stderr__` we definitively know what standard error is going to be from the start of the Python program.

awscli/customizations/codeartifact/login.py
```diff
@@ -84,13 +84,16 @@ def _run_commands(self, tool, commands, dry_run=False):
...
+            self.subprocess_utils.run(
                 command,
-                stdout=self.subprocess_utils.PIPE,
-                stderr=self.subprocess_utils.PIPE
+                capture_output=True,
+                check=True
             )
         except subprocess.CalledProcessError as ex:
             if not ignore_errors:
+                sys.stdout.write(
```
kyleknap, line 94: Instead of writing out before raising the called process error, it would be interesting if we can explore just exposing our own exception class that we can raise and control the message to include stderr for. We can the exception something like `CommandFailedError` that accepts the instance of a `CalledProcessError` and we can then format the error message to include the normal string repr for `CalledProcessError` and then whatever was displayed from `stderr` to make the message like:
```
Command '['some-command', 'login']' returned non-zero exit status 1. Stderr from command: <stderr content>
```
I like this approach more as:
1. It give a little more context on what command error output represents
2. It guarantees we are writing this error message to stderr by including it in the raised exception. In general, we only want to be including error messages in stderr and not stdout as that is a general expectation from the user.

tests/unit/customizations/codeartifact/test_adapter_login.py
```diff
@@ -52,7 +52,7 @@ def test_get_commands(self):
             )
 
     def test_run_commands_nonexistent_command(self):
```
kyleknap, line 54: We should make sure to add a test case that output from stderr of the subprocess shows up in the stderr propagated to the user.

tests/functional/codeartifact/test_codeartifact_login.py
```diff
@@ -40,7 +40,7 @@ def setUp(self):
         self.pypi_rc_path_mock = self.pypi_rc_path_patch.start()
         self.pypi_rc_path_mock.return_value = self.test_pypi_rc_path
 
-        self.subprocess_patch = mock.patch('subprocess.check_call')
+        self.subprocess_patch = mock.patch('subprocess.run')
         self.subprocess_mock = self.subprocess_patch.start()
         self.subprocess_check_output_patch = mock.patch(
```
kyleknap, line 45: We should be able to remove this patch I'm assuming with our refactoring?

awscli/customizations/codeartifact/login.py
```diff
@@ -84,13 +84,16 @@ def _run_commands(self, tool, commands, dry_run=False):
```
kyleknap, line 84: Let's also make sure to add a changelog entry for this change so it ends up in our generated changelog. We should be able to run the `new-change` script from the root of the repository:
```
./scripts/new-change -t enhancement -c '``codeartifact login``' -d 'Include stderr output from underlying login tool when subprocess fails'
```
Then we should be able to add the json file found in the `.changes/next-release` directory.

kyleknap (commented):
Looks good. I just had some smaller comments.

tests/functional/codeartifact/test_codeartifact_login.py
dafangc-amazon: No, nuget and dotnet still use `check_output` to capture responses from list nuget source. I didn't want to change that because it is specific to nuget and dotnet and seems not in the scope of this pr.

tests/unit/customizations/codeartifact/test_adapter_login.py
dafangc-amazon: Added a test

awscli/customizations/codeartifact/login.py
dafangc-amazon: I created a class that inherits `CalledProcessError`. For the most part they are the same, other than when returncode > 1 (exception raised) it will add the info from stderr.

tests/functional/codeartifact/test_codeartifact_login.py
kyleknap: Make sense. Let's keep it as is then.

kyleknap (commented):
Looks good. Just had a few more comments.

awscli/customizations/codeartifact/login.py
```diff
@@ -393,14 +409,14 @@ class TwineLogin(BaseLogin):
...
     def __init__(
-        self,
-        auth_token,
-        expiration,
-        repository_endpoint,
-        domain,
-        repository,
-        subprocess_utils,
-        pypi_rc_path=None
+            self,
```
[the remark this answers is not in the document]
dafangc-amazon: Sure, accidentally checked in these changes but I'll definitely undo them.

kyleknap (approved):
Looks good! :ship:

Issue #6630, opened 2021-12-28
Allow omitting timestamps


RobertoUa:
Add a new format, just like `--format short` but without timestamps.
My logs already have timestamps, and on a small screen it takes up precious space. Moreover, timestamps are added even to multiline messages, which makes it hard for copying data from `aws logs tail`

kdaily:
Hi @RobertoUa,

Thanks for the post. Just to clarify, you're asking for omitting timestamps specifically from the `aws logs tail` command? 

This command is a convenience wrapper around the `aws logs filter-log-events` command, so any multiline messages are coming back from CloudWatch as is. You wouldn't be able to use the `tail` command to filter on your own timestamps, and there's no expectation that other users would be providing their own timestamps, so that might be confusing.

Can you provide some more details about how this would work and what your use case is? Thanks!

RobertoUa:
>Just to clarify, you're asking for omitting timestamps specifically from the `aws logs tail` command

Yeah, I'm asking for adding a parameter to `aws logs tail` for omitting timestamps. Probably `--format none`.
My current workaround is piping to a `cut` command, like this `aws logs tail <log> --follow --format short | cut -d\  -f2-`

Compare logs
with `cut`
```
2021-12-28T11:32:23,38 [io-compute-1] ERROR software.amazon.awssdk.http.crt.internal.AwsCrtAsyncHttpStreamAdapter - Response Encountered an Error. 
Caused by: software.amazon.awssdk.crt.http.HttpException: The connection has closed or is closing.
  software.amazon.awssdk.http.crt.internal.AwsCrtAsyncHttpStreamAdapter.onResponseComplete(AwsCrtAsyncHttpStreamAdapter.java:116)
  software.amazon.awssdk.crt.http.HttpStreamResponseHandlerNativeAdapter.onResponseComplete(HttpStreamResponseHandlerNativeAdapter.java:33)
```

vs current `aws logs tail --format short`

```
2021-12-29T05:44:43 2021-12-29T05:44:43,714 [io-compute-0] ERROR software.amazon.awssdk.http.crt.internal.AwsCrtResponseBodyPublisher - Error processing Response Body - task: 43310b8fade840d8843a7c4129ede64b, instance: 1d392c06193f44eea31ffe7996423499, ec2: i-0ed77c9972cf974eb
2021-12-29T05:44:43 Caused by: software.amazon.awssdk.crt.http.HttpException: The connection has closed or is closing.
[...]

PR #8185, opened 2023-09-20
Adding functionality in aws logs tail  to log message as a single json without additional formatting

albinpaul:
*Issue #, if available:*
Fixes https://github.com/aws/aws-cli/issues/6630

*Description of changes:*
Added a new formatter that logs entire log in a single line. So that it can be used with other formatters like https://github.com/pinojs/pino-pretty.


tim-finnigan:
Hi all, thanks for reaching out and for your patience here. After bringing this up with discussion with the team, the consensus was to close this as not planned. The reason being that you should already be able to use command line tools to process the output as needed and remove the timestamps, for example `aws logs tail /aws/lambda/testJs --format short | cut -c 21-`.

PR #8221, opened 2023-10-04
Use builtin mock

JordonPhillips:
This removes the dependency on the mock package since it's built in to the python versions we support.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -40,7 +40,7 @@ from awscli.compat import StringIO
 from ruamel.yaml import YAML
 
 try:
-    import mock
+    from unittest import mock
 except ImportError as e:
     # In the off chance something imports this module
     # that's not suppose to, we should not stop the CLI
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,5 +1,4 @@
 jsonschema==4.7.2
-mock==1.3.0
 pytest==7.2.0
 coverage==7.0.1
 pytest-cov==4.0.0
--- a/tests/functional/autoprompt/test_autoprompt.py
+++ b/tests/functional/autoprompt/test_autoprompt.py
@@ -1,4 +1,4 @@
-import mock
+from unittest import mock
 import os
 
 import pytest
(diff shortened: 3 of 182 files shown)


nateprewitt:
I have no idea if this (ff5961571f534c2e5c5417c3d628d6597d49317c) still ports forward (unlikely after 2 years) but we may want to do something closer to CLI v1 to keep the codebases as close as we can.

yan12125:
> I have no idea if this ([ff59615](https://github.com/aws/aws-cli/commit/ff5961571f534c2e5c5417c3d628d6597d49317c)) still ports forward (unlikely after 2 years) but we may want to do something closer to CLI v1 to keep the codebases as close as we can.

How about pushing similar changes for both v1 and v2 branches? `unittest.mock` is available since Python 3.3 [1], and aws-cli v1 already requires Python 3.7 [2].

[1] https://docs.python.org/3/library/unittest.mock.html
[2] https://github.com/aws/aws-cli/blob/develop/setup.py

nateprewitt:
Thanks for the suggestion, @yan12125. That's what we currently do in CLI v1. I think the proposal was to move that functionality into the CLI v2 instead of doing a different implementation. I have a port of the above commit here that we may be able to use (f255378269aa862e7f2c91722605cd159e641bc6). Botocore will need it's changes too though.

yan12125:
Thanks for the explanation and sorry for misunderstanding the intention. Your proposal is definitely a good move.

For botocore, porting https://github.com/boto/botocore/commit/f22b2cddcc894fd6db1654aaca93861448f6c33b may be enough.

yan12125:
> For botocore, porting https://github.com/boto/botocore/commit/f22b2cddcc894fd6db1654aaca93861448f6c33b may be enough.

I gave it a try, but I didn't go far as upstream botocore and bundled botocore v2 are quite different. As bundled botocore already deviates from upstream, how about doing a simple replacement for bundled botocore like this pull request, and using https://github.com/aws/aws-cli/commit/f255378269aa862e7f2c91722605cd159e641bc6 for the remaining?

By the way, this pull request can no longer be merged as-is due to conflicts.

PR #8291, opened 2023-11-02
[v2] Add support for stdin/stdout streams for CRT client

kyleknap:
This PR pulls in changes from this s3transfer PR: https://github.com/boto/s3transfer/pull/277. However, it only pulls in a subset of changes proposed specific to streaming support. The other proposed changes from the s3transfer PR will be considered/pulled in via subsequent PRs. This PR currently targets a feature branch that will consist of all CRT transfer client improvements that we are looking to make, and this branch will eventually be merged into the `v2` branch. Once merged, the improvements to the `s3transfer` code will be backported to the upstream `s3transfer` project.

In supporting steams at the `s3transfer` CRT layer, I also did the following refactorings:

* Refactoring the s3transfer crt layer to better organize the branching introduced in supporting both file objects and file name.
* Introduce additional helper test methods to reduce verbosity of s3transfer crt test and improve reusability. Even with the updates, there was a fair amount of duplication. In the long term, I'd prefer we switch these over to pytest parameterization, but I suspect that would be larger than the intended scope of this PR.

### DIFF
--- a/awscli/s3transfer/crt.py
+++ b/awscli/s3transfer/crt.py
@@ -428,19 +428,12 @@ class BotocoreCRTRequestSerializer(BaseCRTRequestSerializer):
                 headers_list.append((name, str(value, 'utf-8')))
 
         crt_headers = awscrt.http.HttpHeaders(headers_list)
-        # CRT requires body (if it exists) to be an I/O stream.
-        crt_body_stream = None
-        if aws_request.body:
-            if hasattr(aws_request.body, 'seek'):
-                crt_body_stream = aws_request.body
-            else:
-                crt_body_stream = BytesIO(aws_request.body)
 
         crt_request = awscrt.http.HttpRequest(
             method=aws_request.method,
             path=crt_path,
             headers=crt_headers,
-            body_stream=crt_body_stream,
+            body_stream=aws_request.body,
         )
         return crt_request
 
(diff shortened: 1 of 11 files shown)


awscli/s3transfer/crt.py
```diff
@@ -555,39 +567,19 @@ def __init__(self, crt_request_serializer, os_utils):
...
+            self,
+            f'_get_make_request_args_{request_type}',
+            self._default_get_make_request_args,
+        )(
+            request_type=request_type,
+            call_args=call_args,
+            coordinator=coordinator,
+            future=future,
+            on_done_before_calls=[],
+            on_done_after_calls=on_done_after_calls,
```
nateprewitt, line 580: nit; this seems auto-magic enough to be difficult to debug if we have a failure in getattr or the function call itself since they'll all be on the same line. It may be worth splitting this into a variable for the function and then its invocation.

nateprewitt (approved):
One minor comment but otherwise I think this looks good. Feel free to merge if you disagree.

awscli/s3transfer/crt.py
```diff
@@ -613,6 +605,96 @@ def invoke_all_callbacks(*args, **kwargs):
...
+            send_filepath = call_args.fileobj
+            data_len = self._os_utils.get_file_size(send_filepath)
+            call_args.extra_args["ContentLength"] = data_len
+        else:
+            call_args.extra_args["Body"] = call_args.fileobj
+
+        # Suppress botocore's automatic MD5 calculation by setting an override
+        # value that will get deleted in the BotocoreCRTRequestSerializer.
+        # The CRT S3 client will automatically compute checksums as part of
+        # requests it makes.
```
graebm, line 628: CRT S3 Client doesn't compute checksums unless you pass an [S3ChecksumConfig](https://github.com/awslabs/aws-crt-python/blob/fbf82f23cecb754beb6df7c1a30f11700774f2b9/awscrt/s3.py#L94). See me setting the checksum algorithm and location [here](https://github.com/boto/s3transfer/pull/277/files#diff-d7d493766dac8cf549533003bff769f7bc5d766645facdbe6c53d188284c953aR635-R640)

So this (and the currently released versions) aren't doing any checksums or MD5 for uploads

graebm (commented):
LGTM

awscli/s3transfer/crt.py
kyleknap: Gotcha. Yeah one of the follow ups is to automatically set checksums for these requests as you had it in your original PR. For now, I'll just delete/update the second sentence in that comment.

awscli/s3transfer/crt.py
kyleknap: That's reasonable. I'll break it up.

graebm (approved):

PR #8298, opened 2023-11-03
[v2] Turn on checksum validation for CRT S3 client

kyleknap:
For uploads, the CRT S3 client will add CRC32 trailing checksums. For downloads, the CRT S3 client will validate checksums associated to the object when possible. This is also inline with the proposed changes in this s3transfer PR: https://github.com/boto/s3transfer/pull/277. The plan is backport this to s3transfer once merged.

### DIFF
--- a/awscli/s3transfer/crt.py
+++ b/awscli/s3transfer/crt.py
@@ -206,6 +206,7 @@ class CRTTransferManager:
             extra_args = {}
         if subscribers is None:
             subscribers = {}
+        self._validate_checksum_algorithm_supported(extra_args)
         callargs = CallArgs(
             bucket=bucket,
             key=key,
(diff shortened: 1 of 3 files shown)


awscli/s3transfer/crt.py
```diff
@@ -206,6 +206,7 @@ def upload(self, fileobj, bucket, key, extra_args=None, subscribers=None):
             extra_args = {}
         if subscribers is None:
             subscribers = {}
+        self._validate_checksum_algorithm_supported(extra_args)
```
graebm, line 209: somewhat offtopic: Noticing that CRTTransferManager doesn't validate input as extensively as the classic TransferManager:
https://github.com/aws/aws-cli/blob/d61d07799c5c4d2d40d5e26c6d6efed474f98501/awscli/s3transfer/manager.py#L318-L319

graebm (approved):

awscli/s3transfer/crt.py
kyleknap: Yeah I think we should follow up on validating extra args in another PR.

awscli/s3transfer/crt.py
```diff
@@ -231,6 +232,17 @@ def delete(self, bucket, key, extra_args=None, subscribers=None):
     def shutdown(self, cancel=False):
         self._shutdown(cancel)
 
+    def _validate_checksum_algorithm_supported(self, extra_args):
+        checksum_algorithm = extra_args.get('ChecksumAlgorithm')
+        if checksum_algorithm is None:
+            return
+        if checksum_algorithm not in awscrt.s3.S3ChecksumAlgorithm.__members__:
```
nateprewitt, line 239: Do we guarantee casing will be correct here by calling `upper()` somewhere else? Otherwise, we'd be able to end up with something like `Crc32` right?

nateprewitt: Nevermind, this isn't configurable by the end user currently so we're only using the constants which should match.

kyleknap: I pushed up a commit to make the checksum algorithm case insensitive. In the normal transfer manager, it accepts lowercase values. So it makes sense to accept it here because we will be porting it back to `s3transfer`.

nateprewitt (approved):

PR #8304, opened 2023-11-06
Extend docutils line length limit

### LINKED ISSUE #8269: Error message for help on ec2 create-launch-template-version
### Describe the bug

Running the command `aws ec2 create-launch-template-version help` generates an error message with `line-length-limit`.

### Expected Behavior

To display the help document for ec2 create-launch-template-version

### Current Behavior

The output of running `aws ec2 create-launch-template-version help`:

```
<string>:: (ERROR/3) Line 2533 exceeds the line-length-limit.
()                                                                          ()

NAME
        -

       System Message: ERROR/3 (<string>:)
              Line 2533 exceeds the line-length-limit.

                                                                            ()
```

### Reproduction Steps

`aws ec2 create-launch-template-version help`

### Possible Solution

_No response_

`aws --version` output: `aws-cli/2.13.29 Python/3.11.6 Darwin/22.6.0 source/arm64 prompt/off`

Also received the same issue running the command while on CloudShell:
```
aws ec2 create-launch-template-version help
<string>:: (ERROR/3) Line 2533 exceeds the line-length-limit.
```

with version `aws-cli/2.13.27 Python/3.11.6 Linux/4.14.255-322-265.538.amzn2.x86_64 exec-env/CloudShell exe/x86_64.amzn.2 prompt/off`

### CLI version used

2.13.29

macos 13.5.2

kyleknap:
**Note: This PR targets the `develop` branch even though the related issue currently does not affect v1. This is because once v1 does upgrade `docutils`, it will be impacted. In addition, by merging this into `develop` first, it will become merged into v2. I also confirmed that the changes in this PR are compatible (i.e. ``help`` commands still work) with the ranges of `docutils` supported for v1.**

Addresses: https://github.com/aws/aws-cli/issues/8269

This is a temporary fix to handle JSON snippets in the man pages that are longer than the default docutils line limit of 10,000 and cause the man pages for some EC2 examples to not render.

Long term, we should consider breaking up the enums to multiple lines or potentially not enumerate all enums in the JSON snippets.

The value 50,000 was chosen based on the line length for ``InstanceType`` enums for the ``ec2 create-launch-template-version`` command. Currently, the line length for the ``InstanceType`` enums is just over 10,000. Assuming each enum is 10 characters long, a line length of 50,000 would give us the runway for roughly 4000 more instance types. The line length was not made magnitudes higher (i.e. to avoid hitting this issue ever again) because lines of this length are already not ideal and we should make sure to implement a more long term solution.

### DIFF
--- a/.changes/next-release/bugfix-help-587.json
+++ b/.changes/next-release/bugfix-help-587.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``help``",
+  "description": "Relax line length limit for rendered ``help`` pages"
+}
--- a/awscli/help.py
+++ b/awscli/help.py
@@ -66,6 +66,16 @@ class PagingHelpRenderer(object):
         self.output_stream = output_stream
 
     PAGER = None
+    _DEFAULT_DOCUTILS_SETTINGS_OVERRIDES = {
+        # The default for line length limit in docutils is 10,000. However,
+        # currently in the documentation, it inlines all possible enums in
+        # the JSON syntax which exceeds this limit for some EC2 commands
+        # and prevents the manpages from being generated.
+        # This is a temporary fix to allow the manpages for these commands
+        # to be rendered. Long term, we should avoid enumerating over all
+        # enums inline for the JSON syntax snippets.
+        'line_length_limit': 50_000
+    }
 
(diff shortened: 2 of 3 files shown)


nateprewitt (approved):
:shipit:

PR #8315, opened 2023-11-09
feat: set None string as None value during aws configure

Thibault-Brocheton:
During `aws credentials` command, the CLI ask for the output type
image

Some users of my company, instead of typing enter, type the "None" string that is "advised" by the command
image

Their lack of knowledge about how this kind of advice is handled is the cause, however, I think we could transform the "None" string to None value

I might be missing some edge cases, if my change doesn't make sense, just close the PR

*Description of changes:*

During `aws configure`, we consider 'None' string values to be None values


tim-finnigan:
Thanks for creating this PR and your patience here. If still interested in this, please create a feature request outlining why you think this is needed, and we can track the discussion there to get input from others. The content in brackets follows the convention of denoting the default value, so I don't think this change would be considered.

PR #8327, opened 2023-11-14
[v2] Update CRT throughput target defaults

kyleknap:
Instead of defaulting to five gigabits per second when no target througput is configured, the AWS CLI will use the CRT to determine if there is a recommended target throughput to use. If it is unable, to determine a recommended target throughput, it will default to ten gigabits per second, which is the current throughput default for the CRT S3 client.

This behavior was updated to help minimize the amount of configuration required to get the fastest throughputs from the CRT transfer client. The fallback throughput was updated from five to ten gigabits so that the CRT S3 integration is consistent with the underlying CRT S3 client's defaults.

This also fixes an issue where it was documented that a user could provide integers (e.g. `1024`) and human readable rates that lack a magnitude prefix (e.g. `1024B/s`) for S3 rate-related configurations. However, providing these sorts of values resulted in value errors. This adds support for these formats so the behavior matches the documentation and also makes it easier to specify exact rates especially when you are expecting powers of 10 (and not 2) when using magnitude suffixes (`MB`, `GB`, etc.)

### DIFF
--- a/awscli/customizations/s3/transferconfig.py
+++ b/awscli/customizations/s3/transferconfig.py
@@ -30,7 +30,7 @@ DEFAULTS = {
     'max_queue_size': 1000,
     'max_bandwidth': None,
     'preferred_transfer_client': constants.AUTO_RESOLVE_TRANSFER_CLIENT,
-    'target_bandwidth': int(5 * (1024 ** 3) / 8),  # which is 5 Gb/s
+    'target_bandwidth': None,
 }
 
 
--- a/awscli/customizations/s3/transferconfig.py
+++ b/awscli/customizations/s3/transferconfig.py
@@ -10,6 +10,8 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import re
+
 from s3transfer.manager import TransferConfig
 
 from awscli.customizations.s3 import constants
@@ -57,6 +59,7 @@ class RuntimeConfig(object):
             'default': constants.CLASSIC_TRANSFER_CLIENT
         }
     }
+    _INTEGER_STR_REGEX = re.compile(r'^[0-9]+$')
 
     @staticmethod
     def defaults():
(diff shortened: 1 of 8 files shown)


awscli/customizations/s3/transferconfig.py
```diff
@@ -93,18 +96,44 @@ def _convert_human_readable_rates(self, runtime_config):
...
-                    runtime_config[attr] = human_readable_to_int(value[:-2])
+                    runtime_config[attr] = self._human_readable_rate_to_int(
+                        value
+                    )
                 elif value.endswith('b/s'):
-                    bits_per_sec = human_readable_to_int(value[:-2])
+                    bits_per_sec = self._human_readable_rate_to_int(value)
                     bytes_per_sec = int(bits_per_sec / 8)
                     runtime_config[attr] = bytes_per_sec
+                elif self._INTEGER_STR_REGEX.match(value):
```
nateprewitt, line 106: Just curious why you went with a regex as opposed to a try/except. If a user passes in the value as `10.0` are we considering that as unsupported? Also it looks like the message only discusses rates. Do we want to update that to reflect this behavior?

awscli/s3transfer/crt.py
```diff
@@ -114,7 +119,12 @@ def create_s3_crt_client(
 
     :type target_throughput: Optional[int]
     :param target_throughput: Throughput target in Bytes.
-        Default is 0.625 GB/s (which translates to 5 Gb/s).
+        By default, CRT will automatically attempt to choose a target
+        throughput that matches the system's maximum network throughput.
+        Currently, if CRT is unable to determine the maximum network
+        throughput, a fallback target throughput of 1.25 GB/s
+        (which translates to 10Gb/s) is used. To guarantee a specific target
+        throughput, set a value for this parameter.
```
nateprewitt, line 127: > To guarantee a specific target throughput, set a value for this parameter.

semantic nit; should we use a word like "modify" or "set" instead of "guarantee"? I don't think the CRT can ever guarantee it hits the target.

awscli/customizations/s3/transferconfig.py
kyleknap: I went with the regex to avoid being entirely coupled to Python's `int()` interface. By restricting the values to be strings of digits 0-9, it ensures that we don't necessarily support all strings that Python supports for `int()` and we have more control over what we want to support. For example, I think you can specify underscores in string integer values:
```python
>>> int('1_000_000')
1000000
```
In the end, I don't have a strong preference about this. I could go either way. Thoughts?

> If a user passes in the value as 10.0 are we considering that as unsupported? 

Yeah that would not be supported. Right now all rates are in terms of bytes/sec and I don't think there is much value in supporting a fraction of a byte for these configurations.

> Also it looks like the message only discusses rates. Do we want to update that to reflect this behavior?

Good catch. I can update that.

awscli/s3transfer/crt.py
kyleknap: I can change that. I can see how that is misleading.

awscli/customizations/s3/transferconfig.py
kyleknap: Following up on this. I guess we already sort of bounded to what Python accepts for `int()`. In the currently released logic, you can set the configuration rate value with underscores and it will properly interpret values like this: `8_000Mb/s`. I'm just going to update the logic to try to do an `int()` conversion instead of the regex to keep it all consistent. I don't think the current set of values that Python supports is an issue.

awscli/s3transfer/crt.py
```diff
@@ -176,6 +187,23 @@ def create_s3_crt_client(
...
+        target_gbps = get_recommended_throughput_target_gbps()
+        logger.debug(
+            'Recommended CRT throughput target in gbps: %s', target_gbps
+        )
+        if target_gbps is None:
+            target_gbps = 10.0
+    else:
+        # NOTE: The GB constant in s3transfer is technically a gibibyte. The
+        # GB constant is not used here because the CRT interprets a gigabyte
+        # as a base power of 10 (i.e. 1000 ** 3 instead of 1024 ** 3).
```
graebm, line 201: utterly trivial
```suggestion
        # NOTE: The GB constant in s3transfer is technically a gibibyte. The
        # GB constant is not used here because the CRT interprets gigabits-for-networking
        # as a base power of 10 (i.e. 1000 ** 3 instead of 1024 ** 3).
```

awscli/s3transfer/crt.py
```diff
@@ -114,7 +119,12 @@ def create_s3_crt_client(
 
     :type target_throughput: Optional[int]
     :param target_throughput: Throughput target in Bytes.
-        Default is 0.625 GB/s (which translates to 5 Gb/s).
+        By default, CRT will automatically attempt to choose a target
+        throughput that matches the system's maximum network throughput.
+        Currently, if CRT is unable to determine the maximum network
+        throughput, a fallback target throughput of 1.25 GB/s
+        (which translates to 10Gb/s) is used. To set a specific target
```
graebm, line 126: maybe spell things out? Since GB is already used meaning powers-of-2 in this codebase
```suggestion
        throughput, a fallback target throughput of 125000000 Bytes/s
        (which translates to 10 gigabits/s, or 1.16 gibibytes) is used. To set a specific target
```

graebm (commented):
Looks good. Just nitpicking comments

graebm (approved):

nateprewitt (approved):

PR #8335, opened 2023-11-16
[v2] Move CRT S3 configurations from experimental section

kyleknap:
In doing so, the CRT S3 transfer client and the ``preferred_transfer_client`` and ``target_bandwidth`` S3 configurations are now GA and stable.


nateprewitt (approved):
:shipit:

Issue #8342, opened 2023-11-20
V2 support for Python 3.12


ericbn:
### Describe the feature

aws-cli V2 [currently](https://github.com/aws/aws-cli/blob/66af26fd0de98269959b5d570ebfd91e8018faef/README.rst?plain=1#L14-L17) supports:

> * 3.8.x
> * 3.9.x
> * 3.10.x
> * 3.11.x

Add support for Python 3.12.x too.

### Use Case

I installed aws-cli V2 using Homebrew, and it's the last tool I have installed still using Python 3.11. But more importantly, I believe this helps keep V2 up-to-date.

### Proposed Solution

Add support for Python 3.12.x too.

### Other Information

Python 3.12 was first released on 2023-10-02 and has EOL scheduled for 2028-10. See https://devguide.python.org/versions/

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### CLI version used

2.13.37

### Environment details (OS name and version, etc.)

Darwin/21.6.0

tim-finnigan:
Thanks @ericbn - this was [recently added](https://github.com/aws/aws-cli/pull/8303/files) to the v1 README and looks like it just needs to be added to the v2 README. Created a PR to do that: https://github.com/aws/aws-cli/pull/8343. 

Also for installing v2 of the AWS CLI we recommend using one of the options documented here: https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html.

Andrewpk:
@tim-finnigan not one of the [options listed](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) mention a package manager for the installation, which is a [surprisingly](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-macos) [common](https://www.linode.com/docs/products/tools/cli/guides/install/) [thing to want](https://cloud.google.com/sdk/docs/downloads-snap), so it's easy for a fellow user to see where @ericbn is coming from.

ncopa:
Is there any time frame for adding python 3.12 support? Alpine Linux is upgrading python to 3.12.

onlined:
#8217 needs to be merged first. `ruamel.yaml.clib` versions before `0.2.8` don't support Python 3.12.

yan12125:
[...]

PR #8303, opened 2023-11-06
update readme to include py 3.12

Volatus:
Currently the CI jobs already do builds for Python 3.12 but the README
did not explcitly specify support for 3.12

Signed-off-by: Ismayil Mirzali <ismayilmirzeli@gmail.com>

*Issue #, if available:*

*Description of changes:*
Updates the README to explicitly state Python 3.12 support.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #8343, opened 2023-11-20
Add 3.12.x to README

tim-finnigan:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/8342

*Description of changes:* Add 3.12.x to README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,7 @@ The aws-cli package works on Python versions:
 -  3.9.x and greater
 -  3.10.x and greater
 -  3.11.x and greater
+-  3.12.x and greater
 
 Notices
 ~~~~~~~


ericbn:
@tim-finnigan isn’t this PR updating the V1 README instead of the [V2 README](https://github.com/aws/aws-cli/blob/v2/README.rst)? I also noticed that the [V2 tox.ini](https://github.com/aws/aws-cli/blob/v2/tox.ini) still does not include Python 3.12.

tim-finnigan:
@ericbn thanks for noting that - [this PR](https://github.com/aws/aws-cli/pull/8303/files) already updated the README for v1 noting Python 3.12 support. There is still some testing required before Python 3.12 is officially supported in v2, but the README will be updated when that is complete.

PR #8449, opened 2023-12-23
 Replace string concatenation with f-string

Sergius-Nyah:
The code "session.user_agent_extra = 'botocore/%s' % botocore_version" uses old-style string formatting. This could be replaced with an f-string for better readability and performance."


kdaily:
Thanks for the suggestion. However, this PR makes many more changes out of scope of f-string replacements, making it difficult to review and only pull in the related changes. Closing this PR.

PR #8450, opened 2023-12-23
Replace the strings in the code with these constants.

Sergius-Nyah:
The strings 'aws-cli', 'botocore', 'awscli', 's3transfer', 'urllib3' are used in the code. These could be replaced with constants. This Pull Request improves code readability and maintainability by avoiding the use of magic strings."

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -312,6 +312,17 @@ AVAILABLE_COMMANDS = 'Available Commands'
 class ProviderDocumentEventHandler(CLIDocumentEventHandler):
 
     def doc_breadcrumbs(self, help_command, event_name, **kwargs):
+        """
+        This method generates breadcrumbs for the documentation of a command.
+
+        Args:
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @Sergius-Nyah, we've noticed you created a large number of PRs in the last day. Generally when providing contributions for open source projects, it's courteous to limit contributions to a few PRs and wait for feedback unless you've previously discussed a larger effort with the maintainers.

For a number of the PRs, it seems there may be some confusion around concepts like magic strings and expected error handling.

For this one specifically, there don't appear to be any magic strings. The current patch would imply _all_ strings must be constants which isn't convention. A [magic string](https://en.m.wikipedia.org/wiki/Magic_string) implies there is some non-obvious or divergent behavior that occurs with a given input.

Unfortunately, I don't think we'd accept this as it's written. Thanks for your interest in contributing though. For future PRs, it's also important to start with a fresh branch for each contribution. Otherwise you end up with all your commits in each PR (this one has 24 commits from all your previous PRs).

PR #8471, opened 2024-01-09
Pin alabaster version

hssyoo:
`alabaster` is a transitive dependency we take via `Sphinx`. This commit pins the version of `alabaster` we use to ensure compatibility with our version of `Sphinx`.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,3 +1,6 @@
 Jinja2<3.1.0
 Sphinx==3.0.2
 sphinx-notfound-page==0.4
+# alabaster 0.7.14 dropped support for Sphinx<3.4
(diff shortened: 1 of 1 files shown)


requirements-docs.txt
```diff
@@ -1,3 +1,4 @@
 Jinja2<3.1.0
 Sphinx==3.0.2
 sphinx-notfound-page==0.4
+alabaster==0.7.13
```
nateprewitt, line 4: May be worth adding a comment noting this is because support for Sphinx<3.4 was dropped in alabaster 0.7.14, and Sphinx 3.0.2 has an unbounded dependency. Then we'll know when it can be cleaned up if we update Sphinx later.

PR #8482, opened 2024-01-17
Pin sphinx extension versions

kdaily:
*Issue #, if available:*

Docs will not build.

*Description of changes:*

Sphinx extensions removed dependency on Sphinx 5 in latest versions, but still require it. Since we are still pinning to Sphinx 3, we need to pin to previous versions of the extensions.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -4,3 +4,12 @@ sphinx-notfound-page==0.4
 # alabaster 0.7.14 dropped support for Sphinx<3.4
 # and Sphinx 3.0.2 bounds alabaster>=0.7,<0.8.
 alabaster==0.7.13
+# Latest versions of these packages support only
+# Sphinx>5, which we are not using. Set explicit pins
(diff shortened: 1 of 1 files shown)


requirements-docs.txt
```diff
@@ -3,4 +3,13 @@ Sphinx==3.0.2
 sphinx-notfound-page==0.4
 # alabaster 0.7.14 dropped support for Sphinx<3.4
 # and Sphinx 3.0.2 bounds alabaster>=0.7,<0.8.
-alabaster==0.7.13
\ No newline at end of file
+alabaster==0.7.13
+# These packages removed dependency on Sphinx>5,
```
kyleknap, line 7: Nit: Update the comment to be: `Latest versions of these packages support only Sphix>5`. The packages removed the explicit dependency in the `pyproject.toml` but they still require it at runtime of the extension.

kyleknap (approved):
:ship:

PR #8511, opened 2024-01-29
Introduce dependency test suite for monitoring dependency closure

kyleknap:
The plan is to build out this test suite to test against the AWS CLI's dependencies to help facilitate dependency upgrades.

To start, this test suite contains the following new test cases to better monitor the overall dependency closure of the awscli package:

* Assert expected packages in runtime closure. This will alert us if a dependency introduces a new transitive depenency to the AWS CLI closure.

* Assert expected unbounded dependencies in runtime closure. Specifically these are dependencies that do not have a version ceiling. This will alert us if a new unbounded dependency is introduced into the AWS CLI runtime dependency closure.

* Assert expected packages used in building the awscli package and any of its runtime dependencies. Specifically, this targets the use case of a user installing the awscli (and all of its dependencies) from sdist as it will capture which build packages are pulled into the isolated virtual environment when building the package.

* Assert expected unbounded dependencies in building the awscli package and any of its runtime dependencies. This helps monitor out of all of the build packages used, which ones may have no version ceiling when installed into an isolated virutal environment

See additional implementation notes below:

* These tests were broken into a separate test suite (i.e. instead of adding them to the unit and functional test suite) to allow more granularity when running them. Specifically, it is useful for:

  1. Avoiding the main unit and functional CI test suite from failing if a dependency changes from underneath of us (e.g. a new build dependency is added that we cannot control).

  2. For individuals that package the awscli, they generally will not want to run this test suite as it is fairly specific to how pip installs dependencies.

  3. The tests currently run between 3-5 minutes, which would significantly slow down the unit/functional test runs. Having it in a seperate directory allows you to only run the dependency test suite if you are making a change in dependencies.

* To determine the runtime dependency closure, the `Package` and `DependencyClosure` utilities traverse the dist-info `METADATA` files of the packages installed in the current site packages to build the runtime graph. This approach was chosen because:

  1. Since pip already installed the package, this logic avoids having to reconstruct the logic of how pip decides to resolve dependencies to figure out how to traverse the runtime graph. Any custom logic may deviate from how pip behaves which is what most users will be using to install the `awscli` as a Python package
  2. It's faster. The runtime closure test cases do not require downloading or installing any additional packages.

* To determine the build dependency closure of each package in the `awscli` runtime closure, the Package and `DependencyClosure` utilities leverage the `build` package to invoke PEP517 hooks to collect which packages are needed to build both the sdist and wheel. Relying on PEP517 hooks is really the only standard that can be relied on to derive the build dependencies as they can be dynamically generated as part of invoking a build backend. The `build` package was chosen to invoke the PEP517 hooks because:

  1. It has helpful higher level utilities to capture the results from the hooks if provided a virtualenv and project path. This would have been a significant amount of scaffolding to produce our own PEP517 hook invoker utilities.
  2. It has compatibility logic on how to handle projects that do not include a `pyproject.toml` and automatically can invoke the correct build backend (e.g. falls back to the legacy setuptools backend when needed) when a build backend is not declared.

### DIFF
--- a/.github/workflows/run-dep-tests.yml
+++ b/.github/workflows/run-dep-tests.yml
@@ -1,0 +1,27 @@
+name: Run dependency tests
+
+on:
+  push:
+  pull_request:
+    branches-ignore: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: python scripts/ci/install
+    - name: Run tests
+      run: python scripts/ci/run-dep-tests
--- a/requirements-dev-lock.txt
+++ b/requirements-dev-lock.txt
@@ -7,6 +7,10 @@
 atomicwrites==1.4.1 \
     --hash=sha256:81b2c9071a49367a7f770170e5eec8cb66567cfbbc8c73d20ce5ca4a8d71cf11
     # via -r requirements-dev.txt
+build==1.0.3 \
+    --hash=sha256:538aab1b64f9828977f84bc63ae570b060a8ed1be419e7870b8b4fc5e6ea553b \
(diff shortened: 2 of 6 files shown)


kyleknap:
Closing this PR in favor of this one: https://github.com/aws/aws-cli/pull/8512.

After talking offline, we decided that adding tests for the build dependency closure is too complex and too unstable for the value it would provide. Specifically, the current test implementation proves that the project already has a large build dependency and most of the dependencies are unbounded. This likely means it will not be uncommon for these tests to fail as there are little version bounds preventing a build dependency in the closure from changing. Furthermore, if a dependency does change their build dependencies, there is not much the `awscli` project could do to handle it (other than update the test) as `pip` by default will install build dependencies in a separate virtualenvs that the `awscli` project itself has no control over. If in the future, there are improvements made to Python/`pip` to allow more granular control over the build dependencies of a dependency, these sorts of tests may be worthwhile to explore again.

PR #8512, opened 2024-01-30
Introduce dependency test suite for runtime closure

kyleknap:
The plan is to build out this test suite to test against the AWS CLI's dependencies to help facilitate dependency upgrades.

To start, this test suite contains the following new test cases to better monitor the overall dependency closure of the awscli package:

* Assert expected packages in runtime closure. This will alert us if a dependency introduces a new transitive depenency to the AWS CLI closure.

* Assert expected unbounded dependencies in runtime closure. Specifically these are dependencies that do not have a version ceiling. This will alert us if a new unbounded dependency is introduced into the AWS CLI runtime dependency closure.

See additional implementation notes below:

* These tests were broken into a separate test suite (i.e. instead of adding them to the unit and functional test suite) to allow more granularity when running them. Specifically, it is useful for:

  1. Avoiding the main unit and functional CI test suite from failing if a dependency changes from underneath of us (e.g. a new build dependency is added that we cannot control).

  2. For individuals that package the `awscli`, they generally will not want to run this test suite as it is fairly specific to how pip installs dependencies.

* To determine the runtime dependency closure, the `Package` and `DependencyClosure` utilities traverse the dist-info `METADATA` files of the packages installed in the current site packages to build the runtime graph. This approach was chosen because:

  1. Since pip already installed the package, this logic avoids having to reconstruct the logic of how pip decides to resolve dependencies to figure out how to traverse the runtime graph. Any custom logic may deviate from how pip behaves which is what most users will be using to install the `awscli` as a Python package
  2. It's faster. The runtime closure test cases do not require downloading or installing any additional packages.

### DIFF
--- a/.github/workflows/run-dep-tests.yml
+++ b/.github/workflows/run-dep-tests.yml
@@ -1,0 +1,27 @@
+name: Run dependency tests
+
+on:
+  push:
+  pull_request:
+    branches-ignore: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+
+    steps:
+    - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: python scripts/ci/install
+    - name: Run tests
+      run: python scripts/ci/run-dep-tests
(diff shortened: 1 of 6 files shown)


.github/workflows/run-dep-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+
+    steps:
+    - uses: actions/checkout@v4
```
nateprewitt, line 19: nit; can we pin these to specific hashes so we're not flagging on OSSF. Here's an example: https://github.com/boto/boto3/blob/develop/.github/workflows/run-crt-test.yml#L21

nateprewitt (approved):
Minor nit but otherwise looks good.

.github/workflows/run-dep-tests.yml
kyleknap: I can update that. I'll do it before merging.

.github/workflows/run-dep-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+
+    steps:
+    - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@a5c61591373683505ea898e09a3ea4f39ef2b9c
```
nateprewitt, line 21: I think this was missing a leading `0` for the commit ID.
```suggestion
      uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c
```

PR #8519, opened 2024-02-01
Update QueryOutFileArgument permission logic

kyleknap:
This allows the permission mode to be set prior to writing any content for new created files.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -20,6 +20,7 @@ import zipfile
 import signal
 import contextlib
 from configparser import RawConfigParser
+from functools import partial
 
 from botocore.compat import six
 #import botocore.compat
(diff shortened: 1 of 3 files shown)


awscli/compat.py
```diff
@@ -196,19 +182,32 @@ def _get_text_writer(stream, errors):
...
             stdout = sys.stdout
 
         stdout.write(statement)
 
 
+def compat_open(filename, mode='r', encoding=None, access_permissions=None):
+    """Back-port open() that accepts an encoding argument.
+
+    In python3 this uses the built in open() and in python2 this
+    uses the io.open() function.
```
nateprewitt, line 196: Is this part still relevant given we don't support Python 2 anymore?

kyleknap: No it's not. I removed it because we retained it in v2 and I wanted to minimize difference between the implementations: https://github.com/kyleknap/aws-cli/commit/692314a97d0f415e0f846ab6546b0eb6247d5394

kyleknap: I don't have strong opinions though on whether to keep it.

nateprewitt: Yeah, I don't either. Just noted it as looking odd. We can do cleanup later.

nateprewitt (approved):

PR #8563, opened 2024-03-06
add UnhealthyNodeReplacement to CLI

jonathan343:
add UnhealthyNodeReplacement to CLI

### DIFF
--- a/tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
+++ b/tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
@@ -1,0 +1,50 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from tests.unit.customizations.emr import EMRBaseAWSCommandParamsTest as \
+    BaseAWSCommandParamsTest
+from awscli.testutils import BaseAWSHelpOutputTest
+
+class TestSetUnhealthyNodeReplacement(BaseAWSCommandParamsTest):
+    prefix = 'emr set-unhealthy-node-replacement'
+
+    def assert_error_message_has_field_name(self, error_msg, field_name):
+        self.assertIn('Missing required parameter', error_msg)
+        self.assertIn(field_name, error_msg)
+
+    def test_unhealthy_node_replacement(self):
+        args = ' --cluster-id j-ABC123456 --unhealthy-node-replacement'
+        cmdline = self.prefix + args
+        result = {'JobFlowIds': ['j-ABC123456'], 'UnhealthyNodeReplacement': True}
+        self.assert_params_for_cmd(cmdline, result)
+
+    def test_no_unhealthy_node_replacement(self):
+        args = ' --cluster-id j-ABC123456 --no-unhealthy-node-replacement'
+        cmdline = self.prefix + args
+        result = {'JobFlowIds': ['j-ABC123456'], 'UnhealthyNodeReplacement': False}
+        self.assert_params_for_cmd(cmdline, result)
+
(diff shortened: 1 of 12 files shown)


tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
```diff
@@ -0,0 +1,50 @@
...
+        self.assert_error_message_has_field_name(stderr, 'UnhealthyNodeReplacement')
+
+class TestSetUnhealthyNodeReplacementHelp(BaseAWSHelpOutputTest):
+    def test_set_unhealthy_node_replacement_is_undocumented(self):
+        self.driver.main(['emr', 'help'])
+        self.assert_not_contains('set-unhealthy-node-replacement')
+
+
+if __name__ == "__main__":
+    unittest.main()
```
kyleknap, line 50: Let's remove this. `pytest` is used for running tests and generally we won't call the test file directly to invoke it's `__main__`.

tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def assert_error_message_has_field_name(self, error_msg, field_name):
+        self.assertIn('Missing required parameter', error_msg)
+        self.assertIn(field_name, error_msg)
+
+    def test_unhealthy_node_replacement(self):
+        args = ' --cluster-id j-ABC123456 --unhealthy-node-replacement'
+        cmdline = self.prefix + args
+        result = {'JobFlowIds': ['j-ABC123456'], 'UnhealthyNodeReplacement': True}
+        self.assert_params_for_cmd(cmdline, result)
```
kyleknap, line 29: I'm thinking we just keep this test case out of all the `TestSetUnhealthyNodeReplacement` cases. We really only have a test case for this as a sanity check to make sure the command exists and is runnable. We do not need to test the full interface of the command as it is a model driven command, which generally has tests.

tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
```diff
@@ -0,0 +1,50 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from tests.unit.customizations.emr import EMRBaseAWSCommandParamsTest as \
+    BaseAWSCommandParamsTest
```
kyleknap, line 15: I'd recommend either 1) Not renaming `EMRBaseAWSCommandParamsTest` or 2) Just use `awscli.testutils.BaseAWSCommandParamsTest` directly. Mainly it is confusing to override the name of the child class to it's parent class at import. I don't have a strong preference on which option

tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
```diff
@@ -0,0 +1,50 @@
...
+        cmdline = self.prefix + args
+        result = {'JobFlowIds': ['j-ABC123456'], 'UnhealthyNodeReplacement': False}
+        self.assert_params_for_cmd(cmdline, result)
+
+    def test_at_least_one_option(self):
+        args = ' --cluster-id j-ABC123456'
+        cmdline = self.prefix + args
+        stderr = self.run_cmd(cmdline, 255)[1]
+        self.assert_error_message_has_field_name(stderr, 'UnhealthyNodeReplacement')
+
```
kyleknap, line 42: There should be two newlines here between classes

tests/unit/customizations/emr/test_set_unhealthy_node_replacement.py
```diff
@@ -0,0 +1,50 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from tests.unit.customizations.emr import EMRBaseAWSCommandParamsTest as \
+    BaseAWSCommandParamsTest
+from awscli.testutils import BaseAWSHelpOutputTest
+
+class TestSetUnhealthyNodeReplacement(BaseAWSCommandParamsTest):
```
kyleknap, line 18: There should be two newlines between classes and imports

kyleknap (commented):
Looks good. I just had some comments about the tests

PR #8579, opened 2024-03-14
S3 Config Documentation Update

goabv:
*Issue #, if available:*

*Description of changes:* Updating S3 configuration doc to encourage using CRT S3 client

### DIFF
--- a/awscli/topics/s3-config.rst
+++ b/awscli/topics/s3-config.rst
@@ -22,7 +22,7 @@ Configuration Values
 These are the configuration values you can set specifically for the ``aws s3``
 command set:
 
-* ``preferred_transfer_client`` - Determines the underlying Amazon S3 transfer client to use for transferring files to and from S3. For most workloads, choosing ``crt`` will enable faster uploads and downloads.
+* ``preferred_transfer_client`` - This determines the underlying Amazon S3 transfer client to use for transferring files to and from S3. For most workloads, choosing ``crt`` will enable faster uploads and downloads.
 * ``max_concurrent_requests`` - The maximum number of concurrent requests.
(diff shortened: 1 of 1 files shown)


awscli/topics/s3-config.rst
```diff
@@ -22,6 +22,7 @@ Configuration Values
 These are the configuration values you can set specifically for the ``aws s3``
 command set:
 
+* ``preferred_transfer_client`` - This determines the underlying Amazon S3 transfer client to use for transferring files to and from S3. For most workloads, choosing ``crt`` will enable faster uploads and downloads.
 * ``max_concurrent_requests`` - The maximum number of concurrent requests.
 * ``max_queue_size`` - The maximum number of tasks in the task queue.
```
graebm, line 27: Some of these config options only apply to `classic`, and some only apply to `crt`. If we're going to suggest `crt` we probably shouldn't show a bunch of settings that don't affect it.

This is currently detailed further down in the doc: https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client

Should we do a better job in this doc of breaking up which options affect both, which affect crt-only, and which are classic-only?

nateprewitt:
Thanks for the PR, @goabv. Most of this information is already in the CLI v2 docs [here](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client). We won't want to add this to the CLI v1 docs, as this PR does, because the CRT isn't available by default.

Was there something specifically insufficient with the current docs? Any changes should be made [here](https://github.com/aws/aws-cli/blob/v2/awscli/topics/s3-config.rst) on the `v2` branch.

PR #8608, opened 2024-03-29
Add S3 bucket validation to s3 mv

hssyoo:
When the `--validate-same-s3-paths parameter` is used or the env var `AWS_CLI_S3_MV_VALIDATE_SAME_S3_PATHS` is set to `true`, the s3 mv command will resolve underlying bucket names of source and destination S3 URIs and validate that an object is not being moved onto itself.


kyleknap (approved):
Looks good :ship:

PR #8610, opened 2024-03-30
Fix pytest-8.0.0 test failures due to incorrect `pytest.warns()`

yan12125:
*Issue #, if available:*

*Description of changes:*

This is a port of https://github.com/boto/botocore/pull/3109

Remove two incorrect `pytest.warns()` invocations
for `.get_component('internal')`, in order to fix test failures with pytest 8.0.0.  This version of pytest fixes `pytest.warns()` to work inside `pytest.raises()`, and therefore causes the tests to fail because the warning is not emitted when `get_component()` raises `ValueError`.

### DIFF
--- a/tests/unit/botocore/test_session.py
+++ b/tests/unit/botocore/test_session.py
@@ -640,9 +640,7 @@ class TestSessionComponent(BaseSessionTest):
         self.assertIs(
             self.session._get_internal_component('internal'), component)
         with self.assertRaises(ValueError):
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Original review here: https://github.com/boto/botocore/pull/3109#pullrequestreview-1855067095

These cases don't actually make sense where they've been placed. They should be safe to remove.

PR #8613, opened 2024-04-01
Add importlib-metadata requirement

hssyoo:
Adds the `importlib-metadata` package to the build requirements.


requirements/portable-exe-extras.txt
```diff
@@ -1 +1,2 @@
 pyinstaller==5.12.0
+importlib_metadata>=7.0,<=7.1.0
```
kyleknap, line 2: We should add a comment here why we are declaring `importlib_metadata` in this requirements file even though the build system does not directly require it.

kyleknap (commented):
Looks good. Just had a small comment.

kyleknap (approved):
Looks good :ship:

PR #8625, opened 2024-04-08
Move completions files

hssyoo:
Move completions model files from `awscli/data/` to `awscli/botocore/data`.

The diffs look like the commit is dropping 2 `completions-1.json` files, but it only looks that way because `cloudfront` [currently has 3 API versions](https://github.com/aws/aws-cli/tree/2.15.36/awscli/data/cloudfront) for `completions-1.json`, and they're all empty models. Because there's other services with empty completions models, git is having problems differentiating between them. Since we only need the latest API versions, we're dropping 2 of the older `cloudfront` completions files, which makes it look like we're dropping files.


kyleknap (approved):
Looks good :ship:

PR #8656, opened 2024-04-30
Sanitize __type metadata out of response parsing for tagged unions

nateprewitt:
This PR ports forward a bug fix from Botocore to properly parse tagged unions that contain a `__type` attribute in JSON RPC services.


kyleknap (approved):
:ship:

PR #8657, opened 2024-04-30
Fix mac builds for v2

nateprewitt:
It looks like tests have been broken for Python 3.8 and 3.9 on macOS since the release of macOS 14 which moved the runners to arm64 hardware. This will set both distribution and dependency workflows to use macos-13 which keeps tests on the amd64 hardware for the time being.


kyleknap (approved):
:ship:

PR #8664, opened 2024-05-02
User agent v2

kdaily:
*Issue #, if available:*

*Description of changes:*


tests/unit/test_clidriver.py
```diff
@@ -766,15 +768,15 @@ def test_idempotency_token_is_not_required_in_help_text(self):
     @mock.patch('awscli.clidriver.distro.major_version', return_value='1')
     def test_user_agent_for_linux(self, *args):
         driver = create_clidriver()
-        expected_user_agent = 'source/x86_64.amzn.1'
+        expected_user_agent = 'md/installer#source md/distrib#amzn.1'
         self.assertEqual(expected_user_agent,
                          driver.session.user_agent_extra)
 
     def test_user_agent(self, *args):
         machine = platform.machine()
```
kdaily, line 776: This is unused, should remove it.

tests/unit/test_clidriver.py
```diff
@@ -766,15 +768,15 @@ def test_idempotency_token_is_not_required_in_help_text(self):
...
         driver = create_clidriver()
         user_agent_extra_pattern = re.compile(
-            fr'^source/{machine}(\.[a-z_]+)?(\.[0-9]+)?$'
+            r'^md/installer#source'
```
kdaily, line 779: This was failing specifically on Ubuntu since we split the installer source and the arch/distribution into separate components.

awscli/utils.py
```diff
@@ -539,3 +540,30 @@ def __init__(self):
 
     def visit_shape(self, shape):
         self.visited.append(shape)
+
+
+def add_component_to_user_agent_extra(session, component):
+    if session.user_agent_extra and not session.user_agent_extra.endswith(" "):
+        session.user_agent_extra += " "
+    session.user_agent_extra += f"{component.to_string()}"
+
```
kdaily, line 549: Nit: needs another new line

tests/unit/test_clidriver.py
kdaily: `{machine}` is no longer needed because its no longer part of the user agent extra.

awscli/utils.py
```diff
@@ -539,3 +540,30 @@ def __init__(self):
...
+_LINEAGE_COMMAND_METADATA_REGEX = re.compile(r'md\/command#[\w\.]*')
+
+
+def add_command_lineage_to_user_agent_extra(session, lineage):
+    # Find existing command metadata and append the last item in the lineage to
+    # it
+    if comp := _LINEAGE_COMMAND_METADATA_REGEX.search(session.user_agent_extra):
+        session.user_agent_extra = \
+            _LINEAGE_COMMAND_METADATA_REGEX.sub(
+                f"{comp.group()}.{lineage[0]}", session.user_agent_extra
```
kdaily, line 566: This is wrong. It should be `lineage[-1]`.

awscli/utils.py
```diff
@@ -539,3 +540,31 @@ def __init__(self):
...
+    )
+
+
+_LINEAGE_COMMAND_METADATA_REGEX = re.compile(r'md\/command#[\w\.]*')
+
+
+def add_command_lineage_to_user_agent_extra(session, lineage):
+    # Find existing command metadata and append the last item in the lineage to
+    # it
+    if comp := _LINEAGE_COMMAND_METADATA_REGEX.search(session.user_agent_extra):
```
kyleknap, line 564: Looking at this now, I plan take a deeper look into this to see if we actually need this logic. This is definitely a bit messy and it would be nice if we could avoid it all together.

kyleknap (commented):
The implementation looks good. I only had some minor feedback. I plan on taking another pass later to go over the tests, but it's looking good so far.

awscli/botocore/compat.py
```diff
@@ -289,6 +289,17 @@ def get_tzinfo_options():
         return (tzlocal,)
 
 
+# Detect if CRT is available for use
```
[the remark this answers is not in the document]
kdaily: Agree, this was an unknown from my end. Removing the checks where it is used is resolved OK and tests pass.

awscli/botocore/__init__.py
```diff
@@ -16,7 +16,7 @@
 import os
 import re
 
-__version__ = '2.0.0dev155'
```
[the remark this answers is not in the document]
kdaily: Done.

awscli/botocore/__init__.py
```diff
@@ -16,7 +16,7 @@
...
+__version__ = '2.0.0'
```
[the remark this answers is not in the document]
kdaily: Done

awscli/utils.py
kyleknap: Circling back on this... So I think this pull request made the chance of the session having the command component less significant: https://github.com/aws/aws-cli/pull/7399. Specifically, the log will eagerly delegate to the subcommand's `__call__` instead of updating the user agent.

That being said there are instances such as `s3 rb --force` where a command is called directly: https://github.com/aws/aws-cli/blob/4613a2ef56e4df28325e0fd9d1c9187ae2477717/awscli/customizations/s3/subcommands.py#L913 but that ends up in unexpected updates to the command component (e.g. `md/command#s3.rb.rm` on the DeleteObject calls).

I think with this knowledge, we can probably just update it to check if `md/command` is defined and only add the command if it is not in the user agent. That should help simplify this logic.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,32 @@
+from awscli import __version__ as awscli_version
+from tests import CLIRunner
+
```
kyleknap, line 3: We should make sure there are two newlines here.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,32 @@
...
+def assert_user_agent_has_necessary_components(result, service, operation):
```
kyleknap, line 4: We can probably shorten this name to `assert_expected_user_agent`. I like that because we can then potentially expand it to include expected values as part of the function signature for some of these components down the road (e.g. `expected_installer`, `expected_arch`, etc.)

tests/integration/test_cli.py
```diff
@@ -270,11 +270,17 @@ def test_table_output(self):
...
+        version_output = p.stdout
+
+        # There should be four components: aws-cli version, Python version,
+        # platform version, and installation source. The CLI and Python
+        # version have fixed format; the platform and install source can
+        # change based on the system they are invoked.
+        self.assertTrue(len(version_output.split(" ")) == 4, p.stderr)
+
+        self.assertTrue(version_output.startswith('aws-cli'), p.stderr)
+        self.assertTrue("Python/" in version_output, p.stderr)
```
kyleknap, line 282: I wonder if we should make these assertions a regex? I think it will make the assertion a bit more exact and self documenting.

kyleknap (commented):
Looks good. Just had a few more comments on the tests.

kyleknap (approved):
Looks good! :ship:

PR #8738, opened 2024-06-12
[v2] Handle OSError and OverflowError in parse_timestamp

hssyoo:
Port of https://github.com/boto/botocore/pull/2972

Fall back to timedelta method for negative and post-2038 timestamps in parse_timestamp utility.


nateprewitt (approved):
:shipit:

PR #8744, opened 2024-06-13
Bump the ceiling for botocore memory leak tests to 15 MiB

AdamWill:
*Issue #, if available:*

*Description of changes:*
See https://github.com/boto/botocore/issues/3205 for the background on this. In rebuilding awscli2 for Python 3.13 in Fedora Rawhide, we found that two of these tests fail because they now top out around 11MiB of memory usage, rather than around 1.6MiB. We don't understand why this is yet, but it's not a memory *leak*, so bumping the ceiling seems appropriate. I'm sending this upstream so I have a reference for the downstream package and to raise awareness of the issue, but the correct fix may be something else.

### DIFF
--- a/tests/functional/botocore/leak/test_resource_leaks.py
+++ b/tests/functional/botocore/leak/test_resource_leaks.py
@@ -22,8 +22,8 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
     # a substantial amount of time to the total test run time.
     INJECT_DUMMY_CREDS = True
     # We're making up numbers here, but let's say arbitrarily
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks for this PR — I just [commented](https://github.com/boto/botocore/issues/3205#issuecomment-2176625718) on the issue you opened providing more context. After bringing team decided to close this PR as more research is required into the underlying issue, but please refer to https://github.com/boto/botocore/pull/3185 for further discussion around memory usage and testing in Python 3.13.

PR #8751, opened 2024-06-18
replace six.StringIO() by io.StringIO() in remaining places

a-detiste:
Description: trim `six` usage in a controlled way.


nateprewitt:
Thanks for the PR @a-detiste! We actually have a larger push for changes around `six` in both CLI v1 and v2 that will be released shortly. I believe it already encompasses this fix, so we'll defer to the larger PR. We do appreciate the contribution though!

a-detiste:
Thanks !

PR #8754, opened 2024-06-19
Remove six from awscli codebase

nateprewitt:
This PR removes usage of six in the CLI v1 codebase and minimizes its presence to one import from Botocore. This is left in place for backwards compatibility but is considered deprecated and may be removed in a future version. Downstream distributors may choose to patch out this single import to remove the requirement on keeping `six` in their ecosystem.

The primary changes here are in three classes:

1. `six.*_type(s)`, six maintains references to things like `bytes`, `str`, `basestr`, `unicode` depending on the Python version being used. We're removing this for the direct call to either `str` for `six.text_type` or `bytes` for `six.binary_type`. This also applies to `six.integer_types` and `six.string_types` which are used for type checking.

  ```python
  >>> import six
  >>> six.text_type
  <class 'str'>
  >>> six.binary_type
  <class 'bytes'>
  >>> six.string_types
  (<class 'str'>,)
  >>> six.integer_types
 (<class 'int'>,)
  ```
  
2. `six.moves.*`, these changes are placeholders that map between changed references to the same class. This was to account for modules that got renamed in Python 3. These are no longer needed so they've been moved to their Python 3 equivalent, you can find the full mapping table [here](https://six.readthedocs.io/#module-six.moves).

3. `six.b() six.u(), six.BytesIO(), six.StringIO()`, these are used for bringing StringIO/BytesIO back to Python 2 and dealing with the differences of bytes/unicode between the two major versions. StringIO and BytesIO are already present in the standard library `io` module, so we no longer need `six` for these. For `six.b` and `six.u`, these are just syntax sugar around `value.encode('latin-1')` and a literal no-op for `six.u` in Python 3. There's some discussion to be had on whether this should really be `value.encode('utf-8')` where we're using `six.b` but I've left it exactly as it was with `six` to make sure we're not breaking anything.

If you have more questions, you can read the full six docs [here](https://six.readthedocs.io).

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -12,7 +12,6 @@
 # language governing permissions and limitations under the License.
 import argparse
 import sys
-from awscli.compat import six
 from difflib import get_close_matches
 
 
@@ -106,12 +105,12 @@ class CLIArgParser(argparse.ArgumentParser):
             # default to utf-8.
             terminal_encoding = 'utf-8'
         for arg, value in vars(parsed).items():
-            if isinstance(value, six.binary_type):
+            if isinstance(value, bytes):
                 setattr(parsed, arg, value.decode(terminal_encoding))
             elif isinstance(value, list):
                 encoded = []
                 for v in value:
-                    if isinstance(v, six.binary_type):
+                    if isinstance(v, bytes):
                         encoded.append(v.decode(terminal_encoding))
                     else:
                         encoded.append(v)
--- a/awscli/argprocess.py
+++ b/awscli/argprocess.py
@@ -13,7 +13,6 @@
 """Module for processing CLI args."""
 import os
 import logging
-from awscli.compat import six
 
 from botocore.compat import OrderedDict, json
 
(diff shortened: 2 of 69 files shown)


hssyoo (commented):
Thanks for putting this together. Just had a couple things:
* Is there a reason for not removing `six` usage from tests? Looks like these 2 in particular directly import `six`:
https://github.com/aws/aws-cli/blob/577fd3a96969071bdc61095773ff324c094e24f5/tests/unit/customizations/cloudformation/test_deploy.py#L14
https://github.com/aws/aws-cli/blob/577fd3a96969071bdc61095773ff324c094e24f5/tests/unit/customizations/cloudformation/__init__.py#L14
* It looks like there's some leftover `six` usage here:
https://github.com/aws/aws-cli/pull/8754/files#diff-61b73a8cd44d177aa984c8a0e4dc8b9f29898fc269243b7b846aab73c203d2f5R103-R108

nateprewitt:
Good call out, it looks like there's actually a bunch of usage in the test directory. The grep expression I was using missed a number of these, I've updated it and will get those pushed up.

hssyoo (approved):
🏆

PR #8757, opened 2024-06-20
[v2] Generate dependency lockfiles

hssyoo:
To provide a stable build environment, this PR does the following:
* Move `requirements.txt` -> `requirements-dev.txt` to align with AWS CLI v1 and boto projects
* Alias `requirements.txt` to `requirements-dev.txt` for backwards compatibility
* Create a `requirements-build-win.txt` file as an extra source for `pip-compile` so lockfiles include Windows dependencies
* Move from `mock` to `unittest.mock` import to remove the dependency and simplify dependency resolution
* Generate lockfiles from `requirements*.txt` files and replace their references in install scripts with `requirements*-lock.txt`


requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
+-r requirements-base.txt
+tox==4.4.12
+build==0.7.0
```
nateprewitt, line 3: Just curious why `build` is here and not in either `-base` or `-build`?

requirements-base-lock.txt
```diff
@@ -0,0 +1,14 @@
+#
```
nateprewitt, line 1: If we install every lockfile at once do we have anything ensuring we don't have conflicts there? i.e. if we update one lockfile but forget to do the others, will we know before we merge?

nateprewitt (changes requested):
A couple quick comments about the mock changes but those otherwise look good. A quick note that splitting these kinds of changes into two PRs may be helpful in the future. That lets us keep review overhead lower and minimizes rollback impact. Ideally if one piece is broken we don't need to write a rollback by hand, and don't rollback unnecessary changes if we use `git revert`.

requirements-dev.txt
hssyoo: Based on historic commits and comments...
* `requirements-build` is specifically for dependencies required for building installers
* `build ` was added in 2021 while `requirements-base` was added in 2022 and seems it was missed

I would agree that it makes sense to move `build` to `requirements-base`

requirements-base-lock.txt
hssyoo: Hm that's a good point. At a minimum we should rework [regenerate-lock-files](https://github.com/aws/aws-cli/blob/v2/scripts/regenerate-lock-files) or make use of it to generate all of these at once.

This also raises another concern where we make a dependency in `pyproject.toml` more restrictive but forget to regenerate `requirements-docs-lock.txt` (which takes `pyproject.toml` as input).

PR #8758, opened 2024-06-20
Six removal v2

nateprewitt:
Follow up to #8754 to start moving the six removals into the CLI v2. This should be an almost identical patch from v1 with some minor additions to botocore and net-new CLI code in v2.

Currently in draft, this should encompass all usages of six and bring the files in the CLI v2 inline with botocore/s3transfer to minimize conflicts with future merges.


tests/functional/docs/test_help_output.py
```diff
@@ -21,15 +21,15 @@
...
 
 """
-import io
 import os
 
 from awscli.testutils import BaseAWSHelpOutputTest
 from awscli.testutils import FileCreator
 from awscli.testutils import mock
 from tests import CLIRunner
+from awscli.testutils import aws
```
hssyoo, line 30: What's this import for?

tests/functional/s3/test_cp_command.py
```diff
@@ -18,16 +18,16 @@
 from awscli.customizations.s3.utils import relative_path
 from awscli.testutils import BaseAWSCommandParamsTest
 from awscli.testutils import capture_input, mock
-from awscli.compat import six, OrderedDict
+from awscli.compat import BytesIO, OrderedDict
 from tests.functional.s3 import (
     BaseS3TransferCommandTest, BaseS3CLIRunnerTest, BaseCRTTransferClientTest
 )
-
+from tests import requires_crt
```
hssyoo, line 25: Same here, looks like an unused import

hssyoo (commented):
Looks good, thanks for this PR. Just had a couple questions on added imports

tests/functional/s3/test_cp_command.py
nateprewitt: Same note, this got auto-added with the cherry-pick. v2 requires the CRT so this isn't needed. In the future once we get `ruff` in place this will be handled automatically for us.

tests/functional/docs/test_help_output.py
nateprewitt: It looks like this got added automatically when cherry-picking the v1 commit onto v2. It looks like Kyle rewrote the [original v1 test](https://github.com/aws/aws-cli/commit/9b3b86c58eadb96023698f84f86fefc01b70c171#diff-07670404753d70f985622d5db67d414d39dd65720a7bbdfb41b0e56e92a9971dR492) in v2 to use `CLIRunner` instead of `aws`. I'll remove it.

hssyoo (approved):
🏆

PR #8765, opened 2024-06-25
Correct JSON sample output of sts assume-role-with-web-identity command

DlieBG:
In the example output of the sts assume-role-with-web-identity command, two commas are missing to be valid JSON.


elysahall (approved):
LGTM, sending for maintainer review.

AndrewAsseily (approved):

PR #8792, opened 2024-07-09
Add changelog for Python interpreter version update

hssyoo:
This is applicable to the bundled Python interpreter versions in official AWS CLI v2 installers and Docker images.


nateprewitt (approved):
⛵

PR #8794, opened 2024-07-11
[v2] Add s3 integration tests using directory buckets

hssyoo:
Set up S3 directory bucket creation/teardown for integration tests with high-level `s3` commands. For the existing teardown logic to be compatible with directory buckets, we switch from `ListObjects` to `ListObjectsV2` when clearing out test buckets.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -181,6 +181,46 @@ def create_bucket(session, name=None, region=None):
     return bucket_name
 
 
+def create_dir_bucket(session, name=None, region=None, az=None):
+    """
+    Creates a S3 directory bucket
+    :returns: the name of the bucket created
+    """
+    if not region:
+        region = 'us-west-2'
+    client = session.create_client('s3', region_name=region)
+    if not az:
+        az = 'usw2-az1'
+    if name:
+        bucket_name = name
+    else:
+        bucket_name = f"{random_bucket_name()}--{az}--x-s3"
+    params = {
+        'Bucket': bucket_name,
+        'CreateBucketConfiguration': {
+            'Location': {
+                'Type': 'AvailabilityZone',
+                'Name': az
+            },
+            'Bucket': {
+                'Type': 'Directory',
+                'DataRedundancy': 'SingleAvailabilityZone'
+            }
+        }
+    }
+    try:
+        client.create_bucket(**params)
+    except ClientError as e:
+        if e.response['Error'].get('Code') == 'BucketAlreadyOwnedByYou':
+            # This can happen in the retried request, when the first one
+            # succeeded on S3 but somehow the response never comes back.
+            # We still got a bucket ready for test anyway.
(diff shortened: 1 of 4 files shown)


awscli/testutils.py
```diff
@@ -181,6 +181,46 @@ def create_bucket(session, name=None, region=None):
...
+def create_dir_bucket(session, name=None, region=None, az=None):
+    """
+    Creates a S3 directory bucket
+    :returns: the name of the bucket created
+    """
+    if not region:
+        region = 'us-west-2'
+    client = session.create_client('s3', region_name=region)
+    if not az:
+        az = 'usw2-az1'
```
nateprewitt, line 193: We might consider making the `region` and `az` input a tuple since they're inherently bound. i.e. if I pass `us-east-1` but forget to add the `az`, I don't think we want those disjoint, do we?

Something like this:
```suggestion
    if not location:
        location = ('us-west-2', 'usw2-az1')
    region, az = location
    client = session.create_client('s3', region_name=region)
```

nateprewitt (commented):
Quick interface comment, but otherwise looks good.

nateprewitt (approved):

awscli/testutils.py
```diff
@@ -181,6 +181,45 @@ def create_bucket(session, name=None, region=None):
...
 
 
+def create_dir_bucket(session, name=None, location=None):
+    """
+    Creates a S3 directory bucket
+    :returns: the name of the bucket created
+    """
+    if not location:
+        location = ('us-west-2', 'usw2-az1')
+    region, az = location
```
kdaily, line 191: nit: name the parameter `availability_zone` instead of `az` to avoid using abbreviations.

awscli/testutils.py
```diff
@@ -181,6 +181,45 @@ def create_bucket(session, name=None, region=None):
     return bucket_name
 
 
+def create_dir_bucket(session, name=None, location=None):
```
kdaily, line 184: It would be helpful to a developer to put a docstring for the type of `location`, since its a tuple.

tests/integration/customizations/s3/conftest.py
```diff
@@ -47,6 +47,20 @@ def shared_copy_bucket(s3_utils):
...
 
 
+@pytest.fixture(scope='package')
+def shared_dir_bucket(s3_utils):
+    bucket_name = s3_utils.create_dir_bucket()
+    yield bucket_name
+    s3_utils.delete_bucket(bucket_name)
+
+
+@pytest.fixture(scope='package')
```
kdaily, line 57: Document this the same as `shared_copy_bucket`:

```
# Directory bucket for cross-S3 copies
```

kdaily (changes requested):
Only a nit on the naming and docs, looks good otherwise.

kdaily (approved):

PR #8795, opened 2024-07-11
Add s3 integration tests using directory buckets

hssyoo:
Set up S3 directory bucket creation/teardown for integration tests with high-level s3 commands. For the existing teardown logic to be compatible with directory buckets, we switch from `ListObjects` to `ListObjectsV2` when clearing out test buckets.


kdaily (changes requested):
Looks good overall, a couple minor things. just around some naming consistency.

kdaily (approved):

PR #8820, opened 2024-07-26
[v2] Disable S3 Express support for s3 mb command

aemous:
*Description of changes:*
Disable S3 Express support for s3 mb command


hssyoo (approved):
🏆

PR #8832, opened 2024-07-31
Add validation for numerical pagination argument values to be non-negative

aemous:
*Description of changes:*
I've added validation on numerical pagination arguments to ensure they are non-negative. This fixes a bug where using negative numbers for pagination arguments can cause erroneous results.


awscli/customizations/paginate.py
```diff
@@ -251,11 +251,21 @@ def _get_cli_name(param_objects, token_name):
...
+        raise ValueError("invalid literal for nonnegative int() with base 10: '%s'" % value)
+    return ival
+
+
 class PageArgument(BaseCLIArgument):
     type_map = {
         'string': str,
-        'integer': int,
-        'long': int,
+        'integer': nonnegative_int,
```
aemous, line 267: One possible concern here is that we are now enforcing ALL numerical page arguments be non-negative. As of today, that's what we want. In the future, if only a subset of pagination arguments should be validated as non-negative, some refactoring would be necessary. Overall, I think this is reasonable, but I'm happy to take feedback.

aemous:
Going with another approach for this

PR #8833, opened 2024-07-31
[V2] Add validation for numerical pagination argument values to be non-negative

aemous:
*Description of changes:*
I've added validation on numerical pagination arguments to ensure they are non-negative. This fixes a bug where using negative numbers for pagination arguments can cause erroneous results.


awscli/customizations/paginate.py
```diff
@@ -298,11 +298,21 @@ def check_should_enable_pagination_call_parameters(
...
+        raise ValueError("invalid literal for nonnegative int() with base 10: '%s'" % value)
+    return ival
+
+
 class PageArgument(BaseCLIArgument):
     type_map = {
         'string': str,
-        'integer': int,
-        'long': int,
+        'integer': nonnegative_int,
```
aemous, line 314: Possible future concern: see my comment on V1 PR about this.

aemous:
Going with another approach to this

PR #8838, opened 2024-08-02
[v2] Use higher key size for TestSha256RSADigestValidator

fossdd:
*Issue #, if available:*

*Description of changes:*

[cryptography 43.0.0 ](https://github.com/pyca/cryptography/blob/43.0.x/CHANGELOG.rst) requires bumped the minimum key size to 1024.

Since `rsa.generate_private_key()` is only used in this test, this is fine.

### DIFF
--- a/tests/unit/customizations/cloudtrail/test_validation.py
+++ b/tests/unit/customizations/cloudtrail/test_validation.py
@@ -362,7 +362,7 @@ class TestSha256RSADigestValidator(unittest.TestCase):
         self._digest_data['_signature'] = 'aeff'
 
     def test_validates_digests(self):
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @fossdd! Thanks for opening this. I came across this separately and bumped up the keysize to 2048 (#8903). Closing this out for now. We are still working towards moving the pin for cryptography.

fossdd:
> I came across this separately and bumped up the keysize to 2048.

That's great! Thanks!

> We are still working towards moving the pin for cryptography.

Yeah, downstream at Alpine (and other distros) we mostly use latest version of libraries and notice changes like that early. But I don't mind fixing them and upstream them, if that makes life easier (and hopefully doesn't duplicate work).

PR #8846, opened 2024-08-06
[V2] Load smoke tests from data file

aemous:
*Description of changes:*
Replaced the hard-coded smoke tests with a JSON data file.


tests/integration/test_smoke.py
```diff
@@ -12,109 +12,20 @@
...
 import os
-import random
 import time
+import json
 
 import pytest
 
 from awscli.testutils import aws
 
+INTEG_TESTS_DIR = os.path.dirname(os.path.join(os.path.abspath(__file__)))
```
hssyoo, line 22: It looks like we don't need `os.path.join` here.

tests/integration/smoke-test.json
```diff
@@ -0,0 +1,332 @@
+{
```
hssyoo, line 1: nit: Let's name this file `smoke-tests.json`

hssyoo (changes requested):
Looks solid, just had a couple minor comments.

tests/integration/smoke-test.json
aemous: I've updated this in my latest revision.

tests/integration/test_smoke.py
aemous: Updated in latest revision

tests/integration/test_smoke.py
```diff
@@ -12,109 +12,20 @@
...
-]
-
-
-# These services require a particular region to run.
-REGION_OVERRIDES = {
-    'route53domains': 'us-east-1'
-}
+with open(os.path.join(INTEG_TESTS_DIR, "smoke-test.json"), "r") as definitions_file:
+    SMOKE_TESTS = json.load(definitions_file)
+    COMMANDS = [" ".join(SMOKE_TESTS["Commands"][i]) for i in range(len(SMOKE_TESTS["Commands"]))]
```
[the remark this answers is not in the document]
aemous: Updated in latest revision

hssyoo (approved):
🏆

PR #8848, opened 2024-08-06
Add smoke-tests JSON and load them from smoke tests.

aemous:
*Description of changes:*
Replaced the hard-coded smoke tests with a JSON data file.


aemous:
Actually, this change is not needed for v1.

PR #8855, opened 2024-08-08
[V2] Add warning for invalid max-items value

aemous:
*Description of changes:*
Explicitly warn users that we do not support non-positive values for the max-items global pagination argument. 

Example output:
``` 
aws s3api list-objects --bucket bucket-123 --max-items 0
2024-08-08 12:11:58,993 - awscli.customizations.paginate - CRITICAL - Non-positive values for --max-items are unsupported and may yield undefined behavior.

{
    "RequestCharged": null
} 
```


awscli/customizations/paginate.py
```diff
@@ -225,6 +234,13 @@ def check_should_enable_pagination(input_tokens, shadowed_args, argument_table,
             argument_table[key] = value
 
 
+def check_and_warn_negative_max_items(parsed_args, parsed_globals, **kwargs):
+    parsed_max_items = getattr(parsed_args, 'max_items', None)
+    if parsed_globals.paginate and parsed_max_items is not None:
+        if int(parsed_max_items) <= 0:
+            logger.critical("Non-positive values for --max-items are unsupported and may yield undefined behavior.")
```
hssyoo, line 241: For warnings in the AWS CLI, we have a couple conventions:
* Use `uni_print` to print to stderr, like [this example](https://github.com/aws/aws-cli/blob/v2/awscli/customizations/s3/subcommands.py#L1315)
* We format the string to start with `warning: `

awscli/customizations/paginate.py
```diff
@@ -164,23 +164,32 @@ def unify_paging_params(argument_table, operation_model, event_name,
...
+    # The reason we have to do the third is that someone could put
     # operation-specific pagination args in the CLI input JSON file
     # directly and this bypasses all of the CLI args processing.
     session.register(
         parsed_args_event,
         partial(check_should_enable_pagination,
                 list(_get_all_cli_input_tokens(paginator_config)),
                 shadowed_args, argument_table))
+
+    session.register(
```
hssyoo, line 189: I'm not sure it makes sense to do the validation here. Mainly, this function is responsible for ensuring the AWS CLI's custom paging parameters aren't conflicting with the underlying service model parameters.

Instead, let's defer the validation until the max items object for `PageArgument` calls `add_to_params`. This also makes it more explicit that the parameter we're validating is for the paginator, and not some parameter named `MaxItems` that a service model defined.

hssyoo (changes requested):
Looks like a good start! You've got the right idea, I just had a couple suggestions on where to do the validation.

awscli/customizations/paginate.py
aemous: Thanks, I addressed this in my latest revision.

awscli/customizations/paginate.py
aemous: Thanks, I addressed this in my latest revision.

tests/integration/test_smoke.py
```diff
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 import re
 import os
+import random
```
kdaily, line 15: Doesn't look like this is needed!

tests/unit/customizations/test_paginate.py
```diff
@@ -19,7 +23,6 @@
 from awscli.customizations import paginate
 from awscli.customizations.exceptions import ParamValidationError
 
-
```
kdaily, line 22: nit: we should keep this line for formatting reasons, two new lines after the imports.

.changes/next-release/enhancement-paginator-41862.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "paginator",
+  "description": "Add critical warning when the user enters a non-positive value for the max-items pagination parameter."
```
kdaily, line 4: I think its enough to say we added a warning. I'm guessing the "critical" descriptor came when you were using `logging.critical`, but since thats removed we should update here too.

 ```suggestion
  "description": "Add warning when a non-positive value is provided for the max-items pagination parameter."
```

**Review by kdaily — DISMISSED** [review]

File: awscli/customizations/paginate.py [hunk 6]
```diff
@@ -24,6 +24,7 @@
 
 """
 import logging
+import sys
```
kdaily, line 27: This might not be needed either!

kdaily (changes requested):

awscli/customizations/paginate.py
```diff
@@ -337,7 +338,15 @@ def add_to_parser(self, parser):
         parser.add_argument(self.cli_name, dest=self.py_name,
                             type=self.type_map[self._parse_type])
 
+    def _check_and_warn_negative_max_items(self, value):
+        if int(value) <= 0:
+            uni_print(
+                "warning: Non-positive values for --max-items are unsupported and may yield undefined behavior.\n",
```
kdaily, line 344: Nit: Saying they are unsupported seems that they would raise an error; they are allowed but will likely result in nonsense.
 
```suggestion
                "warning: Non-positive values for --max-items may result in undefined behavior.\n",
```

tests/unit/customizations/test_paginate.py
```diff
@@ -350,3 +354,25 @@ def test_can_handle_missing_page_size(self):
         del self.parsed_args.page_size
         self.assertIsNone(paginate.ensure_paging_params_not_set(
             self.parsed_args, {}))
+
+class TestNonPositiveIntWarnings(TestPaginateBase):
```
hssyoo, line 358: I know that you're following the current format of the test module, but we should try to write new tests using `pytest`. I think this is a good test to use `pytest` because we don't need any scaffolding from `TestPaginateBase`. I recommend setting up a fixture at the top of the module for the page argument:
```
@pytest.fixture
def page_arg():
    return PageArgument('max-items', 'documentation', int, 'MaxItems')
```

And instead of using `io.StringIO` to capture stderr, we can use [`pytest`'s `capsys`](https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html):
```
class TestNonPositiveMaxItems:
    def test_positive_integer_does_not_raise_warning(self, page_arg, capsys):
        page_arg.add_to_params({}, 1)
        captured = capsys.readouterr()
        assert captured.err == ""

    def test_zero_raises_warning(self, page_arg, capsys):
        page_arg.add_to_params({}, 0)
        captured = capsys.readouterr()
        assert "Non-positive values for --max-items" in captured.err

    def test_negative_integer_raises_warning(self, page_arg, capsys):
        page_arg.add_to_params({}, -1)
        captured = capsys.readouterr()
        assert "Non-positive values for --max-items" in captured.err
```

And a nit: The test class name could be more accurate since we're specifically testing the `MaxItems` parameter.

hssyoo (changes requested):

aemous:
Iterated on all requested feedback. If this looks good I'll open a V1 version with the same code. @hssyoo

tests/unit/customizations/test_paginate.py
```diff
@@ -350,3 +353,23 @@ def test_can_handle_missing_page_size(self):
         del self.parsed_args.page_size
         self.assertIsNone(paginate.ensure_paging_params_not_set(
             self.parsed_args, {}))
+
+class TestNonPositiveMaxItems:
+    @pytest.fixture
+    def max_items_page_arg(self):
+        return PageArgument('max-items', 'documentation', int, 'MaxItems')
```
hssyoo, line 360: I meant we can put this at the top of the module. The nice thing about fixtures is that `pytest` has a predictable ordered list of locations where it looks for fixtures, so they don't need to be tied down to test classes. Keeping this at the top will make it more reusable. I know there's a question of whether or not this fixture will be used by other tests in the future, but I can imagine generalizing this to be a `page_arg` fixture that acts as sort of a factory for returning different `PageArgument` instances.

awscli/customizations/paginate.py
```diff
@@ -337,8 +338,16 @@ def add_to_parser(self, parser):
...
 
+    def _emit_non_positive_max_items_warning(self, value):
+        if int(value) <= 0:
+            uni_print(
+                "warning: Non-positive values for --max-items may result in undefined behavior.\n",
+                sys.stderr)
+
     def add_to_params(self, parameters, value):
         if value is not None:
+            if self._serialized_name == 'MaxItems':
```
hssyoo, line 349: We should check `int(value) <=0` here and remove it from `_emit_non_positive_max_items_warning`.

hssyoo (changes requested):

aemous:
Iterated on all requested feedback. If this looks good I'll open a V1 version with the same code. @hssyoo

awscli/customizations/paginate.py
```diff
@@ -337,8 +338,15 @@ def add_to_parser(self, parser):
         parser.add_argument(self.cli_name, dest=self.py_name,
                             type=self.type_map[self._parse_type])
 
+    def _emit_non_positive_max_items_warning(self):
+            uni_print(
+                "warning: Non-positive values for --max-items may result in undefined behavior.\n",
+                sys.stderr)
```
hssyoo, line 344: nit: The formatting looks off here

aemous: Updated in latest revision. Please let me know if we have a formatter / formatting guide anywhere for this repository.  I kept it split into 2 lines to avoid too-long of a line but not sure if we have formal requirements on that anywhere.

hssyoo (approved):

PR #8865, opened 2024-08-13
Fix IPv6 support

nateprewitt:
This PR ports https://github.com/boto/botocore/pull/2574 forward into the CLIv2 as it appears to have been missed. This will avoid incorrect validation failures within botocore and defer validation/routing decisions down the stack to urllib3.


hssyoo (approved):
🏆

PR #8869, opened 2024-08-15
Fix IPv6 SigV4 Signing

nateprewitt:
This PR fixes missing brackets from the `Host` header when supplying a hostname that matches an IP-literal as defined in [RFC 3986 3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). This will resolve issues between how we're signing requests and how they're being interpreted by AWS services which is currently leading to signature mismatches.


kdaily (approved):
Yep, looks like per the RFC we should be doing that. 📈

PR #8897, opened 2024-09-04
[V2] Add Python 3.12 Support

aemous:
*Issue #, if available:*
- https://github.com/aws/aws-cli/issues/8342 

*Description of changes:*
This PR adds changes needed to support Python version 3.12. A highlight of the changes are:
- Updating Github workflows to run with Python 3.12.
- Updating pyproject.toml, tox.ini, and test files to support changes introduced in Python 3.12, and deprecation warnings added in dependencies.
- Install setuptools in all CI builds with Python 3.12+ where it's no longer included by default (similar to #8106).
- Fixing EKS test suite (similar to #8106).
- Fixing botocore smoke test integration suite. Specifically, I removed our assertion that no warnings are emitted. PyTest already guarantees this, and which we configure with `pyproject.toml`. 
- Updated a botocore unit test suite against Python 3.12's update to BufferedIOBase.


aemous:
In reply to the only failing check, it is due to a failed assertion on the version of urllib3. For some reason, the the regenerated portable-exe-win-lock.txt file expects 1.26.19, but our lockfile has 1.26.20 pinned. Both satisfy our pyproject.toml requirement of `urllib3>=1.25.4,<1.27`.

pyproject.toml
```diff
@@ -36,11 +38,11 @@ dependencies = [
     # less than or equal to Python 3.10. In order to ensure we have
     # a consistent dependency closure across all Python versions,
     # we explicitly include ruamel.yaml.clib as a dependency.
-    "ruamel.yaml.clib>=0.2.0,<=0.2.7",
+    "ruamel.yaml.clib>=0.2.0,<=0.2.8",
     "prompt-toolkit>=3.0.24,<3.0.39",
     "distro>=1.5.0,<1.9.0",
     "awscrt>=0.19.18,<=0.21.2",
-    "python-dateutil>=2.1,<=2.8.2",
+    "python-dateutil>=2.1,<=2.9.0",
```
hssyoo, line 45: Why do we need to increase the ceiling?

pyproject.toml
```diff
@@ -36,11 +38,11 @@ dependencies = [
     # less than or equal to Python 3.10. In order to ensure we have
     # a consistent dependency closure across all Python versions,
     # we explicitly include ruamel.yaml.clib as a dependency.
-    "ruamel.yaml.clib>=0.2.0,<=0.2.7",
+    "ruamel.yaml.clib>=0.2.0,<=0.2.8",
```
hssyoo, line 41: Why do we need to increase the ceiling?

requirements-build.txt
```diff
@@ -1,4 +1,5 @@
 # Requirements we need to run our build jobs for the installers.
 # We create the separation for cases where we're doing installation
 # from a local dependency directory instead of requirements.txt.
-PyInstaller==5.12.0
+PyInstaller==5.13.2
+pip==24.0
```
hssyoo, line 5: Why are we pinning `pip`?

scripts/ci/install
```diff
@@ -13,6 +14,13 @@ os.chdir(REPO_ROOT)
 def run(command):
     return check_call(command, shell=True)
 
+if sys.version_info[:2] >= (3, 12):
+    # Python 3.12+ no longer includes setuptools by default.
+
+    # Setuptools 71+ now prefers already installed versions
+    # of packaging _and_ broke the API for packaging<22.0.
+    # We'll pin to match what's in requirements-dev.txt.
+    run("pip install setuptools==71.1.0 packaging==24.1")
```
hssyoo, line 23: Why do we need to pin `packaging`? Wouldn't a compatible version be installed when we pip install `setuptools==71.1.0`?

tests/integration/botocore/test_smoke.py
```diff
@@ -273,11 +271,8 @@ def test_can_make_request_with_client(
     # instead of service/operations.
     client = _get_client(botocore_session, service_name)
     method = getattr(client, xform_name(operation_name))
-    with warnings.catch_warnings(record=True) as caught_warnings:
-        response = method(**kwargs)
-        err_msg = f"Warnings were emitted during smoke test: {caught_warnings}"
-        assert len(caught_warnings) == 0, err_msg
-        assert 'Errors' not in response
```
hssyoo, line 280: Why do we need to do this? The PR description doesn't go into what this is fixing. It just looks like we're getting less coverage. It looks like [botocore never removed the warnings assertion here](https://github.com/boto/botocore/blob/develop/tests/integration/test_smoke.py#L286). We should avoid drifting if it's not necessary.

hssyoo (changes requested):
Looks fine, had questions on some changes being made. Let's also remember to add a new change entry.

pyproject.toml
aemous: 2.9.0 removes a call to a deprecated function

pyproject.toml
aemous: was facing a build issue with this library without this

scripts/ci/install
aemous: Ported change from python 3.12 support for v1.

PR #8917, opened 2024-09-12
[V2] Add Python 3.12 Support

aemous:
*Issue #, if available:*
- https://github.com/aws/aws-cli/issues/8342 

*Description of changes:*
This PR adds changes needed to support Python version 3.12. A highlight of the changes are:
- Updating Github workflows to run with Python 3.12.
- Updating pyproject.toml, tox.ini, and test files to support changes introduced in Python 3.12, and deprecation warnings added in dependencies.
- Install setuptools in all CI builds with Python 3.12+ where it's no longer included by default (similar to #8106).
- Fixing EKS test suite (similar to #8106).
- Modified botocore smoke test integration test to explicitly ignore a known warning emitted as a result of a dependency calling a deprecated function
- Updated a botocore unit test suite against Python 3.12's update to BufferedIOBase.


tests/functional/eks/test_kubeconfig.py
```diff
@@ -120,7 +120,7 @@ def test_load_simple(self):
         ])
         loaded_config = self._loader.load_kubeconfig(simple_path)
         self.assertEqual(loaded_config.content, content)
-        self._validator.validate_config.called_with(Kubeconfig(simple_path,
+        self._validator.validate_config.assert_called_with(Kubeconfig(simple_path,
                                                                content))
```
hssyoo, line 124: nit: The formatting looks off here.

tests/unit/botocore/test_utils.py
```diff
@@ -1049,16 +1049,24 @@ def test_is_unaffected_by_sigv4(self):
 
 
 class TestSwitchToChunkedEncodingForNonSeekableObjects(unittest.TestCase):
-    def test_switch_to_chunked_encodeing_for_stream_like_object(self):
+    def test_chunked_encoding_used_for_stream_like_object(self):
```
hssyoo, line 1052: There's nothing technically wrong with this, but we should stay consistent with how the [test was changed in `botocore`](https://github.com/boto/botocore/blob/develop/tests/unit/test_utils.py#L1122-L1139).

hssyoo (changes requested):

hssyoo (approved):
🏆

Issue #6750, opened 2022-02-26
Support S3 additional checksums in high-level S3 commands


m-radzikowski:
**Is your feature request related to a problem? Please describe.**

Newly released [additional S3 checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/) feature enhances the SDKs operations by calculating selected checksum value on file upload. This also includes multipart upload. However, this new feature is not present in the high-level S3 commands.

**Describe the solution you'd like**

`--checksum-algorithm` parameter in the `aws s3` commands, especially in the `aws s3 cp`.

**Describe alternatives you've considered**

Using low-level commands.

tim-finnigan:
Hi @m-radzikowski thanks for the feature request. I’m going to bring this up for discussion with the team and will post an update here.

rajivnarayan:
This would be a useful addition to the high-level commands. For reference here is a solution using s3api:
Would have been nice if MD5 digests were included as an option.

```
# aws-cli version 2.7.16
# https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/

# compute and save checksum on upload / copy
# algorithms supported: crc32 crc32c sha1 sha256
aws s3api put-object --body <file_name> --checksum-algorithm crc32 --bucket <bucket_name> --key <key_name>

# retrieve the checksum```
# ChecksumCRC32 ChecksumCRC32C ChecksumSHA1 ChecksumSHA256
aws s3api head-object --bucket <bucket_name> --key <key_name> --checksum-mode Enabled --Query ChecksumCRC32 --output text
```

jonathansampson:
+1 to support for checksums when syncing.

saksham:
+1 on this feature.

genvidkyle:
+1

ashepherd:
+1

MaksymSimchuk-prxt:
+1

khilnani:
+1

sarthakjain271095:
@rajivnarayan i stumbled upon [this](https://github.com/aws/aws-cli/issues/6750#issuecomment-1195959947). The sha256 checksum value being returned from aws doesn't seem to be right. Additionally, as per the [cli doc](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-object.html), `--checksum-algorithm` param is only supported when using SDK. Have you faced issues about the sha256 value not being rightly calculated?

[...]

PR #8933, opened 2024-09-20
[v2] S3 high level checksums

aemous:
*Issue #, if available:*
- https://github.com/aws/aws-cli/issues/6750

Low-level `aws s3api` commands support checksums other than MD5 for verifying end-to-end data integrity. This pull request ports over similar functionality to high-level `aws s3` commands. 

*Description of changes:*
- Added `--checksum-algorithm` flag for file uploads using `aws s3 cp`, `aws s3 sync`, and `aws s3 mv`.
  - Supported algorithms are CRC32, SHA256, SHA1, CRC32C.
- Added `--checksum-mode` flag for file downloads using `aws s3 cp`, `aws s3 sync`, and `aws s3 mv`.
- Added unit and functional tests to assert the correctness of these changes.

### DIFF
--- a/.changes/next-release/feature-s3-46667.json
+++ b/.changes/next-release/feature-s3-46667.json
@@ -1,5 +1,5 @@
 {
   "type": "feature",
   "category": "s3",
-  "description": "Adds support for checksums other than MD5 to high-level S3 commands."
+  "description": "Adds ``--checksum-mode`` and ``--checksum-algorithm`` parameters to high-level ``s3`` commands."
 }
--- a/.changes/next-release/feature-s3-46667.json
+++ b/.changes/next-release/feature-s3-46667.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "s3",
+  "description": "Adds support for checksums other than MD5 to high-level S3 commands."
+}
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -483,7 +483,8 @@ VALIDATE_SAME_S3_PATHS = {
 
 CHECKSUM_MODE = {
         'name': 'checksum-mode', 'choices': ['ENABLED'],
-        'help_text': 'To retrieve the checksum, this mode must be enabled.'
+        'help_text': 'To retrieve the checksum, this mode must be enabled. If the object has a '
+                     'checksum, it will be verified.'
 }
 
 CHECKSUM_ALGORITHM = {
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -481,6 +481,17 @@ VALIDATE_SAME_S3_PATHS = {
     )
 }
 
+CHECKSUM_MODE = {
+        'name': 'checksum-modee', 'choices': ['ENABLED'],
+        'help_text': 'To retrieve the checksum, this mode must be enabled.'
+}
+
+CHECKSUM_ALGORITHM = {
+        'name': 'checksum-algorithm', 'choices': ['CRC32', 'SHA256', 'SHA1', 'CRC64NVME', 'CRC32C'],
+        'help_text': 'Indicates the algorithm used to create the checksum for the object when you use the SDK. '
+        'This header will not provide any additional functionality if you don’t use the SDK.'
+} # TODO truncated some of the doc description from low-level api until i see exactly how this affects things.
+
 TRANSFER_ARGS = [DRYRUN, QUIET, INCLUDE, EXCLUDE, ACL,
                  FOLLOW_SYMLINKS, NO_FOLLOW_SYMLINKS, NO_GUESS_MIME_TYPE,
                  SSE, SSE_C, SSE_C_KEY, SSE_KMS_KEY_ID, SSE_C_COPY_SOURCE,
@@ -489,7 +500,7 @@ TRANSFER_ARGS = [DRYRUN, QUIET, INCLUDE, EXCLUDE, ACL,
                  CONTENT_DISPOSITION, CONTENT_ENCODING, CONTENT_LANGUAGE,
                  EXPIRES, SOURCE_REGION, ONLY_SHOW_ERRORS, NO_PROGRESS,
                  PAGE_SIZE, IGNORE_GLACIER_WARNINGS, FORCE_GLACIER_TRANSFER,
-                 REQUEST_PAYER]
+                 REQUEST_PAYER, CHECKSUM_MODE, CHECKSUM_ALGORITHM]
 
 
 class S3Command(BasicCommand):
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1390,7 +1390,7 @@ class CommandParameters(object):
             }
             raise ParamValidationError(
                 f"Expected {param} parameter to be used with one of following path formats: "
-                f"{', '.join(allowed_paths)} but received {expected_usage_map[paths_type]}"
+                f"{', '.join([expected_usage_map[path] for path in allowed_paths])}. Instead, received {expected_usage_map[paths_type]}."
             )
 
     def _normalize_s3_trailing_slash(self, paths):
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -487,7 +487,7 @@ CHECKSUM_MODE = {
 }
 
 CHECKSUM_ALGORITHM = {
-        'name': 'checksum-algorithm', 'choices': ['CRC32', 'SHA256', 'SHA1', 'CRC64NVME', 'CRC32C'],
+        'name': 'checksum-algorithm', 'choices': ['CRC32', 'SHA256', 'SHA1', 'CRC32C'],
         'help_text': 'Indicates the algorithm used to create the checksum for the object when you use the SDK. '
         'This header will not provide any additional functionality if you don’t use the SDK.'
 } # TODO truncated some of the doc description from low-level api until i see exactly how this affects things.
@@ -1286,6 +1286,18 @@ class CommandParameters(object):
                 self._validate_same_underlying_s3_paths()
             if self._should_emit_validate_s3_paths_warning():
                 self._emit_validate_s3_paths_warning()
+        # If we're using the checksum-algorithm flag, the path type must be locals3
+        if params['checksum_algorithm']:
+            self._raise_if_paths_type_incorrect_for_param(
+                CHECKSUM_ALGORITHM['name'],
+                params['paths_type'],
+                'locals3')
+        # If we're using the checksum-mode flag, the path type must be s3local
+        elif params['checksum_mode']:
+            self._raise_if_paths_type_incorrect_for_param(
+                CHECKSUM_MODE['name'],
+                params['paths_type'],
+                's3local')
 
         # If the user provided local path does not exist, hard fail because
         # we know that we will not be able to upload the file.
@@ -1368,6 +1380,19 @@ class CommandParameters(object):
             raise ParamValidationError(
                 "Cannot mv a file onto itself: "
                 f"{self.parameters['src']} - {self.parameters['dest']}"
+            )
+
+    def _raise_if_paths_type_incorrect_for_param(self, param, paths_type, expected_paths_type):
+        if paths_type != expected_paths_type:
+            expected_usage_map = {
+                'locals3': '<LocalPath> <S3Uri>',
+                's3s3': '<S3Uri> <S3Uri>',
+                's3local': '<S3Uri> <LocalPath>',
+                's3': '<S3Uri>'
+            }
+            raise ParamValidationError(
+                "The %s flag is only compatible with the format: %s. You used the format: %s." %
+                (param, expected_usage_map[expected_paths_type], expected_usage_map[paths_type])
             )
 
     def _normalize_s3_trailing_slash(self, paths):
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -484,13 +484,12 @@ VALIDATE_SAME_S3_PATHS = {
 CHECKSUM_MODE = {
         'name': 'checksum-mode', 'choices': ['ENABLED'],
         'help_text': 'To retrieve the checksum, this mode must be enabled.'
-} # TODO finalize help text
+}
 
 CHECKSUM_ALGORITHM = {
         'name': 'checksum-algorithm', 'choices': ['CRC32', 'SHA256', 'SHA1', 'CRC32C'],
-        'help_text': 'Indicates the algorithm used to create the checksum for the object when you use the SDK. '
-        'This header will not provide any additional functionality if you don’t use the SDK.'
-} # TODO truncated some of the doc description from low-level api until i see exactly how this affects things.
+        'help_text': 'Indicates the algorithm used to create the checksum for the object.'
+}
 
 TRANSFER_ARGS = [DRYRUN, QUIET, INCLUDE, EXCLUDE, ACL,
                  FOLLOW_SYMLINKS, NO_FOLLOW_SYMLINKS, NO_GUESS_MIME_TYPE,
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1286,18 +1286,16 @@ class CommandParameters(object):
             if self._should_emit_validate_s3_paths_warning():
                 self._emit_validate_s3_paths_warning()
 
-        # If we're using the checksum-algorithm flag, the path type must be locals3
         if params.get('checksum_algorithm'):
             self._raise_if_paths_type_incorrect_for_param(
                 CHECKSUM_ALGORITHM['name'],
                 params['paths_type'],
-                'locals3')
-        # If we're using the checksum-mode flag, the path type must be s3local
+                ['locals3', 's3s3'])
         if params.get('checksum_mode'):
             self._raise_if_paths_type_incorrect_for_param(
                 CHECKSUM_MODE['name'],
                 params['paths_type'],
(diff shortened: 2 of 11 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -519,6 +521,7 @@ def map_upload_part_params(cls, request_params, cli_params):
         """Map CLI params to UploadPart request params"""
         cls._set_sse_c_request_params(request_params, cli_params)
         cls._set_request_payer_param(request_params, cli_params)
+        cls._set_checksum_algorithm_param(request_params, cli_params)
```
hssyoo, line 524: Something interesting I noticed is that the `map_upload_part_params` isn't actually called anywhere. Based on the [commit history](https://github.com/aws/aws-cli/commit/fc25229aeebf3f13721b307fd658a57d3c617246#diff-a06cd12835dfbf1c3436863e220d13fc24a1fab17ace4de80cce8fa4b9158409), I think this is actually dead code that wasn't cleaned up after integration with `s3transfer`.

Digging into `s3transfer`'s upload logic, it looks like `UploadSubmissionTask` handles both regular and [multipart uploads](https://github.com/aws/aws-cli/blob/9419d473fdb7f9017ff1c935da5b5cd8834ce86f/awscli/s3transfer/upload.py#L696). And the AWS CLI's `UploadRequestSubmitter` [always sets `map_put_object_params` as the request mapper](https://github.com/aws/aws-cli/blob/v2/awscli/customizations/s3/s3handler.py#L341).

You can verify this by commenting out this line and calling `aws s3 cp large-file s3://my-bucket --checksum-algorithm CRC32` and observing that `CreateMultipartUpload` and `UploadPart` are called with the checksum header.

For this PR, let's remove this line but keep the method for now. I want to avoid polluting this PR with changes not totally relevant to the task. We can create a backlog item to address potential dead code.

hssyoo, line 524: Something interesting I noticed is that the `map_upload_part_params` isn't actually called anywhere. Based on the [commit history](https://github.com/aws/aws-cli/commit/fc25229aeebf3f13721b307fd658a57d3c617246#diff-a06cd12835dfbf1c3436863e220d13fc24a1fab17ace4de80cce8fa4b9158409), I think this is actually dead code that wasn't cleaned up after integration with `s3transfer`.

Digging into `s3transfer`'s upload logic, it looks like `UploadSubmissionTask` handles both regular and [multipart uploads](https://github.com/aws/aws-cli/blob/9419d473fdb7f9017ff1c935da5b5cd8834ce86f/awscli/s3transfer/upload.py#L696). And the AWS CLI's `UploadRequestSubmitter` [always sets `map_put_object_params` as the request mapper](https://github.com/aws/aws-cli/blob/v2/awscli/customizations/s3/s3handler.py#L341).

You can verify this by commenting out this line and calling `aws s3 cp large-file s3://my-bucket --checksum-algorithm CRC32` and observing that `CreateMultipartUpload` and `UploadPart` are called with the checksum header.

For this PR, let's remove this line but keep the method for now. I want to avoid polluting this PR with changes not totally relevant to the task. We can create a backlog item to address potential dead code.

awscli/customizations/s3/subcommands.py
```diff
@@ -481,6 +481,16 @@
     )
 }
 
+CHECKSUM_MODE = {
+        'name': 'checksum-mode', 'choices': ['ENABLED'],
+        'help_text': 'To retrieve the checksum, this mode must be enabled.'
+}
+
+CHECKSUM_ALGORITHM = {
+        'name': 'checksum-algorithm', 'choices': ['CRC32', 'SHA256', 'SHA1', 'CRC32C'],
```
hssyoo, line 490: non-blocking: We should consider retrieving the choices from the service model instead of hardcoding the values here.

awscli/customizations/s3/subcommands.py
```diff
@@ -1359,6 +1382,19 @@ def _raise_if_mv_same_paths(self, src, dest):
...
+        if paths_type != expected_paths_type:
+            expected_usage_map = {
+                'locals3': '<LocalPath> <S3Uri>',
+                's3s3': '<S3Uri> <S3Uri>',
+                's3local': '<S3Uri> <LocalPath>',
+                's3': '<S3Uri>'
+            }
+            raise ParamValidationError(
+                "The %s flag is only compatible with the format: %s. You used the format: %s." %
+                (param, expected_usage_map[expected_paths_type], expected_usage_map[paths_type])
```
hssyoo, line 1395:
```suggestion
                f"Expected {param} parameter to be used with one of following path formats: "
                f"{', '.join(allowed_paths)} but received {expected_usage_map[paths_type]}"
```
Where `allowed_paths` is a list of the mapped paths types.

awscli/customizations/s3/subcommands.py
```diff
@@ -1359,6 +1382,19 @@ def _raise_if_mv_same_paths(self, src, dest):
                 f"{self.parameters['src']} - {self.parameters['dest']}"
             )
 
+    def _raise_if_paths_type_incorrect_for_param(self, param, paths_type, expected_paths_type):
+        if paths_type != expected_paths_type:
```
hssyoo, line 1386: Instead of `expected_paths_type` let's make the arg a list and call it  `allowed_paths_types`. And here we can make the check `if paths_type not in allowed_paths_types`.

awscli/customizations/s3/subcommands.py
```diff
@@ -1276,6 +1286,19 @@ def _validate_path_args(self):
             if self._should_emit_validate_s3_paths_warning():
                 self._emit_validate_s3_paths_warning()
 
+        # If we're using the checksum-algorithm flag, the path type must be locals3
```
hssyoo, line 1289: We can remove this comment. It's just describing what the code is doing and doesn't have any explanatory benefits.

awscli/customizations/s3/subcommands.py
```diff
@@ -1276,6 +1286,19 @@ def _validate_path_args(self):
...
+        if params.get('checksum_algorithm'):
+            self._raise_if_paths_type_incorrect_for_param(
+                CHECKSUM_ALGORITHM['name'],
+                params['paths_type'],
+                'locals3')
+        # If we're using the checksum-mode flag, the path type must be s3local
```
hssyoo, line 1295: Same here, we can remove.

awscli/customizations/s3/subcommands.py
```diff
@@ -1276,6 +1286,19 @@ def _validate_path_args(self):
             if self._should_emit_validate_s3_paths_warning():
                 self._emit_validate_s3_paths_warning()
 
+        # If we're using the checksum-algorithm flag, the path type must be locals3
+        if params.get('checksum_algorithm'):
+            self._raise_if_paths_type_incorrect_for_param(
+                CHECKSUM_ALGORITHM['name'],
+                params['paths_type'],
+                'locals3')
```
hssyoo, line 1294: I think we should also allow `s3s3`. See my top-level comment for more info.

tests/unit/customizations/s3/test_utils.py
```diff
@@ -653,6 +653,33 @@ def test_upload_part_copy(self):
              'SSECustomerKey': 'my-sse-c-key'})
 
 
+class TestRequestParamsMapperFlexibleChecksumsUploads:
```
hssyoo, line 656:
```suggestion
class TestRequestParamsMapperChecksumAlgorithm:
```

tests/unit/customizations/s3/test_utils.py
```diff
@@ -653,6 +653,33 @@ def test_upload_part_copy(self):
...
+        RequestParamsMapper.map_put_object_params(request_params, cli_params)
+        assert request_params == {'ChecksumAlgorithm': 'CRC32'}
+
+    def test_upload_part(self, cli_params):
+        request_params = {}
+        RequestParamsMapper.map_upload_part_params(request_params, cli_params)
+        assert request_params == {'ChecksumAlgorithm': 'CRC32'}
+
+
+class TestRequestParamsMapperFlexibleChecksumsDownloads:
```
hssyoo, line 672:
```suggestion
class TestRequestParamsMapperChecksumMode:
```

tests/functional/s3/test_cp_command.py
```diff
@@ -760,6 +760,72 @@ def test_cp_with_error_and_warning_permissions(self):
         self.assertIn('upload failed', stderr)
         self.assertIn('warning: File has an invalid timestamp.', stderr)
 
+    def test_upload_with_flexible_checksum_crc32(self):
```
hssyoo, line 763: Let's avoid using the term "flexible checksum". Instead, refer to the parameter names `checksum_algorithm` and `checksum_mode`. This is mainly to clarify what exactly the test is for.

hssyoo (changes requested):
Looks like a solid start.

My main feedback here is that `--checksum-algorithm` should be supported for copies (ie `s3s3`). I can definitely see strong use cases for wanting to copy the object checksums between different S3 paths (or even generating different checksums). In order to support this, we're going to need to port [this PR](https://github.com/boto/s3transfer/pull/242) for the vended `s3transfer` package. Let's make this port a separate PR to reduce noise in this one.

We should also prefer f-strings for string formatting in all new code. ie `f"var: {my_var}"` instead of `"var: %s" % myvar`.

awscli/customizations/s3/subcommands.py
aemous: Thanks for the feedback; could be worth looking into in a future PR.

tests/functional/s3/test_sync_command.py
```diff
@@ -353,6 +353,64 @@ def test_with_copy_props(self):
             ]
         )
 
+    def test_upload_with_checksum_algorithm_sha1(self):
```
hssyoo, line 356: Let's also add copy (`s3s3`) tests here.

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -400,6 +400,48 @@ def test_validate_no_streaming_paths(self):
         cmd_params.add_paths(paths)
         self.assertFalse(cmd_params.parameters['is_stream'])
 
+    def test_validate_checksum_algorithm_download_error(self):
+        paths = ['s3://bucket/key', self.file_creator.rootdir]
+        parameters = {'checksum_algorithm': 'CRC32'}
+        cmd_params = CommandParameters('cp', parameters, '')
+        with self.assertRaises(ParamValidationError):
```
hssyoo, line 407: `ParamValidationError` is used enough throughout the codebase that when testing we should assert that the error message is what we expect (and not propagating some other `ParamValidationError`).

tests/unit/customizations/s3/test_utils.py
```diff
@@ -653,6 +653,28 @@ def test_upload_part_copy(self):
              'SSECustomerKey': 'my-sse-c-key'})
 
 
+class TestRequestParamsMapperChecksumAlgorithm:
```
hssyoo, line 656: Let's also add test cases for:
* Copy object
* Empty `cli_params` and assert `ChecksumAlgorithm` isn't present

tests/unit/customizations/s3/test_utils.py
```diff
@@ -653,6 +653,28 @@ def test_upload_part_copy(self):
...
+    def cli_params(self):
+        return {'checksum_algorithm': 'CRC32'}
+
+    def test_put_object(self, cli_params):
+        request_params = {}
+        RequestParamsMapper.map_put_object_params(request_params, cli_params)
+        assert request_params == {'ChecksumAlgorithm': 'CRC32'}
+
+
+class TestRequestParamsMapperChecksumMode:
```
hssyoo, line 667: Let's add a test case for empty `cli_params` and assert `ChecksumAlgorithm` isn't present.

hssyoo (changes requested):

tests/functional/s3/test_cp_command.py
```diff
@@ -760,6 +760,66 @@ def test_cp_with_error_and_warning_permissions(self):
         self.assertIn('upload failed', stderr)
         self.assertIn('warning: File has an invalid timestamp.', stderr)
 
+    def test_upload_with_checksum_algorithm_crc32(self):
```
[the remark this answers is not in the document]
hssyoo: Oh, we should also be sure to add test cases for multipart uploads.

hssyoo (commented):
Playing around with it some more, I'm not sure `--checksum-mode ENABLED` provides anything useful without the `debug` flag:
```
aws s3 cp s3://hssyoo-brawn-test/README.rst ./ --checksum-mode ENABLED

download: s3://hssyoo-brawn-test/README.rst to ./README.rst
```

Compare this to the low-level `get-object`:
```
aws s3api get-object --bucket hssyoo-brawn-test --key README.rst --checksum-mode ENABLED ./README.rst

{
    ...
    "ChecksumCRC32C": "ac1Isw==",
    ...
}
```

With the current changes, the `--checksum-mode` parameter in high-level S3 commands:
* Doesn't do client-side validation of the object
* Doesn't output the checksum the user could use to validate the object

We should decide what behavior we want to support.

EDIT: Sorry, validation is automatically [supported with `botocore`](https://github.com/aws/aws-cli/blob/5f942a86a529e48aa15fa959a55a24ab605cfa19/awscli/botocore/httpchecksum.py#L372). Ignore.

tests/functional/s3/test_cp_command.py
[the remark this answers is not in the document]
aemous: Resolving in next revision

tests/functional/s3/test_sync_command.py
```diff
@@ -353,6 +353,64 @@ def test_with_copy_props(self):
...
+                'ContentLength': '100',
+                'LastModified': '00:00:00Z',
+                'ETag': 'foo-1',
+                'ChecksumSHA1': 'checksum',
+                'Body': BytesIO(b'foo')
+            }
+        ]
+        cmdline = f'{self.prefix} s3://bucket/foo {self.files.rootdir} --checksum-mode ENABLED'
+        self.run_cmd(cmdline, expected_rc=0)
+        self.assertEqual(self.operations_called[0][0].name, 'ListObjectsV2')
```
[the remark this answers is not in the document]
aemous: I ran the command manually with debug and verified that a download sync does not call GetObject. I also added a print statement to output all operations called when running this test and ListObjectV2 was the only one.

tests/functional/s3/test_sync_command.py
aemous: Good idea thanks, adding one in next revision

tests/functional/s3/test_sync_command.py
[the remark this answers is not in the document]
hssyoo: But not having `GetObject` defeats the purpose of this test since we're explicitly testing the `ChecksumMode` parameter with `sync`'s download feature. It looks like the issue here is that we have `self.head_object_response()` as the stubbed response for `ListObjectsV2` when it should be `self.list_objects_response(keys['foo-1'])` instead.

tests/unit/customizations/s3/test_subcommands.py
aemous: Good point, resolving in next revision

tests/unit/customizations/s3/test_utils.py
aemous: Good points, resolving in next iteration

tests/unit/customizations/s3/test_utils.py
aemous: Good point thank you, resolving in next revision

tests/unit/customizations/s3/test_subcommands.py
```diff
@@ -400,6 +400,42 @@ def test_validate_no_streaming_paths(self):
         cmd_params.add_paths(paths)
         self.assertFalse(cmd_params.parameters['is_stream'])
 
+    def test_validate_checksum_algorithm_download_error(self):
+        paths = ['s3://bucket/key', self.file_creator.rootdir]
+        parameters = {'checksum_algorithm': 'CRC32'}
+        cmd_params = CommandParameters('cp', parameters, '')
+        with self.assertRaises(ParamValidationError) as cm:
+            cmd_params.add_paths(paths)
+            self.assertIn('Expected checksum-algorithm parameter to be used with one of following path formats', cm.msg)
```
hssyoo, line 409: Let's also make this assertion for the tests below.

hssyoo (changes requested):

awscli/customizations/s3/subcommands.py
```diff
@@ -1359,6 +1380,19 @@ def _raise_if_mv_same_paths(self, src, dest):
...
+        if paths_type not in allowed_paths:
+            expected_usage_map = {
+                'locals3': '<LocalPath> <S3Uri>',
+                's3s3': '<S3Uri> <S3Uri>',
+                's3local': '<S3Uri> <LocalPath>',
+                's3': '<S3Uri>'
+            }
+            raise ParamValidationError(
+                f"Expected {param} parameter to be used with one of following path formats: "
+                f"{', '.join(allowed_paths)} but received {expected_usage_map[paths_type]}"
```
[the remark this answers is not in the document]
aemous: Ah, will fix in next revision, with a tiny adjustment to grammar/wording to be more clear.

.changes/next-release/feature-s3-46667.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "s3",
+  "description": "Adds support for checksums other than MD5 to high-level S3 commands."
```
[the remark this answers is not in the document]
aemous: Will include in next revision

tests/unit/customizations/s3/test_subcommands.py
aemous: Will do in next revision

hssyoo (approved):
Looks good! 🏆

PR #8935, opened 2024-09-20
[v2] S3 expires timestamp port

aemous:
*Issue #, if available:*

*Description of changes:*
- Ported https://github.com/boto/botocore/pull/3232 from botocore, which gracefully handles invalid expires timestamps from S3. 
- Update all s3api documentation that includes the string `Expires -> (timestamp)` in the `Output` section with the following changes:
  - Add a deprecation notice under the `Expires` field
  - Inject a new synthetic `ExpiresString` field
  - The above documentation changes are consistent with [botocore's documentation ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_object.html)changes as of writing.

*Testing:*
In addition to passing existing and newly added tests for this functionality, I did manual testing to verify the new field is included in the response body of the GetObject operation. Below is an example manual test using the command `aws s3api get-object --bucket bucketname --key filename outfile`:

```
}
    "AcceptRanges": "bytes",
    "LastModified": "2024-09-20T18:56:59+00:00",
    "ContentLength": 23230,
    "ETag": "\"207d91b3fd7953aac7b28b8ddf882953\"",
    "ContentType": "binary/octet-stream",
    "Expires": "2025-09-20T07:28:00+00:00",
    "ServerSideEncryption": "AES256",
    "Metadata": {},
    "ExpiresString": "Sat, 20 Sep 2025 07:28:00 GMT"
}
```

This can be compared to the output of the exact same command using the current latest release:

```
{
    "AcceptRanges": "bytes",
    "LastModified": "2024-09-20T18:56:59+00:00",
    "ContentLength": 23230,
    "ETag": "\"207d91b3fd7953aac7b28b8ddf882953\"",
    "ContentType": "binary/octet-stream",
    "Expires": "2025-09-20T07:28:00+00:00",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}
```

*Screenshots:*
#### Example of the injected deprecation notice on the `Expires` output field, and the newly injected synthetic `ExpiresString` field
image


awscli/customizations/s3events.py
```diff
@@ -57,6 +65,41 @@ def replace_event_stream_docs(help_command, **kwargs):
...
+    current = ''
+    while current != 'Expires -> (timestamp)':
+        try:
+            current = doc.pop_write()
+            popped.append(current)
+        except IndexError:
+            # Do nothing if Expires is not in the modeled response
+            break
+    # Put back the expires field and description
+    doc.push_write(popped.pop())
```
aemous, line 80: This requires the condition that each entry (including description) in the output syntax is made up of exactly four components on the stack. Not the cleanest solution in my opinion

awscli/customizations/s3events.py
```diff
@@ -57,6 +65,46 @@ def replace_event_stream_docs(help_command, **kwargs):
...
+            popped.append(current)
+        except IndexError:
+            # Do nothing if Expires is not in the modeled response
+            caught_index_error = True
+            break
+
+    # Only inject the deprecation notice if the Expires timestamp arg was found in the document
+    if not caught_index_error:
+        # Put back the expires field and description
+        doc.push_write(popped.pop())
```
aemous, line 85: This depends on the precondition that each field of the output body is made up of exactly 4 items on the documentation stack. This appears true as of writing. 

I believe this feels convoluted, and I prefer botocore's use of the DocumentStructure class, which makes it easier to inject notices like this. It would take non-trivial work to migrate over to use DocumentStructure, and it might not be worth the effort.

awscli/customizations/s3events.py
```diff
@@ -57,6 +65,46 @@ def replace_event_stream_docs(help_command, **kwargs):
...
+        doc.push_write(popped.pop())
+        doc.push_write(popped.pop())
+        doc.push_write(popped.pop())
+
+        doc.push_write('\n\n\n' + doc.style.spaces())
+        doc.push_write('.. note::')
```
aemous, line 91: I chose note mostly to stay consistent with the current [botocore docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_object.html). Another alternative considered is Warning, which may semantically be more fitting (?)

ashovlin (approved):

awscli/customizations/s3events.py
```diff
@@ -57,6 +65,46 @@ def replace_event_stream_docs(help_command, **kwargs):
               "object content is written to the specified outfile.\n")
 
 
+def document_expires_string(help_command, **kwargs):
```
hssyoo, line 68: There doesn't seem to be a cleaner way to handle this with the existing interfaces. [Writes are just stored in a list](https://github.com/aws/aws-cli/blob/c6305d03c636481d39016b678ed9b3405fc3f0b7/awscli/botocore/docs/bcdoc/restdoc.py#L32) in `ReSTDocument` objects, so I wonder if it's worth implementing a couple extra methods to help simplify injecting lines.
```
# Only copying/including relevant details
class ReSTDocument(object):
    def __init__(self, target='man'):
        self._writes = []

    def find_write(self, content):
        try:
            return self._writes.index(content)
        except ValueError:
            return None

    def insert_write(self, before_index, content):
        # This may need some massaging for correct formatting
        self._writes.insert(before_index, content)
```

Then in `document_expires_string` we could return early if `doc.find_write('Expires -> (timestamp)'` returns `None`. Otherwise we inject writes.

Thoughts?

hssyoo: Also, in the past, when we had to [add notes for specific types](https://github.com/aws/aws-cli/blob/c6305d03c636481d39016b678ed9b3405fc3f0b7/awscli/clidocs.py#L291-L308), we implemented them in `CLIDocumentEventHandler` so the notes would get written at the same time as the rest of the docs and not after. I'm not sure if we want to include S3-specific logic here but it's another option.

aemous: In reply to the suggestions for find and insert writes, I prefer your suggestions and will implement them in the next revision.

In terms of CLIDocumentEventHandler, I think it'd make more sense to have it under the customizations directory since it's service-specific. But I think a generic `inject_note_on_output_field(service_name, field_name, field_type, note_description)` function and `inject_output_field_after(service_name, prior_field_name, prior_field_type, synthetic_field_name, synthetic_field_type, synthetic_field_desc)` function could be useful, and we could call it from s3events to make that code a lot cleaner and simpler.

The first function is self-documenting in its name, but the second function I'd imagine would inject a new field named `synthetic_field_name` with type `synthetic_field_type` with description `synthetic_field_desc`. Specifically, it would inject this field right after the field with name `prior_field_name` and type `prior_field_type` if such field exists under service `service_name`. If no such field is found under such service, the synthetic field will not be injected. 

But maybe this is too specific and we should wait for at least a couple duplications of this use-case before abstracting it in this way?

hssyoo: > But maybe this is too specific and we should wait for at least a couple duplications of this use-case before abstracting it in this way?

Yeah let's go with a simpler solution for now and we can revisit it if we find ourselves doing this more often.

hssyoo (changes requested):

awscli/bcdoc/restdoc.py
```diff
@@ -67,6 +67,23 @@ def push_write(self, s):
         """
         self._writes.append(s)
 
+    def find_last_write(self, content):
+        """
+        Returns the index of the last occurrence of the content argument
+        in the stack, or raises a ValueError if content is not on the stack.
+        """
+        try:
+            return len(self._writes) - self._writes[::-1].index(content)
```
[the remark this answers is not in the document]
aemous: Fixing in next revision thank you

awscli/bcdoc/restdoc.py
```diff
@@ -67,6 +67,23 @@ def push_write(self, s):
         """
         self._writes.append(s)
 
+    def find_last_write(self, content):
+        """
+        Returns the index of the last occurrence of the content argument
+        in the stack, or raises a ValueError if content is not on the stack.
```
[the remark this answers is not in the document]
aemous: Fixing in next revision

tests/functional/s3api/test_get_object.py
```diff
@@ -85,6 +85,24 @@ def test_streaming_output_arg_with_error_response(self):
         self.assert_params_for_cmd(
             cmdline, {'Bucket': 'mybucket', 'Key': 'mykey'})
 
+    def test_invalid_expires_timestamp(self):
```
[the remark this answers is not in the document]
aemous: Oh, I was under the presumption that this was mocking the response from botocore rather than the service. It was just a sanity check to make sure we don't raise an error in the case botocore does what it should in the case of an invalid Expires field in the response body from S3. Not sure the test has much value anyways, so I'll remove it for the next revisions.

awscli/bcdoc/restdoc.py
```diff
@@ -67,6 +67,23 @@ def push_write(self, s):
         """
         self._writes.append(s)
 
+    def find_last_write(self, content):
```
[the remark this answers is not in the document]
aemous: Fixed in latest revision

tests/unit/bcdoc/test_restdoc.py
```diff
@@ -0,0 +1,53 @@
...
+# WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+import pytest
+
+from awscli.bcdoc.restdoc import ReSTDocument
+
+
+class TestReSTDocument:
```
hssyoo, line 27: Thanks for adding these tests. Can we actually add them to the existing `TestReSTDocument` [tests here](https://github.com/aws/aws-cli/blob/07dee1855177584cd82d9016254d46b046d095ab/tests/unit/bcdoc/test_document.py#L27)? I want to avoid splitting tests for this interface.

aemous: Fixed in latest revision

hssyoo (approved):
Let's rerun failed CI before merging but otherwise looks good! 🏆

PR #8940, opened 2024-09-23
[v2] S3Transfer handle checksums on copy

aemous:
*Issue #, if available:*

*Description of changes:*
- Ported over https://github.com/boto/s3transfer/pull/242/files to handle checksums on copy to our vendored S3Transfer.

*Descriptions of tests:*
- Tested changes using the new unit/functional tests.
- Tested the changes manually using aws s3 cp command via the `s3-high-level-checksums` branch (as of writing).

### DIFF
--- a/awscli/s3transfer/copies.py
+++ b/awscli/s3transfer/copies.py
@@ -226,6 +226,8 @@ class CopySubmissionTask(SubmissionTask):
                 num_parts,
                 transfer_future.meta.size,
             )
+            # Get the checksum algorithm of the multipart request.
+            checksum_algorithm = call_args.extra_args.get("ChecksumAlgorithm")
             part_futures.append(
                 self._transfer_coordinator.submit(
                     request_executor,
@@ -240,6 +242,7 @@ class CopySubmissionTask(SubmissionTask):
                             'extra_args': extra_part_args,
                             'callbacks': progress_callbacks,
                             'size': size,
+                            'checksum_algorithm': checksum_algorithm,
                         },
                         pending_main_kwargs={
                             'upload_id': create_multipart_future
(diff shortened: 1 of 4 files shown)


awscli/s3transfer/copies.py
```diff
@@ -226,6 +226,8 @@ def _submit_multipart_request(
                 num_parts,
                 transfer_future.meta.size,
             )
+            # Get the checksum algorithm of the multipart request.
+            checksum_algorithm = call_args.extra_args.get("ChecksumAlgorithm")
```
hssyoo, line 230: non-blocking: In the future, we should consider mapping the results of [`HeadObject`](https://github.com/aws/aws-cli/blob/9419d473fdb7f9017ff1c935da5b5cd8834ce86f/awscli/s3transfer/copies.py#L109-L113) and using that to automatically pass `ChecksumAlgorithm`. I'm mainly suggesting this because the low-level [`CopyObject` API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html) states:
> When you copy an object, if the source object has a checksum, that checksum value will be copied to the new object by default.

We can potentially add it in a separate PR since we want to stay aligned with the `s3transfer` package. It's also not totally straightforward to add it to [this mapping](https://github.com/aws/aws-cli/blob/9419d473fdb7f9017ff1c935da5b5cd8834ce86f/awscli/s3transfer/copies.py#L33) since `HeadObject` will return something like `ChecksumCRC32` instead of `ChecksumAlgorithm`.

aemous: Thanks for the feedback. Agreed with your points for potentially adding it in a separate PR.

hssyoo (approved):

PR #8992, opened 2024-10-18
[V2] Improve performance when parsing invalid shorthand syntax

ashovlin:
*Description of changes:* This avoids exponential backtracking when parsing shorthand syntax that:
1. Starts with either `'` or `"`, but is not closed
2. Contains a lot of `\`

Now, these will finish quickly and throw the expected `ShorthandParseError`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #8999, opened 2024-10-21
[V1] Improve performance when parsing invalid shorthand syntax

ashovlin:
*Issue #, if available:* V1 port of #8992

*Description of changes:* This avoids exponential backtracking when parsing shorthand syntax that:
1. Starts with either `'` or `"`, but is not closed
2. Contains a lot of `\`

Now, these will finish quickly and throw the expected `ShorthandParseError`.

### DIFF
--- a/.changes/next-release/bugfix-shorthand-99230.json
+++ b/.changes/next-release/bugfix-shorthand-99230.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "shorthand",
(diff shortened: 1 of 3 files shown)


hssyoo (approved):

aemous (approved):
Non-blocking: Not sure if I philosophically agree with failing unit tests based on execution time. In the future, maybe we could consider having this be a performance benchmark instead.

Issue #8013, opened 2023-06-29
Support fileb:// for nested objects


NukaCody:
### Describe the feature

Looks like fileb:// works fine for simple arguments "--file fileb://some-local-file.txt" but it does not work for items that have an embedded file options. For example:

```sh
aws rolesanywhere create-trust-anchor --name MyCompanyTrustAnchorImported \
  --source sourceData={x509CertificateData=fileb://root-ca.crt},sourceType="CERTIFICATE_BUNDLE"  \
  --enabled
```

Does not work as intended. After running with --debug, I see the request gets wrapped as a string "fileb://root-ca.crt" instead of it filling in with the file contents

### Use Case

PEM encoded certificates can get quite long and it's not practical to copy it manually at the command line as a "string"

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

aws-cli/2.12.2 Python/3.11.4 Linux/4.14.255-314-253.539.amzn2.x86_64 exec-env/CloudShell exe/x86_64.amzn.2 prompt/off

### Environment details (OS name and version, etc.)

Amazon Linux 2 / Cloudshell

aemous:
Feature is implemented. The usage of the command in the request would be

```
aws rolesanywhere create-trust-anchor --name MyCompanyTrustAnchorImported \
  --source sourceData={x509CertificateData@=fileb://root-ca.crt},sourceType="CERTIFICATE_BUNDLE"  \
  --enabled
```
The only difference is the need for the `@=` operator.

aemous:
Reopening as this issue was prematurely closed before the feature was released. It is currently merged. I will close the issue after our next release.

aemous:
Feature is implemented. The usage of the command in the request would be

```
aws rolesanywhere create-trust-anchor --name MyCompanyTrustAnchorImported \
  --source sourceData={x509CertificateData@=file://root-ca.crt},sourceType="CERTIFICATE_BUNDLE"  \
  --enabled
```

The only difference from the request is the need for the `@=` operator. Similarly, `fileb://` can be used the same way. However, for this particular API `create-trust-anchor`, `source.sourceData. x509CertificateData` must be of type string, not bytes.

PR #9012, opened 2024-10-25
[V2] Enable support for loading files into nested parameters

aemous:
*Related Issues:*
- https://github.com/aws/aws-cli/issues/8013

*Description of changes:*
- Enable support for loading files into nested parameters via usage of `file://` or `fileb://` prefix.

*Description of tests:*
- Added unit test cases to ensure the parser works against the expected grammar updates.
- Added a test suite to verify that the parser correctly loads contents from local files as expected.
- Ran and passed all existing and new tests.
- Manually tested the feature via self-crafted input.

*Usage Example:*
For example, suppose there is a file named `tagval.txt` located at `path/to/tagval.txt` on the host machine with contents `TagValue`.

` 
aws codecommit tag-resource --resource-arn <RESOURCE_ARN> --tags TagKey@=file://path/to/tagval.txt
`

The above command will parse the tags parameter as `"tags": {"TagKey": "TagValue"}}`. Notice the need for the assignment operator `@=`. This operator is required to prevent breaking existing customers that may be using a nested parameter value that happens to start with `file://` already.


awscli/shorthand.py
```diff
@@ -91,10 +94,12 @@ def __init__(self, value, expected, actual, index):
         super(ShorthandParseSyntaxError, self).__init__(msg)
 
     def _construct_msg(self):
+        expected_txt = ' or '.join(self.expected) \
```
aemous, line 97: For reviewer, do we suspect this could potentially be a breaking change? We are modifying the text of an exception that gets thrown on incorrect shorthand syntax input. 

This felt like a good solution because there are now cases in which we must expect one of multiple substrings while parsing shorthand syntax. For example, after parsing the key of a key-value pair, the parser now expects either '=' or '@='. If the user instead supplied neither, I think it is useful to enumerate all of the correct possible operators.

awscli/shorthand.py
```diff
@@ -332,17 +361,51 @@ def _consume_quoted(self, regex, escaped_char=None):
...
+                (paramfile := get_paramfile(val, LOCAL_PREFIX_MAP)) is not None):
+            return paramfile
+        if parent_key is not None and not resolve_paramfiles:
+            self._print_file_warnings_if_prefixed(parent_key, val)
+        return val
+
+    def _print_file_warnings_if_prefixed(self, key, val):
+        for prefix in LOCAL_PREFIX_MAP.keys():
+            if val.startswith(prefix):
+                _LOGGER.warning(f'Usage of the {prefix} prefix was detected '
```
aemous, line 403: For reviewer, initially I planned to use `uni_print` instead of `logging` for visibility to the user: I wanted it to print to the command line even without the `--debug` flag. However, my concern is that there is no theoretical upper bound on the number of times this gets printed for a single command. It would print once per `file://` usage without the `@=` operator. This could spam the command line if a list of strings prefixed with `file://` is supplied. 

For this reason, I chose to log the warning only in debug mode. Any feedback on this?

hssyoo (commented):
I only had time to review the implementation for parsing the `@=` operator. I should be able to take a look at the implementation for quoted keys, tests, and the questions you had later this week.

Moving forward, we should try to cut separate PRs targeting a feature branch when implementing multiple features. IMO, support for shorthand paramfiles and quoted keys are related, but distinct features.

awscli/shorthand.py
```diff
@@ -191,14 +196,26 @@ def _parameter(self):
...
         key = self._key()
-        self._expect('=', consume_whitespace=True)
-        values = self._values()
+        assignment_strings = ['=', _FILE_ASSIGNMENT]
+        resolve_paramfiles = False
+        assignment_op = self._expect_strings(assignment_strings, consume_whitespace=True)
+        if assignment_op == _FILE_ASSIGNMENT:
+            resolve_paramfiles = True
+        values = self._values(key, resolve_paramfiles)
         return key, values
```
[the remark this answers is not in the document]
aemous: I agree with most of these changes. I have one concern with the second code block. 

I'm currently printing warnings that contains the key and file prefix in the warning:

```
log.warning(f'Usage of the {prefix} prefix was detected '
                f'without the file assignment operator in parameter {key}. '
                f'To load nested parameters from a file, you must use the file '
                f'assignment operator \'{_FILE_ASSIGNMENT}\'. For example, '
                f'{key}{_FILE_ASSIGNMENT}<...>.')
```
To address the file prefix in the warning, I could return the prefix that was found in `_should_parse_as_paramfile`. But I'd still have to thread the key down to the lower-level methods, right? Would you prefer I make it an instance variable similar to your resolve_paramfiles suggestion?

tests/unit/test_shorthand.py
```diff
@@ -184,6 +202,63 @@ def test_parse(data, expected):
...
     assert actual == expected
 
+class TestShorthandParser:
+    @pytest.fixture()
+    def files(self):
+        files = FileCreator()
+        yield files
+        files.remove_all()
+
+    def test_paramfile(self, files):
```
hssyoo, line 212: I noticed a lot of these tests have the same structure. Let's use `pytest.mark.parametrize` to clean it up a bit. You can see an [example here](https://github.com/aws/aws-cli/blob/88e1f9c6697e118652787caf3194c6464c9c8068/tests/unit/test_shorthand.py#L197).

hssyoo (changes requested):

awscli/paramfile.py
```diff
@@ -15,8 +15,7 @@
 import copy
 
 from awscli.compat import compat_open
-from awscli.argprocess import ParamError
-
+from awscli import argprocess
```
aemous, line 18: For reviewer, this change was made to avoid a circular import error.

tests/unit/test_shorthand.py
```diff
@@ -184,6 +202,61 @@ def test_parse(data, expected):
...
     assert actual == expected
 
+class TestShorthandParserParamFile:
+    @pytest.fixture()
+    def files(self):
+        files = FileCreator()
+        yield files
+        files.remove_all()
+
+    @pytest.mark.parametrize(
```
hssyoo, line 212: I was thinking we could generalize the following tests:
* `test_paramfile`
* `test_file_assignment_no_file`
* `test_file_param_without_file_assignment`

by parameterizing the file contents, the input to the `parse` method, and the expected result.

hssyoo (changes requested):

hssyoo (changes requested):

awscli/shorthand.py
```diff
@@ -342,7 +359,13 @@ def _second_value(self):
             return self._double_quoted_value()
         else:
             consumed = self._must_consume_regex(self._SECOND_VALUE)
-            return consumed.replace('\\,', ',').rstrip()
+            processed = consumed.replace('\\,', ',').rstrip()
+            return self._should_resolve_paramfiles(processed) if self._resolve_paramfiles else processed
+
+    def _should_resolve_paramfiles(self, val):
```
hssyoo, line 365: I meant the boolean should be renamed. `_resolve_paramfiles` -> `_should_resolve_paramfiles`

This one should remain `_resolve_paramfile`

tests/unit/test_shorthand.py
```diff
@@ -184,6 +204,38 @@ def test_parse(data, expected):
...
+    def test_paramfile(self, mock_compat_open, file_contents, data, expected):
+        mock_compat_open.return_value.__enter__.return_value.read.return_value = file_contents
+        result = shorthand.ShorthandParser().parse(data)
+        assert result == expected
+
+    @patch('awscli.paramfile.compat_open')
+    def test_paramfile_list(self, mock_compat_open):
+        f1_contents = 'file-contents123'
+        f2_contents = 'contents2'
+        mock_compat_open.return_value.__enter__.return_value.read.side_effect = [f1_contents, f2_contents]
```
hssyoo, line 227: We should avoid changing the return type of `f.read()`, even if we're mocking it. Instead, since this test is just making sure paramfiles can be parsed in lists, we can just stick with 1 file used twice. And then we can also just add this to the parametrized test above:
```
('file-contents123', 'Foo@=[a, file://foo1, file://foo1]', {'Foo': ['a', 'file-contents123', 'file-contents123']})
```

hssyoo (changes requested):

tests/unit/test_shorthand.py
aemous: This code is a little different since its setting the value of `side_effect`, which let's you mock subsequent calls to the function via an iterable. So this array is actually patching the first read with element 0, and the second read with element 1.

I do feel there is value in having a test with 2 different files in a list. It verifies there is no bug of using the same file for every encounter of file:// and correctly resolves each element of the list independently.

hssyoo: Ah I missed that. Okay that sounds good to me then

hssyoo (approved):

PR #9051, opened 2024-11-01
Use AWS CRT instead of cryptography for EC2 decrypt password

kdaily:
*Issue #, if available:*

NA

*Description of changes:*

Replace use of `cryptography` methods with AWS CRT to perform password decryption when calling the `aws ec2 get-password-data` command with a supplied private key file.

Testing:

- Existing test suites cover decryption methods, and they pass.
- Manually compared results of retrieving password from the AWS Console for a Windows EC2 instance to the result using  the new implementation.


hssyoo (approved):
🏆

PR #9052, opened 2024-11-04
Use randomized bucket names for S3 integration tests.

jonathan343:
This change updates existing S3 integration tests to more away from static names and instead use randomized bucket names consistent with other tests.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -32,6 +32,7 @@ import sys
 import tempfile
 import time
 import unittest
+import uuid
 from pprint import pformat
 from subprocess import PIPE, Popen
 from unittest import mock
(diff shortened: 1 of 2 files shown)


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -307,8 +307,9 @@ def test_mv_with_large_file(self):
                          len(file_contents.getvalue()))
 
     def test_mv_to_nonexistent_bucket(self):
+        bucket_name = random_bucket_name()
```
nateprewitt, line 310: Should this be static for the entire test runner? Otherwise, we're increasing probability of a collision. We should be able to come up with one "non-existant" name (maybe verify it?) and then run all the tests.

nateprewitt (changes requested):

tests/integration/customizations/s3/test_plugin.py
jonathan343: In the latest commit I created a `_NON_EXISTENT_BUCKET` constant which is verified in the `setup_module` function where we verify other buckets.

nateprewitt (approved):
Looks good assuming tests pass.

awscli/testutils.py
```diff
@@ -32,6 +32,7 @@
 import tempfile
 import time
 import unittest
+import uuid
```
kdaily, line 35: This doesn't seem to be used.

PR #9055, opened 2024-11-05
[v2] S3 200 errors

aemous:
*Issue #, if available:*
- https://github.com/boto/botocore/pull/3276
- https://github.com/boto/botocore/pull/3285
- https://github.com/boto/botocore/pull/3288

*Description of changes:*
- Ports botocore changes to handle S3 HTTP 200 OK errors.

*Description of tests:*
- Ran and passed all unit & functional tests included in the PRs linked above.
- Ran all existing tests to ensure these changes don't break existing correctness.
- Added 1 functional test to ensure the CLI does not break given a response from botocore in the case a 200 OK error is encountered from S3.
- Manually tested the built application via S3 commands to ensure heuristic correctness of the product.

### DIFF
--- a/awscli/botocore/handlers.py
+++ b/awscli/botocore/handlers.py
@@ -1143,6 +1143,7 @@ def _handle_200_error(operation_model, response_dict, **kwargs):
     if not _should_handle_200_error(operation_model, response_dict):
         # Operations with streaming response blobs are excluded as they
         # can't be reliably distinguished from an S3 error.
+        print(f'Should not handle this since detectes as non-200 error: {response_dict}')
         return
     if _looks_like_special_case_error(
         response_dict['status_code'], response_dict['body']
--- a/awscli/botocore/handlers.py
+++ b/awscli/botocore/handlers.py
@@ -159,7 +159,6 @@ def _looks_like_special_case_error(status_code, body):
             # 500 Service Errors and try again.
             return True
         if root.tag == 'Error':
-            print(f'RETURNING TRUE! LOOKS LIKE SPECIAL CASE')
             return True
     return False
 
@@ -1144,7 +1143,6 @@ def _handle_200_error(operation_model, response_dict, **kwargs):
     if not _should_handle_200_error(operation_model, response_dict):
         # Operations with streaming response blobs are excluded as they
         # can't be reliably distinguished from an S3 error.
-        print(f'SHOULD NOT HANDLE THIS ONE. ')
         return
     if _looks_like_special_case_error(
         response_dict['status_code'], response_dict['body']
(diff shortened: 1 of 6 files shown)


awscli/botocore/handlers.py
```diff
@@ -115,6 +115,7 @@ def escape_xml_payload(params, **kwargs):
 
 
 def check_for_200_error(response, **kwargs):
+    """This function has been deprecated, but is kept for backwards compatibility."""
```
SamRemis, line 118: This isn't necessary for the CLIv2.  We should delete this function (assuming it's no longer called).

SamRemis (dismissed):

tests/functional/s3/test_mv_command.py
```diff
@@ -267,6 +269,22 @@ def test_download_with_checksum_mode_crc32(self):
         self.assertEqual(self.operations_called[1][0].name, 'GetObject')
         self.assertEqual(self.operations_called[1][1]['ChecksumMode'], 'ENABLED')
 
+    def test_slowdown_raises_error(self):
```
aemous, line 272: For reviewer, since these changes do not add new functionality, and only impact the CLI via exception-raising in particular scenarios, I did not see the need for too many CLI-specific tests. Instead, I implemented this simple regression test to verify expected behavior in the case botocore response with its expected output in the case it receives a S3 200 OK error.

awscli/botocore/handlers.py
aemous: I verified this function is not called in our code base and removed it.

aemous: It seems it's being called by 3 pre-existing unit tests. For this reason, I think I'd prefer to leave it in. Let me know if you think removing the three unit tests and the deprecated function make sense.

nateprewitt: We should delete the tests if they're still calling this. It's a dead path and shouldn't be needed for the CLI v2 implementation.

awscli/botocore/handlers.py
```diff
@@ -135,20 +136,22 @@ def check_for_200_error(response, **kwargs):
         # trying to retrieve the response.  See Endpoint._get_response().
         return
     http_response, parsed = response
-    if _looks_like_special_case_error(http_response):
+    if _looks_like_special_case_error(
+            http_response.status_code, http_response.content
+    ):
```
nateprewitt, line 141:
```suggestion
    if _looks_like_special_case_error(
        http_response.status_code, http_response.content
    ):
```

ashovlin (approved):
We can probably rebase on https://github.com/aws/aws-cli/commit/23fe8984e5bd0da32db5ad28304a53a64942def0 to fix the CI, but the change looks good to me.

tests/functional/s3/test_mv_command.py
hssyoo: This looks like it's really testing the message format of modeled exceptions and not specifically S3 200 behavior

hssyoo (changes requested):

tests/functional/s3/test_mv_command.py
aemous: Removed the test in next revision

hssyoo (approved):

PR #9060, opened 2024-11-07
[v2] End of support for macOS 10.15

hssyoo:
Update macOS source distribution and PKG installer warnings following the end of support for macOS 10.15. For more, see the macOS support policy blog post:

https://aws.amazon.com/blogs/developer/macos-support-policy-updates-for-the-aws-cli-v2/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,9 +34,9 @@ The aws-cli package works on Python versions:
 -------
 Notices
 -------
-On 2024-06-20, support for macOS versions 10.14 and prior was dropped.
-To use up-to-date versions of AWS CLI v2, customers using macOS 10.14 or prior
-should upgrade to macOS 10.15 or later. For more information, please see
+On 2024-11-11, support for macOS version 10.15 was dropped.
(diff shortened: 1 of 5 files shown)


kdaily (approved):

README.rst
```diff
@@ -34,9 +34,9 @@ The aws-cli package works on Python versions:
 -------
 Notices
 -------
-On 2024-06-20, support for macOS versions 10.14 and prior was dropped.
-To use up-to-date versions of AWS CLI v2, customers using macOS 10.14 or prior
-should upgrade to macOS 10.15 or later. For more information, please see
+On 2024-11-11, support for macOS version 10.15 was dropped.
```
nateprewitt, line 37: high-level question; do we record the last version that had support (or the version that dropped support) anywhere in our docs? That will probably ease any searching customers have to do down the line.

hssyoo: Ah yeah that would be useful. We don't, but I'll talk to someone who can help with that.

Issue #8988, opened 2024-10-18
aws s3 cp exit code 255: botocore.awsrequest.AWSRequest object at


fishcharlie:
### Describe the bug

When running `aws s3 cp code.zip s3://my-bucket/folder/code.zip`, it exits with code 255 and prints the following:

```
<botocore.awsrequest.AWSRequest object at 0x7f367a0c9af0>
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

It to successfully upload the file.

### Current Behavior

Fails with exit code 255 and prints:

```
<botocore.awsrequest.AWSRequest object at 0x7f367a0c9af0>
```

### Reproduction Steps

Honestly I have no idea what is even going wrong. aws cli doesn't even print a good error messages. So I have no idea how to even reproduce this or what is even going wrong...

Would be more than happy to help debug further but need some guidance here since the error message is so poor.

### Additional Information/Context

```
runner@fv-az1496-34:~/work/EchoServer/EchoServer$ ls -al
total 613764
drwxr-xr-x  10 runner docker      4096 Oct 18 04:06 .
drwxr-xr-x   3 runner docker      4096 Oct 18 04:05 ..
-rw-r--r--   1 runner docker 628228556 Oct 18 04:06 code.zip


runner@fv-az1496-34:~/work/Server/Server$ aws --version
aws-cli/2.18.0 Python/3.12.6 Linux/6.5.0-1025-azure exe/x86_64.ubuntu.22
```

### CLI version used

2.18.0

### Environment details (OS name and version, etc.)

Python/3.12.6 Linux/6.5.0-1025-azure exe/x86_64.ubuntu.22

adev-code:
Hello @fishcharlie, thanks for reaching out. I have replicated the command you ran and did not get the same error. For a further look, please include the full debug logs by adding --debug to your command and redacting any sensitive information. Thanks!

HajraRafiq28:
@fishcharlie  Try to add a region while configuring the credential. It will work

dmartin-isp:
Even with a bad command, the output shouldn't be `<botocore.awsrequest.AWSRequest object at 0x7f367a0c9af0>` but some kind of error message. I've got another aws cli command that I'm tryign to use and am getting a similarly unhelpful output.

nateprewitt:
Hi @dmartin-isp, we definitely agree that's not intended output. We'll need debug logs as requested above, or a simple repro to investigate further though.

[...]

PR #9073, opened 2024-11-13
Fixed ambiguous error message raised when resolving region from IMDS

nateprewitt:
An issue raised in https://github.com/aws/aws-cli/issues/8988 points out an unhandled error in the AWS CLI region fetcher. In the base IMDS credential fetcher, this is handled by adding a debug log and turning the request into a no-op. In the CLI's region fetcher, it's treated as terminal and surfaced to the end user with an unhelpful object reference.

This was later ported to botocore in a different location (`botocore/utils.py`). I'll address that in a separate PR upstream.

### DIFF
--- a/.changes/next-release/bugfix-IMDS-70100.json
+++ b/.changes/next-release/bugfix-IMDS-70100.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "IMDS",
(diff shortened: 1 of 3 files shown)


ashovlin (approved):
👍 looks like same pattern we have https://github.com/aws/aws-cli/blob/8916d371799db87de9d5b7c804d65c0f47552087/awscli/botocore/utils.py#L530-L531

PR #9076, opened 2024-11-14
Add MRAP support for CRT s3transfer integration

nateprewitt:
This PR patches a gap with SigV4a signing for the CRT S3 transfer manager. On performance-optimized instance types [[list](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client)], the CRT will auto-enable itself for all high-level `aws s3` commands. This may lead to signature mismatches for [Multi-Region Access Points](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPointRequests.html) backed by buckets in other regions.

Customers currently encountering this issue can bypass the problem by setting `preferred_transfer_client` to `classic` ([docs](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client)). That mitigation will no longer be needed after this patch is merged.

### DIFF
--- a/awscli/s3transfer/crt.py
+++ b/awscli/s3transfer/crt.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
+import re
 import threading
 from io import BytesIO
 
@@ -41,7 +42,7 @@ from botocore import UNSIGNED
 from botocore.compat import urlsplit
 from botocore.config import Config
 from botocore.exceptions import NoCredentialsError
-from botocore.utils import is_s3express_bucket
+from botocore.utils import ArnParser, InvalidArnException, is_s3express_bucket
 
 from s3transfer.constants import MB
 from s3transfer.exceptions import TransferNotDoneError
@@ -840,7 +841,19 @@ class S3ClientArgsCreator:
                 x.title() for x in request_type.split('_')
             )
 
-        if is_s3express_bucket(call_args.bucket):
+        arn_handler = _S3ArnParamHandler()
+        if (
+            accesspoint_arn_details := arn_handler.handle_arn(call_args.bucket)
+            and accesspoint_arn_details['region'] == ""
+        ):
+            # Configure our region to `*` to propogate in `x-amz-region-set`
+            # for multi-region support in MRAP accesspoints.
+            make_request_args['signing_config'] = AwsSigningConfig(
+                algorithm=AwsSigningAlgorithm.V4_ASYMMETRIC,
+                region="*",
+            )
+            call_args.bucket = accesspoint_arn_details['resource_name']
+        elif is_s3express_bucket(call_args.bucket):
             make_request_args['signing_config'] = AwsSigningConfig(
                 algorithm=AwsSigningAlgorithm.V4_S3EXPRESS
             )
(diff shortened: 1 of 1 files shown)


awscli/s3transfer/crt.py
```diff
@@ -840,7 +841,19 @@ def _default_get_make_request_args(
                 x.title() for x in request_type.split('_')
             )
 
-        if is_s3express_bucket(call_args.bucket):
+        arn_handler = _S3ArnParamHandler()
+        if (
+            accesspoint_arn_details := arn_handler.handle_arn(call_args.bucket)
```
kdaily, line 846:
```suggestion
            (accesspoint_arn_details := arn_handler.handle_arn(call_args.bucket))
```

Without the parentheses to scope it, this failed for me with:

```
cannot access local variable 'accesspoint_arn_details' where it is not associated with a value
```

awscli/s3transfer/crt.py
```diff
@@ -840,7 +841,19 @@ def _default_get_make_request_args(
...
+        arn_handler = _S3ArnParamHandler()
+        if (
+            (accesspoint_arn_details := arn_handler.handle_arn(call_args.bucket))
+            and accesspoint_arn_details['region'] == ""
+        ):
+            # Configure our region to `*` to propogate in `x-amz-region-set`
+            # for multi-region support in MRAP accesspoints.
+            make_request_args['signing_config'] = AwsSigningConfig(
+                algorithm=AwsSigningAlgorithm.V4_ASYMMETRIC,
+                region="*",
```
waahm7, line 853: I don't know how the integration works, but are we setting the other fields correctly here?

Defaults used by S3: https://github.com/awslabs/aws-c-s3/blob/5877f40f87c77ccf2b278839995a6ee91983080f/source/s3_util.c#L272

Some parameters like service_type are required https://github.com/awslabs/aws-crt-python/blob/main/awscrt/auth.py#L620.

waahm7: Looked more into it and c-s3 will use defaults if not set.

nateprewitt: I don't believe that's how the CRT code for the S3Client works. During client creation, if a signing config isn't passed, the CRT is creating a default signing config ([docs](https://github.com/awslabs/aws-crt-python/blob/084e141e9b87aea833b53e5b3a9175320218ed57/awscrt/s3.py#L162C1-L162C111)). Right below the line you linked, we see it correctly sets the service to S3 ([ref](https://github.com/awslabs/aws-c-s3/blob/5877f40f87c77ccf2b278839995a6ee91983080f/source/s3_util.c#L276)).

We can also see in [the docs](https://github.com/awslabs/aws-crt-python/blob/084e141e9b87aea833b53e5b3a9175320218ed57/awscrt/s3.py#L362-L375) that all fields are considered optional at the request level as we've already configured during client creation. We also have this pattern from previous integrations for things like [S3 Express](https://github.com/aws/aws-cli/blob/ad791c0798f690ff29e7959ea075af295ee3ec87/awscli/s3transfer/crt.py#L843-L846) which have been functioning as expected since they launched last year.

TingDaoK: yeah, it's a bit confusing since we used the same object, in auth, that those fields are required, while in S3, which those fields are optional. 

Since in S3, we will need to modify the signing config for different type of requests. We are not expecting user to pass in the "totally right" config for signing, S3 client will just derive the right config for signing process based on what user passed in.

I have a TODO [here](https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L412), that we should probably use a different object to avoid the confusing.

PR #9078, opened 2024-11-15
Add MRAP support for CRT s3transfer integration

nateprewitt:
This PR (previously #9076) patches a gap with SigV4a signing for the CRT S3 transfer manager. On performance-optimized instance types [[list](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client)], the CRT will auto-enable itself for all high-level `aws s3` commands. This may lead to signature mismatches for [Multi-Region Access Points](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPointRequests.html) backed by buckets in other regions.

Customers currently encountering this issue can bypass the problem by setting `preferred_transfer_client` to `classic` ([docs](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#preferred-transfer-client)). That mitigation will no longer be needed after this patch is merged.

### DIFF
--- a/tests/functional/s3transfer/test_crt.py
+++ b/tests/functional/s3transfer/test_crt.py
@@ -69,6 +69,8 @@ class TestCRTTransferManager(unittest.TestCase):
         self.region = 'us-west-2'
         self.bucket = "test_bucket"
         self.s3express_bucket = 's3expressbucket--usw2-az5--x-s3'
+        self.mrap_accesspoint = 'arn:aws:s3::123456789012:accesspoint/mfzwi23gnjvgw.mrap'
+        self.mrap_bucket = 'mfzwi23gnjvgw.mrap'
         self.key = "test_key"
         self.expected_content = b'my content'
         self.expected_download_content = b'new content'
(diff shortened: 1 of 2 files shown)


tests/functional/s3transfer/test_crt.py
```diff
@@ -156,6 +160,26 @@ def _assert_exected_s3express_request(
             awscrt.auth.AwsSigningAlgorithm.V4_S3EXPRESS
         )
 
+    def _assert_exected_mrap_request(
```
kdaily, line 163: nit:

```suggestion
    def _assert_expected_mrap_request(
```

This is a copy paste from other mis-named tests, so fine to clean this up at another time.

nateprewitt: I'll do a quick follow up PR cleaning up everything so we don't lose it if one of these needs to get reverted.

kdaily:
LGTM. Tested on a `trn1.32xlarge` and that worked as expected.

kdaily (approved):

tests/functional/s3transfer/test_crt.py
nateprewitt: https://github.com/aws/aws-cli/pull/9087

PR #9099, opened 2024-11-22
[v2] Fix flaky prompt-toolkit tests

hssyoo:
The [`PromptToolkitAppRunner` test utility](https://github.com/aws/aws-cli/blob/37a410088eb035d52da38545f1c8b9330b0a95e7/tests/__init__.py#L390) only waits 2 seconds for the app to finish rendering. This causes flaky test failures because prompt-toolkit is tested using EC2 service model, which is ~3MB and can take a few seconds to load on hosts with low resources. This PR changes the test service model to CloudWatch, which is ~230KB.

Also bumps the event timeout to 5 seconds to allow more time for the app to render for input buffer tests. Unlike other flaky tests, there's not a slimmer command/model to switch to.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -387,7 +387,7 @@ class AppRunContext:
 
 
 class PromptToolkitAppRunner:
-    _EVENT_WAIT_TIMEOUT = 3
+    _EVENT_WAIT_TIMEOUT = 5
 
     def __init__(self, app, pre_run=None):
         self.app = app
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -387,7 +387,7 @@ class AppRunContext:
 
 
 class PromptToolkitAppRunner:
-    _EVENT_WAIT_TIMEOUT = 2
+    _EVENT_WAIT_TIMEOUT = 3
 
     def __init__(self, app, pre_run=None):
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Assuming we get a clean run without having to retry 🤞

nateprewitt:
The one failure seems to match a lot of the cases I've seen:

```
functional/autoprompt/test_prompttoolkit.py::TestPromptToolkitPrompterBuffer::test_input_buffer_initialization[args3-s3 ls ] - AssertionError: assert 's3 ls' == 's3 ls '
```

For whatever reason we're missing trailing spaces on some of the tests periodically. I haven't looked at this at all but 1.) is the trailing space relevant and 2.) if not can we just normalize the strings by trimming before comparison?

hssyoo:
Played around with this a bit on my fork. I timed the test durations and saw a couple instances of prompt-toolkit tests taking ~3.5 seconds. Setting the timeout to 5 seconds for now and the plan is to monitor future CI runs to see if we need to bump it higher or look into improving performance of the autoprompt feature. Not too worried about the bump causing longer test times since this is only observed in GitHub CI and it's faster than manually retrying failed builds.

PR #9109, opened 2024-11-27
Use AWS CRT instead of cryptography for Cloudfront url signing

kdaily:
*Issue #, if available:*

*Description of changes:*

Replace use of `cryptography` methods with AWS CRT methods to perform CloudFront URL  when calling the `aws cloudfront sign` [command](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudfront/sign.html) (commit fdd8ddd651b2c51e20efab79d2cb7a661f22525b).

### Testing

All unit and functional tests pass.

I manually ran the `aws cloudfront sign` command before and after the change and did a `diff` of the generated URL, which were identical. I added a test that confirms with an expected signature to be sure that the required hashing/padding is used.

### DIFF
--- a/awscli/customizations/cloudfront.py
+++ b/awscli/customizations/cloudfront.py
@@ -10,14 +10,12 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import hashlib
 import sys
 import time
 import random
 
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import hashes
-from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15
-from cryptography.hazmat.primitives.serialization import load_pem_private_key
+from awscrt.crypto import RSA, RSASignatureAlgorithm
 
 from botocore.utils import parse_to_aware_datetime
 from botocore.signers import CloudFrontSigner
@@ -258,9 +256,11 @@ class SignCommand(BasicCommand):
 
 class RSASigner(object):
     def __init__(self, private_key):
-        backend = default_backend()
         key_bytes = private_key.encode('utf8')
-        self.priv_key = load_pem_private_key(key_bytes, None, backend)
+        self.priv_key = RSA.new_private_key_from_pem_data(key_bytes)
 
     def sign(self, message):
-        return self.priv_key.sign(message, PKCS1v15(), hashes.SHA1())
+        return self.priv_key.sign(
+            RSASignatureAlgorithm.PKCS1_5_SHA1,
+            hashlib.sha1(message).digest()
+        )
--- a/tests/functional/cloudfront/test_sign.py
+++ b/tests/functional/cloudfront/test_sign.py
@@ -64,9 +64,17 @@ class TestSign(BaseAWSCommandParamsTest):
         cmdline = (
             self.prefix + '--private-key file://' + self.private_key_file +
             ' --date-less-than 2016-1-1')
+        expected_signature = (
+            "UiEmtMsInU-gXoa1O7-bTRJmZ~ocphB0ONMxyEHs2r8Y9dwzeB~DkbgzPMX3jbdb"
+            "wIwVX3f4VcY4HBLdPSkbF~D6KbUlxPw1ju8mlXeu2C436XxZdrJrrJaiEDaTpKsl"
+            "Xpn9ngaCzVfCVPfkC3a0NBWBySi5ezCG2yzb0c-djNgI1wkogwtmtZuOxAoKF1sR"
+            "TyFX9ZitUiUIl~65nkJ94s~GGwxzTf1kMi7Wdm~9rFrJpx0O7nJEBy5O578s2UHr"
+            "ejtwyedUR5BqXTkgu~A51NcjAN9LErATV7SVBYicoZ76AOfB-TKay7g6-MWCK6-T"
+            "-4Q5x6XH4yzII3JpbCmVwA__"
+        )
         expected_params = {
             'Key-Pair-Id': ['my_id'],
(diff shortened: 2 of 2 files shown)


awscli/customizations/cloudfront.py
```diff
@@ -10,14 +10,12 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import hashlib
```
aemous, line 13: Thoughts on using crt's SHA1 hashing instead? https://github.com/awslabs/aws-crt-python/blob/main/awscrt/crypto.py#L10 . I'd expect it to perform better due to native bindings

aemous (commented):
Looks good overall. Was any testing done to ensure the sign functions returns the same value for the same private key and message before/after this change? Do you think such a test would be useful?

.changes/next-release/enhancement-awscrt-51540.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "awscrt",
+  "description": "Update minimum awscrt version requirement to 0.23.1"
```
aemous, line 4: Nit: also add that we replaced cryptography with CRT for CloudFront URL signing

awscli/customizations/cloudfront.py
aemous: Though on the flip side I see we tend to use hashlib for similar hashes in the rest of our codebase.

kdaily: > Was any testing done to ensure the sign functions returns the same value for the same private key and message before/after this change?

Yes, I can update PR description to include an example.

> Do you think such a test would be useful?

We would still need to keep the dependency to `cryptography` in order to support such a test. However, there is no existing test that checks that an expected signing result occurs; indeed, changing to SHA256 hashing, the test still passes because all values are mocked. I can investigate adding a new test with an expected case.

> Thoughts on using crt's SHA1 hashing instead?

Sure, not opposed. Python's `hashing` is standard library, and I haven't investigated performance improvements. I wouldn't expect that much as the overhead for invoking a Python interpreter to run a CLI command is going to be much higher on a single invocation.

.changes/next-release/enhancement-awscrt-51540.json
kdaily: Oops! In rebasing I lost my feature changelog entry. Thanks for catching that.

awscli/customizations/cloudfront.py
kdaily: It also has a more verbose interface. Usage would look like:

```
    def sign(self, message):
        hasher = Hash.sha1_new()
        hasher.update(message)

        return self.priv_key.sign(
            RSASignatureAlgorithm.PKCS1_5_SHA1,
            hasher.digest()
        )
```

kdaily: Since we use `hashlib` elsewhere I'm going to leave that change out for now.

aemous: > We would still need to keep the dependency to cryptography in order to support such a test.

How about a comparison against a hard-coded pre-calculated hash? 

> However, there is no existing test that checks that an expected signing result occurs

My main reason for suggesting this is to ensure we did not break hash calculation by changing the underlying algorithm, and that this remains the case in the future. If that doesn't seem valuable to you I'm also okay with just a note that you manually verified this is the case.

aemous (approved):
Looks good to me. I'd like the changelog updated (as discussed) before we merge the feature branch.

.changes/next-release/enhancement-awscrt-51540.json
ashovlin: I'm imagining we may need to adjust or combine all of these when merging the feature branch later anyway.

ashovlin (approved):

tests/functional/cloudfront/test_sign.py
```diff
@@ -64,18 +64,21 @@ def test_canned_policy(self):
         cmdline = (
             self.prefix + '--private-key file://' + self.private_key_file +
             ' --date-less-than 2016-1-1')
+        expected_signature = "UiEmtMsInU-gXoa1O7-bTRJmZ~ocphB0ONMxyEHs2r8Y9dwzeB~DkbgzPMX3jbdbwIwVX3f4VcY4HBLdPSkbF~D6KbUlxPw1ju8mlXeu2C436XxZdrJrrJaiEDaTpKslXpn9ngaCzVfCVPfkC3a0NBWBySi5ezCG2yzb0c-djNgI1wkogwtmtZuOxAoKF1sRTyFX9ZitUiUIl~65nkJ94s~GGwxzTf1kMi7Wdm~9rFrJpx0O7nJEBy5O578s2UHrejtwyedUR5BqXTkgu~A51NcjAN9LErATV7SVBYicoZ76AOfB-TKay7g6-MWCK6-T-4Q5x6XH4yzII3JpbCmVwA__"
```
hssyoo, line 67: nit: Let's break this into multiple lines. Also in the the `test_custom_policy` test.

hssyoo (changes requested):
Looks good, just a small nit. Let's also make sure to fix the failing sdist tests.

tests/functional/cloudfront/test_sign.py
kdaily: Split into multiple lines.

.changes/next-release/enhancement-awscrt-51540.json
kdaily: Yeah, agree. I'll leave this out for now and will write a single feature update on the `cryptography-to-crt` branch.

kdaily:
I updated the docs requirements file to address the versions of `awscrt` and `zipp` as part of b06503523be0602ebf6caee52319204e35ebed60 to address the dependency test failures. All test suites are passing.

hssyoo (approved):

Issue #7602, opened 2023-01-19
SSO does not honor custom ca-bundle configuration


kjluedke:
### Describe the bug

Executing an sso subcommand does not honor the --ca-bundle option or ca-bundle configuration.  The value provided to the CLI is ignored in favor of the default certificate bundle.

### Expected Behavior

I expected my specified bundle to be used.

### Current Behavior

The default bundle at `C:\Program Files\Amazon\AWSCLIV2\awscli\botocore\cacert.pem` is used.

### Reproduction Steps

Execute the following command:

`aws sso login --ca-bundle C:\Users\myUser\aws-ca-oidc-bundle.pem --debug`

Observe the following line in the debug output:
`2023-01-17 09:53:20,292 - MainThread - botocore.httpsession - DEBUG - Certificate path: C:\Program Files\Amazon\AWSCLIV2\awscli\botocore\cacert.pem`

### Possible Solution

Our temporary workaround has been to replace the contents of `C:\Program Files\Amazon\AWSCLIV2\awscli\botocore\cacert.pem` with our custom bundle.

### Additional Information/Context

This defect might be related to https://github.com/aws/aws-cdk/issues/21328 

### CLI version used

aws-cli/2.9.15 Python/3.9.11

### Environment details (OS name and version, etc.)

Windows/10 exe/AMD64

tim-finnigan:
Thanks @kjluedke for reporting this. The team confirmed the issue and will need to investigate it further. We will share any updates here.

esmcelroy:
Reporting that I'm having the same issue. Version: aws-cli/2.11.0 Python/3.11.2 Windows/10 exe/AMD64
Pulling the bottom-most traceback after running `aws configure sso --no-verify-ssl --debug`:
```
Traceback (most recent call last):
  File "awscli\clidriver.py", line 460, in main
  File "awscli\customizations\commands.py", line 151, in __call__
  File "awscli\customizations\commands.py", line 205, in __call__
  File "awscli\customizations\configure\sso.py", line 516, in _run_main
  File "awscli\customizations\sso\utils.py", line 72, in do_sso_login
  File "awscli\botocore\utils.py", line 3052, in fetch_token
  File "awscli\botocore\utils.py", line 3037, in _token
  File "awscli\botocore\utils.py", line 2947, in _poll_for_token
  File "awscli\botocore\utils.py", line 2918, in _registration
  File "awscli\botocore\utils.py", line 2875, in _register_client
[...]

PR #9112, opened 2024-11-29
Support the `--ca-bundle` option on SSO commands

ashovlin:
*Issue #, if available:* #7602

*Description of changes:*
Before, when constructing the SSO-OIDC client used inside the `aws sso login`, `aws sso logout`, and `aws configure sso` commands, we were not setting the `verify` parameter. These commands did not honor the global `--ca-bundle` and `--no-verify-ssl` options.

Now, we plumbed through `parsed_globals` to set `verify` appropriately.

### DIFF
--- a/.changes/next-release/bugfix-sso-68327.json
+++ b/.changes/next-release/bugfix-sso-68327.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``sso``",
+  "description": "Support the ``--ca-bundle`` option on SSO commands."
+}
--- a/.changes/next-release/bugfix-sso-68327.json
+++ b/.changes/next-release/bugfix-sso-68327.json
@@ -1,5 +1,5 @@
 {
   "type": "bugfix",
   "category": "``sso``",
-  "description": "Support the ``--ca-bundle`` option on SSO commands."
+  "description": "Support the ``--ca-bundle`` and ``--no-verify-ssl`` options on SSO commands."
 }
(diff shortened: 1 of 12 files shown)


.changes/next-release/bugfix-sso-68327.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``sso``",
+  "description": "Support the ``--ca-bundle`` option on SSO commands."
```
aemous, line 4: Can we also add that we support the ``--no-verify-ssl`` option now?

aemous (changes requested):
Looks good overall, can we also get a couple corresponding tests that include the ``--no-verify-ssl`` arg on the sso commands?

TheerapatSrisombat (commented):
`import numpy as np

# กำหนดพิกัดเริ่มต้น
x = 19.3476535
y = 100.533892

# สร้างเมทริกซ์เพื่อแทนค่าต่างๆ
matrix = np.array([[x, y], [0, 1], [2, 3]])

# ตัวอย่างการคำนวณง่ายๆ
result = np.sum(matrix)

print(result)`

aemous (approved):

aemous:
Looks good to me

.changes/next-release/bugfix-sso-68327.json
ashovlin: Added via https://github.com/aws/aws-cli/pull/9112/commits/7ea5e364410bc8e4c1ed63d7e87bf7f566b66816 (changelog message too).

hssyoo (approved):

PR #9123, opened 2024-12-05
Add precommit for linting and formatting

kdaily:
*Description of changes:*

First in a series of commits to lint and format the `v2` branch of this repository.

Adds a precommit hook configuration file and enables whitespace linting, `isort` for imports, and `ruff` to lint and format.  It adds configuration to `pyproject.toml` for `ruff` as well with the same configuration as the `develop` branch, with the exception of ignoring unused imports (`F401`). This exception will be dropped at a later date.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,29 @@
+exclude: "\
+    ^(\
+    .changes|\
+    .github|\
+    awscli/examples|\
+    awscli/topics|\
+    awscli/botocore|\
+    awscli/s3transfer|\
+    awscli/doc|\
+    exe/assets|\
+    tests/functional/cloudformation/deploy_templates/booleans/input.yml|\
+    tests/functional/cloudformation/deploy_templates/nested-tag/input.yml|\
+    tests/|\
+    CHANGELOG.rst|\
+    configure\
+    )"
+repos:
+  - repo: 'https://github.com/pre-commit/pre-commit-hooks'
+    rev: v4.5.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.8
+    hooks:
+      - id: ruff
+        args: [ --fix ]
+      - id: ruff-format
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -134,3 +134,64 @@ markers = [
 
 [tool.black]
 line-length = 80
+
+[tool.ruff]
+exclude = [
+    ".bzr",
+    ".direnv",
+    ".eggs",
+    ".git",
+    ".git-rewrite",
+    ".hg",
+    ".ipynb_checkpoints",
+    ".mypy_cache",
+    ".nox",
+    ".pants.d",
+    ".pyenv",
+    ".pytest_cache",
+    ".pytype",
+    ".ruff_cache",
+    ".svn",
+    ".tox",
+    ".venv",
+    ".vscode",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
+
+force-exclude = true
+
+# Format same as Black.
+line-length = 79
+indent-width = 4
+
+[tool.ruff.lint]
+# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
+# Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
+# McCabe complexity (`C901`) by default.
+select = ["E4", "E7", "E9", "F", "I", "UP"]
+ignore = ["F401"]
+
+# Allow fix for all enabled rules (when `--fix`) is provided.
+fixable = ["ALL"]
+unfixable = []
+
+# Allow unused variables when underscore-prefixed.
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
+exclude: "\
+    ^(\
+    .changes|\
```
aemous, line 3: Non-blocking: should `awscli/data`, be on this list?

aemous (approved):
Looks good to me. Left one non-blocking comment to consider.

.pre-commit-config.yaml
kdaily: I think we should leave that directory in. There are only JSON files in it, and they should be subject to the end of file fixer check; the rest of the checks aren't applicable anyways.

Indeed, running the basic linting catches that one file does not have a trailing newline:

https://github.com/aws/aws-cli/pull/9157/files#diff-031ec82ae188f9206c76aa4ae0fc601a34c9af77e340adba262922c9f4b7012e

nateprewitt: Does the linter actively change these? If they're getting re-updated with each release, we should avoid them.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - repo: 'https://github.com/pre-commit/pre-commit-hooks'
+    rev: v4.5.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: 'https://github.com/PyCQA/isort'
+    rev: 5.12.0
+    hooks:
+      - id: isort
```
nateprewitt, line 27: I don't think we actually need this for CLI v2. It's included for backwards compat in botocore imports. IMO we should just delete those here.

pyproject.toml
```diff
@@ -134,3 +134,70 @@ markers = [
...
+    "node_modules",
+    "site-packages",
+    "venv",
+]
+
+# Format same as Black.
+line-length = 79
+indent-width = 4
+
+target-version = "py38"
```
nateprewitt, line 178: Is this still right? Are we waiting until botocore raises pins? I think the CLI v2 can be a little more aggressive.

.pre-commit-config.yaml
kdaily: I'm not sure I understand what you're suggesting here. `isort` makes changes in the CLI v2 codebase.

pyproject.toml
kdaily: I assumed since our minimum version is Python 3.8, thats what we would have here. After a closer reading of the `ruff` docs, it would effectively be set to `py38` because we have this in our `pyproject.toml` too:

https://github.com/aws/aws-cli/blob/d4619a326382e52c9f0faeb29648df2b9a1489ed/pyproject.toml#L16

From the [docs](https://docs.astral.sh/ruff/settings/#target-version):

> If you're already using a pyproject.toml file, we recommend project.requires-python instead, as it's based on Python packaging standards, and will be respected by other tools. For example, Ruff treats the following as identical to target-version = "py38":
> 
>
> ```
> [project]
> requires-python = ">=3.8"
> ```

.pre-commit-config.yaml
nateprewitt: `ruff` has the same functionality as `isort` built in, you just add the `I` flag in the config.

The only reason we still have isort in botocore is because ruff does cleanup on all unused imports. There should be a full explanation in the original PR. Our ideal end state doesn't use `isort` at all.

pyproject.toml
```diff
@@ -134,3 +134,70 @@ markers = [
...
+
+[tool.ruff.lint]
+# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
+# Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
+# McCabe complexity (`C901`) by default.
+select = ["E4", "E7", "E9", "F", "UP"]
```
nateprewitt, line 184: Specifically here you'd add `"I"` to enable the isort functionality.

.pre-commit-config.yaml
nateprewitt: Original PR: https://github.com/boto/botocore/pull/3206#issue-2353772183

kdaily: Ah, I saw the part about quote configuration but not the part about `isort`. I don't have `I` right now but I can make that change. Thanks!

kdaily: > ruff does cleanup on all unused imports

This caused a lot of issues for the CLI codebase too, so I am disabling that with `ruff` until it can be fixed separately:

https://github.com/aws/aws-cli/pull/9123/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R185

```
ignore = ["F401"]
```

kdaily: I dropped `isort` from the precommit config and removed the configuration section from `pyproject.toml`, and added the `I` flag to the ruff configuration section.

pyproject.toml
kdaily: Done!

pyproject.toml
nateprewitt: Cool, we can probably yank this then and let `python_requires` be the single source of truth.

.pre-commit-config.yaml
kdaily: Done, resolving.

nateprewitt (approved):

hssyoo (approved):

PR #9136, opened 2024-12-13
Use AWS CRT for cryptographic functionality in CloudTrail, CloudFront, and EC2 customizations

kdaily:
*Issue #, if available:*

*Description of changes:*

This PR replaces cryptographic methods from the `cryptography` package with corresponding methods from the `awscrt` package. This applies to the the following customizations:

- CloudTrail log validation (`aws cloudtrail validate-logs`)
- CloudTrail query result validation (`aws cloudtrail verify-query-results`)
- EC2 password decryption (`aws ec2 get-password-data`)
- CloudFront URL signing (`aws cloudfront sign`) 


.changes/next-release/enhancement-cloudfront-36119.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "cloudfront",
+  "description": "Replace cryptographic functions from cryptography with awscrt for the sign command."
```
ashovlin, line 4: nit: I think we usually do double-backticks on the service and command names.

ashovlin (approved):

.changes/next-release/enhancement-cloudfront-36119.json
kdaily: Fixed, thanks!

aemous (approved):
Looks good to me. Something to keep on our radar going forward, is we may want to consider having a shared `tests/test_resources` folder that we load in any test/customization that uses private or public keys, to avoid duplicating the file across test types and customizations. 

Similarly, we have a hardcoded example private key in `tests/functional/cloudfront/test_sign.py`; in the long-term if we continue to add tests relying on example keys we may want to unify our tests to load from this shared folder.

However I don't think we have that much duplication that it's a problem as of now.

PR #9153, opened 2024-12-19
Add support for `stringArray` and `operationContextParams`

ashovlin:
_**Note for reviewers:** I intentionally did not include a changelog file, as this should not result in any user-observable changes for now, but let me know if you think we should._

*Issue #, if available:* CLI-4648

*Description of changes:*

### Primary Change
In the rules-based endpoints resolver, this adds support for:
1. `stringArray` as parameters into endpoint resolution
2. `operationContextParams` - this is a new trait in the service model, which can extract more complicated endpoint resolution parameters (including lists of strings) out of an operation's request.

This is a port of https://github.com/boto/botocore/pull/3301/.

### Supporting Changes
This includes two more PRs from `botocore`:
1. https://github.com/boto/botocore/pull/3302/ - this is fixing a bug in `get_attr`, which is a built-in function used within endpoint resolution to resolve named keys and/or indices into arrays or dicts.
2. https://github.com/boto/botocore/pull/2934 - this was an optimization made for `boto3` to reduce memory usage when using multiple `Session` objects. While this is less of an issue for the CLI given a single command invocation will typically use a single `Session`, by overriding the LRU cache we're using for `resolve_endpoint`, we can support unhashable types (i.e. `list`, which is used for our new `stringArray`), which the `lru_cache` in `functools` does not.

### Testing
In addition to porting the automated tests, I manually tested an internal, future service update that is taking advantage of this, and confirmed that the `unit` and `functional` tests pass locally against the updated model and endpoints files.

### DIFF
--- a/.changes/next-release/enhancement-endpoints-85370.json
+++ b/.changes/next-release/enhancement-endpoints-85370.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "endpoints",
+  "description": "Add support for ``stringArray`` parameters and the ``operationContextParams`` trait when resolving service endpoints."
+}
--- a/awscli/botocore/endpoint_provider.py
+++ b/awscli/botocore/endpoint_provider.py
@@ -24,7 +24,6 @@ or you can look at the test files in /tests/unit/data/endpoints/valid-rules/
 import logging
 import re
 from enum import Enum
-from functools import lru_cache
 from string import Formatter
 from typing import NamedTuple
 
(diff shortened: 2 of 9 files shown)


aemous (approved):

kdaily (approved):

hssyoo (approved):
Just needs a new change entry.

EDIT: Oops, just saw the note in the description. I do think it's still nice to have since it makes it easier to identify the version that introduced this change.

ashovlin:
> Just needs a new change entry.
> 
> EDIT: Oops, just saw the note in the description. I do think it's still nice to have since it makes it easier to identify the version that introduced this change.

Added via https://github.com/aws/aws-cli/pull/9153/commits/838d7c019841c515c0ac3788c124368be7a25f55

hssyoo (approved):

PR #9163, opened 2024-12-23
[v2] Expose `--bucket-name-prefix` and `--bucket-region` to `s3 ls`

hssyoo:
S3's `ListBuckets` API released filter parameters `Prefix` and `BucketRegion`. This PR exposes the low-level parameters to the high-level `s3 ls` command by exposing the `--bucket-name-prefix` and `--bucket-region` parameters. These parameters are only used if no S3 URI is supplied to the command:
* `aws s3 ls --bucket-name-prefix foo` is fine.
* `aws s3 ls s3://mybucket --bucket-name-prefix foo` is functionally the same as `aws s3 ls s3://mybucket`.

Decisions made:
* Map `Prefix` to `--bucket-name-prefix`. `s3 ls` calls `ListObjectsV2` under the hood if an S3 URI is supplied. In that case, the bucket name and prefix/key are parsed, and the `Prefix` parameter is automatically (or implicitly) supplied. Without an S3 URI, `ListBuckets` is called and the `Prefix` parameter must be explicitly supplied by the high-level `s3 ls` parameter. To avoid confusion, `ListBuckets`'s `Prefix` parameter is exposed as `--bucket-name-prefix`.
* The client doesn't attempt to validate that the 2 new parameters are only used when an S3 URI is not supplied. So `aws s3 ls s3://mybucket --bucket-region us-east-1` does not throw an error/warning.
    1. This is consistent with how parameters are generally handled for high-level `s3` commands.
    2. The parameter names clearly refer to buckets (not objects) and additional details are available in the help docs.

v1 PR will be opened after this is approved.

### DIFF
--- a/.changes/next-release/enhancement-s3ls-20704.json
+++ b/.changes/next-release/enhancement-s3ls-20704.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``s3 ls``",
+  "description": "Expose low-level ``ListBuckets` parameters ``Prefix`` and ``BucketRegion`` to high-level ``s3 ls`` command as ``--bucket-name-prefix`` and ``--bucket-region``."
+}
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -492,6 +492,26 @@ CHECKSUM_ALGORITHM = {
         'help_text': 'Indicates the algorithm used to create the checksum for the object.'
 }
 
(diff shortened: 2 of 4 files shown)


aemous (changes requested):
Looks good overall. I'd like to a see a unit & functional tests against specifying one of these args with an S3 URI, to verify it ignores the arg.

ashovlin (approved):

aemous (approved):
Looks good to me

kdaily (approved):

Issue #9132, opened 2024-12-11
Permit suppression of the `aws configure sso` example command


richardkmichael:
### Describe the feature

Concerning the example printed at the end of `aws configure sso`:

```
To use this profile, specify the profile name using --profile, as shown:

aws s3 ls --profile myProfile
```

Either:

- Add a flag to suppress it; or,
- Simply remove it, as it seems noisy to use a flag, and a suitably terse flag name is not obvious.

### Use Case

In our AWS configuration, `ListBuckets` is prohibited, so a user receives `AccessDenied` when running this command.

This confuses our staff (especially newcomers to the AWS CLI), who follow our setup instructions, run the suggested command (believing it to be a _test_), which then fails and they have the impression they made a mistake in the setup.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

aws-cli/2.22.4 Python/3.12.7 Darwin/23.6.0 source/arm64

### Environment details (OS name and version, etc.)

macOS 14

tim-finnigan:
Thanks for the feature request — I think the example command is helpful to show how users can reference the new profile. That being said, maybe there's a better example to use there than S3 — maybe [aws sts get-caller-identity](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sts/get-caller-identity.html). Will leave this open for further tracking/discussion.

richardkmichael:
@tim-finnigan I agree an example is helpful, and I considered suggesting a replacement but I couldn't think of one off-hand that would be almost certain to succeed.  Yours is a good one.

PR #9203, opened 2025-01-13
Refine SSO configure prompt titles

AndrewAsseily:
*Issue #, if available:*
* CLI-3756

*Description of changes:*
* Remove "CLI" from the prompts that apply to all SDKs
* Clarify that the default output format is json, not just "None"
* Changed the example message to address the following issue: https://github.com/aws/aws-cli/issues/9132
* Included unit tests for the aforementioned changes in addition to manual testing


awscli/customizations/configure/sso.py
```diff
@@ -488,6 +488,10 @@ def _prompt_for_cli_output_format(self):
 
     def _prompt_for_profile_config(self, config_name, text, completions=None):
         current_value = self._profile_config.get(config_name)
+
+        if config_name == 'output' and not current_value:
+            current_value = 'json'
```
ashovlin, line 493: Is this going to persist `output = json` in the profile which we don't do today?

If so, I'd try to avoid a change in behavior with a quick cosmetic change, largely so we don't have to hunt down all of the implications. 
1. If using `current_value` -> `new_value`, is there a way to not persist it if the user just pressed enter to accept the default.
2. Or, can we just communicate the default behavior through the `text` and `prompt_fmt` strings, then still actually collect and "save" `None`?

ashovlin (changes requested):

awscli/customizations/configure/sso.py
AndrewAsseily, line 1: Alternative to updating the prompt text:
 
In `_prompt_for_profile_config`, 

```
    # Only save if user actually typed something different
    if new_value and new_value != current_value:
        self._new_profile_config_values[config_name] = new_value
 ```

File: tests/unit/customizations/configure/test_sso.py [hunk 2]
```diff
@@ -1513,6 +1514,37 @@ def test_single_account_single_role_device_code_fallback(
...
+        assert f"Successfully configured SSO for profile: {profile_name}" in captured.out
+
+    def test_print_conclusion_default_profile_case_insensitive(selfself, sso_cmd, capsys):
+        sso_cmd._print_conclusion(True, 'DEFAULT')
+        captured = capsys.readouterr()
+        assert "The AWS CLI is now configured to use the default profile." in captured.out
+        assert "aws sts get-caller-identity" in captured.out
+
+    def test_print_conclusion_empty_profile_name(self, sso_cmd, capsys):
+        sso_cmd._print_conclusion(True, '')
```
ashovlin, line 1544: Is this even realistic? Doesn't empty fallback to default?

ashovlin (approved):

tests/unit/customizations/configure/test_sso.py
AndrewAsseily: After manual testing, I confirmed that an empty profile does indeed fall back to the default profile. I'll remove this test since it's not testing a valid edge case.

PR #9254, opened 2025-01-24
[v2] [customization] Create/delete role associations for EMR on EKS

hssyoo:
### Description
EKS Pod Identity requires users to create pod identity associations for cluster service accounts that the pod container is going to use with a configured job execution IAM role. This requires users to provide service account names and the IAM role arn. However, service account information is opaque in the EMR on EKS service so the `--create-role-associations` and `--delete-role-associations` customizations simplify management by resolving the correct service accounts based on the input cluster and role names.

### Manual testing
Create an IAM role with the following assume role policy document:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Service": "pods.eks.amazonaws.com"
            },
            "Action": [
                "sts:AssumeRole",
                "sts:TagSession"
            ]
        }
    ]
}
```

Create an EKS cluster.

Run the create command, replacing the `my-cluster` and `my-role` values:
```
aws emr-containers create-role-associations --cluster-name my-cluster --namespace default --role-name my-role
```

Run the delete command, replacing the `my-cluster` and `my-role` values:
```
aws emr-containers delete-role-associations --cluster-name my-cluster --namespace default --role-name my-role
```


awscli/customizations/emrcontainers/delete_role_associations.py
```diff
@@ -0,0 +1,287 @@
...
+                "eks",
+                region_name=self._region,
+                verify=parsed_globals.verify_ssl,
+            )
+        )
+        account_id = eks_client.get_account_id(self._cluster_name)
+
+        results = []
+        # If service account provided, delete association with provided service account and role
+        if self._service_account_name:
```
aemous, line 128: What should the user expect if they specify both a type and service account name ? Should we add validation to prevent this? If so, should we also update the description rst files to explain this?

Same comment for create_role_assocations

awscli/examples/emr-containers/create-role-associations/_description.rst
```diff
@@ -0,0 +1,19 @@
...
+    <FRAMEWORK> = EMR on EKS framework such as spark, flink, livy
+    <COMPONENT> = Task component for the framework. Such as client, driver, executor for spark; flink-operator, jobmanager, taskmanager for flink.
+    <AWS_ACCOUNT_ID> = AWS Account ID of the EKS cluster
+    <BASE36_ENCODED_ROLE_NAME> = Base36 encoded form of the IAM Role name
+
+You can use the **--type** option to select which submission model to associate.
+
+You can use the **--operator-namespace** option if your operator/livy jobs are running in a different operator namespace other than your job/application namespace. **--namespace** should be the Job/Application Namespace, and this option is for operator namespace to associate operator/livy service account.
+
+You can use the **--service-account-name** option to associate a custom service account with the role.
```
aemous, line 19: nit: Should we explain how specifying a service account overrides specifying a type? Or that user must specify one or the other but not both?

awscli/examples/emr-containers/delete-role-associations.rst
```diff
@@ -0,0 +1,55 @@
...
+
+EKS allows associations with non existing resources (namespace, service account), so EMR on EKS suggest to delete the associations if the namespace is deleted or the role is not in use to release the space for other associations.
+
+* Command::
+
+    aws emr-containers delete-role-associations \
+        --cluster-name example_cluster \
+        --namespace example_namespace \
+        --role-name example_iam_role \
+        (--type example_type) \
```
aemous, line 14: All my comments on create-role-associations.rst apply here too

awscli/examples/emr-containers/delete-role-associations/_description.rst
```diff
@@ -0,0 +1,19 @@
+Deletes role associations of given IAM role with EMR service accounts such that it will be removed from Amazon EMR on EKS with the given namespace from the given EKS cluster.
+
+Note:
+The command would dissociate EMR service accounts with provided IAM role to EKS pod identity:
```
aemous, line 4: nit: to -> from

awscli/customizations/emrcontainers/create_role_associations.py
```diff
@@ -0,0 +1,298 @@
...
+                "If jobs are running in a different namespace than the operator installation namespace, "
+                "this parameter needs to be set as where the operator is running on."
+            ),
+            "required": False,
+        },
+        {
+            "name": "service-account-name",
+            "help_text": (
+                "Specify the service account name that you want to associate "
+                "with the IAM role. Default to Amazon EMR on EKS service accounts."
```
aemous, line 92: The code implies that if a service account name is unspecified then it falls back to relying on the value of type, so I'm confused to what "Default to Amazon EMR on EKS service accounts." means here. If it's meant to imply it falls back to type, I think we could be more explicit on how these two args are dependent / mutually exclusive.

Similar comment for delete-role-associations

tests/unit/customizations/emrcontainers/test_create_role_associations.py
```diff
@@ -0,0 +1,596 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aemous, line 1: 2024 -> 2025

tests/unit/customizations/emrcontainers/conftest.py
```diff
@@ -0,0 +1,177 @@
+import pytest
```
aemous, line 1: Add copyright

tests/unit/customizations/emrcontainers/test_create_role_associations.py
```diff
@@ -0,0 +1,596 @@
...
+        cli_runner.add_response(
+            HTTPResponse(body=create_pod_identity_association_response)
+        )
+        cli_runner.add_response(
+            HTTPResponse(body=create_pod_identity_association_response)
+        )
+        cli_runner.add_response(
+            HTTPResponse(body=create_pod_identity_association_response)
+        )
+        result = cli_runner.run(
```
aemous, line 59: These tests seem like functional tests rather than unit tests. Can we update their directory locations accordingly?

tests/unit/customizations/emrcontainers/test_delete_role_associations.py
```diff
@@ -0,0 +1,578 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aemous, line 1: 2024 -> 2025

tests/unit/customizations/emrcontainers/test_delete_role_associations.py
```diff
@@ -0,0 +1,578 @@
...
+            request_idx += 1
+            self.assert_list_call_matches(
+                result.aws_requests[request_idx],
+                livy_service_accounts[i],
+                cluster_name,
+                namespace,
+            )
+            request_idx += 1
+            self.assert_delete_call_matches(
+                result.aws_requests[request_idx], cluster_name, str(i + 1)
```
aemous, line 434: Could be worth adding an assertion that there are no more than the expected number of requests in result.aws_requests list

tests/unit/customizations/emrcontainers/test_delete_role_associations.py
```diff
@@ -0,0 +1,578 @@
...
+        self.assert_list_call_matches(
+            result.aws_requests[1], "test_sa", cluster_name, namespace
+        )
+        self.assert_delete_call_matches(
+            result.aws_requests[2], cluster_name, "1"
+        )
+
+    # Use case: Expect to do nothing on deletion when resource not found but print warning message
+    # Expected results: Association deletions are skipped
+    def test_create_role_associations_already_exists_for_start_job_run(
```
aemous, line 545: create -> delete

aemous (changes requested):

.changes/next-release/api-change-emrcontainers-7930.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "api-change",
```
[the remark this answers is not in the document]
hssyoo: Hm yeah it looks like new customizations are introduced as features. I'll update.

awscli/customizations/emrcontainers/create_role_associations.py
```diff
@@ -0,0 +1,298 @@
...
+
+                # Raise the error if EKS throws exceptions other than ResourceInUseException
+                if error.response["Error"]["Code"] != "ResourceInUseException":
+                    for result in results:
+                        uni_print(
+                            f"Rolling back association for service account {service_account_name} "
+                            + f"and role {self._role_name} in namespace {namespace} as encountered an error",
+                            out_file=sys.stderr,
+                        )
+                        eks_client.delete_pod_identity_association(
```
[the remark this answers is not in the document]
hssyoo: An error will be raised as a standard client exception.

awscli/customizations/emrcontainers/delete_role_associations.py
```diff
@@ -0,0 +1,287 @@
...
+                    self._cluster_name, namespace, service_account_name
+                )
+                if not association.get("associations", []):
+                    uni_print(
+                        f"Skipping deletion as no pod identity association found for service account {service_account_name} "
+                        + f"and role {self._role_name} in namespace {namespace}\n",
+                        out_file=sys.stderr
+                    )
+                    continue
+                association_id = association["associations"][0]["associationId"]
```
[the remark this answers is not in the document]
hssyoo: My opinion is that creating these associations outside of this workflow would be considered undefined behavior and we shouldn't do anything special, but tagging @zhuyulin27 for her take.

awscli/customizations/emrcontainers/delete_role_associations.py
```diff
@@ -0,0 +1,287 @@
...
+            (
+                ServiceAccount.LIVY_SERVICE_ACCOUNT.value,
+                (
+                    self._operator_namespace
+                    if self._operator_namespace
+                    else self._namespace
+                ),
+            ),
+            (ServiceAccount.LIVY_SPARK_SERVICE_ACCOUNT.value, self._namespace),
+        ]
```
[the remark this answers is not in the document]
hssyoo: They are identical. I intentionally didn't refactor it because I don't see them being reused by other commands. We can always revisit this if we do end up adding commands that use them in the future.

awscli/customizations/emrcontainers/create_role_associations.py
```diff
@@ -0,0 +1,298 @@
...
+            )
+            try:
+                result = eks_client.create_pod_identity_association(
+                    self._cluster_name,
+                    namespace,
+                    role_arn,
+                    service_account_name,
+                )
+                results.append(
+                    result["association"] if "association" in result else result
```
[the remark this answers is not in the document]
hssyoo: @zhuyulin27

awscli/customizations/emrcontainers/delete_role_associations.py
hssyoo: I think it's reasonable for a user defined service account name to take precedence over other parameters. Parameter validation based on mutual exclusion isn't something we generally do unless there are security concerns or potential footguns.

awscli/customizations/emrcontainers/eks.py
```diff
@@ -40,3 +40,25 @@ def get_account_id(self, cluster_name):
...
+                 associationId = association_id
+            )
+    
+    def list_pod_identity_associations(self, cluster_name, namespace, service_account):
+            return self.eks_client.list_pod_identity_associations(
+                 clusterName = cluster_name,
+                 namespace = namespace,
+                 serviceAccount = service_account
+            ) 
+
```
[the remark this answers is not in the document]
hssyoo: @zhuyulin27 and I talked about this. In the end, the service team would prefer treating the `EKS` class as a high-level client and expose client methods that delegate to the low-level eks client.

awscli/customizations/emrcontainers/delete_role_associations.py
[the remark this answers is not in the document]
zhuyulin27: He Ahmed, the reason being, in EKS concept, each service account can only be associated to 1 role, so the list filter (cluster, namespace, service account) should only return 1 result, regardless of how many associations the role has.

awscli/customizations/emrcontainers/create_role_associations.py
zhuyulin27: This command is mainly for simplifying customer experience to create role associations with EMR service accounts under the hood. So that customer don't need to figure out EMR service accounts pattern and perform manual actions. This extra `service-account-name` is optional if they would like to do it manually or associate their roles to EKS with specific service account. But the original idea for this command is to create associations with EMR service accounts based on submission models. (Spark/Spark Operator/..)

Service account name and type are both optional and yes mutually exclusive, once service account name is set, we will directly use the name for the association and will not generate the EMR ones based on the type.

aemous: Thanks for the clarification. My suggestion in this comment is to clarify this for the user. 

I think it's confusing to say there's a default value since if service-account-name is unspecified we don't utilize it. Typically default implies if the argument is unspecified, then a default value for the argument is used. This is not the case here, we instead fall back to using a different argument (type).

 My (non-required) suggestion is to replace the "default" wording with specifying that this overrides the type argument if it's supplied. 

Feel free to resolve this conversation if you feel this would not be confusing to your customers.

awscli/customizations/emrcontainers/delete_role_associations.py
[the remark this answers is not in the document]
aemous: Thank you for this clarification.

awscli/customizations/emrcontainers/create_role_associations.py
zhuyulin27: Sounds reasonable, we can change it to 
                "(Optional) Specify the service account name that you want to associate "
                "with the IAM role. If this is not set, Amazon EMR on EKS service accounts will be used for association."

tests/functional/emrcontainers/test_delete_role_associations.py
```diff
@@ -0,0 +1,587 @@
...
+                "--role-name",
+                role_name,
+            ]
+        )
+
+        expected_out = ""
+        for service_account in start_job_run_service_accounts:
+            expected_out += (
+                f"Skipping deletion as no pod identity association found for service account {service_account} "
+                + f"and role {role_name} in namespace {namespace}\n"
```
aemous, line 585: non-blocking nit: Unnecessary concatenation operator `+`

aemous (changes requested):
Looks good overall. Requesting changes on the 2 examples, and a few non-blocking suggestions otherwise.

awscli/examples/emr-containers/delete-role-associations.rst
```diff
@@ -0,0 +1,50 @@
...
+            "serviceAccount": "emr-spark-executor-service-account-example",
+            "roleArn": "arn:aws:iam::111111111111:role/example_iam_role",
+            "associationArn": "arn:aws:eks:us-east-1:111111111111:podidentityassociation/example_cluster/c-bgyr1umgdmrk1kdtq",
+            "associationId": "c-bgyr1umgdmrk1kdtq",
+            "tags": {},
+            "createdAt": "2022-11-15T10:49:00+00:00",
+            "modifiedAt": "2022-11-15T10:49:00+00:00"
+        }
+    ]   
+
```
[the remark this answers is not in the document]
aemous: Resolved offline

awscli/examples/emr-containers/create-role-associations.rst
```diff
@@ -0,0 +1,48 @@
...
+            "serviceAccount": "emr-spark-executor-service-account-example",
+            "roleArn": "arn:aws:iam::111111111111:role/example_iam_role",
+            "associationArn": "arn:aws:eks:us-east-1:111111111111:podidentityassociation/example_cluster/c-bgyr1umgdmrk1kdtq",
+            "associationId": "c-bgyr1umgdmrk1kdtq",
+            "tags": {},
+            "createdAt": "2022-11-15T10:49:00+00:00",
+            "modifiedAt": "2022-11-15T10:49:00+00:00"
+        }
+    ]   
+
```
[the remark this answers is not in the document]
aemous: Resolved offline

aemous (approved):
LGTM

PR #9258, opened 2025-01-28
Revert "Use CRT for cloudfront url signing"

kdaily:
This reverts commit 2ecf7c5a32bc4256754ef5f291319d273c6c0bbb.

*Issue #, if available:*

*Description of changes:*

Switching from `cryptography` to `awscrt` for cryptographic operations lost the ability to use PKCS8-like public keys (those that start with a header `BEGIN PUBLIC KEY`). Reverting this change to restore the previous behavior until the functionality can be added to `awscrt`.


nateprewitt (approved):
:shipit:

hssyoo (approved):
Looks good, can we provide a brief reason for reverting in the description?

PR #9264, opened 2025-01-30
[v2] Benchmark harness improvements & add more service benchmarks

aemous:
*Issue #, if available:*
N/A

*Description of changes:*
- Adds support for specifying file literals in the benchmark definitions JSON.
- Updated README with documentation on specifying file literals.
- Added dynamodb benchmarks for `get-item`, `batch-get-item`, `put-item` commands, and a benchmark for the customization `ddb put`.
- Aded ec2, ecs, cloudwatch logs, and cloudwatch benchmarks.
- Added STS benchmarks for `AssumeRole`, `GetCallerIdentity`, `AssumeRoleWithWebIdentity`.
- Removed S3 benchmarks using the CRT client because they are currently non-functional with a non-trivial fix. Will possibly revisit getting them working later.
- Fix bug in benchmark summarizer with computing averages.
- Add support to the benchmark harness for debugging the child process.
- Redirect child process output to files so that the only output of `run-benchmarks` script on success is the benchmark results.


scripts/performance/benchmarks.json
```diff
@@ -136,30 +107,154 @@
     ]
   },
   {
-    "name": "s3.sync.upload",
-    "command": ["s3", "sync", "./test_dir", "s3://bucket/test_dir", "--quiet"],
+    "name": "dynamodb.getitem",
+    "command": ["dynamodb", "get-item", "--table-name", "MyTable", "--key", "file://key.json", "--return-consumed-capacity", "TOTAL"],
```
ashovlin, line 111: How much does the `file://` usage impact this output as opposed to specifying the key inline? I see we use it in examples, but curious if it's worth separating DDB's performance from that of `file://`? But if it's relatively small compared to the full operation, okay with this.

ashovlin: After chatting internally, it's not significant relative to the overall benchmark (and is how we have the DDB examples documented). Okay with it as-is.

scripts/performance/benchmarks.json
```diff
@@ -20,36 +20,7 @@
...
-        "body": "<CompleteMultipartUploadResult><Bucket>bucket</Bucket>  <Key>key</Key>  <ETag>etag-123</ETag></CompleteMultipartUploadResult>"
-      }
-    ]
-  },
-  {
-    "name": "s3.cp.upload",
-    "command": ["s3", "cp", "test_file", "s3://bucket/test_file", "--quiet"],
-    "dimensions": [
-      {"FileSize": "32MB"},
-      {"S3TransferClient": "CRT"}
```
ashovlin, line 36: > Removed S3 benchmarks using the CRT client because they are currently non-functional with a non-trivial fix. Will possibly revisit getting them working later.

Could you explain what the issue and non-trivial fix are?

aemous: All of our benchmarks stub the HTTP layer via mocked API responses. For the CRT S3 client, the HTTP layer lives in the underlying C++ code. At best, we would write CRT-specific logic for stubbing the `awscrt.S3Client` class, which is one layer/abstraction higher than HTTP. 

Here are the pros and cons of this that I've noted.

Pros:
- Could detect performance regressions in our CRT-specific code paths.

Cons:
- CRT benchmarks would be different than all others due to not being able to stub the HTTP layer. As a result, we would not be able to compare CRT benchmark results to any other benchmarks. Though, perhaps we don't need/intend to.

ashovlin (approved):

PR #9265, opened 2025-01-30
Ignore Content-Length header when setting 'chunked' for Transfer-Encoding 

jonathan343:
> Note: This is the AWS CLI v2 version of https://github.com/boto/botocore/pull/3360

This PR removes the `Content-Length` header when a request is sent using chunked upload.

See the guidance below from [Signature Calculations for the Authorization Header: Transferring Payload in Multiple Chunks](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html) for more details 

> When transferring data in a series of chunks, you must do one of the following:
>
>Explicitly specify the total content length (object length in bytes plus metadata in each chunk) using the Content-Length >HTTP header. To do this, you must pre-compute the total length of the payload, including the metadata that you send in >each chunk, before starting your request.
>
>Specify the Transfer-Encoding HTTP header. If you include the Transfer-Encoding header and specify any value other >than identity, you must omit the Content-Length header.
>
>For all requests, you must include the x-amz-decoded-content-length header, specifying the size of the object in bytes.

### DIFF
--- a/awscli/botocore/httpchecksum.py
+++ b/awscli/botocore/httpchecksum.py
@@ -388,7 +388,13 @@ def _apply_request_trailer_checksum(request):
         # Send the decoded content length if we can determine it. Some
         # services such as S3 may require the decoded content length
         headers["X-Amz-Decoded-Content-Length"] = str(content_length)
-
+        
+        if "Content-Length" in headers:
+            del headers["Content-Length"]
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
:shipit:

PR #9286, opened 2025-02-09
remove the very last reference to "six"

a-detiste:
Hi, it was a very long way but it seems this is the very last reference to six.

Greetings

*Description of changes:*

https://wiki.debian.org/Python3-six-removal


nateprewitt:
Hi @a-detiste, thanks for the PR. I'm not sure we'd call these imports indirection. We're specifically managing shared pieces we may shift or update in the future in one centralized place across multiple packages. They've existed there for the majority of the library's lifetime and are subject to further changes. The set of PRs you're currently proposing doesn't seem to add significant value and come at a higher expense for any future changes.

Removing the six reference from compat doesn't actually remove six from this library's dependency closure since it's still used in python-dateutils. We'd be open to accepting a PR that specifically removes the `six` import from `awscli.compat`, but that's likely the extent of the changes we'd want at this time.

a-detiste:
Please just consider that the most of the library lifetime was this ugly Python2 + 3 era and that is not coming back, the Python core team learned from their mistakes and won't repeat them anytime soon. 

The current codebase os already a hodgepoge of the old way and new way intermixed: so many places import OrderedDict from "botocore.compat" and so many other - newer - places import it directly from "collections".

I ve spent the best of these last two years sending PR to remove all this old residual scar tissue and am pretty confident Python 2 won't return. Sometimes the perfect reason these compat.py got lost in time and the reason to keep them around are akin to cargo cult.

a-detiste:
I looked into botocore codebase itself and I see they did much progress to remove unneeded usage of their own compat.py . Only the test cases are laging behind and need more cleanup, I provided a followup PR there too.

PR #9293, opened 2025-02-11
[v2] Move requirements changes to its own requirements file.

aemous:
*Description of changes:*
* Revert all changes to requirements files made in #9278 
* Add psutil requirement to a separate file in `scripts/performance/requirements.txt`.

*Description of tests:*
* Created a new fresh python3.12 virtual environment and ran `python3.12 -m pip install -e .` then `python3.12 -m pip install -r scripts/performance/requirements.txt` and confirmed the script runs successfully.


scripts/performance/requirements.txt
```diff
@@ -0,0 +1 @@
+psutil==6.1.1
```
ashovlin, line 1: I was wondering if we want a lockfile, but when testing locally it looks like `psutil` would be the only thing included so I think that's okay.

ashovlin (approved):

PR #9304, opened 2025-02-19
Remove old jQuery and fix copyright year in HTML docs [V2]

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* Some low-hanging fruit for the V2 API reference:
1. Remove jQuery 1.9.1 which is included via our Sphinx theme. 
     * This is a common source of internal tickets from scanners looking for outdated versions.
     * 3.4.1 is being included via Sphinx itself: https://awscli.amazonaws.com/v2/documentation/api/latest/_static/jquery.js, we should still try to keep this updated, but lets remove the old one first
2. Update the copyright year in the date, and compute it during generation.

### Testing
```
cd doc/
make html
python -m http.server --directory build/html
```

1. Clicked around, verified no script errors.
2. Searched from the side bar (http://127.0.0.1:8000/index.html), verified that there were no script errors and that the search worked.
3. Searched from the search page itself (http://127.0.0.1:8000/search.html), verified that there were no script errors and that the search worked


doc/source/conf.py
```diff
@@ -61,8 +61,8 @@
 master_doc = 'index'
 
 # General information about the project.
-project = u'AWS CLI Command Reference'
-copyright = u'2018, Amazon Web Services'
+project = 'AWS CLI Command Reference'
+copyright = f'{datetime.datetime.now().year}, Amazon Web Services'
```
nateprewitt, line 65: We technically don't need a year with this anymore (or to update it at all) but this is fine for now.

nateprewitt (approved):

doc/source/conf.py
ashovlin: I did consider removing it, I kind of like some sort of year or timestamp in docs to indicate "freshness", but not a strong opinion if we want to adjust again.

PR #9326, opened 2025-02-26
[v2] Fix sdist lockfile test

hssyoo:
[GitHub Action that runs on Python 3.12 and macOS 13 is failing sdist tests](https://github.com/aws/aws-cli/actions/runs/13552377342/job/37878861338?pr=9324) because 1 of the lockfiles is generated with a newline at the end. This commit strips leading/trailing newlines to ensure consistent comparisons.

### DIFF
--- a/tests/backends/build_system/functional/test_lock_files.py
+++ b/tests/backends/build_system/functional/test_lock_files.py
@@ -29,7 +29,7 @@ def should_read_line(line):
 
 def read_lock_file(path):
     with open(path, 'r') as f:
-        lines = f.read().split('\n')
+        lines = f.read().strip('\n').split('\n')
 
     # Find source files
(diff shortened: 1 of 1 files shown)


ashovlin (approved):
I think we fixed some of them up manually when merging the performance work since we were adding a new lock file and then moved it. Maybe that threw it off? But seems okay to fix here.

hssyoo:
> I think we fixed some of them up manually when merging the performance work since we were adding a new lock file and then moved it. Maybe that threw it off? But seems okay to fix here.

This is specifically for source distribution lockfiles and it's oddly only affecting 1 test matrix, so I'd assume they're unrelated. But yeah doing basic sanitization here is a marginal improvement either way.

PR #9333, opened 2025-02-28
[v2] User agent 2.1: Track client features

hssyoo:
Ports https://github.com/boto/botocore/pull/3389 and https://github.com/boto/s3transfer/pull/335

I also added a CLI-specific change to wrap the CLI entrypoint under the `start_as_current_context` context manager so future work can add CLI-specific features.

```
> aws s3 cp myfile s3://mybucket
> aws history show

...
"User-Agent": "aws-cli/2.24.14 ... m/G cfg/retry-mode#standard ..."
...
```

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -15,6 +15,7 @@ import logging
 import sys
 
 from botocore.client import Config
+from botocore.useragent import register_feature_id
 from botocore.utils import is_s3express_bucket, ensure_boolean
 from dateutil.parser import parse
 from dateutil.tz import tzlocal
@@ -783,6 +784,7 @@ class PresignCommand(S3Command):
 class S3TransferCommand(S3Command):
     def _run_main(self, parsed_args, parsed_globals):
         super(S3TransferCommand, self)._run_main(parsed_args, parsed_globals)
+        register_feature_id('S3_TRANSFER')
         self._convert_path_args(parsed_args)
         params = self._get_params(parsed_args, parsed_globals, self._session)
         source_client, transfer_client = self._get_source_and_transfer_clients(
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -15,7 +15,6 @@ import logging
 import sys
 
 from botocore.client import Config
-from botocore.useragent import register_feature_id
 from botocore.utils import is_s3express_bucket, ensure_boolean
 from dateutil.parser import parse
 from dateutil.tz import tzlocal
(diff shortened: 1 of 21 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -783,6 +784,7 @@ def _run_main(self, parsed_args, parsed_globals):
 class S3TransferCommand(S3Command):
     def _run_main(self, parsed_args, parsed_globals):
         super(S3TransferCommand, self)._run_main(parsed_args, parsed_globals)
+        register_feature_id('S3_TRANSFER')
```
hssyoo, line 787: Thinking through it more, although we currently only use S3 transfer clients in the high-level `s3` command, it may be used in other customizations in the future.

hssyoo: We'd also want to track usage if users are using custom plugins. I addressed this in https://github.com/aws/aws-cli/pull/9333/commits/8797100e67d330d1aae5ad400943eb87bd059cb3

ashovlin (commented):
Changes look good. For the S3 high-level commands, can we track the `HeadObject` calls too (unless the internal design says not to)?

tests/functional/test_useragent.py
```diff
@@ -10,7 +10,7 @@ def assert_expected_user_agent(result, service, operation):
     assert ' md/arch#' in ua_string
     assert ' md/prompt#off' in ua_string
     assert f' md/command#{service}.{operation}' in ua_string
-    assert ' ua/2.0 ' in ua_string
+    assert ' ua/2.1 ' in ua_string
```
ashovlin, line 13: Not a blocker, but might be nice to assert the `m/C` for the `aws s3 ls` case below while we're here.

hssyoo: Addressed in https://github.com/aws/aws-cli/pull/9333/commits/edb58d5e13871c45da53eab79a9e8d88ffc48281

hssyoo:
> Changes look good. For the S3 high-level commands, can we track the `HeadObject` calls too (unless the internal design says not to)?

Addressed in https://github.com/aws/aws-cli/pull/9333/commits/35e6716d08d29efc2310dd4281e5832829e7ab7f

I considered only registering at the point of `HeadObject`/`ListObjectsV2` calls ([code](https://github.com/aws/aws-cli/blob/e498f32228142e58e8a8f2fe435e6fe31872988d/awscli/customizations/s3/filegenerator.py#L305C9-L305C16)), but decided to just move it under `S3TransferCommand` so that we consider any API calls made in this class of commands to have used S3Transfer.

ashovlin (approved):

PR #9343, opened 2025-03-04
Pin Python version of source dist tests until PyInstaller upgrade

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* Pinning the Python versions of 3.9 and 3.10 that are still compatible with our current PyInstaller version until we address the issues that led to #9332


kdaily (approved):
LGTM. I see in the CI for windows that its using 3.9.13 and 3.10.11, earlier versions as expected than the ones pinned for Mac/Linux.

kdaily:
`macos-latest` 3.8 and 3.9 tests taking quite a while.

Issue #8910, opened 2024-09-10
Memory leak in aws s3 cp when piping .zstd files through stdout to a slow consumer


dandrei:
### Describe the bug

Like the title says, when doing an `aws s3 cp` to stdout on a `.zst` file piped to `zstd -dc` and then to a slow consumer, after a while the AWS CLI exhibits memory leak, leading to the complete freeze of the OS within minutes.

### Expected Behavior

The file should be streamed without incident.

### Current Behavior

After a certain amount of time, the AWS CLI starts using too much memory until the OS freezes.

### Reproduction Steps

Below is a script that streams a `.zst` file to stdout (`-`), pipes it to `zstd -dc` and then to a slow consumer: a Python script that reads from stdin and sleeps for 1 second every 1000 lines.

For this to work, you need access to an S3 bucket with a large enough `.zst` file for the bug to manifest. In my tests, it consistently happened before 1 million lines were read, but you could use a larger file to be sure. The production file I was reading when I stumbled upon this bug contained one JSON object / line, and lines had an average length of ~1000-2000 characters, in case this matters.

The `ulimit` is there to prevent the memory leak from freezing your machine, the process will instead get killed once the allowed memory is exceeded (2G). It also demonstrates that the leak happens in the call to `aws s3 cp`, and not at any other step.

I have tested this with higher values like 8G, 16G, 32G, no matter how much memory you have once the bug happens your machine's RAM will get filled up within minutes, even though up to that point memory use remained largely linear.

    (ulimit -v 2097152; aws s3 cp s3://your-bucket/path/to/file.zst -) | zstd -dc | python -c "
    import sys, time
    line_count = 0
    for line in sys.stdin: 
        line_count += 1
        if line_count % 1000 == 0:
            print(str(line_count))
            time.sleep(1)
    "

### Possible Solution

My best guess is that it has to do with the multi-threaded nature of `aws s3 cp`, and the way `zstd -dc` reads data from stdin. The bug doesn't manifest with other file formats (e.g. `gzip -dc`), or with `aws s3api get-object`.

### Additional Information/Context

This was reproduced on multiple CLI versions, and on multiple operating systems (Ubuntu, Amazon Linux).

### CLI version used

2.15.43

### Environment details (OS name and version, etc.)

Ubuntu 22.04.4 LTS

tim-finnigan:
[...]

PR #9345, opened 2025-03-05
[v2] Bug fix and optimization in s3transfer's download

aemous:
*Issue #, if available:*
- Possibly related: https://github.com/aws/aws-cli/issues/8910 

# Context

This PR aims to address a bug that occurs when an `IncompleteReadError` is encountered while downloading a multipart AWS S3 object to a non-seekable stream (e.g. stdout). In this case, all bytes of the object with offset larger than the offset of the incomplete read get queued in a buffer, and the CLI terminates before the buffer is flushed. 

The results of this are (1) the object does not fully get written to the stream, and (2) the memory usage of the CLI grows linearly with respect to the amount of object bytes with offset larger than the offset of the incomplete read. This means that most of the object may be stored in memory no matter how large the object is.

The scope of the changes in this PR to s3transfer will _only_ impact users downloading multi-part AWS S3 objects to a non-seekable stream (e.g. stdout).

# Description of changes

## s3transfer

- Fix bug that occurs when an `IncompleteReadError` is encountered while downloading a multipart AWS S3 object to a non-seekable stream (e.g. stdout).
 - Changed `DeferQueue` so that it will overwrite pending writes to the same offset with whichever write request has the most data. So, when the request with the incomplete read is retried, it will overwrite the incomplete data in the queue assuming the retry contains more data (e.g. the full part).
 - Changed `DeferQueue` so that if it has already dequeued an incomplete part, it will queue the subset of the bytes in the retry that were not previously dequeued.

## Performance Benchmark Scripts

- Added two new benchmark definitions to `benchmarks.json`: one simulates a standard multi-part download of a 10GB file (1,192 parts), and the other simulates the same download except an `IncompleteReadError` occurs (with retry) on the 32nd part.
- Updated the benchmark harness to support loading mocked HTTP response bodies from files. Updated the performance scripts README with the updated definition schema.
- Added support for a `--debug-dir <path>` path to make benchmark-definition debugging easier. When this path is specified, all output of the child (benchmarked) processes gets written to files in this directory. 

# Description of tests

## Correctness

- Added 2 new unit tests to cover the 2 main modifications of `DeferQueue`. Also updated names/documentation of existing related tests.
- Verified that the new benchmark that simulates an incomplete read successfully writes the entire object to stdout by piping the output to a file and checking its size. 
  - Before the changes to `DeferQueue` contained in this PR, I verified that the full object does NOT successfully get written. 
- Ran and passed all existing tests (unit, functional, integration, etc.)
- Manually used this CLI build to successfully download a 10GB object from S3 to stdout using the default concurrency options.

## Performance Testing

- Ran the performance tests before and after the changes to `DeferQueue` contained in this PR and observed significant optimizations to memory usage and execution time. The results are summarized below

### Summary <a href="#user-content-summary" id="summary">#</a>

max_mem_issue
execution_time_mem_issue

### Raw Data

Each cell in the table below is the average value of the metric taken over 5 iterations.
| - | 2.24.14 | 2.24.14 | 2.24.14 (w/ DeferQueue improvements) |  2.24.14 (w/ DeferQueue improvements) |
| -- | -- | -- | -- | -- |
| - | Happy Path | IncompleteReadError | Happy Path | IncompleteReadError |
| Max Memory (GB) | 0.7246 | 5.242 | 0.207 | 0.20836 |
| p95 Memory (GB) | 0.7101 | 5.025 | 0.207 | 0.19397 |
| Max CPU (%) | 19.72 | 11.82 | 18.42 | 18.42 |
| p95 CPU (%) | 1.5 | 1.5 | 2.5 | 2.5 |
| Total Time (s) | 179.4 | 9.848 | 111.7 | 110.18 |


ashovlin (commented):
Is it worth running this new benchmark continuously? 

Mostly asking for ergonomics, since GitHub won't even show the file diff now. If we don't want to run it as part of the new suite, wondering if it'd be better to have a separate `benchmarks-s3-advanced.json` (or whatever naming) separately.

aemous:
> Is it worth running this new benchmark continuously?
> 
> Mostly asking for ergonomics, since GitHub won't even show the file diff now. If we don't want to run it as part of the new suite, wondering if it'd be better to have a separate `benchmarks-s3-advanced.json` (or whatever naming) separately.

@ashovlin Agreed that would be a good idea, I iterated on this feedback in the latest revision.

hssyoo:
So this seems like it would work, but I'm not sure I understand why we're writing an incomplete get at all. It looks like `IncompleteReadError` is one of the retryable errors ([ref](https://github.com/aws/aws-cli/blob/f486310392c959424a5b803078fc2f04cee0f800/awscli/s3transfer/utils.py#L44)) ([caught here](https://github.com/aws/aws-cli/blob/f486310392c959424a5b803078fc2f04cee0f800/awscli/s3transfer/download.py#L598)). I think we should be exploring fixing the underlying issue here. Do you have any debug logs for the failure case?

aemous:
> So this seems like it would work, but I'm not sure I understand why we're writing an incomplete get at all. It looks like `IncompleteReadError` is one of the retryable errors ([ref](https://github.com/aws/aws-cli/blob/f486310392c959424a5b803078fc2f04cee0f800/awscli/s3transfer/utils.py#L44)) ([caught here](https://github.com/aws/aws-cli/blob/f486310392c959424a5b803078fc2f04cee0f800/awscli/s3transfer/download.py#L598)). I think we should be exploring fixing the underlying issue here. Do you have any debug logs for the failure case?

@hssyoo 

### IncompleteReadErrors are unavoidable

Based on my investigation, IncompleteReadErrors are unavoidable; for example, if we pause the program for 10 minutes while its reading from the socket, S3 will likely close the connection in such case due to this idle time. As a result, when the program resumes, it will find nothing left in the socket to read. Then, it compares the amount of data read to the `Content-Length` header, and since the amount read is less than the amount specified in the header, we raise an IncompleteReadError ([code](https://github.com/aws/aws-cli/blob/v2/awscli/botocore/response.py#L127-L135)).

So, since these errors are unavoidable, I decided we need to at least be able to handle it gracefully and performantly.

### Why write an incomplete read?

We don't know the read is incomplete until we read the entire part from S3 (described above). So, if we wanted to only write to the DeferQueue after we're sure it's not an incomplete read, we would have to (1) wait for the entire part to be received over the wire, and (2) load the entire part in memory before sending it to the DeferQueue. 

(1) By waiting for the entire part to be received, we would be decreasing throughput of writing the object to the queue/stream. 

(2) An average part size in practice with default config values can be about 8.4 MB (or higher depending on what user sets for `multipart_chunksize`, and since we're processing multiple parts concurrently this memory can multiply up with the number of threads.

For reasons (1) and (2), I decided that sending incomplete data to the DeferQueue is preferable for performance reasons.

awscli/s3transfer/download.py
```diff
@@ -768,23 +766,49 @@ def request_writes(self, offset, data):
...
+                # request with this one since it contains more data.
+                self._pending_offsets[offset] = data
+        else:
+            heapq.heappush(self._writes, offset)
+            self._pending_offsets[offset] = data
+        while self._writes and self._writes[0] == self._next_offset:
+            next_write_offset = heapq.heappop(self._writes)
+            next_write = self._pending_offsets[next_write_offset]
+            writes.append({'offset': next_write_offset, 'data': next_write})
+            del self._pending_offsets[next_write_offset]
```
SamRemis, line 812: Any reason we are using `del` here instead of just `next_write = self._pending_offsets.pop(next_write_offset)`?

kdaily (approved):
Looks good, discussed offline one change - the `benchmarks-s3-advanced.json` is quite a large file (805KB) to include in the repository. We should consider what benchmarks need to be run continuously or ways to reduce the size of the data files added to the repository. Good to ship otherwise.

PR #9356, opened 2025-03-11
[v2] SSOTokenLoader Expiration Check

aemous:
*Description of changes:*
- Updated `SSOTokenLoader` to check if the cached legacy token is expired according to the local clock. If expired, it will raise an `UnauthorizedSSOTokenError` instead of sending an expired token to Identity Center's `GetRoleCredentials` API.
- Added a unit test to verify when an expired token is cached, that GetRoleCredentials is not called and `UnauthorizedSSOTokenError` is raised.

*Description of tests:*
- Ran all test suites (unit, functional, integration, etc.)
- The following manual workflow was completed to test the code.
  1. Login to a user account using the legacy flow via `aws sso login`.
  2. Let the cached token expire.
  3. Ran a command using the SSO profile (e.g. `aws s3 ls --profile SSOProfile --debug`)
  4. Verified an `UnauthorizedSSOTokenError` was raised, and checked the stacktrace (pasted below) to verify this exception was raised _without_  `GetRoleCredentials` being called.

**Stacktrace from manual workflow on this branch**

```
 File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 519, in _protected_refresh
    metadata = self._refresh_using()
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 660, in fetch_credentials
    return self._get_cached_credentials()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 670, in _get_cached_credentials
    response = self._get_credentials()
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 2056, in _get_credentials
    raise UnauthorizedSSOTokenError()
```

**Stacktrace from manual workflow on v2**

```
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 518, in _protected_refresh
    metadata = self._refresh_using()
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 659, in fetch_credentials
    return self._get_cached_credentials()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 669, in _get_cached_credentials
    response = self._get_credentials()
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 2055, in _get_credentials
    raise UnauthorizedSSOTokenError()
botocore.exceptions.UnauthorizedSSOTokenError: The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
2025-03-13 10:45:14,232 - MainThread - awscli.clidriver - DEBUG - Exception caught in main()
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/credentials.py", line 2053, in _get_credentials
    response = client.get_role_credentials(**kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/client.py", line 365, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/context.py", line 124, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aemous/GitHub/aws-cli2/awscli/botocore/client.py", line 752, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.UnauthorizedException: An error occurred (UnauthorizedException) when calling the GetRoleCredentials operation: Session token not found or invalid
```

Notice the above exception is raised due to a returned `UnauthorizedException` from `GetRoleCredentials`, while this is not the case in the previous stacktrace.

### DIFF
--- a/awscli/botocore/credentials.py
+++ b/awscli/botocore/credentials.py
@@ -2050,11 +2050,8 @@ class SSOCredentialFetcher(CachedCredentialFetcher):
             # raise an UnauthorizedSSOTokenError if the loaded legacy token
             # is expired to save a call to GetRoleCredentials with an
             # expired token.
-            print(f'expiresAt field to be passed to parse function: {token_dict["expiresAt"]}')
-            print(f'entire cached token: {token_dict}')
             expiration = dateutil.parser.parse(token_dict['expiresAt'])
             remaining = total_seconds(expiration - self._time_fetcher())
-            print(f'remaining: {remaining}')
             if remaining <= 0:
                 raise UnauthorizedSSOTokenError()
 
--- a/awscli/botocore/credentials.py
+++ b/awscli/botocore/credentials.py
@@ -234,10 +234,6 @@ def _local_now():
     return datetime.datetime.now(tzlocal())
 
 
-def _utc_now():
-    return datetime.datetime.now(tzutc())
-
-
 def _parse_if_needed(value):
     if isinstance(value, datetime.datetime):
         return value
@@ -1995,7 +1991,7 @@ class SSOCredentialFetcher(CachedCredentialFetcher):
     def __init__(self, start_url, sso_region, role_name, account_id,
                  client_creator, token_loader=None, cache=None,
                  expiry_window_seconds=None, token_provider=None,
-                 sso_session_name=None, time_fetcher=_utc_now):
+                 sso_session_name=None, time_fetcher=_local_now):
         self._client_creator = client_creator
         self._sso_region = sso_region
         self._role_name = role_name
--- a/awscli/botocore/credentials.py
+++ b/awscli/botocore/credentials.py
@@ -1991,7 +1991,7 @@ class SSOCredentialFetcher(CachedCredentialFetcher):
     def __init__(self, start_url, sso_region, role_name, account_id,
                  client_creator, token_loader=None, cache=None,
                  expiry_window_seconds=None, token_provider=None,
-                 sso_session_name=None, time_fetcher=_local_now):
+                 sso_session_name=None, time_fetcher=_utc_now):
         self._client_creator = client_creator
         self._sso_region = sso_region
         self._role_name = role_name
--- a/awscli/botocore/credentials.py
+++ b/awscli/botocore/credentials.py
@@ -234,6 +234,10 @@ def _local_now():
     return datetime.datetime.now(tzlocal())
 
 
+def _utc_now():
+    return datetime.datetime.now(tzutc())
+
+
 def _parse_if_needed(value):
     if isinstance(value, datetime.datetime):
         return value
--- a/awscli/botocore/credentials.py
+++ b/awscli/botocore/credentials.py
@@ -2050,6 +2050,7 @@ class SSOCredentialFetcher(CachedCredentialFetcher):
             # raise an UnauthorizedSSOTokenError if the loaded legacy token
             # is expired to save a call to GetRoleCredentials with an
             # expired token.
+            print(f'expiresAt field to be passed to parse function: {token_dict["expiresAt"]}')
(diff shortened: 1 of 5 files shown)


ashovlin (changes requested):
1. I think it's still worth a changelog entry explaining the change in behavior.
2. The Windows test failures look valid.
3. Can you add a test case asserting that the expected error was thrown and that no calls were made to `GetRoleCredentials`?

awscli/botocore/credentials.py
```diff
@@ -2042,7 +2044,16 @@ def _get_credentials(self):
...
+            token_dict = self._token_loader(self._start_url)
+            token = token_dict['accessToken']
+
+            # raise an UnauthorizedSSOTokenError if the loaded legacy token
+            # is expired to save a call to GetRoleCredentials with an
+            # expired token.
+            expiration = dateutil.parser.parse(token_dict['expiresAt'])
+            remaining = total_seconds(expiration - self._time_fetcher())
+            if remaining <= 0:
+                raise UnauthorizedSSOTokenError()
```
aemous, line 2056: For reviewer:

On one hand, I think it's unfair to claim the token is "Unauthorized" without the server stating this, since the server is the source of truth on this.

On the other hand, if we use any other exception (e.g. a new ExpiredTokenException), then that would theoretically be a breaking change, since we're changing the exception in a code path that always (?) throws an exception.

awscli/botocore/credentials.py
```diff
@@ -1990,7 +1991,7 @@ class SSOCredentialFetcher(CachedCredentialFetcher):
     def __init__(self, start_url, sso_region, role_name, account_id,
                  client_creator, token_loader=None, cache=None,
                  expiry_window_seconds=None, token_provider=None,
-                 sso_session_name=None):
+                 sso_session_name=None, time_fetcher=_local_now):
```
ashovlin, line 1994: Double-checking, is local time okay? It looks like we save off the expiry in UTC.

aemous: The time_fetcher for this class is only used for expiration time check via subtracting two datetime objects.

My understanding is that datetime will correctly convert the datetimes even if their in different timezones, since each datetime instance is aware of what timezone it's in. 

Simple script to verify this:

```
from datetime import datetime, timezone
from dateutil import tz, parser

eastern = tz.gettz('America/New_York')
dt1 = datetime(2025, 3, 12, 10, 0, 0, tzinfo=eastern)
dt2 = parser.parse('2025-3-12T10:00:00Z')
difference = dt1 - dt2

print(f'difference: {difference.total_seconds()}')

```
Output: `difference: 14400.0`

aemous: And, my original reasoning for using tzlocal (and relying on datetime auto-conversion) instead of tzutc here is since tzlocal is what's used in this credentials file, it would be the first use of tzutc in this file.

I guess because credentials have arbitrary timezones and tokens are stored with UTC.

ashovlin (approved):

PR #9360, opened 2025-03-13
[v2] Update Sphinx from 3.0.2 to 6.2

hssyoo:
Changes:
- Updates Sphinx to 6.2
- Updates sphinx-notfound-page to 1.1.0
- Removes explicit pins on Sphinx dependencies since we're now on Sphinx>5
- Sphinx removed jQuery support. Only the searchbox uses it, so replace jQuery use with HTML
- sphinx-notfound-page dropped support for `notfound_default_language`. Replace it with `notfound_urls_prefix`

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,12 +42,6 @@ notfound_pagename = '_404'
 # to set our path prefix
 notfound_urls_prefix = '/v2/documentation/api/latest/'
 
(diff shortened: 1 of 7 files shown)


kdaily (approved):
LGTM overall!

One nit in `doc/source/conf.py`, there's a commented out reference to `notfound_default_language` on line 50 that should probably be removed also since its deprecated.

Issue #8554, opened 2024-02-28
role parameter in eks update-kubeconfig is not being used for aws cli connection


xcompass:
### Describe the bug

It seems that the role parameter in `aws eks update-kubeconfig --role arn:aws:iam::1234567890:role/ASSUMEDROLE` is only being inserted in to kube config and used for kubectl connection, but not `aws eks` command for retrieving kubeconfig.

My use case: I have 2 AWS accounts with one eks cluster in each account. I would like to manage both clusters with account A's credential without switching back and forth accounts. So I have setup a role (ASSUMEDROLE) in account B with AssumeRole permission from a role from account A. Everything works fine except the `update-kubeconfig` command for cluster B in account B. I expect to get the cluster B kubeconfig by running `aws eks update-kubeconfig --name clusterB --role arn:aws:iam::ACCOUNTB#:role/ASSUMEDROLE`, where aws cil should use the ASSUMEDROLE in account B to connect and retrieve the config and also insert the role to kubeconfig user `get-token` command.

Currently, I have to create a new AWS profile and specify the role_arn to ASSUMEDROLE and source as account A profile and run `aws eks update-kubeconfig --name clusterB --role arn:aws:iam::ACCOUNTB#:role/ASSUMEDROLE --profile=NEWPROFILE` to get the config. However, it will also double assumes the role as the NEWPROFILE environment var is also added to kubeconfig generated by the command.
```
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - ca-central-1
      - eks
      - get-token
      - --cluster-name
      - clusterB
      - --role
      - arn:aws:iam::ACCOUNTB#:role/ASSUMEDROLE
      - --output
      - json
      command: aws
      env:
      - name: AWS_PROFILE
        value: newprofile
```
The workaround is to remove `--role` in the `update-kubeconfig` command. However, I would like to just use a single profile. If the `--role` parameter is actually being used for aws eks connection, it would solve this problem. 

This is also an inconsistent behavior than other commands. e.g. `aws eks get-token --role`, where the role was actually used for the `aws eks` command connection. 

Others are also having into this issue: 
https://github.com/aws/aws-cli/issues/5823
https://github.com/aws/aws-cli/issues/6389

### Expected Behavior

[...]

PR #9364, opened 2025-03-13
Assume Role to find EKS cluster for config update

5nafu:
*Issue #, if available: #8554*

*Description of changes:*

This PR will assume a given role when trying to update the kubeconfig.

While `aws eks update-kubeconfig --role-arn...` would add the required parameters to the user section of the configfile, but will fail finding the cluster if the original profile does not have access. This PR will fix the issue and assume the role before getting the cluster description.

### DIFF
--- a/awscli/customizations/eks/update_kubeconfig.py
+++ b/awscli/customizations/eks/update_kubeconfig.py
@@ -250,15 +250,15 @@ class EKSClient(object):
         describe-cluster will only be called once.
         """
         if self._cluster_description is None:
-            if self._parsed_globals is None:
-                client = self._session.create_client("eks")
-            else:
-                client = self._session.create_client(
-                    "eks",
-                    region_name=self._parsed_globals.region,
-                    endpoint_url=self._parsed_globals.endpoint_url,
-                    verify=self._parsed_globals.verify_ssl
-                )
+            client_kwargs = {}
+            if self._parsed_globals is not None:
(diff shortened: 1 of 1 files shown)


djtung:
Hi, I'm on the team working on EKS at AWS. As I understand:

1. The current documentation specifies that --role-arn is meant for "cluster authentication" (kubectl connection), not for the AWS CLI's cluster discovery process.
2. This fix overloads the --role-arn parameter by using it for both:
    - AWS CLI cluster discovery
    - Kubernetes cluster authentication
3. This change would be breaking for existing customers who may have tools/scripts relying on the current documented behaviour

Better would be to implement the suggestions from [here](https://github.com/aws/aws-cli/issues/8554#issuecomment-2115174034), perhaps with the latter of providing another parameter so that the two roles can be clearly separated. If you're willing to make that change for adding a new parameter (and update the documentation), we'd be happy to approve it.

sdomme:
HI @djtung, thank you very much for your feedback. We reworked this contribution in https://github.com/aws/aws-cli/pull/9443 in favour to your suggestions. We confirmed with @5nafu, that this PR is superseded by the new request.
Please have a look. Thnaks

aemous:
> We confirmed with @5nafu, that this PR is superseded by the new request.

Closing this PR as it has been superseded, and the reworked version has been merged.

PR #9375, opened 2025-03-17
Run pre-commit on CLI codebase

kdaily:
*Issue #, if available:*

*Description of changes:*

Runs the entire pre-commit hook for linting and formatting on the CLI codebase, including `ruff check` fixes. This still manually excludes the vended `botocore` and `s3transfer` packages and tests.

~It also adds an exclusion for the Sphinx guzzle theme (`doc/source/guzzle_sphinx_theme`).~ (added in 0d734345cdf0c8c1e518493937dc650061f253e1)


scripts/installers/make-exe
```diff
@@ -124,16 +126,15 @@ def main():
...
-        )
+        help=('Path to ac.index file to include in the exe.'),
     )
     args = parser.parse_args()
 
     output = os.path.abspath(args.output)
     if args.src_dir:
         print(
-            'Installing dependencies from local directory: %s' % args.src_dir)
+            'Installing dependencies from local directory: %s' % args.src_dir
```
kdaily, line 136:
```suggestion
            f'Installing dependencies from local directory: {args.src_dir}'
```

scripts/installers/make-macpkg
```diff
@@ -45,27 +46,30 @@ def stage_files(workdir):
...
-        ))
+        )
+    )
+    with tmp_dir() as formatted_resource_dir:
+        render_resources(
+            formatted_resource_dir, RESOURCES_DIR, {'version': version}
+        )
+        print(
+            run(
+                ('productbuild --distribution %s ' '--resources %s ' '%s')
```
kdaily, line 68:
```suggestion
                ('productbuild --distribution %s --resources %s %s')
```

kdaily:
I intentionally left my own suggested changes unresolved as breadcrumbs. Will merge all of those after a 2P review.

hssyoo:
FYI as I'm going through the changes I'm reacting 👍 / 👎 if I agree/disagree with your suggestion commits.

awscli/customizations/servicecatalog/generatebase.py
```diff
@@ -33,7 +33,7 @@ def _run_main(self, parsed_args, parsed_globals):
             self.s3_uploader.upload(
                 parsed_args.file_path, get_s3_path(parsed_args.file_path)
             )
-        except OSError as ex:
+        except OSError:
             raise RuntimeError("%s cannot be found" % parsed_args.file_path)
```
hssyoo, line 37: Is ruff not changing these to f-strings?

hssyoo (approved):
Looks good, just had 1 non-blocking comment. I think in general, if ruff fixed 1 type of formatting, say splitting a long line into multiple lines, then it won't fix another formatting issue within that, say converting to f-strings. Does running ruff multiple times generate diffs between each ruff call?

Assuming internal builds succeed, 🏆

kdaily:
> Does running ruff multiple times generate diffs between each ruff call?

No; it won't ever introduce changes that would break the current linting/formatting. There are many more fixes that are identified that aren't fixable by default with the current settings (`ruff check --fix`). To adjust f-strings in long lines or some of the more complicated f-strings that use string formatting, we can run `ruff check --fix --unsafe-fixes`, or examine the stderr output of the call to see where else there are things that can be improved.

Also, from the ruff pre-commit hooks repo:

> When running with --fix, Ruff's lint hook should be placed before Ruff's formatter hook, and before Black, isort, and other formatting tools, as Ruff's fix behavior can output code changes that require reformatting.
> 
> When running without --fix, Ruff's formatter hook can be placed before or after Ruff's lint hook.
> 
> (As long as your Ruff configuration avoids any [linter-formatter incompatibilities](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules), ruff format should never introduce new lint errors, so it's safe to run Ruff's format hook after ruff check --fix.)

awscli/customizations/servicecatalog/generatebase.py
kdaily: I think it would introduce a formatting exception, so it would need to be done manually. I've intentionally chosen to let the tool do most of the work, and can clean up as we go anything else. I manually made the changes where I knew the line length would be fine with the trivial change, but only those.

kdaily:
Also, for altering the f-string usage in log statements, I ff-strings are used in logging, Python will always compute the string, even if the log level isn't set to output it. If the old syntax is used (`log.debug("Value is %s", foo)`) then the string will only be computed if the log level actually will output it.

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -234,12 +234,10 @@ def choose_kubeconfig(self, cluster_name):
                 loaded_config = self._loader.load_kubeconfig(candidate_path)
 
                 if loaded_config.has_cluster(cluster_name):
-                    LOG.debug(
-                        "Found entry to update at {0}".format(candidate_path)
-                    )
+                    LOG.debug(f"Found entry to update at {candidate_path}")
```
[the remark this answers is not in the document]
kdaily: Discussed below to not use f-strings in log statements.

awscli/customizations/eks/update_kubeconfig.py
```diff
@@ -234,12 +234,10 @@ def choose_kubeconfig(self, cluster_name):
...
                     return loaded_config
             except KubeconfigError as e:
-                LOG.warning("Passing {0}:{1}".format(candidate_path, e))
+                LOG.warning(f"Passing {candidate_path}:{e}")
```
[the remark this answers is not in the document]
kdaily: Discussed below to not use f-strings in log statements.

awscli/customizations/flatten.py
```diff
@@ -67,7 +67,7 @@ def add_to_params(self, parameters, value):
         cli_type = self._container.cli_type_name
         key = self._property
 
-        LOG.debug('Hydrating {0}[{1}]'.format(container, key))
+        LOG.debug(f'Hydrating {container}[{key}]')
```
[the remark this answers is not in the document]
kdaily: Discussed below to not use f-strings in log statements.

awscli/customizations/flatten.py
```diff
@@ -231,16 +231,14 @@ def _find_nested_arg(self, argument, name):
         """
         if SEP in name:
             # Find the actual nested argument to pull out
-            LOG.debug('Finding nested argument in {0}'.format(name))
+            LOG.debug(f'Finding nested argument in {name}')
```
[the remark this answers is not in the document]
kdaily: Discussed below to not use f-strings in log statements.

awscli/customizations/servicecatalog/generatebase.py
kdaily: Hmm, this one I'm not sure, will double check. Line length would actually be the same.

kdaily: Unsure why it didn't want to fix it, but its noted in the output:

```
$ ruff check awscli/customizations/servicecatalog/generatebase.py
awscli/customizations/servicecatalog/generatebase.py:37:32: UP031 Use format specifiers instead of percent format
   |
35 |             )
36 |         except OSError:
37 |             raise RuntimeError("%s cannot be found" % parsed_args.file_path)
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP031
38 |
39 |     def get_and_validate_region(self, parsed_globals):
   |
   = help: Replace with format specifiers
```

There are still 1257 string formatting improvements to fix that would need to be done manually or with the `--unsafe-fixes` flag. Running it with that flag replaces nearly all of the changes with `'format()` usage, not f-strings, so I'm not inclined to do that at the moment.

awscli/customizations/cloudformation/artifact_exporter.py
```diff
@@ -247,9 +244,7 @@ def export(self, resource_id, resource_dict, parent_dir):
 
         if isinstance(property_value, dict):
             LOG.debug(
-                "Property {0} of {1} resource is not a URL".format(
-                    self.PROPERTY_NAME, resource_id
-                )
+                f"Property {self.PROPERTY_NAME} of {resource_id} resource is not a URL"
```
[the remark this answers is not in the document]
kdaily: Fixed in acd932948997ad3cfffedf89f0df4a3b063aadd6

kdaily:
Internal build successful.

PR #9384, opened 2025-03-18
Relax pre-commit config to run on whole codebase

kdaily:
*Issue #, if available:*

*Description of changes:*

This change allows pre-commit to run everywhere. The exclusion list is an intersection of the configuration from both the upstream botocore and s3transfer repositories.

The pre-commit hook is still not enabled by default, and is not run in GitHub CI.


aemous (approved):
LGTM

PR #9385, opened 2025-03-18
run pre-commit on s3transfer

kdaily:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/unit/s3transfer/test_s3transfer.py
+++ b/tests/unit/s3transfer/test_s3transfer.py
@@ -33,6 +33,7 @@ from s3transfer import (
     random_file_extension,
 )
 from s3transfer.exceptions import RetriesExceededError, S3UploadFailedError
+
 from tests import mock, unittest
 
 
@@ -377,7 +378,6 @@ class TestMultipartUploader(unittest.TestCase):
 
 
 class TestMultipartDownloader(unittest.TestCase):
-
     maxDiff = None
 
     def test_multipart_download_uses_correct_client_calls(self):
@@ -435,7 +435,7 @@ class TestMultipartDownloader(unittest.TestCase):
         response_body = b'foobarbaz'
         stream_with_errors = mock.Mock()
         stream_with_errors.read.side_effect = [
-            socket.error("fake error"),
+            OSError("fake error"),
             response_body,
         ]
         client.get_object.return_value = {'Body': stream_with_errors}
@@ -470,7 +470,7 @@ class TestMultipartDownloader(unittest.TestCase):
         client = mock.Mock()
         response_body = b'foobarbaz'
         stream_with_errors = mock.Mock()
-        stream_with_errors.read.side_effect = socket.error("fake error")
+        stream_with_errors.read.side_effect = OSError("fake error")
         client.get_object.return_value = {'Body': stream_with_errors}
         config = TransferConfig(multipart_threshold=4, multipart_chunksize=4)
 
@@ -679,7 +679,7 @@ class TestS3Transfer(unittest.TestCase):
         }
         self.client.get_object.side_effect = [
             # First request fails.
-            socket.error("fake error"),
+            OSError("fake error"),
             # Second succeeds.
             {'Body': BytesIO(b'foobar')},
         ]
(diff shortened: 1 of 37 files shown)


tests/unit/s3transfer/test_s3transfer.py
```diff
@@ -435,7 +435,7 @@ def test_retry_on_failures_from_stream_reads(self):
         response_body = b'foobarbaz'
         stream_with_errors = mock.Mock()
         stream_with_errors.read.side_effect = [
-            socket.error("fake error"),
+            OSError("fake error"),
```
aemous, line 438: any justification for why this change is safe?

tests/unit/s3transfer/test_s3transfer.py
```diff
@@ -470,7 +470,7 @@ def test_exception_raised_on_exceeded_retries(self):
         client = mock.Mock()
         response_body = b'foobarbaz'
         stream_with_errors = mock.Mock()
-        stream_with_errors.read.side_effect = socket.error("fake error")
```
aemous, line 473: Similar question; why's this safe?

tests/unit/s3transfer/test_s3transfer.py
```diff
@@ -679,7 +679,7 @@ def test_get_object_stream_is_retried_and_succeeds(self):
         }
         self.client.get_object.side_effect = [
             # First request fails.
-            socket.error("fake error"),
```
aemous, line 682: same question for all instances of this change

tests/unit/s3transfer/test_utils.py
```diff
@@ -329,7 +329,7 @@ def test_allocate(self):
 
     @mock.patch('s3transfer.utils.fallocate')
     def test_allocate_with_io_error(self, mock_fallocate):
-        mock_fallocate.side_effect = IOError()
+        mock_fallocate.side_effect = OSError()
         with self.assertRaises(IOError):
```
aemous, line 333: should be worth (and safe) swapping this IoError for oserror too. And also renaming to reflect this

aemous (changes requested):
a couple small requested changes.

One overarching concern I have is, will some of these cause us to deviate from boto/s3transfer? Do we want to match our formatting options to be identical to the ones used in that repo? For whitespaces it may not be a big deal, but for reordering imports and swapping interchangeable exceptions I wonder if this becomes a concern

tests/unit/s3transfer/test_s3transfer.py
kdaily: From https://docs.python.org/3/library/socket.html#socket.error:

> exception socket.error
> 
>    A deprecated alias of [OSError](https://docs.python.org/3/library/exceptions.html#OSError).
> 
>    Changed in version 3.3: Following [PEP 3151](https://peps.python.org/pep-3151/), this class was made an alias of [OSError](https://docs.python.org/3/library/exceptions.html#OSError).

kdaily:
Regarding `socket.error` to `OSError`:

Also, just to check its really an alias (confirmed this back to Python 3.8):

```
>>> import socket
>>> raise socket.error
  File "<input>", line 1, in <module>
    raise socket.error
OSError
```

tests/unit/s3transfer/test_utils.py
kdaily: Sure; raising an `IOError` or `OSError` result in identical behavior.

```
>>> try:
...     raise OSError
... except IOError:
...     print("hooray")
...
...
hooray
>>> try:
...     raise IOError
... except OSError:
...     print("hooray")
...
...
hooray
```

kdaily:
> will some of these cause us to deviate from boto/s3transfer

In the short term, yes, only because we enabled Ruff to do the import sorting (instead of `isort`). Offline discussions resulted in us going ahead with the more modern approach and the `boto` repositories (`botocore` and `s3transfer`) will catch up. For `botocore`, there is already a customer PR to modernize import sorting:

https://github.com/boto/botocore/pull/3380

See discussion when enabling ruff import sorting on the CLI repository here:

- https://github.com/aws/aws-cli/pull/9123#discussion_r1904712331
- https://github.com/aws/aws-cli/pull/9123#discussion_r1904752369

kdaily:
Force-pushed to pull in latest `v2` changes, required for running internal build.

tests/unit/s3transfer/test_utils.py
aemous: I agree, I just figured if we're replacing the deprecated IOError in line 332, that we may as well also do a similar swap on 333.

I understand they're functionally equivalent.

Also non-blocking, feel free to resolve if you disagree

aemous (approved):
LGTM. Have a couple non-blocking unresolved comments if you agree with them, otherwise I'm good with merging this

Issue #9234, opened 2025-01-19
AWS CLI v2 support for Python 3.13


aureq:
### Describe the bug

Hi,

I'm unable to install the AWS CLI version 2 using Python 3.13.1 by running `/usr/local/bin/pip install https://github.com/aws/aws-cli/archive/refs/tags/2.23.2.tar.gz`  (this work fine on Python 3.12.x, and this works fine with AWS CLI 1.37.2 too).

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The AWS  CLI v2 should install as expected on Python 3.13.x.

### Current Behavior

The build fails with the following errors and warnings

```
Collecting https://github.com/aws/aws-cli/archive/refs/tags/2.23.2.tar.gz
  Using cached https://github.com/aws/aws-cli/archive/refs/tags/2.23.2.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... error
  error: subprocess-exited-with-error
  
  × pip subprocess to install backend dependencies did not run successfully.
  │ exit code: 1
  ╰─> [518 lines of output]
      Collecting distro<1.9.0,>=1.5.0
        Using cached distro-1.8.0-py3-none-any.whl.metadata (6.9 kB)
      Collecting docutils<0.20,>=0.10
        Using cached docutils-0.19-py3-none-any.whl.metadata (2.7 kB)
      Collecting awscrt==0.23.4
        Using cached awscrt-0.23.4-cp313-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (4.1 kB)
      Collecting urllib3<1.27,>=1.25.4
        Using cached urllib3-1.26.20-py2.py3-none-any.whl.metadata (50 kB)
      Collecting colorama<0.4.7,>=0.2.5
        Using cached colorama-0.4.6-py2.py3-none-any.whl.metadata (17 kB)
      Collecting zipp<3.21.0
        Using cached zipp-3.20.2-py3-none-any.whl.metadata (3.7 kB)
      Collecting ruamel.yaml.clib<=0.2.8,>=0.2.0
        Using cached ruamel.yaml.clib-0.2.8.tar.gz (213 kB)
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'done'
        Preparing metadata (pyproject.toml): started
[...]

PR #9391, opened 2025-03-21
[v2] Add support for Python 3.13

hssyoo:
Adds support for Python 3.13 and updates the bundled Python interpreter in installers to 3.13.2.

- Due to dependency conflicts, Python 3.8 support needs to be dropped.
- Upgrade `ruamel.yaml.clib` to 0.2.12 for Python 3.13 support.
  - Wheels aren't available for macOS < 14 on x86. Opened a ticket here, but it should be fine to require source install since macOS comes with C compilers: https://sourceforge.net/p/ruamel-yaml-clib/tickets/43/
- Remove explicit dependency on `zipp` since the latest version supports Python 3.9+.
- Some packages installed from source will have their site packages directories replace `.` with `_`. This causes some dependency tests to fail because it'll make an assertion like `assert ruamel.yaml.clib == ruamel_yaml_clib`. If it fails, we normalize the package name and try 1 more time.
- Python 3.13 no longer considers `/` an absolute path in Windows, which is what `os.path.sep` returns. For tests, we wrap it in `os.path.abspath`, converting it to `C://`.
  - https://docs.python.org/3/whatsnew/3.13.html
  - > On Windows, [isabs()](https://docs.python.org/3/library/os.path.html#os.path.isabs) no longer considers paths starting with exactly one slash (\ or /) to be absolute. (Contributed by Barney Gale and Jon Foster in [gh-44626](https://github.com/python/cpython/issues/44626).)


AlexandreEXFO:
https://github.com/aws/aws-cli/issues/9234

PR #9409, opened 2025-03-28
Use awscrt for cryptographic functionality in EC2 Instance Connect

kdaily:
*Issue #, if available:*

*Description of changes:*

This is a continuation of the work started in https://github.com/aws/aws-cli/pull/9136.

This PR replaces cryptographic methods from the cryptography package with corresponding methods from the awscrt package for the AWS EC2 Instance Connect command:

- `aws ec2instanceconnect ssh`


awscli/customizations/ec2instanceconnect/ssh.py
```diff
@@ -448,19 +442,23 @@ def _ssh(
 class KeyManager:
     def generate_key(self):
         logger.debug('Generate Ed25519 Key')
-        return Ed25519PrivateKey.generate()
+        return ED25519.new_generate()
+
+    def _get_public_key(self, private_key):
+        return f"ssh-ed25519 {ED25519.export_public_key(private_key, export_format=ED25519ExportFormat.OPENSSH_B64).decode()}"
```
kdaily, line 448: This should probably call `.encode()` on the resulting string to return bytes like `public_bytes` did.

kdaily: Fixed in cc20e49cb32730e13930c74cd03654e3c18b432d

ashovlin (approved):
Tested from my local Mac to an instance both with and without `--private-key-file`, both worked.

PR #9420, opened 2025-04-03
Fix installing setup dependencies for bundled installer on newer versions of pip

ashovlin:
*Issue #, if available:* CLI-6153

Don't think we have a GitHub issue, but this has been reported on a few Q&A sites:
1. https://askubuntu.com/questions/1513592/installing-aws-cli-v1-on-ubuntu-24-does-not-work-but-works-fine-on-ubuntu-22 
2. https://stackoverflow.com/questions/77658049/getting-pip-error-while-installing-aws-cli

*Description of changes:*
1. When installing the CLI V1 via the bundled installer, we use the `--find-links file://{local path}` syntax to install the packages from the bundle, rather than PyPI. At some point (not apparent from pip's changelog, still trying to pinpoint), this started throwing `ValueError: non-local file URIs are not supported on this platform: 'file://.`. 
    * To address this, I removed the `file://` syntax since `--find-links` can accept local paths (see [example 10](https://pip.pypa.io/en/stable/cli/pip_install/#examples) on `pip install` docs).
2. When testing this on 3.12, installing the bundle fails with `ModuleNotFoundError: No module named 'setuptools'` since it's no longer included by default.
    * To address this, I added `setuptools` to our `BUILDTIME_DEPS` that we install first.
    * I struggled to install `setuptools` from sdist, which looks like it needs to bootstrap itself. I switched to wheels to alleviate this, though just for the `BUILDTIME_DEPS` to start. https://github.com/aws/aws-cli/pull/4756 alluded to this as future work.
3. I also pinned `urllib3` for the bundled installer to a version before `2.0.0` (where they started requiring `hatchling`). We previously pinned `colorama` for a similar reason.

### Testing
Added new GitHub action to test making the bundle and installing it across our usual Python version + OS matrix.

I'll manually test a dry-run through the internal build system.

### DIFF
--- a/scripts/ci/test-bundle
+++ b/scripts/ci/test-bundle
@@ -5,23 +5,12 @@
 
 import os
 import re
-from contextlib import contextmanager
 from subprocess import check_output
+from awscli.testutils import cd
 
 _dname = os.path.dirname
 
 REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
-
-
-@contextmanager
-def cd(path):
-    """Change directory while inside context manager."""
-    cwd = os.getcwd()
-    try:
-        os.chdir(path)
-        yield
-    finally:
-        os.chdir(cwd)
 
 
 def run(command):
--- a/scripts/ci/test-bundle
+++ b/scripts/ci/test-bundle
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+# Don't run tests from the root repo dir.
+# We want to ensure we're importing from the installed
+# binary package not from the CWD.
+
+import os
+import re
+from contextlib import contextmanager
+from subprocess import check_output
+
+_dname = os.path.dirname
+
+REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
+
+
+@contextmanager
+def cd(path):
+    """Change directory while inside context manager."""
+    cwd = os.getcwd()
+    try:
+        os.chdir(path)
+        yield
+    finally:
+        os.chdir(cwd)
+
+
+def run(command):
+    print(f'Running {command}')
+    return check_output(command, shell=True)
+
+
+def run_make_bundle():
+    """
+    Builds the bundled installer, and returns its path
+    """
+    output = run(f'{REPO_ROOT}/scripts/make-bundle')
+    match = re.search(
+        r'Zipped bundle installer is at: (.+?\.zip)', output.decode('utf-8')
+    )
+    if not match:
+        raise RuntimeError("Could not find bundle path in make-bundle output")
+
+    return match.group(1)
+
+
+def install_from_bundle(zip_path):
+    run(f'unzip -o {bundle_path}')
+    path_without_zip = bundle_path[:-4]
+    run(
+        f'sudo {path_without_zip}/install -i /usr/local/aws -b /usr/local/bin/aws'
+    )
+
+
+def verify_installation():
+    version_output = run("aws --version")
+    print(f"Installed AWS CLI version: {version_output}")
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


scripts/ci/test-bundle
```diff
@@ -0,0 +1,63 @@
...
+from contextlib import contextmanager
+from subprocess import check_output
+
+_dname = os.path.dirname
+
+REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
+
+
+@contextmanager
+def cd(path):
```
aemous, line 17: I think we have an equivalent function in `awscli/testutils.py`. Is it worth importing that one to minimize duplicate code?

scripts/make-bundle
```diff
@@ -84,6 +87,16 @@ def download_package_tarballs(dirname, packages):
             ))
 
 
+def download_package_wheels(dirname, packages):
+    with cd(dirname):
+        for package, package_version in packages:
+            run(
+                '%s -m pip download %s==%s --no-build-isolation '
```
aemous, line 94: nit: I know % is consistent with the rest of this file, but can we use f-strings for new code/functions?

scripts/make-bundle
```diff
@@ -174,9 +187,10 @@ def main():
     # manually install them. We isolate them to a particular directory so we
     # can run the install before the things they're dependent on. We have to do
     # this because pip won't actually find them since it doesn't handle build
-    # dependencies.
+    # dependencies. We use wheels for this, to avoid bootstrapping setuptools
+    # in 3.12+ where it's no longer included by default.
     setup_dir = os.path.join(package_dir, 'setup')
-    download_package_tarballs(
+    download_package_wheels(
```
aemous, line 193: Have we verified that each package in `BUILDTIME_DEPS` publishes wheels to PyPi supporting the Linux & Mac versions we support?

Risk: one of those packages may stop publishing wheels for our supported Linux / Mac versions, or drop support for a version we support before we do, which might break our installer again. What's our stance on this risk?

aemous (changes requested):
Looks good. I left a couple comments/questions.

scripts/ci/test-bundle
ashovlin: Moved to an import in https://github.com/aws/aws-cli/pull/9420/commits/0408ebe220a819aa76f6eebb4b608cf434b8fc92

scripts/make-bundle
ashovlin: Ran `uvx ruff format` over the file in https://github.com/aws/aws-cli/pull/9420/commits/0408ebe220a819aa76f6eebb4b608cf434b8fc92

scripts/make-bundle
ashovlin: Added `validate_that_wheels_are_universal` in https://github.com/aws/aws-cli/pull/9420/commits/0408ebe220a819aa76f6eebb4b608cf434b8fc92 to validate this while building the bundle.

aemous (approved):
LGTM

scripts/make-bundle
```diff
@@ -79,17 +83,33 @@ def create_scratch_dir():
...
+def download_package_wheels(dirname, packages):
+    with cd(dirname):
+        for package, package_version in packages:
+            run(
+                f'{sys.executable} -m pip download {package}=={package_version}'
+                f' --only-binary :all:'
+            )
+
+
+def validate_that_wheels_are_universal(dirname):
```
hssyoo, line 101: non-blocking: I don't think it's required that the wheels are universal, only that wheels are available for the platform the user is installing AWS CLI on. It doesn't look like this will be an issue anytime soon and if it is, it'll be caught the by CI so I'm fine with this.

hssyoo (approved):
LGTM 🏆

Issue #9433, opened 2025-04-10
aws ecr describe-repositories fails with linking error


codeimpossible:
### Describe the bug

Our CI/CD pipeline began failing roughly when v2.26 was published with the following error:

```bash
[Publishing Runtime Image] Checking if ECR repository exists
[15:52:40][Publishing Runtime Image] less: /usr/local/aws-cli/v2/2.26.0/dist/libtinfo.so.5: no version information available (required by less)
[15:52:40][Publishing Runtime Image] 
[15:52:40][Publishing Runtime Image] less: Symbol `ospeed' has different size in shared object, consider re-linking
[15:52:40][Publishing Runtime Image] 
[15:52:40][Publishing Runtime Image] WARNING: terminal is not fully functional
```

The command we're running is `aws ecr describe-repositories --repository-names $service_lower --region us-east-1`

It seems the latest cli release may not have the correct dependencies?

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

For the command to work and return whether or not a repository exists.

### Current Behavior

It throws the exception mentioned and hangs, waiting for user input. 

### Reproduction Steps

run `aws ecr describe-repositories --repository-names <repo_name> --region us-east-1`

### Possible Solution

Revert to 2.25

### CLI version used

2.2.6

### Environment details (OS name and version, etc.)

ubuntu 20.x

codeimpossible:
Our workaround ended up being to downgrade to v2.25.13

hssyoo:
Thanks for opening this issue. Version 2.26.0 upgraded the bundled Python interpreter from 3.12 to 3.13. This caused the Linux installers to dynamically link `libtinfo.so.5`, which is bundled with the Linux installers.

Can you provide more details on the environment? Specifically, are there configurations set that may be changing dynamic linking behavior? Ubuntu is part of our build system's test matrix, I'm wondering what differences in the environment may be the cause.

codeimpossible:
hi @hssyoo , I originally listed Ubuntu 20.x as the version our CI was running, but i was in the middle of firefighting and that's not correct. We run our CI in a containerized pipeline and the container this failed under is `ubuntu:18.04`. I've included the relevant lines from our dockerfile below.

```dockerfile
FROM ubuntu:18.04

## Install Wget, Gpg, Curl
RUN apt-get update
RUN apt-get install -y --no-install-recommends \
    wget \
    apt-transport-https ca-certificates curl \
[...]

PR #9434, opened 2025-04-11
[v2] Remove PyInstaller LD_LIBRARY_PATH from pager

hssyoo:
Fixes https://github.com/aws/aws-cli/issues/9433

The AWS CLI calls out to the system's pager, which may share dynamic libraries with the AWS CLI. This can cause issues if an AWS CLI library is incompatible with the system's pager. Although the system owns the pager, because it's executed as a child of the AWS CLI's Python process, it inherits PyInstaller's `LD_LIBRARY_PATH`. This commit fixes the issue by retrieving kwargs for pager popen under the `original_ld_library_path` context manager. The effect is that the system's pager uses the the system's dependencies instead of the AWS CLI's.

### DIFF
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -64,10 +64,7 @@ class LazyPager:
     def __init__(self, popen, **kwargs):
         self._popen = popen
         self._popen_kwargs = kwargs
-        # When calling out to the system's pager, we want to avoid using
-        # shared libraries bundled with the AWS CLI so that the pager uses
-        # the system's shared libraries.
-        self._popen_kwargs['env'].pop('LD_LIBRARY_PATH', None)
+        self._replace_or_remove_ld_library_path()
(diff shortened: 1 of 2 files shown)


awscli/utils.py
```diff
@@ -64,6 +64,10 @@ class LazyPager:
     def __init__(self, popen, **kwargs):
         self._popen = popen
         self._popen_kwargs = kwargs
+        # When calling out to the system's pager, we want to avoid using
+        # shared libraries bundled with the AWS CLI so that the pager uses
+        # the system's shared libraries.
+        self._popen_kwargs['env'].pop('LD_LIBRARY_PATH', None)
```
hssyoo, line 70: Thinking about it now we probably want to use the original path if it exists.

ashovlin (approved):

PR #9450, opened 2025-04-18
Alias socialmessaging command

SamRemis:
Due to a human error while customizing the name of a socialmessaging command, we are creating an alias to allow both the old name and the newly corrected name to work.  This generates the correct docs as shown below.  This must be released on the same day as the associated botocore PR.  

image


alexgromero (approved):
LGTM 🚢

PR #9458, opened 2025-04-24
Fix incorrect socialmessaging command

SamRemis:
This PR fixes a relatively new command that had a human error due to manually renaming it to preserve branding.  All customers currently using this command are aware of this breaking change.  

### DIFF
--- a/.changes/next-release/bugfix-socialmessaging-6243.json
+++ b/.changes/next-release/bugfix-socialmessaging-6243.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "socialmessaging",
+  "description": "Fixes incorrect customization for renaming a socialmessaging command"
+}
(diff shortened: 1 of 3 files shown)


.changes/next-release/bugfix-socialmessaging-6243.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "socialmessaging",
+  "description": "Fixes incorrect customization for renaming a socialmessaging command"
```
ashovlin, line 4: I'd describe what is changing from a user's perspective, since "customization for renaming" is internal.

ashovlin (commented):
Why a different approach than https://github.com/aws/aws-cli/pull/9450?

SamRemis:
> Why a different approach than #9450?

The usage of this command in CliV1 and botocore is significantly higher, so renaming it in a braking way for them is not an option.  Usage is low enough in CLIv2 that we can fix this at the source while the service is still early on; all customers currently calling this command have already been notified of this breaking change, so we can safely make it.

ashovlin (approved):

Issue #9440, opened 2025-04-14
OpenSSL 3.0.16


plancksecond:
### Describe the feature

Update OpenSSL to 3.0.16 for cve-2024-13176 and cve-2024-9143

### Use Case

Security patch

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

Latest

### Environment details (OS name and version, etc.)

Windows

adev-code:
Hello, thanks for raising the feature request. The team will update OpenSSL but we cannot provide an ETA for release.  Please refer to an AWS CLI changelog for updates about this going forward https://raw.githubusercontent.com/aws/aws-cli/v2/CHANGELOG.rst . Please let me know if you have any other questions. Thanks.

plancksecond:
I checked the latest version and OpenSSL has not been updated. It is still 3.0.15

aemous:
Closing issue as the latest (`2.27.11` and later) AWS CLI v2 Windows and Mac installers now ship with OpenSSL 3.0.16 bundled.

PR #9462, opened 2025-04-28
[v2] Upgrade bundled interpreter to Python 3.13.3

aemous:
*Issue #, if available:*

- https://github.com/aws/aws-cli/issues/9440 

*Description of changes:*

- Upgrade bundled Python interpreter to 3.13.3
- Regenerate sdist lockfiles

*Description of tests:*

- Ran all test suites and CI


kdaily (commented):
This looks good, and would probably be safe. In the past, we've only added a changelog and not updated transitive dependencies (looks like these are dependencies of `pyinstaller`). 

https://github.com/aws/aws-cli/pull/9309

aemous:
> This looks good, and would probably be safe. In the past, we've only added a changelog and not updated transitive dependencies (looks like these are dependencies of `pyinstaller`).
> 
> #9309

I just ran the regenerate lock files github action, should we split it to a separate PR? We may have missed regenerating in a previous dependency update @kdaily

aemous:
@kdaily As discussed offline, I removed the lockfile updates.

kdaily (approved):

PR #9464, opened 2025-04-28
CI: Only run stale actions for upstream

hugovk:
*Description of changes:*

For forks with GitHub Actions enabled, there's no need to run the scheduled stale ones.

Let's disable them for forks.

See also https://hugovk.dev/blog/2023/til-how-to-disable-cron-for-github-forks/


nateprewitt (approved):
Thanks @hugovk!

ashovlin (approved):

PR #9470, opened 2025-04-30
Add initial Q Developer customization

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* Adds the initial commands for integration with the [Amazon Q Developer CLI](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line.html).

This is targeting a feature branch, future PRs will cover downloading the extension executable. Command naming may change as well.

For now if you want to test it, install it following https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html, then copy the `q` executable from the installed path to `~/.aws/cli/extensions/q/q`.


ashovlin:
In https://github.com/aws/aws-cli/pull/9470/commits/594184674393a9b86bc8b139ce5e8a7f4907a409, updated the main entrypoint to use `aws --agent-mode` instead of `aws q chat` after internal discussions. 

We'll likely move the "lifecycle" commands as well from `aws q` to `aws configure` or possibly a new `aws extensions`, but that'll be in a future PR.

awscli/customizations/q/utils.py
```diff
@@ -0,0 +1,26 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from pathlib import Path
+
+EXTENSIONS_DIR = Path.home() / '.aws' / 'cli' / 'extensions'
```
nateprewitt, line 16: Do we need to consider the `AWS_DATA_PATH` here as well if they've configured something other than `~/.aws/cli/...`?

PR #9471, opened 2025-04-30
Skip bundled installer tests on pushes to develop

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* #9420 recently added a new GitHub Action that runs tests for the CLI V1 bundled installer for Mac and Linux.

During the daily release, we push to GitHub before PyPI. These tests are kicked off, but then fail because the new botocore isn't in PyPI yet.

Example failure: https://github.com/aws/aws-cli/actions/runs/14739584274

I'll look at the internal canaries as a better spot to test that the mainline branch is working.

### DIFF
--- a/.github/workflows/run-bundle-test.yml
+++ b/.github/workflows/run-bundle-test.yml
@@ -2,6 +2,8 @@ name: Run bundle test
 
 on:
   push:
+    branches-ignore:
+      - develop # the daily release reaches GitHub before PyPI, and these fail in that window
   pull_request:
     branches-ignore: [ master ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/run-bundle-test.yml
```diff
@@ -2,6 +2,8 @@ name: Run bundle test
 
 on:
   push:
+    branches-ignore:
+      - develop # the daily release reaches GitHub before PyPI, and these fail in that window
```
jonathan343, line 6: We should also add `master` to this list right?

ashovlin: Added via https://github.com/aws/aws-cli/pull/9471/commits/80a12258718206484e0493cd2a175fedec6b9aaa, forgot we still push there.

jonathan343 (approved):
:shipit:

PR #9475, opened 2025-05-02
Update new-change script enhancement description.

aemous:
*Description of changes:*

* Update new-change file enhancement's description for more clarity.

### DIFF
--- a/scripts/new-change
+++ b/scripts/new-change
@@ -57,7 +57,7 @@ TEMPLATE = """\
 # feature: A larger feature or change in behavior, usually resulting in a
 #          minor version bump.
 # bugfix: Fixing a bug in an existing code path.
-# enhancement: Adding a small feature or a change to an underlying 
+# enhancement: An incremental improvement or a change to an underlying 
            implementation detail.
 # api-change: Changes to a modeled API.
 type: {change_type}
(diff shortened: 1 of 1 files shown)


scripts/new-change
```diff
@@ -57,7 +57,8 @@ TEMPLATE = """\
 # feature: A larger feature or change in behavior, usually resulting in a
 #          minor version bump.
 # bugfix: Fixing a bug in an existing code path.
-# enhancement: Small change to an underlying implementation detail.
+# enhancement: An added small feature, or small change to an underlying 
+           implementation detail.
```
kdaily, line 61: nit: we can try to keep the same voice for each of these. How about:

```suggestion
# enhancement: Adding a small feature or a change to an underlying 
           implementation detail.
```

aemous: Resolved in latest revision.

kdaily (dismissed):

scripts/new-change
```diff
@@ -57,7 +57,8 @@ TEMPLATE = """\
 # feature: A larger feature or change in behavior, usually resulting in a
 #          minor version bump.
 # bugfix: Fixing a bug in an existing code path.
-# enhancement: Small change to an underlying implementation detail.
+# enhancement: Adding a small feature or a change to an underlying 
+           implementation detail.
```
nateprewitt, line 61:
```suggestion
# enhancement: An incremental improvement or a change to an underlying 
           implementation detail.
```

nateprewitt (commented):
It might be better to avoid the word "feature" here since it denotes a separate change above. Anything that's adding net-new functionality is fine as an enhancement as long as we're not making changes to fundamental systems. Those would prompt a version bump which is when `feature` is used.

hssyoo (approved):

PR #9481, opened 2025-05-07
Remove cloudwatch xml unit tests

arandito:
*Issue #, if available:*

*Description of changes:*
- Removes Cloudwatch xml response parsing unit tests.


ashovlin (commented):
Should we delete the corresponding `.json` files also? Looking at the README, I don't think they're functional once the XML files are removed.

arandito:
> Should we delete the corresponding `.json` files also? Looking at the README, I don't think they're functional once the XML files are removed.

That's correct, will remove them.

ashovlin (approved):

PR #9485, opened 2025-05-09
[v2] Add support for chunking performance tests

aemous:
*Context:*

The main purpose of this PR is adding 'chunking' support to performance tests to enable sharding of the tests across parallel workers in a distributed workflow. The secondary purpose of this PR is to refactor the architecture of the performance testing code to make it easier to integrate more complex performance tests. The architectural decisions were led by implementing more complex RPCv2 CBOR performance test cases and observing points of friction.

*Description of changes:*

* Adds support for evenly chunking performance test cases and only running a specified chunk
* Implemented a `BaseBenchmarkSuite` class which defines the abstract functions that must be implemented to define new performance tests. The default performance test cases were abstracted out into a `JSONStubbedBenchmarkSuite` implementation.
* Modifications to `JSONStubbedBenchmarkSuite`
  * Modified `dimensions` format to require the usage of `name` and `value` keys to reflect the CloudWatch Metrics model.
  * Added support for writing binary content via a new `mode` key on `file_literals` definition.
* Abstracted code for serializing benchmarks results into its own `BenchmarkResultsSerializer` class. This makes it easier to override this class for one-off performance testing when a different output format is expected.
* Architectural changes to performance testing
  * Change `Summarizer` to return metric results in a new `Metric` data model, which includes additional data like description and unit.

*Description of tests:*

* Ran the default JSON stubbed tests via `./scripts/performance/run-benchmarks --num-iterations 20` and observed expected results.


scripts/performance/README.md
```diff
@@ -257,6 +257,9 @@ the specified contents.
       - Each element is an object with the following keys:
         - `name` (string): Name of the file to create
         - `content` (string): The contents of the file.
+        - `mode` (string) **(optional)**: The write mode to use for writing the
+file contents.
+          - Default: `w`
```
ashovlin, line 262: Is anything actually using this? It looks hardcoded below in `_create_file_with_size` and `begin_iteration`.

scripts/performance/run-benchmarks
```diff
@@ -41,4 +51,63 @@ if __name__ == "__main__":
...
+        ),
+    )
+    parsed_args = parser.parse_args()
+    if (
+        sum(
+            x is not None
+            for x in [parsed_args.num_chunks, parsed_args.chunk_id]
+        )
+        == 1
+    ):
```
ashovlin: Is this pattern common? For readability summing bools seems odd.

scripts/performance/tests/__init__.py
```diff
@@ -0,0 +1,24 @@
+class BaseBenchmarkSuite:
```
ashovlin: Do we need a base class if there's only one implementation of it? 

If keeping, could you add class-level comments when when someone would use this?

ashovlin (commented):
When I saw the new `/tests` directory, I thought those would be unit/integration/etc. tests for the performance framework. I'd rather keep in the top level if it's still just implementation.

scripts/performance/README.md
aemous: It is used in line 136 of `simple_stubbed_tests.py`

scripts/performance/tests/__init__.py
aemous: I like having this base class so that it forces us to implement our only test suite via a contract that we know is extendable to other use cases. This base class is meant to be used / extended for more complex performance test suites that are not easily representable in JSON or actually integrate with the service rather than stubbing a client.

I previously extended this base class to implement CBOR benchmarks for a couple other services, and those tests fit nicely against the functions on this base class.
Even though we won't ship those CBOR tests, they served to show locally that the function contract on this class is extendible/reusable.

I will add those class-level comments in the next iteration

scripts/performance/run-benchmarks
aemous: Agreed this is non-idiomatic, will revise.

scripts/performance/README.md
ashovlin: Oops, I just saw the `w` and thought it was hardcoded, but that's just the fallback.

ashovlin (approved):

Issue #9372, opened 2025-03-17
`aws lambda wizard new-function` presents outdated list of runtimes


allanlewis:
### Describe the bug

The command `aws lambda wizard new-function` initially asks for a function name and then a runtime. The latter is presented as a combo box with a several options, many of which are obsolete.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The list of runtimes should be the list of currently-supported runtimes.

### Current Behavior

The list of runtimes appears to be hard-coded here: https://github.com/aws/aws-cli/blob/fed31fc877fac05142ca632d31db135cb9ddf570/awscli/customizations/wizard/wizards/lambda/new-function.yml#L12-L37

### Reproduction Steps

Run `aws lambda wizard new-function` and open the list of runtimes.

### Possible Solution

Update `awscli/customizations/wizard/wizards/lambda/new-function.yml` dynamically such that it's consistent with the current list of runtimes.

### CLI version used

aws-cli/2.24.24 Python/3.12.9 Darwin/24.3.0 source/x86_64

### Environment details (OS name and version, etc.)

MacOS 15.3.1 (24D70)

RyanFitzSimmonsAK:
Hi @allanlewis, thanks for reaching out.  I've cut a PR to update the runtimes in the wizard.  I'll leave this open for tracking until it is merged.  Thanks for bringing this to our attention!

PR #9489, opened 2025-05-12
Update new-function.yml with current runtimes

RyanFitzSimmonsAK:
*Issue #, if available:*

https://github.com/aws/aws-cli/issues/9372

*Description of changes:*

Updated the runtimes in the new-function wizard to match the supported runtimes from https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html


allanlewis (commented):
This looks like it will solve the immediate issue but surely the wizard could get the list of runtimes dynamically?

Also, given the underlying issue, is this wizard even much used? Could it be deprecated after this change?

PR #9491, opened 2025-05-13
Document required top level params

kdaily:
*Issue #, if available:*

*Description of changes:*

Update the documentation of required parameters. This adds documentation in the 'Options' section for each parameter that is required. There is now a `[required]` field in line with the parameter name. This also updates the previously implemented required documentation for nested parameters from #9463.

Also updates string interpolation to use f-strings in `clidocs.py`.

### DIFF
--- a/awscli/clidocs.py
+++ b/awscli/clidocs.py
@@ -180,6 +180,14 @@ class CLIDocumentEventHandler:
     def doc_option(self, arg_name, help_command, **kwargs):
         doc = help_command.doc
         argument = help_command.arg_table[arg_name]
+        if (
+            argument.required
+            or getattr(argument, '_DOCUMENT_AS_REQUIRED', False)
+        ):
+            required_field = ' [required]'
+        else:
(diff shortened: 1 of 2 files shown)


awscli/clidocs.py
```diff
@@ -283,20 +293,23 @@ def _do_doc_member(self, doc, member_name, member_shape, stack, required=False):
         type_name = self._get_argument_type_name(
             member_shape, member_shape.type_name
         )
+
         if member_name:
-            doc.write('%s -> (%s)' % (member_name, type_name))
+            tmp = f'{member_name} -> ({type_name})'
         else:
-            doc.write('(%s)' % type_name)
+            tmp = f'({type_name})'
```
kdaily, line 296: should rename this from `tmp`.

kdaily: Fixed in 1325c4e2f2e7530de019fb91d5c4e2df5adf9b52

PR #9494, opened 2025-05-14
Provide ETag to transfer manager

hssyoo:
Companion PR to https://github.com/boto/s3transfer/pull/345

The S3 transfer manager will call HeadObject during a download if object size and ETag are not provided. To avoid a double HeadObject, this PR updates the high level S3 library to provide the ETag when it makes its own initial HeadObject request.


nateprewitt (changes requested):
After we release the s3transfer update, we need to raise the floor on s3transfer with this change. Otherwise, this is going to be potentially breaking for anyone not updating to the latest version of s3transfer.

nateprewitt (approved):

PR #9496, opened 2025-05-15
[v2] Web-based help 

kdaily:
### Description of changes

Add in a configuration option that can open CLI help pages in the browser or print the URL to the remote CLI reference guide.

The new configuration option is available in the shared configuration file as `cli_help_output`. It has four valid values:

1. Unset: default behavior opening the man page in the terminal.
1. `terminal`: default behavior opening the man page in the terminal.
1. `browser`: open the man page as HTML in the default browser.
1. `url`: print the URL to the remote AWS CLI reference guide for the version of the AWS CLI installed.

When set to `browser`, a warning message is printed to the terminal notifying the user that the default web browser is being opened, and an error message in the case the browser cannot be opened.

When set to `url`, the value for client side paging is respected (e.g., controlled through `AWS_PAGER` environment variable).

###  Implementation notes

This refactors the existing help renderer to move the previous parent class `PagingHelpRenderer` to a sub-class of a `HelpRenderer`. It refactors to potentially send output to different destinations, currently a pager or a browser.

Then, this adds a new renderer base class called the `BrowserHelpRenderer` that swaps out a private method that sends rendered output to the pager with one that sends it to the browser. The rendered content is written to a temporary file and then the users default browser is opened. Separate OS-specific renders inherit, as the paging-based renderers did. An OS-specific sub-class is added that uses this behavior (`PosixBrowserHelpRenderer`) to render the output using `groff` or `mandoc` using `html` instead of `ascii` output, and a `WindowsBrowserHelpRenderer` swaps out the text writer for an HTML writer.

Help command classes (derived from `HelpCommand`) now have a `url` property. Additionally, the `HelpCommand` now uses its own botocore session to determine if its help should be the documentation contents or the URL.


aemous:
> When set to url, the value for client side paging is respected (e.g., controlled through `AWS_PAGER` environment variable).

Is there a case where this would print using the pager? I would think since it's just printing a URL it would just be a single line. Is this for the case that the URL is sufficiently long that the client would benefit from a paging program?

awscli/help.py
```diff
@@ -88,54 +138,62 @@ def get_pager_cmdline(self):
...
+
+    _DEFAULT_DOCUTILS_SETTINGS_OVERRIDES = {
+        # The default for line length limit in docutils is 10,000. However,
+        # currently in the documentation, it inlines all possible enums in
+        # the JSON syntax which exceeds this limit for some EC2 commands
+        # and prevents the manpages from being generated.
+        # This is a temporary fix to allow the manpages for these commands
+        # to be rendered. Long term, we should avoid enumerating over all
+        # enums inline for the JSON syntax snippets.
+        'line_length_limit': 50_000
```
aemous, line 171: Do we have any validation tests that verify no line in our docs exceed this limit? Mainly for future compatibility

awscli/help.py
```diff
@@ -88,54 +138,62 @@ def get_pager_cmdline(self):
...
+    def _send_output_to_destination(self, output):
+        self._send_output_to_browser(output)
+
+    def _send_output_to_browser(self, output):
+        html_file = tempfile.NamedTemporaryFile("wb", suffix=".html", delete=False)
+        html_file.write(output)
+        html_file.close()
 
+        try:
+            print("Opening help file in the default browser.")
```
aemous, line 185: nit: Should we prefer LOG here instead of print to be consistent with the rest of the file?

aemous (changes requested):
Looks good overall. Had some small nits and questions.

Would also like to see some tests added as you already mentioned in the PR description.

Would be nice to see some example output for each config value in the PR description for a single service/operation.

awscli/help.py
kdaily: I think here we want this to be explictly shown to the user on stdout. This would be similar to this message on SSO login:

https://github.com/aws/aws-cli/blob/86520617901307c5078fe74712ea61ebf5f38310/awscli/customizations/sso/utils.py#L199-L207

awscli/help.py
kdaily: Good question - this is actually taken from here though, its pre-existing and I hadn't abstracted it.

https://github.com/aws/aws-cli/blob/86520617901307c5078fe74712ea61ebf5f38310/awscli/help.py#L72-L81

kdaily:
> > When set to url, the value for client side paging is respected (e.g., controlled through `AWS_PAGER` environment variable).
> 
> Is there a case where this would print using the pager? I would think since it's just printing a URL it would just be a single line. Is this for the case that the URL is sufficiently long that the client would benefit from a paging program?

I erred on the side that when we print the result of a command to the console, we always pass it through the pager, to maintain consistency. Other commands that only output less than a full page also go through the pager by default; users can configure their pager to not page if the content is small, like setting `export LESS=-FRX`.

kdaily:
> Looks good overall. Had some small nits and questions.
> 
> Would also like to see some tests added as you already mentioned in the PR description.
> 
> Would be nice to see some example output for each config value in the PR description for a single service/operation.

Yep, next revision will add in tests, now that we got some feedback on the feature. Will update the example output as well.

kdaily:
Added tests in ff4dd079ca2fa876c31376b65cc817c09d0591cb. These are still a bit rough, want to get full coverage check on the different operating systems.

awscli/__init__.py
```diff
@@ -104,3 +104,5 @@ def find_spec(self, fullname, path, target=None):
 
 
 TopLevelImportAliasFinder.add_alias_finder(sys.meta_path)
+
+_DEFAULT_BASE_REMOTE_URL = f"https://awscli.amazonaws.com/v2/documentation/api/{__version__}"   # noqa
```
kdaily, line 108: Should get "help" or "documentation" in the variable name.

awscli/help.py
aemous: Alright, since it was not newly introduced in this PR I won't expect/require adding a new test for it here.

aemous:
> I added the ability to set the help output format via an environment variable. This provides users a way to set it per invocation without adding a global flag, which is tricky to add since it should only apply to `aws * help` commands.

1. Have we considered just adding an argument to the `aws help` command rather than global? 
2. (non-blocking) I find it a bit weird that we are adding an env/config var that only modifies the behavior of a single command. But this isn't the first env var to do so, `AWS_CLI_S3_MV_VALIDATE_SAME_S3_PATHS` is an env var that only modifies `aws s3 mv`, so there is precedent. This is just a note/consideration, not a requested change.

awscli/customizations/commands.py
```diff
@@ -383,6 +384,12 @@ def examples(self):
     def event_class(self):
         return '.'.join(self.obj.lineage_names)
 
+    @property
+    def url(self):
+        if len(self.command_table) > 0:
```
aemous, line 388: non-blocking: Can you explain what this if statement is doing here? A one-liner comment explaining it would be nice too

awscli/customizations/commands.py
```diff
@@ -408,7 +415,12 @@ def __call__(self, args, parsed_globals):
         # We pass ourselves along so that we can, in turn, get passed
         # to all event handlers.
         docevents.generate_events(self.session, self)
-        self.renderer.render(self.doc.getvalue())
+
+        if self._help_output_format == 'url':
+            self.renderer.render(self.url.encode())
+        else:
+            self.renderer.render(self.doc.getvalue())
+
```
aemous, line 422: 1. For defensive programming purposes, can we switch to use if-elif with the accepted known values? 

2. Do we want some kind of validation on the value of the config var? Whether yes/no should be consistent with other similar enum vars that we use. If we don't tend to use validation I'd be fine with this, at a minimum let's be aware of what output we get when users supplies an invalid config value here.

awscli/help.py
```diff
@@ -68,6 +91,34 @@ class PagingHelpRenderer:
...
+
+    def _popen(self, *args, **kwargs):
+        return Popen(*args, **kwargs)
+
+    def _convert_doc_content(self, contents):
+        return contents
+
+
+class PagingHelpRenderer(HelpRenderer):
+    """Interface for a help renderer.
```
aemous, line 116: nit: should there be newline after """ for consistency of formatting

awscli/help.py
```diff
@@ -88,59 +139,69 @@ def get_pager_cmdline(self):
...
+
+    def _send_output_to_destination(self, output):
+        self._send_output_to_browser(output)
+
+    def _send_output_to_browser(self, output):
+        html_file = tempfile.NamedTemporaryFile(
+            "wb", suffix=".html", delete=False
+        )
+        html_file.write(output)
+        html_file.close()
```
aemous, line 182: non-blocking: This appears to do 1 temp write each time a help command is executed. a future optimization may consider maintaining a cache of these / the whole docs, but this is unnecessary at this point and can always be changed later if we notice performance issues

awscli/help.py
aemous: Understood, should we use uni_print then for consistency?

awscli/customizations/commands.py
```diff
@@ -383,6 +384,12 @@ def examples(self):
     def event_class(self):
         return '.'.join(self.obj.lineage_names)
 
+    @property
+    def url(self):
```
aemous, line 383: Do we plan to port this PR to v1? There may be some v1 edge cases where this won't work.

For example, with `aws sms-voice list-configuration-sets` in AWS CLI v1 the event_class is `sms-voice.list-configuration-sets`, but the docs page https://docs.aws.amazon.com/cli/v1/sms-voice/list-configuration-sets does not exist, because we mark the sms-voice command as `UNDOCUMENTED`.

aemous (changes requested):
Looks like some tests are still failing, may be flaky based on the OS/runtime versions.

Requested some (mostly minor) changes and left some questions

awscli/customizations/commands.py
kdaily: No, we are not going to port this to v1. However, I did confirm that the URL in v2 is handled properly for an undocumented operation, like this one:

https://github.com/aws/aws-cli/blob/787bbf556b6f0ce0d5aac1e4db86c15d3e0535ae/awscli/customizations/codecommit.py#L58-L66

But there is indeed a docs page generated for it.

https://awscli.amazonaws.com/v2/documentation/api/2.27.37/reference/codecommit/credential-helper/erase.html

awscli/help.py
kdaily: Noted as an improvement to determine a caching strategy.

awscli/customizations/commands.py
kdaily: I looked at the parameters that control the autoprompt mode. It follows a similar pattern that falls back to a default if an unknown value is used. I tested this by setting an unknown value in the config file like this:

```
[default]
cli_auto_prompt = absolutely
```

The only valid values for `cli_auto_prompt` are `off`, `on`, and `on-partial`. This code falls back to `off` if any other value is given.

https://github.com/aws/aws-cli/blob/a8d8f06152c3c4b8a5726e73c01fb30a9e3b2975/awscli/clidriver.py#L224-L236

kdaily: But you are correct, it should still be an `elif` there!

awscli/customizations/commands.py
kdaily: Customizations, including custom commands and things like wizards, don't have a separate help command interface, they inherit from `BasicCommand`. This logic is taken from the docs generator to differentiate between a command (which indexes sub-commands) and sub-commands that are the leaves of documentation pages.

awscli/help.py
kdaily: I think its fine to leave it. Even this module is inconsistent (see `ProviderHelpCommand` or `OperationHelpCommand`, for example). The PEP for docstrings suggests first line is in line with the opening triple quote.

kdaily:
> > I added the ability to set the help output format via an environment variable. This provides users a way to set it per invocation without adding a global flag, which is tricky to add since it should only apply to `aws * help` commands.

For the time being until an interface is agreed upon, I removed the environment variable option. 

>     1. Have we considered just adding an argument to the `aws help` command rather than global?

To add to `help` only, it would need a different interface. It doesn't look like `arg_table` is used for help commands at the moment. An offline suggestion to add a new help subcommand (like `aws help browser`), which would be alongside `aws help topics`, is also possible. However, this would need to be available on every sub-command too (so `aws sts get-caller-identity help browser`). It would be possible to add a DIFFERENT sub-command, like `aws help-browser`), but this is adding yet another help system, and it was deemed undesirable.
 
>     2. (non-blocking) I find it a bit weird that we are adding an env/config var that only modifies the behavior of a single command. But this isn't the first env var to do so, `AWS_CLI_S3_MV_VALIDATE_SAME_S3_PATHS` is an env var that only modifies `aws s3 mv`, so there is precedent. This is just a note/consideration, not a requested change.

I'm not sure that I understand this concern? Setting the config variable (or the since removed env variable) would change the behavior for every help command provided in the AWS CLI; though maybe what you mean is it ONLY applies when adding the sub-command `help`, and yes this is indeed novel.

awscli/help.py
kdaily: Looks like `uni_print` is specifically for customizations. We use the std lib `print` in autocomplete as well:

https://github.com/aws/aws-cli/blob/9c8f131b89c2f13c87132446c93c21b467d9dafe/awscli/autocomplete/main.py#L62

awscli/customizations/commands.py
aemous: Alright thanks for confirming that.

aemous (approved):
LGTM.

aemous (approved):
LGTM

kdaily:
Merging with a failed but unrelated prompttoolkit test.

PR #9498, opened 2025-05-18
[v2] Add session id to user agent string

hssyoo:
This PR generates and appends a session id to the user agent string.

### Description
A session id is an identifier that represents an AWS CLI usage session (ie a specific task or logical grouping of tasks). It has the following properties:
- Hashed using a randomly generated host id, TTY path (if available), and timestamp
- A session id is reused from the cache if it has been less than 30 minutes since the id was last used

Session ids are cached in a SQLite database stored in the `session` table. An entry in the database consists of:
- `key`: Primary key that represents a cache key, which is a hash of the host id + TTY path
- `session_id`: The session id associated with a host id + TTY path
- `timestamp`: Most recent timestamp of the session id's usage

A single host id is generated using uuid4 and persisted in the `host_id` table. The primary key is always hardcoded to `0` to ensure there's only ever 1 host id that doesn't change.

A cached session id is considered to be expired if it has been at least 30 minutes since the associated timestamp. On every AWS CLI invocation, the cache is cleared by deleting all entries with timestamps < current timestamp - (60*30) seconds. The cleanup job is processed in a daemon thread that terminates as soon as the main CLI process terminates to prevent blocking on cleanup tasks.

When determining the session id, it first checks the cache.
- If cache hit, check if the timestamp is expired
  - If not expired, update only the timestamp and write it back to table
  - If expired, generate and update session id and timestamp and write it back to table
- If cache miss, generate a new session id and write it to table

The session id is appended to the user agent string under the `sid` prefix. eg:
```
sid/090bcefb7f7cd3a77d9db442e378c0e1
```

### Caveats
- Separating sessions between 30 minutes of inactivity is arbitrary and requires client code updates if we decide to adjust it. Future improvements could involve a smarter way to determine unique sessions, outside of relying on timestamps.
- TTY path is used to generate a session id because we want different screens to represent different sessions. There's not a reliable way to get something similar on Windows machines, so it'll only ever have 1 active session id at a time.

### Why SQLite over files?
Caching session data in files can lead to race conditions when multiple processes are writing/reading to/from the same files. We can work around this by either implementing cross-file locks or writing to temporary files and moving them, but both approaches add complexity to implementation.

SQLite offers atomic primitives off the shelf, along with a few useful features:
- Write-ahead logging for improved concurrency
- `INSERT OR REPLACEINTO` for simpler writes
- `DELETE .. WHERE` for bulk deletes


awscli/telemetry.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class CLISessionGenerator:
+    def generate_session_id(self, host_id, tty, timestamp):
+        return self._generate_md5_hash(host_id, tty, timestamp)
+
+    def generate_cache_key(self, host_id, tty):
+        return self._generate_md5_hash(host_id, tty)
+
+    def _generate_md5_hash(self, *args):
```
ashovlin: In light of https://github.com/aws/aws-cli/pull/9535 recently, should we use a different hash in case MD5 isn't available? Or at least use `botocore.compat.get_md5`?

awscli/telemetry.py
```diff
@@ -0,0 +1,279 @@
...
+        CLISessionDatabaseWriter(conn),
+        CLISessionDatabaseReader(conn),
+        CLISessionDatabaseSweeper(conn),
+    )
+
+
+def add_session_id_component_to_user_agent_extra(session, orchestrator=None):
+    cli_session_orchestrator = orchestrator or _get_cli_session_orchestrator()
+    add_component_to_user_agent_extra(
+        session, UserAgentComponent("sid", cli_session_orchestrator.session_id)
```
ashovlin: Reading the internal docs again, should this be `md/sid#<value>` instead? Not sure if individual tools should declare new "top-level" pairs like this.

ashovlin (dismissed):

awscli/telemetry.py
hssyoo: This was discussed with Balaji. We wanted to keep it short

awscli/telemetry.py
hssyoo: Addressed via https://github.com/aws/aws-cli/pull/9498/commits/e8f4732355169cb04de034abbba21f4975c2cc17

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+
+from awscli.compat import is_windows
+from awscli.utils import add_component_to_user_agent_extra
+
+_CACHE_DIR = Path.home() / '.aws' / 'cli' / 'cache'
+_DATABASE_FILENAME = 'session.db'
+_SESSION_LENGTH_SECONDS = 60 * 30
+_SESSION_ID_LENGTH = 12
+
+_CACHE_DIR.mkdir(parents=True, exist_ok=True)
```
aemous: Since this `mkdir` call is "guardless", it'll run each time it is imported in another class. Is there any better place to put it to ensure it's only run once per CLI command execution?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def _get_checksum():
+    hashlib_params = {"usedforsecurity": False}
+    try:
+        checksum = get_md5(**hashlib_params)
+    except MD5UnavailableError:
+        import hashlib
+
+        checksum = hashlib.sha256(**hashlib_params)
```
aemous, line 44: What if MD5 is initially unavailable, so we use SHA256 to compute checksums, then in a future invocation MD5 is available (maybe this is impossible within the same TTY?), so we'll use a different checksum, is this an edge case that would cause unexpected behavior? Would it ever cause overwriting the cache before the 30 minute expiration?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+            check_same_thread=False,
+            isolation_level=None,
+        )
+        self._ensure_database_setup()
+
+    def execute(self, query, *parameters):
+        try:
+            return self._connection.execute(query, *parameters)
+        except sqlite3.OperationalError:
+            # Process timed out waiting for database lock.
```
aemous, line 92: non-blocking: What are the implications of returning an empty cursor here?  Would be nice for devs less familiar with sqlite to have that commented. I'm guessing writes to an empty cursor will result in no actual write? Is it worth logging this to debug logger?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+        INSERT OR REPLACE INTO session (
+            key, session_id, timestamp
+        ) VALUES (?, ?, ?)
+    """
+
+    def __init__(self, connection):
+        self._connection = connection
+
+    def write(self, data):
+        self._connection.execute(
```
aemous, line 146: non-blocking: connection.execute creates a new cursor and calls execute on the cursor, and returns the cursor. Will we ever call write multiple times within the same CLI command? if so, consider storing the returned cursor, and reusing it in subsequent calls (if it's not empty) as a slight performance optimization

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def read(self, key):
+        cursor = self._connection.execute(self._READ_RECORD, (key,))
+        result = cursor.fetchone()
+        if result is None:
+            return
+        return CLISessionData(*result)
+
+    def read_host_id(self):
+        cursor = self._connection.execute(self._READ_HOST_ID)
```
aemous, line 179: Non-blocking: similar comment here, is it worth storing the cursor for subsequent calls?  (also applies to all places that connection.execute is called)

separately, what if the cursor returned is empty due to db being locked? Would this throw an error?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+    def cache_key(self):
+        return self._generator.generate_cache_key(self._host_id, self._tty)
+
+    @cached_property
+    def _session_id(self):
+        return self._generator.generate_session_id(
+            self._host_id, self._tty, self._timestamp
+        )
+
+    @cached_property
```
aemous, line 237: Are there edge cases where session id would update in the middle of a single command execution? Maybe a command that takes over 30 mins to run? Since cached_property is used, this will always return the same value even if session id should be updated. Is this a concern?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+            # Standard input was redirected to a pseudofile.
+            # This can happen when running tests on IDEs or
+            # running scripts with redirected input, etc.
+            return
+
+    @cached_property
+    def _host_id(self):
+        return self._reader.read_host_id()
+
+    @cached_property
```
aemous, line 275: Using cached property will return the same timestamp each time, is this intended? Would we ever request a session id multiple times in one command execution?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+            orchestrator or _get_cli_session_orchestrator()
+        )
+        add_component_to_user_agent_extra(
+            session,
+            UserAgentComponent("sid", cli_session_orchestrator.session_id),
+        )
+    except Exception:
+        # Ideally, the AWS CLI should never throw if the session id
+        # can't be generated since it's not critical for users. Issues
+        # with session dada should instead be caught server-side.
```
aemous: Nit: is dada a typo here?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+        CLISessionDatabaseReader(conn),
+        CLISessionDatabaseSweeper(conn),
+    )
+
+
+def add_session_id_component_to_user_agent_extra(session, orchestrator=None):
+    try:
+        cli_session_orchestrator = (
+            orchestrator or _get_cli_session_orchestrator()
+        )
```
aemous, line 302: Ah, so currently my comments on cached_property and reusing same value may not be relevant since we're adding the user agent to the botocore session once, and reusing that for each operation invoked in the command? Do we want to support theoretically updating the user agent between operations within the same command (e.g. for long-running customization commands, like s3 sync). In theory a sync can take an hour, and so a user can have 2 different session ids within that one command?

tests/functional/test_telemetry.py
```diff
@@ -0,0 +1,324 @@
...
+    session = MagicMock(Session)
+    session.user_agent_extra = ''
+    orchestrator = MagicMock(CLISessionOrchestrator)
+    orchestrator.session_id = 'my-session-id'
+    add_session_id_component_to_user_agent_extra(session, orchestrator)
+    assert session.user_agent_extra == 'sid/my-session-id'
+
+
+def test_entrypoint_catches_bare_exceptions():
+    class FakeOrchestrator(CLISessionOrchestrator):
```
aemous: non-blocking nit: is it worth using a Mock object here that throws Exception when session_id is called rather than defining a class?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+        WHERE timestamp < ?
+    """
+
+    def __init__(self, connection):
+        self._connection = connection
+
+    def sweep(self, timestamp):
+        try:
+            self._connection.execute(self._DELETE_RECORDS, (timestamp,))
+        except Exception:
```
aemous, line 195: Can we catch the specific exceptions that SQLite guarantees can be thrown instead? If some other serious exception is occurring, maybe we want to know about it? Maybe at least worth a debug log entry?

awscli/telemetry.py
```diff
@@ -0,0 +1,303 @@
...
+def add_session_id_component_to_user_agent_extra(session, orchestrator=None):
+    try:
+        cli_session_orchestrator = (
+            orchestrator or _get_cli_session_orchestrator()
+        )
+        add_component_to_user_agent_extra(
+            session,
+            UserAgentComponent("sid", cli_session_orchestrator.session_id),
+        )
+    except Exception:
```
aemous, line 307: Can we scope this down to the specific exceptions guaranteed to be thrown by the functions called? Is a debug log entry worth it despite session id not being critical?

awscli/telemetry.py
hssyoo: Not too worried about this case. This fallback in primarily for handling FIPS-compliant environments where MD5 would never be allowed

awscli/telemetry.py
hssyoo: For the purpose of a user session, the session id should be the same for all requests from a single command

awscli/telemetry.py
hssyoo: See comment above

awscli/telemetry.py
hssyoo: That's a good consideration, but ultimately we want the same session id for single command. I also think there are performance considerations with re-calculating the session id between each request in a single command that aren't justified for capturing session ids

awscli/telemetry.py
hssyoo: This is a safeguard to ensure we never throw on session ids. I don't think we should add unnecessary risk by catching specific exceptions. I'm avoiding debug logs in this module because this is purely a server-side "feature" that should be completely transparent to users. ie There's nothing actionable for users if they see funny debug logs on their end

awscli/telemetry.py
hssyoo: Same as comment above

awscli/telemetry.py
hssyoo: Session ids are generated just once per command invocation

awscli/telemetry.py
hssyoo: Ah so the code's never writing to an empty cursor. All executions happen directly on a database connection, which returns a cursor object. For writes, the cursor object is effectively throw away after returning. For reads, the cursor object holds a pointer to the database records that can then be fetched. The empty cursor is being returned as a fallback so that when a downstream caller tries to read the returned cursor, we won't get an exception like "None object has no method blah"

awscli/telemetry.py
hssyoo: Hm good callout on potentially empty cursors. I don't think returning `None` or some default value makes sense here because that would generate a different session id.

Thinking about it more, this is a read query for a single record, so it should never acquire a blocking lock since Sqlite supports shared reads. The only time it may be blocked is when another process is writing/updating the record. I'm leaning towards first checking to see if the host_id table is empty before attempting to write a host id to it. This would ensure that the initial insert/ignore query only happens once, then all subsequent queries for the host_id table would be read-only, ensuring that a value is always returned. Thoughts?

awscli/telemetry.py
aemous: Ah, thanks for the clarification

tests/functional/test_telemetry.py
hssyoo: Addressed via https://github.com/aws/aws-cli/pull/9498/commits/6b8b905d2dd309924865d37f2d43fa4c36d30ab2

awscli/telemetry.py
hssyoo: Addressed via https://github.com/aws/aws-cli/pull/9498/commits/facdd6434eba407074bdb272bb4b1a2bd53f49e8

awscli/telemetry.py
hssyoo: Addressed via https://github.com/aws/aws-cli/pull/9498/commits/3cc0974e7f5f390fc81e2ecc4e1abb41af091d76

awscli/telemetry.py
hssyoo: Addressed via https://github.com/aws/aws-cli/pull/9498/commits/6daeaefa76ab8bfe9a6212b1e4263743079149ce

aemous: Nice, LGTM

PR #9512, opened 2025-05-24
Merging duplicate code

0xAJX:
There are multiple if statements with duplicate code, so merging them for easier readability.


ashovlin:
Thanks for opening, but we prefer to leave it as is since we rarely touch this area and in case their behavior does diverge. See also https://github.com/aws/aws-cli/issues/9523#issuecomment-2956814492.

PR #9534, opened 2025-06-06
Add SHA256 fallback if MD5 isn't available on the system

nateprewitt:
*Description of changes:*
This PR will move our usage of md5 to enable the usedforsecurity flag set to false since these checksumming functions are only intended to create a deterministic path. If MD5 is still not available on the system, we'll attempt to use a FIPS compliant algorithm (SHA256 in this case) to derive an alternative path.

Path names should stay deterministic on similarly configured systems but customers uploading from both FIPS compliant and non-compliant environments to the same bucket may see some redundant uploads for the same template. Risk of this should be low given compliance requirements. 

### DIFF
--- a/awscli/customizations/s3uploader.py
+++ b/awscli/customizations/s3uploader.py
@@ -11,7 +11,6 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
-import hashlib
 import logging
 import threading
 import os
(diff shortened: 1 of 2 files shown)


kdaily:
Implementation looks good to me.

We should update the command description here to account for this change:

https://github.com/aws/aws-cli/blob/60ae3cbd865d8cdaa5c9a425b3b743027e621a3a/awscli/examples/cloudformation/_package_description.rst

> Before the command uploads artifacts, it checks if the artifacts are already present in the S3 bucket to prevent unnecessary uploads. The command uses MD5 checksums to compare files. If the values match, the command doesn't upload the artifacts. Use the --force-upload flag to skip this check and always upload the artifacts.

Suggestion:

Before the command uploads artifacts, it checks if the artifacts are already present in the S3 bucket to prevent unnecessary uploads. The command uses a hashing function to compare files. If the values match, the command doesn't upload the artifacts. Use the --force-upload flag to skip this check and always upload the artifacts. By default, an MD5 checksum is used as a hash. If MD5 is not available in the environment, a SHA256 checksum is used.

kdaily (approved):

PR #9545, opened 2025-06-12
[v2] S3 progress indicator frequency

kdaily:
*Issue #, if available:*

*Description of changes:*

Add the ability to control the frequency of updating the S3 progress indicator bar as well as print the status messages to multiple lines.

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -188,7 +188,7 @@ class ResultRecorder(BaseResultHandler):
         if not isinstance(result, BaseResult):
             raise ValueError(
                 'Any result using _get_ongoing_dict_key must subclass from '
-                'BaseResult. Provided result is of type: %s' % type(result)
+                f'BaseResult. Provided result is of type: {type(result)}'
             )
         key_parts = []
         for result_property in [result.transfer_type, result.src, result.dest]:
@@ -316,7 +316,12 @@ class ResultPrinter(BaseResultHandler):
     SRC_TRANSFER_LOCATION_FORMAT = '{src}'
 
     def __init__(
-        self, result_recorder, out_file=None, error_file=None, frequency=0, oneline=True
+        self,
+        result_recorder,
+        out_file=None,
+        error_file=None,
+        frequency=0,
+        oneline=True,
     ):
         """Prints status of ongoing transfer
 
@@ -356,26 +361,9 @@ class ResultPrinter(BaseResultHandler):
 
     def __call__(self, result):
         """Print the progress of the ongoing transfer based on a result"""
-        result_handler = self._result_handler_map.get(type(result), self._print_noop)
-        if type(result) is ProgressResult:                
-            result_handler = self._override_progress_result_handler(
-                result, result_handler
-            )
-        result_handler(result=result)
-
-    def _override_progress_result_handler(self, result, result_handler):
-        if (
-            type(result) in [ProgressResult]
-            and (
-                self._first
-                or (self._frequency == 0)
-                or (time.time() - self._now >= self._frequency)
-            )
-        ):
-            self._now = time.time()
-            self._first = False
-            return result_handler
-        return self._print_noop
+        self._result_handler_map.get(type(result), self._print_noop)(
+            result=result
+        )
 
     def _print_noop(self, **kwargs):
         # If the result does not have a handler, then do nothing with it.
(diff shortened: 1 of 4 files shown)


awscli/customizations/s3/results.py
```diff
@@ -331,6 +333,8 @@ def __init__(self, result_recorder, out_file=None, error_file=None):
         """
```
aemous, line 349: Can we add docs for `frequency` and `oneline`?

awscli/customizations/s3/results.py
```diff
@@ -347,12 +351,31 @@ def __init__(self, result_recorder, out_file=None, error_file=None):
...
+        result_handler = self._result_handler_map.get(type(result), self._print_noop)
+        if type(result) is ProgressResult:                
+            result_handler = self._override_progress_result_handler(
+                result, result_handler
+            )
+        result_handler(result=result)
+
+    def _override_progress_result_handler(self, result, result_handler):
+        if (
+            type(result) in [ProgressResult]
```
aemous: nit: could this be `type(result) is ProgressResult` to be consistent with the same check in `__call__` ?

awscli/customizations/s3/results.py
```diff
@@ -347,12 +351,31 @@ def __init__(self, result_recorder, out_file=None, error_file=None):
             DryRunResult: self._print_dry_run,
             FinalTotalSubmissionsResult: self._clear_progress_if_no_more_expected_transfers,
         }
+        self._now = time.time()
```
aemous, line 370: non-blocking: seems this initial value is never used, consider initializing to `None`

awscli/customizations/s3/results.py
```diff
@@ -347,12 +351,31 @@ def __init__(self, result_recorder, out_file=None, error_file=None):
...
-            result=result
-        )
+        result_handler = self._result_handler_map.get(type(result), self._print_noop)
+        if type(result) is ProgressResult:                
+            result_handler = self._override_progress_result_handler(
+                result, result_handler
+            )
+        result_handler(result=result)
+
+    def _override_progress_result_handler(self, result, result_handler):
```
aemous: non-blocking question: have we considered modifying `_print_progress` to check `frequency` and `first` with a similar check, and only print as needed, instead of adding a new override handler function?

aemous (changes requested):
LGTM overall. Any thoughts on adding tests for the multiline progress to verify it does print to multiple lines when the arg is supplied?

awscli/customizations/s3/results.py
kdaily: Worth exploring! I'll look into what that would be like and if there are any pros or cons.

awscli/customizations/s3/results.py
kdaily: Wouldn't it be an issue here?

https://github.com/aws/aws-cli/pull/9545/files#diff-dd9d793860521b9a92ab0ba7eed780d48b8b4dc0cc920e73ec39f4c945cf2878R372

awscli/customizations/s3/results.py
kdaily: This seems to work too.

kdaily:
Also added tests for multiline and frequency configured progress status updates.

aemous (changes requested):
Looks good overall, can we fix the failing test and address the last unresolved comment?

tests/unit/customizations/s3/test_results.py
```diff
@@ -1147,7 +1342,7 @@ def test_failure_but_no_expected_files_transferred_provided(self):
         ref_statement = (
             'upload failed: file to s3://mybucket/mykey my exception\n'
             'Completed 1.0 MiB/~1.0 MiB (0 Bytes/s) with ~0 file(s) '
-            'remaining (calculating...)\r'
```
kdaily, line 1150: Inadvertent change, need to remove this one.

kdaily: Fixed with f89532d993ca4024456d0ac4072679914c322a29

awscli/customizations/s3/results.py
kdaily: Fixed in b8897a66ceeb5227c3e7989be9eb46a1a1414f73

aemous (approved):
LGTM

PR #9546, opened 2025-06-12
v2 CLI only example for API Gateway

daniil-millwood:


elysahall (approved):
LGTM

ashovlin (approved):
Might be nice to explain that this is base64 encoded in the preamble, but if the current example is not working then okay with this as an improvement.

daniil-millwood:
Okay, @ashovlin Let's leave it as-is.

daniil-millwood:
Thank you.

PR #9552, opened 2025-06-17
[v2] Upgrade prompt-toolkit to 3.0.51

hssyoo:
*Issue #, if available:*

*Description of changes:*


hssyoo:
Okay so prompt-toolkit began identifying its version using `importlib.metadata.version` [[source]](https://github.com/prompt-toolkit/python-prompt-toolkit/commit/68e3c3981f785f810578cd099e48de38fb2ad359). There are 2 issues here..
1. The AWS CLI looks for certain path finders and wraps them inside a custom path finder [[source]](https://github.com/aws/aws-cli/blob/911c50f2dd349ec865337665ec5ae03bbc906697/awscli/__init__.py#L97). This prevents prompt-toolkit's metadata from being found at build time and runtime.
2.  PyInstaller doesn't copy package metadata by default when building the frozen executable. This prevents prompt-toolkit's metadata from being found at runtime. There's a way to configure this, but I haven't gotten it to work yet [[docs](https://pyinstaller.org/en/stable/hooks.html#PyInstaller.utils.hooks.copy_metadata)].

PR #8345, opened 2023-11-21
Fix sse-c HeadObject headers for s3-s3 copy

jstastny:
Also fixing https://github.com/aws/aws-cli/issues/6012, which describes the unencrypted to encrypted copy, while we were running into the same thing with encrypted to encrypted (different key) scenario.

*Description of changes:*
This fixes the S3 to S3 copy when using SSE-C keys for multipart object.

Without this, the head object, which is part of the copy flow, fails because it uses target sse-c keys when accessing the objects from the source.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #9559, opened 2025-06-18
Fix handling of SSE-C keys when copying unencrypted to encrypted objects or objects with different keys

ashovlin:
*Issue #, if available:* #6012, and supersedes #8345

*Description of changes:* When copying a file that originates in S3, the high level `aws s3` commands make a `HeadObject` call for the file that is being copied. 

When using [server-side encryption with customer-provided keys (SSE-C)](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html), callers can specify the key for the source object(s) `--sse-c-copy-source`/`--sse-c-copy-source-key` and/or the destination object (`--sse-c`/`--sse-c-key`). 

The CLI is incorrectly handling this for S3 -> S3 copies, by applying destination key to the copy source object.
1. Copying an unencrypted object to an encrypted one: `aws s3 cp s3://$bucket/file.txt s3://$bucket/ssec-copy-c.txt --sse-c AES256 --sse-c-key $key_1`
2. Copying an object with different encryption keys: `aws s3 cp s3://$bucket/ssec.txt s3://$bucket/ssec-copy-diff-key.txt --sse-c-copy-source AES256 --sse-c-copy-source-key $key_1 --sse-c AES256 --sse-c-key $key_2`

*Arguably Incorrect Behavior, Leaving for Backwards Compatibility*

I think we're arguably handling S3 -> local copies as well. `aws s3 cp s3://$bucket/ssec.txt decrypt.txt --sse-c AES256 --sse-c-key $key_1` works, but I can see a case for applying `--sse-c-copy-source`/`--sse-c-copy-source-key` to the object that's originating to S3. But I'm preserving this behavior for backwards compatibility, the updated logic only runs for S3 -> S3 copies.

### DIFF
--- a/.changes/next-release/bugfix-s3-83199.json
+++ b/.changes/next-release/bugfix-s3-83199.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``s3``",
+  "description": "Fix handling of SSE-C keys when copying unencrypted to encrypted objects or objects with different encryption keys"
+}
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1481,16 +1481,18 @@ class CommandArchitecture:
         # SSE-C key and algorithm. Note the reverse FileGenerator does
         # not need any of these because it is used only for sync operations
         # which only use ListObjects which does not require HeadObject.
-        RequestParamsMapper.map_head_object_params(
-            request_parameters['HeadObject'], self.parameters
-        )
         if paths_type == 's3s3':
+            head_params = self.parameters.copy()
+            head_params['sse_c'] = self.parameters.get('sse_c_copy_source')
+            head_params['sse_c_key'] = self.parameters.get(
+                'sse_c_copy_source_key'
+            )
             RequestParamsMapper.map_head_object_params(
-                request_parameters['HeadObject'],
-                {
-                    'sse_c': self.parameters.get('sse_c_copy_source'),
-                    'sse_c_key': self.parameters.get('sse_c_copy_source_key'),
-                },
+                request_parameters['HeadObject'], head_params
+            )
+        else:
+            RequestParamsMapper.map_head_object_params(
+                request_parameters['HeadObject'], self.parameters
             )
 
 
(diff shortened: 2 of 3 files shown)


FilipPyrek:
Awesome! Thanks @ashovlin 👏

FilipPyrek:
Hi @ashovlin when do you think we could get this merged? 🙂

ashovlin:
> Looks good. But for discussion:
> 
> > `aws s3 cp s3://$bucket/ssec.txt decrypt.txt --sse-c AES256 --sse-c-key $key_1` works, but I can see a case for applying `--sse-c-copy-source`/`--sse-c-copy-source-key` to the object that's originating to S3.
> 
> `GetObject` itself uses `SSECustomerAlgorithm` over `CopySourceSSECustomerAlgorithm`, so isn't the current S3->local behavior correct?

Ah, maybe it's more open to interpretation then:
1. My mental model was that the `--sse-c-copy-source` pair is for the source object, and `--sse-c` for the destination object regardless of local vs. S3, so the current behavior slightly surprised me.
2. But if you understand that the`--see-c` pair as maps to`SSECustomerAlgorithm`, then this makes sense.

m1so:
Hi @ashovlin & @hssyoo, following up on the PR after taking over from @FilipPyrek. Happy to provide any inputs to move this forward 👍 

Our main usecase at Deepnote was summarized by Filip in this thread: https://github.com/aws/aws-cli/pull/8345#issuecomment-2994157728

ashovlin:
Reset approvals and pushed a new revision in https://github.com/aws/aws-cli/pull/9559/changes/ebe7c0670519ee746033cb08aedec0e42f5353cf with increased test coverage.

The original commit, only updating `_map_sse_c_params` in `subcommands.py` is called for all invocations.

However if doing either a `cp -r` or `sync`, we also issue an `HeadObject` call for objects over the multipart threshold from `SetMetadataDirectivePropsSubscriber`, since we need to copy metadata properties manually instead of using the `x-amz-metadata-directive` on `CopyObject` as we do for objects under the multipart threshold. 

The new commit address both places, using the approach from https://github.com/aws/aws-cli/pull/8345 which this is superseding.

ashovlin:
Internal build passed via https://github.com/aws/aws-cli/pull/10162, but planning to merge this one since we reviewed here.

PR #9579, opened 2025-06-28
Update aws_zsh_completer.sh

rrotter:
*Issue #, if available:*
n/a

*Description of changes:*
- remove `-i` flag from `bashcompinit`: this command takes no flags, the flag has no effect (see: https://github.com/zsh-users/zsh/blob/master/Completion/bashcompinit, it contains no argument parsing at all)
- remove patching of _bash_complete: this was patched in zsh 5.0.3 which was released in 2013; no OS or package manager has shipped a zsh this old in 10+ years
- only run bashcompinit if it hasn't been run already: there are plenty of other bash only completers; re-running bashcompinit for each one is unnecessary and slows down shell initialization

### DIFF
--- a/bin/aws_zsh_completer.sh
+++ b/bin/aws_zsh_completer.sh
@@ -5,56 +5,10 @@
 # % source /path/to/zsh_complete.sh
 #
 # Typically that would be called somewhere in your .zshrc.
-#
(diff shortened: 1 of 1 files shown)


rrotter:
#10327 mostly addresses this.

PR #9590, opened 2025-07-10
added support for AWS temp creds being set with AWS configure, added aws MFA-login

liwadman:
*Issue #, if available:*
N/A, find me on slack

*Description of changes:*
Per our discussion

*Added aws configure mfa-login sub command, which takes the user interactively through starting an MFA session for an IAM user
*Added support for temporary AWS security credentials with aws configure

### DIFF
--- a/.changes/next-release/configure-tempcreds.json
+++ b/.changes/next-release/configure-tempcreds.json
@@ -1,12 +1,11 @@
-[]
 {
   "type": "enhancement",
-  "category": "``Configure``",
-  "description": "Added Support for AWS temp creds with AWS configure, e.g. CLI will ask for aws_session_token if the access key id is temporary"
+  "category": "``configure``",
+  "description": "Added support for temporary credentials with ``aws configure``. The CLI will prompt for an ``aws_session_token`` if the provided access key ID is temporary."
 },
 {
   "type": "enhancement",
-  "category": "``Configure``",
-  "description": "Added new sub-command to aws confugre, `mfa-login`, which is an interactive wrapper for the aws sts get-session-token command"
+  "category": "``configure``",
+  "description": "Added the new ``aws configure mfa-login` command, which creates a profile with temporary credentials corresponding to an IAM user with an MFA code."
 },
 ]
--- a/.changes/next-release/configure-tempcreds.json
+++ b/.changes/next-release/configure-tempcreds.json
@@ -1,6 +1,12 @@
-
+[]
 {
   "type": "enhancement",
   "category": "``Configure``",
   "description": "Added Support for AWS temp creds with AWS configure, e.g. CLI will ask for aws_session_token if the access key id is temporary"
-}
+},
+{
+  "type": "enhancement",
+  "category": "``Configure``",
+  "description": "Added new sub-command to aws confugre, `mfa-login`, which is an interactive wrapper for the aws sts get-session-token command"
+},
+]
--- a/.changes/next-release/configure-tempcreds.json
+++ b/.changes/next-release/configure-tempcreds.json
@@ -1,0 +1,6 @@
+
+{
+  "type": "enhancement",
+  "category": "``Configure``",
+  "description": "Added Support for AWS temp creds with AWS configure, e.g. CLI will ask for aws_session_token if the access key id is temporary"
+}
--- a/awscli/customizations/configure/configure.py
+++ b/awscli/customizations/configure/configure.py
@@ -102,6 +102,7 @@ class ConfigureCommand(BasicCommand):
         # (logical_name, config_name, prompt_text)
         ('aws_access_key_id', "AWS Access Key ID"),
         ('aws_secret_access_key', "AWS Secret Access Key"),
+        ('aws_session_token', "AWS Session Token"),
         ('region', "Default region name"),
         ('output', "Default output format"),
     ]
@@ -115,17 +116,13 @@ class ConfigureCommand(BasicCommand):
             config_writer = ConfigFileWriter()
         self._config_writer = config_writer
 
+    def _needs_session_token(self, new_values):
+        """Check if session token is needed based on access key ID."""
+        access_key = new_values.get('aws_access_key_id')
+        return access_key and access_key.startswith('ASIA')
+
     def _run_main(self, parsed_args, parsed_globals):
         # Called when invoked with no args "aws configure"
-        # Check if there are any remaining unparsed arguments that might be invalid subcommands
-        if hasattr(parsed_args, 'remaining') and parsed_args.remaining:
-            sys.stderr.write(
-                f"Invalid subcommand: {' '.join(parsed_args.remaining)}\n"
-            )
-            sys.stderr.write(
-                "Valid subcommands are: list, get, set, add-model, import, list-profiles, sso, sso-session, mfa-login, export-credentials\n"
-            )
-            return 1
 
         new_values = {}
         # This is the config from the config file scoped to a specific
@@ -135,10 +132,11 @@ class ConfigureCommand(BasicCommand):
         except ProfileNotFound:
             config = {}
 
-        # Track if we need to prompt for session token
-        needs_session_token = False
-
         for config_name, prompt_text in self.VALUES_TO_PROMPT:
+            # Skip session token if not needed
+            if config_name == 'aws_session_token' and not self._needs_session_token(new_values):
+                continue
+                
             current_value = config.get(config_name)
             new_value = self._prompter.get_value(
                 current_value, config_name, prompt_text
@@ -146,32 +144,10 @@ class ConfigureCommand(BasicCommand):
             if new_value is not None and new_value != current_value:
                 new_values[config_name] = new_value
 
-            # Check if this is a temporary credential (starts with ASIA)
-            if (
-                config_name == 'aws_access_key_id'
-                and new_value
-                and new_value.startswith('ASIA')
-            ):
-                needs_session_token = True
-
-            # Prompt for session token after secret key but before region
-            if config_name == 'aws_secret_access_key' and needs_session_token:
-                session_token_current = config.get('aws_session_token')
-                session_token_new = self._prompter.get_value(
-                    session_token_current,
-                    'aws_session_token',
-                    'AWS Session Token',
-                )
-                if (
-                    session_token_new is not None
-                    and session_token_new != session_token_current
-                ):
-                    new_values['aws_session_token'] = session_token_new
-
         # Remove session token for non-temporary credentials
         if (
             'aws_access_key_id' in new_values
-            and not needs_session_token
+            and not self._needs_session_token(new_values)
             and config.get('aws_session_token')
         ):
             new_values['aws_session_token'] = None
--- a/awscli/customizations/configure/configure.py
+++ b/awscli/customizations/configure/configure.py
@@ -121,6 +121,16 @@ class ConfigureCommand(BasicCommand):
         access_key = new_values.get('aws_access_key_id')
         return access_key and access_key.startswith('ASIA')
 
+    def _should_prompt_for_session_token(self, new_values, config):
+        """Determine if we should prompt for session token."""
+        # Don't prompt if explicitly switching to long-term credentials
+        new_access_key = new_values.get('aws_access_key_id')
+        if new_access_key and not self._needs_session_token(new_values):
+            return False
+        
+        # Prompt if needed for temporary credentials or if already exists
+        return self._needs_session_token(new_values) or config.get('aws_session_token')
+
     def _run_main(self, parsed_args, parsed_globals):
         # Called when invoked with no args "aws configure"
 
@@ -133,8 +143,7 @@ class ConfigureCommand(BasicCommand):
             config = {}
 
         for config_name, prompt_text in self.VALUES_TO_PROMPT:
-            # Skip session token if not needed
-            if config_name == 'aws_session_token' and not self._needs_session_token(new_values):
+            if config_name == 'aws_session_token' and not self._should_prompt_for_session_token(new_values, config):
                 continue
                 
             current_value = config.get(config_name)
--- a/awscli/customizations/configure/configure.py
+++ b/awscli/customizations/configure/configure.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 import logging
 import os
+import sys
 
 from botocore.exceptions import ProfileNotFound
 
@@ -25,6 +26,7 @@ from awscli.customizations.configure.get import ConfigureGetCommand
 from awscli.customizations.configure.importer import ConfigureImportCommand
 from awscli.customizations.configure.list import ConfigureListCommand
 from awscli.customizations.configure.listprofiles import ListProfilesCommand
+from awscli.customizations.configure.mfalogin import ConfigureMFALoginCommand
 from awscli.customizations.configure.set import ConfigureSetCommand
 from awscli.customizations.configure.sso import (
     ConfigureSSOCommand,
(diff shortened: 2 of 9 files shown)


awscli/customizations/configure/configure.py
```diff
@@ -110,20 +117,64 @@ def __init__(self, session, prompter=None, config_writer=None):
...
+        # Check if there are any remaining unparsed arguments that might be invalid subcommands
+        if hasattr(parsed_args, 'remaining') and parsed_args.remaining:
+            sys.stderr.write(
+                f"Invalid subcommand: {' '.join(parsed_args.remaining)}\n"
+            )
+            sys.stderr.write(
+                "Valid subcommands are: list, get, set, add-model, import, list-profiles, sso, sso-session, mfa-login, export-credentials\n"
+            )
+            return 1
+
```
ashovlin: Was there a test case that prompted this?

For something invalid like `aws configure foo`, I'm hitting this block which already generates a similar message: https://github.com/aws/aws-cli/blob/a7a912426e2b9f86b876753126e34a291e3be13f/awscli/argparser.py#L79-L85

awscli/customizations/configure/configure.py
```diff
@@ -110,20 +117,64 @@ def __init__(self, session, prompter=None, config_writer=None):
...
+                session_token_new = self._prompter.get_value(
+                    session_token_current,
+                    'aws_session_token',
+                    'AWS Session Token',
+                )
+                if (
+                    session_token_new is not None
+                    and session_token_new != session_token_current
+                ):
+                    new_values['aws_session_token'] = session_token_new
```
ashovlin: Could we rework a bit for readability? Setting state in one iteration then conditionally inserting another while in the loop can be tough to reason about:
1. Add `aws_session_token` to `VALUES_TO_PROMPT`
2. Extract `needs_session_token` to a helper method that takes in `new_values` (so we're reusing the existing state)
3. Then skip the `aws_session_token` iteration based on something like `if config_name == 'aws_session_token' and not self._needs_session_token(new_values)`

.changes/next-release/configure-tempcreds.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "enhancement",
+  "category": "``Configure``",
+  "description": "Added Support for AWS temp creds with AWS configure, e.g. CLI will ask for aws_session_token if the access key id is temporary"
+},
+{
+  "type": "enhancement",
+  "category": "``Configure``",
+  "description": "Added new sub-command to aws confugre, `mfa-login`, which is an interactive wrapper for the aws sts get-session-token command"
+},
+]
```
ashovlin:
```suggestion
{
  "type": "enhancement",
  "category": "``configure``",
  "description": "Added support for temporary credentials with ``aws configure``. The CLI will prompt for an ``aws_session_token`` if the provided access key ID is temporary."
},
{
  "type": "enhancement",
  "category": "``configure``",
  "description": "Added the new ``aws configure mfa-login` command, which creates a profile with temporary credentials corresponding to an IAM user with an MFA code."
},
]
```

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+        'Sets up temporary credentials for MFA authentication. '
+        'This command creates a new profile with temporary credentials '
+        'obtained using the AWS STS get-session-token API.'
+    )
+    SYNOPSIS = (
+        'aws configure mfa-login [--profile profile-name] '
+        '[--update-profile profile-to-update] [--duration-seconds seconds] '
+        '[--serial-number mfa-serial-number]'
+    )
+    EXAMPLES = (
```
ashovlin: For both `DESCRIPTION` and `EXAMPLES`, you can load this from the `.rst` files instead of defining again here. See how `get.py` does it.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+        response = compat_input(f"{prompt_text}: ")
+        if not response:
+            return None
+        return response
+
+
+class ConfigureMFALoginCommand(BasicCommand):
+    """Configures MFA login for AWS CLI by creating temporary credentials."""
+
+    NAME = 'mfa-login'
```
ashovlin, line 53: Wondering if `login` is unnecessary since we're already working with credentials. `aws configure mfa` would be more concise? But will defer to you since you're in this space more.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+        },
+        {
+            'name': 'duration-seconds',
+            'help_text': (
+                'The duration, in seconds, that the credentials should remain valid. '
+                'Default is 43200 seconds (12 hours).'
+            ),
+            'action': 'store',
+            'required': False,
+            'cli_type_name': 'integer',
```
ashovlin: Can you use the `default` property here? That'd avoid need to define the `43200` again later.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+        if config_writer is None:
+            config_writer = ConfigFileWriter()
+        self._config_writer = config_writer
+
+    def _check_profile_exists(self, profile_name):
+        """Check if a profile exists using botocore's native profile handling."""
+        session = botocore.session.Session()
+        return profile_name in session.available_profiles
+
+    def _run_main(self, parsed_args, parsed_globals):
```
ashovlin: Throughout, can we either do something like:
1.  `configure.py`, which is iterating over `VALUES_TO_PROMPT`
2. `sso.py`, which has the separate `prompt_for_foo` helpers for each option

Breaking up the big `_run_main` will improve readability/maintainability/testability of this.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+                f"Error accessing profile {source_profile}: {str(e)}\n"
+            )
+            return 1
+
+        # Get MFA serial number
+        mfa_serial = parsed_args.serial_number or source_config.get(
+            'mfa_serial'
+        )
+        if not mfa_serial:
+            if sys.stdin.isatty():
```
ashovlin: Throughout other custom commands, we don't really guard against `isatty`. Was there something specific that prompted it here?

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+        else:
+            sys.stderr.write("MFA token code is required\n")
+            return 1
+
+        # If no target profile is specified, generate a default name
+        if not target_profile:
+            random_suffix = ''.join(
+                random.choices(string.ascii_lowercase + string.digits, k=5)
+            )
+            target_profile = f"session-{random_suffix}"
```
ashovlin: Would it be helpful to use something deterministic here instead, like how `aws configure sso` does `{role name}-{account id}`?

I don't use MFA day-to-day, but wondering if a user would run it ~daily to keep updating the same long-running profile? I know we have `--update-profile`, but at least I'm lazy sometimes.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+            )
+        except ClientError as e:
+            sys.stderr.write(f"An error occurred: {e}\n")
+            return 1
+
+        # Extract credentials from response
+        temp_credentials = response['Credentials']
+
+        # Write credentials to the credentials file
+        credentials_file = os.path.expanduser('~/.aws/credentials')
```
ashovlin: Can you load this path via `os.path.expanduser(self._session.get_config_variable('credentials_file')` like in `configure.py`? 

That'll honor if the user is using an alternative path via ` AWS_SHARED_CREDENTIALS_FILE`.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+            'aws_session_token': temp_credentials['SessionToken'],
+        }
+
+        # Get expiration time as a string
+        try:
+            expiration_time = temp_credentials['Expiration'].strftime(
+                '%Y-%m-%d %H:%M:%S UTC'
+            )
+        except Exception:
+            # Handle case where Expiration might not be a datetime object
```
ashovlin: Can you catch that specific exception then?

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,400 @@
...
+            f"Temporary credentials written to profile '{target_profile}'\n"
+        )
+        sys.stdout.write(f"Credentials will expire at {expiration_time}\n")
+        sys.stdout.write(
+            f"To use these credentials, specify --profile {target_profile} when running AWS CLI commands\n"
+        )
+
+        return 0
+
+    def _handle_missing_default_profile(self, parsed_args, duration_seconds):
```
ashovlin: A few chunks in here are duplicated of above, could we consolidate?

awscli/customizations/configure/configure.py
liwadman: no, removing.

awscli/customizations/configure/configure.py
liwadman: good callout, added this in forthcoming commit.

awscli/customizations/configure/mfalogin.py
liwadman: I think I prefer mfa-login because of the verb "login" - you're "logging in" in this workflow, not configuring an MFA for future use.

awscli/customizations/configure/mfalogin.py
liwadman: that's good feedback. Because we have a deterministic principal here always (IAM user), we can make a predictable profile based on things we know.

awscli/customizations/configure/mfalogin.py
liwadman: agreed not necesarry. Though the command is setup to be run interactively like aws configure sso.

awscli/customizations/configure/mfalogin.py
liwadman: done

awscli/customizations/configure/mfalogin.py
liwadman: done

awscli/customizations/configure/mfalogin.py
liwadman: done

awscli/customizations/configure/configure.py
```diff
@@ -95,35 +102,55 @@ class ConfigureCommand(BasicCommand):
...
         # profile.
         try:
             config = self._session.get_scoped_config()
         except ProfileNotFound:
             config = {}
+
         for config_name, prompt_text in self.VALUES_TO_PROMPT:
+            # Skip session token if not needed
+            if config_name == 'aws_session_token' and not self._needs_session_token(new_values):
+                continue
```
ashovlin: If I have a profile configure with temporary credentials, and I run through `aws configure` again (say I want to change my region), it doesn't show the session token prompt.
```
✗ aws configure
AWS Access Key ID [****************NUNU]: 
AWS Secret Access Key [****************VbHX]: 
Default region name [us-east-1]: 
Default output format [json]: 
```

Granted I don't think you'd change it without updating the access key and secret, but for consistency I think it'd be better to show it.

We might want to consider something like `(config already has a session token and we're not changing the access key) or (new_values has a ASIA... access key)`?

awscli/examples/configure/mfa-login.rst
```diff
@@ -0,0 +1,49 @@
...
+
+The following ``mfa-login`` command creates a new profile with temporary credentials obtained using MFA authentication. ::
+
+    aws configure mfa-login
+
+Output::
+
+    MFA serial number or ARN: arn:aws:iam::123456789012:mfa/user
+    MFA token code: 123456
+    Profile to update [session-12345]:
```
ashovlin: Minor, but update `session-12345` throughout both examples files with the new format

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,327 @@
...
+    def __init__(self, session, prompter=None, config_writer=None):
+        super().__init__(session)
+        if prompter is None:
+            prompter = InteractiveMFAPrompter()
+        self._prompter = prompter
+        if config_writer is None:
+            config_writer = ConfigFileWriter()
+        self._config_writer = config_writer
+
+    def _check_profile_exists(self, profile_name):
```
ashovlin: nit: Nothing is calling this anymore? If so, remove.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,327 @@
...
+        try:
+            expiration_time = temp_credentials['Expiration'].strftime(
+                '%Y-%m-%d %H:%M:%S UTC'
+            )
+        except AttributeError:
+            expiration_time = str(temp_credentials['Expiration'])
+
+        credential_values['#Credentials expire at: '] = (
+            f"{expiration_time}"
+        )
```
ashovlin: It might be happening in `writer.py`, but as I refresh by re-running `mfa-login` I'm seeing the expiration lines stack up
```
[12345-google-authenticator]
aws_access_key_id = ...
aws_secret_access_key = ...
aws_session_token = ...
#Credentials expire at:  = 2025-08-28 01:09:31 UTC
#Credentials expire at:  = 2025-08-28 01:50:53 UTC
#Credentials expire at:  = 2025-08-28 01:51:16 UTC
#Credentials expire at:  = 2025-08-28 01:51:38 UTC
```

ashovlin: I could take or leave this. I'm not in love with the `:  = ` syntax, but don't want to ask you to plumb through official "comment support" in this PR. I like trying to be helpful, but on the other hand we won't stamp these if you configure temp creds through `aws configure`.

An upcoming project might need an actual expiration field, so if that shakes out we can convert this.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,327 @@
...
+            f"To use these credentials, specify --profile {target_profile} when running AWS CLI commands\n"
+        )
+        return 0
+
+    def _run_main(self, parsed_args, parsed_globals):
+        source_profile = parsed_globals.profile or 'default'
+        duration_seconds = parsed_args.duration_seconds
+
+        # Setup session and validate profile
+        session, source_config = self._setup_session_with_profile(source_profile)
```
ashovlin: As long as `source_profile` corresponds to `--profile` (or `default`), it'd be better to use the `session` that is passed into `__init__`. 

The CLI will already initialize that using `--profile`, as well as other global options like `--region`, `--endpoint_url`, `--ca-bundle`, etc. One of our footguns is custom commands defining their own sessions/clients, which then don't honor the global options that users expect.

awscli/customizations/configure/mfalogin.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+    EXAMPLES = BasicCommand.FROM_FILE('configure', 'mfa-login', '_examples.rst')
+    ARG_TABLE = [
+        {
+            'name': 'profile',
+            'help_text': ('Use a specific profile from your credential file.'),
+            'action': 'store',
+            'required': False,
+            'cli_type_name': 'string',
+        },
```
ashovlin: This is causing a test failure in CI because the CLI already exposes `--profile` for every command.
```
FAILED functional\test_shadowing.py::test_no_shadowed_builtins[configure-command_table415-builtins415] - AssertionError: 
Shadowing/Prefixing a top level option: configure.mfa-login.profile
```

As long as we interpret the CLI's existing/global `--profile` to mean the source profile for this workflow, we don't need to define it again here.

Jives with https://github.com/aws/aws-cli/pull/9590/files#r2304086080

awscli/customizations/configure/mfalogin.py
liwadman: I think it's probably useful enough to print them in stdout like we do below. will remove.

ashovlin (approved):

PR #9591, opened 2025-07-14
[v2] Add test to ensure retries reuse request checksum

hssyoo:
Ports https://github.com/boto/botocore/pull/3516

This commit adds a test to ensure that request checksums are only calculated and applied once, even in the event of retries.

### DIFF
--- a/tests/functional/botocore/test_s3.py
+++ b/tests/functional/botocore/test_s3.py
@@ -19,6 +19,7 @@ from botocore.compat import datetime, parse_qs, urlsplit
 from botocore.config import Config
 from botocore.exceptions import (
     ClientError,
+    ConnectionError,
     ParamValidationError,
     UnsupportedS3AccesspointConfigurationError,
     UnsupportedS3ConfigurationError,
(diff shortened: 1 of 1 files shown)


tests/functional/botocore/test_s3.py
```diff
@@ -2191,6 +2192,32 @@ def test_checksum_content_encoding(content_encoding, expected_header):
...
+    mock_urllib3_session_send.side_effect = ConnectionError(error='Fake error')
+    op_kwargs = {
+        "Bucket": "mybucket",
+        "Key": "mykey",
+        "Body": b"foo",
+        "ChecksumAlgorithm": "CRC32",
+    }
+    s3 = _create_s3_client(
+        retries={
+            'max_attempts': 2,
```
aemous, line 2210: Just curious, why is the max attempts for CLI v2 configured as 2 while it is configured as 1 for boto3?

hssyoo: So I believe retry logic in `botocore` predates standardized retry behavior across SDKs. `botocore` already had a `max_attempts` argument you pass into the `Config()` object, which represents the max retry attempts. But when we added the `max_attempts` environment and config file variable, those represented max attempts (including initial request). In order to support something similar in the `Config()` object, `botocore` also added an argument called `total_max_requests`, which represents max attempts.

AWS CLI v2 was released after all this, so it simplified the logic. Now our vended `botocore`'s `Config()` object takes an argument `max_attempts`, which represents max attempts and is consistent with `max_attempts` in the config file and env var.

aemous: Makes sense, thanks

aemous (approved):
LGTM

PR #9569, opened 2025-06-24
Implemented No-overwrite for uploads using cp command

rakshil14-2:
*Issue :*  Github Issue #2874 

*Description of changes:*

Customers requested to prevent of overwriting of objects. This is done by providing `no-overwrite` header using `ifNoneMatch` feature of Amazon S3 which only allows to upload objects that are not present on the bucket. This header also supports multi-part upload as well as works well with other cp header.
If the user tries to upload the object which is already present at the target bucket using command line, a `warning` about skipping the file will be displayed and that object is not allowed to upload to S3 bucket.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #9592, opened 2025-07-14
Adding No overwrite for copy operation

rakshil14-2:
*Issue :*  Github Issue #2874 

*Description of changes:*

Customers requested to prevent overwriting of objects. This is done by providing `no-overwrite` header using `ifNoneMatch` with Etags  to transfer objects that are not present on the bucket. However, retrieval of etags is difficult hence during copy operation always perform multipart copy since it follows multipart upload approach which indeed supports `ifNoneMatch` header without Etags. The `no-overwrite` header works well with other `cp` command headers. Customer can be implemented using `aws s3 cp <source> <destination> --no-overwrite` which allows successful transfers when object with the same name is not present on the destination. However, if the user tries to transfer the object which is already present at the target bucket using command line, a `warning` about skipping the file will be generated and that object is not allowed to transfer to S3 bucket.

_Testing_:
 Functional testing is performed to validate the object copy process. This included both scenarios copying an object when object with same key is already present in the bucket as well as those when the object with same key is not present on bucket. These tests were conducted for both object having size less than multipart threshold and one with greater than multipart threshold

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -23,7 +23,6 @@ from awscli.compat import ensure_text_type, queue
 from awscli.customizations.s3.subscribers import OnDoneFilteredSubscriber
 from awscli.customizations.s3.utils import (
     WarningResult,
-    create_warning,
     human_readable_size,
 )
 from awscli.customizations.utils import uni_print
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -13,6 +13,7 @@
 import logging
 import os
 
+from botocore.exceptions import ClientError
 from s3transfer.manager import TransferManager
 
 from awscli.compat import get_binary_stdin
@@ -481,7 +482,46 @@ class CopyRequestSubmitter(BaseTransferRequestSubmitter):
         )
 
     def _get_warning_handlers(self):
-        return [self._warn_glacier]
+        return [
+            self._warn_glacier,
+            self._warn_if_zero_byte_file_exists_with_no_overwrite,
+        ]
+
+    def _warn_if_zero_byte_file_exists_with_no_overwrite(self, fileinfo):
+        """
+        Warning handler to skip zero-byte files when no_overwrite is set and file exists.
+
+        This method handles the transfer of zero-byte objects when the no-overwrite parameter is specified.
+        To prevent overwrite, it uses head_object to verify if the object exists at the destination:
+        If the object is present at destination: skip the file (return True)
+        If the object is not present at destination: allow transfer (return False)
+
+        :type fileinfo: FileInfo
+        :param fileinfo: The FileInfo object containing transfer details
+
+        :rtype: bool
+        :return: True if file should be skipped, False if transfer should proceed
+        """
+        if not self._cli_params.get('no_overwrite') or (
+            getattr(fileinfo, 'size') and fileinfo.size > 0
+        ):
+            return False
+
+        bucket, key = find_bucket_key(fileinfo.dest)
+        client = fileinfo.source_client
+        try:
+            client.head_object(Bucket=bucket, Key=key)
+            LOGGER.debug(
+                "Warning: Skipping file %s as it already exists on %s",
+                fileinfo.src,
+                fileinfo.dest,
+            )
+            return True
+        except ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                return False
+            else:
+                raise
 
     def _format_src_dest(self, fileinfo):
         src = self._format_s3_path(fileinfo.src)
--- a/awscli/s3transfer/copies.py
+++ b/awscli/s3transfer/copies.py
@@ -13,6 +13,7 @@
 import copy
 import math
 
+from botocore.exceptions import ClientError
 from s3transfer.tasks import (
     CompleteMultipartUploadTask,
     CreateMultipartUploadTask,
@@ -67,6 +68,7 @@ class CopySubmissionTask(SubmissionTask):
         'CopySourceSSECustomerKeyMD5',
         'MetadataDirective',
         'TaggingDirective',
+        'IfNoneMatch',
     ]
 
     COMPLETE_MULTIPART_ARGS = [
@@ -75,6 +77,11 @@ class CopySubmissionTask(SubmissionTask):
         'SSECustomerKeyMD5',
         'RequestPayer',
         'ExpectedBucketOwner',
+        'IfNoneMatch',
+    ]
+
+    COPY_OBJECT_ARGS_BLOCKLIST = [
+        'IfNoneMatch',
     ]
 
     def _submit(
@@ -98,6 +105,7 @@ class CopySubmissionTask(SubmissionTask):
         :param transfer_future: The transfer future associated with the
             transfer request that tasks are being submitted for
         """
+        call_args = transfer_future.meta.call_args
         # Determine the size if it was not provided
         if transfer_future.meta.size is None:
             # If a size was not provided figure out the size for the
@@ -105,7 +113,6 @@ class CopySubmissionTask(SubmissionTask):
             # the TransferManager. If the object is outside of the region
             # of the client, they may have to provide the file size themselves
             # with a completely new client.
-            call_args = transfer_future.meta.call_args
             head_object_request = (
                 self._get_head_object_request_from_copy_source(
                     call_args.copy_source
@@ -127,10 +134,24 @@ class CopySubmissionTask(SubmissionTask):
             transfer_future.meta.provide_transfer_size(
                 response['ContentLength']
             )
-
-        # If it is greater than threshold do a multipart copy, otherwise
-        # do a regular copy object.
-        if transfer_future.meta.size < config.multipart_threshold:
+        # Check for ifNoneMatch is enabled and file has content
+        # Special handling for 0-byte files: Since multipart copy works with object size
+        # and divides the object into smaller chunks, there's an edge case when the object
+        # size is zero. This would result in 0 parts being calculated, and the
+        # CompleteMultipartUpload operation throws a MalformedXML error when transferring
+        # 0 parts because the XML does not validate against the published schema.
+        # Therefore, 0-byte files are always handled via single copy request regardless
+        # of the multipart threshold setting.
+        should_overwrite = (
+            call_args.extra_args.get("IfNoneMatch")
+            and transfer_future.meta.size != 0
+        )
+        # If it is less than threshold and ifNoneMatch is not in parameters
+        # do a regular copy else do multipart copy.
+        if (
+            transfer_future.meta.size < config.multipart_threshold
+            and not should_overwrite
+        ):
             self._submit_copy_request(
                 client, config, osutil, request_executor, transfer_future
             )
@@ -147,19 +168,25 @@ class CopySubmissionTask(SubmissionTask):
         # Get the needed progress callbacks for the task
         progress_callbacks = get_callbacks(transfer_future, 'progress')
 
-        # Submit the request of a single copy.
+        # Submit the request of a single copy and make sure it
+        # does not include any blocked arguments.
+        copy_object_extra_args = {
+            param: val
+            for param, val in call_args.extra_args.items()
+            if param not in self.COPY_OBJECT_ARGS_BLOCKLIST
+        }
         self._transfer_coordinator.submit(
             request_executor,
             CopyObjectTask(
                 transfer_coordinator=self._transfer_coordinator,
                 main_kwargs={
-                    'client': client,
-                    'copy_source': call_args.copy_source,
-                    'bucket': call_args.bucket,
-                    'key': call_args.key,
-                    'extra_args': call_args.extra_args,
-                    'callbacks': progress_callbacks,
-                    'size': transfer_future.meta.size,
+                    "client": client,
+                    "copy_source": call_args.copy_source,
+                    "bucket": call_args.bucket,
+                    "key": call_args.key,
+                    "extra_args": copy_object_extra_args,
+                    "callbacks": progress_callbacks,
+                    "size": transfer_future.meta.size,
                 },
                 is_final=True,
             ),
--- a/awscli/s3transfer/manager.py
+++ b/awscli/s3transfer/manager.py
@@ -188,6 +188,7 @@ class TransferManager:
         'SSEKMSEncryptionContext',
         'Tagging',
         'WebsiteRedirectLocation',
+        'IfNoneMatch',
     ]
 
     ALLOWED_UPLOAD_ARGS = (
@@ -195,7 +196,6 @@ class TransferManager:
         + [
             'ChecksumType',
             'MpuObjectSize',
-            'IfNoneMatch',
         ]
         + FULL_OBJECT_CHECKSUM_ARGS
     )
--- a/tests/functional/s3/test_cp_command.py
+++ b/tests/functional/s3/test_cp_command.py
@@ -335,7 +335,6 @@ class TestCPCommand(BaseCPCommandTest):
         # Create a large file that will trigger multipart upload
         full_path = self.files.create_file('foo.txt', 'a' * 10 * (1024**2))
         cmdline = f'{self.prefix} {full_path} s3://bucket --no-overwrite'
-
         # Set up responses for multipart upload
         self.parsed_responses = [
             {'UploadId': 'foo'},  # CreateMultipartUpload response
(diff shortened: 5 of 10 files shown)


awscli/s3transfer/copies.py
```diff
@@ -127,10 +133,17 @@ def _submit(
...
                 response['ContentLength']
             )
-
-        # If it is greater than threshold do a multipart copy, otherwise
-        # do a regular copy object.
-        if transfer_future.meta.size < config.multipart_threshold:
+        # Check if no-overwrite is enabled and file has content
+        no_overwrite_requested = (
+            call_args.extra_args.get("IfNoneMatch")
+            and transfer_future.meta.size != 0
```
aemous, line 147: Considering an edge case

Let's say a file `foo.txt` is in bucket B. Let's say it has a lot of content, 1GB size.
Then, let's say i have a file `foo.txt` in my bucket A, and it is empty. 

If I do `aws s3 cp s3://A/foo.txt s3://B/foo.txt --no-overwrite`, I would expect `foo.txt` in bucket B to not be updated. According to this code, it seems it may actually overwrite the file, deleting the 1GB of contents.

Is this correct? I'd think we would want to change this behavior. For example, what if the source file has size 0, we use ListObjectsV2 to check if the object exists? Could be worth discussing with team

awscli/s3transfer/copies.py
```diff
@@ -127,10 +133,17 @@ def _submit(
             transfer_future.meta.provide_transfer_size(
                 response['ContentLength']
             )
-
-        # If it is greater than threshold do a multipart copy, otherwise
-        # do a regular copy object.
-        if transfer_future.meta.size < config.multipart_threshold:
+        # Check if no-overwrite is enabled and file has content
+        no_overwrite_requested = (
```
kdaily: Nit: the name here is a little misleading, as it also includes something the user is not directly setting (the file size). Would rename to something like `should_not_overwrite` (confirm the logic in the name matches the intention)?

kdaily: I think we should leave a comment on why a file size of 0 gets special treatment here too.

awscli/s3transfer/manager.py
```diff
@@ -188,14 +188,14 @@ class TransferManager:
...
         'WebsiteRedirectLocation',
+        'IfNoneMatch',
     ]
 
     ALLOWED_UPLOAD_ARGS = (
         _ALLOWED_SHARED_ARGS
         + [
             'ChecksumType',
             'MpuObjectSize',
-            'IfNoneMatch',
```
kdaily, line 198: Why is this getting removed here?

awscli/customizations/s3/results.py
```diff
@@ -23,7 +23,6 @@
 from awscli.customizations.s3.subscribers import OnDoneFilteredSubscriber
 from awscli.customizations.s3.utils import (
     WarningResult,
-    create_warning,
```
kdaily, line 26: Is this being removed because we forgot to clean it up/remove it in #9569?

awscli/s3transfer/copies.py
```diff
@@ -75,6 +76,11 @@ class CopySubmissionTask(SubmissionTask):
         'SSECustomerKeyMD5',
         'RequestPayer',
         'ExpectedBucketOwner',
+        'IfNoneMatch',
+    ]
+
+    COPY_OBJECT_BLOCKLIST = [
```
kdaily: Nit: should probably be `COPY_OBJECT_ARGS_BLOCKLIST`

awscli/s3transfer/copies.py
```diff
@@ -147,7 +160,13 @@ def _submit_copy_request(
...
         progress_callbacks = get_callbacks(transfer_future, 'progress')
 
-        # Submit the request of a single copy.
+        # Submit the request of a single copy and make sure it
+        # does not include any blocked arguments.
+        copy_object_extra_args = {}
+        for param, val in call_args.extra_args.items():
+            if param not in self.COPY_OBJECT_BLOCKLIST:
+                copy_object_extra_args[param] = val
+
```
aemous: Why is this needed for single-part copies? 

Does the CLI S3 customization ever add IfNoneMatch to single-part copies?

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    def test_no_overwrite_flag_on_copy_when_small_object_exists_on_target(
+        self,
+    ):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket --no-overwrite'
+        )
+        # Set up responses for multipart copy (since no-overwrite always uses multipart)
+        self.parsed_responses = [
+            self.head_object_response(),  # HeadObject to get source metadata
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: nit: Can we use `create_mpu_response()` here?

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            {'CopyPartResult': {'ETag': '"foo-1"'}},  # UploadPartCopy response
```
aemous: nit: Can we use `upload_part_copy_response` here?

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            {
+                'ResponseMetadata': {'HTTPStatusCode': 412},
+                'Error': {
+                    'Code': 'PreconditionFailed',
+                    'Message': 'At least one of the pre-conditions you specified did not hold',
+                    'Condition': 'If-None-Match',
+                },
```
aemous: * Can we remove ResponseMetadata to be consistent with our existing tests? Even though it's returned by botocore, as long as there is an Error key, that is a sufficient condition for being an error. (if S3 makes a breaking change to change the HTTP code of the response, this test may no longer be representative of the response.)
* Can we create a new `precondition_failed_error_response(self, condition)` helper function in  `tests/functional/s3/__init__.py`? And we can use that in place of duplicating the response structure in each test.

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+
+    def test_no_overwrite_flag_on_copy_when_object_is_of_zero_size(self):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/file.txt s3://bucket2 --no-overwrite'
+        )
+        self.parsed_responses = [
+            {
+                "ContentLength": 0,
+                "LastModified": "00:00:00Z",
+                'ETag': '"d41d8cd98f00b204e9800998ecf8427e"',
```
aemous: nit: Consider using `head_object_response(ContentLength=100)` instead.

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            },  # HeadObject with content length 0
+            {
+                'CopyObjectResult': {
+                    'ETag': '"d41d8cd98f00b204e9800998ecf8427e"'
+                }
```
aemous: * Consider using `copy_object_response()` instead

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket --no-overwrite'
+        )
+        # Set up responses for multipart copy with large object
+        self.parsed_responses = [
+            self.head_object_response(
+                ContentLength=10 * (1024**2)
+            ),  # HeadObject with large content
+            {'TagSet': []},  # GetObjectTagging response
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: consider `create_mpu_response()` instead

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            },  # UploadPartCopy response part 1
+            {
+                'CopyPartResult': {'ETag': '"foo-2"'}
+            },  # UploadPartCopy response part 2
+            {
+                'ResponseMetadata': {'HTTPStatusCode': 412},
+                'Error': {
+                    'Code': 'PreconditionFailed',
+                    'Message': 'At least one of the pre-conditions you specified did not hold',
+                    'Condition': 'If-None-Match',
```
aemous: similar comment as before, remove ResponseMetadata, reuse precondition_failed_error() helper

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket --no-overwrite'
+        )
+        # Set up responses for multipart copy with large object
+        self.parsed_responses = [
+            self.head_object_response(
+                ContentLength=10 * (1024**2)
+            ),  # HeadObject with large content
+            {'TagSet': []},  # GetObjectTagging response
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: create_mpu_response instead ?

tests/functional/s3/test_cp_command.py
```diff
@@ -397,6 +396,177 @@ def test_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    ):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket --no-overwrite'
+        )
+        # Set up responses for multipart copy with large object
+        self.parsed_responses = [
+            self.head_object_response(
+                ContentLength=10 * (1024**2)
+            ),  # HeadObject with large content
+            {'TagSet': []},  # GetObjectTagging response
```
aemous: `get_object_tagging_response([])` instead ?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    def test_mv_no_overwrite_flag_on_copy_when_small_object_does_not_exist_on_target(
+        self,
+    ):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket2 --no-overwrite'
+        )
+        # Set up responses for multipart copy (since no-overwrite always uses multipart)
+        self.parsed_responses = [
+            self.head_object_response(),  # HeadObject to get source metadata
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: `create_mpu_response('foo')` instead ?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    def test_mv_no_overwrite_flag_on_copy_when_small_object_exists_on_target(
+        self,
+    ):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key.txt s3://bucket2 --no-overwrite'
+        )
+        # Set up responses for multipart copy (since no-overwrite always uses multipart)
+        self.parsed_responses = [
+            self.head_object_response(),  # HeadObject to get source metadata
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: `create_mpu_response` ?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            {'CopyPartResult': {'ETag': '"foo-1"'}},  # UploadPartCopy response
+            {
+                'ResponseMetadata': {'HTTPStatusCode': 412},
+                'Error': {
+                    'Code': 'PreconditionFailed',
+                    'Message': 'At least one of the pre-conditions you specified did not hold',
+                    'Condition': 'If-None-Match',
+                },
```
aemous: similar comment as before on this error

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    def test_mv_no_overwrite_flag_on_copy_when_object_is_of_zero_size(self):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/file.txt s3://bucket2 --no-overwrite'
+        )
+        self.parsed_responses = [
+            {
+                "ContentLength": 0,
+                "LastModified": "00:00:00Z",
+                'ETag': '"d41d8cd98f00b204e9800998ecf8427e"',
+            },  # HeadObject with content length 0
```
aemous: `head_object_response(ContentLength=0)` instead?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+        self.assertEqual(self.operations_called[1][0].name, 'CopyObject')
+        self.assertEqual(self.operations_called[2][0].name, 'DeleteObject')
+
+    def test_mv_no_overwrite_flag_when_large_object_exists_on_target(self):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key1.txt s3://bucket --no-overwrite'
+        )
+        self.parsed_responses = [
+            self.head_object_response(ContentLength=10 * (1024**2)),
+            {'TagSet': []},  # GetObjectTagging response
```
aemous: `get_object_tagging_response([])` instead?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: `create_mpu_response('foo')` instead?

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+            {
+                'CopyPartResult': {'ETag': '"foo-2"'}
+            },  # UploadPartCopy response part 2
+            {
+                'ResponseMetadata': {'HTTPStatusCode': 412},
+                'Error': {
+                    'Code': 'PreconditionFailed',
+                    'Message': 'At least one of the pre-conditions you specified did not hold',
+                    'Condition': 'If-None-Match',
+                },
```
aemous: similar response on this error

tests/functional/s3/test_mv_command.py
```diff
@@ -422,6 +422,176 @@ def test_mv_no_overwrite_flag_multipart_upload_when_object_exists_on_target(
...
+    def test_mv_no_overwrite_flag_when_large_object_does_not_exist_on_target(
+        self,
+    ):
+        cmdline = (
+            f'{self.prefix} s3://bucket1/key1.txt s3://bucket --no-overwrite'
+        )
+        self.parsed_responses = [
+            self.head_object_response(ContentLength=10 * (1024**2)),
+            {'TagSet': []},  # GetObjectTagging response
+            {'UploadId': 'foo'},  # CreateMultipartUpload response
```
aemous: `get_object_tagging_response([])` and `create_mpu_response('foo')`

awscli/s3transfer/copies.py
```diff
@@ -319,6 +322,59 @@ def _get_transfer_size(
             return total_transfer_size - (part_index * part_size)
         return part_size
 
+    def _check_destination_object_existence(self, client, call_args):
+        """
+        Check if destination object exists when IfNoneMatch is specified.
+
+        This method implements the no-overwrite functionality by checking if the
```
aemous: Let's not refer to "no-overwrite" in any code in `s3transfer`; no code in S3transfer should be specific to the CLI. 

S3Transfer is also used by boto3, and no-overwrite is not a concept in boto3. Same comment for line 141

awscli/s3transfer/copies.py
```diff
@@ -319,6 +322,59 @@ def _get_transfer_size(
...
+            self._transfer_coordinator.set_exception(
+                client.exceptions.ClientError(
+                    {
+                        'Error': {
+                            'Code': 'PreconditionFailed',
+                            'Message': 'At least one of the pre-conditions you specified did not hold',
+                            'StatusCode': 412,
+                        }
+                    },
+                    'HeadObject',
```
aemous: You're throwing `PreConditionFailed` from `HeadObject` here, but HeadObject actually succeeded. I think this should be CompleteMultipartUpload in my opinion; that's the request we expect to return PreConditionFailed with IfNoneMatch supplied.

awscli/s3transfer/copies.py
```diff
@@ -319,6 +322,59 @@ def _get_transfer_size(
...
+                )
+            )
+            return False
+        except Exception as e:
```
aemous: I'd prefer catching a more specific exception type, such as `ClientError`. It's generally better practice to catch the most specific error type you expect the try body to raise.

awscli/s3transfer/copies.py
```diff
@@ -319,6 +322,59 @@ def _get_transfer_size(
...
+            error_code = None
+            if hasattr(e, 'response'):
+                error_code = e.response.get('Error', {}).get('Code')
+                if (
+                    error_code == '404'
```
aemous: Do we expect error_code to be `404` or `NoSuchKey`? We should find out which to expect and only check that value (unless both can be returned depending on the case, but I don't think that's true).

awscli/s3transfer/copies.py
```diff
@@ -319,6 +322,59 @@ def _get_transfer_size(
...
+          and returns False (terminate process)
+        If IfNoneMatch is specified and destination object doesn't exist (404/NoSuchKey):
+          Returns True (proceed with copy)
+        If other errors occur: Re-raises the exception
+
+        """
+        if not call_args.extra_args.get("IfNoneMatch"):
+            return True
+        try:
+            client.head_object(Bucket=call_args.bucket, Key=call_args.key)
```
aemous: If the user executes `aws s3 cp s3://bucketA s3://bucketB --recursive`? It seems this will add one extra `HeadObject` request for each object in the source bucket; a more performant/frugal solution would be to call ListObjectsV2 in the CLI *before* submitting the request to S3transfer. We could likely use our FileGenerator code for this: we currently have one filegenerator (for the source) for cp/mv when --recursive is supplied; we could add an extra check to see if --no-overwrite is supplied, and add a *second* file generator (for the destination) when no-overwrite is used with --recursive.

The solution I just proposed would be cleaner too, since we would know to cancel an upload *before* trying to submit the file to s3transfer.

May not be a strict requirement from me as an approver, let's discuss.

awscli/s3transfer/copies.py
rakshil14-2: My thinking is that we are trying to replicate `ifNoneMatch` functionality, where if the object is present we are getting `PreConditionFailed` Error hence, with `headObject` if object is found at the destination, it should also throw the same error

kdaily: Raising an HTTP exception for an operation that is not currently able to return that error seems off to me.

awscli/s3transfer/copies.py
```diff
@@ -300,6 +323,52 @@ def _get_transfer_size(
...
+        client: The S3 client to use for the head_object call
+        call_args: The transfer call arguments containing bucket, key, and extra_args
+
+        Returns:
+            bool: True if copy should proceed, False if process should terminate
+
+        Behavior:
+        If IfNoneMatch is not specified: Returns True (proceed normally)
+        If IfNoneMatch is specified and destination object exists:
+          Sets PreconditionFailed exception achieveing behaviour of ifNoneMatch header
```
kdaily: nit: spelling
```suggestion
          Sets PreconditionFailed exception achieving behaviour of ifNoneMatch header
```

awscli/s3transfer/copies.py
```diff
@@ -300,6 +323,52 @@ def _get_transfer_size(
             return total_transfer_size - (part_index * part_size)
         return part_size
 
+    def _check_destination_object_existence(self, client, call_args):
```
kdaily: nit: generally method names also use an active voice. How about:

```suggestion
    def _destination_object_exists(self, client, call_args):
```

so now `if _destination_object_exists()` sounds better.

awscli/s3transfer/copies.py
```diff
@@ -147,23 +161,32 @@ def _submit_copy_request(
...
-                },
-                is_final=True,
-            ),
-        )
+        # Submit the request of a single copy and make sure it
+        # does not include any blocked arguments.
+        copy_object_extra_args = {}
+        for param, val in call_args.extra_args.items():
+            if param not in self.COPY_OBJECT_BLOCKLIST:
+                copy_object_extra_args[param] = val
```
kdaily: nit, could use a list comprehension here.

```suggestion
        copy_object_extra_args = {
            param, val in call_args.extra_args.items() if param not in self.COPY_OBJECT_BLOCKLIST
        }
```

awscli/s3transfer/copies.py
```diff
@@ -300,6 +323,52 @@ def _get_transfer_size(
             return total_transfer_size - (part_index * part_size)
         return part_size
 
+    def _check_destination_object_existence(self, client, call_args):
+        """
+        Check if destination object exists when IfNoneMatch is specified.
+
+        This method implements the ifNoneMatch header functionality by checking if the
+        destination object already exists before proceeding with the copy operation.
+        Args:
```
kdaily: nit: if documenting args, then we should use existing standards for docstrings, which is to use Sphinx markup. See here for an example for parameters and return values.

https://github.com/aws/aws-cli/blob/e0dba6bc2aa29ab31302419d6764fcd5fb069f96/awscli/argprocess.py#L138-L145

awscli/s3transfer/copies.py
```diff
@@ -300,6 +323,52 @@ def _get_transfer_size(
             return total_transfer_size - (part_index * part_size)
         return part_size
 
+    def _check_destination_object_existence(self, client, call_args):
+        """
+        Check if destination object exists when IfNoneMatch is specified.
+
+        This method implements the ifNoneMatch header functionality by checking if the
```
kdaily: This statement is confusing. I'm not sure I understand why all of the implementation is based around `IfNoneMatch`? For example, we could base all functionality around head object pre-checks, and then any logic using `IfNoneMatch` would be altered to have that behavior?

kdaily (changes requested):

awscli/customizations/s3/results.py
rakshil14-2: yes

awscli/s3transfer/manager.py
rakshil14-2: Because it used  for copying as well so placed as a Shared args for both upload and copy

rakshil14-2: Because this is shared argument for both upload and copy

awscli/customizations/s3/s3handler.py
```diff
@@ -488,6 +496,36 @@ def _format_src_dest(self, fileinfo):
         dest = self._format_s3_path(fileinfo.dest)
```
rakshil14-2: File Object coming from Subcommands. This is generated in FileInfo class

kdaily (changes requested):
Good start! I think the logic makes sense. I think that the function names and documentation could be clarified a bit.

Also, see the suggestion about where the check is determined. There is an existing list of warning handlers that is used to skip files - could the logic be handled there, since it only needs the `FileInfo` object to determine?

Also, I noticed that there is a "flicker" in the terminal output when skipping a file with this. We might need to check how the progress status is updated in this case.

awscli/customizations/s3/s3handler.py
```diff
@@ -471,6 +472,13 @@ def _submit_transfer_request(self, fileinfo, extra_args, subscribers):
         bucket, key = find_bucket_key(fileinfo.dest)
         source_bucket, source_key = find_bucket_key(fileinfo.src)
         copy_source = {'Bucket': source_bucket, 'Key': source_key}
+        if not self._should_transfer_begin(fileinfo):
```
kdaily: Could this get registered in `_get_warning_handlers` instead of modifying this method?

awscli/s3transfer/copies.py
```diff
@@ -147,19 +168,25 @@ def _submit_copy_request(
...
-                    'extra_args': call_args.extra_args,
-                    'callbacks': progress_callbacks,
-                    'size': transfer_future.meta.size,
+                    "client": client,
+                    "copy_source": call_args.copy_source,
+                    "bucket": call_args.bucket,
+                    "key": call_args.key,
+                    "extra_args": copy_object_extra_args,
+                    "callbacks": progress_callbacks,
+                    "size": transfer_future.meta.size,
```
aemous, line 189: nit: why did we re-format the ' to " ?

aemous (approved):
LGTM. I made one non-blocking formatting-related comment.

awscli/s3transfer/copies.py
kdaily: Looks like the formatter ran on it?

PR #9594, opened 2025-07-15
[v2] Make inline docs scripts external

aemous:
*Context:*

This change is in preparation for adding a `Content-Security-Policy` header to the AWS CLI API Reference versioned docs. The added header will block the loading of inline-scripts. To avoid breaking inline scripts on older versions of the docs, the header will allowlist these inline scripts. To avoid needing to update the header each time we modify these scripts in the future, we should move them to an external file.

*Description of changes:*

* Replaced use of inline scripts with external script files. 

*Description of tests:*

* Ran `make all` on the docs, and verified the scripts are being loaded.


kdaily (commented):
Looks good. Do we need to include this commit (0bbf4e2a8140dfc03cdc22821ee27d5e3c06b1af)? Looks like the changes it introduced were undone in the later commit.

aemous:
> Looks good. Do we need to include this commit ([0bbf4e2](https://github.com/aws/aws-cli/commit/0bbf4e2a8140dfc03cdc22821ee27d5e3c06b1af))? Looks like the changes it introduced were undone in the later commit.

@kdaily The index.html file is from build, I accidentally included it in the PR, that commit is to remove the unintended file. The end result is as expected (index.html should not be committed).

kdaily (approved):

PR #9609, opened 2025-07-22
Implementing no-overwrite for sync and download

rakshil14-2:
*Issue :*  Github Issue #2874 

*Description of changes:*

Customers requested to prevent overwriting of objects. This is done by providing `no-overwrite` header. This PR contains implementation of `--no-overwrite` flag for 2 operations i.e download and sync operations.

- Download operation 
To prevent overwriting of the file, before initiating the download, file is begin checked at the local system and if the file is absent downloading is allowed else, operation is terminated.
- Sync operation
To implement `--no-overwrite` a new sync strategy has been introduced named to allow transfer of only those files that are not present at the destination. Since files have already been separated before initiating the operation, so again checking for its existence would be redundant and causes extra api calls hence the argument is removed before the transfer begins. This strategy is independent of size and timestamps.

_Testing_:
 Functional testing is performed to validate the object transfer processes. This included both scenarios transferring an object when object with same key is already present in the destination as well as those when the object with same key is not present on destination. These tests were conducted for both object having size less than multipart threshold and one with greater than multipart threshold.


awscli/customizations/s3/s3handler.py
```diff
@@ -438,7 +438,38 @@ def _get_fileout(self, fileinfo):
...
+                False if the download should proceed
+        """
+        if not self._cli_params.get('no_overwrite'):
+            return False
+        fileout = self._get_fileout(fileinfo)
+        if os.path.exists(fileout):
+            LOGGER.debug(
+                "Warning: Skipping file %s as it already exists on %s",
+                fileinfo.src,
+                fileinfo.dest,
```
hssyoo: 1. We should use f-strings for formatting strings: https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals
2. The message itself is misleading because it implies the same file exists at destination, which isn't necessarily true. Let's clarify that _a_ file exists a destination

awscli/customizations/s3/subcommands.py
```diff
@@ -1603,6 +1616,7 @@ def _validate_streaming_paths(self):
             self.parameters['is_stream'] = True
             self.parameters['dir_op'] = False
             self.parameters['only_show_errors'] = True
+            self._validate_no_overwrite_for_download_streaming()
```
hssyoo: I think it would make sense for this method to use `self.parameters['is_stream']` once it's set and then call it right after `_validate_streaming_paths`. So separate the 2 validate methods and make the new one self-contained.

awscli/customizations/s3/subcommands.py
```diff
@@ -1510,6 +1510,19 @@ def _map_sse_c_params(self, request_parameters, paths_type):
                 },
             )
 
+    def _get_s3_handler_params(self):
```
hssyoo, line 1513: We should just take out `NO_OVERWRITE` from `TRANSFER_ARGS` and move it back to `CpCommand` and `MvCommand`. It seems unnecessary to add it to sync only to remove it, and then add it back via the sync strategy class.

awscli/customizations/s3/syncstrategy/nooverwrite.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
hssyoo: `2014` -> `2025`

awscli/customizations/s3/syncstrategy/nooverwrite.py
```diff
@@ -0,0 +1,45 @@
...
+    """Sync strategy that prevents overwriting of existing files at the destination.
+    This strategy will only sync files that don't exist at the destination.
+    If a file exists at both source and destination, it will be skipped regardless
+    of size or modification time differences.
+    """
+
+    ARGUMENT = NO_OVERWRITE
+
+    def determine_should_sync(self, src_file, dest_file):
+        if not dest_file:
```
hssyoo: What does `dest_file` being falsey mean? Is this checking to see if it's `None`?

awscli/customizations/s3/syncstrategy/nooverwrite.py
```diff
@@ -0,0 +1,45 @@
...
+            # Sync if file does not exist at destination
+            return True
+        elif src_file.compare_key != dest_file.compare_key:
+            # Sync if file exists at both source and destination but the
+            # source file's key is different than the destination file's key
+            return True
+        else:
+            # Skip if file exists at both source and destination
+            LOG.debug(
+                "Warning: Skipping file %s as it already exists on %s",
```
hssyoo: nit: Our convention for warning messages is to start with `warning: `

tests/functional/s3/test_cp_command.py
```diff
@@ -550,6 +538,74 @@ def test_no_overwrite_flag_on_copy_when_large_object_does_not_exist_on_target(
         # Verify the IfNoneMatch condition was set in the CompleteMultipartUpload request
         self.assertEqual(self.operations_called[5][1]['IfNoneMatch'], '*')
 
+    def test_no_overwrite_flag_on_download_when_small_object_already_exists_at_target(
```
hssyoo: Instead of small object vs big object, let's make a distinction on single object vs multipart downloads

tests/functional/s3/test_cp_command.py
```diff
@@ -550,6 +538,74 @@ def test_no_overwrite_flag_on_copy_when_large_object_does_not_exist_on_target(
...
+            self.get_object_response(),
+        ]
+        self.run_cmd(cmdline, expected_rc=0)
+        self.assertEqual(len(self.operations_called), 2)
+        self.assertEqual(self.operations_called[0][0].name, 'HeadObject')
+        self.assertEqual(self.operations_called[1][0].name, 'GetObject')
+        with open(full_path) as f:
+            self.assertEqual(f.read(), 'foo')
+
+    def test_no_overwrite_flag_on_download_when_large_object_exists_at_target(
```
hssyoo: I don't think we need this test since it's testing the same code path and behavior as the single object case

tests/functional/s3/test_mv_command.py
```diff
@@ -575,6 +563,78 @@ def test_mv_no_overwrite_flag_when_large_object_does_not_exist_on_target(
...
+        ]
+        self.run_cmd(cmdline, expected_rc=0)
+        self.assertEqual(len(self.operations_called), 3)
+        self.assertEqual(self.operations_called[0][0].name, 'HeadObject')
+        self.assertEqual(self.operations_called[1][0].name, 'GetObject')
+        self.assertEqual(self.operations_called[2][0].name, 'DeleteObject')
+        with open(full_path) as f:
+            self.assertEqual(f.read(), 'foo')
+
+    def test_no_overwrite_flag_on_mv_download_when_large_object_exists_at_target(
```
hssyoo: Same comment here, we can remove

tests/functional/s3/test_sync_command.py
```diff
@@ -540,6 +540,91 @@ def test_download_with_checksum_mode_crc64nvme(self):
             ('ChecksumMode', 'ENABLED'), self.operations_called[1][1].items()
         )
 
+    def test_sync_upload_with_no_overwrite_when_file_does_not_exist_at_destination(
+        self,
+    ):
+        self.files.create_file("new_file.txt", "mycontent")
+        self.parsed_responses = [
+            self.list_objects_response(['file.txt']),
+            self.get_object_response(),
```
hssyoo: Why are we parsing a `GetObject` response?

tests/unit/customizations/s3/syncstrategy/test_nooverwrite.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
hssyoo: `2014` -> `2025`

tests/unit/customizations/s3/syncstrategy/test_nooverwrite.py
```diff
@@ -0,0 +1,110 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import datetime
+
+from awscli.customizations.s3.filegenerator import FileStat
+from awscli.customizations.s3.syncstrategy.nooverwrite import NoOverwriteSync
+from awscli.testutils import unittest
```
hssyoo: For new test modules/classes we use `pytest`, which is more modern and composable

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -679,6 +680,40 @@ def test_submit_move_adds_delete_source_subscriber(self):
         for i, actual_subscriber in enumerate(actual_subscribers):
             self.assertIsInstance(actual_subscriber, ref_subscribers[i])
 
+    def test_skip_download_when_no_overwrite_and_file_exists(self):
+        """Testing warn_if_file_exists_with_no_overwrite handler when file exists at destination"""
```
hssyoo: In general, test names should describe what's being tested. Your test names do that, which makes these docstrings redundant

awscli/customizations/s3/syncstrategy/nooverwrite.py
kdaily: We have guidance that the best thing to do is drop the year, so this can become:

```
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

awscli/customizations/s3/syncstrategy/nooverwrite.py
hssyoo: Spent some more time looking at this. By the time this code is executed, the `Comparator` class [[source](https://github.com/aws/aws-cli/blob/7a3330a6104dcff0e549a632f4495673f7f2358f/awscli/customizations/s3/comparator.py#L20)] will have already compared the existence of source/destination files. Since the no overwrite sync strategy is registered for `file_at_src_and_dest`, this means that `dest_file` is guaranteed to be not `None` and that the file exists at destination. So we should be able to just always return `False` in this case.

awscli/customizations/s3/subcommands.py
hssyoo: Chatted offline, we want to avoid passing `no_overwrite` to `s3s3` transfers since it'll trigger a multipart copy. I'm okay with handling this here. Instead of creating a new dictionary, is it safe for us to just delete `no_overwrite` from `self.parameters`?

awscli/customizations/s3/syncstrategy/register.py
```diff
@@ -48,4 +49,8 @@ def register_sync_strategies(command_table, session, **kwargs):
     # Register the delete sync strategy.
     register_sync_strategy(session, DeleteSync, 'file_not_at_src')
 
+    # Register the noOverwrite sync strategy
+    register_sync_strategy(session, NoOverwriteSync, 'file_not_at_dest')
```
hssyoo: Is there a reason we're not just using the default `MissingFileSync` strategy for `file_not_at_dest`?

awscli/customizations/s3/subcommands.py
```diff
@@ -1824,3 +1836,25 @@ def _validate_sse_c_copy_source_for_paths(self):
...
                 )
+
+    def _validate_no_overwrite_for_download_streaming(self):
+        """
+        Validates that no-overwrite parameter is not used with streaming downloads.
+
+        When downloading from S3 to stdout (streaming download), the no-overwrite
+        parameter doesn't make sense since stdout is not a file that can be checked
+        for existence. This method checks for this invalid combination and raises
+        an appropriate error.
```
hssyoo: On Unix-like systems, stdout is treated as a file-like object and you can check its existence (eg `/dev/stdout`). I think what we're trying to say is that stdout is always going to be there, so it doesn't make sense to support the no overwrite feature

awscli/customizations/s3/syncstrategy/register.py
rakshil14-2: Chatted Offline, If we just use `MissingFileSync` strategy, then only files not present at destination will be transferred and if the file is already present then default sync strategy `SizeAndLastModifiedSync` will be used with syncs the object based on the size and time, so it get overwritten. To address this,  we decided to `override` the default sync strategy.

awscli/customizations/s3/subcommands.py
```diff
@@ -1296,10 +1297,15 @@ def choose_sync_strategies(self):
...
-        )
+        if self.parameters.get('no_overwrite'):
+            sync_strategies['file_at_src_and_dest_sync_strategy'] = (
+                NoOverwriteSync()
+            )
+        else:
+            # Set the default strategies.
+            sync_strategies['file_at_src_and_dest_sync_strategy'] = (
+                SizeAndLastModifiedSync()
+            )
```
hssyoo: We want to avoid resolving the strategy via control flow because it'll get more complex and harder to reason through if we add more strategies in the future. Instead, we should be able to follow the docstrings for this method:
> It defaults to the default sync strategies but a customizable sync strategy can override the default strategy if it returns the instance of its self when the event is emitted.

So for the `NoOverwriteSync` class, we'll want to implement a `use_sync_strategy` method that returns itself if the `no_overwrite` param exists.

hssyoo (changes requested):

awscli/customizations/s3/subcommands.py
hssyoo: Oh wait, it should be even simpler to just set the class variable `ARGUMENT = NO_OVERWRITE`, similar to here: https://github.com/aws/aws-cli/blob/60ebb3c9ccaad0e492af1ca857f3b13ff68241fd/awscli/customizations/s3/syncstrategy/delete.py#L32

rakshil14-2: Yes, declared `NOOVERWRITE` as a class variable, and `default` sync strategy is getting overriden

hssyoo (approved):

Issue #7758, opened 2023-03-18
aws configure import --csv Prompt error, Expected header "User Name" not found


qian-jiahong:
### Describe the bug

 When executing: aws configure import --csv, the wrong file path is entered, and the "error prompt" received is wrong.

### Expected Behavior

$ aws configure import --csv aws_user_credentials.csv
$ aws configure import --csv aws_user_credentials-noexist.csv

It should prompt that the file does not exist.
Instead of: Expected header "User Name" not found

### Current Behavior

$ aws configure import --csv aws_user_credentials.csv
Expected header "User Name" not found

$ aws configure import --csv aws_user_credentials-noexist.csv
Expected header "User Name" not found

$ aws configure import --csv file://./aws_user_credentials.csv
Successfully imported 1 profile(s)

$ aws configure import --csv file://./aws_user_credentials-noexist.csv
Error parsing parameter '--csv': Unable to load paramfile file://./aws_user_credentials-noexist.csv: [Errno 2] No such file or directory: './aws_user_credentials-noexist.csv'

### Reproduction Steps

Download the certificate file (.csv) from the aws website to the current directory. Named as: aws_user_credentials.csv
$ aws configure import --csv aws_user_credentials.csv
$ aws configure import --csv aws_user_credentials-noexist.csv

### CLI version used

aws-cli/2.11.3 Python/3.11.2 Linux/5.15.0-56-generic exe/x86_64.ubuntu.20 prompt/off

### Environment details (OS name and version, etc.)

Linux ubuntu20-s1 5.15.0-56-generic #62~20.04.1-Ubuntu SMP Tue Nov 22 21:24:20 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

tim-finnigan:
Hi @qian-jiahong thanks for reaching out. This is an error that I can reproduce. The [documentation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/configure/import.html#examples) for `aws configure import` does use `file://` as a prefix, but regardless the `Expected header "User Name" not found` is not a helpful error message.

[...]

PR #9617, opened 2025-07-28
Fix unclear configure import error

AndrewAsseily:
*Issue #, if available:* CLI-4867

*Description of changes:*
This enhancement addresses the issue reported in https://github.com/aws/aws-cli/issues/7758, where users were encountering unclear error messages when attempting to import CSV credentials.
This change improves the documentation and error messaging for the `aws configure import --csv` 
command when the file:// prefix is missing from file paths.

*Testing:*
1. Ran Unit tests
2. Built and reviewed documentation changes
3. Manually tested scenarios:
   1. Missing file:// prefix:
      ```
      ./bin/aws configure import --csv test-creds.csv
      ```
      Result: Shows error message
      > You may be passing a file to import without the 'file://' prefix. To import a CSV file, use --csv file://path/to/file.csv

   2. Correct file path format:
      ```
      ./bin/aws configure import --csv file://test-creds.csv
      ```
      Result: Shows success message
      > Successfully imported 1 profile(s)

   3. Inline CSV data:
      ```
      ./bin/aws configure import --csv "User Name,Access Key ID,Secret Access Key
      my-user,*************,inlinekeyvalue"
      ```
      Result: Shows success message
      > Successfully imported 1 profile(s)

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute 
this contribution, under the terms of your choice.*


awscli/customizations/configure/importer.py
```diff
@@ -18,6 +18,9 @@
 from awscli.customizations.configure.writer import ConfigFileWriter
 from awscli.customizations.utils import uni_print
 
+class ConfigureImportError(Exception):
```
kdaily: nit: looks like we have previously used a `RuntimeError` for file/directory existence issues. For example:

https://github.com/aws/aws-cli/blob/d4fa0daaf592e9fea2ca9615010d1423c0c03d65/awscli/customizations/s3/subcommands.py#L1614-L1617

nateprewitt: `ValueError` is probably more correct if we're discussing an input issue. The question that comes to mind here though is if `file://` is a required prefix, why are we not just appending it if it doesn't exist?

kdaily: > `ValueError` is probably more correct if we're discussing an input issue.

There does seem to be precedence for that too, SGTM.

https://github.com/aws/aws-cli/blob/d4fa0daaf592e9fea2ca9615010d1423c0c03d65/awscli/customizations/cloudformation/artifact_exporter.py#L634-L638

AndrewAsseily: I've updated the code to use ValueError instead of a custom exception. As discussed offline, we've decided to keep the file:// prefix requirement to maintain consistency with our existing CLI patterns and avoid potential ambiguity.

.changes/next-release/enhancement-Configure-43901.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Configure",
```
kdaily: This is a nit, but for consistency with existing changelog messages this should be a lowercase `configure`, like these:

- https://github.com/aws/aws-cli/blob/a95b915dd21b39862930d5c4d21c9caf63e45079/CHANGELOG.rst#L1997
- https://github.com/aws/aws-cli/blob/a95b915dd21b39862930d5c4d21c9caf63e45079/CHANGELOG.rst#L8983

```suggestion
  "category": "configure",
```

tests/unit/customizations/configure/test_importer.py
```diff
@@ -93,6 +93,31 @@ def test_import_downloaded_bad_headers(self):
         with self.assertRaises(CredentialParserError):
             self.import_command(args=['--csv', content], parsed_globals=None)
 
+    def test_raises_error_when_plain_file_path_passed(self):
+        with open('temp_creds.csv', 'w') as f:
+            f.write('User name,Access key ID,Secret access key\nuser,AKID,SAK')
+        try:
+            with self.assertRaises(ValueError) as cm:
+                self.import_command(args=['--csv', 'temp_creds.csv'], parsed_globals=None)
+            self.assertIn("file://", str(cm.exception))
```
kdaily: I think it would be a bit safer to assert that some other unique part of the error message is present because `file://` does occur twice in the message.

kdaily (changes requested):
Only a couple of minor changes, but this change should improve customer experience in adding creds from CSV files!

PR #9621, opened 2025-07-29
[v2] Add completion model for dsql

hssyoo:
Adds a server-side completion model for `dsql`.

### DIFF
--- a/awscli/botocore/data/dsql/2018-05-10/completions-1.json
+++ b/awscli/botocore/data/dsql/2018-05-10/completions-1.json
@@ -1,0 +1,92 @@
+{
+  "version": "1.0",
+  "resources": {
+    "Cluster": {
+      "operation": "ListClusters",
+      "inputParameters": null,
+      "resourceIdentifier": {
+        "identifier": "clusters[].identifier",
+        "arn": "clusters[].arn"
+      }
+    }
+  },
+  "operations": {
+    "DeleteCluster": {
+      "identifier": {
+        "completions": [
+          {
+            "resourceName": "Cluster",
+            "resourceIdentifier": "identifier",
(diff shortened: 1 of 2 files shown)


awscli/botocore/data/dsql/2018-05-10/completions-1.json
```diff
@@ -0,0 +1,92 @@
+{
+  "version": "1.0",
+  "resources": {
+    "Cluster": {
+      "operation": "ListClusters",
+      "inputParameters": null,
```
hssyoo: `inputParameters` is modeled in the completer generator, though it's not currently used/generated, which is why value in `null`. Going to manually remove it for now and fix this in code.

awscli/botocore/data/dsql/2018-05-10/completions-1.json
```diff
@@ -0,0 +1,92 @@
...
+      "resourceArn": {
+        "completions": [
+          {
+            "resourceName": "Cluster",
+            "resourceIdentifier": "arn",
+            "parameters": {}
+          }
+        ]
+      }
+    },
```
hssyoo: LLM is a bit too eager with what operations/parameters should be completable. Going to manually remove for now, but need to massage prompts to avoid models that don't really make sense.

awscli/botocore/data/dsql/2018-05-10/completions-1.json
```diff
@@ -0,0 +1,69 @@
...
```
hssyoo: I'm also going to remove this because there may be other modeled resources in the future that won't be included. It would be interesting if we could group resources together to reference in `operations` so that an operation could target multiple resources.

PR #9622, opened 2025-07-29
Update documentation for default retry strategy change

kdaily:
*Issue #, if available:*

https://github.com/boto/botocore/pull/3513/

*Description of changes:*

Updates the configuration options topic guide for retries to indicate the default is changing from `legacy` to `standard`.

The [AWS CLI user guide](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-retries.html) will be updated separately to reflect the change as well.  

### DIFF
--- a/awscli/topics/config-vars.rst
+++ b/awscli/topics/config-vars.rst
@@ -487,13 +487,13 @@ These configuration variables control how the AWS CLI retries requests.
     A string representing the type of retries the AWS CLI will perform.  Value
     values are:
 
-        * ``legacy`` - The pre-existing retry behavior.  This is default value if
-          no retry mode is provided.
         * ``standard`` - A standardized set of retry rules across the AWS SDKs.
           This includes a standard set of errors that are retried as well as
           support for retry quotas, which limit the number of unsuccessful retries
           an SDK can make.  This mode will default the maximum number of attempts
-          to 3 unless a ``max_attempts`` is explicitly provided.
+          to 3 unless a ``max_attempts`` is explicitly provided. This is default
+          value if no retry mode is provided.
+        * ``legacy`` - The pre-existing retry behavior.
         * ``adaptive`` - An experimental retry mode that includes all the
           functionality of ``standard`` mode along with automatic client side
           throttling.  This is a provisional mode that may change behavior in
--- a/awscli/topics/config-vars.rst
+++ b/awscli/topics/config-vars.rst
@@ -493,12 +493,15 @@ These configuration variables control how the AWS CLI retries requests.
           an SDK can make.  This mode will default the maximum number of attempts
           to 3 unless a ``max_attempts`` is explicitly provided. This is default
           value if no retry mode is provided.
-        * ``legacy`` - The pre-existing retry behavior.
+        * ``legacy`` - The pre-existing retry behavior. This mode will default the
+          maximum number of attempts to 5.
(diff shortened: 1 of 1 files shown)


ashovlin:
Could we include a changelog entry too? Even though the change is from botocore, I think it's important enough to call out.

nateprewitt:
@ashovlin That's actually part of the release process. Everything from botocore gets propogated up in [these commits](https://github.com/aws/aws-cli/commit/ce177f7a40d3785aa8dfad3d6ddb5c2d862f74ee) to the CLI. We used to specifically tag them as botocore changes, but it looks like the current release system stopped doing that in ~1.23.0. Maybe a question for @hssyoo if that wasn't intentional.

The botocore release will push up both of these:
* [STS Changelog](https://github.com/boto/botocore/pull/3517/files#diff-b66d78410c5f68e23eefdd6b33a78d18a79cf279a01f2d326d8f00ebfb0f4550R1-R5)
* [Retries Changelog](https://github.com/boto/botocore/pull/3513/files#diff-465fb078b20857b637213f3da97e5e6e9555ab0926d6b7786bfff6532c5d06a1R1-R5)

ashovlin:
> @ashovlin That's actually part of the release process. Everything from botocore gets propogated up in [these commits](https://github.com/aws/aws-cli/commit/ce177f7a40d3785aa8dfad3d6ddb5c2d862f74ee) to the CLI. We used to specifically tag them as botocore changes, but it looks like the current release system stopped doing that in ~1.23.0. Maybe a question for @hssyoo if that wasn't intentional.
> 

Does that include changes that aren't `api-change`?

Botocore 1.39.15 had the SSO fix: https://github.com/boto/botocore/blob/d71964b8cb377e358f3dd36f30d9a55880d1ae18/CHANGELOG.rst#L21

But I don't see that in the corresponding CLI 1.41.15: https://github.com/aws/aws-cli/blob/4b050a30d0246546da811bd2408e463bfd889e36/CHANGELOG.rst#L15-L20

_**EDIT: Clarified internally, this one is a `feature` so it should be propagated. `bugfix` or `enhancement` are not currently.**_

ubaskota (commented):
Even though it's mentioned under `max_attempts`, can we also include the number of attempts it makes in `legacy` and `standard` retry modes? Something along the lines of 'This retry mode defaults to 4 retries (5 max attempts).' This would be helpful to our users.

kdaily:
@ubaskota I would be more inclined to link to our user guide that details the behavior, specifics, and caveats of each strategy:

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes

What do you think of that?

kdaily:
> Even though it's mentioned under `max_attempts`, can we also include the number of attempts it makes in `legacy` and `standard` retry modes? Something along the lines of 'This retry mode defaults to 4 retries (5 max attempts).' This would be helpful to our users.

Addressed in 93321e0dcce1415376ccf25c9d9faa65d7d343a5.

ubaskota:
> @ubaskota I would be more inclined to link to our user guide that details the behavior, specifics, and caveats of each strategy:
> 
> https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes
> 
> What do you think of that?

Yeah, that works too.

ubaskota (approved):

PR #9628, opened 2025-07-30
Update sts endpoint assertions

kdaily:
*Issue #, if available:*

*Description of changes:*

The default behavior for which STS endpoint to use is changing from global to regional.

These tests previously assumed that the global endpoint will always be used, which is will not be the case for the default behavior from the botocore client once boto/botocore#3517 is merged. This change updates the tests to reflect that the regional endpoint will be used.

### DIFF
--- a/tests/functional/eks/test_get_token.py
+++ b/tests/functional/eks/test_get_token.py
@@ -140,18 +140,23 @@ class TestGetTokenCommand(BaseAWSCommandParamsTest):
     def test_url(self):
         cmd = 'eks get-token --cluster-name %s' % self.cluster_name
         response = self.run_get_token(cmd)
-        self.assert_url_correct(response)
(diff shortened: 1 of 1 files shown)


ashovlin (commented):
Given the CI failures here, it might be better to explicitly set `legacy` or `regional` for the STS setting for each test case. That way we're not reliant on the botocore version or environment settings.

kdaily:
> Given the CI failures here, it might be better to explicitly set `legacy` or `regional` for the STS setting for each test case. That way we're not reliant on the botocore version or environment settings.

We agreed separately that we should have a test for STS defaults, but would follow up separately. To change this test we would need to mock the session or set an environment variable, since the CLI does not have a command line parameter to set `sts_regional_endpoint`.

jonathan343 (approved):

PR #9630, opened 2025-07-31
right padding added

senivasan294:
*Issue #, if available:*

*Description of changes:*


kdaily:
Hi @senivasan294, thanks for the contribution, but I don't think this is required at the time. It also seems that the right padding added isn't accessed or used. Closing this out.

senivasan294:
@kdaily ok thanks for update

Issue #4070, opened 2019-04-11
aws s3 rm flag to remove all versions


adamchainz:
The S3 console has the 'empty' button which deletes all objects and versions thereof to clear a bucket before removal. This is very useful for cleaning up temporary buckets.

`aws s3 rm --recursive s3://mybucket` doesn't work the same as "empty the versions, and there isn't a flag to remove all versions, so it's hard to automate complete S3 bucket removal.

Perhaps a flag `--all-versions` to `rm` could be added, to allow this?

justnance:
@adamchainz - Thanks for your post.  Looks like this would be a great feature and _similar to_ another feature request we are tracking under #1017.  I'm going to keep you issue open pending further review and feedback.

bsmith1310:
+1 to this feature request. Having an easy way to (empty and) delete a bucket would be great.

Comments on [#1017](https://github.com/aws/aws-cli/issues/1017) suggest that simply disabling versioning on the bucket prevents the issue, but from what I'm seeing, that's not true. I have buckets that were never versioned, but `aws s3 rm --recursive s3://mybucket` doesn't empty the bucket, the files remain in their 'soft' deleted state.

acdha:
This (and #1017) is becoming more commonly encountered as tools based on the AWS or CIS security standards encourage enabling versioning for all buckets. It'd be nice if the AWS CLI could be used for normal operational practices without having a special case for this configuration.

PR #9637, opened 2025-08-03
Deleting all versions of objects

rakshil14-2:
*Issue* :https://github.com/aws/aws-cli/issues/4070

Description of changes:
Added a flag for deleting all versions of an object present in a versioned-enable bucket. For dealing with versions, a `list-object-versions` api is called to get the `version ids` as well as `delete markers` of the objects. Once the object and their version ids are retrieved, then a batch of objects with their version id is created which is passed to `delete-objects` api for deletion. 

### DIFF
--- a/.changes/next-release/enhancement-s3-34855.json
+++ b/.changes/next-release/enhancement-s3-34855.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "`s3`",
+  "description": "Add all-versions flag with AWS CLI S3 rm command to delete all versions of objects present in a versioned-enabled bucket. fixes `#4070 <https://github.com/aws/aws-cli/issues/4070>`__"
+}
--- a/awscli/customizations/s3/filegenerator.py
+++ b/awscli/customizations/s3/filegenerator.py
@@ -19,6 +19,7 @@ from dateutil.parser import parse
 from dateutil.tz import tzlocal
 
 from awscli.compat import queue
+from awscli.customizations.s3.fileinfo import VersionedFileInfo
 from awscli.customizations.s3.utils import (
     EPOCH_TIME,
     BucketLister,
@@ -26,6 +27,7 @@ from awscli.customizations.s3.utils import (
     find_bucket_key,
     find_dest_path_comp_key,
     get_file_stat,
+    split_s3_bucket_key,
 )
 
 _open = open
@@ -406,3 +408,109 @@ class FileGenerator:
         response['LastModified'] = last_update.astimezone(tzlocal())
         response['ETag'] = response.pop('ETag', None)
         return s3_path, response
+
+
+class VersionedFileGenerator:
+    """
+    This class generates VersionedFileInfo objects for all versions of objects in a bucket.
+    It uses the BucketLister class to list all versions and creates appropriate
+    VersionedFileInfo objects for each version.
+    """
+
+    def __init__(
+        self,
+        client,
+        operation_name,
+        follow_symlinks=True,
+        page_size=None,
+        result_queue=None,
+        request_parameters=None,
+    ):
+        """
+        Initialize a new VersionedFileGenerator.
+
+        :param client: The S3 client to use.
+        :param operation_name: The name of the operation to perform.
+        :param follow_symlinks: Whether to follow symlinks.
+        :param page_size: The number of items to include in each API response.
+        :param result_queue: Queue for results and warnings.
+        :param request_parameters: Additional parameters for the request.
+        """
+        self._client = client
+        self.operation_name = operation_name
+        self.follow_symlinks = follow_symlinks
+        self.page_size = page_size
+        self.result_queue = result_queue
+        if not result_queue:
+            self.result_queue = queue.Queue()
+        self.request_parameters = {}
+        if request_parameters is not None:
+            self.request_parameters = request_parameters
+        self._version_lister = BucketLister(client)
+
+    def call(self, files):
+        """
+        Generate VersionedFileInfo objects for all versions of objects.
+
+        :param files: Dictionary containing source and destination information.
+        :yields: VersionedFileInfo objects for each version of each object.
+        """
+        source = files['src']['path']
+        src_type = files['src']['type']
+        dest_type = files['dest']['type']
+
+        # Use the list_object_versions method to get all versions
+        file_iterator = self.list_object_versions(source, files['dir_op'])
+
+        for src_path, content, version_id in file_iterator:
+            dest_path, compare_key = find_dest_path_comp_key(files, src_path)
+
+            # Create a VersionedFileInfo for this object version
+            yield VersionedFileInfo(
+                src=src_path,
+                dest=dest_path,
+                compare_key=compare_key,
+                size=content.get('Size', 0),
+                last_update=content.get('LastModified'),
+                src_type=src_type,
+                dest_type=dest_type,
+                operation_name=self.operation_name,
+                associated_response_data=content,
+                version_id=version_id,
+            )
+
+    def list_object_versions(self, s3_path, dir_op):
+        """
+        This function yields the appropriate object versions or all object versions
+        under a common prefix depending if the operation is on objects under a
+        common prefix. It yields the file's source path, content, and version ID.
+
+        :param s3_path: The S3 path to list versions for.
+        :param dir_op: Whether this is a directory operation.
+        :yields: Tuples of (source_path, content, version_id)
+        """
+        bucket, key = split_s3_bucket_key(s3_path)
+
+        # Short circuit path: if we are not recursing into the s3
+        # bucket and a specific path was given, we can just yield
+        # that path and not have to call any operation in s3.
+        # However, for versioned objects, we still need to list all versions
+        # even for a specific object, so we don't have a short circuit path here.
+
+        # List all versions of objects
+        for (
+            src_path,
+            content,
+            version_id,
+        ) in self._version_lister.list_object_versions(
+            bucket=bucket,
+            prefix=key,
+            page_size=self.page_size,
+            extra_args=self.request_parameters.get('ListObjectVersions', {}),
+        ):
+            # If this is not a directory operation and the path doesn't match exactly,
+            # skip it (similar to the behavior in FileGenerator.list_objects)
+            if not dir_op and s3_path != src_path:
+                continue
+
+            yield src_path, content, version_id
--- a/awscli/customizations/s3/fileinfo.py
+++ b/awscli/customizations/s3/fileinfo.py
@@ -109,3 +109,37 @@ class FileInfo:
         # restored back to S3.
         # 'Restore' looks like: 'ongoing-request="false", expiry-date="..."'
         return 'ongoing-request="false"' in response_data.get('Restore', '')
+
+
+class VersionedFileInfo(FileInfo):
+    def __init__(self, version_id=None, is_delete_marker=False, **kwargs):
+        """
+        This class extends FileInfo to include version information for S3 objects.
+        It is specifically designed for operations that need to work with versioned
+        S3 objects, such as deleting all versions of an object.
+
+        :param version_id: The version ID of the S3 object.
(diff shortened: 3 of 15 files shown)


awscli/customizations/s3/utils.py
```diff
@@ -432,6 +432,74 @@ def list_objects(
...
+    def __init__(self, client, date_parser=_date_parser):
+        """
+        Initialize a new BucketVersionLister.
+        :param client: The S3 client to use for listing versions.
+        :param date_parser: A function to parse date strings into datetime objects.
+        """
+        self._client = client
+        self._date_parser = date_parser
+
+    def list_object_versions(
```
aemous, line 434: Could we add this as a function in `BucketLister`?

awscli/customizations/s3/utils.py
```diff
@@ -432,6 +432,74 @@ def list_objects(
...
+        pages = paginator.paginate(**kwargs)
+        for page in pages:
+            # Process versions
+            versions = page.get('Versions', [])
+            for version in versions:
+                source_path = bucket + '/' + version['Key']
+                version['LastModified'] = self._date_parser(
+                    version['LastModified']
+                )
+                version['DeleteMarker'] = False
```
aemous: I'm not a big fan of adding a key that says whether it's a delete marker. Could we have 2 separate functions in this class? list versions, and list delete markers? 

Or, maybe the caller can just check if it has an ETag? I think ETags are always on versions, but not on deletemarkers. Worth double-checking this, or using a different attribute in this way.

awscli/customizations/s3/subcommands.py
```diff
@@ -1365,8 +1385,14 @@ def run(self):
         self._map_request_payer_params(rgen_request_parameters)
         rgen_kwargs['request_parameters'] = rgen_request_parameters
 
-        file_generator = FileGenerator(**fgen_kwargs)
-        rev_generator = FileGenerator(**rgen_kwargs)
+        if self.parameters.get('all_versions'):
```
aemous: For defensive programming purposes, could we make this `if self.cmd == 'rm' and self.parameters.get('all_versions')`?

awscli/customizations/s3/s3handler.py
```diff
@@ -592,3 +608,260 @@ def _submit_transfer_request(self, fileinfo, extra_args, subscribers):
...
+        :return: List of batches in S3 delete-objects format
+        """
+        if not fileinfos:
+            return []
+        bucket = None
+        objects_with_fileinfo = []
+        for fileinfo in fileinfos:
+            bucket, key = split_s3_bucket_key(fileinfo.src)
+            obj_entry = {'Key': key}
+            if hasattr(fileinfo, 'version_id') and fileinfo.version_id:
```
aemous: What's the purpose of this check? `if hasattr(fileinfo, 'version_id') and fileinfo.version_id`

It seems you're trying to check "if the object has version_id and its value is not None"? 

1. Are there cases where a VersionedFileInfo would not have a version_id? 
2. Are there cases where the version_id is None?

awscli/customizations/s3/fileinfo.py
```diff
@@ -109,3 +109,40 @@ def _is_restored(self, response_data):
...
+        :param version_id: The version ID of the S3 object.
+        :type version_id: string
+        :param is_delete_marker: Whether this version is a delete marker.
+        :type is_delete_marker: boolean
+        """
+        super().__init__(**kwargs)
+        self.version_id = version_id
+        self.is_delete_marker = is_delete_marker
+
+    def is_glacier_compatible(self):
```
aemous, line 127: Have you tested this end-to-end against S3 glacier?

awscli/customizations/s3/s3handler.py
```diff
@@ -592,3 +608,260 @@ def _submit_transfer_request(self, fileinfo, extra_args, subscribers):
...
+                # Report success
+                for fileinfo in fileinfos_in_batch:
+                    self._report_sucess(fileinfo)
+
+            except Exception as e:
+                for fileinfo in fileinfos_in_batch:
+                    self._report_failure(fileinfo, e)
+        return total_objects
+
+    def _report_sucess(self, fileinfo):
```
aemous: nit: typo. sucess -> success

awscli/customizations/s3/filegenerator.py
```diff
@@ -406,3 +409,114 @@ def _list_single_object(self, s3_path):
...
+            yield VersionedFileInfo(
+                src=src_path,
+                dest=dest_path,
+                compare_key=compare_key,
+                size=content.get('Size', 0),
+                last_update=content.get('LastModified'),
+                src_type=src_type,
+                dest_type=dest_type,
+                operation_name=self.operation_name,
+                # client=self._client,
```
aemous: Let's remove this comment if it's not needed

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -1253,3 +1267,202 @@ def test_dry_run(self):
...
+            self.result_queue,
+            [self.result_recorder, self.processed_results.append],
+        )
+        self.command_result_recorder = CommandResultRecorder(
+            self.result_queue, self.result_recorder, self.result_processor
+        )
+
+        self.transfer_manager = mock.Mock(spec=TransferManager)
+        self.transfer_manager.__enter__ = mock.Mock()
+        self.transfer_manager.__exit__ = mock.Mock()
```
aemous, line 1288: Why's this needed?

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -1253,3 +1267,202 @@ def test_dry_run(self):
...
+                operation_name='delete',
+                src_type='s3',
+            )
+            fileinfo.version_id = f'version{i}'
+            fileinfos.append(fileinfo)
+
+        mock_submitter = mock.Mock()
+        mock_submitter.can_submit.return_value = True
+        mock_submitter.submit_batch.return_value = num_transfers
+        self.batch_s3_handler._batch_submitters = [mock_submitter]
```
aemous: Can we find a better way to do this? It's generally not good testing practice to access to the private members of a class (the members starting with `_`. Can either modify the batch s3 handler to accept submitters as input.

What do we do for our existing s3 handler ? Do we modify the private field of submitters?

tests/unit/customizations/s3/test_s3handler.py
```diff
@@ -1253,3 +1267,202 @@ def test_dry_run(self):
...
+
+        self.batch_s3_handler.call([fileinfo])
+        mock_submitter.can_submit.assert_called_once_with(fileinfo)
+        mock_submitter.submit_batch.assert_called_once()
+
+
+class TestDeleteBatch:
+    @pytest.fixture(autouse=True)
+    def setUp(self):
```
aemous: If there's only one test case, let's just move the setup code into the test case. setUp is only needed when its shared across multiple test cases in the suite

aemous (changes requested):
Let's also generate a changelog entry. 

looks good overall, a few questions/requests

awscli/customizations/s3/s3handler.py
rakshil14-2: Yes, when the bucket versioning is suspended, the newly transferred objects will have `null` as versionId so we need to handle them

tests/unit/customizations/s3/test_s3handler.py
rakshil14-2: Chatted online: Resolved it

awscli/customizations/s3/fileinfo.py
rakshil14-2: Discussed with the team : The issue is about object locking, so currently decided to display the error to the customer

awscli/customizations/s3/filegenerator.py
```diff
@@ -406,3 +408,109 @@ def _list_single_object(self, s3_path):
         response['LastModified'] = last_update.astimezone(tzlocal())
         response['ETag'] = response.pop('ETag', None)
         return s3_path, response
+
+
+class VersionedFileGenerator:
```
kdaily, line 413: Discussed a bit offline, but one question I had is if a `VersionedFileGenerator` can be a special case of a `FileGenerator` - can it be used in expected downstream places, like an `S3Handler` or `Comparator` as is noted for the `FileGenerator`/`FileInfo` objects? If so, it might be helpful to mirror the implementation structure so future devs can easily learn about the two classes, or even maybe subclass so that downstream uses can swap out for a `VersionedFileGenerator`?

kdaily: One reason I ask is because the implementation subclasses `VersionedFileInfo` from `FileInfo`.

rakshil14-2: Discussed offline: Decided to keep it separate as it is VersionedFileInfo is related to S3 versioned bucket and it can be extended in future. Also decided to keep two separate handlers one handling the current functionality of all the operation and other for dealing with batch operation.

.changes/next-release/enhancement-s3-34855.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "s3",
```
aemous: non-blocking nit: \``s3`` for fomatting.

aemous (approved):
LGTM

awscli/customizations/s3/s3handler.py
```diff
@@ -89,7 +96,10 @@ def __call__(self, transfer_manager, result_queue):
         command_result_recorder = CommandResultRecorder(
             result_queue, result_recorder, result_processor
         )
-
+        if self._requires_batch_handler():
```
kdaily, line 99: I'm not 100% sure about adding logic to change the type returned. In the function signature it says:

```
        :returns: A S3TransferHandler instance
```

Does a `BatchS3TransferHandler` act like the existing `S3TransferHandler`?

rakshil14-2: Yes, but `BatchS3TransferHandler` will handler batching operations

awscli/customizations/s3/s3handler.py
```diff
@@ -592,3 +609,284 @@ def _submit_transfer_request(self, fileinfo, extra_args, subscribers):
...
+        self._cli_params = cli_params
+        self._result_command_recorder = result_command_recorder
+
+        submitter_args = (
+            self._transfer_manager,
+            self._result_command_recorder.result_queue,
+            cli_params,
+        )
+        # Submitter list for batch request submitters
+        self._batch_submitters = [
```
kdaily, line 720: Could this be:

```suggestion
        self._submitters = [
```

to follow the same interface as the `S3TransferHandler`?

awscli/customizations/s3/s3handler.py
```diff
@@ -592,3 +609,284 @@ def _submit_transfer_request(self, fileinfo, extra_args, subscribers):
...
+            BatchDeleteRequestSubmitter(*submitter_args),
+        ]
+
+    def call(self, fileinfos):
+        """
+        Process iterable of VersionedFileInfo for batch transfers
+
+        :type versionedFileInfo: Set of VersionedFileInfo to submit to batch submitter
+        :param fileinfos: The fileinfos to submit to the batch submitter
```
kdaily, line 729: I think keeping the same structure as the existing handler would be preferrable to describe, since the only underlying changes are the type of iterable and the submitters being called on them.

```suggestion
        :type fileinfos: iterable of VersionedFileInfos
        param fileinfos: Set of VersionedFileInfos to submit to underlying transfer
            request submitters to make transfer API calls to S3
```

.changes/next-release/enhancement-s3-34855.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "`s3`",
```
aemous, line 3: nit: I think this should have two ticks on both sides. \``s3``

kdaily (changes requested):

PR #9647, opened 2025-08-10
Implement MCP server

tonyxwz:
This is an MCP server that can execute any AWS CLI command supported by the current user's installed aws cli.

This is more of a feature request than a PR. The implementation is far from prod ready, just a demo to show that what are possible.

*Issue #, if available:* No issues. Just an idea that I feel worthy of exploring by the aws cli devs. Currently there are too many AWS MCP servers in the `awslabs/mcp` repository, and to be honest, almost none of them are very useful.

I think the AWS CLI itself should be the place where AWS MCP server is implemented (at least the local version).

After all, the github repository description says:
>  Universal Command Line Interface for Amazon Web Services 

*Description of changes:*

Just add one command

```sh
aws mcp-server start
```
or

```sh
aws mcp-server start --transport streamable-http
```

Result (testing with `npx @modelcontextprotocol/inspector`):

Screenshot 2025-08-10 at 17 26 29

You need to install the mcp python SDK to test this PR:
```
pip install mcp
```

Check your MCP client such as <https://code.visualstudio.com/docs/copilot/chat/mcp-servers> to add the server to your coding assistant.

### DIFF
--- a/awscli/customizations/mcp/_mcp_server.py
+++ b/awscli/customizations/mcp/_mcp_server.py
@@ -54,16 +54,16 @@ def run_aws_cli(ctx: Context, aws_cli: str) -> dict:
     sub_command = tokens[1:]
     LOG.debug(f"Running AWS CLI command {sub_command!r}")
     try:
-        output = StringIO()
-        _patch_aws_cli_pager(output)
-        with redirect_stdout(output):
+        buf = StringIO()
+        _patch_aws_cli_pager(buf)
+        with redirect_stdout(buf):
             clidriver = create_clidriver(sub_command)
             return_code = clidriver.main(sub_command)
+            output = buf.getvalue()
             try:
-                parsed_json = json.load(output)
-                return parsed_json
+                return json.loads(output)
             except json.JSONDecodeError:
-                return {"output": output.getvalue()}
+                return {"output": output}
 
     except SystemExit as sys_exit:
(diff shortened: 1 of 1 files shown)


ashovlin:
Thanks for kicking off this discussion @tonyxwz! I think it does make sense as a feature request if you want to create one in paralle to thisl, that might be more visible to the community for collecting feedback and 👍 s.

The closest one in `awslabs/mcp` is probably `aws-api-mcp-server`, which is built on top of the CLI:
* https://awslabs.github.io/mcp/servers/aws-api-mcp-server/
* https://github.com/awslabs/mcp/tree/main/src/aws-api-mcp-server

Curious if you've tried that one?

tonyxwz:
> that might be more visible to the community for collecting feedback and 👍 s.

@ashovlin Thanks for your response. Should I create a feature request issue instead?

I tried them, but I feel `aws` cli is a better place to expose arbitrary aws command knowledge to my LLM assistant because I do not want to instead another command line program for aws which relies on a different version of awscli ([v1 instead of v2](https://github.com/awslabs/mcp/blob/29efc4aa57178b9765ee070b72dc95d758ebe38c/src/aws-api-mcp-server/pyproject.toml#L13)). All the v2 commands do not work with aws api mcp server, yet almost all the new knowledge of LLM is for v2.

PR #9655, opened 2025-08-15
Improve User Experience in Linux and Windows

rakshil14-2:
*Description of changes:*
In Current Implementation all the events(Success, Failure, Warning) are handled by result queue. Where every file_transferred is accounted for any of the three events. In case of `--include` and `--exclude`, the files which are supposed to be filtered out are done before initiating the actual transfer hence, they is no issue in those cases.

While in case of `no-overwrite` transfer actually happens so for the files which are skipped a new result queue must be added and those files that are skipped must be properly handled. Here is the example how we can deal with the solution.


kdaily:
Fixed via https://github.com/aws/aws-cli/pull/9717

PR #9678, opened 2025-08-27
[v2] Expose io_chunksize S3 transfer config

hssyoo:
This PR exposes an existing S3 transfer config `io_chunksize` so that it's configurable via the shared config file.

Copied from doc update in the PR:
> When a GET request is called for downloads, the response contains a file-like
object that streams data fetched from S3. Chunks are read from the stream and
queued in-memory for writes. ``io_chunksize`` configures the maximum size of
elements in the IO queue. This value can be specified using the same semantics
as ``multipart_threshold``, that is either as the number of bytes as an
integer, or using a size suffix.
Increasing this value may result in higher overall throughput by preventing
blocking in cases where large objects are downloaded in environments where
network speed exceeds disk write speed.

### DIFF
--- a/awscli/customizations/s3/transferconfig.py
+++ b/awscli/customizations/s3/transferconfig.py
@@ -30,6 +30,7 @@ DEFAULTS = {
     'max_bandwidth': None,
     'preferred_transfer_client': constants.AUTO_RESOLVE_TRANSFER_CLIENT,
     'target_bandwidth': None,
+    'io_chunksize': 256 * 1024,
 }
 
 


kdaily:
Question: should this get added to the `TransferConfig` class as well?

https://github.com/aws/aws-cli/blob/efc11546c9a7f744dd0a5ccbdb7dcf2eaf60f433/awscli/s3transfer/__init__.py#L667

I ask because I don't think sending it through `kwargs` is going to persist it on the object created here:

https://github.com/aws/aws-cli/pull/9678/files#diff-280c24f20aa46551fbfd53f1910ece8ac396fae5de0956cf3555015592d45161R217

awscli/topics/s3-config.rst
```diff
@@ -208,6 +210,26 @@ threads having to wait unnecessarily which can lead to excess resource
 consumption and connection timeouts.
 
 
+io_chunksize
+------------
+
+**Default** - ``256KB``
```
kdaily, line 216: How is the default set? I see we have some defaults configured as well, should it be added there?
https://github.com/aws/aws-cli/pull/9678/files#diff-280c24f20aa46551fbfd53f1910ece8ac396fae5de0956cf3555015592d45161L25

hssyoo: Default is set here: https://github.com/aws/aws-cli/blob/b9b9e44e44d79378d695e11db3e44f5e87ea73dc/awscli/s3transfer/manager.py#L63

We don't need to add it there since it's an optional param with a default value

hssyoo:
> Question: should this get added to the `TransferConfig` class as well?
> 
> https://github.com/aws/aws-cli/blob/efc11546c9a7f744dd0a5ccbdb7dcf2eaf60f433/awscli/s3transfer/__init__.py#L667
> 
> I ask because I don't think sending it through `kwargs` is going to persist it on the object created here:
> 
> https://github.com/aws/aws-cli/pull/9678/files#diff-280c24f20aa46551fbfd53f1910ece8ac396fae5de0956cf3555015592d45161R217

The referenced code is effectively dead. It's only used for a GameLift customization [[source](https://github.com/aws/aws-cli/blob/b9b9e44e44d79378d695e11db3e44f5e87ea73dc/awscli/customizations/gamelift/uploadbuild.py#L106)]. The `TransferConfig` object used everywhere else is here: https://github.com/aws/aws-cli/blob/b9b9e44e44d79378d695e11db3e44f5e87ea73dc/awscli/s3transfer/manager.py#L53

Are you actually seeing the value not being persisted when configuring `io_chunksize`?

awscli/topics/s3-config.rst
kdaily: Maybe I'm missing something, since I don't see the `RuntimeConfig` object gets used anywhere else, but there is a docstring that was munged during lint/format changes that specifies the use of `DEFAULTS`:

```
# If the user does not specify any overrides,
# these are the default values we use for the s3 transfer
# commands.
```

https://github.com/aws/aws-cli/pull/9678/files#diff-280c24f20aa46551fbfd53f1910ece8ac396fae5de0956cf3555015592d45161L13-L15

The rest of the parameters respected from the `s3` sub-section in the config file are also set in `DEFAULTS`. 
There is a unit test for `RuntimeConfig` that checks if the object created matches what is in `DEFAULTS`. As is, `io_chunksize` won't be set when creating the `RuntimeConfig` object, though, only when creating the `TransferConfig`, like here:

https://github.com/aws/aws-cli/blob/01779c1a4f40f9ad05d1d3ee4b8f28a73d9f9efc/awscli/customizations/s3/subcommands.py#L756

If this object has a new member, shouldn't that test reflect it?

https://github.com/aws/aws-cli/blob/9824a9a3338dfc1579eca3d8f978b561e1a548f1/tests/unit/customizations/s3/test_transferconfig.py#L30

hssyoo: I guess I'm not sure I understand what the problem is when the `TransferConfig` object's initializer has an explicit default value for `io_chunksize`. This is the behavior it uses today, ie AWS CLI will always set `io_chunksize` to 256KB. It's possible the CLI-specific defaults are there from v1, before `botocore` was vended just in case `botocore` decided to change its defaults. But I can just add it to our defaults if it'll make it easier

Issue #9514, opened 2025-05-27
Add search filter to aws account select menu in `aws configure sso` command


frauniki:
### Describe the feature

In large environments, there are cases where hundreds of AWS Accounts are tied to a specific aws org.
Currently, in such cases, finding and selecting the desired aws account is a very tedious task. (We have to use the arrow keys to move items one by one to find the desired account.)

```
$ aws configure sso
SSO session name (Recommended): sample-session-001
There are 150 AWS accounts available to you.
> development-team-a, dev+team-a@example-domain.com (111111111111)
  staging-environment, qa+staging@example-domain.com (222222222222)
  production-web-app, ops+webapp@example-domain.com (333333333333)
  data-lake-project, analytics+datalake@example-domain.com (444444444444)
  test-sandbox-env, sandbox+user@example-domain.com (555555555555)
  security-audit-logs, security+logs@example-domain.com (666666666666)
  company-main-account, admin+main@example-domain.com (777777777777)
```

Adding key bindings to the select menu and the ability to search for aws accounts would make these tasks much easier.

### Use Case

Large-scale AWS Organizations (AWS Accounts 100+)

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

latest

### Environment details (OS name and version, etc.)

none

adev-code:
Hello @frauniki thanks for reaching out. I will bring this up to the team and I will let you know as soon as there are any updates. If you have any questions, please do let me know. Thanks.

johnmayou:
Hi there,

I’m also experiencing this issue, with my dropdown/select showing 401 accounts. It seems that in my organization (~1k developers), all developers are given "read-only" access to all accounts across the org. This drastically increases the number of accounts to search through, making `aws configure sso` challenging to use efficiently.

A search/filter feature would be an incredibly helpful addition for users in large organizations like mine. Thank you for considering this improvement!

adev-code:
We are tracking this similar issue to this issue : https://github.com/aws/aws-cli/issues/6108 , for updates please see the related issue.

Issue #6108, opened 2021-04-21
Sort accounts in aws configure sso


krzysztof-magosa:
**Is your feature request related to a problem? Please describe.**
At the moment `aws configure sso` displays select box with all accounts in random order.
This is problematic having hundreds of accounts.

**Describe alternatives you've considered**
Search functionality would be even better, but sorting still makes sense.

**Describe the solution you'd like**
Alphabetic ordering would be great.

kdaily:
Thanks for the feature request, @krzysztof-magosa.

ryansonshine:
@kdaily , is the aws team open to a PR for this?

ryansonshine:
@krzysztof-magosa , I went ahead and made the assumption that we would want case-insensitive sorting on PR #6205 . If you had other ideas though I'm open to hearing them.

kdaily:
Thanks @ryansonshine, We'll take a look and review.

ruicovelo:
Yes, sorting helps... But I have thousands of accounts to go through. So for me, search is the way to go. I can't use it any other way.

LozanoMatheus:
Can we add the sort in the roles as well? I also thought about having the `--account-id` as a parameter in the CLI, should it be in a new feature request or ...?

YAMLcase:
What's a possible workaround for this?  I need to set up sso for an account and have thousands of unsorted accounts to scroll through.  Can we at least get a clue how the items are ordered?

RyanFitzSimmonsAK:
This is now supported starting in version `2.34.24` of the AWS CLI.  Please feel free to reopen this issue if you have any problems using this feature.  Thank you for contributing!

PR #6205, opened 2021-06-03
Add profile and role sorting to sso configure

ryansonshine:
*Issue #, if available:*
Closes #6108

*Description of changes:*
This change adds sorting to profiles and roles when running `aws sso configure` listing them in ascending alphabetical order by default (case insensitive).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #9692, opened 2025-09-01
[v2] feat: Add account filtering in aws configure sso

frauniki:
*Issue #6108, #9514*

*Description of changes:*
Enable keyboard filtering when selecting AWS accounts during SSO configuration.
Users can now type to filter accounts by name, email, or ID.


frauniki:
I cannot understand why this PR has been left unattended for so long.
#6205

This is undoubtedly a feature many AWS users need right now.
Even in my own environment, I already manage hundreds of AWS accounts and endure significant pain when running `aws configure sso`.
There must be customers managing thousands of AWS accounts. For them, `aws configure sso` must be causing unimaginable pain.

Costya-Y:
Is there a way to bump the priority for this one?

PR #9702, opened 2025-09-04
Fix HTTP 100 Continue parsing to accept optional reason phrase

adev-code:
Merged in Botocore: https://github.com/boto/botocore/pull/3543

Fixes HTTP 100 Continue parsing to accept responses without reason phrases, as allowed by RFC9110. The `_is_100_continue_status()` method currently requires exactly 3 parts (`HTTP/1.1 100 Continue`), which prevents botocore from recognizing valid HTTP 100 responses, causing the client to never see "100 Continue response seen". This breaks compatibility with legitimate HTTP servers that follow the current HTTP specification. The fix changes `len(parts) >= 3` to `len(parts) >= 2` to accept both formats (`HTTP/1.1 100 Continue` or `HTTP/1.1 100`) while maintaining backward compatibility with existing servers.

### DIFF
--- a/awscli/botocore/awsrequest.py
+++ b/awscli/botocore/awsrequest.py
@@ -221,7 +221,7 @@ class AWSConnection:
         parts = maybe_status_line.split(None, 2)
         # Check for HTTP/<version> 100 Continue\r\n
         return (
-            len(parts) >= 3
+            len(parts) >= 2
             and parts[0].startswith(b'HTTP/')
             and parts[1] == b'100'
         )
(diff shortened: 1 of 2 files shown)


adev-code:
> Could we add a changelog entry? I see a few 👍 on the PR, so it'd be good to mention in the changelog.

Added in commit 4cfce15c7f382493996bd46b0004f518d0233513

.changes/next-release/bugfix-botocore-39521.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``botocore``",
```
ashovlin, line 3: Don't see a precedent for `botocore`, which should just be an internal detail for V2. That said, don't really have a better suggestion so I'm okay with it, can always adjust later. See one occurrence of `HTTP`

ashovlin (approved):

PR #9703, opened 2025-09-05
Create setup and initial runtime checkers

aemous:
*Description of changes:*

* Create code architecture for migration readiness feature.
* Implement ecr get-login breaking change checker.
* Implement hidden aliases breaking change checker.
* Implement URL in params breaking change checker.
* Add new-change file.


AndrewAsseily:
Looks like we have a couple of CI tests that need attention

awscli/examples/global_options.rst
```diff
@@ -70,3 +70,7 @@
   
   The maximum socket connect time in seconds. If the value is set to 0, the socket connect will be blocking and not timeout. The default value is 60 seconds.
   
+``--migrate-v2`` (boolean)
+  
+  Enable AWS CLI v2 migration assistance. Prints warnings if the command would face a breaking change after swapping AWS CLI v1 for AWS CLI v2 in the current environment. Prints one warning for each breaking change detected.
+
```
ashovlin: Thinking about it during review, `--migrate-v2` feels a little too "action-y" like we're performing some sort of migration. What do folks think about `--v2-debug` or `--migration-debug`?

awscli/clidriver.py
```diff
@@ -310,6 +320,19 @@ def _handle_top_level_args(self, args):
...
+
+
+    def _print_migration_warnings(self):
+        migration_warnings = []
+        for key, value in MIGRATION_CHECKERS.items():
+            if value.triggered:
+                migration_warnings.append(value.warning_message)
+        for warning in migration_warnings:
+            uni_print(f"AWS CLI v2 MIGRATION WARNING: {warning}")
+
```
ashovlin: I'd rather not build up all of this state in classes and pollute `clidriver.py` like this. Could we treat it like to our existing `logger.debug` calls?

For this case, can we just move the printing to the call site 
https://github.com/aws/aws-cli/blob/44f4c6f4c9dd7aa001d0342ff9097638624832a4/awscli/customizations/ecr.py#L69-L70

with something like 

```
    if parsed_globals.migrate_v2:
       uni_print(...) # or maybe a migration_warning_print(...) if we wanted to centralize formatting
    ecr_client = create_client_from_parsed_globals( ...
```

That way we don't need to worry about printing later if an error happened once we passed the call site.

ashovlin: Thought might be tough if there are places we don't have access to `parsed_globals`.

.changes/next-release/enhancement-Migration-48043.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Migration",
+  "description": "Implement a --migrate-v2 flag that detects breaking changes for AWS CLI v2 for entered commands.\""
```
ashovlin:
```suggestion
  "description": "Implement a ``--migrate-v2`` flag that detects breaking changes for AWS CLI v2 for entered commands."
```

awscli/clidriver.py
aemous: To add a bit of context, the ecr get-login may be the only breaking change that occurs within a customization. 

Here's the full list of breaking changes that depend *only* on the arguments entered to the CLI:

1. A blob type parameter starts with `file://` (args only, globalargs.py)
2. A parameter starts with `http://` or `https://` (args only, globalargs.py)
3. ecr get-login replaced with ecr get-login-password (args only, you're proposing ecr customization)
4. Removed ~20 aliases (args only, globalargs.py)
5. Specifying pagination params via `--cli-input-json` (args only, globalargs.py)

4 of them are not tied to any specific customization and just exist on the level of args/values. Do you agree 1-3, and 5 make sense to implement in globalargs.py since all they depend on is the value of the args? I think its a good place because we're implementing a global arg, and we'd have access to all parsed args from this file.

Also, one breaking change doesn't depend on any information at all (all output will pass through default pager), we may also be able to lump that into `globalargs.py`. We may be able to lump others into globalargs, such as ones depending on env vars and configs. Are we good with implementing as many of these as we can in `globalargs.py` instead of `clidriver.py`, because I think that's what I'm leaning towards now.

And with this context in mind, would you still prefer 3 be implemented within the ecr customization? I feel `globalargs.py` may be a better place since it only depends on value of the args rather than anything specific to ecr such as ecr client creation, or the command implementation itself, etc., and it would be consistent with a lot of other checks, but still open to take more feedback on this, or any of the other breakages I listed.

aemous: Also, for the point on not building up state in a new migrationcheckers class, I can address that in next revision. Rather than the dictionary approach I'll just print from wherever we choose to implement the checks.

aemous: Also, for the point about printing later if an error happens, I think I was being overcautious there and we can likely remove the printing from within the exceptions. The only concern is if the error occurs *before* the migration checkers get printed. And if that occurs, the command won't run anyways, and the error is within somewhere else in CLI that would need to be addressed by the user or us, and would likely even occur in v2 anyways.

ashovlin: - Yeah, fine with shifting over to `globalargs.py` if that's where the majority of them will be - just didn't want to do that here in `clidriver.py` if possible.
- Not too worried about printing in the error mode, and like printing as we go.

awscli/customizations/globalargs.py
```diff
@@ -90,6 +95,21 @@ def resolve_cli_connect_timeout(parsed_args, session, **kwargs):
     arg_name = 'connect_timeout'
     _resolve_timeout(session, parsed_args, arg_name)
 
+def detect_migration_breakage(parsed_args, remaining_args, argument_table, session, **kwargs):
+    if parsed_args.migrate_v2:
+        url_params = [param for param in remaining_args if param.startswith('http://') or param.startswith('https://')]
+        uni_print("AWS CLI v2 MIGRATION WARNING: By default, AWS CLI v2 returns all output through your operating system’s default pager program. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-output-pager.\n")
```
ashovlin: I'm still on the fence for whether this one is useful, since users cannot address it while still on V1, right?

awscli/customizations/globalargs.py
```diff
@@ -90,6 +95,21 @@ def resolve_cli_connect_timeout(parsed_args, session, **kwargs):
...
+        if parsed_args.command == 'ecr' and remaining_args[0] == 'get-login':
+            uni_print('AWS CLI v2 MIGRATION WARNING: The ecr get-login command has been removed in AWS CLI v2. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-ecr-get-login.\n')
+        if url_params:
```
ashovlin: Should this also check `cli_follow_urlparam == true`? If they've already opted out, then V2 should behave the same way.

aemous: I considered this. I ended up not adding the check, because `cli_follow_urlparam` was removed as a config variable in v2. So, if `cli_follow_urlparam == true`, then the user would have a variable set in their config that does not exist in v2. Are we okay with this? 

I do see the benefit of adding the check, it would be nice for users resolving these issues via their config/environment, which is a big point of this tool

awscli/customizations/globalargs.py
aemous: Good point, I'll remove this in next revision.

awscli/customizations/globalargs.py
ashovlin: I'm imagining that if I see this warning:
* If I am relying on the download behavior, I need to refactor that before upgrading to V2. Once I refactor, the warning should disappear. 
* If I am treating it as a string via `cli_follow_urlparam = false`, then I'm already compliant V2's behavior so don't need to adjust my script. Yes, you don't need that setting anymore for V2, but it doesn't hurt to leave it in place. 

I still like considering it here.

awscli/examples/global_options.rst
aemous: Resolved in latest revision

awscli/customizations/globalargs.py
aemous: Addressed in latest revision

ashovlin (approved):

PR #9704, opened 2025-09-05
Fix use_ssl parameter ignored when custom HTTPS endpoint_url provided

adev-code:
Copy from botocore and it was merged: https://github.com/boto/botocore/pull/3542

Fixes a bug where setting `use_ssl=False` with a custom HTTPS endpoint incorrectly downgrades the connection to HTTP. According to the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html), when `endpoint_url` is provided, the `use_ssl` parameter should be ignored. The fix adds a check to ensure this downgrade only happens when no custom endpoint is specified, preserving the HTTPS protocol when customers explicitly provide an HTTPS URL.

### DIFF
--- a/tests/unit/botocore/test_endpoint_provider.py
+++ b/tests/unit/botocore/test_endpoint_provider.py
@@ -16,6 +16,7 @@ import logging
 import os
 
 import pytest
+
 from botocore.endpoint_provider import (
     EndpointProvider,
     EndpointRule,
@@ -23,7 +24,7 @@ from botocore.endpoint_provider import (
     RuleCreator,
     RuleSet,
     RuleSetStandardLibary,
-    TreeRule,
+    TreeRule, RuleSetEndpoint,
 )
 from botocore.exceptions import (
     EndpointResolutionError,


tests/unit/botocore/test_endpoint_provider.py
```diff
@@ -486,3 +488,77 @@ def test_auth_schemes_conversion_first_authtype_unknown(
...
+        service_model=None,
+        builtins={},
+        client_context=None,
+        event_emitter=None,
+        use_ssl=use_ssl,
+        requested_auth_scheme=None,
+    )
+
+    with patch.object(resolver._provider, 'resolve_endpoint') as mock_resolve:
+        mock_resolve.return_value = RuleSetEndpoint(
```
ashovlin, line 558: Looks like we're missing an import that is causing this test to fail in CI.

adev-code: Thanks @ashovlin, fixed and appended the right Import and now waiting for the test to succeed.

ashovlin (approved):

ashovlin:
Okay with no changelog entry for this, CLI users cannot manipulate the client config to set `use_ssl`, so this should be no impact.

PR #9711, opened 2025-09-08
Fix CancelledError retry loops to enable immediate S3 upload cancellation

adev-code:
Copy from Botocore: https://github.com/boto/botocore/pull/3552 

This fix ensures that Ctrl-C and other cancellation signals immediately stop S3 uploads by preventing CancelledError retry loops in the HTTP session, allowing users to terminate operations promptly instead of waiting through lengthy retry cycles. Adding a specific exception handler for CancelledError in httpsession.py that re-raises the error unchanged before it reaches the generic handler. This prevents the cancellation signal from being misclassified as a retryable HTTP error.

### DIFF
--- a/.changes/next-release/bugfix-s3-89705.json
+++ b/.changes/next-release/bugfix-s3-89705.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``s3``",
+  "description": "Fix CancelledError retry loops to enable immediate S3 upload cancellation"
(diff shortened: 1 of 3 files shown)


adev-code:
> Change looks okay, and the botocore version has been out "baking" for a while.
> 
> Can you add a changelog entry?

Added in commit 0620fdea65edbb17bbb480ad7a891bf2439cc207

PR #9717, opened 2025-09-11
Fix printing bugs when using no-overwrite flag

aemous:
*Description of changes:*

* Fixed bug when using `--no-overwrite` parameter where the last character printed to CLI would be a carriage return (`\r`), which would cause incompatibility with some terminal program (e.g. Windows Powershell).
  * Keep track of files skipped for determining whether there is progress remaining in `ResultPrinter`.
  * Update files remaining progress to take into account files skipped.
  * Add support for results that do not print newlines (e.g. new `SkipFileResult`).
  * Modify result handling in `ResultPrinter` so that they can opt-out of resetting `progress_length` to 0. Useful for result handlers that do not print newlines (e.g. new `SkipFileResult`).

*Description of tests:*

* Manually tested changes on Windows Powershell, with uploads, bucket-to-bucket copies, and downloads. Verified the bug is fixed in all three cases.

### DIFF
--- a/awscli/customizations/s3/results.py
+++ b/awscli/customizations/s3/results.py
@@ -457,24 +457,16 @@ class ResultPrinter(BaseResultHandler):
         if self._has_remaining_progress():
             self._print_progress()
         else:
-            LOGGER.debug(f"No progress remaining, length: {self._progress_length}")
             self._clear_progress_if_no_more_expected_transfers(ending_char='\r')
 
     def _print_progress(self, **kwargs):
-        # Get all of the statistics in the correct form.
-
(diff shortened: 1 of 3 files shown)


kdaily (changes requested):
This LGTM. I ran it through the test suite after pulling in the latest changes from `v2` and that was OK. I think we should add a test to exercise the newly added `SkipFileResult`?

kdaily:
LGTM!

Lets make sure to clean up the commit history or squash before merging, preferably getting some of the details from the description on this PR into the final commit(s). Thanks!

kdaily (approved):

Issue #9718, opened 2025-09-14
New AWS CLI v2 still expecting Rosetta on Apple Silicon


technologytailor:
### Describe the bug

Same issue as https://github.com/aws/aws-cli/issues/7252. It was marked as closed. Last comment suggested opening a new issue instead of commenting.

Issue: AWS CLI V2 still requiring Rosetta on macOS.
macOS version: Sequoia 15.2

Link: https://awscli.amazonaws.com/AWSCLIV2.pkg

Result:
<img width="605" height="438" alt="Image" src="https://github.com/user-attachments/assets/4256029b-bdf7-4f77-a708-9907a82c703a" />

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

AWS CLI V2 should install without requiring Rosetta.

### Current Behavior

AWS CLI V2 installation package expects Rosetta installation on macOS.

<img width="605" height="438" alt="Image" src="https://github.com/user-attachments/assets/4256029b-bdf7-4f77-a708-9907a82c703a" />

### Reproduction Steps

1. Download AWS CLI V2 package from https://awscli.amazonaws.com/AWSCLIV2.pkg
2. Run installer

### CLI version used

V2

### Environment details (OS name and version, etc.)

macOS. Sequoia 15.2

technologytailor:
Also tried downloading using versioned URL - https://awscli.amazonaws.com/AWSCLIV2-2.30.0.pkg
Same result.

charles-br1:
Same here - https://github.com/aws/aws-cli/issues/7252#issuecomment-3283844261

rafrafek:
This issue was first reported on Sep 6, 2022, and it remains unresolved. Since Apple Silicon is now widely adopted, it seems important for the installer to be tested natively.

A renewed M1 Mac Mini can be purchased for around ~$250 on Amazon: https://www.amazon.com/mini-Apple-Memory-256GB-MGNR3LL/dp/B08PF2RF76

Alternatively, AWS itself offers Apple Silicon EC2 instances in the cloud: https://aws.amazon.com/ec2/instance-types/mac/

In my opinion, it should be within reach for AWS (Amazon) to acquire hardware for proper testing. Relying on the community for testing such a basic functionality of the company's core utility (AWS CLI) does not seem ideal.

miyohide:
I encountered the same issue.

After trying to install the latest version, I noticed it’s a Universal Binary, so it might be a problem with the installer.
[...]

PR #9721, opened 2025-09-15
Disable rosetta prompt

AndrewAsseily:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/9718


scripts/installers/make-macpkg
```diff
@@ -20,13 +20,17 @@ ROOT = os.path.dirname(
...
-DISTRIBUTION_PATH = os.path.join(PKG_DIR, 'distribution.xml')
 
 TEMP_PKG_NAME = 'aws-cli.pkg'
 PKG_NAME = 'AWS-CLI-Installer.pkg'
 EXE_ZIP_NAME = 'awscli-exe.zip'
 EXE_NAME = 'aws'
 
+def select_distribution_path():
+    base = os.path.join(PKG_DIR, 'distribution.xml')
+    uni = os.path.join(PKG_DIR, 'distribution.universal.xml')
```
hssyoo: What's the rationale for maintaining 2 separate distribution files? What happens if we update `distribution.xml` with the host architectures and then produce an x86-only installer?

macpkg/distribution.universal.xml
```diff
@@ -0,0 +1,29 @@
...
+        my.result.type = 'Warning';
+        my.result.message = 'On November 13, 2024, the AWS CLI v2 dropped support for macOS 10.15. Please upgrade to macOS 11 or later to use up-to-date versions of AWS CLI v2. For more information, please visit: https://aws.amazon.com/blogs/developer/macos-support-policy-updates-for-the-aws-cli-v2/';
+        return false;
+      }
+      return true;
+    }
+    ]]>
+    </script>
+    <domains enable_anywhere="true" enable_currentUserHome="true" enable_localSystem="true" />
+    <options customize="always" require-scripts="false" hostArchitectures="x86_64,arm64" />
```
hssyoo: What docs did you use to determine the correct values here? Asking because Apple's docs [[source](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW41)] doesn't show `arm64` as a valid code for `hostArchitectures`, though I understand it hasn't been updated since 2012

scripts/installers/make-macpkg
AndrewAsseily: It just seems wrong/unintuitive to specify both architectures in the distribution file but build only one (claim "x86_64,arm64" support in the metadata when building x86_64-only). Maintaining a single file would make it tedious to correctly format and replace architecture values between different build types.

macpkg/distribution.universal.xml
AndrewAsseily: Referenced from a developer discussion on Apple's forums [here](https://developer.apple.com/forums/thread/667672?answerId=648995022#648995022)

scripts/installers/make-macpkg
hssyoo: I think maybe I'm not seeing the value of having distribution files for non-universal PKGs at all. It made sense to use `BUILD_UNIVERSAL_BINARY` for non-universal source dist builds since they use PyInstaller, but this is specifically for an actual Mac PKG installer, which I'm not sure is used externally

The situation you described would be wrong in the case that someone goes out of their way to build a Mac PKG with x86-only support, and then distribute it to ARM environments, where they may fail if Rosetta isn't installed

Issue #9715, opened 2025-09-11
`aws ec2-instance-connect ssh` didn't work with updated instance connect endpoint


perotinus:
### Describe the bug

It looks like the `ec2-instance-connect` `ssh` and `open-tunnel` commands might not work correctly with EC2 Instance Connect Endpoints that have been updated.

I adjusted the security group on one of my EC2 Instance Connect Endpoints (using `aws ec2-instance-connect modify-instance-connect-endpoint`, and verified that the update worked, but when I tried to connect to an EC2 instance over SSH via the endpoint, the command failed:

```
$ aws ec2-instance-connect ssh --debug --instance-id i-XXXX
...
There are no available instance connect endpoints.
```

I looked at the debug output and noticed this line:
```
2025-09-10 18:18:54,428 - MainThread - botocore.endpoint - DEBUG - Making request for OperationModel(name=DescribeInstanceConnectEndpoints) with params: {'url_path': '/', 'query_string': '', 'method': 'POST', 'headers': {'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8', 'User-Agent': 'aws-cli/2.29.1 md/awscrt#0.27.6 ua/2.1 os/macos#24.6.0 md/arch#x86_64 lang/python#3.13.7 md/pyimpl#CPython m/Z,C,b,E cfg/retry-mode#standard md/installer#exe md/prompt#off md/command#ec2-instance-connect.ssh'}, 'body': {'Action': 'DescribeInstanceConnectEndpoints', 'Version': '2016-11-15', 'Filter.1.Name': 'state', 'Filter.1.Value.1': 'create-complete', 'Filter.2.Name': 'vpc-id', 'Filter.2.Value.1': 'vpc-XXXXXX'}, 'url': 'https://ec2.us-east-1.amazonaws.com/', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x1131e8050>, 'has_streaming_input': False, 'auth_type': None, 'unsigned_payload': None, 'auth_options': ['aws.auth#sigv4']}}
```

Specifically I noticed `'Filter.1.Name': 'state', 'Filter.1.Value.1': 'create-complete'` in the body. I looked at the instance connect endpoints using the describe command, and noticed that the state of the one that I updated was `update-complete`, which seemed like it would explain why it wasn't found. I recreated the endpoint exactly as it was after the update (except for the state and other automated values), and then SSH started working.

[...]

PR #9731, opened 2025-09-18
Update ec2-instance-connect to handle new endpoint states

CharlesReinhardt:
*Issue number*: https://github.com/aws/aws-cli/issues/9715

Update ec2-instance-connect customizations to support additional states (update-complete, update-in-progress, update-failed) introduced by ModifyInstanceConnectEndpoint. For more information on ModifyInstanceConnectEndpoint, see [Modify an EC2 Instance Connect Endpoint][1]

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/modify-ec2-instance-connect-endpoint.html

### DIFF
--- a/awscli/customizations/ec2instanceconnect/eicefetcher.py
+++ b/awscli/customizations/ec2instanceconnect/eicefetcher.py
@@ -51,7 +51,17 @@ class InstanceConnectEndpointRequestFetcher:
         self, ec2_client, instance_connect_endpoint_id
     ):
         args = {
-            "Filters": [{"Name": "state", "Values": ["create-complete"]}],
+            "Filters": [
+                {
+                    "Name": "state",
+                    "Values": [
+                        "create-complete",
+                        "update-in-progress",
+                        "update-failed",
+                        "update-complete",
+                    ],
+                }
+            ],
             "InstanceConnectEndpointIds": [instance_connect_endpoint_id],
         }
         describe_eice_response = (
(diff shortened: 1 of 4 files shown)


awscli/customizations/ec2instanceconnect/eicefetcher.py
```diff
@@ -51,7 +51,17 @@ def _get_instance_connect_endpoint_by_id(
...
     ):
         args = {
-            "Filters": [{"Name": "state", "Values": ["create-complete"]}],
+            "Filters": [
+                {
+                    "Name": "state",
+                    "Values": [
+                        "create-complete",
+                        "update-in-progress",
+                        "update-failed",
```
hssyoo, line 60: Do we want users to attempt to connect to an endpoint when an update is in progress or failed? I'm wondering if there are any concerns with users transparently connecting to an endpoint where they expect some modified state but is actually in progress/failed.

CharlesReinhardt: Great question. We do want to allow users to be able to attempt connections when an endpoint is in these two states. 

The result of allowing connections during `update-in-progress` and `update-failed` would be *possibly* unexpected behavior when trying to connect depending on the update being performed and the user's network configurations. This unexpected behavior is something the user can troubleshoot by calling `aws ec2 describe-instance-connect-endpoints` to see that their endpoint is in an `update-in-progress` or `update-failed` state, which we feel is better than blocking all users from connecting to an `update-in-progress` or `update-failed` endpoint. 

---

For example, say user A might be performing an update to allow their endpoint to work with existing network configurations. When the endpoint is in `update-in-progress` or `update-failed`, their network configurations would likely block their connections until the update finishes or they make subsequent updates to reach an `update-complete` state. 

Differently, user B might be performing a "nice to have" update that wouldn't have any effect on the endpoint's behavior in relation/cooperation with their network configurations. When the endpoint is in `update-in-progress` or `update-failed`, their endpoint might perform just the same as if it was in `update-complete`. 

If we don't include `update-in-progress` and `update-failed` here, user A might get a better error message, but user B is blocked from using their functional endpoint. We'd rather include `update-in-progress` and `update-failed` to avoid blocking user B.

PR #9734, opened 2025-09-18
[v2] Implement full object checksum validation for multipart downloads

hssyoo:
This PR implements full object checksum validation for multipart downloads when using the high-level `s3` command.

Currently, checksum validation is only done at the low-level S3 client level via Botocore. This current validation only happens when either the object is retrieved in a single GET request, or if the requested range of an object happens to fall in the same boundary as the part size (if the object was uploaded via MPU).

At a high-level:
1. Each download task calls an initial `HeadObject` (existing behavior), where it retrieves the stored checksum value and algorithm, if in the response.
2. If there's a stored checksum and the checksum type is `FULL_OBJECT` and the checksum algorithm is CRC-based, then it provides the checksum value and algorithm to the future meta.
3. If a multipart download is required and the checksum properties are provided to the future meta, then it'll create a `FullObjectChecksum` object that's responsible for storing part checksums and then later combining them to a full object checksum and validation.
4. As parts are being downloaded, the response streams are wrapped by `PartStreamingChecksumBody`. This class calculates the checksum for each part, unless the underlying stream is already calculating the checksum in which case the underlying stream's checksum is reused.
5. Once all part download tasks are completed, a final validation task is invoked. `FullObjectChecksum` combines all part-level checksums and then validates the calculated checksum against the stored checksum.

Note that `s3transfer.checksums.combine_crc32` will only be added to the S3Transfer library, not in AWS CLI v2's vended library because AWS CLI v2 can just use CRT's future bindings. It's included in this PR for now so reviewers can play around with it.

**To manually test:**
Upload an object with CRC32 with a single PUT request:
```
aws s3api put-object --body myobject --bucket mybucket --key mykey --checksum-algorithm CRC32
```

Ensure checksum type is `FULL_OBJECT` and has CRC32 checksum value:
```
aws s3api head-object --bucket mybucket --key mykey --checksum-mode ENABLED
{
    "AcceptRanges": "bytes",
    "LastModified": "2025-09-15T14:13:18+00:00",
    "ContentLength": 39542919,
    "ChecksumCRC32": "mychecksum",
    "ChecksumType": "FULL_OBJECT",
    "ETag": "\"myetag"",
    "ContentType": "application/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}
```

Download object using multipart ranged GETs:
```
aws s3 cp s3://mybucket/mykey /tmp/ --debug
```

In the debug logs you should see something like:
```
s3transfer.checksums - DEBUG - Successfully validated stored checksum mychecksum against calculated checksum mychecksum
```

### DIFF
--- a/awscli/s3transfer/checksums.py
+++ b/awscli/s3transfer/checksums.py
@@ -18,6 +18,20 @@ subject to abrupt breaking changes. Please do not use them directly.
 from copy import copy
 
 
+def provide_checksum_to_meta(response, transfer_meta):
+    stored_checksum = None
+    checksum_algorithm = None
+    checksum_type = response.get("ChecksumType")
+    if checksum_type and checksum_type == "FULL_OBJECT":
+        for crc_checksum in _CRC_CHECKSUM_TO_COMBINE_FUNCTION.keys():
+            if checksum_value := response.get(crc_checksum):
+                stored_checksum = checksum_value
+                checksum_algorithm = crc_checksum
+                break
+    transfer_meta.provide_checksum_algorithm(checksum_algorithm)
+    transfer_meta.provide_stored_checksum(stored_checksum)
+
+
 def combine_crc32(crc1, crc2, len2):
     """Combine two CRC32 values.
 
--- a/awscli/s3transfer/checksums.py
+++ b/awscli/s3transfer/checksums.py
@@ -20,6 +20,7 @@ from copy import copy
 from functools import cached_property
 
 from botocore.httpchecksum import CrtCrc32Checksum
+from s3transfer.exceptions import S3ValidationError
 
 
 class PartStreamingChecksumBody:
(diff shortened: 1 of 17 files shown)


awscli/s3transfer/checksums.py
```diff
@@ -0,0 +1,222 @@
...
+        for offset, next_offset in zip(remaining_offsets, next_offsets):
+            part_checksum = self._part_checksums[offset]
+            offset_len = next_offset - offset
+            combined = self._combine_function(
+                combined, part_checksum, offset_len
+            )
+
+        self._calculated_checksum = base64.b64encode(
+            combined.to_bytes(4, byteorder='big')
+        ).decode('ascii')
```
hssyoo, line 113: This is actually dependent on the checksum algorithm. I think instead I'll instantiate the appropriate checksum object, set the `int_crc` property to the combined value, then call the `digest()` method

PR #9737, opened 2025-09-19
[v2] Adjust error message for accessibility

kdaily:
*Issue #, if available:*

*Description of changes:*

Make error messages more accessible.

Per user feedback, the error messages are not easy to discover when using assistive technologies, like screen readers. This change adds more structure and differentiation to the error messages when running a non-existent AWS CLI command or sub-command by capitalizing and surrounding in square brackets the "error" string in the error message.

Previous error:

```
aws: error: Some error message.
```

New error:

```
aws: [ERROR]: Some error message.
```


aemous (approved):
LGTM

PR #9772, opened 2025-10-08
Implement v2-debug checks for customizations, s3, and environment variables

aemous:
*Description of changes:*

* Implemented runtime checks for the following breaking changes via `--v2-debug`:
  * Usage of `PYTHONUTF8` and `PYTHONIOENCODING` env vars unsupported in v2.
  * S3 copies executed under `aws s3` namespace may call additional APIs for multipart S3 copies 
  * Signer used for S3 request is Sigv2 (I expect it's very rare for customers to be using sigv2 and run into this)
  * [!] `aws cloudformation deploy` will not fail on empty changesets.
  * Timestamps in responses will be formatted in ISO 8601.
* Line-length formatting of other `--v2-debug` code.
* Added tests to cover some added functionality

*Description of tests:*

* Ran and passed all new tests.
* Manually tested the file encoding checker by supplying  `PYTHONUTF8`:
```
$ PYTHONUTF8=1 python3 -m awscli s3 ls --v2-debug
AWS CLI v2 MIGRATION WARNING: The PYTHONUTF8 and PYTHONIOENCODING environment variables are unsupported in AWS CLI v2. AWS CLI v2 uses AWS_CLI_FILE_ENCODING instead, set this environment variable to resolve this. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-encodingenvvar.

An error occurred (ExpiredToken) when calling the ListBuckets operation: The provided token has expired.
```
* Manually tested the s3 copies checker via `aws s3 cp`:
```
$ python3 -m awscli s3 cp s3://BUCKET1/file1 s3://BUCKET2 --v2-debug
AWS CLI v2 MIGRATION WARNING: In AWS CLI v2, object properties will be copied from the source in multipart copies between S3 buckets. This may result in extra S3 API calls being made. Breakage may occur if the principal does not have permission to call these extra APIs. This warning cannot be resolved. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-s3-copy-metadata

copy: s3://BUCKET1/file1 to s3://BUCKET2
```
* Manually tested the timestamps checker via `aws s3 ls`:
```
AWS CLI v2 MIGRATION WARNING: In AWS CLI v2, all timestamp response values are returned in the ISO 8601 format. To migrate to v2 behavior and resolve this warning, set the configuration variable `cli_timestamp_format` to `iso8601`. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-timestamp.
[...]
```

### DIFF
--- a/awscli/customizations/cloudformation/deploy.py
+++ b/awscli/customizations/cloudformation/deploy.py
@@ -24,6 +24,7 @@ from awscli.customizations.cloudformation.yamlhelper import yaml_parse
 
 from awscli.customizations.commands import BasicCommand
 from awscli.compat import get_stdout_text_writer
+from awscli.customizations.utils import uni_print
 from awscli.utils import create_nested_client, write_exception
 
 LOG = logging.getLogger(__name__)
@@ -321,12 +322,13 @@ class DeployCommand(BasicCommand):
                            parsed_args.execute_changeset, parsed_args.role_arn,
                            parsed_args.notification_arns, s3_uploader,
                            tags, parsed_args.fail_on_empty_changeset,
-                           parsed_args.disable_rollback)
+                           parsed_args.disable_rollback, parsed_args.v2_debug)
 
     def deploy(self, deployer, stack_name, template_str,
                parameters, capabilities, execute_changeset, role_arn,
                notification_arns, s3_uploader, tags,
-               fail_on_empty_changeset=True, disable_rollback=False):
+               fail_on_empty_changeset=True, disable_rollback=False,
+               v2_debug=False):
         try:
             result = deployer.create_and_wait_for_changeset(
                 stack_name=stack_name,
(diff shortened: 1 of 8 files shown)


aemous:
I'm curious how reviewers feel about printing the timestamp field warning for each timestamp in the response. The PR description shows an example of `aws s3 ls`, where many S3 timestamps are returned, and it prints the warning for each timestamp field encountered over all responses. This is because there is no good way to check for the presence of timestamp fields outside of the response parser, and that gets called for each timestamp.

If we want to make sure each warning gets printed at most once per command, we would have to likely introduce some state that tracks each warning type and whether it has been triggered yet (e.g. [the original design](https://github.com/aws/aws-cli/pull/9703#discussion_r2337892577)). Curious to get feedback on this.

PR #9775, opened 2025-10-09
Implement v2-debug checks for args, config, customizations, s3, and environment variables

aemous:
*Description of changes:*

* Implemented runtime checks for the following breaking changes via `--v2-debug`:
  * Usage of `PYTHONUTF8` and `PYTHONIOENCODING` env vars unsupported in v2.
  * S3 copies executed under `aws s3` namespace may call additional APIs for multipart S3 copies 
  * Signer used for S3 request is Sigv2 (I expect it's very rare for customers to be using sigv2 and run into this)
  * [!] `aws cloudformation deploy` will not fail on empty changesets.
  * Timestamps in responses will be formatted in ISO 8601.
  * Usage of `file://` with blob-type parameters.
  * `us-east-1` will use the regional endpoint in v2 rather than global.
  * Usage of `api_versions` in config will not be supported in v2.
  * Support for `[plugins]` will be provisional in v2.
  * Pagination params used in `--cli-input-json` will disable automatic paging in v2.
* Line-length formatting of other `--v2-debug` code.
* Added tests to cover some added functionality

*Description of tests:*

* Ran and passed all new tests.
* Manually tested the file encoding checker by supplying  `PYTHONUTF8`:
```
$ PYTHONUTF8=1 python3 -m awscli s3 ls --v2-debug
AWS CLI v2 MIGRATION WARNING: The PYTHONUTF8 and PYTHONIOENCODING environment variables are unsupported in AWS CLI v2. AWS CLI v2 uses AWS_CLI_FILE_ENCODING instead, set this environment variable to resolve this. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-encodingenvvar.

An error occurred (ExpiredToken) when calling the ListBuckets operation: The provided token has expired.
```
* Manually tested the s3 copies checker via `aws s3 cp`:
```
$ python3 -m awscli s3 cp s3://BUCKET1/file1 s3://BUCKET2 --v2-debug
AWS CLI v2 MIGRATION WARNING: In AWS CLI v2, object properties will be copied from the source in multipart copies between S3 buckets. This may result in extra S3 API calls being made. Breakage may occur if the principal does not have permission to call these extra APIs. This warning cannot be resolved. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-s3-copy-metadata

copy: s3://BUCKET1/file1 to s3://BUCKET2
```
* Manually tested the timestamps checker via `aws s3 ls`:
```
AWS CLI v2 MIGRATION WARNING: In AWS CLI v2, all timestamp response values are returned in the ISO 8601 format. To migrate to v2 behavior and resolve this warning, set the configuration variable `cli_timestamp_format` to `iso8601`. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-timestamp.
[...]
```
* Manually tested specifying pagination params through `--cli-input-json`:
```
 $ aws s3api list-buckets --v2-debug --cli-input-json file://pagination-params.json 
AWS CLI v2 MIGRATION WARNING: In AWS CLI v2, if you specify pagination parameters by using a file with the `--cli-input-json` parameter, automatic pagination will be turned off. This is not the case in v1. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-skeleton-paging.
[...]
```
* Manually tested specifying `api_versions` in config file:
```
api_versions =
    ec2 = 2016-11-15
    s3 = 2006-03-01
```
```
$ aws s3 ls --v2-debug
AWS CLI v2 MIGRATION WARNING: The AWS CLI v2 does not support calling earlier versions of AWS service APIs via the `api_versions` configuration file setting. To migrate to v2 behavior and resolve this warning, remove the `api_versions` setting in the configuration file. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-api-versions.
```
* Manually tested using `file://` for a blob-type parameter:

```
$ python3 -m awscli secretsmanager update-secret --secret-id SECRET-NAME --secret-binary file://BINARY-SECRET.json --v2-debug
AWS CLI v2 MIGRATION WARNING: When specifying a blob-type parameter starting with `file://`, AWS CLI v2 will assume the content of the file is already base64-encoded. To maintain v1 behavior after upgrading to v2, set the `cli_binary_format` configuration variable to `raw-in-base64-out`. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-binaryparam.
{
    "ARN": "ARN",
    "Name": "SECRET-NAME",
    "VersionId": "VERSION-ID"
}
```

After setting  `cli_binary_format` to `raw-in-base64-out`, I've verified that the warning above was not printed.

* Manually tested configuring region to `us-east-1` and executing an S3 command:

```
$ python3 -m awscli s3 ls --v2-debug 
AWS CLI v2 MIGRATION WARNING: When you configure AWS CLI v2 to use the `us-east-1` region, it uses the true regional endpoint rather than the global endpoint. To continue using the global endpoint on v2, configure the region to be `aws-global`. See https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-s3-regional-endpoint.
```

After setting `region` to `aws-global`, I've verified that the warning above was not printed.

### DIFF
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -697,7 +697,8 @@ class ServiceOperation:
                     '`cli_binary_format` configuration variable to '
                     '`raw-in-base64-out`. See https://docs.aws.amazon.com/cli/'
                     'latest/userguide/cliv2-migration-changes.html#'
-                    'cliv2-migration-binaryparam.\n'
+                    'cliv2-migration-binaryparam.\n',
+                    out_file=sys.stderr
                 )
 
 
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -546,7 +546,7 @@ class ServiceOperation:
             parsed_args, self.arg_table
         )
 
-        self._detect_migration_breakage(
+        self._detect_binary_file_migration_change(
             self._session,
             parsed_args,
             parsed_globals,
@@ -668,7 +668,13 @@ class ServiceOperation:
         parser = ArgTableArgParser(arg_table)
         return parser
 
-    def _detect_migration_breakage(self, session, parsed_args, parsed_globals, arg_table):
+    def _detect_binary_file_migration_change(
+            self,
+            session,
+            parsed_args,
+            parsed_globals,
+            arg_table
+    ):
         if (
                 session.get_scoped_config()
                         .get('cli_binary_format', None) == 'raw-in-base64-out'
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -669,7 +669,10 @@ class ServiceOperation:
         return parser
 
     def _detect_migration_breakage(self, session, parsed_args, parsed_globals, arg_table):
-        if session.get_config_variable('cli_binary_format') == 'raw-in-base64-out':
+        if (
+                session.get_scoped_config()
+                        .get('cli_binary_format', None) == 'raw-in-base64-out'
+        ):
             # if cli_binary_format is set to raw-in-base64-out, then v2 behavior will
             # be the same as v1, so there is no breaking change in this case.
             return
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -529,7 +529,6 @@ class ServiceOperation:
         operation_parser = self._create_operation_parser(self.arg_table)
         self._add_help(operation_parser)
         parsed_args, remaining = operation_parser.parse_known_args(args)
-        # print(f"operation parser parsed args: {parsed_args}")
         if parsed_args.help == 'help':
             op_help = self.create_help_command()
             return op_help(remaining, parsed_globals)
@@ -547,8 +546,12 @@ class ServiceOperation:
             parsed_args, self.arg_table
         )
 
-        self._detect_migration_breakage(self._session, parsed_args, parsed_globals, self.arg_table)
-
+        self._detect_migration_breakage(
+            self._session,
+            parsed_args,
+            parsed_globals,
+            self.arg_table
+        )
         event = f'calling-command.{self._parent_name}.{self._name}'
         override = self._emit_first_non_none_response(
             event,
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -79,8 +79,6 @@ def main():
 def create_clidriver():
     session = botocore.session.Session(EnvironmentVariables)
     _set_user_agent_for_session(session)
-    # TODO check if full config plugins is empty or not. if it's not, we signal the warning for plugin support being provisional
-    # similarly, we check for api_versions config value here.
     load_plugins(
         session.full_config.get('plugins', {}),
         event_hooks=session.get_component('event_emitter'),
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -41,6 +41,7 @@ from awscli.arguments import (
 )
 from awscli.commands import CLICommand
 from awscli.compat import get_stderr_text_writer
+from awscli.customizations.utils import uni_print
 from awscli.formatter import get_formatter
 from awscli.help import (
     OperationHelpCommand,
@@ -528,6 +529,7 @@ class ServiceOperation:
         operation_parser = self._create_operation_parser(self.arg_table)
         self._add_help(operation_parser)
         parsed_args, remaining = operation_parser.parse_known_args(args)
+        # print(f"operation parser parsed args: {parsed_args}")
         if parsed_args.help == 'help':
             op_help = self.create_help_command()
             return op_help(remaining, parsed_globals)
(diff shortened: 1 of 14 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -1019,7 +1019,7 @@ def choose_sync_strategies(self):
 
         return sync_strategies
 
-    def run(self):
+    def run(self, v2_debug=False):
```
ashovlin: Could this go through `self.parameters` instead? That looks like where we're handling other info from `parsed_globals`.

awscli/customizations/scalarparse.py
```diff
@@ -64,8 +68,23 @@ def add_timestamp_parser(session):
         # parser (which parses to a datetime.datetime object) with the
         # identity function which prints the date exactly the same as it comes
         # across the wire.
-        # TODO create an inner function that wraps identity here. it'll signal to print the runtime check.
-        timestamp_parser = identity
+        encountered_timestamp = False
+        def identity_with_warning(x):
+            nonlocal encountered_timestamp
+            if not encountered_timestamp:
+                encountered_timestamp = True
```
ashovlin, line 78: Can you explain this flow better in comments? Is it so we're only printing a single warning per response?

aemous: Yes, without adding these guards I was getting flooded with messages (one per timestamp in the responses), which nearly made the feature unusable.

I'll update with comments

awscli/customizations/paginate.py
```diff
@@ -253,6 +275,32 @@ def _get_cli_name(param_objects, token_name):
             return param.cli_name.lstrip('-')
 
 
+# This function checks for pagination args in the actual calling
+# arguments passed to the function.  If the user is using the
+# --cli-input-json parameter to provide JSON parameters they are
+# all in the API naming space rather than the CLI naming space
+# and would be missed by the processing above.  This function gets
+# called on the calling-command event.
```
ashovlin: Generally prefer docstrings throughout. Though we don't generate user-facing docs from these, they're more helpful in IDEs for us and contributors.

awscli/customizations/globalargs.py
```diff
@@ -97,18 +97,127 @@ def resolve_cli_connect_timeout(parsed_args, session, **kwargs):
...
+                    'https://docs.aws.amazon.com/cli/latest/userguide'
+                    '/cliv2-migration-changes.html#cliv2-migration-aliases.\n',
+                    out_file=sys.stderr
+                )
+        session.register('choose-signer.s3.*', warn_if_sigv2)
+
+def warn_if_east_configured_global_endpoint(request, operation_name, **kwargs):
+    # The regional us-east-1 endpoint is used in certain cases (e.g.
+    # FIPS/Dual-Stack is enabled). Rather than duplicating this logic
+    # from botocore, we check the endpoint URL directly.
```
ashovlin, line 194: Is this why we need the actual endpoint check in addition to the `us_east_1_regional_endpoint` + `us-east-1` logic above?

awscli/clidriver.py
```diff
@@ -663,6 +668,39 @@ def _create_operation_parser(self, arg_table):
         parser = ArgTableArgParser(arg_table)
         return parser
 
+    def _detect_migration_breakage(self, session, parsed_args, parsed_globals, arg_table):
```
ashovlin: If this is only dealing with the `file://`, then I think it'd be good to add that to the method name. I wouldn't want someone else to add logic that gets skipped by the early return.

awscli/customizations/globalargs.py
```diff
@@ -97,18 +97,127 @@ def resolve_cli_connect_timeout(parsed_args, session, **kwargs):
 
 def detect_migration_breakage(parsed_args, remaining_args, session, **kwargs):
     if parsed_args.v2_debug:
```
ashovlin: Given how long this got, it'd be nice to return early here if the mode isn't enabled to save an indent level.

awscli/customizations/paginate.py
```diff
@@ -253,6 +275,32 @@ def _get_cli_name(param_objects, token_name):
...
+def check_should_enable_pagination_call_parameters(
+    input_tokens, call_parameters, parsed_args, parsed_globals, **kwargs
+):
+    if parsed_globals.v2_debug:
+        pagination_params_in_input_tokens = [
+            param for param in call_parameters if param in input_tokens
+        ]
```
ashovlin, line 304: How is this limited to `--cli-input-json`?

awscli/customizations/globalargs.py
aemous: Yes

awscli/customizations/paginate.py
aemous: Good point! Did some testing and it actually prints the warning even when its specified on the command directly. I'll need to fix this to maybe check what params came from cli-input-json and access them in paginate.py. 

Will address in next revision

ashovlin (approved):

Issue #5943, opened 2021-02-14
Please update pinned versions


edrozenberg:
AWS CLI is unique on my system in pinning many things to specific versions, and pinning them to versions that are very old and no longer in common usage by "most people". I tested just now with the latest aws cli versions available.

Would be great to see AWS CLI remove the pins because they create ongoing, irresolvable conflicts.

**awscli==2.1.26**
```
 - colorama [required: >=0.2.5,<0.4.4, installed: 0.4.4]
 - docutils [required: >=0.10,<0.16, installed: 0.16]
 - cryptography [required: >=2.8.0,<=2.9.0, installed: 3.4.4]
 - ruamel.yaml [required: >=0.15.0,<0.16.0, installed: ?]
 - wcwidth [required: <0.2.0, installed: 0.2.5]
 - prompt-toolkit [required: >=2.0.0,<3.0.0, installed: 3.0.16]
```

**awscli==1.19.7**
```
 - docutils [required: >=0.10,<0.16, installed: 0.16]
 - PyYAML [required: >=3.10,<5.4, installed: 5.4.1]
 - colorama [required: >=0.2.5,<0.4.4, installed: 0.4.4]
 - rsa [required: >=3.1.2,<=4.5.0, installed: 4.7]
```

kdaily:
Hi @edrozenberg,

Thanks for your comment. We will continue to be conservative with version ranges and won't be removing the ceiling by default. This is to be sure that we maintain backwards compatibility as much as possible, and we have the chance to review any interface changes to the dependencies. 

However, we can improve the process for raising the ceiling when new versions become available. It would be a nice feature to get notified on new dependency versions and automatically run interface testing on them so that we can respond to these changes more quickly. For the time being, if you have specific packages that you feel that need to be bumped up, please file an issue for them and describe the conflicts that you're getting with them.

edrozenberg:
@kdaily thanks, maybe the common usage to dedicate a machine or VM to be the "aws cli" machine, because the pinned versions of the aws cli reqs can prevent running other packages on the same machine that require newer versions.

weddige:
I would really appreciate, if you could update `colorama` and `cryptography`. If you don't want to remove the ceiling, you maybe could at least unpin the minor versions.

kdaily:
[...]

PR #9790, opened 2025-10-14
Add support for Python 3.14

hssyoo:
Add support for Python 3.14. Mostly chores:
- Add 3.14 to GH actions testing matrix
- Update README.md, setup.py, and tox.ini

It also replaces `%` with `%%` when adding AWS CLI command parameters to the arg parser. The latest versions of `argparse` added a check to raise exceptions when invalid formatted strings are passed to the help docs: https://github.com/python/cpython/pull/124899. Some command parameters will fail this check because they have `%` characters in them, eg from `aws s3api list-objects help`:
```
NOTE:
     When using the URL encoding type, non-ASCII characters that are
     used in an object's key name will be percent-encoded according to
     UTF-8 code values. For example, the object test_file(3).png will
     appear as test_file%283%29.png .
```

`argparse` sees `%` and thinks it's a formatted string that needs to be expanded. And then it throws an exception when it fails to do so. We escape it with `%%` so it's interpreted as a literal `%`.

Verified that the bulit reference guide does not contain double `%`:
image

### DIFF
--- a/awscli/arguments.py
+++ b/awscli/arguments.py
@@ -449,7 +449,7 @@ class CLIArgument(BaseCLIArgument):
         cli_name = self.cli_name
         parser.add_argument(
             cli_name,
-            help=self.documentation,
+            help=self.documentation.replace('%', '%%'),
             type=self.cli_type,
             required=self.required,
         )


chenrui333:
- there is some issue of building awscli 2.31.18 against py3.14, need https://github.com/aws/aws-cli/pull/9662

chenrui333:
- Not sure about that the team thinks of this idea, https://github.com/aws/aws-cli/issues/5943

PR #9797, opened 2025-10-16
Updating customizations/utils.py

varungovinda:
Using botocore to fecth partitions

### DIFF
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -16,8 +16,10 @@ Utility functions to make it easier to work with customizations.
 """
 import copy
 import sys
+import re
 
 from botocore.exceptions import ClientError
+from botocore.loaders import Loader
 from awscli.utils import create_nested_client
 
 
@@ -222,9 +224,12 @@ def uni_print(statement, out_file=None):
 def get_policy_arn_suffix(region):
     """Method to return region value as expected by policy arn"""
     region_string = region.lower()
-    if region_string.startswith("cn-"):
-        return "aws-cn"
-    elif region_string.startswith("us-gov"):
-        return "aws-us-gov"
-    else:
-        return "aws"
+    loader = Loader()
+    partitions_data = loader.load_data('partitions')
+
+    for partition in partitions_data['partitions']:
+        region_regex = partition['regionRegex']
+        if re.match(region_regex, region_string):
+            return partition['id']
+
+    return "aws"
--- a/tests/unit/customizations/test_utils.py
+++ b/tests/unit/customizations/test_utils.py
@@ -259,6 +259,7 @@ class TestUniPrint(unittest.TestCase):
 
 class TestGetPolicyARNSuffix(unittest.TestCase):
     def test_get_policy_arn_suffix(self):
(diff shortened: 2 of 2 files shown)


tests/unit/customizations/test_utils.py
```diff
@@ -259,6 +259,7 @@ def test_encoding_statement_fails_are_replaced(self):
 
 class TestGetPolicyARNSuffix(unittest.TestCase):
     def test_get_policy_arn_suffix(self):
+        self.assertEqual("aws-cn", utils.get_policy_arn_suffix("eusc-de-east-1"))
```
AndrewAsseily, line 262: The assertion for "eusc-de-east-1" expecting "aws-cn" seems incorrect

AndrewAsseily: Please make sure to test this locally as well (https://aws.github.io/aws-cli/testing.html)

AndrewAsseily: We should try to maintain consistency across v1 and v2 codebases (Code formatting and spacing, Test case ordering, Function organization and sequencing)

awscli/customizations/utils.py
```diff
@@ -222,9 +224,12 @@ def uni_print(statement, out_file=None):
...
     """Method to return region value as expected by policy arn"""
     region_string = region.lower()
-    if region_string.startswith("cn-"):
-        return "aws-cn"
-    elif region_string.startswith("us-gov"):
-        return "aws-us-gov"
-    else:
-        return "aws"
+    loader = Loader()
+    partitions_data = loader.load_data('partitions')
```
AndrewAsseily, line 228: I don't think it's necessary now, but would be nice to eventually cache this to avoid loading it every time.

AndrewAsseily (changes requested):
Thanks for opening this PR. Left a few comments that apply (mostly) to both v1 and v2.
Also, to ensure the CI checks run properly, could you please create a new feature branch (e.g feature/botocore-util-partitions) from develop, push your changes there and update this PR to use that branch as the source? Pushing directly to develop seems to be preventing CI from triggering.

tests/unit/customizations/test_utils.py
```diff
@@ -259,6 +259,7 @@ def test_encoding_statement_fails_are_replaced(self):
 
 class TestGetPolicyARNSuffix(unittest.TestCase):
```
AndrewAsseily, line 260: Thinking about this a little more. It would be nice to include tests that ensure we are backwards compatible. i.e explicit tests testing the previous hard-coded mapping (aws-cn and aws-us-gov).

awscli/customizations/utils.py
```diff
@@ -222,9 +224,12 @@ def uni_print(statement, out_file=None):
...
+
+    for partition in partitions_data['partitions']:
```
AndrewAsseily, line 230: Botocore already exposes EndpointResolver.get_partition_for_region(region) which handles this logic internally. Using it would remove the manual regex loop.

AndrewAsseily (changes requested):

PR #9800, opened 2025-10-17
Add 'wire' as a valid value for cli_timestamp_format.

aemous:
*Description of changes:*

* Added `wire` as a valid value for `cli_timestamp_format` configuration setting, to align with our [published documentation](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#cli-config-cli_timestamp_format).

*Description of tests:*

* Added a unit test that tests setting the config value to `wire`. Verified all unit and functional tests pass locally.

### DIFF
--- a/.changes/next-release/enhancement-timestamps-52310.json
+++ b/.changes/next-release/enhancement-timestamps-52310.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "timestamps",
-  "description": "Add `wire` as a valid value for `cli_timestamp_format`."
+  "description": "Add ``wire`` as a valid value for ``cli_timestamp_format``."
 }
--- a/.changes/next-release/enhancement-timestamps-52310.json
+++ b/.changes/next-release/enhancement-timestamps-52310.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "timestamps",
+  "description": "Add `wire` as a valid value for `cli_timestamp_format`."
+}
--- a/awscli/customizations/scalarparse.py
+++ b/awscli/customizations/scalarparse.py
@@ -71,7 +71,7 @@ def add_timestamp_parser(session):
         timestamp_parser = iso_format
     else:
         raise ValueError('Unknown cli_timestamp_format value: %s, valid values'
-                         ' are "wire" or "iso8601"' % timestamp_format)
+                         ' are "none", "wire" or "iso8601"' % timestamp_format)
     factory.set_parser_defaults(timestamp_parser=timestamp_parser)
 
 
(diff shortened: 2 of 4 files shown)


.changes/next-release/enhancement-timestamps-52310.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "timestamps",
+  "description": "Add `wire` as a valid value for `cli_timestamp_format`."
```
ashovlin: nit: double `` for RST

awscli/customizations/scalarparse.py
```diff
@@ -69,7 +71,7 @@ def add_timestamp_parser(session):
         timestamp_parser = iso_format
     else:
         raise ValueError('Unknown cli_timestamp_format value: %s, valid values'
-                         ' are "none" or "iso8601"' % timestamp_format)
+                         ' are "wire" or "iso8601"' % timestamp_format)
```
ashovlin: I'd probably leave `none` here in the message since it's still supported, one fewer possible surprise for users.

ashovlin (commented):
Could you also update here? https://github.com/aws/aws-cli/blob/582ad8d87f4d09fdc2747ced944663b351deaca9/awscli/topics/config-vars.rst?plain=1#L101

awscli/customizations/scalarparse.py
aemous: Fixed in latest revision

.changes/next-release/enhancement-timestamps-52310.json
aemous: Fixed in latest revision

awscli/customizations/scalarparse.py
```diff
@@ -49,17 +49,19 @@ def add_timestamp_parser(session):
     try:
         timestamp_format = session.get_scoped_config().get(
             'cli_timestamp_format',
-            'none')
```
kdaily, line 52: Do we need to change this here, since the backwards compatible behavior is preserved? Don't we only really need to handle the new case when a user actually sets the value to `wire`?

aemous: Technically no we don't *need* to change that, for the reason you stated. `wire` and `none` are equivalent and interchangeable in this code. I think I prefer to change it to `wire` just because that's the official/documented option we have published in our docs, and I want to treat `none` as an option we only support because we happen to support it before.

aemous:
> Could you also update here?
> 
> https://github.com/aws/aws-cli/blob/582ad8d87f4d09fdc2747ced944663b351deaca9/awscli/topics/config-vars.rst?plain=1#L101

Revised to align with the dev guide.

ashovlin (approved):

PR #9803, opened 2025-10-21
Initial architecture for AWS CLI v1-to-v2 upgrade linting tool

aemous:
*Description of changes:*

* Setup project structure for AWS CLI v1-to-v2 upgrade linting tool, including docs, requirements files, and a Makefile.
* Implemented architecture for AWS CLI v1-to-v2 upgrade linting tool CLI.
* Implemented one breaking change: https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-binaryparam

*Description of tests:*

* Wrote and passed unit tests asserting expected behavior of the primary components and interfaces.
* Thoroughly tested the tool manually and verified expected behavior.

*Example output:*

```
$ upgrade-aws-cli --script examples/upload_s3_files.sh -i --output upload_s3_files_v2.sh

[1/2] binary-params-base64
In AWS CLI v2, an input parameter typed as binary large object (BLOB) expects the input to be base64-encoded. To retain v1 behavior after upgrading to AWS CLI v2, add `--cli-binary-format raw-in-base64-out`.

@@ -1,8 +1,8 @@
#!/bin/bash
# Example script with AWS CLI v1 patterns

-aws secretsmanager put-secret-value --secret-id secret1213 \
-  --secret-binary file://data.json
+aws secretsmanager put-secret-value --secret-id secret1213 \
+  --secret-binary file://data.json --cli-binary-format raw-in-base64-out                                                                                                                                        

if
 aws kinesis put-record --stream-name samplestream --data file://data --partition-key samplepartitionkey ; then

Apply this fix? [y]es, [n]o, [u]pdate all, [q]uit: y

[2/2] binary-params-base64
In AWS CLI v2, an input parameter typed as binary large object (BLOB) expects the input to be base64-encoded. To retain v1 behavior after upgrading to AWS CLI v2, add `--cli-binary-format raw-in-base64-out`.

@@ -5,7 +5,7 @@
  --secret-binary file://data.json

if
- aws kinesis put-record --stream-name samplestream --data file://data --partition-key samplepartitionkey ; then
+ aws kinesis put-record --stream-name samplestream --data file://data --partition-key samplepartitionkey --cli-binary-format raw-in-base64-out ; then
    echo "command succeeded."
fi


Apply this fix? [y]es, [n]o, [u]pdate all, [q]uit: y
Fixed script written to: upload_s3_files_v2.sh
```


awsclilinter/Makefile
hssyoo, line 1: Not sure I understand the purpose of this Makefile. It looks like it can be largely replaced with modern Python dev tools like `uv` and `ruff` (also curious why we're using `black` and `isort` here).

awsclilinter/README.md
```diff
@@ -0,0 +1,74 @@
+# AWS CLI v1-to-v2 Upgrade Linter
+
+A CLI tool that lints bash scripts for AWS CLI v1 usage and updates them to avoid breaking 
+changes introduced in AWS CLI v2. Not all breaking changes can be detected statically, 
+thus not all of them are supported by this tool.
+
+For a full list of the breaking changes introduced with AWS CLI v2, see 
+[Breaking changes between AWS CLI version 1 and AWS CLI version 2](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-changes-breaking).
+
+## Installation
```
hssyoo, line 10: I thought the idea was to publish this as a separate PyPI package? Why are we instructing users to build from source?

awsclilinter/README.md
```diff
@@ -0,0 +1,74 @@
...
+1. Create a virtual environment:
+```bash
+python3.12 -m venv venv
+source venv/bin/activate
+```
+
+2. Install dependencies:
+```bash
+pip install -r requirements.txt
+pip install -r requirements-dev-lock.txt
```
hssyoo: Why do we anticipate users to require dev dependencies?

awsclilinter/setup.py
hssyoo, line 1: Why do we have `setup.py` in addition to `pyproject.toml`?

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+def get_user_choice(prompt: str) -> str:
+    """Get user input for interactive mode."""
+    while True:
+        choice = input(prompt).lower().strip()
+        if choice in ["y", "n", "u", "q"]:
+            return choice
+        print("Invalid choice. Please enter y, n, u, or q.")
+
+
+def display_finding(finding: LintFinding, index: int, total: int, script_content: str):
```
hssyoo, line 30: Something that's confusing is that a line can show up as a diff even if nothing changed. Using the example file in this commit:
```
-aws secretsmanager put-secret-value --secret-id secret1213 \
-  --secret-binary file://data.json
+aws secretsmanager put-secret-value --secret-id secret1213 \
+  --secret-binary file://data.json --cli-binary-format raw-in-base64-out
```

Is this intended behavior?

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+    """Display a finding to the user with context."""
+    src_lines = script_content.splitlines()
+    dest_lines = finding.edit.inserted_text.splitlines()
+    start_line = finding.line_start
+    end_line = finding.line_end
+    src_lines_removed = end_line - start_line + 1
+    new_lines_added = len(dest_lines)
```
hssyoo: If I'm understanding everything correctly, then `src_lines_removed` should always be equal to `new_lines_added`, right? If so, should we add validation on that to guard against potentially corrupting the source file?

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Create a map from line numbers to their indices within the full script file
+    line_positions = []
+    pos = 0
+    for i, line in enumerate(src_lines):
+        line_positions.append((pos, pos + len(line)))
+        pos += len(line) + 1
```
hssyoo: Why do we need to map the entire script file instead of just the affected lines + context lines? If we want to map the entire file, can we just do it once instead of per finding?

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+        choice = get_user_choice("\nApply this fix? [y]es, [n]o, [u]pdate all, [q]uit: ")
+
+        if choice == "y":
+            accepted.append(finding)
+        elif choice == "u":
+            accepted.extend(findings[i - 1 :])
+            break
+        elif choice == "q":
+            print("Cancelled.")
+            sys.exit(0)
```
hssyoo: It might be a bit confusing if a user accepts a few fixes and then presses `q` to exit early with the expectation that the accepted fixes would be applied.
- We should explicitly document this behavior.
- [non-blocking suggestion] And also consider adding a [s]ave option that "flushes" accepted fixes and applies them. If you think we should do this, okay with doing it in a followup PR.

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+            print(f"{RED}-{line}{RESET}")
+
+            if i == end_line:
+                line_start_pos, _ = line_positions[i]
+                start_pos_in_line = max(0, finding.edit.start_pos - line_start_pos)
+                end_pos_in_line = min(len(line), finding.edit.end_pos - line_start_pos)
+                new_line = line[:start_pos_in_line] + finding.edit.inserted_text + line[end_pos_in_line:]
+                # In case the inserted text takes up multiple lines,
+                # inject a + at the start of each line.
+                new_line = new_line.replace("\n", "\n+")
```
hssyoo: When would this be the case?

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if start_line <= i <= end_line:
+            # This line is being modified
+            print(f"{RED}-{line}{RESET}")
+
+            if i == end_line:
+                line_start_pos, _ = line_positions[i]
+                start_pos_in_line = max(0, finding.edit.start_pos - line_start_pos)
+                end_pos_in_line = min(len(line), finding.edit.end_pos - line_start_pos)
+                new_line = line[:start_pos_in_line] + finding.edit.inserted_text + line[end_pos_in_line:]
```
hssyoo: I'm a little confused on the purpose of character positions in this function. Is there a reason we can't just replace line by line? Asking since we know the start/end lines of a finding and it looks like we already split lines on the finding edit.

hssyoo (commented):
It's looking good, I like the general experience. Left a few questions and suggestions.

awsclilinter/awsclilinter/cli.py
```diff
@@ -0,0 +1,162 @@
...
+        choice = input(prompt).lower().strip()
+        if choice in ["y", "n", "u", "q"]:
+            return choice
+        print("Invalid choice. Please enter y, n, u, or q.")
+
+
+def display_finding(finding: LintFinding, index: int, total: int, script_content: str):
+    """Display a finding to the user with context."""
+    src_lines = script_content.splitlines()
+    dest_lines = finding.edit.inserted_text.splitlines()
```
hssyoo: To calculate `new_lines_added`, is there a reason we're not using `LintFinding.line_start` and `LintFinding.line_end`?

awsclilinter/Makefile
aemous: That's fair, I'll look into `uv` and `ruff`, and potentially delete this Makefile and update the README dev guide

awsclilinter/README.md
aemous: Ah, so this was moreso intended for the devs contributing to this package. I agree most customers would be installing it from PyPi. I can try to make that more clear in the next revision with adding clarifications to this README, let me know if you have additional asks for this

awsclilinter/README.md
aemous: Similar to my last comment, this workflow was intended for dev contributors (i.e. us). The dev dependencies are for PyTest and format packages to test and format the package.

awsclilinter/setup.py
aemous: Good point, after some research it seems we only really need one in our case, and that `pyproject.toml` is the favored way. I will delete `setup.py` in the next revision

awsclilinter/awsclilinter/cli.py
aemous: Good point. I tried to keep the diff-generation code simple while also giving a somewhat familiar experience to `git diff`, and I think this behavior is an artifact of that.

We can pursue more complex diffing logic to address this. If we go down this route, I'd likely utilize `difflib` which is included in the standard Python library. It has a [unified_diff](https://docs.python.org/3/library/difflib.html) function which may produce what we're looking for. 

Below is some simple code I tested locally, seems to show the desired behavior:

```
>>> src_lines = 'aws secretsmanager put-secret-value --secret-id secret1213 \\\n  --secret-binary file://data.json\n'.splitlines(keepends=True)
>>> dest_lines = 'aws secretsmanager put-secret-value --secret-id secret1213 \\\n  --secret-binary file://data.json --cli-binary-format raw-in-base64-out\n'.splitlines(keepends=True)
>>> diff2 = difflib.unified_diff(src_lines, dest_lines)
>>> sys.stdout.writelines(diff2)
--- 
+++ 
@@ -1,2 +1,2 @@
 aws secretsmanager put-secret-value --secret-id secret1213 \
-  --secret-binary file://data.json
+  --secret-binary file://data.json --cli-binary-format raw-in-base64-out
```

I'm leaning towards going this route, please share feedback on this. It would likely simplify some of our diff-generation-code.

awsclilinter/awsclilinter/cli.py
aemous: As of now, yes I expect them to be equal. By calculating it like this it does not lock us in to that as a requirement.

I can imagine down the line us suggesting a 'fix' that takes up multiple lines, in that case they would not be equal. However, we currently have no plans for any of the linting rules to suggest fixes that take up multiple lines. It'll pretty much always be adding a flag to the last line or modifying the command or argument.

I can add this assertion, and there's always the option to remove it if we ever have suggestions for adding multi-line suggestions. Are you recommending it's added as a runtime check? Or introducing a test for this? 

I'm leaning towards a runtime check, and raising `RuntimeError` if they're not equal, with comments explaining some things in this thread. That'll be my plan for next revision unless you share disagreements

awsclilinter/awsclilinter/cli.py
aemous: That makes sense, we can just map the affected lines + context lines. 

We'd likely still need the for loop that goes through the entire file up until all the context lines are mapped so that we can count up the line lengths, then we can break from the loop. 

The reason for doing this per-finding instead of once is based on behavior when we have more than 1 linting rule. Let's say we have two linting rules A and B. Let's say we lint the file for rule A first, and then B. If we apply the fix from rule A, then the "intermediate script file" will be updated with the suggested fix, which updates the file indices. Therefore, the indices for rule B become stale and need to be recomputed after A is applied, and we'll need to recalculate this map to get the new diff based on the updated "intermediate script file."

Let me know if this is clear. There's optimizations that can be done, but they get relatively complex, and since we'll only have at most ~10 linting rules, this is all technically constant-time anyways and I didn't want to write something more complex as a pre-optimization.

awsclilinter/awsclilinter/cli.py
aemous: This makes sense. Now that I think about this, I think we can let `q` be the save and exit option that you're suggesting. Users can always use Ctrl+C to terminate the program immediately.

Let me know if you agree with this. In either case, I would like to do it in a followup PR though.

awsclilinter/awsclilinter/cli.py
aemous: This is currently the case for all commands that take up multiple lines. For example, the example you gave in a previous comment:

```
+aws secretsmanager put-secret-value --secret-id secret1213 \
+  --secret-binary file://data.json --cli-binary-format raw-in-base64-out
```

The code you're commenting on here is how a + was added to the start of the second line.

If we use `difflib` instead (discussed in previous reply), this logic could be removed and handled by `difflib` anyways.

awsclilinter/awsclilinter/cli.py
aemous: The reason is because the `ast-grep` library returns the `Edit` objects based on script indices rather than line-by-line. 

For example, you may have a CLI command in an if-statement:

```
if aws s3api head-object --bucket your-bucket-name --key path/to/your/object.txt >/dev/null 2>&1; then
    echo "Object exists."
fi
```

`ast-grep` library will give us an `Edit` object, with the start and end index of the new AWS CLI command. The `Edit` object has 3 things:

1. New text: `aws s3api head-object --bucket your-bucket-name --key path/to/your/object.txt --cli-binary-format raw-in-base64-out`
2. Start pos: the start index of the CLI command (3 in this case, due to `if` on the same line)
3. End pos: the end index of the CLI command (83 in this case, does not include the `>/dev/null 2>&1; then` on the same line)

Following the code in the snippet:

* `line[:start_pos_in_line]` will be 'if '
* `finding.edit.inserted_text` will be `aws s3api head-object --bucket your-bucket-name --key path/to/your/object.txt --cli-binary-format raw-in-base64-out`
* `line[end_pos_in_line:]` will be ` >/dev/null 2>&1; then`

So, this all lets us show the full lines in the diff, rather than just the CLI command.

awsclilinter/awsclilinter/cli.py
aemous: I think this relates to a previous reply in another comment. I left open the possibility that there could be more "new lines" than "source lines". `line_start` and `line_end` are the line numbers in the source file. 

If we end up deciding to be more rigid and force that the number of "new lines" equals the number of "source lines" with an assertion, we may be able use the variables you suggested instead.

With this information, please advise on if you want me to proceed with the more rigid approach, with the assertion that source == new lines.

awsclilinter/awsclilinter/cli.py
hssyoo: I think that looks great for our use case

awsclilinter/awsclilinter/cli.py
hssyoo: Yeah I think we should just stick to line-by-line changes (ie always 1 line changed per 1 line affected). Go ahead and add the runtime check

awsclilinter/awsclilinter/cli.py
hssyoo: Oh the multi-linting rule case makes sense. I'm okay with not complicating this any further

awsclilinter/awsclilinter/cli.py
hssyoo: Hmm let's discuss this with the team. I can see advantages to different approaches here

awsclilinter/awsclilinter/cli.py
hssyoo: Echoing my other reply: Yeah let's go ahead and make the assertion

awsclilinter/README.md
```diff
@@ -0,0 +1,122 @@
...
+or for your user:
+
+```shell
+$ python3 -m pip install --user awsclilinter
+```
+
+If you have the `awsclilinter` package installed and want to upgrade to the latest version, you can run:
+
+```shell
+$ python3 -m pip install --upgrade awscli
```
hssyoo: Think this should be `--upgrade awsclilinter`

awsclilinter/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+]
+dependencies = [
+    "ast-grep-py>=0.39.6",
+]
+
+[project.optional-dependencies]
+dev = [
+    "pytest>=8.0.0",
+    "ruff>=0.8.0",
+]
```
hssyoo, line 35: We should set ceilings on these dependencies

hssyoo (commented):
Just a couple small comments. Also:
- Did we add the runtime check / still plan to?
- I think we should still have a [q]uit or [a]bort option so users can exit without needing to Ctrl+C. We don't need to block this PR on it though.

awsclilinter/README.md
aemous: Good catch! Revising

hssyoo (approved):
🏆

PR #9810, opened 2025-10-22
Track usage of --v2-debug usage and final --v2-debug changes

aemous:
*Description of changes:*

* Usage of the new migration debug flag is tracked in the User-Agent header.
* Add support for the pager output breaking change, with resolution determined by setting `cli_pager` config to the empty string.
* An environment variable exists as an alternative to the `--v2-debug` command line option
* A --no-v2-debug CLI option exists (boolean switch).
* The text "AWS CLI v2 MIGRATION WARNING" is updated to "AWS CLI v2 UPGRADE WARNING"
* Updated scalarparse runtime detection to only print warning if the user did not configure any setting for `cli_timestamp_format`.
* Replaced cloudformation deploy's warning's reliance on the service response: output it even if the changeset is non-empty.
* Updated warnings to give more guidance, especially retaining v1 behavior on v2
* Updated the binary-params-base64 use case. It does not need to be prefixed with `file://`. 
* Replace use of uni print outside of customizations with print.

*Description of tests:*

* Ran and passed all unit and functional tests.

### DIFF
--- a/.changes/next-release/enhancement-Migration-48043.json
+++ b/.changes/next-release/enhancement-Migration-48043.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "Migration",
-  "description": "Implement a ``--migrate-v2`` flag that detects breaking changes for AWS CLI v2 for entered commands."
+  "description": "Implement a ``--v2-debug`` flag that detects breaking changes for AWS CLI v2 for entered commands."
 }
--- a/.changes/next-release/enhancement-Migration-48043.json
+++ b/.changes/next-release/enhancement-Migration-48043.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "Migration",
-  "description": "Implement a ``--v2-debug`` flag that detects breaking changes for AWS CLI v2 for entered commands."
+  "description": "Implement a ``--v2-debug`` flag and ``AWS_CLI_UPGRADE_DEBUG_MODE`` environment variable that detects breaking changes for AWS CLI v2 for entered commands."
 }
--- a/awscli/customizations/globalargs.py
+++ b/awscli/customizations/globalargs.py
@@ -106,6 +106,20 @@ def detect_migration_breakage(parsed_args, remaining_args, session, **kwargs):
     ]
     region = parsed_args.region or session.get_config_variable('region')
     s3_config = session.get_config_variable('s3')
+    if (
+            not session.get_scoped_config().get('cli_pager', None)
+            == '' and 'AWS_PAGER' not in os.environ
+    ):
+        uni_print(
+            'AWS CLI v2 UPGRADE WARNING: By default, the AWS CLI version 2 '
+            'returns all output through your operating system’s default pager '
+            'program. To retain AWS CLI v1 behavior, set the `cli_pager` '
+            'configuration setting, or the `AWS_PAGER` environment variable, '
+            'to the empty string. See https://docs.aws.amazon.com/cli/latest/'
+            'userguide/cliv2-migration-changes.html'
+            '#cliv2-migration-output-pager.\n',
+            out_file=sys.stderr
+        )
     if 'PYTHONUTF8' in os.environ or 'PYTHONIOENCODING' in os.environ:
         if 'AWS_CLI_FILE_ENCODING' not in os.environ:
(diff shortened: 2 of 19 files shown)


.changes/next-release/enhancement-Migration-48043.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "Migration",
-  "description": "Implement a ``--migrate-v2`` flag that detects breaking changes for AWS CLI v2 for entered commands."
+  "description": "Implement a ``--v2-debug`` flag that detects breaking changes for AWS CLI v2 for entered commands."
```
ashovlin: Add the env var here too.

awscli/data/cli.json
```diff
@@ -69,6 +69,11 @@
             "action": "store_true",
             "dest": "v2_debug",
             "help": "<p>Enable AWS CLI v2 migration assistance. Prints warnings if the command would face a breaking change after swapping AWS CLI v1 for AWS CLI v2 in the current environment. Prints one warning for each breaking change detected.</p>"
+        },
+        "no-v2-debug": {
+            "action": "store_true",
+            "dest": "no_v2_debug",
+            "help": "<p>Disable AWS CLI v2 migration assistance.</p>"
```
ashovlin: This is for if you've set the environment variable, but want to opt-out for a single invocation? 

If so, I don't think we need that at launch, can add it if folks ask for it. We have the slightly more verbose `AWS_CLI_UPGRADE_DEBUG_MODE=false aws s3 ls` as a workaround sooner.

awscli/customizations/globalargs.py
```diff
@@ -151,18 +169,18 @@ def detect_migration_breakage(parsed_args, remaining_args, session, **kwargs):
...
-            'cli/latest/userguide/cliv2-migration-changes.html'
-            '#cliv2-migration-ecr-get-login.\n',
+            'AWS CLI v2 UPGRADE WARNING: The `ecr get-login` command has '
+            'been removed in AWS CLI v2. You must use `ecr get-login-password` '
+            'instead. See https://docs.aws.amazon.com/cli/latest/userguide/'
+            'cliv2-migration-changes.html#cliv2-migration-ecr-get-login.\n',
             out_file=sys.stderr
         )
-    if url_params and session.full_config.get('cli_follow_urlparam', True):
+    if url_params and session.get_scoped_config().get('cli_pager', None):
```
ashovlin: Should this still be `cli_follow_urlparam`?

awscli/customizations/globalargs.py
ashovlin: I'm not seeing the warning printed for `aws s3api list-directory-buckets --region us-east-1 --v2-debug` with profile

```

[default]
region = us-east-1
cli_history=enabled
retry_mode=standard
output = json
```

ashovlin (changes requested):

awscli/data/cli.json
aemous: Yes, that was the intended use, which is consistent with some other global args that have env vars, such as auto prompt mode and pager.

I'm not against removing it, will do this in next revision.

awscli/customizations/globalargs.py
aemous: Good catch! The config we are checking should be updated to scoped config, but yes it should still be `cli_follow_urlparam`, looks like a copy-paste error.

Will revise.

awscli/customizations/globalargs.py
aemous: Good catch! Looks like I likely tested with high-level `s3` commands for this particular breaking change, though it should apply to the modeled `s3api` commands too. 

Will revise.

aemous: I addressed this, and also added 2 regression tests to test this (for s3 and s3api).

aemous:
An update is needed to detecting URL parameter usage (parameters starting with `http://` or `https://`). The current solution has false positives. There is a blocklist of parameters that do not use the follow param feature, and any `CLIArgument` with attribute `follow_urlparam = False` do not use the feature. We do not check these two conditions in this PR currently.

To avoid bloating this PR with new changes, I'll follow up with this change in a future PR.

ashovlin (approved):

RatStar811 (commented):
Issue

PR #9816, opened 2025-10-27
updating customizations/utility

varungovinda:
*Issue #, if available:*

*Description of changes:*  This PR updates the `get_policy_arn_suffix` function to dynamically load AWS partition information using botocore's `RuleSetStandardLibary` and `partitions.json` instead of hardcoded region mappings.

**Affected commands:**
- `aws dlm create-default-role`
- `aws emr create-default-roles` 
- `aws emr-containers update-role-trust-policy`

### DIFF
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -17,9 +17,9 @@ Utility functions to make it easier to work with customizations.
 import copy
 import sys
 
-from botocore.exceptions import ClientError, UnknownRegionError
+from botocore.exceptions import ClientError
 from botocore.loaders import Loader
-from botocore.regions import EndpointResolver
+from botocore.endpoint_provider import RuleSetStandardLibary
 from awscli.utils import create_nested_client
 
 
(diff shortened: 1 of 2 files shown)


awscli/customizations/utils.py
```diff
@@ -222,9 +224,12 @@ def uni_print(statement, out_file=None):
...
     region_string = region.lower()
-    if region_string.startswith("cn-"):
-        return "aws-cn"
-    elif region_string.startswith("us-gov"):
-        return "aws-us-gov"
-    else:
-        return "aws"
+    loader = Loader()
+    endpoints_data = loader.load_data('endpoints')
+    resolver = EndpointResolver(endpoints_data)
```
AndrewAsseily: I've been advised that we should prioritize using partitions.json over endpoints.json when possible rather than manually parsing partitions.json

AndrewAsseily (commented):
Thanks for this PR! We'll likely need to remove the forward-compatible test for `cn-northwest-2` unless we reintroduce the old string prefix fallback behavior, since aws_partition should handle this but the current partitions datadoes not include that specific test case.

PR #9817, opened 2025-10-27
updating customizations/utils

varungovinda:
*Description of changes:* This PR updates the `get_policy_arn_suffix` function to dynamically load AWS partition information using botocore's `RuleSetStandardLibary` and `partitions.json` instead of hardcoded region mappings.

**Affected commands:**
- `aws dlm create-default-role`
- `aws emr create-default-roles` 
- `aws emr-containers update-role-trust-policy`
- `emr-containers create-role-associations`

### DIFF
--- a/awscli/customizations/utils.py
+++ b/awscli/customizations/utils.py
@@ -24,7 +24,6 @@ from botocore.exceptions import ClientError
 from botocore.loaders import Loader
 from botocore.endpoint_provider import RuleSetStandardLibary
 
-
 from awscli.customizations.exceptions import ParamValidationError
 
 _SENTENCE_DELIMETERS_REGEX = re.compile(r'[.:]+')


awscli/customizations/utils.py
```diff
@@ -208,12 +210,15 @@ def uni_print(statement, out_file=None):
...
     region_string = region.lower()
-    if region_string.startswith("cn-"):
-        return "aws-cn"
-    elif region_string.startswith("us-gov"):
-        return "aws-us-gov"
-    else:
-        return "aws"
+    loader = Loader()
+    endpoints_data = loader.load_data('endpoints')
+    resolver = EndpointResolver(endpoints_data)
```
AndrewAsseily: I've been advised that we should prioritize using partitions.json over endpoints.json when possible. Rather than manually parsing partitions.json, we can leverage the existing aws_partition function from `RuleSetStandardLibrary` while still maintaining 'aws' as the default fallback:

```python
from botocore.endpoint_provider import RuleSetStandardLibary

loader = Loader()
partitions_data = loader.load_data('partitions')
rule_lib = RuleSetStandardLibary(partitions_data)
partition_result = rule_lib.aws_partition(region.lower())
return partition_result['name']

**Review by AndrewAsseily — COMMENTED** [review]
Thanks for this PR! We'll likely need to remove the forward-compatible test for `cn-northwest-2` unless we reintroduce the old string prefix fallback behavior, since aws_partition should handle this but the current partitions data does not include that specific test case.

File: .changes/next-release/bugfix-DLMEMR-99146.json [hunk 2]
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "DLM, EMR",
```
aemous, line 3: nit: we don't have existing changelog entries in this format with two entities separated by a comma. may be worth considering using a single more generic category that captures both.

PR #9823, opened 2025-10-29
Update model-validation tests 

aemous:
*Description of changes:*

* For each of our model-validation tests, record the service, operation, and optionally the shape, involved in the violation. Also, add the operation name to test errors where they were previously missing. 

*Description of tests:*

* Ran all tests and CI workflows successfully.


tests/functional/botocore/test_paginator_config.py
```diff
@@ -150,36 +150,39 @@ def _pagination_configs():
...
 
 
-def _validate_result_key_exists(page_config):
+def _validate_result_key_exists(operation_name, page_config):
     if 'result_key' not in page_config:
         raise AssertionError(
             "Required key 'result_key' is missing "
-            f"from pagination config: {page_config}"
+            f"from pagination config: {page_config}\n"
+            f"Target={operation_name}.result_key"
```
SamRemis: What is the word "Target" doing here and in all of the below examples?   This is functional, but it doesn't seem relevant to the public facing tests.

aemous: Resolve in latest revision. We utilize custom PyTest properties in the latest revision.

tests/functional/test_shadowing.py
```diff
@@ -58,13 +61,21 @@ def test_no_shadowed_builtins(command_name, command_table, builtins):
     errors = []
     for sub_name, sub_command in command_table.items():
         op_help = sub_command.create_help_command()
+        model = op_help.obj
         arg_table = op_help.arg_table
         for arg_name in arg_table:
             if any(p.startswith(arg_name) for p in builtins):
+                if isinstance(model, OperationModel):
```
ashovlin, line 68: When won't it be an `OperationModel`?

aemous: If I understand correctly, I think it always should be, though I did not do a thorough investigation to confirm this. This if statement is mainly for (1) static type checkers and (2) defensive programming.

For (1), my IDE knows `service_model` and `name` are valid attributes of `model` within the `if` statement.
For (2), it prevents breaking tests due to this change, which is for our own internal usage. If any validates-models test fails and does not output a service or operation field, we'll fail the build anyways internally.

ashovlin (approved):

PR #9835, opened 2025-11-06
fix: add validation for restricted service identifiers

quinnypig:
This change adds validation to prevent access attempts to restricted
AWS service identifiers that are not available for general use.

When users attempt to access restricted services like 'chaos2', they
will receive a clear error message indicating the service does not
exist in their available service catalog.

This improves the user experience by providing explicit feedback
for service identifiers that are reserved for internal AWS operations.

## Changes
- Modified CLIArgParser._check_value() to detect restricted services
- Added custom error messaging for clarity
- Added unit and functional tests for coverage

## Testing
- Unit tests verify error message formatting
- Functional tests confirm end-to-end behavior
- Existing tests pass without modification

## Verification
\`\`\`bash
$ aws chaos2 help
aws: error: argument command: service 'chaos2' does not exist

(But thank you for checking. Your curiosity has been logged.)
\`\`\`

All tests passing:
- 2/2 unit tests (chaos2 validation)
- 2/2 integration tests (end-to-end)
- 58/58 total unit tests (no regressions)

### DIFF
--- a/tests/unit/test_clidriver.py
+++ b/tests/unit/test_clidriver.py
@@ -731,6 +731,7 @@ class TestAWSCommand(BaseAWSCommandParamsTest):
         self.assertNotIn("Your curiosity has been logged",
                          stderr_output)
 
+
 class TestHowClientIsCreated(BaseAWSCommandParamsTest):
     def setUp(self):
         super(TestHowClientIsCreated, self).setUp()


hssyoo:
Hi @quinnypig, the AWS CLI already generally handles invalid inputs (in this case an unavailable service name). I'm curious why we'd want to hardcode `chaos2` as a special case.
1. This would prevent someone who _does_ have a local `chaos2` model available from using it in the AWS CLI. I'm not familiar with `chaos2` but if it's a custom model someone is currently using, this commit will prevent them from doing so.
2. I'm not familiar with "restricted services", but as long as service models aren't locally available, there's never any danger of someone unintentionally calling them. The AWS CLI does not ship with a service model named `chaos2`, so there's no risk of this.

If there's a specific issue this PR is trying to address, we'd be happy to discuss potential solutions in an open issue.

Issue #9479, opened 2025-05-06
Error message for `aws s3 ls` refers to wrong endpoint.


wonko:
### Describe the bug

Given an s3 bucket in `eu-west-1`, when setting `AWS_REGION=eu-west-1` and `AWS_ENDPOINT_URL=https://s3.amazonaws.com` env vars and issuing an `aws s3 ls s3://bucketname/some/path/to/object`, it will give you the following error message:

```
An error occurred (PermanentRedirect) when calling the ListObjectsV2 operation: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint: bucketname.s3.amazonaws.com
You can fix this issue by explicitly providing the correct region location using the --region argument, the AWS_DEFAULT_REGION environment variable, or the region variable in the AWS CLI configuration file.  You can get the bucket's location by running "aws s3api get-bucket-location --bucket BUCKET".
```

It is right that it gives the error, as the endpoint should either be unset, or set to the region-specific `s3.eu-west-1.amazonaws.com` endpoint. The endpoint which is suggested in the error message is however wrong, and will result in NotFound errors.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The error message should either give a link to how/where to find the correct endpoint, or it should give the correct endpoint. Giving a non-functional endpoint is a bug to me.

### Current Behavior

See the output in the above snippet

### Reproduction Steps

Create a new bucket in non us-east-1, make sure you have access to it, and then:

```
bash-5.1$ set | grep AWS
AWS_ENDPOINT_URL=https://s3.amazonaws.com
AWS_REGION=eu-west-1
AWS_SHARED_CREDENTIALS_FILE=/tmp/aws-credfile
_=AWS_ENDPOINT_URL=https://s3.amazonaws.com
bash-5.1$ aws s3 ls s3://[BUCKET-REDACTED]/some/path/ --summarize

An error occurred (PermanentRedirect) when calling the ListObjectsV2 operation: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint: [BUCKET-REDACTED].s3.amazonaws.com
You can fix this issue by explicitly providing the correct region location using the --region argument, the AWS_DEFAULT_REGION environment variable, or the region variable in the AWS CLI configuration file.  You can get the bucket's location by running "aws s3api get-bucket-location --bucket BUCKET".
[...]

PR #9850, opened 2025-11-17
[Documentation] Warning against the use of S3 Global Endpoint (s3.amazonaws.com) as endpoint-url

adev-code:
## Summary
Adds documentation improvements to [S3 configuration documentation](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html) (in the AWS CLI) for the --endpoint-url parameter:

1. Warning against S3 Global endpoint: Advises users to avoid using `https://s3.amazonaws.com` or `http://s3.amazonaws.com` as the --endpoint-url parameter value, as it can cause PermanentRedirect errors and configuration confusion.

2. General endpoint_url guidance: Adds a note explaining the --endpoint-url parameter usage, including caution about S3 unintended behavior such as S3 redirect issues. References the AWS SDK reference guide for additional endpoint configuration information.

## Testing
• [x] Documentation builds successfully with make html
• [x] Warning and note display correctly in generated docs

Screenshot 2025-11-28 at 9 40 03 AM

Addresses #9479

### DIFF
--- a/awscli/topics/s3-config.rst
+++ b/awscli/topics/s3-config.rst
@@ -18,7 +18,7 @@ for performance reasons or to account for the specific environment where these
 .. note::
    S3 commands have an option to use a custom endpoint using ``--endpoint-url``.
    This overrides the default endpoint the command will use.
-   Use caution when configuring this parameter as it can cause S3 unintended behavior including S3 redirect issues.
+   Use caution when configuring this parameter as it can cause unintended behavior including S3 redirect issues.
    See `Service-specific endpoints <https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html>`_ page in the *AWS SDK reference guide* for more information.
 
 Configuration Values


awscli/topics/s3-config.rst
```diff
@@ -15,6 +15,11 @@ are provided in the case where you need to modify one of these values, either
 for performance reasons or to account for the specific environment where these
 ``aws s3`` commands are being run.
 
+.. warning::
```
kdaily: I would move this into the "Configuration Values" section as a general comment, not a warning.

awscli/topics/s3-config.rst
```diff
@@ -15,6 +15,11 @@ are provided in the case where you need to modify one of these values, either
...
+   S3 commands have a parameter ``--endpoint-url`` that overrides the command's
+   default URL with the given URL. Avoid using the S3 Global Endpoint
```
kdaily: I think we could be a bit more general here and just say you should avoid configuring the endpoint URL at all when using the AWS S3 service? Generally endpoint URL is only going to be needed for 3P endpoints when using S3.

kdaily:
This might be a good place to point customers to:

https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html

awscli/topics/s3-config.rst
```diff
@@ -15,6 +15,11 @@ are provided in the case where you need to modify one of these values, either
 for performance reasons or to account for the specific environment where these
 ``aws s3`` commands are being run.
 
+.. note::
+   S3 commands have an option to use a custom endpoint using ``--endpoint-url``.
+   This overrides the default endpoint the command will use.
+   Use caution when configuring this parameter as it can cause S3 unintended behavior including S3 redirect issues.
```
kdaily:
```suggestion
   Use caution when configuring this parameter as it can cause unintended behavior including S3 redirect issues.
```

awscli/topics/s3-config.rst
adev-code: Ok thanks, made it as a note

awscli/topics/s3-config.rst
adev-code: Ok thanks, made it look general rather than discussing about the global endpoint. As other endpoints may also lead to unintended behaviors, not just the global endpoint.

kdaily (approved):
LGTM, once CI tests pass. Please squash commits before merging, thanks!

PR #9859, opened 2025-11-20
Support multiple linting rules

aemous:
*Description of changes:*

* Updated interactive and auto-fix mode to support multiple linting rules. The main technical complexity with supporting multiple rules is that as we apply fixes to the script, older linter findings become stale and would need to be 
'refreshed'. We avoid this problem by linting the script one-rule-at-a-time.
* Updated `Base64BinaryFormatRule` so that it does not require `file://` prefix in an argument value to trigger a detection. It now flags all AWS CLI commands.
* Added `PaginationRule` that flags all AWS CLI commands, and adds `--no-cli-paginate` as a fix.
* Replaced `ScriptLinter` class with utility functions. Updated their interface to let callers pass in their own AST.
* Added save and exit mode.
* Added tests for save-and-exit mode and the new `PaginationRule`.
* Misc. refactoring for maintainability/readability.

*Description of tests:*

* Ran and passed all tests.
* Manually tested the tool in interactive and auto-fix mode on the example script, and verified correct behavior.

### DIFF
--- a/awsclilinter/awsclilinter/cli.py
+++ b/awsclilinter/awsclilinter/cli.py
@@ -26,7 +26,9 @@ def prompt_user_choice_interactive_mode() -> str:
     """Get user input for interactive mode."""
     while True:
         choice = (
-            input("\nApply this fix? [y]es, [n]o, [u]pdate all, [s]ave and exit, [q]uit: ")
+            input(
+                "\nApply this fix? [y] yes, [n] no, [u] update all, [s] save and exit, [q] quit: "
+            )
             .lower()
             .strip()
         )
@@ -35,7 +37,7 @@ def prompt_user_choice_interactive_mode() -> str:
         print("Invalid choice. Please enter y, n, u, s, or q.")
 
 
-def display_finding(finding: LintFinding, index: int, total: int, script_content: str):
+def display_finding(finding: LintFinding, index: int, script_content: str):
     """Display a finding to the user with context."""
     src_lines = script_content.splitlines(keepends=True)
 
@@ -61,7 +63,7 @@ def display_finding(finding: LintFinding, index: int, total: int, script_content
             f"{len(src_context)} != {len(dest_context)}."
         )
 
-    print(f"\n[{index}/{total}] {finding.rule_name}")
+    print(f"\n[{index}] {finding.rule_name}")
     print(f"{finding.description}")
 
     diff = difflib.unified_diff(src_context, dest_context, lineterm="")
@@ -119,7 +121,6 @@ def interactive_mode_for_rule(
     findings: List[LintFinding],
     ast: SgRoot,
     finding_offset: int,
-    total_findings: int,
 ) -> Tuple[SgRoot, bool, Optional[str]]:
     """Run interactive mode for a single rule's findings.
 
@@ -127,7 +128,6 @@ def interactive_mode_for_rule(
         findings: List of findings for this rule.
         ast: Current script content, represented as an AST.
         finding_offset: Offset for display numbering.
-        total_findings: Total number of findings across all rules.
 
     Returns:
         Tuple of (ast, changes_made, last_choice)
@@ -139,7 +139,7 @@ def interactive_mode_for_rule(
     last_choice: Optional[str] = None
 
     for i, finding in enumerate(findings):
-        display_finding(finding, finding_offset + i + 1, total_findings, ast.root().text())
+        display_finding(finding, finding_offset + i + 1, ast.root().text())
         last_choice = prompt_user_choice_interactive_mode()
 
         if last_choice == "y":
@@ -158,7 +158,7 @@ def interactive_mode_for_rule(
                 ast = parse(linter.apply_fixes(ast, accepted_findings))
             return ast, len(accepted_findings) > 0, last_choice
         elif last_choice == "q":
-            print("Quit without saving.")
+            print("Quitting without saving.")
             sys.exit(0)
 
     if accepted_findings:
@@ -177,7 +177,7 @@ def main():
     parser.add_argument(
         "--fix", action="store_true", help="Apply fixes to the script (modifies in place)"
     )
-    parser.add_argument("--output", help="Output path for the fixed script")
+    parser.add_argument("--output", help="Output path for the modified script")
     parser.add_argument(
         "-i",
         "--interactive",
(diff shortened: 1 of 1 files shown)


aemous:
Just a note, the pagination linting rule should actually be based around the `--no-cli-pager` argument instead of the `--no-paginate` argument. This will be addressed in a follow-up PR, I already have the fix staged locally with the next round of changes.

awsclilinter/awsclilinter/cli.py
```diff
@@ -18,13 +22,17 @@
...
 
 
-def get_user_choice(prompt: str) -> str:
+def prompt_user_choice_interactive_mode() -> str:
     """Get user input for interactive mode."""
     while True:
-        choice = input(prompt).lower().strip()
-        if choice in ["y", "n", "u", "s"]:
+        choice = (
+            input("\nApply this fix? [y]es, [n]o, [u]pdate all, [s]ave and exit, [q]uit: ")
```
kdaily: For accessibility, we should make the letter to select separate from the word that describes. A screen reader would not be able to read this well. How about:


```suggestion
            input("\nApply this fix? [y] yes, [n] no, [u] update all, [s] save and exit, [q] quit: ")
```

awsclilinter/awsclilinter/rules/pagination_rule.py
```diff
@@ -0,0 +1,50 @@
+from typing import List
+
+from ast_grep_py.ast_grep_py import SgRoot
+
+from awsclilinter.rules import LintFinding, LintRule
+
+
+class PaginationRule(LintRule):
```
kdaily, line 8: Shouldn't this be `--no-cli-pager`?

https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html#cli-usage-pagination-clientside

awsclilinter/awsclilinter/rules/pagination_rule.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def name(self) -> str:
+        return "no-cli-paginate"
+
+    @property
+    def description(self) -> str:
+        return (
+            "AWS CLI v2 uses pagination by default for commands that return large result sets. "
+            "Add --no-cli-paginate to disable pagination and match v1 behavior."
```
kdaily, line 19: Users can also disable pagination via env variable or config file option, which they may be already doing. Would that make detection a false positive?. If so, how do we account for that?

awsclilinter/awsclilinter/cli.py
```diff
@@ -75,22 +83,89 @@ def display_finding(finding: LintFinding, index: int, total: int, script_content
...
+
+    Returns:
+        Modified script content
+    """
+    current_ast = ast
+
+    # Group findings by rule
+    findings_by_rule: Dict[str, List[LintFinding]] = {}
+    for finding, rule in findings_with_rules:
+        if rule.name not in findings_by_rule:
```
kdaily, line 109: nit: could use `collections.defaultdict` to reduce checking here, not a strong preference.

awsclilinter/awsclilinter/cli.py
```diff
@@ -75,22 +83,89 @@ def display_finding(finding: LintFinding, index: int, total: int, script_content
...
+            if accepted_findings:
+                ast = parse(linter.apply_fixes(ast, accepted_findings))
+            return ast, True, last_choice
+        elif last_choice == "s":
+            # Apply accepted findings and stop processing
+            if accepted_findings:
+                ast = parse(linter.apply_fixes(ast, accepted_findings))
+            return ast, len(accepted_findings) > 0, last_choice
+        elif last_choice == "q":
+            print("Quit without saving.")
```
kdaily: nit:


```suggestion
            print("Quitting without saving.")
```

awsclilinter/awsclilinter/cli.py
```diff
@@ -127,31 +202,80 @@ def main():
...
-    if not findings:
-        print("No issues found.")
-        return
+    rules = [Base64BinaryFormatRule(), PaginationRule()]
 
     if args.interactive:
-        findings = interactive_mode(findings, script_content)
-        if not findings:
+        # Do an initial parse-and-lint with all the rules simultaneously to compute the total
+        # number of findings for displaying progress in interactive mode.
```
kdaily: What's the value of displaying progress here? It seems a bit unnecessary to completely run through the entire set of possible changes only to do it again without just saving the diffs of what would have been changed and then accepting/denying them in the later loop. That would increase complexity of the solution, and may not matter for performance (scripts might not be that big).

kdaily: I would also be fine with a summary at the end that said how many changes could have been applied and how many were.

awsclilinter/awsclilinter/cli.py
```diff
@@ -127,31 +202,80 @@ def main():
...
+        if not findings_with_rules:
+            print("No issues found.")
+            return
+
+        # Process one rule at a time, re-parsing between rules
+        current_script = script_content
+        any_changes = False
+        finding_offset = 0
+
+        # Calculate total findings for display
```
kdaily: See comment above, but this should probably go above the comment "# Process one rule at a time" for clarity of where the logic starts.

awsclilinter/awsclilinter/cli.py
```diff
@@ -127,31 +202,80 @@ def main():
...
             return
 
-    if args.fix or args.output or args.interactive:
-        # Interactive mode is functionally equivalent to --fix, except the user
-        # can select a subset of the changes to apply.
-        fixed_content = linter.apply_fixes(script_content, findings)
         output_path = Path(args.output) if args.output else script_path
-        output_path.write_text(fixed_content)
+        output_path.write_text(current_script)
+        print(f"Fixed script written to: {output_path}")
```
kdaily: nit: 


```suggestion
        print(f"Modified script written to: {output_path}")
```

awsclilinter/awsclilinter/rules/pagination_rule.py
aemous: Good question. Since this is a static linter, we cannot resolve the user's config profile values, and if that behavior is desired, we can point users to use the `--v2-debug` runtime migration tool instead, which has that functionality. Being static, we should be processing these scripts independent of the environment: the only input is a script, and the only output is a script (possibly with a detections summary).

I generally agree that would be a false positive case. As a workaround, if a user knows they configured pagination to disable the pager, then they can reject the auto-fix in interactive mode by entering "n" when the fix is suggested. In any case, even if the auto-fix *is* made, the command-line option overrides the configuration in precedence order, and the behavior would not change (pager gets disabled anyways). So, the worst case scenario is that the flag is added even when it's unnecessary, *however*, it being unnecessary does not cause breakage.

awsclilinter/awsclilinter/cli.py
aemous: Good point, I was second-guessing this myself. I figured users may want to know how many findings are left especially if the script is large. However, that may not be super valuable, can always add it back if users request it.

Will remove this in next revision.

awsclilinter/awsclilinter/rules/pagination_rule.py
aemous: Yes, good catch. Discussed internally, this will be addressed in a future PR.

PR #9861, opened 2025-11-20
Fix start-live-tail tests

kdaily:
*Issue #, if available:*

*Description of changes:*

This fix addresses failures encountered on some systems when running the startlivetail unit tests through the AWS CLI test runner script at `scripts/ci/run-tests`. This change adds an optional `app_input` parameter so that the test can be run with `create_pipe_input()` to replace the input for unit testing.

According to prompt toolkit documentation:

> During the creation of a prompt_toolkit Application, we can specify
> what input and output device to be used. By default, these are output
> objects that correspond with sys.stdin and sys.stdout. In unit tests
> however, we want to replace these.
> - For the input, we want a “pipe input”. This is an input device,
> in which we can programmatically send some input. It can be created
> with create_pipe_input(), and that return either a PosixPipeInput or a
> Win32PipeInput depending on the platform.

Reference: https://python-prompt-toolkit.readthedocs.io/en/stable/pages/advanced_topics/unit_testing.html


kdaily:
Test failures are unrelated and will be mitigated by #9865.

kdaily:
Closing, replaced with #9877

PR #9873, opened 2025-11-24
[v2] Use AssertionError instead of unittest.TestCase.fail for S3 memory assertion integration tests

kdaily:
*Issue #, if available:*

*Description of changes:*

Integration tests were ported from CLI v1, but some were altered to not inherit from a `unittest.TestCase`, so it does not have a fail method. Tests using the `assert_max_memory_used` will fail incorrectly and not print the failure message. This change makes the test raise an `AssertionError` with the existing failure message.

### DIFF
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -1554,7 +1554,7 @@ class TestMemoryUtilization(BaseS3IntegrationTest):
                     peak_memory / 1024.0 / 1024.0,
                 )
             )
-            self.fail(failure_message)
+            raise AssertionError(failure_message)
 
     def test_transfer_single_large_file(self, files, shared_bucket):
         # 40MB will force a multipart upload.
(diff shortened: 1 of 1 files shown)


tests/integration/customizations/s3/test_plugin.py
aemous: non-blocking: would `assert peak_memory <= max_mem_allowed, failure_message` be more consistent/idiomatic with the rest of our codebase? This way it only fails when assertions are enabled

aemous (changes requested):
LGTM overall. one nit, open to discussion

tests/integration/customizations/s3/test_plugin.py
kdaily: I'm fine moving it to an idiomatic expression, but wouldn't raising an `AssertionError` and calling `assert` result in the same thing, and would fail identically if assertions are enabled?

kdaily: Addressed in edfe92aac046f0f532d40e3aa12803208cd90a81

aemous: Yeah to clarify my comment was more about being idiomatic. Secondarily if you run it with assertions *disabled* then the behavior is different, this may be a moot point since PyTest enables assertions.

aemous (approved):
LGTM

Issue #9864, opened 2025-11-21
ECR docker login does not work anymore with docker image amazon/aws-cli:2.32.2


b3n3d1k7:
### Describe the bug

We have been using the amazon/aws latest image in our pipelines to push docker images to private ECR repositories using authorization token retrieved by aws ecr get-login-password.

With amazon/aws-cli:2.32.2 the docker login call fails with the error message: "400 Bad Request"
Reverting to amazon/aws-cli:2.32.1 fixes the issue.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Docker authentication using the AWS CLI docker image amazon/aws-cli:2.32.2 works as described here: https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html#registry-auth-token

### Current Behavior

With amazon/aws-cli:2.32.2 the docker login call fails with the error message: "400 Bad Request"

### Reproduction Steps

Run the command
`aws ecr get-login-password --region region | docker login --username AWS --password-stdin aws_account_id.dkr.ecr.region.amazonaws.com`
with the docker image amazon/aws-cli:2.32.2

I.e.
```
docker run -t --env AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION}" --env AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}" --env AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" --env AWS_SESSION_TOKEN="${AWS_SESSION_TOKEN}" amazon/aws-cli:2.32.2 ecr get-login-password --region "${AWS_DEFAULT_REGION}" | docker login --username AWS --password-stdin "${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
```

### CLI version used

aws-cli/2.32.2

### Environment details (OS name and version, etc.)

ubuntu 24.04

wilsongee:
I am also seeing the same issue:

```
> make ecr_login
docker pull amazon/aws-cli:2.32.2
2.32.2: Pulling from amazon/aws-cli
Digest: sha256:1f78cc694527183e91e4a914bc4d426eeee16ae99eb718069c048a1f3af06612
Status: Image is up to date for amazon/aws-cli:2.32.2
docker.io/amazon/aws-cli:2.32.2
[...]

PR #9878, opened 2025-11-25
Fix Docker output by moving colorama init for ECS expressgateway commands

ashovlin:
*Issue #, if available:* #9864

*Description of changes:* The recent ECS expressgateway commands added a call to `colorama.init()` on import, which happens as we register the commands via the handlers. So this was always getting called, even if the user is using other AWS CLI commands.

This caused the linked issue, where if a user is piping command output _out of_ a Docker container with the `--tty` flag, there'd be a new reset control code (`^[[m`) in the output.

Now, this moves the `colorama.init` call into the no-longer-static ECS `ColorUtils`, so it will only be called if the user is using the new ECS commands.

**Note:** I tried to add a test that mimicked the Docker pseudo-TTY but couldn't get it working quickly. We have Docker smoke tests in the internal pipeline that we can add this to. For now, added a test that ensures colorama is not initialized outside of where we use it.

### DIFF
--- a/tests/functional/test_output.py
+++ b/tests/functional/test_output.py
@@ -31,7 +31,7 @@ class TestOutput(BaseAWSCommandParamsTest):
         self.patch_colorama_init = mock.patch('colorama.init')
         self.mock_colorama_init = self.patch_colorama_init.start()
 
-        super(__class__, self).setUp()
+        super().setUp()
         self.files = FileCreator()
 
         self.patch_popen = mock.patch('awscli.utils.Popen')
@@ -53,7 +53,7 @@ class TestOutput(BaseAWSCommandParamsTest):
         self.expected_content = self.get_expected_content(self.parsed_response)
 
     def tearDown(self):
-        super(__class__, self).tearDown()
+        super().tearDown()
         self.files.remove_all()
         self.patch_popen.stop()
         self.patch_tty.stop()
--- a/tests/unit/customizations/ecs/expressgateway/test_color_utils.py
+++ b/tests/unit/customizations/ecs/expressgateway/test_color_utils.py
@@ -10,65 +10,53 @@ def color_utils():
 
 def test_make_cyan(color_utils):
     result = color_utils.make_cyan("test text")
-    assert "test text" in result
-    # Should contain ANSI color codes for cyan
-    assert "\033[" in result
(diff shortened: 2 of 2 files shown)


tests/functional/test_output.py
```diff
@@ -49,7 +53,7 @@ def setUp(self):
         self.expected_content = self.get_expected_content(self.parsed_response)
 
     def tearDown(self):
-        super(TestOutput, self).tearDown()
+        super(__class__, self).tearDown()
```
nateprewitt: Are we doing something special here? I think `super().tearDown()` is all we need. Adding the class to [`super`](https://docs.python.org/3/library/functions.html#super) hasn't been required since Python 2. The only cases we should need to do this is if we're trying to invoke `super()` for a class that's not in our class hierarchy. We cleaned these up in the other codebases, but it might be a CLI v2 miss?

tests/unit/customizations/ecs/expressgateway/test_color_utils.py
```diff
@@ -1,140 +1,159 @@
...
+@pytest.fixture
+def color_utils():
+    return ColorUtils()
+
+
+def test_make_cyan(color_utils):
+    result = color_utils.make_cyan("test text")
+    assert "test text" in result
+    # Should contain ANSI color codes for cyan
+    assert "\033[" in result
```
nateprewitt: Are these cases where we should be asserting explicit equivalence since we control the input? I'm just wondering if we have future impact where other control characters may be sneaking in that we're not aware of.

tests/functional/test_output.py
ashovlin: Had just accepted ruff's suggestion, but agree, fixed in https://github.com/aws/aws-cli/pull/9878/commits/cd39388c433e38d73d2a966ded2c072df945f089

Will chat with Kenny about the proactive clean-up.

tests/unit/customizations/ecs/expressgateway/test_color_utils.py
ashovlin: Agreed, switched to exact assertions in https://github.com/aws/aws-cli/pull/9878/commits/cd39388c433e38d73d2a966ded2c072df945f089

nateprewitt (approved):

Issue #2688, opened 2017-06-28
Exception as JSON in aws client


andre2007:
While executing a command in the AWS client I specify JSON as output.
When an exception occurs, this exception is written to stdout as text instead of JSON. I would like to have the Exceptions also returned as JSON Objects like they are defined in the APIs: https://github.com/aws/aws-sdk-js/tree/master/apis

stealthycoin:
What do you plan to do with it? Would you want to be able to run `--query` on it?

andre2007:
Based on the Javascript api definitions located here https://github.com/aws/aws-sdk-js/tree/master/apis I created a SDK for the D Programming language. Instead of calling the rest services I use the AWS client to execute the requests. Thats working just fine.
If an exception occurs in a service request I want to throw the specific exception defined in the api definitions. If the aws client would return the exception as JSON object it would be much simpler.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

karolba:
Exceptions in `--output json` still aren't being output as json, so reopening

karolba:
Or.. not reopening, I don't have permissions to reopen issues :sweat_smile: The bot message above is misleading

ashovlin:
@karolba or others, are you looking to interact with the error details programmatically? Or just want to see additional information in the console as a human user?

[...]

PR #9890, opened 2025-12-02
Structured Error Output

AndrewAsseily:
## Issue #
CLI-5136 & CLI-7572 & https://github.com/aws/aws-cli/issues/2688

## Description of changes

Adds structured error formatting support for both AWS service errors and CLI exceptions. Users can now view errors in JSON, YAML, legacy, enhanced, text, or table formats using `--cli-error-format`.

**For service errors (ClientError):** Exposes modeled error members that AWS services return. For example, DynamoDB's `TransactionCanceledException` includes `CancellationReasons` with details about which operations failed.

**For CLI exceptions:** Provides structured output for errors like `NoCredentialsError`, `NoRegionError`, and `ParamValidationError`.

**New `--output off` format:** Suppresses all stdout output while preserving stderr

### DynamoDB TransactionCanceledException

Command:
```bash
aws dynamodb transact-write-items \
  --transact-items '[
    {
      "Update": {
        "TableName": "TestErrorTable",
        "Key": {"id": {"S": "existing-item"}},
        "UpdateExpression": "SET #s = :val",
        "ExpressionAttributeNames": {"#s": "status"},
        "ExpressionAttributeValues": {":val": {"S": "updated"}},
        "ConditionExpression": "attribute_not_exists(id)"
      }
    },
    {
      "Put": {
        "TableName": "TestErrorTable",
        "Item": {"id": {"S": "new-item"}}
      }
    }
  ]'
```

**Enhanced format (default):**
```

aws: [ERROR]: An error occurred (TransactionCanceledException) when calling the TransactWriteItems operation: Transaction cancelled, please refer cancellation reasons for specific reasons [ConditionalCheckFailed, None]

Additional error details:
CancellationReasons: <complex value> (Use --cli-error-format with json or yaml to see full details)
```

**JSON format (`--cli-error-format json`):**
```json
{
    "Message": "Transaction cancelled, please refer cancellation reasons for specific reasons [ConditionalCheckFailed, None]",
    "Code": "TransactionCanceledException",
    "CancellationReasons": [
        {
            "Code": "ConditionalCheckFailed",
            "Message": "The conditional request failed"
        },
        {
            "Code": "None"
        }
    ]
}
```

**YAML format (`--cli-error-format yaml`):**
```yaml
CancellationReasons:
- Code: ConditionalCheckFailed
  Message: The conditional request failed
- Code: None
Code: TransactionCanceledException
Message: Transaction cancelled, please refer cancellation reasons for specific reasons
  [ConditionalCheckFailed, None]
```

### Simple error with additional fields

Command:
```bash
aws s3api get-object --bucket not-a-real-bucket-0000 --key file.txt out.txt
```

**Enhanced format:**
```

aws: [ERROR]: An error occurred (NoSuchBucket) when calling the GetObject operation: The specified bucket does not exist

Additional error details:
BucketName: not-a-real-bucket-0000
```

**JSON format:**
```json
{
  "Code": "NoSuchBucket",
  "Message": "The specified bucket does not exist",
  "BucketName": "not-a-real-bucket-0000"
}
```

## Enhanced format behavior

The enhanced format handles different error shapes:

**Simple values:**
- Strings, numbers, booleans, null
- Small lists (≤5 simple items): `AllowedValues: [GET, PUT, POST, DELETE]`
- Small dicts (≤5 simple key-value pairs): `Metadata: {key1: value1, key2: value2}`

**Complex values:**
- Large collections (>5 items)
- Nested structures (lists of dicts, nested dicts)
- Mixed-type collections

When a field is complex, we show:
```
FieldName: <complex value>
(Use --cli-error-format with json or yaml to see full details)
```

## Configuration

**Config file (`~/.aws/config`):**
```ini
[default]
cli_error_format = json
```

**Environment variable:**
```bash
export AWS_CLI_ERROR_FORMAT=yaml
```

**Command-line flag:**
```bash
aws <command> --cli-error-format enhanced
```

**Valid formats:** `enhanced` (default), `json`, `yaml`, `text`, `table`, `legacy`

Use `legacy` to get the old behavior with no structured error details.

## Technical notes

AWS services return modeled error fields at the top level of the error response, not nested under the `Error` key. For example, DynamoDB returns `CancellationReasons` at `response['CancellationReasons']`, not `response['Error']['CancellationReasons']`. 

The implementation:
1. Copies all fields from `response['Error']` (Code, Message, and any service-specific fields there)
2. Merges in top-level fields from the response (excluding Error, ResponseMetadata, and duplicates)
3. Returns the combined error dict

## Tests
- All existing tests pass

## For reviewers

**MAX_INLINE_ITEMS threshold:**
Currently set to 5. Could be reduced to 4 for better terminal width compatibility, or we could add a character length limit (~70 chars) in addition to the item count. Open to feedback.

**Sensitive data:**
AWS service models include `@sensitive` traits for fields. Right now we display all error fields as returned by the service. Using the [Smithy CLI](https://smithy.io/2.0/guides/smithy-cli/index.html) to analyze AWS service models, I found 0 error response fields marked with `@sensitive` traits.

### DIFF
--- a/awscli/errorhandler.py
+++ b/awscli/errorhandler.py
@@ -123,7 +123,7 @@ def construct_cli_error_handlers_chain(session=None):
 
 
 class BaseExceptionHandler:
-    def handle_exception(self, exception, stdout, stderr):
+    def handle_exception(self, exception, stdout, stderr, **kwargs):
         raise NotImplementedError('handle_exception')
 
 


.changes/next-release/feature-Output-59989.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "Output",
+  "description": "AWS service errors containing modeled fields beyond Code and Message now write structured output to stdout in the configured output format. Set cli_error_format=LEGACY to disable or use --output off to suppress stdout."
+}
```
ashovlin, line 5: `--output off` is substantial, please add its own entry in a separate file.

awscli/examples/global_options.rst
```diff
@@ -29,6 +29,8 @@
   
   *   yaml-stream
   
+  *   off
```
ashovlin, line 32: While we're here, can you catch up the section in the API reference too? 
https://github.com/aws/aws-cli/blob/33dbfd2c85e9bb548ba1e043e0b4a3d818f022a7/awscli/topics/config-vars.rst?plain=1#L92-L96

awscli/structured_error.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2012-2013 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ashovlin: nit: Throughout, either drop years, or use the short version:
```
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0
```

awscli/examples/global_options.rst
ashovlin: Also add new information there on the new error format setting.

.changes/next-release/feature-Output-59989.json
ashovlin: Don't need to see changes, but in hindsight I'd rather have done `--output off` in a separate PR, these features aren't dependent on one another technically.

awscli/clidriver.py
```diff
@@ -1023,21 +1045,54 @@ def invoke(self, service_name, operation_name, parameters, parsed_globals):
...
 
         formatter = get_formatter(output, parsed_globals)
-        with self._output_stream_factory.get_output_stream() as stream:
-            formatter(command_name, response, stream)
+        try:
+            with self._output_stream_factory.get_output_stream() as stream:
+                formatter(command_name, response, stream)
+        except ClientError as e:
+            self._display_structured_error_for_exception(e, parsed_globals)
+            raise
```
ashovlin: Why do we need this block? Can the formatter raise a structured error that the actual API call wouldn't have?

awscli/clidriver.py
```diff
@@ -368,6 +373,20 @@ def _construct_cli_auto_prompt_chain(self):
...
+        providers = [
+            EnvironmentProvider(
+                name='AWS_CLI_ERROR_FORMAT',
+                env=os.environ,
+            ),
+            ScopedConfigProvider(
+                config_var_name='cli_error_format',
+                session=self.session,
+            ),
+            ConstantProvider(value='STANDARD'),
```
ashovlin: Few things on the enum values:
1. Left a comment elsewhere, but please start by documenting valid values in `global_options.rst`
2. Add validation like we do for `output` for invalid values.
3. Only `STANDARD | LEGACY` were never discussed as options, we should revisit with the team internally.

awscli/structured_error.py
```diff
@@ -0,0 +1,94 @@
...
+
+from botocore.exceptions import ClientError
+
+from awscli.formatter import get_formatter
+
+
+LOG = logging.getLogger('awscli.structured_error')
+
+
+class StructuredErrorHandler:
```
ashovlin: Can this be handled through the existing `ClientErrorHandler`? That already has an overridable method with access to stdout/stderr. 

We'd need less new plumbing in the driver, and we already should have confidence that it is catching `ClientError` instances.

I also wonder if that'll help apply more of this to custom commands.

awscli/structured_error.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return True
+
+    def display(self, error_response, parsed_globals):
+        output = self._get_output_format(parsed_globals)
+
+        try:
+            formatter = get_formatter(output, parsed_globals)
+
+            with self._output_stream_factory.get_output_stream() as stream:
```
ashovlin: 1. In the last internal review of the design doc, we weren't keen anymore on writing these to stdout. Users checking for any output and assuming that means the command was successful may be broken. 
2. I definitely don't think we should feed errors into the pager. It's odd to need to `q` to see the original message. Though they can contain lists, there is no mechanism for pagination.

awscli/clidriver.py
AndrewAsseily: Pagination returns a lazy itertaor that formatting can consume and trigger a later page request. If that fails, catching around the formatter preserves the structured error for pagination-time failures

awscli/clidriver.py
AndrewAsseily: Leaving this comment until we discuss as a team.

awscli/clidriver.py
```diff
@@ -30,6 +30,7 @@
     ScopedConfigProvider,
 )
 from botocore.context import start_as_current_context
+from botocore.exceptions import ClientError
```
ashovlin: Doesn't look used?

awscli/clidriver.py
```diff
@@ -516,12 +536,14 @@ def main(self, args=None):
...
         parser = self.create_parser(command_table)
         self._add_aliases(command_table, parser)
+        parsed_globals = None
         try:
             # Because _handle_top_level_args emits events, it's possible
             # that exceptions can be raised, which should have the same
             # general exception handling logic as calling into the
             # command table.  This is why it's in the try/except clause.
             parsed_args, remaining = parser.parse_known_args(args)
+            parsed_globals = parsed_args
```
ashovlin: Why do we need to call this `parsed_globals`? It doesn't look like we're manipulating it beyond the rename. Could we just pass `parsed_args` into `handle_exception`?

awscli/formatter.py
```diff
@@ -368,12 +368,26 @@ def _format_response(self, response, stream):
...
+    """Formatter that suppresses all output.
+    Only stdout is suppressed; stderr (error messages) remains visible.
+    """
+
+    def __call__(self, command_name, response, stream=None):
+        if is_response_paginated(response):
+            for page in response:
+                self._get_transformed_response_for_output(page)
+        else:
+            self._get_transformed_response_for_output(response)
```
ashovlin: Why do we still do these transformations? Isn't that wasted?

awscli/errorhandler.py
```diff
@@ -60,7 +115,7 @@ def construct_cli_error_handlers_chain():
         NoCredentialsErrorHandler(),
         PagerErrorHandler(),
         InterruptExceptionHandler(),
-        ClientErrorHandler(),
+        ClientErrorHandler(session),
```
ashovlin, line 119: I know I wrote about it a year ago, but I'm leaning (or at least wondering) about whether we should attempt to format our own errors in the specified output style as well. 

More consistency might be preferable? We'd probably have fill `Code` and `Message` with the exception class name and message?

awscli/errorhandler.py
```diff
@@ -108,6 +165,115 @@ class ClientErrorHandler(FilteredExceptionHandler):
...
+                return True
+
+            temp_parsed_globals = argparse.Namespace()
+            temp_parsed_globals.output = error_format
+            temp_parsed_globals.query = None
+            temp_parsed_globals.color = (
+                getattr(parsed_globals, 'color', 'auto')
+                if parsed_globals
+                else 'auto'
+            )
```
ashovlin: Is it necessary to copy all of this into a new namespace object? The only use of `self._args` I see over in the formatters is `query`?

tests/unit/test_structured_error.py
```diff
@@ -0,0 +1,588 @@
...
+        assert 'NoSuchBucket' in stderr_output
+        assert 'test' in stderr_output
+
+
+class TestEnhancedErrorFormatter:
+    def setup_method(self):
+        self.formatter = EnhancedErrorFormatter()
+
+    def test_is_simple_value(self):
+        assert self.formatter._is_simple_value('string')
```
ashovlin: I wouldn't test internal methods here. I think the public `format_error`, with an assertion against the full output (not just contains) will be sufficient.

awscli/errorhandler.py
```diff
@@ -36,10 +37,64 @@
...
+
+        stream.write('\nAdditional error details:\n')
+        for key, value in additional_fields.items():
+            if self._is_simple_value(value):
+                stream.write(f'{key}: {value}\n')
+            elif self._is_small_collection(value):
+                stream.write(f'{key}: {self._format_inline(value)}\n')
+            else:
+                stream.write(
+                    f'{key}: <complex value>\n'
```
ashovlin: Does the line break here break up the `key:value` lines if there are more keys after the complex one? Maybe just print all one one line?

awscli/clidriver.py
AndrewAsseily: Removed in latest revision

awscli/errorhandler.py
AndrewAsseily: Fixed, now prints on one line

tests/unit/test_structured_error.py
AndrewAsseily: Refactored to test public API only with full output assertions

awscli/errorhandler.py
AndrewAsseily: Changed to reuse parsed_globals when available. Only creates fallback when it's None

awscli/formatter.py
AndrewAsseily: Removed to now just consume the iterator without transforming

awscli/clidriver.py
AndrewAsseily: Removed in revision. Passing parsed_args directly now

tests/unit/test_structured_error.py
```diff
@@ -0,0 +1,700 @@
...
+        client_error = ClientError(error_response, 'GetObject')
+
+        stdout = io.StringIO()
+        stderr = io.StringIO()
+
+        rc = self.handler.handle_exception(client_error, stdout, stderr)
+
+        assert rc == CLIENT_ERROR_RC
+        stderr_output = stderr.getvalue()
+        assert 'aws: [ERROR]:' in stderr_output
```
ashovlin: For all tests in this file, can you assert the entire expected output at once instead of substrings with `in`? Like how `test_format_error_with_simple_fields` below does it.

I don't think the output should vary at all? This would catch if we ever messed up whitespace or other formatting.

awscli/errorhandler.py
```diff
@@ -36,11 +37,63 @@
...
+    def format_error(self, error_info, stream):
+        additional_fields = self._get_additional_fields(error_info)
+
+        if not additional_fields:
+            return
+
+        stream.write('\nAdditional error details:\n')
+        for key, value in additional_fields.items():
+            if self._is_simple_value(value):
+                stream.write(f'{key}: {value}\n')
```
ashovlin, line 61: Wondering if we should indent these under the "Additional error details header"

Something like:
```
aws: [ERROR]: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist

Additional error details:
  BucketName: nonexistant-bucket
```

I'm leaning towards yes, but not a strong opinion. Can maybe check with the internal team quickly.

awscli/errorhandler.py
ashovlin: Add `**kwargs` here too in case anyone makes a new subclass?

awscli/errorhandler.py
AndrewAsseily: Good call, I debated this earlier on and went against indentation because I thought it didn't look as clean when there's only a single additional field. But I agree it's worth revisiting. I'll bring it up with the team internally and see what they think.

tests/unit/test_structured_error.py
AndrewAsseily: Agree. Updated in new revision.

awscli/errorhandler.py
AndrewAsseily: Fixed in latest revision.

awscli/errorhandler.py
ashovlin: I'm okay either way, and it'd be safe to change post-launch too.

ashovlin (approved):

PR #9899, opened 2025-12-08
Reduced false positive upgrade warnings for URL parameters

aemous:
*Description of changes:*

* Reduced false positives for URL params upgrade warning.
* Improved spacing around upgrade warnings by prefixing all upgrade warnings with newline character `\n`.
* Added `parsed_globals` globals args to the `load-cli-arg` event.
* Updated tests to support the changes above.

*Description of tests:*

* Manually tested the changes with a command that supports URL params. 


awscli/customizations/cloudformation/deploy.py
```diff
@@ -333,7 +333,7 @@ def deploy(self, deployer, stack_name, template_str,
         try:
             if v2_debug and fail_on_empty_changeset:
                 uni_print(
-                    'AWS CLI v2 UPGRADE WARNING: In AWS CLI v2, '
+                    '\nAWS CLI v2 UPGRADE WARNING: In AWS CLI v2, '
```
ashovlin: Might not matter at this point, but could have been nice to wrap all of these in a `print_migration_debug_message`

aemous: True, one caveat is, code under `/customizations` should use `uni_print`, and all other code should use `print`. May need a boolean arg to choose between either. Does this approach sound good with you?

ashovlin: Don't need to see changes at this point, more for future projects if we ever follow a similar pattern again.

ashovlin (approved):

PR #9901, opened 2025-12-09
Port botocore #3592

SamRemis:
Port of https://github.com/boto/botocore/pull/3592 to add host prefix injection to RPCv2.  Without this change, incorrect hosts may be resolved for service relying on the host prefix.  

### DIFF
--- a/.changes/next-release/bugfix-Serializers-42643.json
+++ b/.changes/next-release/bugfix-Serializers-42643.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Serializers",
+  "description": "Adds support for host prefix to BaseRpcV2Serializer class"
+}


ashovlin:
Is this a runtime change in behavior? If so, can you add a changelog entry? 

And please write a more meaningful commit message: https://aws.github.io/aws-cli/git_workflows.html

SamRemis:
> Is this a runtime change in behavior? If so, can you add a changelog entry?

Not to services that have been released yet.  [The only service](https://github.com/search?q=repo%3Aboto%2Fbotocore+++path%3Abotocore%2Fdata+smithy-rpc-v2-cbor&type=code) that uses RPCv2 currently is arc-region-switch, which also supports JSON.  JSON is what winds up getting used by botocore in this case.

> And please write a more meaningful commit message: https://aws.github.io/aws-cli/git_workflows.html

Absolutely.  I can write something in the commit message along the lines of "adds support for host prefix to BaseRpcV2Serializer class".

PR #9906, opened 2025-12-10
Modify debug migration s3 us-east-1 detections

aemous:
*Description of changes:*

* Modify the us-east-1 S3 endpoint check so that:
  * It restricts the `s3.amazonaws.com` check to the hostname of the URL. This prevents false positives, for example, when the object key/prefix contains `s3.amazonaws.com`.
  * It does not trigger detection when `--endpoint-url` is used.
  * If the region is configured to something other than `us-east-1`, it will not trigger detection. 
* Switch changelog entry from `enhancement` type to `feature`.


awscli/customizations/globalargs.py
```diff
@@ -133,12 +133,13 @@ def detect_migration_breakage(parsed_args, remaining_args, session, **kwargs):
...
-                'legacy'
-            ) == 'legacy' and region in ('us-east-1', None)
-        )
+            (
+                s3_config is None
+                or s3_config.get('us_east_1_regional_endpoint', 'legacy')
+                == 'legacy'
+            )
+            and region in ('us-east-1', None)
+            and parsed_args.endpoint_url is None
```
ashovlin: Does this include if a custom endpoint is set via `AWS_ENDPOINT_URL`/`endpoint_url` , `AWS_ENDPOINT_URL_S3`, or the service-specific config override?

Might not be too bad to always register this handler if `--v2-debug` is passed, rather than trying to complete this logic here.

awscli/customizations/globalargs.py
```diff
@@ -220,7 +221,8 @@ def warn_if_east_configured_global_endpoint(request, operation_name, **kwargs):
     # The regional us-east-1 endpoint is used in certain cases (e.g.
     # FIPS/Dual-Stack is enabled). Rather than duplicating this logic
     # from botocore, we check the endpoint URL directly.
-    if 's3.amazonaws.com' in request.url:
+    parsed_url = urlparse.urlparse(request.url)
```
ashovlin, line 223: We're confident custom endpoints won't hit this, right? If not, I'd rather be defensive and either swallow or handle a possible `ValueError`.

awscli/customizations/globalargs.py
aemous: Good point. Looks like if I set `AWS_ENDPOINT_URL_S3` to `s3.amazonaws.com`, and `--region us-east-1`, the warning gets printed.

I agree we likely don't want to duplicate all the logic for endpoint resolution here. I'll remove the --endpoint-url check.

awscli/customizations/globalargs.py
aemous: Custom endpoints may reach this point, for example if a user provides a custom endpoint for `AWS_ENDPOINT_URL_S3` and sets region to `us-east-1`. 

However, a custom endpoint that is not affiliated with AWS will not have their hostname end in `s3.amazonaws.com` .

ashovlin: I was more thinking about if `urlparse` throws an exception. I guess the request would ultimately have failed anyway, so as long as the stack trace is reasonable from here, that's fine.

aemous: Agreed, if the URL is invalid in some way, the request should fail.

Tested locally, provided an invalid url, and it fails with the expected error:

```
Bad value for --endpoint-url "1": scheme is missing.  Must be of the form http://<hostname>/ or https://<hostname>/
```

awscli/customizations/globalargs.py
aemous: revised

ashovlin (approved):

PR #9915, opened 2025-12-12
Creating stream display to be used for text only monitoring

jutyler1:
*Issue #, if available:*

*Description of changes:*
Creating stream display to be used for text only monitoring. This display is not hooked up for usage yet.

### DIFF
--- a/tests/unit/customizations/ecs/expressgateway/test_managedresource.py
+++ b/tests/unit/customizations/ecs/expressgateway/test_managedresource.py
@@ -1,4 +1,6 @@
-import unittest
+import re
+
+import pytest
 
 from awscli.customizations.ecs.expressgateway.managedresource import (
     TERMINAL_RESOURCE_STATUSES,
@@ -6,33 +8,33 @@ from awscli.customizations.ecs.expressgateway.managedresource import (
 )
 
 
-class TestManagedResource(unittest.TestCase):
+class TestManagedResource:
     def test_is_terminal_active(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "ACTIVE", 1761230543.151
         )
-        self.assertTrue(resource.is_terminal())
+        assert resource.is_terminal()
 
     def test_is_terminal_failed(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "FAILED", 1761230543.151
         )
-        self.assertTrue(resource.is_terminal())
+        assert resource.is_terminal()
 
     def test_is_terminal_provisioning(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "PROVISIONING", 1761230543.151
         )
-        self.assertFalse(resource.is_terminal())
+        assert not resource.is_terminal()
 
     def test_get_status_string_active(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "ACTIVE", 1761230543.151
         )
         status_string = resource.get_status_string("⠋")
-        self.assertIn("LoadBalancer", status_string)
-        self.assertIn("lb-123", status_string)
-        self.assertIn("ACTIVE", status_string)
+        assert "LoadBalancer" in status_string
+        assert "lb-123" in status_string
+        assert "ACTIVE" in status_string
 
     def test_get_status_string_failed_with_reason(self):
         resource = ManagedResource(
@@ -43,8 +45,8 @@ class TestManagedResource(unittest.TestCase):
             "Connection timeout",
         )
         status_string = resource.get_status_string("⠋")
-        self.assertIn("FAILED", status_string)
-        self.assertIn("Connection timeout", status_string)
+        assert "FAILED" in status_string
+        assert "Connection timeout" in status_string
 
     def test_get_status_string_active_with_reason(self):
         resource = ManagedResource(
@@ -55,8 +57,8 @@ class TestManagedResource(unittest.TestCase):
             "Load balancer ready",
         )
         status_string = resource.get_status_string("⠋")
-        self.assertIn("ACTIVE", status_string)
-        self.assertIn("Load balancer ready", status_string)
+        assert "ACTIVE" in status_string
+        assert "Load balancer ready" in status_string
 
     def test_combine_newer_resource(self):
         older = ManagedResource(
@@ -66,7 +68,7 @@ class TestManagedResource(unittest.TestCase):
             "LoadBalancer", "lb-123", "ACTIVE", 1761230600.151
         )
         result = older.combine(newer)
-        self.assertEqual(result, newer)
+        assert result == newer
 
     def test_combine_older_resource(self):
         older = ManagedResource(
@@ -76,33 +78,33 @@ class TestManagedResource(unittest.TestCase):
             "LoadBalancer", "lb-123", "ACTIVE", 1761230600.151
         )
         result = newer.combine(older)
-        self.assertEqual(result, newer)
+        assert result == newer
 
     def test_combine_with_none(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "ACTIVE", 1761230543.151
         )
         result = resource.combine(None)
-        self.assertEqual(result, resource)
+        assert result == resource
 
     def test_is_terminal_deleted_status(self):
         resource = ManagedResource("LoadBalancer", "lb-123", "DELETED")
-        self.assertTrue(resource.is_terminal())
+        assert resource.is_terminal()
 
     def test_is_terminal_no_status(self):
         resource = ManagedResource("LoadBalancer", "lb-123", None)
-        self.assertFalse(resource.is_terminal())
+        assert not resource.is_terminal()
 
     def test_init_with_string_timestamp(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "ACTIVE", "2025-11-05T18:00:00Z"
         )
-        self.assertIsInstance(resource.updated_at, float)
-        self.assertGreater(resource.updated_at, 0)
+        assert isinstance(resource.updated_at, float)
+        assert resource.updated_at > 0
 
     def test_init_with_none_timestamp(self):
         resource = ManagedResource("LoadBalancer", "lb-123", "ACTIVE", None)
-        self.assertIsNone(resource.updated_at)
+        assert resource.updated_at is None
 
     def test_combine_with_no_timestamp(self):
         resource1 = ManagedResource(
@@ -112,7 +114,7 @@ class TestManagedResource(unittest.TestCase):
             "LoadBalancer", "lb-123", "PROVISIONING", None
         )
         result = resource1.combine(resource2)
-        self.assertEqual(result, resource1)
+        assert result == resource1
 
     def test_combine_equal_timestamps(self):
         timestamp = 1761230543.151
@@ -123,7 +125,7 @@ class TestManagedResource(unittest.TestCase):
             "LoadBalancer", "lb-123", "PROVISIONING", timestamp
         )
         result = resource1.combine(resource2)
-        self.assertEqual(result, resource1)
+        assert result == resource1
 
     def test_get_status_string_with_depth(self):
         resource = ManagedResource(
@@ -132,7 +134,7 @@ class TestManagedResource(unittest.TestCase):
         status_string = resource.get_status_string("⠋", depth=2)
         # Should have proper indentation
         lines = status_string.split('\n')
-        self.assertTrue(lines[0].startswith("  "))  # 2 spaces for depth=2
+        assert lines[0].startswith("  ")  # 2 spaces for depth=2
 
     def test_get_status_string_with_additional_info(self):
         resource = ManagedResource(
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+)
+from awscli.customizations.ecs.expressgateway.stream_display import (
+    StreamDisplay,
+)
+
+
+class TestStreamDisplay:
+    """Test StreamDisplay for text-based monitoring output."""
+
+    def setup_method(self):
```
aemous: PyTest recommends we use fixtures instead. Can we use a fixture?

e.g.

```
@pytest.fixture
def display(self):
    return StreamDisplay(use_color=True)
```

For tests that need to use a display, add `display` as an argument, and PyTest will pass the fixture as an argument

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def test_show_monitoring_data_with_new_resource(self, capsys):
+        """Test output when new resources are added"""
+        resource1 = ManagedResource(
+            "LoadBalancer", "lb-123", "ACTIVE", None, None
+        )
+        resource_group1 = ManagedResourceGroup(resources=[resource1])
+
+        self.display.show_monitoring_data(resource_group1, None)
+        capsys.readouterr()
```
aemous: Can you explain what this is doing in relation to the test?

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+    def test_use_color_parameter(self):
+        """Test use_color parameter is stored"""
+        display_with_color = StreamDisplay(use_color=True)
+        assert display_with_color.use_color is True
+
+        display_no_color = StreamDisplay(use_color=False)
+        assert display_no_color.use_color is False
+
+    def test_print_flattened_resources_with_reason(self, capsys):
+        """Test resource with reason prints on separate line"""
```
aemous, line 153: Are the assertions really verifying the resource and reason are on separate lines? I think the assertions just verify the substrings are in the output. Would be nice if we can cleanly test that they're on separate lines. Maybe using `splitlines()` or using `\n`. Similar comment for other tests that are trying to test that they're on separate lines.

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+    def test_diff_detects_status_change(self, capsys):
+        """Test diff detects when status changes"""
+        resource1 = ManagedResource(
+            "LoadBalancer", "lb-123", "CREATING", None, None
+        )
+        resource_group1 = ManagedResourceGroup(resources=[resource1])
+
+        # First poll
+        self.display.show_monitoring_data(resource_group1, None)
+        capsys.readouterr()
```
aemous: What does this do in relation to the test?

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+    def test_diff_detects_reason_change(self, capsys):
+        """Test diff detects when reason changes"""
+        resource1 = ManagedResource(
+            "LoadBalancer", "lb-123", "CREATING", None, "Creating resources"
+        )
+        resource_group1 = ManagedResourceGroup(resources=[resource1])
+
+        # First poll
+        self.display.show_monitoring_data(resource_group1, None)
+        capsys.readouterr()
```
aemous: What does this do in relation to the test?

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+        self.display.show_monitoring_data(resource_group, None)
+
+        output = capsys.readouterr().out
+        assert "mystery-resource-123" in output
+        assert "FAILED" in output
+
+    def test_resource_with_none_type_and_none_identifier(self, capsys):
+        """Test resources with both resource_type=None and identifier=None show only status"""
```
aemous: So in the case resource_type and identifier are both None, it seems text-only will just show 
`[timestamp] ACTIVE`, is that right? Is this desired behavior? I feel it could be confusing. 

Any alternatives?

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
```diff
@@ -1,288 +0,0 @@
-import unittest
```
aemous, line 1: I see we're removing tests for ManagedResourceGroup, but I don't see removed functionality in ManagedResourceGroup in this PR. Does this result in decreased test coverage for these functions? Are the tests in test_stream_display meant to cover everything that we are deleting here?

awscli/customizations/ecs/expressgateway/managedresource.py
```diff
@@ -113,6 +113,52 @@ def get_status_string(self, spinner_char, depth=0, use_color=True):
         lines.append("")
         return '\n'.join(lines)
 
+    def get_stream_string(self, timestamp, use_color=True):
```
aemous, line 116: I'd like to see some added tests for `get_stream_string`.

awscli/customizations/ecs/expressgateway/managedresourcegroup.py
```diff
@@ -57,6 +57,86 @@ def _create_key(self, resource):
         identifier = resource.identifier if resource.identifier else ""
         return resource_type + "/" + identifier
 
+    def get_stream_string(self, timestamp, use_color=True):
```
aemous, line 60: Would like to see some tests for this get_stream_string too

awscli/customizations/ecs/expressgateway/stream_display.py
```diff
@@ -0,0 +1,110 @@
...
+            resource_group: ManagedResourceGroup or None
+            info: Additional info text to display
+        """
+        timestamp = self._get_timestamp()
+
+        if resource_group:
+            (
+                changed_resources,
+                updated_dict,
+                removed_keys,
```
aemous, line 58: Does a removed key imply a resource was deleted in the account? Should that correspond to an updated status in monitoring?

awscli/customizations/ecs/expressgateway/stream_display.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+            self.previous_resources_by_key = updated_dict
+
+            if changed_resources:
+                self._print_flattened_resources_list(
+                    changed_resources, timestamp
+                )
+
+        if info:
+            uni_print(f"[{timestamp}] {info}\n")
```
aemous, line 71: Started a discussion offline, but I'm curious what cases this would be printed, in terms of AWS account resources? When would callers provide a non-None info?

aemous (changes requested):
Looks good overall, requested a few small changes and had a few questions

awscli/customizations/ecs/expressgateway/stream_display.py
aemous: Discussed offline, can be used in cases such as not yet achieving eventual consistency , etc.

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
jutyler1: these are testing the output of the delta, so this clears the output of the initial state

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
jutyler1: in practice this should not happen, i'll update to say something like unknown resource to indicate that something is there, but there are issues with it

awscli/customizations/ecs/expressgateway/stream_display.py
jutyler1: it means that we are no longer tracking a resource in the computed view of the service.

if in resource mode, dropping the resource would mean that it is no longer associated. that should only happen for terminal service revision resources, which would have a status of DELETED to begin with. 

if in deployment mode, dropping the resource would mean that it was found to be in common with an existing resources, once the identifier was populated async. we should ignore this to not give the appearance that something has changed.

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
jutyler1: asserted different lines

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
jutyler1: updated

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
jutyler1: it should not have been dropped, add them back

awscli/customizations/ecs/expressgateway/managedresource.py
jutyler1: added

awscli/customizations/ecs/expressgateway/managedresourcegroup.py
jutyler1: added

awscli/customizations/ecs/expressgateway/stream_display.py
aemous: Makes sense, thanks

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
aemous: Thanks for clarifying.

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
aemous: Answered in other thread. Resolved.

tests/unit/customizations/ecs/expressgateway/test_stream_display.py
aemous: Answered in other thread. Resolved.

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
aemous: Can we migrate this to use PyTest?

tests/unit/customizations/ecs/expressgateway/test_managedresource.py
aemous: Can we migrate this to use PyTest?

aemous (changes requested):
Looks good! Just want to migrate the expressgateway test suites to use PyTest, `unittest` is our legacy tests and we have since then used PyTests for new test suites.

tests/unit/customizations/ecs/expressgateway/test_managedresource.py
jutyler1: updated

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
jutyler1: updated

tests/unit/customizations/ecs/expressgateway/test_managedresource.py
```diff
@@ -205,16 +205,18 @@ def test_get_stream_string_with_additional_info(self):
...
     def test_get_stream_string_with_updated_at(self):
         resource = ManagedResource(
             "LoadBalancer", "lb-123", "ACTIVE", 1761230543.151
         )
         stream_string = resource.get_stream_string("2025-12-15 10:00:00")
-        self.assertIn("Last Updated At:", stream_string)
+        assert "Last Updated At:" in stream_string
         # Check timestamp format YYYY-MM-DD HH:MM:SS
-        self.assertRegex(stream_string, r"\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}")
+        import re
```
aemous: nit: can we move this import to the top of the file for consistency?

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
```diff
@@ -8,69 +8,71 @@
...
-        self.assertIn("endpoint-1", status_string)
+        assert "IngressPaths" in status_string
+        assert "endpoint-1" in status_string
 
-    def test_create_key(self):
+    def test_create_key(self, resource1):
         group = ManagedResourceGroup()
-        key = group._create_key(self.resource1)
-        self.assertEqual(key, "LoadBalancer/lb-123")
+        key = group._create_key(resource1)
```
aemous: let's avoid testing private functions. can we either test this indirectly or refactor?

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
```diff
@@ -190,70 +190,70 @@ def test_compare_resource_sets_excludes_matching_types_without_identifier(
...
             resource_type="IngressPaths", identifier="test-id", status="ACTIVE"
         )
         result = group.get_status_string("⠋")
-        self.assertIn("IngressPaths", result)
-        self.assertIn("test-id", result)
-        self.assertIn("✓", result)  # Green checkmark for ACTIVE
-        self.assertIn("ACTIVE", result)
+        assert "IngressPaths" in result
+        assert "test-id" in result
+        assert "✓" in result  # Green checkmark for ACTIVE
```
aemous, line 144: Can you confirm that text-only mode will not use symbols in this way to convey information? If I recall correctly we previously agreed to use text such as "ACTIVE" without symbols for text-only mode. Similar for other tests.

tests/unit/customizations/ecs/expressgateway/test_managedresource.py
jutyler1: moved it

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
jutyler1: removed

tests/unit/customizations/ecs/expressgateway/test_managedresourcegroup.py
jutyler1: that will be produced via `get_stream_string` instead of `get_status_string`. The former will not include symbols

PR #9918, opened 2025-12-12
[v2] Add GitHub Actions build workflow

hssyoo:
Add a GitHub Actions workflow to trigger internal builds when PR is opened and in a non-draft state

### DIFF
--- a/.github/workflows/pull-request-build.yaml
+++ b/.github/workflows/pull-request-build.yaml
@@ -1,0 +1,54 @@
+name: Build SDK
+on:
+  pull_request_review:
+    types: [ opened, synchronize, ready_for_review ]
+
+concurrency:
+  group: start-pull-request-build-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # constants
+  DOWNLOAD_FOLDER: '.build-scripts/'
+  SCRIPT_LOCATION: 'workflows/start-pull-request-build/pull-request-build-v1.sh'
+
+  # custom variables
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
...
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      issues: write
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@main
```
ashovlin, line 30: Could we pin to commit in case there's a breaking change?

hssyoo: In this case I'd rather prioritize getting latest commits with potential security fixes than ensuring stability since the worst case is we need to manually trigger builds. Also don't want to have to regularly update this, though open to changing this later if we find that there's frequent breakages.

ashovlin: That's fair, since it's not automated or truly required that's okay.

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
+name: Build SDK
```
ashovlin, line 1: non-blocker: Switch to "CLI" or give a more descriptive name, line 20 too?

Something with "internal", "installers", "integration tests", etc.

PR #9923, opened 2025-12-16
Extract mode specific display into DisplayStrategy

jutyler1:
*Issue #, if available:*

*Description of changes:*
Extract mode specific display into DisplayStrategy.

This should be no-op. A new strategy/mode will be added next comit

### DIFF
--- a/awscli/customizations/ecs/expressgateway/display_strategy.py
+++ b/awscli/customizations/ecs/expressgateway/display_strategy.py
@@ -27,7 +27,7 @@ class DisplayStrategy:
     Each strategy controls its own execution model, timing, and output format.
     """
 
-    def execute(self, collector, start_time, timeout_minutes):
+    def execute_monitoring(self, collector, start_time, timeout_minutes):
         """Execute the monitoring loop.
 
         Args:
@@ -47,10 +47,17 @@ class InteractiveDisplayStrategy(DisplayStrategy):
     """
 
     def __init__(self, display, use_color):
+        """Initialize the interactive display strategy.
+
+        Args:
+            display: Display instance from prompt_toolkit_display module
+                providing the interactive terminal interface
+            use_color: Whether to use colored output
+        """
         self.display = display
         self.use_color = use_color
 
-    def execute(self, collector, start_time, timeout_minutes):
+    def execute_monitoring(self, collector, start_time, timeout_minutes):
         """Execute async monitoring with spinner and keyboard controls."""
         final_output, timed_out = asyncio.run(
             self._execute_async(collector, start_time, timeout_minutes)
@@ -143,6 +150,8 @@ class InteractiveDisplayStrategy(DisplayStrategy):
                     pass
 
         finally:
+            # Ensure display app is properly shut down
+            self.display.app.exit()
             spinner_task.cancel()
             if display_task is not None and not display_task.done():
                 display_task.cancel()
--- a/tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
+++ b/tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
@@ -28,7 +28,7 @@ class TestDisplayStrategy:
         """Test base class raises NotImplementedError."""
         strategy = DisplayStrategy()
         with pytest.raises(NotImplementedError):
-            strategy.execute(None, None, None)
+            strategy.execute_monitoring(None, None, None)
 
 
 @pytest.fixture
@@ -71,7 +71,9 @@ class TestInteractiveDisplayStrategy:
         mock_sleep.side_effect = KeyboardInterrupt()
 
         start_time = time.time()
-        strategy.execute(mock_collector, start_time, timeout_minutes=1)
+        strategy.execute_monitoring(
+            mock_collector, start_time, timeout_minutes=1
+        )
 
         # Verify display was called
         assert mock_display.display.called
@@ -106,7 +108,9 @@ class TestInteractiveDisplayStrategy:
         mock_sleep.side_effect = KeyboardInterrupt()
 
         start_time = time.time()
-        strategy.execute(mock_collector, start_time, timeout_minutes=1)
+        strategy.execute_monitoring(
+            mock_collector, start_time, timeout_minutes=1
+        )
 
         captured = capsys.readouterr()
         assert "Monitoring Complete!" in captured.out
@@ -128,7 +132,9 @@ class TestInteractiveDisplayStrategy:
         mock_sleep.side_effect = KeyboardInterrupt()
 
         start_time = time.time()
-        strategy.execute(mock_collector, start_time, timeout_minutes=1)
+        strategy.execute_monitoring(
+            mock_collector, start_time, timeout_minutes=1
+        )
 
         captured = capsys.readouterr()
         assert unique_output in captured.out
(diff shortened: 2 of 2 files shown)


awscli/customizations/ecs/expressgateway/display_strategy.py
```diff
@@ -0,0 +1,140 @@
...
+
+        try:
+            display_task = asyncio.create_task(self.display.run())
+
+            done, pending = await asyncio.wait(
+                [display_task, data_task], return_when=asyncio.FIRST_COMPLETED
+            )
+
+            if data_task in done:
+                await data_task
```
aemous: Just want to clarify that I'm understanding this correctly.

1. Wait for at least one of display or data tasks to complete.
2. If data task is done, await its completion.

If that's correct, I'm a bit confused by the if-statement. Why must we await the data_task if it was done?

awscli/customizations/ecs/expressgateway/display_strategy.py
```diff
@@ -0,0 +1,140 @@
...
+
+        finally:
+            spinner_task.cancel()
+            if display_task is not None and not display_task.done():
+                display_task.cancel()
+                try:
+                    await display_task
+                except asyncio.CancelledError:
+                    pass
```
aemous, line 163: Can you explain this, and potentially add a comment with the explanation for why it's needed?

tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
```diff
@@ -0,0 +1,224 @@
...
+            strategy.execute(None, None, None)
+
+
+@pytest.fixture
+def app_session():
+    """Fixture that creates and manages an app session for prompt_toolkit."""
+    session = create_app_session(output=DummyOutput())
+    session.__enter__()
+    yield session
+    session.__exit__(None, None, None)
```
aemous:
```suggestion
@pytest.fixture
def app_session():
    """Fixture that creates and manages an app session for prompt_toolkit."""
    with create_app_session(output=DummyOutput()) as session:
        yield session
```

tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
```diff
@@ -0,0 +1,224 @@
...
+        self, mock_sleep, app_session, capsys
+    ):
+        """Test that collector output appears in final output."""
+
+        async def mock_run_async():
+            await asyncio.sleep(0.01)
+
+        mock_display = Mock()
+        mock_display.display = Mock()
+        mock_display.run = Mock(return_value=mock_run_async())
```
aemous: optional suggestion: consider extracting this into a mock_display fixture, since this pattern is duplicated a few times across some test cases.

aemous (changes requested):
some questions, and very minor change(s) requested

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: The data_task could have exceptions so we call await to reraise them if they are there

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: similarly, when cancelling, we await for the cancellation exception so that we know it actually shut down

tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
jutyler1: updated

tests/unit/customizations/ecs/expressgateway/test_display_strategy.py
jutyler1: updated

awscli/customizations/ecs/expressgateway/display_strategy.py
aemous: Thanks for clarifying!

Can we instead call `data_task.exception()` to check for an exception ([docs](https://quip-amazon.com/enfJAEczLgRE/Dev-Guide-Draft-Using-upgrade-debug-mode))? 

I'd prefer that for a few reasons

1. What if an exception was raised in the `asyncio.wait` but when we re-run to check for exceptions there's no exception? I'd be concerned of hiding/swallowing exceptions for this reason.
2. We're assuming re-running data_task here does not have unintended consequences. This may not be true in the future; i'd prefer to avoid relying on this implicit assumption.

awscli/customizations/ecs/expressgateway/display_strategy.py
aemous: Can we use `display_task.cancelled()` instead? It returns `True` if the task was cancelled. ([source](https://quip-amazon.com/enfJAEczLgRE/Dev-Guide-Draft-Using-upgrade-debug-mode)).

awscli/customizations/ecs/expressgateway/display_strategy.py
```diff
@@ -126,12 +126,26 @@ async def update_spinner():
...
 
+            # Cancel pending tasks
+            for task in pending:
+                task.cancel()
+                # Await cancelled task to ensure proper cleanup and prevent
+                # warnings about unawaited tasks
+                try:
+                    await task
+                except asyncio.CancelledError:
+                    pass
```
aemous, line 150: Similar to other replies, we can use `task.cancelled()` instead to check if the task was cancelled.

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: we are not re-running tasks, we are just retrieving the result/exceptions. I can use exception to be more explicit

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: cancelled will just check the status rather than await for the task to shut down and clean up

aemous: Great point, thanks for clarifying!

awscli/customizations/ecs/expressgateway/display_strategy.py
```diff
@@ -0,0 +1,156 @@
...
+                        )
+                        if (
+                            "Cannot call DescribeServiceRevisions for a service that is INACTIVE"
+                            in error_message
+                        ):
+                            current_output = "Service is inactive"
+                        else:
+                            raise
+                    else:
+                        raise
```
aemous, line 108: Should we be catching (and re-raising) CancelledError for cleaning up the display with self.display.app.exit()? Is it safe not to exit the app?

awscli/customizations/ecs/expressgateway/display_strategy.py
```diff
@@ -0,0 +1,156 @@
...
+class InteractiveDisplayStrategy(DisplayStrategy):
+    """Interactive display strategy with async spinner and keyboard navigation.
+
+    Uses dual async tasks:
+    - Data task: Polls ECS APIs every 5 seconds
+    - Spinner task: Updates display every 100ms with rotating spinner
+    """
+
+    def __init__(self, display, use_color):
+        self.display = display
```
aemous, line 57: What's the expected type/interface of display?

aemous (commented):
couple more questions

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: good point on app.exit, added app.exit() in the finally block at the end of _async_execute. no need to catch and re-raise CancelledError, it should eventually hit the finally block.

awscli/customizations/ecs/expressgateway/display_strategy.py
jutyler1: added docstring. its a prompt toolkit display

aemous (approved):
LGTM. Thanks for adding the new test for verifying the display is closed.

Issue #9869, opened 2025-11-22
aws login credentials are silently overridden by static access keys


k-ibaraki:
### Describe the bug

When static AWS credentials (access keys) exist in ~/.aws/credentials, the `aws login` command displays a success message saying it "Updated profile default to use [new credentials]", but those new credentials are silently ignored. The static credentials from ~/.aws/credentials continue to be used due to credential precedence rules, with no warning to the user.

This creates a dangerous situation where users believe they have switched to different credentials (via aws login) but are actually still using their old static access keys.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

When a user runs `aws login` successfully:

1. If static credentials exist in ~/.aws/credentials for the same profile, AWS CLI should:
   - Display a WARNING that static credentials will take precedence, OR
   - Prompt the user to choose whether to remove the conflicting static credentials, OR
   - Fail with a clear error message about the credential conflict

2. Users should never see "Updated profile to use X" when X will not actually be used

3. The success message should accurately reflect whether the login credentials will actually be used

### Current Behavior

1. User has static access keys in ~/.aws/credentials for [default] profile
2. User runs `aws login` and successfully authenticates
3. AWS CLI displays: "Updated profile default to use arn:aws:sts::XXXXXXXXXXXX:assumed-role/RoleName/username credentials."
4. User believes the profile now uses the new login credentials
5. User runs AWS CLI commands
6. Commands silently use the static credentials from ~/.aws/credentials
7. The login credentials are completely ignored
8. No warning or indication is given that a different credential is being used

### Reproduction Steps

1. Configure static access keys in ~/.aws/credentials:
   [default]
   aws_access_key_id = AKIA...
   aws_secret_access_key = xxx...

2. Verify current identity:
   $ aws sts get-caller-identity
   {
       "Account": "111111111111",
       "Arn": "arn:aws:iam::111111111111:user/test-user"
   }

3. Run `aws login` and complete browser authentication
   Output shows:
   "Updated profile default to use arn:aws:sts::222222222222:assumed-role/AdminRole/username credentials."

4. Check identity again:
   $ aws sts get-caller-identity
   {
       "Account": "111111111111",
       "Arn": "arn:aws:iam::111111111111:user/test-user"
   }
   
   ⚠️ Still using static credentials despite success message claiming profile was updated

5. Manually remove static credentials from ~/.aws/credentials

6. Check identity again:
   $ aws sts get-caller-identity
   {
[...]

PR #9924, opened 2025-12-16
Prevent 'aws login' from updating a profile with existing credentials

ashovlin:
*Issue #, if available:* #9869

*Description of changes:*

~This adds a prompt to `aws login` to warn users when running it for a profile that already has a different style of credentials:~
```
$ aws login

Warning: Profile 'default' is already configured with Access Key credentials. If you continue to log in, the CLI and other tools may continue to use the existing credentials instead.

You may run 'aws login --profile new-profile-name' to create a new profile, or else you may manually remove the existing credentials from 'default'.

Do you want to continue adding login credentials to 'default'? (y/n):
```

This now prevents the user from proceeding if the profile already has existing credentials.
```
$ aws login

Error: Profile 'default' is already configured with Access Key credentials.

You may run 'aws login --profile new-profile-name' to create a new profile, or you must first manually remove the existing credentials from 'default'.
```

This should help avoid the confusing case where you run `aws login` but subsequent commands are still using different credentials.

I did not add support for clearing the other style of credentials yet. For access keys there wouldn't be an easy way to undo or retrieve them if we just delete them from the `credentials` file, so I'd want to be more careful here. If we get feedback, we could expand in the future.

### DIFF
--- a/awscli/customizations/login/login.py
+++ b/awscli/customizations/login/login.py
@@ -20,6 +20,7 @@ from awscli.customizations.configure.sso import (
     RequiredInputValidator,
 )
 from awscli.customizations.configure.writer import ConfigFileWriter
+from awscli.customizations.exceptions import ConfigurationError
 from awscli.customizations.login.utils import (
     CrossDeviceLoginTokenFetcher,
     LoginType,
@@ -98,8 +99,7 @@ class LoginCommand(BasicCommand):
 
         # Abort if the profile is already configured with a different style
         # of credentials, since they'd still have precedence over login
-        if self.does_profile_have_non_login_credentials(profile_name):
-            return
+        self.ensure_profile_does_not_have_existing_credentials(profile_name)
 
         config = botocore.config.Config(
             region_name=region,
@@ -182,10 +182,10 @@ class LoginCommand(BasicCommand):
             else:
                 uni_print('Invalid response. Please enter "y" or "n"')
 
-    def does_profile_have_non_login_credentials(self, profile_name):
+    def ensure_profile_does_not_have_existing_credentials(self, profile_name):
         """
-        Checks if the specified profile is already configured
-        with a different style of credentials.
+        Raises an error if the specified profile is already
+        configured with a different style of credentials.
         """
         config = self._session.full_config['profiles'].get(profile_name, {})
         existing_credentials_style = None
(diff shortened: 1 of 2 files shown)


awscli/customizations/login/login.py
```diff
@@ -177,6 +182,49 @@ def accept_change_to_existing_profile_if_needed(
...
+    def accept_existing_credentials_warning_if_needed(self, profile_name):
+        """
+        Checks if the specified profile is already configured with a
+        different style of credentials. If so, warn the user and prompt them to
+        continue.
+        """
+        config = self._session.full_config['profiles'].get(profile_name, {})
+        existing_credentials_style = None
+
+        if 'web_identity_token_file' in config:
```
hssyoo, line 193: Is it possible to trigger the credential chain and see which (if any) provider returns? I don't love hardcoding different providers here since we'll have to remember to update it as needed

awscli/customizations/login/login.py
```diff
@@ -177,6 +182,49 @@ def accept_change_to_existing_profile_if_needed(
...
+
+        if not existing_credentials_style:
+            return True
+
+        while True:
+            response = compat_input(
+                f'\nWarning: Profile \'{profile_name}\' is already configured '
+                f'with {existing_credentials_style} credentials. '
+                f'If you continue to log in, the CLI and other tools may '
+                f'continue to use the existing credentials instead.\n\n'
```
hssyoo: This seems like a weird experience and I feel like we should be gatekeeping here. Are there reasons we shouldn't just deny the login attempt here?

awscli/customizations/login/login.py
ashovlin: 1. I don't think we want the full credentials chain - if something low in the hierarchy like the ECS provider or IMDS is present, the user might still want to configure a separate profile.
2. I don't want to actually resolve credentials since that requires a network request for a few flavors 

Each provider does have some internal code similar to this for _"determine if a given profile can be used with this provider_". We could extract that so that it's shared, but not sure it'd be worth diverging from botocore that much.

awscli/customizations/login/login.py
ashovlin: Agreed, updated to abort via https://github.com/aws/aws-cli/pull/9924/commits/3571218a0554ae5c715694aa214aad3f9606e0e9.

I'm still not sold on making it easy to clear existing creds via a prompt here, since that may be difficult to undo, especially for access keys. Maybe an `aws profile unset` could address this, since that'd require more intent? We can wait for more feedback.

awscli/customizations/login/login.py
```diff
@@ -202,28 +201,17 @@ def accept_existing_credentials_warning_if_needed(self, profile_name):
...
-            return True
-
-        while True:
-            response = compat_input(
-                f'\nWarning: Profile \'{profile_name}\' is already configured '
-                f'with {existing_credentials_style} credentials. '
-                f'If you continue to log in, the CLI and other tools may '
-                f'continue to use the existing credentials instead.\n\n'
+        if existing_credentials_style:
+            uni_print(
```
hssyoo: non-blocking: It's a _little_ weird that the command returns 0 and an error message is directed to stdout in this case.

ashovlin: Agreed, think I was still thinking about the prompt in the first iteration. Switched to an error in https://github.com/aws/aws-cli/pull/9924/commits/c45e10e7d8440a3e10c6fe010663d8279c9ea34a

Issue #9596, opened 2025-07-16
aws s3 sync from S3 -> Windows issue when directory casing does not match


mike406:
### Describe the bug

Hello, I am looking to report an issue I am seeing with running aws s3 sync. I am trying to sync an S3 bucket to an FSx SMB share mounted as a drive on a Windows EC2 instance. I am noticing that some folders in the destination FSx share are missing files that are present in S3 and they are not being downloaded. Nothing about these files are strange, they are PDFs and .docx files. Any thoughts on why files present in S3 are being ignored and not downloaded? If I sync the exact subfolder specifically, then they do get downloaded.

The full command I am using is `aws s3 sync s3://{my-bucket} \\share\path --delete`

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I expect all files present in S3 to be downloaded to the FSx share if they are not present there.

### Current Behavior

Certain files are not being downloaded so some folders have less files than they should.

### Reproduction Steps

See my comment below - I did end up figuring out how to reproduce the issue after some testing.

### CLI version used

2.27.48.0

### Environment details (OS name and version, etc.)

Windows Server 2019

mike406:
Quick followup, is it possible it is a casing issue with Windows?

For example, the path in the source S3 bucket currently is /folder1/FOLDER2/folder3/
And the path in the destination Windows FSx share is /folder1/folder2/folder3/

It seems that at some point the casing for one of the folders was changed in the S3 bucket and this does not get updated at the destination in future syncs. Is this possible this is the cause and is there any way to work around this to ensure that all files get downloaded?

Edit: Yes this appears to be the cause I am able to replicate it now - if there is a casing mismatch with directories on Windows and S3 and you are also using the `--delete` parameter it deletes all files at the destination _even if those files are supposed to be there_ and downloads files at the same time which looks like it is resulting in some race condition where some files are missing at when it finishes.

RyanFitzSimmonsAK:
Hi @mike406, thank you for reaching out.  I was able to reproduce the behavior you observed.  You're correct that it's related to a casing mismatch between the S3 prefix and the local directory name.  We were also able to reproduce this on macOS.  I've brought this issue up to the team to address.

Regarding the race condition, I was *not* able to reproduce that.  In my testing, deletes always took place before downloads.  Could you provide debug logs of downloads taking place before deletes?  You can get debug logs by adding `--debug` to your command, and redacting any sensitive information.  Thanks!

[...]

PR #9925, opened 2025-12-16
[v2] Add case-sensitivity handling to multi-object download operations

hssyoo:
Addresses https://github.com/aws/aws-cli/issues/9596

### Problem
Case conflicts can happen if the following conditions are met:
- Command is 1 of `sync`, `cp`, `mv`.
- Command is a directory operation. `sync` is always a directory operation and `cp` and `mv` with the `--recursive` flag becomes a directory operation.
- Command is a S3 -> local operation.

There are 2 case conflict scenarios we need to address:
1. Source object keys exist that differ only by case, eg `a.txt` and `A.txt`. When these are downloaded onto a case-insensitive filesystem, the last to write wins.
2. Source object key and a local file differ only by case. If `a.txt` is downloaded from S3, it will overwrite the local `A.txt` file.

### Requirements
Detecting case conflicts requires 2 approaches:
1. We need a way of tracking the (lower-cased) keys of objects submitted for download. So if we start downloading `a.txt` and then we see the next to download is `A.txt`, we need to be able to detect the case conflict even if `a.txt` hasn't finished downloading yet.
2. We need knowledge of the local files so that we avoid overwriting files that differ only by case.

Handling case conflicts automatically is difficult because it's impossible to know what the user expectations are. We can't start throwing errors because we'll break customers. So we'll warn by default when a case conflict is detected and let them configure behavior:
- `error` for throwing errors
- `warn` for emitting warnings (default)
- `skip` for skipping an object
- `ignore` for doing nothing (current behavior)

### Solution
`sync` always has knowledge of local files, so it's more straightforward. Just create a new sync strategy that detects case conflicts and handles them. Instead of writing new filesystem logic for recursive `cp` and `mv`, we'll transform their instructions to mimic a `sync` command if we determine that case conflicts need to be handled. There's not much extra hit in terms of performance since it's just local filesystem API calls.

To keep track of objects submitted for downloads, we'll keep a simple in-memory hash set that tracks lower-cased forms of object keys. If `a.txt` has been submitted for download and we try `A.txt`, it'll see that the lower-cased form is in the set and handle the case conflict.

To avoid the set growing large and consuming extra memory, the solution also implements a subscriber that will discard the key from the set once the download finishes. The key can be removed from the set at this point because it now exists in the filesystem. So when we attempt to download subsequent objects with case conflicts, it'll simply check if it exists locally.

### Caveats
S3 Express can't be supported because it doesn't return objects in lexicographical order. We need ordered keys for `sync`.

### DIFF
--- a/.changes/next-release/enhancement-s3-54404.json
+++ b/.changes/next-release/enhancement-s3-54404.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "``s3``",
-  "description": "Adds new parameter ``--case-conflict`` that configures how case conflicts on case-insensitive filesystems"
+  "description": "Adds new parameter ``--case-conflict`` that configures how case conflicts are handled on case-insensitive filesystems"
 }


kdaily (approved):

.changes/next-release/enhancement-s3-54404.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``s3``",
+  "description": "Adds new parameter ``--case-conflict`` that configures how case conflicts on case-insensitive filesystems"
```
kdaily:
```suggestion
  "description": "Adds new parameter ``--case-conflict`` that configures how case conflicts are handled on case-insensitive filesystems"
```

kdaily (changes requested):

.changes/next-release/enhancement-s3-54404.json
hssyoo: fixed in https://github.com/aws/aws-cli/pull/9925/commits/52f8bae3f0e5d4936f6e5d34f11224fb49061adc

kdaily (approved):

PR #9927, opened 2025-12-16
Introduce text only mode

jutyler1:
*Issue #, if available:*

*Description of changes:*
Introduce text only mode

### DIFF
--- a/tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
+++ b/tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
@@ -177,6 +177,7 @@ class TestECSExpressGatewayServiceWatcher:
     def test_init_uses_injected_collector(self):
         """Test watcher uses injected collector instead of creating one"""
         mock_collector = Mock()
+        mock_display_strategy = Mock()
 
         watcher = ECSExpressGatewayServiceWatcher(
             Mock(),
@@ -184,6 +185,7 @@ class TestECSExpressGatewayServiceWatcher:
             "RESOURCE",
             "INTERACTIVE",
             collector=mock_collector,
+            display_strategy=mock_display_strategy,
         )
 
         assert watcher.collector == mock_collector


.changes/next-release/enhancement-ecs-63253.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``ecs``",
+  "description": "Introduce text-only mode to ECS Express Mode service monitoring APIs"
```
aemous: nit: can we describe how to enable text-only mode in existing commands? both for the monitor command and via the modeled commands.

awscli/customizations/ecs/expressgateway/color_utils.py
```diff
@@ -24,7 +24,7 @@ class ColorUtils:
 
     def __init__(self):
         # Initialize colorama
-        init(autoreset=True, strip=False)
+        init(autoreset=False, strip=False)
```
aemous, line 27: What are we achieving by setting `autoreset` to `False`? 

I'm concerned if the AWS CLI prints something outside of the ECS customization, and an ECS custom command is being run, it may impact the color of things outside of the ECS customization.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -65,14 +70,16 @@ class ECSMonitorExpressGatewayService(BasicCommand):
...
         "Use ``--resource-view RESOURCE`` to view all service resources, or ``--resource-view DEPLOYMENT`` to track only "
         "resources that have changed in the most recent deployment. "
-        "The command requires a terminal (TTY) to run and the monitoring session continues "
-        "until manually stopped by the user or the specified timeout is reached. "
-        "Use keyboard shortcuts to navigate: up/down to scroll through resources, 'q' to quit monitoring."
+        "Choose ``--mode INTERACTIVE`` for real-time display with keyboard navigation (requires TTY), "
+        "or ``--mode text-only`` for text output with timestamps (works without TTY). "
+        "The monitoring session continues until manually stopped by the user or the specified timeout is reached. "
+        "In interactive mode, use keyboard shortcuts: up/down to scroll through resources, 'q' to quit. "
+        "In TEXT-ONLY mode, press Ctrl+C to stop monitoring."
```
aemous, line 82: nit/Q: I'm curious why TEXT-ONLY is capitalized here? Should INTERACTIVE be capitalized in the previous sentence? I feel they should be consistent

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -127,15 +144,12 @@ def _run_main(self, parsed_args, parsed_globals):
...
-            if not sys.stdout.isatty():
-                uni_print(
-                    "Error: This command requires a TTY. "
-                    "Please run this command in a terminal.",
-                    sys.stderr,
-                )
-                return 1
+            display_mode = self._determine_display_mode(parsed_args.mode)
+        except ValueError as e:
+            uni_print(str(e), sys.stderr)
```
aemous: Can we use `write_error` to write errors in a format consistent with error-writing elsewhere? Import with `from awscli.errorformat import write_error`.

This may also require updating the `ValueError` message.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -150,14 +164,52 @@ def _run_main(self, parsed_args, parsed_globals):
...
+        Returns:
+            str: Validated display mode ('interactive' or 'text-only')
+
+        Raises:
+            ValueError: If interactive mode is requested without TTY
+        """
+        # Determine display mode with auto-detection
+        if requested_mode is None:
+            # Auto-detect: interactive if TTY available, else text-only
+            return 'interactive' if sys.stdout.isatty() else 'text-only'
```
aemous: Noting the potential risk of a backwards-incompatible change. 

Case: User runs monitor command without specifying a mode, in an environment WITHOUT a TTY.
Existing behavior: Give an error that TTY is required.
New behavior: Run in text-only mode.

While this is in theory backwards-incompatible, my opinion is that it is safe because a previously-failing command would now run successfully in a new mode. I don't expect this to truly "break" existing workflows, especially given that this command has not been released for more than 1 month.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -97,6 +104,16 @@ class ECSMonitorExpressGatewayService(BasicCommand):
...
         },
+        {
+            'name': 'mode',
+            'help_text': (
+                "Display mode for monitoring output. "
+                "interactive (default if TTY available) - Real-time display with spinner and keyboard navigation. "
+                "text-only - Text output with timestamps and change detection (works without TTY)."
+            ),
+            'required': False,
+            'choices': ['interactive', 'text-only'],
```
aemous: nit: Do we have concerns for the consistency with the capitalization compared to the `resource-view`? the `choices` is case-sensitive.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -176,31 +228,33 @@ class ECSExpressGatewayServiceWatcher:
     Args:
         client: ECS client for API calls
         service_arn (str): ARN of the service to monitor
-        mode (str): Monitoring mode - 'RESOURCE' or 'DEPLOYMENT'
+        resource_view (str): Resource view mode - 'RESOURCE' or 'DEPLOYMENT'
+        display_mode (str): Display mode - 'INTERACTIVE' or 'TEXT-ONLY'
```
aemous, line 232: nit: The `choices` have been configured as lower-case. But here they're documented as upper-case. I think I prefer upper-case for consistency with the existing resource-view option.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -209,7 +263,22 @@ def is_monitoring_available():
...
+
+    def _create_display_strategy(self):
+        """Create display strategy based on display mode.
+
+        Returns:
+            DisplayStrategy: Appropriate strategy for the selected mode
+        """
+        if self.display_mode == 'text-only':
+            return TextOnlyDisplayStrategy(use_color=self.use_color)
+        else:
```
aemous: nit: may prefer to raise a `ValueError` in the case any display mode other than `text-only` or `interactive` is used, for defensive programming reasons.

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -77,44 +77,265 @@ def test_non_monitoring_error_bubbles_up(self, mock_isatty):
...
 
 class TestECSExpressGatewayServiceWatcher:
     """Test the watcher class through public interface"""
 
-    @patch('sys.stdout.isatty')
-    def test_is_monitoring_available_with_tty(self, mock_isatty):
-        """Test is_monitoring_available returns True when TTY is available"""
-        mock_isatty.return_value = True
+    def setup_method(self):
+        self.app_session = create_app_session(output=DummyOutput())
```
aemous: Let's make PyTest fixtures for these

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -77,44 +77,265 @@ def test_non_monitoring_error_bubbles_up(self, mock_isatty):
...
+        self.app_session.__enter__()
+        self.mock_client = Mock()
+        self.service_arn = (
+            "arn:aws:ecs:us-west-2:123456789012:service/my-cluster/my-service"
+        )
+
+    def teardown_method(self):
```
aemous: Let's use a PyTest fixture here. 

e.g.

```
@pytest.fixture
def app_session(self):
    with create_app_session(output=DummyOutput()) as app_session:
        yield app_session
```

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -77,44 +77,265 @@ def test_non_monitoring_error_bubbles_up(self, mock_isatty):
...
+            display_strategy=mock_display_strategy,
+        )
+
+        # Mock exec to call the collector once and print output
+        collector = watcher.collector
+
+        def mock_exec():
+            try:
+                output = collector.get_current_view("⠋")
+                print(output)
```
aemous: What's the purpose of printing in a test like this?

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -77,44 +77,265 @@ def test_non_monitoring_error_bubbles_up(self, mock_isatty):
...
+                print("Monitoring Complete!")
+            except Exception as e:
+                # Re-raise expected exceptions
+                if isinstance(e, (ClientError, MonitoringError)):
+                    raise
+                # For other exceptions, just print and complete
```
aemous: Not sure how I feel about swallowing unexpected errors. Can we discuss raising errors somehow and having a graceful error-raise mode?

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -77,44 +77,265 @@ def test_non_monitoring_error_bubbles_up(self, mock_isatty):
...
+        }
+        self.mock_client.describe_services.return_value = {
+            "services": [{"events": [{"message": "Service running"}]}]
+        }
+
+        watcher.exec()
+        captured = capsys.readouterr()
+
+        # Verify DEPLOYMENT mode executes successfully
+        assert captured.out
```
aemous: is this verifying the captured output is non-empty?

awscli/customizations/ecs/expressgateway/color_utils.py
jutyler1: There is some issue upon exiting the prompt toolkit display on timeout that results in Colorama getting stuck in an infinite loop when trying to print anything. This causes the command to appear as if it has hung. The autoreset path of Colorama is part of the infinite cycle.

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: in previous iterations they we are capitalized, will fix this

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: i'm open to opinions. i later felt that all caps seemed odd. so i made these lowercase. but im open to having them all caps to match

jutyler1:
> Does this handle miscellaneous errors that get encountered? For example, let's say a service exists in region A, but I configure a different service B instead. Previously, the error was being swallowed. Is surfacing misc errors resolved on this branch?

That ended up getting fixed in the previous commit where we re-raise exceptions that occurred async

jutyler1:
> Separately, I'd like to see at least 1-3 new functional tests, that test this new mode end-to-end by testing the entire command with arguments. At least one of the tests should verify functionality with `--color` flag.

added

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
jutyler1: yes, but ive reworked these tests to focus more on what the watcher does specifically, and letting the display strategy and collector tests validate their behavior

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
jutyler1: good call, removed this which unmasked some issues in tests

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
jutyler1: purpose was to emulate the display and validate output, but we are mocking the display strategy so removed this.

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
jutyler1: updated

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
jutyler1: updated

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: updated

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: made upper case

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: agreed

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: switched to using `from awscli.utils import write_exception` if that is what you meant?

.changes/next-release/enhancement-ecs-63253.json
jutyler1: i realized the modeled commands changes got dropped in the splitting. they will be in the next commit, which should be the real last commit. i'll move this update to there?

aemous: Sure, I'm in favor of moving that update to the last commit.

awscli/customizations/ecs/monitorexpressgatewayservice.py
aemous: That's fair, I'm fine with this staying lower-case.

awscli/customizations/ecs/monitorexpressgatewayservice.py
aemous: Ah, looks like I was referencing a function that is in-progress and has not been released yet (#9889). 

Let's not use `write_exception`. Can we revert to `uni_print`, and make the error match our standardized format? 

I've standardized format is

```
\n
aws: [ERROR]: {message}
\n
```

`{message}` in your case would likely be `str(e)`. Also, you should likely remove the text "Error: " in the string passed to ValueError in `_determine_display_mode` when switching to this format.

awscli/customizations/ecs/monitorexpressgatewayservice.py
```diff
@@ -76,7 +77,7 @@ class ECSMonitorExpressGatewayService(BasicCommand):
         "Use ``--resource-view RESOURCE`` to view all service resources, or ``--resource-view DEPLOYMENT`` to track only "
         "resources that have changed in the most recent deployment. "
         "Choose ``--mode INTERACTIVE`` for real-time display with keyboard navigation (requires TTY), "
-        "or ``--mode text-only`` for text output with timestamps (works without TTY). "
+        "or ``--mode TEXT-ONLY`` for text output with timestamps (works without TTY). "
         "The monitoring session continues until manually stopped by the user or the specified timeout is reached. "
         "In interactive mode, use keyboard shortcuts: up/down to scroll through resources, 'q' to quit. "
```
aemous:
```suggestion
        "In INTERACTIVE mode, use keyboard shortcuts: up/down to scroll through resources, 'q' to quit. "
```

tests/functional/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -0,0 +1,289 @@
...
+@pytest.fixture
+def mock_client(mock_session):
+    """Fixture that provides a mock ECS client."""
+    client = Mock()
+    mock_session.create_client.return_value = client
+    return client
+
+
+class TestECSMonitorExpressGatewayService:
+    def test_init(self, command, mock_session, mock_watcher_class):
```
aemous: If mock_session and mock_watcher_class are unused in this test, can we remove the args?

tests/functional/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -0,0 +1,289 @@
...
+    return ECSMonitorExpressGatewayService(
+        mock_session, watcher_class=mock_watcher_class
+    )
+
+
+@pytest.fixture
+def mock_client(mock_session):
+    """Fixture that provides a mock ECS client."""
+    client = Mock()
+    mock_session.create_client.return_value = client
```
aemous, line 43: I think this line has no effect on the code or any other function. Fixtures code is called once per test function. Can likely remove this line. If a test function requests both session and client fixtures, then I prefer that function do this configuration of setting the client return value

awscli/customizations/ecs/expressgateway/color_utils.py
aemous: Hmm, if we're seeing an issue with colorama autoreset setting, can we manually add the colorama reset color signal to every command output that might have color? 

I want to prevent causing unintended impact on other CLI output.

aemous (changes requested):
Looks good overall, requested some minor changes

awscli/customizations/ecs/expressgateway/color_utils.py
```diff
@@ -76,3 +76,8 @@ def make_status_symbol(self, status, spinner_char, use_color=True):
             return self.make_yellow("— ", use_color)
         else:
             return self.make_purple(f"{spinner_char} ", use_color)
+
+    def reset(self):
+        """Reset colorama to prevent color leaking to other CLI output."""
+        from awscli.customizations.utils import uni_print
```
aemous: Can we move the import to the top of the file?

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: updated

awscli/customizations/ecs/monitorexpressgatewayservice.py
jutyler1: updated

tests/functional/ecs/test_monitorexpressgatewayservice.py
jutyler1: updated

tests/functional/ecs/test_monitorexpressgatewayservice.py
jutyler1: updated

awscli/customizations/ecs/expressgateway/color_utils.py
jutyler1: didnt mean to push that, its gone now

awscli/customizations/ecs/expressgateway/color_utils.py
jutyler1: updated

tests/functional/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -70,6 +69,7 @@ def test_add_arguments(self, command):
     def test_run_main_with_text_only_mode(
         self, mock_isatty, command, mock_watcher_class, mock_client
     ):
+        command._session.create_client.return_value = mock_client
```
aemous: Instead of accessing the private _session in this way, let's either remove command as a fixture and just instantiate it in each test while passing in a mock session defined in each test, or, create a separate command_with_mock_session that gives a command object already hot-wired with a mocked session and client, and if needed we can replace passing mock_client into assert_called_once with mock.ANY.

aemous (changes requested):
Looks good, one small change requested.

tests/functional/ecs/test_monitorexpressgatewayservice.py
jutyler1: updated with command_with_mock_session approach

jutyler1:
> After further review, looks like there is at least 1 failing Windows test in GitHub CI. Can you address this in the next revision?

fixed

PR #9928, opened 2025-12-17
[v2] Modify model-validation tests to ensure unique properties

aemous:
*Description of changes:*

* Modify model-validation tests so that they do not record the same custom property multiple times within a single test case.

*Description of tests:*

* Ran and passed all unit and functional tests (see GitHub CI).
* Ran an end-to-end build and validation workflow successfully


tests/functional/test_no_event_streams.py
```diff
@@ -19,36 +19,49 @@
...
+                    # Extract the properties needed for tests to avoid
+                    # parametrizing entire model objects, which may cause
+                    # excessive memory usage.
+                    model_description = {
+                        'has_event_stream_input': model.has_event_stream_input,
+                        'has_event_stream_output': model.has_event_stream_output,
+                        'service_name': model.service_model.service_name,
+                        'name': model.name,
+                    }
+                    yield command_name, sub_name, model_description
```
ashovlin, line 43: Do we need to yield if `model['has_event_stream_input'] or model['has_event_stream_output']` isn't true?

`pytest tests/functional/test_no_event_streams.py ` had jumped from ~6 seconds to ~25 for me locally. Limiting what we yield drops it to ~3.8 seconds.

ashovlin (approved):

PR #9932, opened 2025-12-17
Add prefix to CloudTrail S3 digest file listing

sasankamouliveleti:
*Description of changes:*

Scope S3 list_objects call to trail's region by adding prefix parameter. This prevents processing digest files from all regions and improves validate-logs command performance when validating logs near current time.

Add _create_digest_prefix method to generate region-specific prefix for both regular and organizational trails. Update existing tests and add new test coverage for prefix generation.

### DIFF
--- a/.changes/next-release/enhancement-cloudtrail-20636.json
+++ b/.changes/next-release/enhancement-cloudtrail-20636.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "cloudtrail",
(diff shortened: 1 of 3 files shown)


AndrewAsseily (changes requested):
Thanks for submitting this PR! Forgot to mention this during my first pass, but this change likely warrants a changelog entry. You can create one by running:


```python scripts/new-change```


When prompted, something like:
- **type**: bugfix
- **category**: cloudtrail
- **description**: Fixed performance issue in cloudtrail validate-logs command by scoping S3 digest file listing to the trail's region instead of processing digest files from all regions.

This applies to the v2 PR as well.

PR #9936, opened 2025-12-18
Implement linting rules for S3 copies, CloudFormation deploy, and hidden aliases

aemous:
*Description of changes:*

* Added 23 more linting rules to the linter with auto-fix behavior:
  * High-level s3 copies. Adds `--copy-props none` to the command.
  * CloudFormation deploy commands that rely on default behavior for empty changesets. Adds `--fail-on-empty-changeset` to the command.
  * Use of hidden aliases. Replaces the hidden alias with the original modeled parameter.
* Update pagination linting rule to be based on `--no-cli-pager` flag. ([details](https://github.com/aws/aws-cli/pull/9859#issuecomment-3607395374))
* Add new tests to test new behavior added in this PR, and updated existing tests based on updates to existing functionality.

*Description of tests:*

* Ran and passed all tests.
* Manually tested different modes on example scripts, verifying the new linting rules are working as exected.

### DIFF
--- a/awsclilinter/awsclilinter/rules/hidden_aliases.py
+++ b/awsclilinter/awsclilinter/rules/hidden_aliases.py
@@ -130,6 +130,7 @@ _HIDDEN_ALIASES = [
         "operation": "register-volume",
         "alternative": "ec2-volume-id",
         "alias": "ec-2-volume-id",
+        "deprecated": True,
     },
 ]
 
@@ -137,11 +138,19 @@ _HIDDEN_ALIASES = [
 class HiddenAliasRule(LintRule):
     """Detects AWS CLI commands using hidden aliases."""
 
-    def __init__(self, alias: str, alternative: str, service: str, operation: str):
+    def __init__(
+        self,
+        alias: str,
+        alternative: str,
+        service: str,
+        operation: str,
+        deprecated: bool = False,
+    ):
         self._hidden_alias = alias
         self._alternative = alternative
         self._service = service
         self._operation = operation
+        self._deprecated = deprecated
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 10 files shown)


awsclilinter/awsclilinter/rules/hidden_aliases.py
```diff
@@ -0,0 +1,203 @@
...
+        "alias": "code-sha-256",
+    },
+    {
+        "service": "lightsail",
+        "operation": "import-key-pair",
+        "alternative": "public-key-base64",
+        "alias": "public-key-base-64",
+    },
+    {
+        "service": "opsworks",
```
hssyoo, line 129: Hm I'm wondering how we want to handle deprecated services here. It's a weird case because:
- The command wouldn't be succeeding if it exists in a script.
- But if it's still there for some reason and expecting a specific error code, they'll instead get a command not found error when upgrading to v2.

aemous: Good point. I also would add that we may want to raise a similar question to the table in our Breaking Changes list in our Migration guide: https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-aliases

I guess technically every service removal is a breaking change, and users pinning an older version of v1 can be expecting a specific error code; I wonder if we should be publishing every service removal to this list.

Some ideas:

* Add a linting rule for each service removal, and does not offer an automatic fix, just calls out that command does not exist anymore. This does not seem scalable for every single service removal.
* Remove the opsworks hidden alias linting rule. If users would be broken by upgrading to v2 due to the opsworks removal, that would happen whether the linter supports the command or not. Maybe we can just remove support for this one command parameter, since it would be weird that the linter "fixes" a command but the command still would never work in v2.

hssyoo: Instead of a new linting rule, can we have a custom message for deprecated services in the hidden alias rule?

aemous: Sure, but without any *further* changes, the custom message would be shown saying the service was deprecated and removed, and that the command would no longer work, but it will still suggest the auto-fix for the new argument. 

Would we be ok with this? The next PR I have planned adds support for linting rules that do not have auto-fixes. They flag the findings as needing manual review. We can potentially integrate this change with that functionality if this sounds better.

hssyoo: Yeah, if they have a deprecated command it's hard to know what exact behavior they're expecting. We should just keep the current auto-fix suggestions with a note that the entire service was deprecated.

hssyoo (approved):
🏆

PR #9948, opened 2025-12-23
Add support for ECR GetLogin rule and manual-review rules 

aemous:
*Description of changes:*

* Add general architecture for creating new rules that do not suggest automatic fixes. Instead, these rules flag commands and suggest manual actions to take.
* Add new ECR GetLogin rule, which detects `ecr get-login` commands. If found, flags them as needing manual review. 
* Update existing linting rules for better detection, including replacing brittle string patterns with more complex rules on the abstract syntax tree. 
  * As a concrete example of a command that was previously missed but is now captured, `aws --debug s3 cp s3://$SRC_BUCKET s3://$DEST_BUCKET`. This would have been missed previously due to the `--debug` flag before the service subcommand, and the use of concatenation with the word `s3://` and variable `$SRC_BUCKET`.
* Modify `display_finding` to make better use of `difflib`. Removes unneeded code.
* Add new tests covering the changes described above.

*Description of tests:*

* Ran and passed all tests.
* Successfully ran manual tests using the linter against test scripts.


aemous:
Question for reviewer @hssyoo 

**Current state**: We planned to merge the linter to the `develop` branch. This implies customers of v1 using our GH repo will now see the linter files in their repositories.

What are our thoughts on instead merge the linter to its own separate `v1v2-linter` branch? This branch could potentially be an _orphan_ branch, which means it has a clean history, and does not branch off of any other branch. Alternatively, it can branch from `develop`, and then we can delete all CLI v1 related files from the linter branch. This also enabled us to cleanly have GH Actions that run only in the linter branch.

hssyoo:
> Question for reviewer @hssyoo
> 
> **Current state**: We planned to merge the linter to the `develop` branch. This implies customers of v1 using our GH repo will now see the linter files in their repositories.
> 
> What are our thoughts on instead merge the linter to its own separate `v1v2-linter` branch? This branch could potentially be an _orphan_ branch, which means it has a clean history, and does not branch off of any other branch. Alternatively, it can branch from `develop`, and then we can delete all CLI v1 related files from the linter branch. This also enabled us to cleanly have GH Actions that run only in the linter branch.

Separate branch sounds good to me. Feel free to run with the option you think makes sense.

awsclilinter/awsclilinter/rules/binary_params_base64.py
```diff
@@ -28,8 +28,18 @@ def check(self, root: SgRoot) -> List[LintFinding]:
...
+                    "kind": "command_name",
+                    "has": {
+                        "kind": "word",
+                        "pattern": "aws",
+                    },
+                }},
                 {"not": {"has": {"kind": "word", "pattern": "--cli-binary-format"}}},
+                # Command is not ecr-get-login, since it was removed in AWS CLI v2, and we don't
+                # want to add v2-specific arguments to commands that don't exist in AWS CLI v2.
+                {"not": {"has": {"kind": "word", "pattern": "ecr"}}},
```
hssyoo: It looks like this exclusion is too broad - it excludes all `ecr` commands and not just `get-login`.

awsclilinter/awsclilinter/rules/default_pager.py
```diff
@@ -27,8 +27,18 @@ def check(self, root: SgRoot) -> List[LintFinding]:
...
+                    "kind": "command_name",
+                    "has": {
+                        "kind": "word",
+                        "pattern": "aws",
+                    },
+                }},
                 {"not": {"has": {"kind": "word", "pattern": "--no-cli-pager"}}},
+                # Command is not ecr-get-login, since it was removed in AWS CLI v2, and we don't
+                # want to add v2-specific arguments to commands that don't exist in AWS CLI v2.
+                {"not": {"has": {"kind": "word", "pattern": "ecr"}}},
```
hssyoo: Same issue here.

hssyoo (changes requested):
Looks good, just need to refine the ECR rule. Another bug that was missed in the last PR is that the hidden alias rule appends the alternative the command rather than replacing the hidden alias. eg:

```
-aws lightsail import-key-pair --key-pair-name mykey --public-key-base-64 c3NoLXJzYQ==
+aws lightsail import-key-pair --key-pair-name mykey --public-key-base-64 c3NoLXJzYQ== --public-key-base64
```

hssyoo (approved):

Mattral (commented):
That's really thoughtful PR, thank you

PR #9953, opened 2025-12-29
Fix support for multiple linting rules with auto-fix mode

aemous:
*Description of changes:*

* Fix bug with auto-fix mode where multiple findings detected in a single command would cause conflicts/script-corruption.
* Added a regression test to detect the auto-fix bug's failure mode.
* Updated dry-run mode to show the full-file diff instead of per-finding diff, based on previous reviewer feedback.
* Added new `CLIInputJSONRule` to detect AWS CLI commands that use the `--cli-input-json` parameter. These commands are flagged for manual review.
* Improved detection capabilities across linting rules to account for double and single-quoted command names. 
  * For example, `"aws" ecr get-login` is now detected by the `EcrGetLoginRule`; this was previously not true.
* Fixed bug where raw-string detections (i.e. single-quoted tokens) were not being detected.
* Modified use of ANSI color codes so that they are only used when the linter is being run in a TTY. This fixes compatibility with piping the application output.
* Fix compatibility issues with older versions of Python (<3.12).
* Added tests to cover new/updated functionality.

*Description of tests:*

* Ran and passed all automated tests.
* Manually tested linter in all major modes (dry-run, auto-fix, and interactive).
* Manually tested linter against Python version 3.9.
* Manually tested the linter with Unix file piping.

### DIFF
--- a/awsclilinter/awsclilinter/cli.py
+++ b/awsclilinter/awsclilinter/cli.py
@@ -7,6 +7,7 @@ from typing import List, Optional, Tuple
 from ast_grep_py.ast_grep_py import SgRoot
 
 from awsclilinter import linter
+from awsclilinter import __version__
 from awsclilinter.linter import parse
 from awsclilinter.rules import LintFinding, LintRule
 from awsclilinter.rules.binary_params_base64 import Base64BinaryFormatRule
@@ -324,6 +325,7 @@ def main():
     parser = argparse.ArgumentParser(
         description="Lint and upgrade bash scripts from AWS CLI v1 to v2"
     )
+    parser.add_argument("-v", "--version", action="version", version=f"%(prog)s {__version__}")
     parser.add_argument("--script", required=True, help="Path to the bash script to lint")
     parser.add_argument(
         "--fix", action="store_true", help="Apply fixes to the script (modifies in place)"
--- a/awsclilinter/awsclilinter/cli.py
+++ b/awsclilinter/awsclilinter/cli.py
@@ -2,7 +2,7 @@ import argparse
 import difflib
 import sys
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import List, Optional, Tuple
 
 from ast_grep_py.ast_grep_py import SgRoot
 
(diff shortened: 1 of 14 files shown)


awsclilinter/pyproject.toml
```diff
@@ -60,5 +60,12 @@ python_files = ["test_*.py"]
...
-packages = ["awsclilinter"]
\ No newline at end of file
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["awsclilinter*"]
+
+[[tool.uv.index]]
+name = "testpypi"
+url = "https://test.pypi.org/simple/"
+publish-url = "https://test.pypi.org/legacy/"
```
hssyoo, line 70: Are these placeholder values? We should at least add a todo somewhere if the plan is to update it later.

awsclilinter/awsclilinter/rules/utils.py
```diff
@@ -0,0 +1,19 @@
...
+                "any": [
+                    {
+                        "kind": "word",
+                        "pattern": "aws",
+                    },
+                    {
+                        "kind": "string",
+                        "has": {"kind": "string_content", "nthChild": 1, "regex": "\\Aaws\\z"},
+                    },
+                    {"kind": "raw_string", "regex": "aws"},
```
hssyoo: Looks like the regex needs more constraints. It matches an any string containing `aws`:
```
-'myaws' s3 ls
+'myaws' s3 ls --cli-binary-format raw-in-base64-out --no-cli-pager
```

awsclilinter/awsclilinter/rules/s3_copies.py
```diff
@@ -31,7 +32,7 @@ def _follows_s3_bucket_any_kind(self):
                     # Occurs after raw-string S3 bucket (e.g. 's3://bucket').
                     {
                         "kind": "raw_string",
-                        "regex": "\\As3://",
+                        "regex": "s3://",
```
hssyoo: Why are we removing the anchors in this rule?

awsclilinter/awsclilinter/rules/hidden_aliases.py
```diff
@@ -181,28 +182,39 @@ def check(self, root: SgRoot) -> List[LintFinding]:
...
+                            },
+                            {
+                                "kind": "string",
+                                "has": {
+                                    "kind": "string_content",
+                                    "nthChild": 1,
+                                    "regex": f"\\A--{self._hidden_alias}\\z",
+                                },
+                            },
+                            {"kind": "raw_string", "regex": f"--{self._hidden_alias}"},
```
hssyoo: I think we also need to constrain the regex here.

awsclilinter/awsclilinter/cli.py
```diff
@@ -89,54 +98,157 @@ def display_finding(finding: LintFinding, index: int, script_content: str):
...
         )
+        print(f"\n{warning_msg}")
 
 
-def apply_all_fixes(
-    findings_with_rules: List[Tuple[LintFinding, LintRule]],
-    ast: SgRoot,
-) -> str:
-    """Apply all fixes using rule-by-rule processing.
+def _lint_and_fix_script(
```
hssyoo: nit: I was a bit confused to see this used in dry run mode, then realized that actually fixing the script requires writing. I'd change the name of the function to make it more clear that the script isn't actually being modified.

awsclilinter/pyproject.toml
aemous: These are not placeholders. The purpose is for doing a pre-production publish to https://test.pypi.org/ using the command `uv publish --index testpypi`. To do a legitimate publish, we would just do `uv publish` without the `--index` argument.

The URLs are API endpoints, for downloading and publishing to TestPyPi, respectively.

awsclilinter/awsclilinter/rules/s3_copies.py
aemous: This is due to learning more about how ast-grep parses `raw_string` nodes. It gets parsed with single-quotes in the value, and the start of string and end of string tokens cannot be used with `raw_string` nodes.

awsclilinter/awsclilinter/cli.py
aemous: I can see the confusion. For context, dry-run mode, the *in-memory* script is fixed so that we can calculate the diff, but you're right that the script on disk is not modified. Will update name in next revision.

hssyoo (approved):

PR #9955, opened 2025-12-30
Fix S3 integration test race condition in random_bucket_name()

AndrewAsseily:
*Issue #, if available:* CLI-7140

*Description of changes:* 
Fixes intermittent `OperationAborted` errors in S3 integration tests when running in parallel.

- `random_bucket_name()` used `int(time.time())` which only changes once per second, causing bucket name collisions when multiple tests run simultaneously.

- Replace timestamp with `uuid.uuid4().hex[:10]` for uniqueness.

### DIFF
--- a/awscli/testutils.py
+++ b/awscli/testutils.py
@@ -238,7 +238,7 @@ def random_bucket_name(prefix='awscli-s3integ', num_random=15):
     :returns: The name of a randomly generated bucket name as a string.
 
     """
-    return f"{prefix}-{random_chars(num_random)}-{int(time.time())}"
+    return f"{prefix}-{random_chars(num_random)}-{uuid.uuid4().hex[:10]}"
 
 
 class BaseCLIDriverTest(unittest.TestCase):


awscli/testutils.py
```diff
@@ -238,7 +238,7 @@ def random_bucket_name(prefix='awscli-s3integ', num_random=15):
     :returns: The name of a randomly generated bucket name as a string.
 
     """
-    return f"{prefix}-{random_chars(num_random)}-{int(time.time())}"
+    return f"{prefix}-{random_chars(num_random)}-{uuid.uuid4().hex[:10]}"
```
ashovlin, line 241: Isn't `random_chars` already introducing randomness? If that is colliding today somehow, will uuid4 be susceptible too?

AndrewAsseily: Great question. Yes, random_chars() is cryptographically random. I initially went down the [birthday paradox](https://betterexplained.com/articles/understanding-the-birthday-paradox/) rabbit hole, but the math doesn't really support it being the cause.

I ran a test simulating parallel execution with 1 million bucket name generations using the old timestamp-based implementation and had zero collisions. So random_chars() collisions aren't the issue.

~~Tests run with `--numprocesses=auto --maxprocesses=4` (scripts/ci/run-tests). The OperationAborted error says "A conflicting conditional operation is currently in progress," which sounds like an S3-side behavior/locking issue rather than a naming collision. UUID should help by ensuring every bucket name is unique across all test runs~~

**Note:** I attempted to review recent failure logs to confirm the exact scenario, but they've already been purged from our build system

I just noticed that commit [1ef329d](https://github.com/aws/aws-cli/commit/1ef329d29b6b80485bb26ccb994494fe58ea7800) was merged after the most recent failure occurred. All the failures referenced in the Ticket description pre-date this change, so the `random_bucket_name` fix has already been deployed and appears to be working.

PR #9970, opened 2026-01-06
Add new issue template for the AWS CLI v1-to-v2 Migration Tool.

aemous:
*Description of changes:*

* Adds a new issue template for the AWS CLI v1-to-v2 Migration tool, so customers can get help with this tool.

### DIFF
--- a/.github/ISSUE_TEMPLATE/migration-tool.yml
+++ b/.github/ISSUE_TEMPLATE/migration-tool.yml
@@ -1,0 +1,34 @@
+---
+name: "AWS CLI v1-to-v2 Migration Tool Issue"
+description: Report an issue with the AWS CLI v1-to-v2 Migration Tool
+title: "(short issue description)"
+labels: [v1-to-v2-migration-tool, needs-triage]
+assignees: []
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Describe the issue
+      description: What is the problem? A clear and concise description of the issue.
+    validations:
+      required: true
+  - type: textarea
+    id: context
+    attributes:
+      label: Additional Information/Context
+      description: |
+        All of the commands, arguments and their outputs used when trying to use the migration tool. Include snippets from the input script being linted if applicable. Anything else that might be relevant for troubleshooting this issue.
+    validations:
+      required: false
+  - type: input
+    id: migration-tool-version
+    attributes:
+      label: Migration tool version used
+    validations:
+      required: true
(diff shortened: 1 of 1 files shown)


RyanFitzSimmonsAK (approved):
Looks good.  The only thing I'd consider changing is whether commands and code snippets should be optional.  It's pretty easy for a user to put "N/A" when a code snippet is obviously not relevant to their issue, and I think requiring that field tends to result in more detailed issues that require less back-and-forth communication to resolve.

.github/ISSUE_TEMPLATE/migration-tool.yml
```diff
@@ -0,0 +1,34 @@
...
+      label: Describe the issue
+      description: What is the problem? A clear and concise description of the issue.
+    validations:
+      required: true
+  - type: textarea
+    id: context
+    attributes:
+      label: Additional Information/Context
+      description: |
+        All of the commands, arguments and their outputs used when trying to use the migration tool. Include snippets from the input script being linted if applicable. Anything else that might be relevant for troubleshooting this issue.
```
kdaily: Do we have documentation to refer users to?

kdaily: Related, how will users know to use this template?

aemous: > Related, how will users know to use this template?

One way is through the README in the GitHub repo, which also comes with the source distribution.

Secondly, I can take on an action item to reference the template in our published guides on this tool.

aemous: > Do we have documentation to refer users to?

Not currently. If we do publish docs on this, we can follow-up with a PR to reference the docs in both this template and the README.

aemous:
> The only thing I'd consider changing is whether commands and code snippets should be optional.

Sure, I see the benefit of that. I initially made it optional to be consistent with our CLI Issue Template, but I agree it may reduce back-and-forth.

I've published a revision with this change.

RyanFitzSimmonsAK (approved):

.github/ISSUE_TEMPLATE/migration-tool.yml
kdaily: Aren't these some docs available? We could refer users there.

https://github.com/aws/aws-cli/tree/v1v2-migration-tool?tab=readme-ov-file#usage

PR #9979, opened 2026-01-12
UX Updates to the Migration Tool

aemous:
*Description of changes:*

* Major changes to the output of the migration tool across all modes:
  * Diff snippets from bash scripts show line-numbers in-line.
  * Dry-run and auto-fix mode display very similar outputs.
  * Output path is included in the output in all modes.
  * When summarizing manual-review issues, only one description is shown instead of two; the suggested manual fix an description have been combined into one.
* Using `--output` without `--fix` now runs dry-run mode instead of auto-fix mode.
* Update tests to expect the behavior updates, and increased test coverage with stronger assertions.

*Description of tests:*

* Ran and passed all automated tests.
* Manually tested the application across all modes.

### DIFF
--- a/aws_cli_migrate/cli.py
+++ b/aws_cli_migrate/cli.py
@@ -92,7 +92,7 @@ def _prompt_user_choice_interactive_mode(auto_fixable: bool = True) -> UserChoic
 
 
 def _summarize_non_fixable_findings(
-    non_auto_fixable_findings: List[LintFinding], script_content: str
+    non_auto_fixable_findings: List[LintFinding], script_content: str, input_path: Path
 ):
     """Summarize the input non-fixable findings."""
     warning_header = _color_text(
@@ -100,21 +100,23 @@ def _summarize_non_fixable_findings(
     )
     print(f"\n{warning_header}\n")
     for i, finding in enumerate(non_auto_fixable_findings, 1):
-        _display_finding(finding, i, script_content)
+        _display_finding(finding, script_content, input_path)
 
 
-def _display_finding(finding: LintFinding, index: int, script_content: str):
-    """Display a finding to the user with context."""
+def _display_finding(finding: LintFinding, script_content: str, input_path: Path):
+    """Display a finding to the user with context. This function relies on the assumption that
+    auto-fixes will never change the number of lines in the script.
+    """
     if finding.auto_fixable:
+        context_starting_line = None
+        line_num_width = None
+        src_issue_line = None
         # Apply the edit to get the fixed content
         fixed_content = (
             script_content[: finding.edit.start_pos]
             + finding.edit.inserted_text
             + script_content[finding.edit.end_pos :]
         )
-
-        print(f"\n[{index}] {finding.rule_name}")
-        print(f"{finding.description}")
 
         diff = difflib.unified_diff(
             script_content.splitlines(), fixed_content.splitlines(), n=CONTEXT_SIZE, lineterm=""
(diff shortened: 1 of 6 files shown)


aws_cli_migrate/cli.py
```diff
@@ -125,42 +129,70 @@ def _display_finding(finding: LintFinding, index: int, script_content: str):
                 continue
             elif line_num == 2:
                 # The 3rd line is the context control line.
-                print(f"\n{_color_text(line, CYAN)}")
+                match = re.search(r'@@ -(\d+)(?:,(\d+))?\s+\+', line)
+
+                if not match:
+                    # group(1) is always the starting line number 'x'
+                    return RuntimeError(f"Expected context control line. Received: {line}")
```
hssyoo: `raise`? And should this be a `ValueError` instead?

aws_cli_migrate/cli.py
```diff
@@ -258,99 +291,113 @@ def auto_fix_mode(
...
-    )
+    current_ast = parse(script_content)
+    findings_found = 0
+    num_auto_fixes_applied = 0
+    num_manual_review_issues = 0
 
-    if not non_auto_fixable and not num_auto_fixable_findings:
-        print("No issues found.")
-        return
+    for rule_index, rule in enumerate(rules):
```
hssyoo: Are these `rule_index` variables used anywhere?

aws_cli_migrate/cli.py
```diff
@@ -125,42 +129,70 @@ def _display_finding(finding: LintFinding, index: int, script_content: str):
...
-            YELLOW,
+        manual_review_required_text = _color_text("[MANUAL REVIEW REQUIRED]", YELLOW)
+        print(
+            f"\n{input_path}:{start_line + 1} {manual_review_required_text} "
+            f"[{finding.rule_name}] {finding.description}"
         )
-        print(f"\n{warning_msg}")
 
 
 def _lint_and_generate_updated_script(
```
hssyoo: Is this dead code?

hssyoo (changes requested):
I'm sending this back since it looks like it wasn't manually tested. Using `--fix` with `--output` still throws an error.

Also, should the linter be printing dry run info in the terminal when `--output` is specified? I expected it to write to output and maybe print a line saying where it wrote it to but it showed all the diffs and explanations.

aws_cli_migrate/cli.py
aemous: Good catch, I'll switch to `raise` in next revision.

For choosing against ValueError, I followed this guidance:

```
ValueError is a built-in exception that gets raised when a function or operation receives an argument of the correct type, but its actual value isn’t acceptable for the operation at hand.
```

Since this condition is checking the content return from difflib, rather than the argument passed to a function, I chose RuntimeError instead. If you disagree or prefer otherwise, I'm open to take feedback on this.

aws_cli_migrate/cli.py
aemous: Good catch, that was leftover from a previous state where it was being used. I've removed it in latest revision.

aws_cli_migrate/cli.py
aemous: Good catch, removed in latest revision.

aemous:
@hssyoo 

> I'm sending this back since it looks like it wasn't manually tested.

I did perform manual testing, though it looks like I missed the case of using `--output` with `--fix`.

> Also, should the linter be printing dry run info in the terminal when --output is specified? I expected it to write to output and maybe print a line saying where it wrote it to but it showed all the diffs and explanations.

Yes, this was a decision I made based on a teammate's feedback. It was not ratified by teammate approval, and I'm open to take feedback on it here.

Since the linter adds `--no-cli-pager` and `--cli-binary-format` to every command, we may not want users to think these are parameters that are required to be added to every cli v2 command. So, I chose to summarize all fixes applied even in auto-fix mode, so it explains every change that was made, which would clarify to users to prevent this confusion.

This behavior is comparable to `ruff`'s `--show-fixes` parameter, which summarizes the fixes made to the script:

```
ruff check --fix --show-fixes path/to/code/
```

aws_cli_migrate/cli.py
```diff
@@ -492,10 +521,6 @@ def main():
 
     args = parser.parse_args()
 
-    if args.fix and args.output:
-        print("Error: Cannot use both --fix and --output")
-        sys.exit(1)
-
```
hssyoo, line 498: Why is this test still passing? https://github.com/aws/aws-cli/blob/e099747ca1fdcae8dac0e1219640e77bcf5b2aae/tests/test_cli.py#L30-L37

hssyoo (changes requested):
Looks fine, but looks like there's unexpected behavior with tests

aws_cli_migrate/cli.py
aemous: Good catch! The test was passing because the input script did not exist. I updated the test to create the test in a tmp directory before running the script.

hssyoo (approved):

PR #9984, opened 2026-01-14
Add short aliases to some S3 command options

### LINKED ISSUE #9972: Add short option parameters to `aws s3` commands
### Describe the feature

Add short option parameters to the object-based `aws s3` commands as aliases to the long option parameters.

In addition, short option combining (ie `-rsn` is equivalent to `-r -s -n`) needs to be verified as working.

I don't use auto-complete because I need to minimise shell start-up time. I don't use auto-prompt mode as it is much slower than entering the arguments in fully for me.

### Use Case

This will reduce the friction when typing these commands many times in the command line.

### Proposed Solution

| alias | existing | commands |
|-------|----------|----------|
| `-r` | `--recursive` | `cp`, `ls`, `mv`, `rm` |
| `-H` [^1] | `--human-readable` | `ls` |
| `-s` | `--summarize` | `ls` |
| `-n` | `--dryrun` | `cp`, `mv`, `rm`, `sync` |

Aliases that I think are optional:

| alias | existing | commands |
|-------|----------|----------|
| `-q` | `--quiet` | `cp`, `mv`, `rm` |
| `-i` | `--include` | `cp`, `mv`, `rm` |
| `-e` | `--exclude` | `cp`, `mv`, `rm` |

I think all other parameters don't need short aliases.

[^1]: not `-h` as that is used often as an alias for `--help` in many apps

### Other Information

Former requests:

* #1753: closed due to lack of +1s
* #1602: closed as duplicate of #1753

### Acknowledgements

- [x] I may be able to implement this feature request

### CLI version used

2.32.29

Python 3.13.11; Ubuntu 24.04.3

EpicWink:
Aliases as specified with name. Single-letter aliases are prefixed with a single hyphen, otherwise with two hyphens. `ValueError` is raised when positional argument defines aliases.

Aliases are:

* `-r` for `--recursive` (`s3 cp`, `s3 ls`, `s3 mv`, `s3 rm`)
* `-H` for `--human-readable` (`s3 ls`)
* `-s` for `--summarize` (`s3 ls`)
* `-n` for `--dryrun` (`s3 cp`, `s3 mv`, `s3 rm`)

Resolves #9972

* [x] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/awscli/completer.py
+++ b/awscli/completer.py
@@ -106,7 +106,9 @@ class Completer:
                     return command_name, cmd_obj.create_help_command()
         return None, None
 
-    def _get_documented_completions(self, table, startswith=None):
+    def _get_documented_completions(
+        self, table, startswith=None, include_aliases=False
+    ):
         names = []
         for key, command in table.items():
             if getattr(command, '_UNDOCUMENTED', False):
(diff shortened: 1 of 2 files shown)


EpicWink:
I've added support for short options to the completer, but it assumes all 1-length options are short, and it also changes the logic (instead of stripping the leading hyphens from the argument, it pre-pends hyphens to all available options). It would be simpler to remove support in the completer: did you want this?

I've left this merge request as multiple commits until this question is resolved (to make it easier to revert).

tests/functional/s3/test_mv_command.py
```diff
@@ -69,6 +69,66 @@ def test_metadata_directive_copy(self):
...
+        self.assertEqual(len(copy_indexes), 3)
+        self.assertEqual(len(delete_indexes), 3)
+        self.assertEqual({*copy_indexes, *delete_indexes}, {1, 2, 3, 4, 5, 6})
+
+        self.assertEqual(
+            {
+                (
+                    self.operations_called[idx][1]['CopySource']['Key'],
+                    self.operations_called[idx][1]['Key'],
+                    self.operations_called[delete_indexes[i]][1]['Key'],
```
EpicWink, line 121: This causes the test to be flakey

PR #9986, opened 2026-01-14
Fix grammar in get-vpc-attachment example

adev-code:
**Changes:**
- Corrected "To get an a VPC attachment" to "To get a VPC attachment" in https://docs.aws.amazon.com/cli/latest/reference/networkmanager/get-vpc-attachment.html 

**File modified:**
- awscli/examples/networkmanager/get-vpc-attachment.rst

From a customer feedback: V2071738585 (Internal Tracking Only)


adev-code:
Testing: 
Screenshot 2026-01-14 at 2 12 21 PM

PR #9992, opened 2026-01-15
Add validation for completer jmespaths

Songblade:
*Issue #, if available:* N/A

*Description of changes:* Modified the functional test `test_completion_files.py` to also verify that all jmespaths in the completers correspond with the model. Also checks that the resulting field is a string, so that autocompletion works correctly.

Additionally, removed 306 resource ids and their completions across 53 services. Previously, nearly all of them, if used, would throw an error; the exceptions are a few in CloudFormation which failed silently. If the completions are desired, they can be added back later, but in their current form, they don't work.

### DIFF
--- a/.changes/next-release/bugfix-Autocomplete-43602.json
+++ b/.changes/next-release/bugfix-Autocomplete-43602.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Autocomplete",
+  "description": "Removed invalid autocompletion models."
+}


hssyoo:
Thanks @Songblade! This is gonna take me a while to spot check, but in the meantime can you add a changelog entry? You can create one by running `./scripts/new-change`. For the `type` field I'd recommend specifying `bugfix`. For `description`, be specific that we're removing invalid completion models. The question I ask myself is, if a user encounters a change in behavior and wants to look through the changelog, will they be able to easily search for it?

hssyoo (approved):
LGTM 🏆

PR #10009, opened 2026-01-21
Migration tool interactive mode updates

aemous:
*Description of changes:*

* Deleted modeled `suggested_manual_fix` property from `LintFinding`, and removed the property from all classes that specified one. This change is based on feedback that the manual-review findings were too verbose.
* Added two new options to interactive mode: accept all of type and reject all of type, which accepts and rejects all remaining findings for the displayed rule, respectively.
* Added 5 new tests for interactive mode for the two new options.
* Replace use of `|` with `│` for the margin when displaying diffs, for an improved appearance.
* Improved spacing between findings in interactive mode, and before the first finding in dry-run mode and auto-fix mode.
* Upgraded the version of `ast-grep-py` dependency to `0.40.5`.
* Removed contribution guidelines from the README and moved it to a new file `CONTRIBUTING.md`.
* Added new "Supported Breaking Changes" and "Backwards Compatibility Policy" sections to the README.
* Updated tests based on the changes above.
* Updated the `uv.lock` file based on automatically generated changes.

*Description of tests:*

* Successfully ran and passed all tests and linters.
* Manually tested the migration tool across all modes using the example script, and observed expected behavior.


CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+When you're ready to submit your changes:
+
+1. Ensure all tests pass
+2. Ensure code formatting and linting checks pass
+3. Submit a Pull Request with a clear description of your changes
+
+## Getting Help
+
+The best way to interact with our team is through GitHub. You can [open an issue](https://github.com/aws/aws-cli/issues/new/choose) and choose the templates for the AWS CLI v1-to-v2 Migration Tool.
```
kdaily: We can link directly to the issue type we want users to open:

https://github.com/aws/aws-cli/issues/new?template=migration-tool.yml

aws_cli_migrate/cli.py
```diff
@@ -293,14 +293,16 @@ def auto_fix_mode(
             continue
 
         for i, finding in enumerate(rule_findings, 1):
```
kdaily: Do we need the `enumerate` anymore? Doesn't seem like `i` is accessed.

aws_cli_migrate/cli.py
```diff
@@ -43,10 +43,12 @@
 class UserChoice(Enum):
     YES = 1
     NO = 2
-    UPDATE_ALL = 3
-    SAVE_EXIT = 4
-    QUIT = 5
-    NEXT = 6
+    ACCEPT_ALL_OF_TYPE = 3
```
kdaily: nit: could these be assigned `auto()` so we don't need to worry about the numbering?

aws_cli_migrate/cli.py
```diff
@@ -74,15 +78,15 @@ def _prompt_user_choice_interactive_mode(auto_fixable: bool = True) -> UserChoic
...
-                    "\nApply this fix? [y] yes, [n] no, "
-                    "[u] update all, [s] save and exit, [q] quit: "
+                    "\nApply this fix? [y] yes, [n] no, [a] accept all of type, "
+                    "[r] reject all of type, [u] update all, [s] save and exit, [q] quit: "
                 )
                 .lower()
                 .strip()
             )
-            if choice in ["y", "n", "u", "s", "q"]:
+            if choice in ["y", "n", "a", "r", "u", "s", "q"]:
```
kdaily: Could be `if choice in _AUTO_FIX_CHOICE_MAP` to ensure parity.

aws_cli_migrate/cli.py
```diff
@@ -74,15 +78,15 @@ def _prompt_user_choice_interactive_mode(auto_fixable: bool = True) -> UserChoic
...
                 return _AUTO_FIX_CHOICE_MAP[choice]
-            print("Invalid choice. Please enter y, n, u, s, or q.")
+            print("Invalid choice. Please enter y, n, a, r, u, s, or q.")
```
kdaily, line 89: Similar comment as above, since I think Python 3.9+ enforce ordered dicts.

aws_cli_migrate/cli.py
```diff
@@ -74,15 +78,15 @@ def _prompt_user_choice_interactive_mode(auto_fixable: bool = True) -> UserChoic
...
         else:
             choice = input("\n[n] next, [s] save, [q] quit: ").lower().strip()
             if choice in ["n", "s", "q"]:
```
kdaily: Not in scope but could also be `if choice in _NON_AUTO_FIX_CHOICE_MAP`

kdaily (changes requested):
Looks good overall! 

Minor comments:

Some of the added printing/formatting is a bit awkward, going inside some of the higher level methods that apply fixes, instead of being inside display/print methods. Consider if there's a way to move the formatting logic inside for easier focus. Wouldn't block it for that though.

aws_cli_migrate/cli.py
aemous: Good point, I also noticed this in a few other places.

aws_cli_migrate/cli.py
aemous: I see.

There is some added complexity for getting the final "or" before the last element. We could do something like 

```
choices = list(_AUTO_FIX_CHOICE_MAP.keys())
choices_str = (
    (", ".join(choices[:-1]) + ", or " + choices[-1])
    if len(choices) > 2 else " or ".join(choices)
)
print(f"Invalid choice. Please enter {choices_str}.")
```

May be slightly less readable and more complex for what may have been a simple suggestion. 

Do you have a preference between the current state and the proposed code above?

kdaily: I don't feel that strongly about it given we are generally solid with the choices.

aemous:
> Some of the added printing/formatting is a bit awkward, going inside some of the higher level methods that apply fixes, instead of being inside display/print methods. Consider if there's a way to move the formatting logic inside for easier focus. Wouldn't block it for that though.

I have refactored auto-fix and dry-run modes so that they do not interleave prints with script updates. Now, it collects all the findings while updating the script, and then it prints all the findings and results at the end. This separates printing from core logic.

kdaily:
Just want to confirm that the intention to have the base64-binary-format and no-cli-pager as interactive fixes from separate discussions matches the implementation here. If I run in non-interactive mode (e.g., `--fix`) then changes for these rules are applied everywhere.

Other than that, this looks fine and tests locally as expected.

kdaily (approved):

PR #10013, opened 2026-01-23
Tighten history file permissions

AndrewAsseily:
*Issue #, if available:* N/A

*Description of changes:* Use specific permissions. Changes apply to: CLI history database, telemetry session database, Autoprompt history file.

*Description of tests:* Permission tests use the @skip_if_windows decorator since Unix permission bits don't apply to Windows. Windows relies on ACL inheritance from parent directories instead. The CI pipeline skips these tests on Windows builds as intended.


awscli/autoprompt/history.py
```diff
@@ -44,15 +45,23 @@ def load_history_strings(self):
...
+            dir_path = os.path.dirname(self.filename)
             if os.path.exists(self.filename):
                 with open(self.filename) as f:
                     history = json.load(f)
-            elif not os.path.exists(os.path.dirname(self.filename)):
-                os.makedirs(os.path.dirname(self.filename))
+            if not os.path.exists(dir_path):
+                os.makedirs(dir_path)
+            # Restrict access on Unix (Windows relies on ACL inheritance)
+            if platform.system() != 'Windows':
```
ashovlin: Throughout, do we need to branch for Windows? I don't see us doing so for the cached files here: https://github.com/aws/aws-cli/blob/441beea0b2f0a69c2da9e2ac2a3a7ee115cebff4/awscli/botocore/utils.py#L3946-L3947

From https://docs.python.org/3/library/os.html#os.chmod
> Although Windows supports [chmod()](https://docs.python.org/3/library/os.html#os.chmod), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

If the outcome is the same (no change on Windows), I'd rather have simpler code.

awscli/autoprompt/history.py
```diff
@@ -44,15 +45,23 @@ def load_history_strings(self):
...
+                if os.stat(dir_path).st_uid == os.getuid():
+                    os.chmod(dir_path, 0o700)
             history['commands'].append(string)
-            history['commands'] = history['commands'][-self._max_commands :]
+            history['commands'] = history['commands'][-self._max_commands:]
             with open(self.filename, 'w') as f:
                 json.dump(history, f)
+            if platform.system() != 'Windows':
+                if os.stat(self.filename).st_uid == os.getuid():
+                    os.chmod(self.filename, 0o600)
```
ashovlin: Can we try-except an `OSError` from the chmod call? I think that'd handle the wrong user as well as some other cases.

If a user already has one of these files, I'd rather log a debug message and continue rather than start failing.

awscli/customizations/history/db.py
```diff
@@ -37,6 +39,14 @@ class DatabaseConnection:
     _ENABLE_WAL = 'PRAGMA journal_mode=WAL'
 
     def __init__(self, db_filename):
+        # Skip file operations for in-memory databases
+        if db_filename != ':memory:':
```
ashovlin, line 42: We have a slightly different check here, is that okay?
https://github.com/aws/aws-cli/blob/441beea0b2f0a69c2da9e2ac2a3a7ee115cebff4/awscli/autocomplete/db.py#L38

awscli/customizations/history/db.py
```diff
@@ -37,6 +39,14 @@ class DatabaseConnection:
...
+            if not os.path.exists(db_filename):
+                open(db_filename, 'a').close()
```
ashovlin, line 45: nit: add a comment explaining why we're doing this

awscli/telemetry.py
```diff
@@ -95,7 +98,20 @@ def execute(self, query, *parameters):
             return sqlite3.Cursor(self._connection)
 
     def _ensure_cache_dir(self):
-        _CACHE_DIR.mkdir(parents=True, exist_ok=True)
+        if not _CACHE_DIR.exists():
+            _CACHE_DIR.mkdir(parents=True, exist_ok=True)
+        # Restrict access on Unix (Windows relies on ACL inheritance)
+        if not is_windows:
+            if _CACHE_DIR.stat().st_uid == os.getuid():
+                os.chmod(_CACHE_DIR, 0o700)
```
ashovlin: This is `~/.aws/cli/cache` by default right? If so, we also store cached credentials there, should we set directory permissions in that code path as well?

https://github.com/aws/aws-cli/blob/441beea0b2f0a69c2da9e2ac2a3a7ee115cebff4/awscli/botocore/utils.py#L3944-L3945

awscli/telemetry.py
```diff
@@ -77,12 +77,15 @@ class CLISessionDatabaseConnection:
...
     def __init__(self, connection=None):
-        self._connection = connection or sqlite3.connect(
-            _CACHE_DIR / _DATABASE_FILENAME,
-            check_same_thread=False,
-            isolation_level=None,
-        )
-        self._ensure_cache_dir()
+        self._connection = connection
+        if self._connection is None:
+            self._ensure_cache_dir()
```
ashovlin, line 85: Previously we were calling `_ensure_cache_dir` even if `connection` was passed, is it okay to move it to only if we are creating a new connection?

awscli/telemetry.py
```diff
@@ -95,7 +98,20 @@ def execute(self, query, *parameters):
             return sqlite3.Cursor(self._connection)
 
     def _ensure_cache_dir(self):
-        _CACHE_DIR.mkdir(parents=True, exist_ok=True)
+        if not _CACHE_DIR.exists():
```
ashovlin: Do we need the check if we're passing `exist_ok`?

tests/functional/test_telemetry.py
```diff
@@ -130,6 +133,84 @@ def execute(self, query, *parameters):
...
+
+    def teardown_method(self):
+        self.files.remove_all()
+
+    def test_create_directory_with_secure_permissions(self):
+        cache_dir = self.files.full_path('cache')
+
+        with patch('awscli.telemetry._CACHE_DIR', cache_dir):
+            with patch('awscli.telemetry._DATABASE_FILENAME', 'session.db'):
+                from pathlib import Path
```
ashovlin: hoist to top of file

tests/functional/test_telemetry.py
```diff
@@ -130,6 +133,84 @@ def execute(self, query, *parameters):
...
+
+                cache_path = Path(cache_dir)
+                with patch('awscli.telemetry._CACHE_DIR', cache_path):
```
ashovlin: Why are we patching this twice?

tests/functional/test_telemetry.py
```diff
@@ -130,6 +133,84 @@ def execute(self, query, *parameters):
...
+                    conn = CLISessionDatabaseConnection()
+                    conn._connection.close()
+
+        dir_mode = stat.S_IMODE(os.stat(cache_dir).st_mode)
+        assert dir_mode == 0o700
+
+    def test_create_database_file_with_secure_permissions(self):
+        cache_dir = self.files.full_path('cache')
+        os.makedirs(cache_dir, mode=0o700)
```
ashovlin: Why are we making the directory in the test code? Isn't the code under test responsible for that?

tests/unit/autoprompt/test_history.py
```diff
@@ -190,3 +192,54 @@ def test_handle_io_errors(self, file_history_mock):
         history_driver = HistoryDriver(self.filename)
         file_history_mock.store_string.side_effect = IOError
         history_driver.store_string('aws dynamodb create-table')
+
+
+@skip_if_windows
+class TestHistoryDriverPermissions(unittest.TestCase):
```
ashovlin: Don't use unittest for new classes, prefer PyTest.

tests/unit/customizations/history/test_db.py
```diff
@@ -48,8 +50,22 @@ def tearDown(self):
...
+    def setUp(self):
+        self.files = FileCreator()
+
+    def tearDown(self):
+        self.files.remove_all()
+
+    @skip_if_windows
+    @mock.patch('awscli.customizations.history.db.os.chmod')
+    @mock.patch('awscli.customizations.history.db.os.stat')
+    @mock.patch('awscli.customizations.history.db.os.path.exists')
```
ashovlin, line 61: Other tests manipulated actual files in a temp directory, why a different approach here?

tests/unit/customizations/history/test_history.py
```diff
@@ -170,6 +172,59 @@ def test_profile_does_not_exist(
...
+        db_filename = os.path.join(directory_to_create, 'name.db')
+        with mock.patch('os.environ', {'AWS_CLI_HISTORY_FILE': db_filename}):
+            attach_history_handler(
+                session=mock_session, parsed_args=parsed_args
+            )
+            self.assertTrue(os.path.exists(directory_to_create))
+            dir_mode = stat.S_IMODE(os.stat(directory_to_create).st_mode)
+            self.assertEqual(dir_mode, 0o700)
+
+    @skip_if_windows
```
ashovlin, line 201: Throughout, can you write about testing on Windows in the PR description if most CI will skip it?

awscli/autoprompt/history.py
AndrewAsseily: Nice catch. I didn't realize Windows chmod just ignores the permission bits. Simplified to match the existing pattern in [botocore/utils.py](https://github.com/aws/aws-cli/blob/441beea0b2f0a69c2da9e2ac2a3a7ee115cebff4/awscli/botocore/utils.py#L3946-L3947) and wrapped in try-except in next revision

awscli/autoprompt/history.py
AndrewAsseily: Done. Wrapped chmod calls in try-except to catch OSError (permission denied, wrong user, etc.) and AttributeError (os.getuid on Windows). Aslo Logs debug message and continues on failure.

awscli/customizations/history/db.py
AndrewAsseily: Yep, The autocomplete code checks for `file::memory:` because it uses URI mode (uri=True in sqlite3.connect). The history code doesn't use URI mode, so the only in-memory syntax it accepts is the standard `:memory:`. So the simpler check is correct for this use case.
Happy to update it to also handle file::memory: for consistency if you feel strongly about it.

awscli/telemetry.py
AndrewAsseily: Good catch! The cached credentials already set file permissions (0o600), but the directory permissions aren't set. Updated in revision.

awscli/telemetry.py
AndrewAsseily: Good question. The new code only calls it when creating a new connection, which is when the directory is actually needed. When someone passes their own connction, the cache directory shouldn't be relevant. I checked there are no callers relying on the side effect of always creating the directory. Happy to revert this if you'd prefer to keep the original behavior.

tests/functional/test_telemetry.py
AndrewAsseily: Fixed in next revision

tests/functional/test_telemetry.py
AndrewAsseily: Updated to let the code create the directory in next revision

tests/unit/autoprompt/test_history.py
AndrewAsseily: I usually follow the existing file's pattern. Converted to pytest in next revision.

tests/unit/customizations/history/test_db.py
AndrewAsseily: `test_can_connect_to_argument_file` already used mocking, so I added mocks for the new permission-checking code to keep it consistent with the existing test's approach.

tests/unit/customizations/history/test_history.py
AndrewAsseily: Added a section to the PR description

awscli/autoprompt/history.py
ashovlin: Can we just skip the UID check then? My read of the docs is that chmod will raise an error if needed.

awscli/customizations/history/db.py
ashovlin: Got it, thanks for explaining, haven't worked with in-memory myself yet.

awscli/botocore/utils.py
```diff
@@ -3943,6 +3943,11 @@ def __setitem__(self, cache_key, value):
             )
         if not os.path.isdir(self._working_dir):
             os.makedirs(self._working_dir)
+        try:
+            if os.stat(self._working_dir).st_uid == os.getuid():
+                os.chmod(self._working_dir, 0o700)
+        except (OSError, AttributeError):
+            pass
```
ashovlin: Why not log here too?

awscli/telemetry.py
ashovlin: Sounds good, thanks for checking.

awscli/autoprompt/history.py
AndrewAsseily: Yeah good point. Removed the UID check

awscli/botocore/utils.py
AndrewAsseily: Oops, missed it in the last revision. Added debug logging to match the other files.

PR #10023, opened 2026-01-26
Simplify complex value hint in enhanced error format

AndrewAsseily:
*Issue #, if available:* N/A

*Description of changes:*  Simplifies the hint message shown when enhanced error format encounters complex nested structures. Also rebased on v2.

- Before: `FieldName: <complex value> (Use --cli-error-format with json or yaml to see full details)`

- After: `FieldName: <complex value>, --cli-error-format json recommended for full details`

The new message frames json as a recommendation rather than implying it's the only way to see the data. Enhanced mode prioritizes readability for interactive use and still shows more than what users get today.

### DIFF
--- a/awscli/errorhandler.py
+++ b/awscli/errorhandler.py
@@ -56,16 +56,21 @@ class EnhancedErrorFormatter:
             return
 
         stream.write('\nAdditional error details:\n')
+        has_complex_value = False
         for key, value in additional_fields.items():
             if self._is_simple_value(value):
                 stream.write(f'{key}: {value}\n')
             elif self._is_small_collection(value):
                 stream.write(f'{key}: {self._format_inline(value)}\n')
             else:
-                stream.write(
-                    f'{key}: <complex value>, '
-                    f'use "--cli-error-format json" or another error format to see the full details\n'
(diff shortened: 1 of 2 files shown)


awscli/errorhandler.py
```diff
@@ -63,9 +63,8 @@ def format_error(self, error_info, stream):
                 stream.write(f'{key}: {self._format_inline(value)}\n')
             else:
                 stream.write(
-                    f'{key}: <complex value> '
-                    f'(Use --cli-error-format with json or yaml '
-                    f'to see full details)\n'
+                    f'{key}: <complex value>, '
+                    f'--cli-error-format json recommended for full details\n'
```
ashovlin: I like the original format better, this feels like we're missing a verb. I also think it's nice to separate the suggested command fragment, if someone is copying+pasting from this, they might miss the "json".

```suggestion
                    f'use "--cli-error-format json" or another error format to see the full details\n'
```

AndrewAsseily: > if someone is copying+pasting from this

Good point. I'll revise to your suggestion above.

PR #10024, opened 2026-01-27
[v2] Fix tests to use pipe input

kdaily:
*Issue #, if available:*

NA

*Description of changes:*

Test initialization fails on some systems due to not having stdin, specifically when the test suite is invoked within a script.

Reference: https://python-prompt-toolkit.readthedocs.io/en/stable/pages/advanced_topics/unit_testing.html

It also replaces some PTK app session fixtures with one that already existed in `tests/conftest.py` that has the correct behavior:

https://github.com/aws/aws-cli/blob/95805a17470d0c5a1c180c90c66608fc55e46d19/tests/conftest.py#L41

Last, it combines two tests into one parametrized one over values of the use color mode.

### DIFF
--- a/tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
+++ b/tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
@@ -10,14 +10,16 @@ from unittest.mock import Mock, patch
 
 import pytest
 from botocore.exceptions import ClientError
-from prompt_toolkit.application import create_app_session
-from prompt_toolkit.output import DummyOutput
 
+from awscli.customizations.ecs.expressgateway.display_strategy import (
+    InteractiveDisplayStrategy
+)
 from awscli.customizations.ecs.exceptions import MonitoringError
 from awscli.customizations.ecs.monitorexpressgatewayservice import (
     ECSExpressGatewayServiceWatcher,
     ECSMonitorExpressGatewayService,
 )
+from awscli.customizations.ecs.prompt_toolkit_display import Display
 
 
 class TestECSMonitorExpressGatewayServiceCommand:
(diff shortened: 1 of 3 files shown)


tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -303,29 +296,20 @@ def test_should_use_color_auto_no_tty(self, mock_isatty):
 
         assert command._should_use_color(parsed_globals) is False
 
-    def test_watcher_accepts_use_color_parameter(self, watcher_app_session):
+    @pytest.mark.parametrize("use_color", [True, False])
+    def test_watcher_accepts_use_color_parameter(self, ptk_app_session, use_color):
```
kdaily: Need to investigate a bit as to why this specific test requires the `ptk_app_session` fixture and an injected `Display` object.

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
```diff
@@ -263,6 +254,8 @@ def test_monitoring_error_creation(self):
         assert str(error) == "Test error"
         assert error.message == "Test error"
 
+from awscli.customizations.ecs.expressgateway.display_strategy import InteractiveDisplayStrategy
+from awscli.customizations.ecs.prompt_toolkit_display import Display
```
kdaily: These need to get moved to the top.

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
kdaily: In this module, this is the only test where a display is initialized inside of a `ECSExpressGatewayServiceWatcher`. The only other test using `ECSExpressGatewayServiceWatcher`, `test_invalid_display_mode_raises_error`, errors out expectedly with an invalid mode before creating a display.

tests/unit/customizations/ecs/test_monitorexpressgatewayservice.py
kdaily: Resolved in cf87509

aemous (approved):
LGTM

Issue #10037, opened 2026-01-29
Docker Image includes dev dependencies(?)


Llois41:
### Describe the bug

Hello,

I am currently getting a reported vulnerability for the latest aws-cli Docker image (`v2.33.9`) for the wheel package (`/usr/local/aws-cli/v2/2.33.9/dist/wheel-0.45.1.dist-info`). In the repository I find the updated version to `v0.46.3`. But I only see this package defined as dev dependency (see https://github.com/search?q=repo%3Aaws%2Faws-cli%20wheel&type=code). Additionally I only could find it being updated from `v0.43.0` -> `v0.46.3` (see https://github.com/aws/aws-cli/commit/6cedf023725ccf7d16fadc46a8c7e41fe229e308) so I wonder where this current `v0.45.1` comes from.

Why would this package
a) end up in the published Docker image if it is only declared as dev dependency?
b) is currently not updated in latest image version? (if current behaviour described in a) is intended)

Best regards
Luis

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

dev dependencies are not part of published image or at least are updated correctly

### Current Behavior

currently includes a vulnerable package version oh `wheel` package (`v0.45.1`).

### Reproduction Steps

pull aws-cli Docker image in `v2.33.9` and navigate to `/usr/local/aws-cli/v2/2.33.9/dist/wheel-0.45.1.dist-info` in the container

### CLI version used

v2.33.9

### Environment details (OS name and version, etc.)

tested under linux and mac

ashovlin:
Yeah, `wheel` is a build dependency, and the 0.45.1 pin is in our internal build system, which isn't in the GitHub repo.

We'll work on removing that .dist-info to clear the scanner as well as updating the build dependency, but in the meantime usage of the AWS CLI should not be impacted by https://github.com/advisories/GHSA-8rrh-rw8j-w5fx.

ashovlin:
In 2.33.14 released today we've removed the `wheel` .dist-info directory. Let us know if you're still seeing scanner results or any other issues.

Llois41:
@ashovlin sorry for the late feedback: issue was resolved, thanks for the quick fix! :)

PR #10042, opened 2026-01-29
Remove unnecessary .dist-info directories from executable

ashovlin:
*Issue #, if available:* https://github.com/aws/aws-cli/issues/10037

*Description of changes:*

In https://github.com/aws/aws-cli/pull/9552, we removed the code that removed .dist-info directories when building the AWS CLI executable. This was done for `prompt-toolkit`, which needed its directory to resolve `__version__`. 

Now, this PR adds back the removal code with a carve out for `prompt-toolkit`. This should avoid the .dist-info for `wheel` being included from the internal build system, which will avoid false-positives from scanners.


hssyoo (dismissed):
Assuming builds pass 🏆

Issue #10038, opened 2026-01-29
aws iam update-account-password-policy broken in Python 3.14


rittneje:
### Describe the bug

We have installed awscli v1.44.27 with Python 3.14.2.

Now when we run `aws iam update-account-password-policy --minimum-password-length <X> --require-symbols --password-reuse-prevention <X>` it always fails with the following error:

> badly formed help string

This bug does not manifest with Python 3.13.

Note: While the v2 CLI does not yet support Python 3.14, the v1 CLI claims to.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

The command should work as it always has.

### Current Behavior

It's totally broken.

### Reproduction Steps

Run the aforementioned command in the aforementioned context.

### CLI version used

1.44.27

### Environment details (OS name and version, etc.)

Python 3.14.2, Alpine 3.23

adev-code:
Hello @rittneje, thanks for reaching out and bringing this up. We have confirmed that this is a bug and we are working on a fix. We have also found another command that gives the same error: `aws secretsmanager get-random-password`. 

We recommend using Python 3.13 for now until the fix is applied.

adev-code:
@rittneje, thanks for the patience. A fix has been merged and I have tried the command:
`aws iam update-account-password-policy --minimum-password-length <X> --require-symbols --password-reuse-prevention <X>`
looks like it seems to work now from CLI v1 version: `aws-cli/1.44.32`. Please try again and let us know if there are still issues.

rittneje:
Thanks, it is working now with the latest release.

adev-code:
@rittneje Thanks for the update. Closing.

PR #10044, opened 2026-02-02
Fix argparse error for BooleanArgument with % in documentation

adev-code:
Python 3.14 is more strict with how `argparse` handles '%' characters in help strings. While Python 3 has always treated '%' as format specifiers, Python 3.14 enforces this more strictly. When service model documentation contains '%' (e.g., IAM's `UpdateAccountPasswordPolicy` `RequireSymbols` parameter), `argparse` raises: 

```
ValueError: unsupported format character '^' (0x5e) at index 129
```

A similar fix was applied for `CLIArgument` in PR #9790. This PR adds the same `.replace('%', '%%')` escaping to `BooleanArgument.add_to_parser()`. It also adds tests to verify the CLI's argument classes properly handle '%' characters.

Fixes compatibility with Python 3.14+ for CLI v1.

Addresses:
#10038

### DIFF
--- a/tests/unit/test_argprocess.py
+++ b/tests/unit/test_argprocess.py
@@ -452,6 +452,7 @@ class TestParamShorthand(BaseArgProcessTest):
         with self.assertRaisesRegex(ParamError, error_msg):
             self.parse_shorthand(p, ['ParameterKey=key,ParameterValue="foo,bar\''])
 
+
 class TestParamShorthandCustomArguments(BaseArgProcessTest):
 
     def setUp(self):


tests/unit/test_argprocess.py
```diff
@@ -451,6 +452,26 @@ def test_csv_syntax_errors(self):
         with self.assertRaisesRegex(ParamError, error_msg):
             self.parse_shorthand(p, ['ParameterKey=key,ParameterValue="foo,bar\''])
 
+    def _test_argument_escapes_percent(self, arg_class, arg_type, doc_string, expected_substring):
+        argument = self.create_argument({'Test': {'type': arg_type}})
+        argument.argument_model.members['Test'].documentation = doc_string
+        arg = arg_class('test-arg', argument.argument_model.members['Test'], None, False)
+        parser = argparse.ArgumentParser()
```
kdaily: It looks like this is testing the standard library, instead of our argument parser abstraction. I think we should be sure to test our own code!

kdaily:
Let's also update the description in this PR to clarify that the `argparse` module in Python 3.14 is more strict with how it handles `%` characters - Python 3 has always treated them as formatting related.

kdaily:
Can you also confirm that the HTML and `man`-page docs render as expected for boolean parameters?

adev-code:
> Can you also confirm that the HTML and `man`-page docs render as expected for boolean parameters?

It renders as expected and only has one %, not %%

Screenshot 2026-02-03 at 11 36 34 AM
Screenshot 2026-02-03 at 11 37 27 AM

tests/unit/test_argprocess.py
```diff
@@ -451,7 +452,6 @@ def test_csv_syntax_errors(self):
         with self.assertRaisesRegex(ParamError, error_msg):
             self.parse_shorthand(p, ['ParameterKey=key,ParameterValue="foo,bar\''])
 
-
```
kdaily, line 454: nit: lets not remove this line, generally have two spaces between classes.

tests/unit/test_argprocess.py
```diff
@@ -894,6 +894,27 @@ def test_json_value_decode_error(self):
         with self.assertRaises(ParamError):
             unpack_cli_arg(self.p, value)
 
+class TestArgumentPercentEscaping(BaseArgProcessTest):
+    def _test_percent_escaping(self, arg_type, arg_class, doc_string):
+        argument = self.create_argument({'Test': {'type': arg_type}})
+        argument.argument_model.members['Test'].documentation = doc_string
+        arg = arg_class('test-arg', argument.argument_model.members['Test'], mock.Mock(), mock.Mock(), is_required=False)
```
kdaily: nit: this is a long line, let's format:


```suggestion
        arg = arg_class(
            'test-arg',
            argument.argument_model.members['Test'],
            mock.Mock(),
            mock.Mock(),
            is_required=False,
        )
```

tests/unit/test_argprocess.py
```diff
@@ -894,6 +894,27 @@ def test_json_value_decode_error(self):
         with self.assertRaises(ParamError):
             unpack_cli_arg(self.p, value)
 
+class TestArgumentPercentEscaping(BaseArgProcessTest):
+    def _test_percent_escaping(self, arg_type, arg_class, doc_string):
+        argument = self.create_argument({'Test': {'type': arg_type}})
+        argument.argument_model.members['Test'].documentation = doc_string
```
kdaily: Is there a reason to add the documentation directly on the instance? Can't it be added in the argument dictionary?


```suggestion
        argument = self.create_argument(
            {
                'Test': {
                    'type': arg_type,
                    'documentation': doc_string,
                }
            }
        )
```

Reference: https://github.com/boto/botocore/blob/9ac4824189f386066fdacd1c977cdff73f8a3963/botocore/model.py#L124

tests/unit/test_argprocess.py
```diff
@@ -894,6 +894,27 @@ def test_json_value_decode_error(self):
...
+        arg = arg_class('test-arg', argument.argument_model.members['Test'], mock.Mock(), mock.Mock(), is_required=False)
+        arg_table = {arg.name: arg}
+        parser = ArgTableArgParser(arg_table)
+        help_output = parser.format_help()
+        self.assertIn(arg.cli_name, help_output)
```
kdaily: I'm not sure I follow what this assertion is checking. Can you confirm that it would be checking that the percent escaping is working properly?

tests/unit/test_argprocess.py
```diff
@@ -894,6 +894,27 @@ def test_json_value_decode_error(self):
...
+
+    def test_cli_argument_escapes_percent(self):
+        self._test_percent_escaping('integer', CLIArgument, 'Symbols: % ^ & *')
```
kdaily: Nit: Would a `CLIArgument` with an `integer` type ever have documentation that describes allowed symbols like percentages? I realize the type and the documentation are distinct and don't directly interact, but it might be more clear if the type was a string.

kdaily:
Make sure to squash before merging! :shipit:

PR #10073, opened 2026-02-12
Update CloudTrail validate-logs for full key query range

IanLeeYT:
*Issue #, if available:*
N/A
*Description of changes:*
There exists an edge case in the CloudTrail validate-logs command where the script fails to query all CloudTrail public keys to validate the digest signature. It can happen when the command end time lands right after the rotation instant. Fix is to add a two hour buffer period.

### DIFF
--- a/.changes/next-release/bugfix-cloudtrail-9158.json
+++ b/.changes/next-release/bugfix-cloudtrail-9158.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "cloudtrail",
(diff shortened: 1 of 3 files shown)


AndrewAsseily (changes requested):
Thanks Ian! Looks good. Could you just add a changelog entry, something like:

'''
{
 "category": "cloudtrail",
 "description": "Fixed edge case in ``validate-logs`` where digest validation could fail with \"public key not found\" when the end time lands near a key rotation boundary",
 "type": "bugfix"
}
'''

PR #10087, opened 2026-02-18
fix(text): avoid trailing None for paginated response + scalar --query

### LINKED ISSUE #10061: Describing ChangeSet with '--output text' flag returns invalid result
### Describe the bug

When trying to describe a change set to get a specific value, such as the ChangeSetId, if using the `--output text` flag, this results in an invalid response. It will include both the ARN for the changeset, but also add 'None' in a new line. It's as if it is pulling from a second field or trying to return two values.

### Regression Issue

### Expected Behavior

The output, in this case, should just return the ARN for the changeset as a single string:

```bash
$ aws cloudformation describe-change-set --stack-name my-cfn-stack --change-set-name 'change-set-1' --query 'ChangeSetId' --output text
arn:aws:cloudformation:us-east-1:00000000000:changeSet/change-set-1/11111111111111111111111111
```

### Current Behavior

Instead of just getting the field back in the `--query` (ChangeSetId in this case), we are getting two responses back when using `--output text`.

```bash
$ aws cloudformation describe-change-set --stack-name my-cfn-stack --change-set-name 'change-set-1' --query 'ChangeSetId'
"arn:aws:cloudformation:us-east-1:00000000000:changeSet/change-set-1/11111111111111111111111111"

$ aws cloudformation describe-change-set --stack-name my-cfn-stack --change-set-name 'change-set-1' --query 'ChangeSetId' --output text
arn:aws:cloudformation:us-east-1:00000000000:changeSet/change-set-1/11111111111111111111111111
None
```

### Reproduction Steps

Create a Change Set
Use the CLI to describe changeset and query for a specific key
Repeat, but add `--output text` and observe 'None' is returned with the expected value.

### Possible Solution

_No response_

I've tested this and observed the behavior on my machine (macOS) and we have observed the same behavior in GitHub actions (AmazonLinux 2023 private hosted runner).

Also happens in version:
aws-cli/2.32.30 Python/3.13.11 Darwin/25.2.0 source/x86_64

### CLI version used

aws-cli/2.33.16 Python/3.13.12 Darwin/25.2.0 source/x86_64

MacOS 26.2

mjfxjas:
Fixes #10061.

When an operation is paginated, the text formatter previously applied `--query` per page. For scalar queries, if subsequent pages did not include the queried field, text output could include an extra trailing `None` line.

This change buffers the full paginated result before applying `--query` when using `--output text`, matching the behavior of the JSON/table formatters.

- Adds a unit test covering the regression.

### DIFF
--- a/tests/unit/test_text_formatter_pagination.py
+++ b/tests/unit/test_text_formatter_pagination.py
@@ -1,0 +1,44 @@
+import io
+
+import jmespath
+from botocore.paginate import PageIterator
+
+from awscli.formatter import TextFormatter
+
+
+class FakePageIterator(PageIterator):
+    # We intentionally don't call the parent constructor; this object is only
+    # used to satisfy `isinstance(response, PageIterator)` and provide a stable
+    # `build_full_result()` for the formatter.
+    def __init__(self, full_result):
+        self._full_result = full_result
+
+    def build_full_result(self):
+        return self._full_result
+
+
+class Args:
+    def __init__(self, query):
+        self.query = query
+
+
+def test_text_formatter_buffers_paginated_response_when_query_is_set():
+    # Regression for aws/aws-cli#10061:
+    # when a response is paginated, text output previously applied the query
+    # per-page. If later pages didn't include the queried scalar field, the
+    # formatter could emit an extra trailing "None" line.
+    args = Args(query=jmespath.compile('ChangeSetId'))
+    formatter = TextFormatter(args)
+
+    response = FakePageIterator(
+        {
+            'ResponseMetadata': {'RequestId': 'abc'},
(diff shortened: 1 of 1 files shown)


mjfxjas:
Friendly follow-up on this behavior fix.

This addresses trailing None output when using paginated responses with scalar --query, and keeps output consistent and clean for scripting use cases. If useful, I can add an additional regression test case in the preferred location.

RyanFitzSimmonsAK:
Hi @mjfxjas, thanks for this pull request.  We've received many issues in the past about the interaction between `--query` and `--output text`.  We chose to document that interaction instead of fixing it because it would be a breaking change to people reliant on the existing pagination behavior.  So unfortunately, this pull request is breaking and we will be closing it.  Thank you for contributing.

Issue #9914, opened 2025-12-11
Python 3.14 Support


aemous:
### Describe the feature

Support Python version 3.14 in AWS CLI v2, and upgrade the bundled Python interpreter to version 3.14.

### Use Case

I want source distribution to support the latest new features in Python 3.14.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

2.32.15

### Environment details (OS name and version, etc.)

MacOS 26.1

adamjstewart:
Note that flit-core 3.10+ is required for Python 3.14 support.

int3rnauta:
At the moment this is also an issue in Fedora 43, which ships with Python 3.14, and newer AWS CLI versions (Fedora is at 2.27, but I needed the new "mfa-login" from 2.30+ or "login" from 2.32).

Installing it directly gives the same error as #9821 , resolved with this workaround:
```
sudo dnf install awscli2 # Actually only its dependencies from the distro are needed
pip install -U --user flit_core
pip install --user --no-build-isolation git+https://github.com/aws/aws-cli.git@v2
```

#9822 should be able to fix that, I think.

Thank you in advance.

tkelman:
> Note that flit-core 3.10+ is required for Python 3.14 support.

Open as #9623

Also see:
- #9814 for ruamel-yaml
- #9853 for ruamel-yaml-clib

hssyoo:
Python 3.14 support was added via https://github.com/aws/aws-cli/pull/10102 in AWS CLI v2.34.13

PR #10102, opened 2026-02-23
[v2] Add support for Python 3.14

hssyoo:
Addresses https://github.com/aws/aws-cli/issues/9914

Changes:
- Port test fixes from https://github.com/boto/botocore/pull/3450 and https://github.com/boto/botocore/pull/3522
- Set EKS YAML object width to inf to avoid adding unnecessary line breaks when dumping. This maintains roundtrip and prevents existing tests from failing
- Fix local URI construction when installing downloaded packages in build script
- Update `finally` clause to comply with [PEP 765](https://peps.python.org/pep-0765/)
- Add Python 3.14 to GitHub Actions test matrix
- Update all Python 3.13 settings to 3.14
- Bump pinned `pip` build dependency to `26.0.1`
- Bump `flit_core` ceiling to `3.12.0`
- Bump `ruamel.yaml` ceiling to `0.19.1`
- Bump `ruamel.yaml.clib` ceiling to `0.2.15`
- Bump `pyinstaller` to `6.19.0`
- Pin `virtualenv` dev dependency to `20.35.4` to avoid [conditional Python version ranges](https://github.com/pypa/virtualenv/blob/eb27e55d716d8836b24296efa3a73b98ff1b29e8/pyproject.toml#L48-L49)

### DIFF
--- a/awscli/customizations/s3/s3handler.py
+++ b/awscli/customizations/s3/s3handler.py
@@ -629,9 +629,7 @@ class LocalDeleteRequestSubmitter(BaseTransferRequestSubmitter):
             self._result_queue.put(SuccessResult(**result_kwargs))
         except Exception as e:
             self._result_queue.put(FailureResult(exception=e, **result_kwargs))
-        finally:
-            # Return True to indicate that the transfer was submitted
-            return True
+        return True
 
     def _format_src_dest(self, fileinfo):
         return self._format_local_path(fileinfo.src), None


tests/functional/botocore/leak/test_resource_leaks.py
```diff
@@ -22,8 +22,8 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
     # a substantial amount of time to the total test run time.
     INJECT_DUMMY_CREDS = True
     # We're making up numbers here, but let's say arbitrarily
-    # that the memory can't increase by more than 10MB.
-    MAX_GROWTH_BYTES = 10 * 1024 * 1024
+    # that the memory can't increase by more than 13MB.
+    MAX_GROWTH_BYTES = 13 * 1024 * 1024
```
aemous, line 26: In https://github.com/boto/botocore/pull/3522/, this threshold was called out as potentially temporarily, and that the tests would be revisited after Python 3.14 released.

I'm curious if you tried lowering the limit back to 10 and noticed test failures?

tests/backends/test_pep517.py
```diff
@@ -70,7 +70,7 @@ def assert_dist_info_correctness(dist_info_dir):
     metadata_file = dist_info_dir.join("METADATA")
     assert metadata_file.check()
     metadata_content = metadata_file.read()
-    assert "Metadata-Version: 2.1" in metadata_content
+    assert "Metadata-Version: 2.4" in metadata_content
```
aemous, line 73: Can you describe this change in the PR description?

tests/functional/botocore/leak/test_resource_leaks.py
hssyoo: It's still failing:
```
FAILED tests/functional/botocore/leak/test_resource_leaks.py::TestDoesNotLeakMemory::test_create_single_paginator_memory_constant - AssertionError: False is not true : 11960320
FAILED tests/functional/botocore/leak/test_resource_leaks.py::TestDoesNotLeakMemory::test_create_single_waiter_memory_constant - AssertionError: False is not true : 11878400
```

We could set a slightly lower threshold like 12MB but I don't think it's a difference maker

tests/backends/test_pep517.py
hssyoo: Change is coming from build dependencies using a new license metadata field introduced in metadata version 2.4: https://github.com/pypa/packaging/blob/main/CHANGELOG.rst#id126:~:text=Change%20our%20license%20metadata%20to%20use%20an%20SPDX%20license%20expression%20(%3Apull%3A%60881%60%2C%20%3Apull%3A%60924%60)

PR #10105, opened 2026-02-25
[v2] Implement opting out of S3 Express session auth via config variables.

aemous:
*Notes:*

* Ports https://github.com/boto/botocore/pull/3631 to the vendored botocore.

*Description of changes:*

* Added support for opting out of S3 Express session authentication via the new ``AWS_S3_DISABLE_EXPRESS_SESSION_AUTH`` environment variable or the ``s3_disable_express_session_auth`` S3 shared configuration setting.
* Added new tests against the new configuration and functionality.

*Description of tests:*

* Successfully ran all unit and functional tests.
* Successfully ran build-and-validate workflow.


ashovlin:
A profile like this is failing for me
```
[default]
cli_history=enabled
retry_mode=standard
output=json
region=us-east-1
s3 = 
    disable_s3_express_session_auth = true
```

with `Value (DisableS3ExpressSessionAuth) is the wrong type. Must be <class 'bool'>.`.

Tried `true`, `True`, `abc123`.

awscli/botocore/configprovider.py
```diff
@@ -192,6 +192,12 @@
         None,
         None,
     ),
+    'disable_s3_express_session_auth': (
+        ('s3', 's3_disable_express_session_auth'),
```
ashovlin: Is this nesting `s3_disable_express_session_auth` under `s3` like:
```
s3 = 
    s3_disable_express_session_auth = true
```


That diverges from https://docs.aws.amazon.com/sdkref/latest/guide/feature-s3-express.html and the internal specification which calls for a top-level profile setting. Do we still nest it like this for new settings?

kdaily:
> A profile like this is failing for me
> 
> ```
> [default]
> cli_history=enabled
> retry_mode=standard
> output=json
> region=us-east-1
> s3 = 
>     disable_s3_express_session_auth = true
> ```
> 
> with `Value (DisableS3ExpressSessionAuth) is the wrong type. Must be <class 'bool'>.`.
> 
> Tried `true`, `True`, `abc123`.

Per the definition, shouldn't it be `s3_disable_express_session_auth` instead of `disable_s3_express_session_auth`?

https://github.com/aws/aws-cli/pull/10105/changes#diff-eb369da3f989b1a2325e07d20f4032ae04fd0abf1739793504240b214f325a3cR196

```
[default]
cli_history=enabled
retry_mode=standard
output=json
region=us-east-1
s3 = 
    s3_disable_express_session_auth = true
```

awscli/botocore/args.py
```diff
@@ -676,6 +676,9 @@ def compute_endpoint_resolver_builtin_defaults(
             EPRBuiltins.AWS_S3_DISABLE_MRAP: s3_config.get(
                 's3_disable_multiregion_access_points', False
             ),
+            EPRBuiltins.AWS_S3_DISABLE_EXPRESS_SESSION_AUTH: s3_config.get(
+                'disable_s3_express_session_auth', False
+            ),
```
ashovlin, line 728: Is the environment variable plumbed into here? I'm not seeing it have any effect when testing, I still see the `CreateSession` call with

```
AWS_S3_DISABLE_EXPRESS_SESSION_AUTH=true aws s3api list-objects-v2 --bucket <express-bucket> --debug 2>&1 | grep "Making request"
```

awscli/botocore/config.py
```diff
@@ -120,6 +120,10 @@ class Config:
           * path -- Addressing style is always by path. Endpoints will be
             addressed as such: s3.amazonaws.com/mybucket
 
+        * ``disable_s3_express_session_auth`` -- Refers to whether to use S3
```
ashovlin, line 123: Please add docs to either topics/s3-config or s3-faq

awscli/botocore/args.py
ashovlin: In configprovider we expanded `BOTOCORE_DEFAUT_SESSION_VARIABLES`, should it be `DEFAULT_S3_CONFIG_VARS` instead?

ashovlin:
> > A profile like this is failing for me
> > ```
> > [default]
> > cli_history=enabled
> > retry_mode=standard
> > output=json
> > region=us-east-1
> > s3 = 
> >     disable_s3_express_session_auth = true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with `Value (DisableS3ExpressSessionAuth) is the wrong type. Must be <class 'bool'>.`.
> > Tried `true`, `True`, `abc123`.
> 
> Per the definition, shouldn't it be `s3_disable_express_session_auth` instead of `disable_s3_express_session_auth`?
> 
> https://github.com/aws/aws-cli/pull/10105/changes#diff-eb369da3f989b1a2325e07d20f4032ae04fd0abf1739793504240b214f325a3cR196
> 
> ```
> [default]
> cli_history=enabled
> retry_mode=standard
> output=json
> region=us-east-1
> s3 = 
>     s3_disable_express_session_auth = true
> ```

Oops, mixed up the client config with the profile config, thanks! 

But then why are we crashing for a wrong/non-existent key?

ashovlin (changes requested):

kdaily:
> But then why are we crashing for a wrong/non-existent key?

Huh, yeah something else is amiss.

awscli/botocore/config.py
aemous: I've pushed a revision to the topic `s3-config`.

awscli/botocore/args.py
aemous: It should be BOTOCORE_DEFAUT_SESSION_VARIABLES, because the variable is not nested under `s3` in shared config.

ashovlin (approved):

PR #10109, opened 2026-02-26
[v2] Add ability to set properties in subsections

kdaily:
This change adds new parameters to the `aws configure set` command to specify a sub-section for setting a property. These parameters are analogous to the existing `--profile` parameter. A parameter will be added for each sub-section type and take a value of the subsection name. Following is the generic pattern for the `aws configure set` command:

```
aws configure set
  --<sub-section-type> <sub-section-name> \
  <property> <value>
```

For example, the following command should set the property `sso_region` to the value `us-west-2` in the `sso-session` sub-section named `my-sso-session`:

```
aws configure set
  --sso-session my-sso-session \
  sso_region us-west-2
```

Following is an example setting a nested property in a sub-section:

```
aws configure set \
  --<sub-section-type> <sub-section-name> \
  <nested-section>.<property> value
```

The only sub-section types allowed are `services` and `sso-session`.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/examples/configure/set/_examples.rst
+++ b/awscli/examples/configure/set/_examples.rst
@@ -59,3 +59,15 @@ will produce the following updated config file::
     [profile testing2]
     region = us-west-2
     cli_pager =
+
+To set a parameter in a sub-section, use one of the available sub-section parameters (``--services`` or ``--sso-session``).
+
+For example, to set the ``sso_start_url`` in the ``my-sso-sesssion`` SSO session sub-section, the following command can be used::
+
+    aws configure set --sso-session my-sso-session sso_start_url https://my-sso-portal.awsapps.com/start
+
+To set a nested property, use dotted notation for the parameter name along with a sub-section parameter.
+For example, to set the a service specific endpoint URL for EC2 in a services sub-section called ``my-services``, the following command can be used::
+
+    aws configure set --services my-services ec2.endpoint_url http://localhost:4567
+


awscli/customizations/configure/__init__.py
```diff
@@ -53,3 +63,27 @@ def get_section_header(section_type, section_name):
     if any(c in _WHITESPACE for c in section_name):
         section_name = shlex.quote(section_name)
     return f'{section_type} {section_name}'
+
+
+def get_subsection_from_args(args):
+    # Validate mutual exclusivity of sub-section type parameters
+    groups = [[param] for param in SUBSECTION_TYPE_ALLOWLIST.values()]
+    validate_mutually_exclusive(args, *groups)
```
kdaily: Usually, mutually exclusive parameter validation is done through an event handler. Example:

https://github.com/aws/aws-cli/blob/9fcb197f0aa26b3264f9dd52f10db5e3d5a7e29a/awscli/customizations/cloudfront.py#L35-L38

This uses the `operation-args-parsed` event in `clidriver.ServiceOperation`:

https://github.com/aws/aws-cli/blob/9fcb197f0aa26b3264f9dd52f10db5e3d5a7e29a/awscli/clidriver.py#L849

However, the classes here do not have an event registered anywhere near where parameter arg-parsing is performed. We could register a new event, if its thought that this is too deep in the stack for performing this check. This is where it would need to go:

https://github.com/aws/aws-cli/blob/9fcb197f0aa26b3264f9dd52f10db5e3d5a7e29a/awscli/customizations/commands.py#L158

awscli/customizations/configure/__init__.py
```diff
@@ -53,3 +63,27 @@ def get_section_header(section_type, section_name):
     if any(c in _WHITESPACE for c in section_name):
         section_name = shlex.quote(section_name)
     return f'{section_type} {section_name}'
+
+
+def get_subsection_from_args(args):
+    # Validate mutual exclusivity of sub-section type parameters
+    groups = [[param] for param in SUBSECTION_TYPE_ALLOWLIST.values()]
```
kdaily: Initially, I had proposed that mutual exclusivity also account for `--profile`. Unfortunately, profile resolution happens before, and we've lost any context of where the user set the profile.

awscli/customizations/configure/set.py
```diff
@@ -41,6 +49,20 @@ class ConfigureSetCommand(BasicCommand):
             'cli_type_name': 'string',
             'positional_arg': True,
         },
+        {
+            'name': 'sso-session',
+            'help_text': 'The name of the SSO session sub-section to configure.',
```
ashovlin: This reads odd via `aws configure set help`. We might need to make this `help_text` more generic if there isn't a way to separate the two new options better?

```
OPTIONS
       varname (string) The name of the config value to set.

       value (string) The value to set.

       --sso-session | --services (string) The name of the SSO session
       sub-section to configure.
```

awscli/customizations/configure/set.py
```diff
@@ -60,10 +82,60 @@ def _get_config_file(self, path):
...
+        validate_mutually_exclusive(args, *groups)
+
+        subsection_name = None
+        subsection_type = None
+
+        for section_type, section_properties in SUBSECTION_TYPE_ALLOWLIST.items():
+            if hasattr(args, section_properties['param_name']):
+                param_value = getattr(args, section_properties['param_name'])
+                if param_value is not None:
+                    if not re.match(r"[\w\d_\-/.%@:\+]+", param_value):
```
ashovlin: 1. This is different than the regex in the internal spec `[A-Za-z0-9_\-/.%@:\+]+` - is accepting unicode via `\w` okay?
2. It doesn't look like `aws configure sso` validates this, just `RequiredInputValidator`. Is there a chance someone has already configured something invalid? Though since this is brand new workflow, I think it's okay to start by matching the spec, and relaxing if needed.
3. Can you add a comment explaining what we're validating here? By the time I got to the regex, I wasn't sure what `param_value` was anymore.

awscli/customizations/configure/set.py
```diff
@@ -60,10 +82,60 @@ def _get_config_file(self, path):
...
+                    break
+
+        return (subsection_type, subsection_name)
+
+
+    def _set_subsection_property(self, section_type, section_name, varname, value):
+        if '.' in varname:
+            parts = varname.split('.')
+            if len(parts) > 2:
+                return 0
```
ashovlin: Could we add a comment explaining this validation?

And if it's what I assume, that we don't have any valid properties deeper than `--services <name> s3.endpoint_url`, should we raise a validation error rather than silently not writing?

awscli/customizations/configure/set.py
kdaily: Ah, I see the issue. The intention was to highlight the mutual exclusivity of the parameter. I think updating the wording is probably the best resolution, but I could split them as well and then rely on the implementation erorring if more than one is provided:

```
OPTIONS  
       varname (string) The name of the config value to set.  

       value (string) The value to set.  

       --services (string) The name of the services   
       sub-section to configure.  

       --sso-session (string) The name of the SSO session  
       sub-section to configure.  
```

awscli/customizations/configure/set.py
kdaily: > This is different than the regex in the internal spec [A-Za-z0-9_\-/.%@:\+]+ - is accepting unicode via \w okay?

We currently accept profile and SSO session names with unicode.

> Can you add a comment explaining what we're validating here? By the time I got to the regex, I wasn't sure what param_value was anymore.

Yep, will update.

awscli/customizations/configure/set.py
kdaily: Agree we could raise validation error. The current implementation silently sets the parameter at the first level. For example:

```
aws --profile brandbrandnew configure set foo.bar.baz "qux"
```

Results in:

```
[profile brandbrandnew]
foo = qux
```

awscli/customizations/configure/set.py
kdaily: Updated in 827edb3395a7ff590f81728bd9ea1a2e328f2f4c

awscli/customizations/configure/set.py
kdaily: Raising error now, fixed in c1a63ba68f268b8f5fd90753cc36ee7727fa237e

awscli/customizations/configure/set.py
kdaily: Fixed in 1271886653003f7b48cc87d99a8642cb7d1d0b22. I kept them in the same parameter group and updated the wording.

kdaily:
> Can we also add a new example or two?

Added in be92fb10650637eb5e81aea62eb7a5e1c961e4ae

PR #10112, opened 2026-02-27
[v2] Add ability to get values from properties in subsections

kdaily:
*Issue #, if available:*

Continuation of #10109.

*Description of changes:*

This change adds new parameters to the `aws configure get` command to specify a sub-section for getting the value of a property. These parameters are analogous to the existing `--profile` parameter. A parameter will be added for each sub-section type and take a value of the subsection name. Following is the generic pattern for the `aws configure get` command:

```
aws configure get
  --<sub-section-type> <sub-section-name> \
  <property>
```

For example, the following command should get the property `sso_region` from the `sso-session` sub-section named `my-sso-session`:

```
aws configure get --sso-session my-sso-session sso_region 
```

Following is an example getting a nested property in a sub-section:

```
aws configure get \
  --<sub-section-type> <sub-section-name> \
  <nested-section>.<property>
```

The only sub-section types allowed are `services` and `sso-session`.

### DIFF
--- a/awscli/customizations/configure/__init__.py
+++ b/awscli/customizations/configure/__init__.py
@@ -10,10 +10,9 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+
 from awscli.compat import shlex
 
-# A map between the command line parameter name and the Python object name
-# For allowed sub-section types
 NOT_SET = '<not set>'
 PREDEFINED_SECTION_NAMES = 'plugins'
 # A map between the command line parameter name and the name used
@@ -47,6 +46,7 @@ class SectionNotFoundError(Exception):
 
 class SubsectionNotFoundError(Exception):
     pass
+
 
 def mask_value(current_value):
     if current_value is None:
--- a/awscli/customizations/configure/get.py
+++ b/awscli/customizations/configure/get.py
@@ -11,14 +11,16 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
-import re
 import sys
 
 from awscli.customizations.commands import BasicCommand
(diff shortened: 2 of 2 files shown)


awscli/customizations/configure/__init__.py
```diff
@@ -10,12 +10,25 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import re
```
ashovlin: Don't see usage of this in the diff?

awscli/customizations/configure/get.py
```diff
@@ -121,3 +160,24 @@ def _get_dotted_config_value(self, varname):
...
+        if '.' in varname:
+            parts = varname.split('.')
+            if len(parts) == 2:
+                value = section_config.get(parts[0])
+                if isinstance(value, dict):
+                    return value.get(parts[1])
+                else:
+                    return None
+        else:
+            return section_config.get(varname)
```
ashovlin, line 207:
```
aws configure get --sso-session non-existent sso_region        

'NoneType' object has no attribute 'get'
```

Can we throw a nicer error when the sub-section doesn't exist? And expand the test for this case to check more than exit code?

Something like we do for profiles?
```
aws configure get region --profile non-existent

The config profile (non-existent) could not be found
```

awscli/customizations/configure/__init__.py
kdaily: Fixed in 945b158c2ec338f0e6ea3f335401d4a8490220a1

awscli/customizations/configure/get.py
kdaily: Fixed in 139528d0a9f207eeb7f8a4d17c2247d2847df1c8

kdaily:
> Could we also add an example or two for `get`?

Addressed in 401de0f2df59b66d2ec20b0c92719faa1c029382

Issue #2458, opened 2017-02-22
[Feature Request] Add tagging support to s3


mikehayesuk:
It would be great if we could tag S3 objects while using s3 commands such as "cp" and "sync".

Thanks!

kyleknap:
Marking as a feature request. So something like adding a ``--tagging`` parameter to ``cp`` and ``sync``?

mhayes14:
> Marking as a feature request.

Thanks!

> So something like adding a `--tagging` parameter to `cp` and `sync`?

Yup, it would be great to take advantage of object tagging while using the CLI.

kyleknap:
Thanks for clarifying. Makes sense to me.

dinvlad:
+1, our use case is to enforce CLI users to tag their multipart uploads (through `s3:RequestObjectTag/<tag-key>` policy condition).

yvele:
Did someone found a quick and simple workaround?

dinvlad:
One workaround is to tag object on the server side after the user uploaded their files. The main limitation to that is eventual consistency, i.e. the uploaded files may not appear in the listing right away, so you'd need to introduce an arbitrary delay before labeling (I also check the combined hash of all keys and ETags).

yvele:
@dinvlad Yes I was thinking using an AWS Lambda that triggers when an S3 object is put.. But that's a bit overkill for such a basic feature.. 😄

dinvlad:
@yvele agreed, it would be nice to have it out of the box. However, it does put more burden on the user because now they also need to remember how to apply tags appropriately.

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

eikenb:
You can tag from the cli using `aws s3api put-object-tagging ...`. I'm currently using it in combo with `aws s3 mv ...` to migrate some data w/ tags. It works, but I'd rather have it in one call so that if I need to interrupt the script I wouldn't have to worry about something missing the tags.

bpo:
PR #2890 with boto/s3transfer#96 adds the support for tagging without a second call.

AbhiPrasad:
Is this still on the table? Would be super helpful to our current workflows!

trav-c:
I'd like to see this added too, it would simplify parts of our deployment workflow quite a bit.

NoahSDS:
I want this, too!

mwhouser:
+1 for an all-in-one "copy and tag".

[...]

PR #10114, opened 2026-03-02
[v2] Add tagging support for the s3 mb command

aemous:
v2 version of https://github.com/aws/aws-cli/pull/10119

*Issues:*

* Contributes to https://github.com/aws/aws-cli/issues/2458

*Description of changes:*

* Added support for specifying tags on bucket creation for the `aws s3 mb` command.
* Added functional tests for testing the new feature.
* Added a new `aws s3 mb` command example using the new flag.

*Description of test:*

* Ran a successful build-and-validate workflow. 

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -1240,16 +1240,17 @@ class MbCommand(S3Command):
                 "Cannot use mb command with a directory bucket."
             )
 
-        bucket_config = {
-            'LocationConstraint': self.client.meta.region_name
-        }
-
         params = {'Bucket': bucket}
+        bucket_config = {}
         bucket_tags = self._create_bucket_tags(parsed_args)
 
+        # Only set LocationConstraint when  the region name is not us-east-1.
+        # Sending LocationConstraint with value us-east-1 results in an error.
+        if self.client.meta.region_name != 'us-east-1':
+            bucket_config['LocationConstraint'] = self.client.meta.region_name
         if bucket_tags:
             bucket_config['Tags'] = bucket_tags
-        if self.client.meta.region_name != 'us-east-1':
+        if bucket_config:
             params['CreateBucketConfiguration'] = bucket_config
 
         # TODO: Consolidate how we handle return codes and errors
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -673,6 +673,15 @@ NO_OVERWRITE = {
     ),
 }
 
+TAGS = {
+    'name': 'tags',
+    'action': 'append',
+    'nargs': 2,
+    'help_text': (
+        "This flag specifies tags to be added to the bucket."
+    ),
+}
+
 
 CASE_CONFLICT = {
     'name': 'case-conflict',
@@ -1206,7 +1215,16 @@ class MbCommand(S3Command):
     NAME = 'mb'
     DESCRIPTION = "Creates an S3 bucket."
     USAGE = "<S3Uri>"
-    ARG_TABLE = [{'name': 'path', 'positional_arg': True, 'synopsis': USAGE}]
+    ARG_TABLE = (
+        [
+            {
+                'name': 'path',
+                'positional_arg': True,
+                'synopsis': USAGE,
+            }
+        ]
+        + [TAGS]
+    )
 
     def _run_main(self, parsed_args, parsed_globals):
         super(MbCommand, self)._run_main(parsed_args, parsed_globals)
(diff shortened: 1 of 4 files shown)


awscli/customizations/s3/subcommands.py
```diff
@@ -1222,8 +1240,15 @@ def _run_main(self, parsed_args, parsed_globals):
...
+        bucket_config = {
+            'LocationConstraint': self.client.meta.region_name
+        }
+
         params = {'Bucket': bucket}
+        bucket_tags = self._create_bucket_tags(parsed_args)
+
+        if bucket_tags:
+            bucket_config['Tags'] = bucket_tags
         if self.client.meta.region_name != 'us-east-1':
```
AndrewAsseily: I'm not sure I understand how this works in `us-east-1`, it looks like `bucket_config` only gets passed as `CreateBucketConfiguration` when the region isn't `us-east-1`. Wouldn't that mean tags get silently dropped for buckets created in that region? Maybe I'm missing something about how the new S3 API handles this.

awscli/customizations/s3/subcommands.py
```diff
@@ -673,6 +673,15 @@
     ),
 }
 
+TAGS = {
+    'name': 'tags',
+    'action': 'append',
+    'nargs': 2,
+    'help_text': (
+        "This flag specifies tags to be added to the bucket."
```
AndrewAsseily: Could the help text include an example of the expected format? Since --tags Key Value with two positional args is a bit unusual, I could see users trying `--tags Key=Value` first. Something like "Usage: --tags Key1 Value1 --tags Key2 Value2" might save people a trip to the docs.

AndrewAsseily (changes requested):

awscli/customizations/s3/subcommands.py
aemous: Good callout.

I made two updates in the latest revision:

1. Added a synopsis to the flag definition. This makes the usage more clear in the API reference and the manpage: 

```
NAME
       mb -

DESCRIPTION
       Creates an S3 bucket.

SYNOPSIS

            mb
          <S3Uri>
          [--tags <key> <value>]
          [--debug]
```

2. I've updated the description to include the usage format.

awscli/customizations/s3/subcommands.py
aemous: Great catch!

For context on why the `us-east-1` check exists, S3 CreateBucket API will return an error if you send `LocationConstraint` in the region `us-east-1`. LocationConstraint is required for all other regions.

I've pushed a revision so that `CreateBucketConfiguration` is still sent with tags but without `LocationConstraint` in the region `us-east-1`.

AndrewAsseily: Thanks for the explanation and the fix

awscli/customizations/s3/subcommands.py
```diff
@@ -1222,9 +1243,17 @@ def _run_main(self, parsed_args, parsed_globals):
                 "Cannot use mb command with a directory bucket."
             )
 
-        bucket_config = {'LocationConstraint': self.client.meta.region_name}
         params = {'Bucket': bucket}
+        bucket_config = {}
+        bucket_tags = self._create_bucket_tags(parsed_args)
+
+        # Only set LocationConstraint when  the region name is not us-east-1.
```
AndrewAsseily: Tiny nit: double space in "when  the region name" on this comment.

AndrewAsseily: Looks good otherwise! Approving 🏆

AndrewAsseily (dismissed):

awscli/customizations/s3/subcommands.py
aemous: Good catch, revised.

AndrewAsseily (dismissed):
:shipit:

kdaily (changes requested):
Question: If the flag can only hold one tag, and you can specify it multiple times, then why is the name the plural `tags` instead of `tag`?

aemous:
> Question: If the flag can only hold one tag, and you can specify it multiple times, then why is the name the plural tags instead of tag?

Do you have a strong preference for the singular form? I'm open to feedback on this.

 My reason for choosing plural is because the flag forms a list of tags (i.e. argparse represents the flag as a list). It may be a bit weird for it to be named `tag` but in code it is a list of multiple tags.

kdaily:
Fair enough, not that strongly. However, the proposed syntax does deviate from other existing commands besides `s3` that have tagging support that use `--tags Key1=Value1 Key2=Value2` syntax. Just confirming we considered that in the design here.

https://github.com/aws/aws-cli/pull/9981

AndrewAsseily:
> Fair enough, not that strongly. However, the proposed syntax does deviate from other existing commands besides `s3` that have tagging support that use `--tags Key1=Value1 Key2=Value2` syntax. Just confirming we considered that in the design here.
> 
> #9981

When thinking about this earlier, I found that --tags (plural) is overwhelmingly the convention across the CLI, though I don't feel too strongly about either in this case.
On the Key=Value syntax, the design doc covers this. The shorthand syntax parser has character limitations (e.g., @ is reserved for key@=file:// loading), which would break for valid S3 tag keys. The Key Value format with nargs=2 avoids that issue

kdaily (dismissed):

ashovlin:
Can you add a changelog entry and a new example of using it?

aemous:
> Can you add a changelog entry and a new example of using it?

I've added both in latest revision.

AndrewAsseily (approved):

PR #10118, opened 2026-03-10
Add model validation test for completions-1.json operation references

AndrewAsseily:
*Issue #, if available:*

*Description of changes:* This supports our internal build system's model validation and adds a new test_completions_operations_exist_in_model test marked with `@pytest.mark.validates_models` that validates every operation referenced in a service's completions-1.json exists in the corresponding service-2.json model.

### DIFF
--- a/tests/functional/autocomplete/test_completion_files.py
+++ b/tests/functional/autocomplete/test_completion_files.py
@@ -107,6 +107,35 @@ def get_models_with_completions():
     return models
 
 
+def _get_invalid_completion_operations():
+    cases = []
+    for test_data in get_models_with_completions():
+        known_ops = set(test_data.service_model['operations'])
+        for op_name in test_data.completions['operations']:
+            if op_name not in known_ops:
+                cases.append((test_data.service_name, op_name))
+    return cases
+
+
+@pytest.mark.validates_models
+@pytest.mark.parametrize(
+    "service_name, operation_name",
+    _get_invalid_completion_operations(),
+    ids=lambda val: str(val),
+)
+def test_completions_operations_exist_in_model(
+    service_name, operation_name, record_property
+):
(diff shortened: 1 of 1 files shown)


tests/functional/autocomplete/test_completion_files.py
```diff
@@ -107,6 +107,35 @@ def get_models_with_completions():
     return models
 
 
+def _get_invalid_completion_operations():
+    cases = []
+    for test_data in get_models_with_completions():
```
ashovlin, line 112: `test_verify_generated_completions_are_valid` is already effectively looping over `get_models_with_completions` once. Can we emit this new message from there instead of introducing a second loop?

AndrewAsseily: The existing test is parametrized per service, but the `validates_models` pattern requires that `record_property` is calld at most once per property per test case.
If a service has multiple missing operations (e.g. s3tables had 6 in the failing build), we can't call record_property('aws_operation', ...) multiple times in one test case. So we need a separate parametrization at the (service, operation) level, which requires a different test.

That said, we can avoid loading the models twice by caching the result of get_models_with_completions() and having both tests reference that. Want me to do that?

aemous: I agree with Andrew's reply. If we only looped over the services like in `test_verify_generated_completions_are_valid`, and the service has multiple missing operations, the best we could do is return a single failure to the service internally that says "6 operations are missing...".

We've received internal guidance that each failing approval we return to services are scoped at the property/operation level, when possible. In other words, our internal guidance is to prefer 6 different failing approvals/test-cases that say "Operation XYZ is missing" rather than 1 approval/test-case that says "6 operations are missing..."

Andrew's method for doing this is aligned with the former, which aligns with internal guidance.

PR #10126, opened 2026-03-12
[v2] Add generic support for client context params to vendored botocore

aemous:
*Description of changes:*

* Ports https://github.com/boto/botocore/pull/3037 to vendored botocore. This PR is expected to have **no customer-facing** impact.
* "Adds `client_context_params` config option. If the service is s3 or s3control, still use the s3 config param over client context params if it is set. Params are modeled as capital case but are expected in snake case by the endpoint resolver. This commit also includes some code cleanup for the existing client context params test."

*Description of tests:*

* Successfully ran the build-and-validate workflow.

### DIFF
--- a/tests/functional/botocore/test_context_params.py
+++ b/tests/functional/botocore/test_context_params.py
@@ -59,6 +59,11 @@ FAKE_RULESET_WITH_CLIENT_CONTEXT_PARAM = {
             "documentation": "",
             "type": "String",
         },
+        "BarClientContextParamName": {
+            "required": False,
+            "documentation": "",
+            "type": "String",
+        },
     },
 }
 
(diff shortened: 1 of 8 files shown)


tests/functional/botocore/test_context_params.py
```diff
@@ -224,86 +233,178 @@
...
         (
             's3control',
             FAKE_S3CONTROL_MODEL_WITHOUT_ANY_CONTEXT_PARAMS,
             FAKE_RULESET_WITHOUT_ANY_CONTEXT_PARAMS,
-            False,
+            CONFIG_WITH_S3,
+            NO_CTX_PARAM_EXPECTED_CALL_KWARGS,
+        ),
+        (
+                's3control',
```
AndrewAsseily: nit: Any reason these three s3control entries are more indented (16 spaces) than the rest of them (12 spaces)? Looks like the original botocore PR uses 12 consistently.

aemous: Good catch! I've pushed a revision that makes the indentation consistent.

AndrewAsseily (approved):
Looks good, thanks!
Approved assuming CI passes 🏆

Issue #9148, opened 2024-12-18
aws sso login: allow to customize ip address for redirect_uri


miki725:
### Describe the feature

Recently `aws sso login` started using Authorization Code flow vs previously default device auth flow. As part of that flow CLI shows a URL which embeds a `redirect_uri` which redirects back to `127.0.0.1` after flow is successfully completed on the AWS side. For example:

```sh
➜ aws sso login
Attempting to automatically open the SSO authorization page in your default browser.
If the browser does not open or you wish to use a different device to authorize this request, open the following URL:

https://oidc.us-east-1.amazonaws.com/authorize?response_type=code&client_id=SECRET&redirect_uri=http%3A%2F%2F127.0.0.1%3A34535%2Foauth%2Fcallback&state=TRUNCATED
```

This works great if you are running the command locally on the machine. However sometimes its useful to run it on a remote machine where `127.0.0.1` does not work for the `redirect_uri`. 

### Use Case

Running `aws sso login` on a remote machine where `127.0.0.1` is not accessible for the `redirect_uri`

### Proposed Solution

The request is to allow to customize the host of the `redirect_uri` which will allow to use the Authorization Code flow from a remote connection:

```sh
➜ aws sso login --redirect-host=<remote-hostname>
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

2.22.15

### Environment details (OS name and version, etc.)

aws-cli/2.22.15 Python/3.12.6 Linux/6.10.8-arch1-1 exe/x86_64.arch

ashovlin:
Double-checking, is the desired alternative hostname still the machine that you're originating the login from? If not, we'd need mechanism to get the authorization code that is returned by the redirect back to the machine that is logging in.

Currently AWS IAM Identity Center requires this to be `127.0.0.1`, since we're relying on the implicit trust users have in software running on their own machine. I'll raise this with them internally, but you may need to rely on `--use-device-code` as you described for now.

miki725:
In my specific use-case, I ssh into my home server where I do all dev work on. I implicitly trust that machine. Just the redirect back to `127.0.0.1` doesnt work for me as I open it the login URL in my local machine browser hence the redirect back to `127.0.0.1` doesnt work.

Ideal flow:

* ssh into server
  ```
  @local ➜ ssh server
  ```
* execute `aws sso login` on a remote server
  ```
  @server ➜ aws sso login --redirect-host=server
[...]

PR #10134, opened 2026-03-15
Allow SSO OAuth callback port override

ashenm:
Fixes https://github.com/aws/aws-cli/issues/9148

This change adds support allowing OAuth callback address port and host bind override used by parameters `--redirect-port` and `--redirect-host` respectively.
 
The main use case is containerized and remote development environments, where the default localhost callback port may not be suitable or easily reachable. By allowing the port to be configured externally, the login flow becomes more flexible across Docker, dev containers, remote workspaces, and other non-standard local setups.

### DIFF
--- a/awscli/customizations/sso/login.py
+++ b/awscli/customizations/sso/login.py
@@ -57,6 +57,10 @@ class LoginCommand(BaseSSOCommand):
             session_name=sso_config.get('session_name'),
             registration_scopes=sso_config.get('registration_scopes'),
             use_device_code=parsed_args.use_device_code,
+            redirect_address=(
+                parsed_args.redirect_host,
+                parsed_args.redirect_port,
+            ),
         )
         success_msg = 'Successfully logged into Start URL: %s\n'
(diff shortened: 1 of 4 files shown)


ashovlin:
Hey @ashenm, I'd prefer we start with an option like `--redirect-port` in #9148.

We try to ensure that `AWS_` environment variables work across SDKs and tools, so if going with `AWS_OAUTH_CALLBACK_SERVER_PORT` we'd want to coordinate with at least the AWS Tools for PowerShell team, and update more documentation. Starting with just an AWS CLI option is simpler, and part of the generated documentation. 

Will this approach work for your use case?

ashenm:
@ashovlin got it! Updated adding `--redirect-port` and `--redirect-host`

ashenm:
@ashovlin let me know if anything missing and needs update

marksteward:
Does this work? https://github.com/aws/aws-cli/issues/9148#issuecomment-2554529973 suggests it's hardcoded on the server side.

PR #10139, opened 2026-03-17
Improve error message when test_models_contain_only_known_file_types fails

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* I had a `.DS_Store` locally that was causing this test to fail. Because it splits on the `.`, the error that `''` was not in `known_types` was confusing. This adds a more verbose error message to the assertion. 


AndrewAsseily:
Thanks!

PR #10142, opened 2026-03-17
Filter unmodeled fields from error output

AndrewAsseily:
*Description of changes:* The CLI's "Additional error details" section previously displayed all fields from the error response, including fields not defined in the service model. This change filters the output to only show fields that are explicitly defined in the service's error shape model.

#### Non-modeled: https://github.com/aws/aws-cli/blob/v2/awscli/botocore/data/s3/2006-03-01/service-2.json#L8516

Before: 

```
(.venv)  aws-cli % aws s3api get-object --bucket not-a-real-bucket-0000 --key file.txt out.txt

aws: [ERROR]: An error occurred (NoSuchBucket) when calling the GetObject operation: The specified bucket does not exist

Additional error details:
BucketName: not-a-real-bucket-0000
```

After: 

```
(.venv)  aws-cli % aws s3api get-object --bucket not-a-real-bucket-0000 --key file.txt out.txt                          

aws: [ERROR]: An error occurred (NoSuchBucket) when calling the GetObject operation: The specified bucket does not exist
```

#### Modeled: https://github.com/aws/aws-cli/blob/v2/awscli/botocore/data/lambda/2015-03-31/service-2.json#L7930

```
(.venv) aws-cli % aws lambda get-function --function-name nonexistent-function-12345

aws: [ERROR]: An error occurred (ResourceNotFoundException) when calling the GetFunction operation: Function not found: arn:aws:lambda:us-west-2:<redacted-for-this-PR>:function:nonexistent-function-12345

Additional error details:
Type: User
```

### DIFF
--- a/awscli/errorhandler.py
+++ b/awscli/errorhandler.py
@@ -303,7 +303,7 @@ class ClientErrorHandler(FilteredExceptionHandler):
                     if key.lower() in modeled_lower:
                         error_dict[key] = value
 
-            # Filter fields inside Error dict as well.
+            # Unmodeled fields can contain data that shouldn't be displayed.
             for key in list(error_dict.keys()):
                 if key.lower() not in modeled_lower:
                     del error_dict[key]


awscli/botocore/client.py
```diff
@@ -903,8 +903,17 @@ def _make_api_call(self, operation_name, api_params):
...
+            error_shape = self._service_model.shape_for_error_code(
+                error_code
+            )
+            modeled_fields = {'Code', 'Message'}
+            if error_shape:
+                modeled_fields |= set(error_shape.members.keys())
             error_class = self.exceptions.from_code(error_code)
-            raise error_class(parsed_response, operation_name)
+            error = error_class(parsed_response, operation_name)
+            error.modeled_fields = modeled_fields
```
ashovlin, line 915: Can we do this through either the constructor or a property that defaults to none so we don't need to be as defensive with the `getattr(exception, 'modeled_fields', None)` later?

awscli/errorhandler.py
```diff
@@ -275,7 +293,10 @@ def _get_formatted_message(self, error_info, exception):
     def _extract_error_info(self, exception):
         error_response = self._extract_error_response(exception)
         if error_response and 'Error' in error_response:
-            return error_response['Error']
+            error_info = error_response['Error']
+            modeled_fields = getattr(exception, 'modeled_fields', None)
+            error_info['_modeled_fields'] = modeled_fields
```
ashovlin: We're extracting `.modeled_fields` from the exception, copying that into `error_info['_modeled_fields`], then later using that to delete keys from `error_info`?

That was tough to follow. Can `_extract_error_response`, when it's building `error_dict`, ignore the unmodeled keys initially?

I think that'll be simpler and easier to follow. `_extract_error_response` remains responsible for extracting the info from the exception we care about, rather than that "leaking" that responsibility into a "display" method.

awscli/botocore/client.py
AndrewAsseily: Done. Added `self.modeled_fields = None` as a default so no getattr needed.

awscli/errorhandler.py
AndrewAsseily: Done. `_extract_error_response` now filters by modeled_fields when building error_dict right away

awscli/errorhandler.py
```diff
@@ -286,15 +290,23 @@ def _extract_error_response(exception):
...
+
+            # Only include fields present in the error shape model.
             excluded_keys = {'Error', 'ResponseMetadata', 'Code', 'Message'}
             for key, value in exception.response.items():
                 if key not in excluded_keys and key not in error_dict:
-                    error_dict[key] = value
+                    if key.lower() in modeled_lower:
+                        error_dict[key] = value
+
+            # Filter fields inside Error dict as well.
```
ashovlin: Why? In general, it's better to comment why we're doing something not what we're doing.

AndrewAsseily: Good point thank you!

PR #10144, opened 2026-03-17
Add tip that suggests 'aws login' during 'aws configure' for new profiles

ashovlin:
*Issue #, if available:* CLI-8031

*Description of changes:* 
    * When running `aws configure` for new profiles, suggests `aws login` instead which does not require the usage of long-term IAM access keys.
    * Also backfills some examples for `aws login`

```
➜ aws configure --profile my-new-profile

Tip: You can now deliver temporary credentials to the AWS CLI using your AWS Console session. Simply run 'aws login'.

AWS Access Key ID [None]: 
```

### DIFF
--- a/awscli/customizations/configure/configure.py
+++ b/awscli/customizations/configure/configure.py
@@ -68,8 +68,8 @@ class ConfigureCommand(BasicCommand):
         'To create a new configuration::\n'
         '\n'
         '    $ aws configure\n'
-        '    Tip: You can now deliver temporary credentials to the AWS CLI\n'
-        '    using your AWS Console session. Simply run \'aws login\'.\n\n'
+        '    Tip: You can deliver temporary credentials to the AWS CLI using \n'
+        '    your AWS Console session by running the command \'aws login\'.\n\n'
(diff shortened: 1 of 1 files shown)


awscli/examples/configure/_description.rst
```diff
@@ -10,6 +10,10 @@ When you are prompted for information, the current value will be displayed in
 config file.  It does not use any configuration values from environment
 variables or the IAM role.
 
+Tip: You can now deliver temporary credentials to the AWS CLI using your AWS
+Console session. Simply run ``aws login``. For more information, see
```
kdaily: Nit: The temporal note will eventually get stale, so I think it would be good as a declarative tip. While running the command is simple, understanding how console session and the CLI work together might not be for some users (especially since there's an entire documentation page dedicated to it).

```suggestion
Tip: You can deliver temporary credentials to the AWS CLI using your AWS
Console session by running the command ``aws login``. For more information, see
```

ashovlin: Applied your suggestion, thanks!

kdaily (approved):
Nice, LGTM!

PR #10148, opened 2026-03-23
Limit --maxprocesses for scripts/ci/run-tests when in GitHub runners

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* 
* Reduce parallelization for unit/functional tests in GItHub to avoid the `Error: The operation was canceled.` we've been seeing recently on the GitHub runners. https://github.com/actions/runner/issues/2468 and other threads about it suggest it may be due to excessive resource consumption.
* Updated some actions pins while I was here to avoid Node deprecation warnings.

### DIFF
--- a/scripts/ci/run-tests
+++ b/scripts/ci/run-tests
@@ -52,7 +52,11 @@ def process_args(args):
         test_args += f'--junit-xml={args.junit_xml_path} '
     if args.markers:
         test_args += f'-m {args.markers} '
-    if os.getenv('GITHUB_ACTIONS', 'false') == 'true':
+    if os.getenv('PYTEST_XDIST_AUTO_NUM_WORKERS'):
+        # If the user set PYTEST_XDIST_AUTO_NUM_WORKERS,
+        # use that to limit max processes
+        max_processes = int(os.environ['PYTEST_XDIST_AUTO_NUM_WORKERS'])
+    elif os.getenv('GITHUB_ACTIONS', 'false') == 'true':
         # Limiting parallelism in GitHub, to avoid the
         # 'Error: The operation was canceled.' issues in 2026
         max_processes=3


ashovlin:
> `PYTEST_XDIST_AUTO_NUM_WORKERS`

In https://github.com/aws/aws-cli/pull/10148/commits/f7f227e68da719bd047ea8b9038b9f2fb0e2e49e, we now check if `PYTEST_XDIST_AUTO_NUM_WORKERS` is set first. If it is, we use that to set `--maxprocesses` to match the user's desired number. This will let users override our defaults without modifying the script.

scripts/ci/run-tests
```diff
@@ -52,7 +52,17 @@ def process_args(args):
...
+        # If the user set PYTEST_XDIST_AUTO_NUM_WORKERS,
+        # use that to limit max processes
+        max_processes = int(os.environ['PYTEST_XDIST_AUTO_NUM_WORKERS'])
+    elif os.getenv('GITHUB_ACTIONS', 'false') == 'true':
+        # Limiting parallelism in GitHub, to avoid the
+        # 'Error: The operation was canceled.' issues in 2026
+        max_processes=3
+    else:
+        max_processes=4
+    test_args += f'--numprocesses=auto --dist=loadfile --maxprocesses={max_processes} '
```
kdaily, line 65: I was thinking that `--max-processes` would no longer be needed if `PYTEST_XDIST_AUTO_NUM_WORKERS` was set and the fallback is needed for backwards compatibility if its not set:

```suggestion
    test_args += f'--numprocesses=auto --dist=loadfile '
    # If the user set PYTEST_XDIST_AUTO_NUM_WORKERS,
    # let pytest use that to limit max processes
    # Otherwise, fall back to 4 since that's what was
    # previously used in this script.
    if not os.getenv('PYTEST_XDIST_AUTO_NUM_WORKERS'):
        test_args += f'--maxprocesses=4 '
```

Then it would be specifically set to 3 in the GitHub Actions yaml configurations that run the test suites.

ashovlin: From discussion, planning to merge this for now since tests are really bad yesterday, we can keep tweaking if there are users setting `PYTEST_XDIST_AUTO_NUM_WORKERS`.

PR #10163, opened 2026-03-25
Update `configure set` to validate configuration keys and values

aemous:
*Description of changes:*

* Update `configure set` so that it returns an error if a newline or carriage return character is present in section/profile names, configuration keys, or values.
* Added tests to verify the error is returned and new expected behavior.

*Description of tests:*

* Passed all CI
* Successfully ran build workflow.

### DIFF
--- a/awscli/customizations/configure/writer.py
+++ b/awscli/customizations/configure/writer.py
@@ -24,11 +24,18 @@ class ConfigFileWriter(object):
         r'(?P<value>.*)$'
     )
 
-    def _validate_no_newlines(self, value, label='value'):
+    def _validate_no_newlines_or_carriage_returns(
+        self,
+        value,
+        label='value',
+        msg_override=None,
+    ):
         if isinstance(value, str) and ('\n' in value or '\r' in value):
-            raise ValueError(
-                f"Invalid {label}: newline characters are not allowed: {value!r}"
+            err_msg = msg_override if msg_override is not None else (
+                f"Invalid {label}: newline "
+                f"characters and carriage returns are not allowed: {value!r}"
             )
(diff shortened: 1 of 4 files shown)


awscli/customizations/configure/writer.py
```diff
@@ -24,6 +24,12 @@ class ConfigFileWriter(object):
         r'(?P<value>.*)$'
     )
 
+    def _validate_no_newlines(self, value, label='value'):
+        if isinstance(value, str) and ('\n' in value or '\r' in value):
+            raise ValueError(
+                f"Invalid {label}: newline characters are not allowed: {value!r}"
```
AndrewAsseily: I don't think we should include the raw value in the error message here. configure set handles sensitive keys like aws_secret_access_key ([and others](https://docs.aws.amazon.com/sdkref/latest/guide/feature-static-credentials.html)), so the {value!r} would end up dumping credentials to stderr, which could leak into logs. I think the error is clear enough without it.

AndrewAsseily: Nit: the changelog mentions "newline or carriage-return" but the error message just says "newline characters are not allowed." Might be worth making them consistent.

AndrewAsseily (changes requested):
Thanks! Looks good, just left two small comments. Same comments apply to the v2 port (https://github.com/aws/aws-cli/pull/10167).

awscli/customizations/configure/writer.py
aemous: Pushed a revision with addressed feedback.

AndrewAsseily (approved):
🏆

PR #10165, opened 2026-03-25
Improving caching of S3 endpoints by omitting unused parameters

ashovlin:
*Issue #, if available:* CLI-8066, variation of https://github.com/boto/botocore/pull/3529

*Description of changes:* When resolving the endpoint for S3 operations, currently `{'Key', 'Prefix', 'CopySource'}` are defined as parameters in S3's rules, but aren't currently referenced. This is an issue for endpoint caching, since objects in the same bucket result in a unique set of endpoint parameters. So when running `sync` or `cp -r` for lots of objects in the same bucket, we need to recalculate the endpoint each object.

Now, we drop these three parameters specifically before passing them into the rules. This comes with a test that will begin failing when the rules start referencing these parameters (or add new, unused parameters).


aemous (approved):
LGTM

PR #10166, opened 2026-03-26
Add per-service CLI flags for client context params

AndrewAsseily:
*Issue #, if available:* N/A

*Description of changes:*

This exposes clientContextParams from the service model as per-service CLI flags. Currently that's S3 (5 params) and S3Control (1 param).  The existing s3 config values still take priority.
Added model validation tests for name collisions between client context params and operation input members, and for unsupported param types (currently only string and boolean are supported).

Changelog type: I used enhancement since the underlying plumbing already exists in and this is just surfacing it in the CLI.


ashovlin (changes requested):
I haven't done a full pass on the code yet, but stepping through this in a debugger I am NOT seeing the new parameters influencing the endpoint rules with a breakpoint at `EndpointProvider`->`resolve_endpoint`.

`aws s3api list-objects-v2 --bucket <mrap arn> --disable-multi-region-access-points` is still working for me.

AndrewAsseily:
Fixed in the latest revision.

`aws s3api list-objects-v2 --bucket <mrap arn> --disable-multi-region-access-points` should now return `Invalid configuration: Multi-Region Access Point ARNs are disabled.` 

https://github.com/aws/aws-cli/blob/v2/awscli/botocore/data/s3/2006-03-01/endpoint-rule-set-1.json#L9257

PR #10169, opened 2026-03-27
[v2] Fix bug with auth_scheme_preference

aemous:
*Notes:*

* Currently, `auth_scheme_preference` only works as intended for modeled auth schemes at the service level (e.g. `bedrock`). It does not work for auth being resolved from other points in the precedence chain (e.g. EP2.0, operations).
  * For example, for S3 Outposts, users cannot use `auth_scheme_preference` to choose `sigv4` over `sigv4a`. Outposts models auth at the EP2.0 level, currently.

*Description of changes:*

* `auth_scheme_preference` is now used to reprioritize auth schemes, if configured, if the service models auth schemes at the **endpoint level**.
* `auth_scheme_preference` is now used to reprioritize auth schemes, if configured, if the service models auth schemes at the **operation level**.

*Description of tests:*

* Successfully ran pre-prod build workflow.
* Manually tested `auth_scheme_preference` shared config file using the following test cases:

## Endpoint-level auth test

User configures `sigv4` over `sigv4a`. We expect `sigv4` to be used.

```
auth_scheme_preference = sigv4,sigv4a
```

```
aws s3api get-object --bucket arn:aws:s3-outposts:us-east-1:ACCOUNT:outpost/op-REDACTED/accesspoint/my-outpost-ap-2 --key test-object.txt --region us-east-1 output.txt --debug
```

```
2026-03-27 16:12:17,315 - MainThread - botocore.endpoint - DEBUG - Sending http request: 
<AWSPreparedRequest stream_output=True, 
method=GET, 
url=https://my-outpost-ap-2-ACCOUNT.op-REDACTED.s3-outposts.us-east-1.amazonaws.com/test-object.txt, 
headers={
'x-amz-checksum-mode': b'ENABLED', 
'User-Agent': b'aws-cli/2.33.13 md/awscrt#0.31.2 ua/2.1 os/macos#25.3.0 md/arch#arm64 lang/python#3.12.6 md/pyimpl#CPython m/n,E,b,Z cfg/retry-mode#standard md/installer#source md/prompt#off md/command#s3api.get-object', 
'X-Amz-Date': b'20260327T201217Z', 
'X-Amz-Security-Token': b'REDACTED', 
'X-Amz-Content-SHA256': b'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', 
'Authorization': b'AWS4-HMAC-SHA256 Credential=REDACTED/us-east-1/s3-outposts/aws4_request, 
SignedHeaders=host;x-amz-checksum-mode;x-amz-content-sha256;x-amz-date;x-amz-security-token, Signature=077ee48a7141356c705cbe03042ef71cb47de9b9e366e7f80021582c3fb62b53'}>
```

Result: `AWS4-HMAC-SHA256` in the `Authorization` header value confirms `sigv4` was used.

-------

User configures `sigv4a` over `sigv4`. We expect `sigv4a` to be used.

```
auth_scheme_preference = sigv4a,sigv4
```

```
aws s3api get-object --bucket arn:aws:s3-outposts:us-east-1:ACCOUNT:outpost/op-REDACTED/accesspoint/my-outpost-ap-2 --key test-object.txt --region us-east-1 output.txt --debug
```

```
2026-03-27 16:17:07,382 - MainThread - botocore.endpoint - DEBUG - Sending http request: 
<AWSPreparedRequest stream_output=True, 
method=GET, 
url=https://my-outpost-ap-2-ACCOUNT.op-REDACTED.s3-outposts.us-east-1.amazonaws.com/test-object.txt, 
headers={
'x-amz-checksum-mode': b'ENABLED', 
'User-Agent': b'aws-cli/2.33.13 md/awscrt#0.31.2 ua/2.1 os/macos#25.3.0 md/arch#arm64 lang/python#3.12.6 md/pyimpl#CPython m/Z,E,n,S,b cfg/retry-mode#standard md/installer#source md/prompt#off md/command#s3api.get-object', 
'host': b'my-outpost-ap-2-ACCOUNT.op-REDACTED.s3-outposts.us-east-1.amazonaws.com', 
'X-Amz-Security-Token': b'', 
'X-Amz-Date': b'20260327T201707Z', 
'X-Amz-Region-Set': b'*', 
'x-amz-content-sha256': b'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', 
'Authorization': b'AWS4-ECDSA-P256-SHA256 Credential=REDACTED/s3-outposts/aws4_request, 
SignedHeaders=host;x-amz-checksum-mode;x-amz-content-sha256;x-amz-date;x-amz-region-set;x-amz-security-token, Signature=30450220429b7f10af468e9d806fcbd10a09a40b0f46ee26bc07881ea2eab3a4b9f90548022100a1578454f460ffc1f240fd85633ec5368a289e44072deff1ec2a747a7c5b7ccb'}>
```

Result: `AWS4-ECDSA-P256-SHA256` in the `Authorization` header value confirms `sigv4a` was used.

### DIFF
--- a/awscli/botocore/regions.py
+++ b/awscli/botocore/regions.py
@@ -700,6 +700,9 @@ class EndpointRulesetResolver:
         if self._requested_auth_scheme == UNSIGNED:
             return 'none', {}
 
+        available_ruleset_names = [
+            s['name'].split('#')[-1] for s in auth_schemes
+        ]
         auth_schemes = [
             {**scheme, 'name': self._strip_sig_prefix(scheme['name'])}
             for scheme in auth_schemes
@@ -723,8 +726,10 @@ class EndpointRulesetResolver:
                 return None, {}
         elif self._auth_scheme_preference is not None:
             prefs = self._auth_scheme_preference.split(',')
-            available_ruleset_names = [s['name'].split('#')[-1] for s in auth_schemes]
-            auth_schemes_by_auth_type = {s['name'].split('#')[-1]: s for s in auth_schemes}
+            auth_schemes_by_auth_type = {
+                self._strip_sig_prefix(s['name'].split('#')[-1]): s
+                for s in auth_schemes
+            }
             name = resolve_auth_scheme_preference(prefs, available_ruleset_names)
             scheme = auth_schemes_by_auth_type[name]
         else:


awscli/botocore/args.py
```diff
@@ -119,6 +119,7 @@ def get_client_args(
         s3_disable_express_session_auth = config_kwargs[
             's3_disable_express_session_auth'
         ]
+        preferred_auth_schemes = config_kwargs['auth_scheme_preference']
```
SamRemis: [Nit, non-blocking]
IMO this variable name should exactly the kwarg name so there's no confusion.  There are already a ton of variables around signature version.  In this file alone:

`signature_version` -> `sig_version` -> `requested_auth_scheme`, then also you're now adding `auth_scheme_preference` -> `preferred_auth_schemes`.   It's a lot of complexity to keep in your head, so while it doesn't make a huge difference, I'd prefer to see this renamed to just `auth_scheme_preference` to make it clear that it's the same thing as the config variable.

awscli/botocore/regions.py
```diff
@@ -698,6 +700,17 @@ def auth_schemes_to_signing_ctx(self, auth_schemes):
         if self._requested_auth_scheme == UNSIGNED:
             return 'none', {}
 
+        # if a preferred auth schemes list is provided, reorder the auth schemes
+        # list based on the preferred ordering.
+        if self._preferred_auth_schemes is not None:
```
SamRemis: [Change requested] 
We already have a function called `resolve_auth_scheme_preference` which handles this and should avoid duplicating preference-resolution / ordering logic here.

Right now we manually reorder the endpoint ruleset authSchemes list based on auth_scheme_preference. Instead, we can reuse resolve_auth_scheme_preference() to pick the desired auth type and then map that back to the actual ruleset scheme dict.  Here's the proof of concept I wrote:

```
        if self._requested_auth_scheme is not None:
             ....
        elif self._preferred_auth_schemes is not None:
            prefs = self._preferred_auth_schemes.split(',')
            available_ruleset_names = [s['name'].split('#')[-1] for s in auth_schemes]
            auth_schemes_by_auth_type = {self._strip_sig_prefix(s['name'].split('#')[-1]): s for s in auth_schemes}
            resolved_auth_type = resolve_auth_scheme_preference(prefs, available_ruleset_names)
            name = resolved_auth_type
            scheme = auth_schemes_by_auth_type[resolved_auth_type]
```

We could definitely refactor this and/or the `resolve_auth_scheme_preference` to make it much cleaner.  What do you think?

.changes/next-release/bugfix-signing-24412.json
```diff
@@ -0,0 +1,5 @@
+{
```
SamRemis, line 1: High level note:

Jonathan has a semi-related bugfix that will likely need to apply here as well - https://github.com/boto/botocore/pull/3663/changes.  

When we resolve v4a via auth scheme preference, we need to make sure we are also respecting the signing region set

awscli/botocore/regions.py
aemous: I think this makes sense. I will revise and go this route.

.changes/next-release/bugfix-signing-24412.json
aemous: Good callout. I approved and merged the v2 port. I'll now merge it into this branch and make sure the signing set is applied when v4a is resolved via preference in all cases (including EP2.0 and operation-level traits).

aemous: After merging in Jonathan's code to this branch, and refactoring the code based on your feedback, I have manually tested various resolution cases (EP2.0, operation-level modeled, with auth scheme preference and sigv4a signing region set configured, and everything seems to work correctly).

SamRemis: Thank you for testing :)

awscli/botocore/regions.py
```diff
@@ -698,6 +700,9 @@ def auth_schemes_to_signing_ctx(self, auth_schemes):
         if self._requested_auth_scheme == UNSIGNED:
             return 'none', {}
 
+        available_ruleset_names = [
```
SamRemis, line 703: [nit, non-blocking]

Since this is only used in the elif block, I think it would make more sense to define in that scope.

aemous: I tried that, but then im forced to make other code dirtier. Available ruleset names needs auth schemes before the sig prefix is stripped. Putting into the elif block would require us to put the sig prefix stripping code into the if and the else blocks.

PR #10177, opened 2026-04-01
Update protocol tests and handle null values in sparse map serialization

AndrewAsseily:
*Description of changes:*

PR #10110 added null handling for lists but not maps. This PR extends the same pattern to maps. Also fixes the CBOR map parser which was silently dropping null values instead of preserving them.

Also updates protocol tests to the latest.

```
tests/unit/botocore/test_protocols.py ......................................................................................................................................................................................................................... [ 22%]
............................................................................................................................................................................................................................................................... [ 49%]
............................................................................................................................................................................................................................................................... [ 76%]
...............................................................................................................................................................................................................................                                 [100%]

========================================================================================================================= 950 passed in 1.77s =========================================================================================================================
```


ashovlin (changes requested):
1. Can you add a changelog entry?
2. I see only the JSON and CBOR serializers are updated - do the rest handle nulls already?
3. Preferably in `protocol-tests-ignore-list.json`, but at least in the PR description - can you write why we're skipping each new skipped case? Do we want to preserve backwards compatibility, or do we have bugs that we're not addressing yet?

AndrewAsseily:
1. Added in revision

2. 

The other serializers don't handle nulls, they'd crash the same way. The original sparse work (PR #10110) only targeted parsing, not serialization. This PR adds null handling to the JSON and CBOR serializers because those are the only protocols with sparse test cases. I checked the service models and the only service with the sparse trait today is connectcases, which uses REST-JSON

That said, services could add sparse at any time, so it might be worth adding to the remaining serializers in this PR to be safe. Let me know what you think.

3. Ignore list

- **XmlAttributesInMiddle**: Backwards compatibility: Same issue as the already-ignored XmlAttributesOnPayload.

- **RestJsonHttpPayloadWithStructureAndEmptyResponseBody**: Backwards compatibility: Changing this could break customers accessing keys on the result without a None check.

- **QueryEmptyInputAndEmptyOutput / QueryNoInputAndOutput**: Bug: The parser crashes with KeyError when the result wrapper element is missing from the response body. Should handle this gracefully.

ashovlin:
> That said, services could add sparse at any time, so it might be worth adding to the remaining serializers in this PR to be safe. Let me know what you think.

Yeah, I think we should, especially if a service making use of it would result in a runtime failure as opposed to something we'd catch before launch via tests.

ashovlin (approved):

AndrewAsseily:
Discussed offline. We're skipping the XML- protocols (Query, EC2, REST-XML) in this PR. The sparse trait is effectively a no-op there.

PR #10195, opened 2026-04-08
[v2] Apply sigv4a_signing_region_set when SigV4A is resolved via auth scheme preference

jonathan343:
*This is a port of https://github.com/boto/botocore/pull/3663 for AWS CLI v2*

## Overview
Fix `sigv4a_signing_region_set` being ignored when SigV4A is selected via `auth_scheme_preference` or explicit `signature_version='v4a'`, causing `X-Amz-Region-Set` to fall back to the endpoint-derived region (e.g., `us-west-2` instead of `*`).

## Problem
SigV4A signing-context setup only ran inside `set_operation_specific_signer()` when `auth_type == 'v4a'`. This path isn't reached when SigV4A is resolved later by `_set_auth_scheme_preference_signer()` or configured directly on the client, so the region set was never applied.

## Solution
Extract the SigV4A signing-context update into a shared helper and call it whenever the resolved signer is `v4a`, preserving existing region-set precedence.

## Testing

Added unit coverage for:
- `auth_scheme_preference='sigv4a'` applying the configured SigV4A region set
- `explicit signature_version='v4a'` applying the configured SigV4A region set

---

### DIFF
--- a/.changes/next-release/bugfix-auth-28411.json
+++ b/.changes/next-release/bugfix-auth-28411.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "auth",
+  "description": "Fix ``sigv4a_signing_region_set`` config being ignored when SigV4a is selected via ``auth_scheme_preference``. The configured region set is now correctly applied to the signing context regardless of how SigV4a is resolved."
+}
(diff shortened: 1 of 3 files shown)


aemous (approved):
LGTM. I've also verified the signing region set config is correctly being used in AWS CLI v2 on this branch.

PR #10196, opened 2026-04-08
Tighten output file permissions

AndrewAsseily:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/customizations/s3events.py
+++ b/awscli/customizations/s3events.py
@@ -12,6 +12,8 @@
 # language governing permissions and limitations under the License.
 """Add S3 specific event streaming output arg."""
 
+import os
+
 from awscli.arguments import CustomArgument
 
 STREAM_HELP_TEXT = 'Filename where the records will be saved'
@@ -59,8 +61,7 @@ def replace_event_stream_docs(help_command, **kwargs):
             # This should never happen, but in the rare case that it does
             # we should be raising something with a helpful error message.
             raise DocSectionNotFoundError(
-                'Could not find the "output" section for the command: %s'
-                % help_command
+                f'Could not find the "output" section for the command: {help_command}'
             )
     doc.write('======\nOutput\n======\n')
     doc.write(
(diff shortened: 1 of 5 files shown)


tests/functional/s3api/test_select_object_content.py
```diff
@@ -82,6 +82,26 @@ def test_can_stream_to_file(self):
...
+            '--expression-type',
+            'SQL',
+            '--input-serialization',
+            '{"CSV": {}}',
+            '--output-serialization',
+            '{"CSV": {}}',
+            filename,
+        ]
+        self.assert_params_for_cmd(cmdline, ignore_params=True)
+        self.assertEqual(os.stat(filename).st_mode & 0o777, 0o600)
```
kdaily, line 109: nit: this might warrant a callout what the `& 0o777` does, I wasn't aware when reading (same in the other test)

awscli/customizations/s3events.py
```diff
@@ -124,7 +124,7 @@ def save_file(self, parsed, **kwargs):
         fd = os.open(
             self._output_file, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600
         )
-        os.fchmod(fd, 0o600)
+        os.chmod(self._output_file, 0o600)
```
kdaily, line 127: Why the change to `os.chmod`? If you could update the commit description, that would be helpful to future readers.

kdaily (changes requested):

awscli/customizations/s3events.py
AndrewAsseily: No strong preference, `os.chmod` with the path is simpler and consistent with other places in the codebase

tests/functional/s3api/test_select_object_content.py
AndrewAsseily: Added a comment in revision

kdaily (approved):

PR #10197, opened 2026-04-08
Tighten output file permissions

AndrewAsseily:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.changes/next-release/bugfix-s3streamingoutput-73196.json
+++ b/.changes/next-release/bugfix-s3streamingoutput-73196.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "s3, streaming output",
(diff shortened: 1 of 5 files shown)


kdaily (approved):

andygeorge:
FWIW this tripped us up, as we were doing something like this:

```
aws lambda invoke ... /dev/null
```

...to suppress the lambda output. With this PR, this changed our `/dev/null` permissions, causing unintended behavior.

AndrewAsseily:
Thanks for reporting this @andygeorge, taking a look.

AndrewAsseily:
Hey @andygeorge, thanks again for reporting. We've rolled back the change and it'll be included in today's release, so this should be resolved for you shortly.

PR #10205, opened 2026-04-09
Tighten CodeDeploy config file permissions

aahallal:
Changed the CodeDeploy on-premises instance configuration file creation to use os.open() with explicit permissions (0o600) instead of the default open() function. This ensures the configuration file is created with owner-only read/write permissions, improving security by preventing other users on the system from accessing potentially sensitive configuration data.


AndrewAsseily (approved):
Thanks! 🏆

PR #10212, opened 2026-04-10
[v2] Auto-detect account regional namespace buckets in s3 mb

AndrewAsseily:
*Issue #, if available:*

*Description of changes:*

The s3 mb command now automatically detects account regional namespace bucket names (matching *-<accountId>-<region>-an) and sets the x-amz-bucket-namespace: account-regional header on the CreateBucket request. Previously, users had to fall back to s3api create-bucket --bucket-namespace for 
this workflow.

v1 port: https://github.com/aws/aws-cli/pull/10213

### DIFF
--- a/.changes/next-release/enhancement-s3-61261.json
+++ b/.changes/next-release/enhancement-s3-61261.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "s3",
+  "description": "Add support for creating S3 account regional namespace buckets with aws s3 mb. The command now automatically detects bucket names matching the account-regional naming pattern and sets the required x-amz-bucket-namespace header."
+}
--- a/.changes/next-release/enhancement-s3-61261.json
+++ b/.changes/next-release/enhancement-s3-61261.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
-  "category": "s3",
-  "description": "Add support for creating S3 account regional namespace buckets with aws s3 mb. The command now automatically detects bucket names matching the account-regional naming pattern and sets the required x-amz-bucket-namespace header."
+  "category": "``S3``",
+  "description": "Add support for creating S3 account regional namespace buckets with ``aws s3 mb``. The command now automatically detects bucket names matching the account-regional naming pattern and sets the required ``x-amz-bucket-namespace`` header."
 }
--- a/awscli/customizations/s3/utils.py
+++ b/awscli/customizations/s3/utils.py
@@ -56,10 +56,6 @@ _S3_OUTPOST_BUCKET_ARN_TO_BUCKET_KEY_REGEX = re.compile(
     r'[a-zA-Z0-9\-]{1,63})[/:]?(?P<key>.*)$'
 )
 
-_S3_ACCOUNT_REGIONAL_NAMESPACE_REGEX = re.compile(
-    r'^.+-\d{12}-[a-z]{2}(-[a-z]+-\d+)?-an$'
-)
-
 _S3_OBJECT_LAMBDA_TO_BUCKET_KEY_REGEX = re.compile(
     r'^(?P<bucket>arn:(aws).*:s3-object-lambda:[a-z\-0-9]+:[0-9]{12}:'
     r'accesspoint[/:][a-zA-Z0-9\-]{1,63})[/:]?(?P<key>.*)$'
@@ -67,7 +63,7 @@ _S3_OBJECT_LAMBDA_TO_BUCKET_KEY_REGEX = re.compile(
 
 
 def is_account_regional_namespace_bucket(bucket):
-    return bool(_S3_ACCOUNT_REGIONAL_NAMESPACE_REGEX.match(bucket))
+    return bucket.endswith('-an')
 
 
 def human_readable_size(value):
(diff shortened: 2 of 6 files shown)


.changes/next-release/enhancement-s3-61261.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "s3",
```
aemous, line 3:
```suggestion
  "category": "``S3``",
```

.changes/next-release/enhancement-s3-61261.json
```diff
@@ -0,0 +1,5 @@
...
+  "description": "Add support for creating S3 account regional namespace buckets with aws s3 mb. The command now automatically detects bucket names matching the account-regional naming pattern and sets the required x-amz-bucket-namespace header."
```
aemous, line 4:
```suggestion
  "description": "Add support for creating S3 account regional namespace buckets with ``aws s3 mb``. The command now automatically detects bucket names matching the account-regional naming pattern and sets the required ``x-amz-bucket-namespace`` header."
```

awscli/customizations/s3/utils.py
```diff
@@ -56,12 +56,20 @@
     r'[a-zA-Z0-9\-]{1,63})[/:]?(?P<key>.*)$'
 )
 
+_S3_ACCOUNT_REGIONAL_NAMESPACE_REGEX = re.compile(
+    r'^.+-\d{12}-[a-z]{2}(-[a-z]+-\d+)?-an$'
```
aemous, line 60: I have concerns for this regex.

According to the [S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html):

> Bucket names can only end with the suffix -an when you are creating buckets in your account regional namespace. For more information, see [Namespaces for general purpose buckets](https://docs.aws.amazon.com/AmazonS3/latest/userguide/gpbucketnamespaces.html).

If our mb command is making a best-effort attempt at creating buckets, then I think we should always add the regional namespace header when the suffix is `-an`.  This would match Console behavior, the front-end does client-side validation that if it ends with `-an` it must be for regional account namespace.

S3 will not allow users to create a bucket that ends with `-an` unless the header is provided, so I think it is on us to provide the header whenever that suffix is provided.

For example, I pulled down your code and tried this command:

```
$ aws s3 mb s3://xyz-an     
        
make_bucket failed: s3://xyz-an An error occurred (MissingNamespaceHeader) when calling the CreateBucket operation: The requested bucket is an account-regional namespace bucket, but your request is missing the required x-amz-bucket-namespace header.
```

I feel this error should not be encountered by our users, since we are in charge of supplying headers.

If the rest of the formatting doesn't match the expected format (e.g user gives bucket name `xyz-an`), S3 will give an API error that the name doesn't match the format. In my view that's ideal since they should remain the authoritative source of truth on the right format, rather than us trying to duplicate their formatting logic in the client.

This would also match how we handle S3 express buckets, we only check for the `-x-s3` suffix, even though S3 requires a zone too.

tests/functional/s3/test_mb_command.py
```diff
@@ -47,48 +47,74 @@ def test_incompatible_with_express_directory_bucket(self):
...
-            }
+                    {'Key': 'Key2', 'Value': 'Value2'},
+                ],
+            },
         }
         self.assert_params_for_cmd(command, expected_params)
 
+    def test_account_regional_namespace_bucket(self):
+        bucket = 'amzn-s3-demo-bucket-111122223333-us-west-2-an'
+        command = self.prefix + 's3://%s --region us-west-2' % bucket
```
aemous, line 93: can we use f-strings on new code instead of formatting operator?

tests/functional/s3/test_mb_command.py
```diff
@@ -47,48 +47,74 @@ def test_incompatible_with_express_directory_bucket(self):
...
+    def test_account_regional_namespace_bucket_us_east_1(self):
+        bucket = 'my-bucket-111122223333-us-east-1-an'
+        command = self.prefix + 's3://%s --region us-east-1' % bucket
+        expected_params = {
+            'Bucket': bucket,
+            'BucketNamespace': 'account-regional',
+        }
+        self.assert_params_for_cmd(command, expected_params)
+
+    def test_regular_bucket_no_namespace(self):
```
aemous, line 111: worth adding a test for bucket name `xyz-an`.

aemous (changes requested):
Left some comments.

Also, consider adding ~1 integration test. We are making some assumption on server behavior so it may be worth baking that into tests. Such as, attempt to create a bucket with name `xyz-an`, and expect s3 to return an error that the formatting is wrong.

awscli/customizations/s3/utils.py
AndrewAsseily: Addressed in https://github.com/aws/aws-cli/pull/10212/commits/5246e898be2eed46c848e913cceba50d167e2326

aemous (dismissed):
LGTM. 

I approve merge assuming all CI pass. At least 1 new integ test may be added before merge as long as they have been tested.

aemous (dismissed):
LGTM assuming CI passes and the new integ test was tested.

🏆

aemous (approved):
LGTM if all CI pass.

PR #10213, opened 2026-04-10
Auto-detect account regional namespace buckets in s3 mb

AndrewAsseily:
*Issue #, if available:*

*Description of changes:*

The s3 mb command now automatically detects account regional namespace bucket names (matching *-<accountId>-<region>-an) and sets the x-amz-bucket-namespace: account-regional header on the CreateBucket request. Previously, users had to fall back to s3api create-bucket --bucket-namespace for 
this workflow.

v2: https://github.com/aws/aws-cli/pull/10212


aemous (approved):
LGTM assuming all CI pass.

PR #10220, opened 2026-04-13
Upgrade debug mode: Add detection for default checksum algorithm breaking change

aemous:
*Description of changes:*

* Added support to upgrade debug mode for detecting CRC32-by-default-in-v1 vs. CRC64NVME-by-default-in-v2.

*Description of tests:*

* Passed all CI (including 4 new functional tests added to verify expected behavior of the new functionality).
* Succeeded manual testing described below

### Manual testing

To test the code changes, I ran the following commands:

```
aws s3 cp ./file.txt s3://bucket --v2-debug
```

I verified that the new warning is printed to STDERR.

Then, I also ran

```
aws s3 cp ./file.txt s3://bucket --checksum-algorithm CRC32 --v2-debug
```

And i verified that the new warning is *not* printed to STDERR.


awscli/customizations/s3/subcommands.py
```diff
@@ -1144,6 +1144,24 @@ def run(self):
...
                 )
+            elif (
+                operation_name == 'upload'
+                and self.parameters.get('checksum_algorithm') is None
+            ):
+                uni_print(
+                    '\nAWS CLI v2 UPGRADE WARNING: In AWS CLI v2, for '
+                    '`aws s3` commands that upload a file to an S3 bucket, '
+                    'Cyclic Redundancy Check 64 (CRC64NVME) will be used to '
+                    'compute object checksums by default and include it in '
```
kdaily: nit: grammar


```suggestion
                    'compute object checksums by default and included in '
```

Also, there isn't really any context on what a "request" is in this scenario. Is this necessary for users to understand the change in behavior?

awscli/customizations/s3/subcommands.py
```diff
@@ -1144,6 +1144,24 @@ def run(self):
...
+                    'the request. This is different from v1 behavior, where '
+                    'Cyclic Redundancy Check 32 (CRC32) checksums will '
+                    'instead be used. For guidance on retaining v1 behavior '
```
kdaily: nit: Grammar. Isn't it currently used?

```suggestion
                    'Cyclic Redundancy Check 32 (CRC32) checksums '
                    'are used. For guidance on retaining v1 behavior '
```

awscli/customizations/s3/subcommands.py
```diff
@@ -1144,6 +1144,24 @@ def run(self):
                     '#cliv2-migration-s3-copy-metadata.\n\n',
                     out_file=sys.stderr
                 )
+            elif (
+                operation_name == 'upload'
+                and self.parameters.get('checksum_algorithm') is None
+            ):
+                uni_print(
+                    '\nAWS CLI v2 UPGRADE WARNING: In AWS CLI v2, for '
```
kdaily, line 1152: nit: From the work on standardized error output, shouldn't warnings be prefixed with `aws: [WARNING]`?

aemous: 1. This is consistent with all the other existing upgrade warnings. If we standardize this one, we should standardize all of them, and I'd rather not expand scope to updating all of these warning messages now.
2. Looks like we only released standardized error format to `v2` and not `v1` (https://github.com/aws/aws-cli/pull/10094). I would be more concerned about this if this was `v2`.

For the above reasons, and because the comment is a nit, I would like to keep the current state.

kdaily (approved):

PR #10223, opened 2026-04-14
[v2] Double-quote Windows alias command args if metacharacters present

hssyoo:
`_windows_shell_quote()` only escapes for the MSVC C runtime argv parser (backslashes and double quotes) but
does not account for `cmd.exe` metacharacters. When `ExternalAliasCommand` passes user-supplied arguments through `subprocess.call(command, shell=True)`, the string is interpreted by `cmd.exe` before reaching the target process, meaning metacharacters aren't properly escaped.

This PR fixes the issue by double-quoting the alias value string when `shell=True` and cmd.exe metacharacters are present, automatically treating metacharacters as literals.

Ideally we could collapse `_windows_argv_quote` and `_windows_cmd_shell_quote` into a single function, but I'm hesitant to change the code path for other CLI features that use `compat_shell_quote`, especially features that don't spawn a subprocess with `shell=True` like `ec2-instance-connect ssh`. 

**NOTE:** `%VAR%` expansion and `!VAR!` delayed expansion cannot be reliably escaped inside double quotes in `cmd.exe /c`. `%%` doubling only works in batch file mode. There is no known reliable escape for `%` or `!` inside double quotes on the command line.

### DIFF
--- a/awscli/compat.py
+++ b/awscli/compat.py
@@ -261,13 +261,14 @@ def compat_input(prompt):
         # This is unused directly when the user pastes the cross-device
         # verification code, which may be longer than some terminal's buffers
         import readline  # noqa: F401
+
         return raw_input()
     except ImportError:
         LOG.debug('readline module not available')
         return raw_input()
 
 
-def compat_shell_quote(s, platform=None):
+def compat_shell_quote(s, platform=None, shell=False):
     """Return a shell-escaped version of the string *s*
 
     Unfortunately `shlex.quote` doesn't support Windows, so this method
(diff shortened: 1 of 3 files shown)


awscli/compat.py
```diff
@@ -345,6 +352,60 @@ def _windows_shell_quote(s):
...
+    command line and are not handled here.
+
+    :param s: A string to escape
+    :return: An escaped string
+    """
+    if not s:
+        return '""'
+
+    buff = []
+    num_backspaces = 0
```
ashovlin: nit: `num_backslashes`? Though see it was already like this in existing function.

ashovlin (dismissed):

awscli/compat.py
hssyoo: Fixed via https://github.com/aws/aws-cli/pull/10223/changes/0c01e473eb7e67cbd50c38a837f75f215c691ea2

ashovlin (approved):

PR #10224, opened 2026-04-15
Expose --delay and --max-attempts for wait commands (#1295)

### LINKED ISSUE #1295: Expose parameters for 'delay' and 'max-attempts'
I'm trying to migrate to the unified cli, and to use waiters instead of polling loops, but the conversion-task-completed waiter consistently times out in us-east-1 for my 12GB disk.

It reports 
Waiter ConversionTaskCompleted failed: Max attempts exceeded
but the task does complete successfully. I can write a loop which calls the waiter, but I'd prefer to increase the number of attempts, or reduce their frequency. They appear to be taken from botocore/data/aws/ec2/2014-09-01.waiters.json, which contains the following:

"ConversionTaskCompleted": {      "delay": 15,      "operation":
"DescribeConversionTasks",      "maxAttempts": 40,

I'd like these to be exposed as parameters in the aws cli command, or perhaps for the timeout to be derived from the size of the disk being converted.

minjcho:
*Issue #, if available:*
Fixes #1295 

*Description of changes:*
* Add ``--delay`` and ``--max-attempts`` arguments to waiter commands.
* Pass the provided values through ``WaiterConfig`` when invoking ``waiter.wait()``.
* Update waiter help text to document the new options.
* Add unit coverage for argument parsing, help output, and waiter config injection.

*Description of tests:*
* Added unit tests covering argument parsing, help output, and waiter config injection. 
* Ran `python -m pytest tests/unit/customizations/test_waiters.py -q` — all pass.
* Succeeded manual testing described below. 
 
*Manual testing:*

Verified the new flags appear in the generated help output:

```
aws ec2 wait instance-running help
```

I confirmed that `[--delay <value>]` and `[--max-attempts <value>]` are listed in the synopsis and documented in the options section.

Then I ran:

```
aws ec2 wait instance-running --instance-ids <YOUR_INSTANCE_ID> --delay 5 --max-attempts 10
```

### DIFF
--- a/.changes/next-release/waiter-delay-max-attempts.json
+++ b/.changes/next-release/waiter-delay-max-attempts.json
@@ -1,0 +1,7 @@
+[
+  {
+    "category": "waiter",
+    "description": "Add ``--delay`` and ``--max-attempts`` arguments to all ``wait`` commands, allowing users to override the default polling interval and maximum number of attempts.",
+    "type": "enhancement"
+  }
+]
--- a/awscli/customizations/waiters.py
+++ b/awscli/customizations/waiters.py
@@ -20,6 +20,7 @@ from awscli.customizations.commands import (
     BasicDocHandler,
     BasicHelp,
 )
+from awscli.customizations.exceptions import ParamValidationError
 
 
 DELAY_HELP = (
@@ -74,6 +75,16 @@ class WaiterArgument(BaseCLIArgument):
 
     def add_to_params(self, parameters, value):
         if value is not None:
+            if self._serialized_name == 'Delay' and value < 0:
+                raise ParamValidationError(
+                    '--delay must be a non-negative integer, '
+                    'got %s' % value
+                )
+            if self._serialized_name == 'MaxAttempts' and value < 1:
+                raise ParamValidationError(
+                    '--max-attempts must be a positive integer, '
+                    'got %s' % value
+                )
             waiter_config = parameters.get('WaiterConfig', {})
             waiter_config[self._serialized_name] = value
             parameters['WaiterConfig'] = waiter_config
--- a/awscli/customizations/waiters.py
+++ b/awscli/customizations/waiters.py
@@ -10,15 +10,73 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from botocore import xform_name
+from botocore import model, xform_name
 from botocore.exceptions import DataNotFoundError
 
+from awscli.arguments import BaseCLIArgument
 from awscli.clidriver import ServiceOperation
 from awscli.customizations.commands import (
     BasicCommand,
     BasicDocHandler,
     BasicHelp,
 )
+
+
+DELAY_HELP = (
+    '<p>The amount of time in seconds to wait between attempts. '
+    'If not specified, the default delay for the waiter is used.</p>'
+)
+
+MAX_ATTEMPTS_HELP = (
+    '<p>The maximum number of attempts to be made. '
+    'If not specified, the default max attempts value for the '
+    'waiter is used.</p>'
+)
+
+
+class WaiterArgument(BaseCLIArgument):
+
+    def __init__(self, name, documentation, serialized_name):
+        self.argument_model = model.Shape(
+            'WaiterArgument', {'type': 'integer'}
+        )
+        self._name = name
+        self._serialized_name = serialized_name
+        self._documentation = documentation
+        self._required = False
+
+    @property
+    def cli_name(self):
+        return '--' + self._name
+
+    @property
+    def cli_type_name(self):
+        return 'integer'
+
+    @property
+    def required(self):
+        return self._required
+
+    @required.setter
+    def required(self, value):
+        self._required = value
+
+    @property
+    def documentation(self):
+        return self._documentation
+
+    def add_to_parser(self, parser):
+        parser.add_argument(
+            self.cli_name,
+            dest=self.py_name,
+            type=int,
+        )
+
+    def add_to_params(self, parameters, value):
+        if value is not None:
+            waiter_config = parameters.get('WaiterConfig', {})
+            waiter_config[self._serialized_name] = value
+            parameters['WaiterConfig'] = waiter_config
 
 
 def register_add_waiters(cli):
@@ -203,7 +261,9 @@ class WaiterStateDocBuilder:
         description = (
             ' It will poll every %s seconds until a successful state '
             'has been reached. This will exit with a return code of 255 '
-            'after %s failed checks.' % (delay, max_attempts)
+            'after %s failed checks. You can override the default polling '
+            'behavior with ``--delay`` and ``--max-attempts``.'
+            % (delay, max_attempts)
         )
         return description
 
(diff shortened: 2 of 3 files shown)


minjcho:
**Added input validation for `--delay` and `--max-attempts`**

Since other CLI params don't check non-positive values at the CLI layer (they rely on botocore's `ParamValidator` or the AWS server to reject them), I initially didn't add validation here either.

But I found that `WaiterConfig` is a special case — it's popped from kwargs via `kwargs.pop('WaiterConfig', {})` before reaching `ParamValidator` or the API server, and consumed locally in `time.sleep()` and a loop counter. **No layer validates these values:**

- `--delay -1` → `ValueError: sleep length must be non-negative` (Python runtime error from `time.sleep(-1)`)
- `--delay 0` → rapid-fire API calls with no sleep interval
- `--max-attempts 0` or `-1` → polls once, then immediately raises `WaiterError`

Added validation in `WaiterArgument.add_to_params` matching the waiter JSON schema constraints (`delay >= 0`, `maxAttempts >= 1` from `botocore/tests/functional/test_waiter_config.py`):

```bash
$ aws ec2 wait instance-running --instance-ids i-123 --delay -1
aws: [ERROR]: --delay must be a non-negative integer, got -1

$ aws ec2 wait instance-running --instance-ids i-123 --max-attempts 0
aws: [ERROR]: --max-attempts must be a positive integer, got 0
```

Added 7 unit tests covering rejection and acceptance boundaries.

I'll also file a separate issue on botocore for adding the same validation in `Waiter.wait()`, since boto3 users are equally affected.

.changes/next-release/waiter-delay-max-attempts.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "category": "waiter",
+    "description": "Add ``--delay`` and ``--max-attempts`` arguments to all ``wait`` commands, allowing users to override the default polling interval and maximum number of attempts.",
```
aemous, line 4: This changelog entry looks hand-made. Did you use our automated script?

Please run `./scripts/new-change` to automatically generate a changelog entry with the proper format and name. Then, add the generated changelog entry to this PR and delete this hand-made one.

tests/unit/customizations/test_waiters.py
```diff
@@ -166,6 +167,13 @@ def test_wait_state_help_command(self):
         self.assert_contains('``--filters`` (list)')
         self.assert_contains('======\nOutput\n======\n\nNone')
 
+    def test_wait_state_help_shows_waiter_params(self):
+        self.driver.main(['ec2', 'wait', 'instance-running', 'help'])
+        self.assert_contains('[--delay <value>]')
+        self.assert_contains('[--max-attempts <value>]')
+        self.assert_contains('``--delay``')
+        self.assert_contains('``--max-attempts``')
```
aemous, line 175: These two assertions are redundant, right?

If the output contains `[--delay <value>]`, the since `[--delay <value>]` contains `--delay`, we immediately know the output contains `--delay`.

Please either remove the redundant assertions or update them so they are not redundant.

tests/unit/customizations/test_waiters.py
```diff
@@ -203,6 +211,53 @@ def test_rds_jobs_complete(self):
...
+        cmdline += ' --delay 10 --max-attempts 100'
+        # WaiterConfig is popped by botocore's Waiter.wait() before
+        # the API call, so only API params appear in last_kwargs.
+        result = {'InstanceIds': ['i-12345678']}
+        self.parsed_response = {
+            'Reservations': [{'Instances': [{'State': {'Name': 'running'}}]}]
+        }
+        self.assert_params_for_cmd(cmdline, result)
+
+    def test_ec2_wait_forwards_waiter_config_end_to_end(self):
```
aemous, line 226: This is not a unit test since it is not testing a single class/function in isolation. Please either move this test under `tests/functional/test_waiters.py` or update the test so that it is a unit test.

The same is true for `test_ec2_wait_with_delay_and_max_attempts` and `test_ec2_wait_without_waiter_flags_omits_waiter_config`.

awscli/customizations/waiters.py
```diff
@@ -203,7 +272,9 @@ def _build_polling_description(self, delay, max_attempts):
         description = (
             ' It will poll every %s seconds until a successful state '
             'has been reached. This will exit with a return code of 255 '
-            'after %s failed checks.' % (delay, max_attempts)
+            'after %s failed checks. You can override the default polling '
+            'behavior with ``--delay`` and ``--max-attempts``.'
+            % (delay, max_attempts)
```
aemous, line 277: Can we switch from using the formatting operator `%` to f-strings to better align with our formatting?

tests/unit/customizations/test_waiters.py
```diff
@@ -399,3 +470,129 @@ def test_invoke(self):
...
+        waiter_caller.invoke(
+            'myservice', 'MyWaiter', parameters, parsed_globals
+        )
+
+        waiter.wait.assert_called_with(
+            Foo='bar',
+            WaiterConfig={'Delay': 10, 'MaxAttempts': 50},
+        )
+
+    def test_invoke_without_waiter_config(self):
```
aemous, line 497: Isn't this test redundant to the existing `test_invoke`? `test_invoke` does not supply waiter params, and it uses `assert_called_with`, so if any waiter params are supplied to `wait`, it would fail with an `AssertionError`.

tests/unit/customizations/test_waiters.py
```diff
@@ -399,3 +470,129 @@ def test_invoke(self):
...
+        waiter_caller = WaiterCaller(session, waiter_name)
+        waiter_caller.invoke(
+            'myservice', 'MyWaiter', parameters, parsed_globals
+        )
+
+        # Without set_waiter_config, no WaiterConfig should be passed.
+        waiter.wait.assert_called_with(Foo='bar')
+
+
+class TestWaiterArgument(unittest.TestCase):
```
aemous, line 518: For net-new test suites let's use PyTest instead of extending the older `unittest.TestCase` base class.

i.e. migrate `TestWaiterArgument` to use PyTest.

tests/unit/customizations/test_waiters.py
```diff
@@ -399,3 +470,129 @@ def test_invoke(self):
...
+        self.assertEqual(arg.cli_type_name, 'integer')
+
+    def test_required_defaults_to_false(self):
+        arg = WaiterArgument('delay', 'help text', 'Delay')
+        self.assertFalse(arg.required)
+
+    def test_delay_rejects_negative(self):
+        arg = WaiterArgument('delay', 'help text', 'Delay')
+        params = {}
+        with self.assertRaises(Exception) as ctx:
```
aemous, line 563: Can we narrow down the scope of the expected Exception to the actual exception that gets raised (`ParamValidationError`)?

Similar comment for `test_delay_rejects_negative_large`, `test_max_attempts_rejects_zero`, and `test_max_attempts_rejects_negative`.

tests/unit/customizations/test_waiters.py
```diff
@@ -166,6 +167,13 @@ def test_wait_state_help_command(self):
         self.assert_contains('``--filters`` (list)')
         self.assert_contains('======\nOutput\n======\n\nNone')
```
aemous, line 169: I saw we added tests that validate there's an error when passing negative values for delay/max-attempts to `WaiterCaller`. Can we also add functional tests that validate there's an error end-to-end when running a command that uses negative values for these new params? The new test(s) would live in `tests/functional/test_waiters.py`.

aemous (changes requested):
Thanks for the contribution. I requested some changes and asked a few questions.

RyanFitzSimmonsAK:
I'm setting this pull request to Draft due to author inactivity.  If you implement the requested changes, please feel free to reopen it.

Issue #9011, opened 2024-10-25
 AWS CLI slow startup


gaul:
### Describe the bug

AWS CLI is three times as slow as s3cmd when tested on a local S3 implementation.  s3fs-fuse invokes AWS CLI hundreds of times so this slows down integration tests.  I tested on an Intel i7-1250U.  I previously reported these symptoms in #6500.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

```
$ time s3cmd ls s3://

real    0m0.091s
user    0m0.073s
sys     0m0.016s
```

### Current Behavior

```
$ time AWS_EC2_METADATA_DISABLED=true aws s3 ls --region us-east-1 --endpoint-url http://127.0.0.1:8081 --profile s3proxy

real    0m0.346s
user    0m0.284s
sys     0m0.055s
```

### Reproduction Steps

See above

### CLI version used

aws-cli/2.17.18 Python/3.13.0 Linux/6.11.4-300.fc41.x86_64 source/x86_64.fedora.41

### Environment details (OS name and version, etc.)

Fedora 41

tim-finnigan:
Thanks for reaching out. Changing this to a feature request for increasing the speed. Running CLI commands with the `--debug` parameter will show how long each step of that process is taking. There are several steps involved, such as loading models and processing configurations. Can you provide any more details regarding your specific use case? For more information on optimizing S3 performance you can refer to the [S3 User Guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html).

gaul:
I attached the requested logs: [aws-cli.log](https://github.com/user-attachments/files/17564470/aws-cli.log)

My use case is running the CLI in CI hundreds of times per instance: https://github.com/s3fs-fuse/s3fs-fuse/blob/master/test/integration-test-main.sh .  A typical operation is `s3api head-object` or `s3 ls` -- these are small numbers and sizes of objects.

gaul:
I addressed my symptom by using `curl --aws-sigv4` instead of the AWS CLI.  This reduced CI run-time from 14-15 minutes to 10-11: https://github.com/s3fs-fuse/s3fs-fuse/pull/2689

PR #10231, opened 2026-04-20
[v2] Implement conditional prompt_toolkit importing

aemous:
*Notes:*

* Contributes to #9011.

*Description of changes:*

* Refactored autoprompt code so that resolving the autoprompt mode does not require importing autoprompt or instantiating a `AutoPromptDriver` object. The mode resolution code has been moved to `clidriver.py` as the standalone functions` resolve_auto_prompt_mode` and `validate_auto_prompt_args_are_mutually_exclusive`. `AutoPromptDriver` is now only imported lazily when autoprompt mode is actually active.
* Moved `resolve_mode` and `validate_auto_prompt_args_are_mutually_exclusive` tests from `tests/unit/autoprompt/test_core.py` to `tests/unit/test_clidriver.py` to reflect the new location of those functions.
* Separated `PrompterKeyboardInterrupt` into a new minimal `awscli/autoprompt/exceptions.py` module, so that `awscli.errorhandler` can import it without pulling in `awscli.autoprompt.factory` (which has many top-level `prompt_toolkit` imports).
* Made the `Display` import in `ecs/monitorexpressgatewayservice.py` lazy, deferring the import of `prompt_toolkit_display` (and therefore `prompt_toolkit`) until the interactive ECS monitoring display is actually used.
* Made the imports of `PTKPrompt` and `RequiredInputValidator` from `configure/sso.py` in `customizations/login/login.py` lazy, deferring the import of `configure/sso.py` until the login region prompt is actually needed.
* Made imports of `StartLiveTailCommand` and `TailCommand` in `customizations/logs/__init__.py` lazy, deferring the import of `prompt_toolkit` until the `aws logs` subcommand is invoked (i.e. the logs subcommand table is built).
* Extracted `ConfigureSSOCommand`, `ConfigureSSOSessionCommand`, and their shared base class `BaseSSOConfigurationCommand` into a new `customizations/configure/sso_commands.py` module, which has no `prompt_toolkit` imports. This allows `configure/configure.py` to keep its reference to these SSO commands and defer the import of `prompt_toolkit` (via `configure/sso.py`) until `aws configure sso` or `aws configure sso-session` is actually invoked.
* Made the import of `select_menu` from `wizard/ui/selectmenu` in `configure/sso.py` lazy, deferring the import of `wizard/ui/__init__.py` (which imports `prompt_toolkit`) until `ConfigureSSOCommand` is actually instantiated.
* Made the import of `create_wizard_app` from `wizard/factory` in `wizard/devcommands.py` lazy, deferring the import of the wizard UI chain (which imports `prompt_toolkit`) until a wizard dev command is actually run.
* Made the import of `wizard/factory` in `wizard/commands.py` lazy by introducing a `_get_runner` method on `TopLevelWizardCommand`, deferring the import of the wizard factory and UI chain until a wizard command is actually executed. 
* Moved the import of `devcommands` in `awscli/customizations/wizard/commands.py` from module-level to within the `register_wizard_commands` function. Although this *currently* gives no behavioral difference since `register_wizard_commands` is always called for every command, in the future when plugin/customization initialization is only called when a relevant command is entered (lazy plugin loading), this change will lead to lazy-importing of `devcommands` only when it's needed.
* Moved code out of `awscli/customizations/logs/startlivetail.py` into a new `awscli/customizations/logs/ui.py` file to defer imports of `prompt_toolkit` to when the `start-live-tail` commands is run.
* Added a new functional test that verifies `prompt_toolkit` is not in `sys.modules` for a handful of chosen commands.

*Description of tests:*

* Successfully ran pre-prod build workflow.
* Benchmarked before and after this patch (results below)

### Benchmark Results: Performance Improvement Analysis

The following benchmarking analysis evaluates the impact of the **Conditional Autoprompt** feature across two standard CLI operations. Tests were performed with 2,000 samples per state to ensure statistical precision.

Tested on m7i.xlarge EC2 instance, x86 architecture, AL2023 os, 64 GB storage, 2,000 iterations each.

#### **Consolidated Performance Gains**

| CLI Command | Metric | Before Patch | After Patch | Change |
| :--- | :--- | :--- | :--- | :--- |
| **aws --version** | Mean Time | 609.883 ms | 528.471 ms | **-81.412 ms** |
| | stdev. | 14.9329 ms | 11.6255 ms | -3.3074 ms |
| | SE | 0.3339 ms | 0.2600 ms | — |
| **aws sts get-caller-identity** | Mean Time | 697.245 ms | 615.703 ms | **-81.542 ms** |
| | stdev. | 9.62246 ms | 9.41794 ms | -0.20452 ms |
| | SE | 0.2152 ms | 0.2106 ms | — |

#### **Statistical Significance & Reliability**
To filter out environmental noise (CPU spikes, OS scheduling), we measured the performance shift in terms of **Standard Error (SE)**.

*   **Distance in Standard Errors:** The improvements represent shifts of **~243.8 SEs** for `aws --version` and **~379 SEs** for `aws sts get-caller-identity`. 
*   **P-Value:** In both tests, **$p < 0.001$**. This indicates that the probability of these improvements being a result of random chance is effectively zero.
*   **Precision:** With a sample size of $n=2000$, our margin of error is extremely small ($<0.3$ ms). The performance gains are systemic and statistically "locked-in."

##### **Average execution times compared**

graph (4)

Significant time savings are shown. We can expect similar improvements across all modeled commands, and _some_ improvement(s) to customization commands.

##### **Probability density functions of execution time**
graph (2)

aws --version execution time probability.

graph (3)

aws sts get-caller-identity execution time probability.

The probability density curves in both graphs show no overlap between the "Before" and "After" states. The narrow, tall peaks confirm that the large sample size has provided a high-confidence estimate of the true performance improvement. For the version command, we note that the peak is taller and narrower in the "After" state, signaling that the patch not only improves execution time but also reduces noise.

---

**Summary:** The feature provides a consistent and significant performance reduction (11.70-13.35%) across different command types with statistical certainty.


hssyoo (changes requested):
The measurements in the PR description show a ~5ms improvement, which is far from the ~66ms mean import time we measured. Based on these numbers, it seems highly likely that `prompt_toolkit` is still being imported elsewhere.

I was able to confirm this by checking `sys.modules` at the end of a `aws --version` call.

aemous:
> The measurements in the PR description show a ~5ms improvement, which is far from the ~66ms mean import time we measured. Based on these numbers, it seems highly likely that `prompt_toolkit` is still being imported elsewhere.
> 
> I was able to confirm this by checking `sys.modules` at the end of a `aws --version` call.

Great catch! I have analyzed `sys.modules` repeatedly to find where to switch eager imports to lazy imports across various files, the resulting `sys.modules` show no more `prompt_toolkit` for unrelated commands, when autoprompt is off.

.changes/next-release/enhancement-Performance-97002.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Performance",
+  "description": "Refactor imports to reduce command initialization time (e.g. loading all imported modules), resulting in about 11.70-13.35% reduced execution time across most commands (modeled commands and ``aws --version``)."
```
hssyoo: 1. We're only deferring `prompt_toolkit` import here, right? Let's be specific about that instead of generally claiming "refactor imports".
2. I wouldn't put any concrete numbers in the changelog. This figure changes between hardware and command run. Just say it avoids `prompt_toolkit` import time for commands that don't need it.

hssyoo (changes requested):
It's looking better but it looks like there's some gaps:
1. I noticed that `aws configure list` still imports `prompt_toolkit` even though it doesn't need it.
2. I'm not seeing anything in the PR that will guard against someone from adding unconditional `prompt_toolkit` imports in the future.

I'd like to see us step back and first enumerate all cases where we know we'll need to import `prompt_toolkit`. Knowing all cases upfront will help us verify that the implementation behaves as intended and help us write tests to ensure future changes don't reintroduce unconditional imports.

I'm also not sure if scattering import statements across different files and within methods is the right approach. I'm wondering if what we want instead is an authoritative source for importing `prompt_toolkit`. For example, say we had 2 modules:
1. `_prompt.py` - The only module in the AWS CLI repository that's allowed to import anything from `prompt_toolkit`.
2. `prompt.py` - A module that exposes factory functions that return objects that need require `prompt_toolkit`.

The imports from `_prompt.py` would happen within the factory functions. This is nice because then the rest of the AWS CLI codebase can import from `prompt.py`, but `prompt_toolkit` won't be imported until the factory function is actually called.

A single entrypoint for importing `prompt_toolkit` also enables us to write linting rules that assert that `prompt_toolkit` is never imported outside of `_prompt.py`, providing an extra layer of protection.

Thoughts?

aemous:
> It's looking better but it looks like there's some gaps:
> 
> 1. I noticed that `aws configure list` still imports `prompt_toolkit` even though it doesn't need it.
> 2. I'm not seeing anything in the PR that will guard against someone from adding unconditional `prompt_toolkit` imports in the future.
> 
> I'd like to see us step back and first enumerate all cases where we know we'll need to import `prompt_toolkit`. Knowing all cases upfront will help us verify that the implementation behaves as intended and help us write tests to ensure future changes don't reintroduce unconditional imports.
> 
> I'm also not sure if scattering import statements across different files and within methods is the right approach. I'm wondering if what we want instead is an authoritative source for importing `prompt_toolkit`. For example, say we had 2 modules:
> 
> 1. `_prompt.py` - The only module in the AWS CLI repository that's allowed to import anything from `prompt_toolkit`.
> 2. `prompt.py` - A module that exposes factory functions that return objects that need require `prompt_toolkit`.
> 
> The imports from `_prompt.py` would happen within the factory functions. This is nice because then the rest of the AWS CLI codebase can import from `prompt.py`, but `prompt_toolkit` won't be imported until the factory function is actually called.
> 
> A single entrypoint for importing `prompt_toolkit` also enables us to write linting rules that assert that `prompt_toolkit` is never imported outside of `_prompt.py`, providing an extra layer of protection.
> 
> Thoughts?

## Configure list

Regarding `aws configure list`, good catch. I have completed an audit of the codebase for prompt_toolkit usage, and root-caused it to the following import chain:

```
configure.py imports sso_commands.py (module-level)
  → ConfigureSSOCommand.__init__ imports sso.py (lazy, but fires during subcommand table build)
    → sso.py has module-level prompt_toolkit imports
```
In summary, every `aws configure` command will import `prompt_toolkit` for this exact reason. It should be easy to apply a patch to fix this so that the module is only imported for the 2 SSO commands that use it (`aws configure sso` and `aws configure sso-session`). Sounds like we should only proceed with this patch _if_ we decide against your proposal in part 2.

## Single point of importing prompt_toolkit (with linting enforcement)

### prompt_toolkit audit results

Below is the audit results for when prompt_toolkit is definitely needed:

1. `aws configure sso` / `aws configure sso-session`

**Always needed** — these are inherently interactive commands that always prompt the user

**Current bug (as you pointed out)**: imported during __init__ of BaseSSOConfigurationCommand, which fires during subcommand table build for any `aws configure` subcommand (including list, get, set, etc.)

**Fix**: move imports from __init__ to _run_main

2. `aws <service> wizard` (iam, lambda, dynamodb, events, configure)
**Always needed** — wizards are inherently interactive TUIs

**Currently correct**: `prompt_toolkit` is deferred to _get_runner() → _run_wizard(). Help and subcommand table build don't trigger it.

3. `aws logs start-live-tail`
**Always needed** — it's a TUI command

**Current bug**: `startlivetail.py` has module-level `prompt_toolkit` imports, and it's imported during `building-command-table.logs` to inject the command. So every `aws logs` command (tail, describe-log-groups, etc.) pays the `prompt_toolkit` cost.

**Fix**: defer `prompt_toolkit` imports in `startlivetail.py` to method level, or restructure so the command class can be instantiated without importing `prompt_toolkit`

4. `aws ecs monitor-express-gateway-service` (with `--mode INTERACTIVE`)

**Conditionally needed** — only when `display_mode == 'INTERACTIVE'` (default when TTY is present, explicit via `--mode`)

**Currently correct**: `prompt_toolkit_display.py` is lazily imported inside the display strategy factory method. Other `aws ecs` commands are clean.

5. Autoprompt mode (`--cli-auto-prompt` / config `cli_auto_prompt`)

**Conditionally needed** — only when autoprompt mode resolves to 'on' or 'on-partial'

**Currently correct**: `AutoPromptDriver` is lazily imported inside `_do_main` only when the mode is active.

6. `aws login` (when no region is configured)
**Conditionally needed** — only when `_resolve_region` falls through to `_prompt_for_region` (no `--region` flag, no region in config/profile)

**Currently correct**: lazy import inside `_prompt_for_region`.


**In summary**, there are two places on this current branch that import prompt_toolkit when it is not needed. One is all of the `aws configure` commands (as you pointed out), and the other is all `aws logs` commands.

### Thoughts on your factory-with-enforcement proposal

> I'm also not sure if scattering import statements across different files and within methods is the right approach. I'm wondering if what we want instead is an authoritative source for importing `prompt_toolkit`. For example, say we had 2 modules:

While I see your point, I don't see how the proposed factory approach solves the concern. So instead of importing prompt_toolkit, devs will need to call the factory functions that import prompt_toolkit. But the devs still have to make sure they call the factory function an optimal place. For example, a dev can just as easily call the factory function in the ConfigureCommand's __init__ function, so it seems to replace the problem of where to scatter the imports with where to scatter the factory function calls.

> A single entrypoint for importing `prompt_toolkit` also enables us to write linting rules that assert that `prompt_toolkit` is never imported outside of `_prompt.py`, providing an extra layer of protection.

I don't understand why forcing prompt_toolkit to be imported by a single module is a good restriction to have, since, as stated in the previous paragraph, the factory function calls still need to be placed well to avoid unnecessarily importing prompt_toolkit.


___


A lint rule saying "only these files may import prompt_toolkit" is just an allowlist — it prevents new files from adding prompt_toolkit imports, but it doesn't prevent the actual problem (those allowed files being imported too early).

The rule we actually want to enforce seems more like "prompt_toolkit must not be in sys.modules after create_clidriver() + driver.main() for commands that don't need it." That seems more of a runtime property, not a static lint check. We can't determine it from source analysis alone because it depends on which code paths execute.

What are your thoughts on adding a test for a subset of commands that asserts that `prompt_toolkit` is not in `sys.modules` while executing these commands? It would prevent future regression of the module sneaking back in over time:

```python
def test_no_prompt_toolkit_imported_for_basic_commands():
    for cmd in ['s3 ls', 'configure list', 'logs describe-log-groups', 'ecs describe-services']:
        # run in subprocess, assert prompt_toolkit not in sys.modules
```

.changes/next-release/enhancement-Performance-97002.json
aemous: 1. Technically there is more than just prompt_toolkit being deferred, though prompt_toolkit is likely the heftiest dependency we deferred. For example, `from awscli.customizations.logs.tail import TailCommand` was deferred until an `aws logs` command is being executed.
2. Good point, I will revise the changelog.

hssyoo: Can we just be realistic here about what we're trying to do? The purpose of this effort is to avoid eagerly importing `prompt_toolkit` based on our benchmark results. What are we saving by avoiding `TailCommand`? Are we making this change because it warrants its own optimization or because we're already moving `StartLiveTailCommand` so we might as well also move `TailCommand`?

hssyoo:
@aemous Don't want to quote the whole thread but to respond:
- `prompt_toolkit` usage audit looks good to me.
- Design approach - This was more of a stylistic suggestion. Mainly, I wanted someone to be able to import at the top-level without actually importing `prompt_toolkit`. I ultimately agree that a functional test that asserts the absence of `prompt_toolkit` in `sys.modules` (or a test that achieves something conceptually similar) is good enough for me. Feel free to leave it as-is.

.changes/next-release/enhancement-Performance-97002.json
aemous: Right, the main thing we are doing here is avoiding import of `prompt_toolkit`. And yes, I also chose to defer `TailCommand` because it is not needed unless we are running `aws logs` and it is consistent with `StartLiveTailCommand`. It also future-proofs things so that if `TailCommand` adds heavier imports in the future it will already be deferred until `aws logs` is run.

Is the updated changelog sufficient?

awscli/autoprompt/core.py
hssyoo: This looks like dead code now, let's remove if that's the case.

awscli/autoprompt/exceptions.py
hssyoo, line 1: Missing a license header.

awscli/customizations/configure/sso_commands.py
```diff
@@ -0,0 +1,491 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""
+Defines SSO-specific configure commands, to be used with the `aws configure`
+command.
+
+The main reason it lives in its own separate module instead of in
+`awscli/customizations/sso.py` is so that these commands can be referenced
+without importing `awscli/customizations/sso.py`, which imports
```
hssyoo: typo: `awscli/customizations/sso.py` -> `awscli/customizations/configure/sso.py`

awscli/customizations/configure/sso_commands.py
```diff
@@ -0,0 +1,491 @@
...
+from `prompt_toolkit`. Importing from `prompt_toolkit` has historically
+increased command execution times.
+
+This separation helps us limit our imports from `prompt_toolit` to when it
```
hssyoo: typo: `prompt_toolit` -> `prompt_toolkit`

awscli/clidriver.py
```diff
@@ -126,6 +131,36 @@ def create_clidriver(args=None):
...
+    if cli_auto_prompt and no_cli_auto_prompt:
+        raise ParamValidationError(
+            'Both --cli-auto-prompt and --no-cli-auto-prompt cannot be '
+            'specified at the same time.'
+        )
+
+
+def resolve_auto_prompt_mode(args, session):
+    # Order of precedence to check:
+    # - check if any arg rom NO_PROMPT_ARGS in args
```
hssyoo: typo: `rom` -> `from`, `NO_PROMPT_ARGS` -> `_NO_AUTO_PROMPT_ARGS`

awscli/customizations/logs/ui.py
```diff
@@ -0,0 +1,804 @@
...
+        self._create_ui(app_output, app_input)
+
+    def _create_ui(self, app_output, app_input):
+        from prompt_toolkit.application import Application
+        from prompt_toolkit.layout import Layout, Window
+        from prompt_toolkit.layout.containers import HSplit, VSplit
+        from prompt_toolkit.layout.controls import (
+            BufferControl,
+            FormattedTextControl
+        )
```
hssyoo: `prompt_toolkit` is already imported in this module at the top-level. Do we need to lazily import at all in this module? Especially since the point of this new module is to separate UI components.

tests/unit/customizations/logs/test_startlivetail.py
```diff
@@ -1,40 +1,27 @@
...
-# Licensed under the Apache License, Version 2.0 (the "License"). You
-# may not use this file except in compliance with the License. A copy of
-# the License is located at
-#
-#     http://aws.amazon.com/apache2.0/
-#
-# or in the "license" file accompanying this file. This file is
-# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
-# ANY KIND, either express or implied. See the License for the specific
-# language governing permissions and limitations under the License.
```
hssyoo, line 12: Keep license header or update it.

awscli/customizations/logs/ui.py
```diff
@@ -0,0 +1,804 @@
...
+
+        self._ui.exit()
+
+    def stop(self):
+        if self._exception is not None:
+            self._output.write(str(self._exception) + "\n")
+            self._output.flush()
+
+    def run(self):
+        (self._collect_log_events())
```
hssyoo: nit: Remove outer parentheses.

awscli/customizations/configure/sso_commands.py
```diff
@@ -0,0 +1,491 @@
...
+        # Initialize self._sso_session_prompter to None. It will be
+        # initialized lazily during command execution.
+        self._sso_session_prompter = None
+
+    def _init_prompt_toolkit(self):
+        from awscli.customizations.configure.sso import (
+            SSOSessionConfigurationPrompter,
+        )
+        if self._prompter is None:
+            from awscli.customizations.configure.sso import PTKPrompt
```
hssyoo: I think these 2 imports can just be at the same level outside of the control flow.

tests/functional/dependencies/test_lazy_imports.py
```diff
@@ -0,0 +1,62 @@
...
+
+# Commands that should never trigger a prompt_toolkit import.
+BASIC_COMMANDS = [
+    's3 ls',
+    'configure list',
+    'logs describe-log-groups',
+    'ecs describe-services',
+]
+
+_CHECK_SCRIPT = """\
```
hssyoo, line 26: It looks like this script is making actual AWS requests. We should stub the clients and avoid any unnecessary network traffic.

hssyoo (changes requested):
Looks good, I mostly had formatting nits. Also just to double-check, are we running pre-commit hooks on these commits? If it still needs to be done, I'd appreciate it as a separate commit.

.changes/next-release/enhancement-Performance-97002.json
hssyoo: Looks fine

awscli/customizations/logs/ui.py
aemous: Good point, I wanted to lazily import where possible just to be cautious, but none of this is imported until the correct commands are being run anyways, so no need to be that cautious here.

awscli/clidriver.py
aemous: Looks like I copied those typos from the original state; I have fixed them though, thanks!

hssyoo (dismissed):
🏆

hssyoo (approved):

PR #10236, opened 2026-04-21
Add @validates_models to test_verify_generated_completions_are_valid

SamRemis:
*Description of changes:*
This PR adds `@validates_models` annotation to a test which validates the completion models.

I've also edited another test which already has the annotation to catch cases that could otherwise be missed in the completions test.  Any operations that don't exist in the service model but are referenced in the completion file's `resources` section would have been missed by the existing `test_completions_operations_exist_in_model` test prior to this PR.  


AndrewAsseily (approved):
Thanks!! 🚢

PR #10239, opened 2026-04-22
Add retries to test fail mb rb

SamRemis:
*Description of changes:*
During a test where we attempt to make a bucket that already exists, S3 will occasionally raise `OperationAborted` exceptions instead of the expected `BucketAlreadyExists`.  

This does not get retried as it is a 409 on an unmodeled exception.  This PR adds up to four attempts to get the expected error before failing, using an exponential backoff

### DIFF
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -1377,7 +1377,7 @@ class TestMbRb(BaseS3IntegrationTest):
         for i in range(4):
             # Choose a bucket name that already exists.
             p = aws('s3 mb s3://mybucket')
-            if "BucketAlreadyExists" in p.stderr:
+            if "OperationAborted" not in p.stderr:
                 break
             time.sleep(2**i)
         assert "BucketAlreadyExists" in p.stderr
--- a/tests/integration/customizations/s3/test_plugin.py
+++ b/tests/integration/customizations/s3/test_plugin.py
@@ -1372,8 +1372,8 @@ class TestMbRb(BaseS3IntegrationTest):
         self.assert_no_errors(p)
 
     def test_fail_mb_rb(self):
-        #S3 can intermittenly return an `OperationAborted` exception instead of
-        # `BucketAlreadyExists`, so we give this test three attempts
+        # S3 can intermittently return an `OperationAborted` exception instead of
+        # `BucketAlreadyExists`, so we give this test four attempts
         for i in range(4):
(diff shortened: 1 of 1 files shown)


tests/integration/customizations/s3/test_plugin.py
```diff
@@ -1372,8 +1372,14 @@ def test_mb_rb(self, s3_utils):
         self.assert_no_errors(p)
 
     def test_fail_mb_rb(self):
-        # Choose a bucket name that already exists.
-        p = aws('s3 mb s3://mybucket')
+        #S3 can intermittenly return an `OperationAborted` exception instead of
+        # `BucketAlreadyExists`, so we give this test three attempts
```
jonathan343:
```suggestion
        # S3 can intermittently return an `OperationAborted` exception instead of
        # `BucketAlreadyExists`, so we give this test four attempts
```

jonathan343 (dismissed):

tests/integration/customizations/s3/test_plugin.py
```diff
@@ -1372,8 +1372,14 @@ def test_mb_rb(self, s3_utils):
...
     def test_fail_mb_rb(self):
-        # Choose a bucket name that already exists.
-        p = aws('s3 mb s3://mybucket')
+        # S3 can intermittently return an `OperationAborted` exception instead of
+        # `BucketAlreadyExists`, so we give this test four attempts
+        for i in range(4):
+            # Choose a bucket name that already exists.
+            p = aws('s3 mb s3://mybucket')
+            if "BucketAlreadyExists" in p.stderr:
+                break
```
hssyoo: Instead of looking for `BucketAlreadyExists` and retrying otherwise, I'd prefer if we only retried on `OperationAborted`. Mainly, I want to avoid swallowing other error codes that may point to genuine regressions from the server or client.

hssyoo (changes requested):

hssyoo (approved):

tests/integration/customizations/s3/test_plugin.py
SamRemis: That's a good point; I've updated the PR with this change.

Just for the PR history in case anyone is curious what scenario an error would get swallowed - if S3 returns a new error a very small percentage of times (say 0.5%), this test would have previously caught it and alerted us.

With this change, we otherwise would have almost certainly swallowed the error and never found it because it such an error would be very unlikely to fail four times in a row and cause this test to fail.

PR #10261, opened 2026-04-29
[v2] Test that example filename matches existing service and operation

kdaily:
*Issue #, if available:*

NA, discovered case while manually reviewing a doc examples PR.

*Description of changes:*

If an RST file exists and doesn't match the name of a service and operation, the examples will not be shown in the documentation.

Verify that the RST filename matches a service and command. It should be at the path `{service}/{command}.rst`.


kdaily:
Moving to a draft while I fix all of the current issues with example file names.

PR #10277, opened 2026-05-05
Fix InstanceMetadataRegionFetcher for local zones

halaney:
InstanceMetadataRegionFetcher queries latest/meta-data/placement/availability-zone/ and strips the trailing character to derive the region. This works for standard AZs (us-east-1a becomes us-east-1) but breaks for local zones (us-east-1-atl-2a becomes us-east-1-atl-2).

Switch to latest/meta-data/placement/region, which returns the parent region directly. This has been around since[0] 2020 so it _should_ be ok to use.

[0]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-categories

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/utils.py
+++ b/awscli/utils.py
@@ -163,7 +163,7 @@ class IMDSRegionProvider(BaseProvider):
 
 
 class InstanceMetadataRegionFetcher(IMDSFetcher):
-    _URL_PATH = 'latest/meta-data/placement/availability-zone/'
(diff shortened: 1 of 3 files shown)


halaney:
related: https://github.com/boto/botocore/pull/3699

halaney:
Hey! Any chance I could get some review on this?

PR #10293, opened 2026-05-08
Improve S3 performance for listing objects in transfer tasks

jamesls:
This improves the rate at which we can list objects for S3 transfer tasks such as recursive download, sync, and s3 to s3 copies.  In high compute environments this has become one of the main bottlenecks affecting the transfer of a large number of objects, particularly when using the CRT transfer client. We aren't able to queue work fast enough.  To speed things up I added three changes.

The first is an improvement in parsing the `ListObjectsV2` response.  We were previously double-parsing the `LastModified` member, which is mostly a historical artifact of when the CLI had differing behavior for parsing timestamps than botocore.  As a result of this custom parsing being left in place in the bucket lister we were parsing the timestamps twice.  To minimize the scope of changes, we keep the existing local-timezone datetime parsing in the bucket lister, but we set the botocore parser used in the bucket lister client to be a noop.  This does make the code slightly more complicated as we only plumb through this behavior for bucket lister so we need new client factory methods for that, so we should decide if it's worth trying to make this behavior the default for all of the S3 client creation used in the CLI.

The remaining changes are related to moving the bucket listing off of the main thread and over to a producer/consumer model, with the main thread now pulling objects off of a shared queue.

The producer thread is further broken down into this "quick page" feature where alternating threads are used to retrieve subsequent pages with an SAX based XML parser being used to do a first pass scan to extract the `NextContinuationToken`.  This allows the network IO work to continue as soon as possible while botocore finishes the standard XML parsing of the response body, and the subsequent "page drain" of processing the S3 key names and queueing files over to the CRT layer.

As for rollout, I've added a new `bucket_lister` config option under S3, with the default being the existing single threaded behavior. Users can opt-in via:

```
s3 =
    bucket_lister = threaded
```

The idea would be that this will flip to the default behavior after some period of bake time.

### DIFF
--- a/awscli/customizations/s3/bucketlister.py
+++ b/awscli/customizations/s3/bucketlister.py
@@ -1,0 +1,406 @@
+# Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import io
+import logging
+import threading
+import xml.sax.handler
+from typing import Optional
+from collections import namedtuple
+from dataclasses import dataclass
+from xml.sax.handler import ContentHandler
+
+from dateutil.parser import parse
+from dateutil.tz import tzlocal
+
+from awscli.compat import queue
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+_THREADED_BUCKET_LISTER_COMPLETE = object()
+_ThreadedBucketListerError = namedtuple(
+    '_ThreadedBucketListerError', ['exception']
+)
+_ThreadedBucketPage = namedtuple('_ThreadedBucketPage', ['contents'])
+
+
+def _date_parser(date_string, tzinfo):
+    return parse(date_string).astimezone(tzinfo)
+
+
+class _StopParsing(Exception):
+    pass
+
+
+class BucketLister:
+    """List keys in a bucket."""
+
+    def __init__(self, client, date_parser=_date_parser):
+        self._client = client
+        self._date_parser = date_parser
+        self._local_tz = tzlocal()
+
+    def _get_list_objects_kwargs(
+        self, bucket, prefix=None, page_size=None, extra_args=None
+    ):
+        kwargs = {
+            'Bucket': bucket,
+            'PaginationConfig': {'PageSize': page_size},
+        }
+        if prefix is not None:
+            kwargs['Prefix'] = prefix
+        if extra_args is not None:
+            kwargs.update(extra_args)
+        return kwargs
+
+    def _get_list_objects_v2_request_kwargs(
+        self, bucket, prefix=None, page_size=None, extra_args=None
+    ):
+        kwargs = {'Bucket': bucket}
+        if prefix is not None:
+            kwargs['Prefix'] = prefix
+        if page_size is not None:
+            kwargs['MaxKeys'] = page_size
+        if extra_args is not None:
+            kwargs.update(extra_args)
+        return kwargs
+
+    def _yield_page_contents(self, bucket, contents):
+        if not contents:
+            return
+        local_tz = self._local_tz
+        for content in contents:
+            source_path = bucket + '/' + content['Key']
+            content['LastModified'] = self._date_parser(
+                content['LastModified'],
+                local_tz,
+            )
+            yield source_path, content
+
+    def list_objects(
+        self, bucket, prefix=None, page_size=None, extra_args=None
+    ):
+        kwargs = self._get_list_objects_kwargs(
+            bucket=bucket,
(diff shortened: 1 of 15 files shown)


awscli/topics/s3-config.rst
```diff
@@ -239,6 +243,20 @@ In cases where network IO is the bottleneck, it is recommended to configure
 ``max_concurrent_requests`` instead.
 
 
+bucket_lister
+-------------
+
+**Default** - ``standard``
+
+Determines the bucket listing implementation to use when the AWS CLI discovers
+S3 objects for transfer commands. Valid choices are:
```
hssyoo, line 252: It's not immediately obvious to users why they may want to configure this option. Should we explicitly mention that threaded may speed up commands that download/copy many objects?

awscli/customizations/s3/filegenerator.py
```diff
@@ -134,6 +134,7 @@ class FileGenerator:
     under the same common prefix.  The generator yields corresponding
     ``FileInfo`` objects to send to a ``Comparator`` or ``S3Handler``.
     """
+    _DEFAULT_BUCKET_LISTER_CLS = ThreadedBucketLister
```
hssyoo: Shouldn't this be `BucketLister`?

awscli/customizations/s3/bucketlister.py
```diff
@@ -0,0 +1,406 @@
...
+        self, bucket, prefix=None, page_size=None, extra_args=None
+    ):
+        kwargs = {'Bucket': bucket}
+        if prefix is not None:
+            kwargs['Prefix'] = prefix
+        if page_size is not None:
+            kwargs['MaxKeys'] = page_size
+        if extra_args is not None:
+            kwargs.update(extra_args)
+        return kwargs
```
hssyoo: nit: The names imply that `BucketLister` uses `ListObjects` instead of `ListObjectsV2`, which isn't the case. Should the names just reflect the fact that 1 is for standard bucket lister and the other for threaded? Or let `ThreadedBucketLister` override this method?

awscli/customizations/s3/bucketlister.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def _yield_page_contents(self, bucket, contents):
+        if not contents:
+            return
+        local_tz = self._local_tz
```
hssyoo: nit: Is this line necessary?

awscli/customizations/s3/bucketlister.py
```diff
@@ -0,0 +1,406 @@
...
+    def __init__(self, client, stop_event):
+        self._client = client
+        self._stop_event = stop_event
+        self._task_queues = [queue.Queue(), queue.Queue()]
+        self._complete_page_queue = queue.Queue(maxsize=self._MAX_PAGES_BUFFER)
+        self._shutdown_triggered = False
+        self._thread_local = threading.local()
+        self._before_parse_unique_id = (
+            'awscli-threaded-bucket-lister-prefetch-before-parse-%s'
+            % id(self)
```
hssyoo: nit: Use f-string

awscli/customizations/s3/bucketlister.py
```diff
@@ -0,0 +1,406 @@
...
+            task_queue=self.next_task_queue,
+            next_task_queue=self.task_queue,
+            quick_page_scheduled=False,
+        )
+
+
+@dataclass
+class ListObjectsV2PageResponse:
+    page_number: int
+    page_response: dict
```
hssyoo: Should this be `Optional[dict]` if we set it to `None` in the case of exceptions? LSP doesn't like it

hssyoo (commented):
Thanks for the contribution, diff's looking good. Need to play around with it a bit but in the meantime had some small comments.

Can we also add a changelog entry?

awscli/customizations/s3/bucketlister.py
jamesls: removed.

awscli/customizations/s3/bucketlister.py
jamesls: Renamed the method to make it clear that one is expected to return params for the botocore paginator and one is for the direct list objects v2 request.  Also moved `_get_list_objects_v2_request_kwargs` down to the `ThreadedBucketLister` since it's only used in that class.

awscli/topics/s3-config.rst
jamesls: Updated.  After reading the explanations though, it still didn't seem clear the difference between `standard` and `threaded` so I renamed `standard` -> `single`.  Reads more clear to me now to see `single` vs. `threaded`, but let me know what you think.

jamesls:
> Can we also add a changelog entry?

Changelog added.

> Thanks for the contribution, diff's looking good. Need to play around with it a bit but in the meantime had some small comments.

One thing I've been using to see only the effects of the bucket lister improvements is to skip the actual transfers and use `--dryrun`, e.g.:

```
$ time aws s3 cp --recursive s3://mybucket/foo/ /tmp/ --dryrun --quiet

real    0m16.369s
user    0m4.287s
sys     0m2.366s

$ aws configure set s3.bucket_lister threaded

$ time aws s3 cp --recursive s3://mybucket/foo/ /tmp/ --dryrun --quiet

real    0m9.999s
user    0m3.692s
sys     0m1.839s
```

hssyoo (approved):
🏆 I saw comparable improvements with ~20k objects on my Mac. Also ran an internal build against this branch to verify builds and integration tests pass.

PR #10297, opened 2026-05-11
Avoid passing ECS session token in plugin argv

NguyenCong2k:
## Summary

- pass the ECS Exec session response to `session-manager-plugin` through `AWS_SSM_START_SESSION_RESPONSE` when the installed plugin supports it
- keep the existing argv fallback for older plugin versions
- add unit coverage for the environment-variable path

## Test

```powershell
python -m pytest tests\unit\customizations\test_sessionmanager.py tests\unit\customizations\ecs\test_executecommand_startsession.py -q

### DIFF
--- a/awscli/customizations/ecs/executecommand.py
+++ b/awscli/customizations/ecs/executecommand.py
@@ -13,10 +13,12 @@
 import logging
 import json
 import errno
+import os
 
-from subprocess import check_call
+from subprocess import check_call, check_output
 from awscli.compat import ignore_user_entered_signals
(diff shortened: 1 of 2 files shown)


NguyenCong2k:
Thanks, addressed both comments.

- Removed the redundant initial `session-manager-plugin` `check_call`; the `--version` call now serves as the plugin existence check.
- Moved the failing-path argv assertion outside the `assertRaises` block so it executes.

Tests:
```powershell
python -m pytest tests\unit\customizations\test_sessionmanager.py tests\unit\customizations\ecs\test_executecommand_startsession.py -q

PR #10299, opened 2026-05-11
[v2] Implement LazyCommand for deferring modules until their commands are invoked, and make PrefixTrie public

aemous:
*Issue #, if available:*

* Contributes to #9011.

*Notes:*

* This PR is part of an ongoing effort to reduce AWS CLI initialization time by lazily deferring plugins until they are needed. This PR makes an existing PrefixTrie implementation public, which will be used in a future PR to efficiently model a plugin event-dependency registry. Also, it implements `LazyCommand`, which will be used in a future PR to automatically defer plugin imports for plugins that add custom commands to AWS CLI.

*Description of changes:*

* Switched PrefixTrie from a private interface to a public interface (i.e. rename `_PrefixTrie` to `PrefixTrie`).
* Implemented unit tests for the `PrefixTrie` class.
* Implemented new` LazyCommand` class, which can wrap `CLICommand` instances and defer importing or invoking them until the command is called.
* Implemented unit and functional tests for the `LazyCommand` class.

*Description of tests:*

* Successfully ran all test suites and pre-production build workflow (see GitHub CI).

### DIFF
--- a/tests/unit/botocore/test_hooks.py
+++ b/tests/unit/botocore/test_hooks.py
@@ -14,7 +14,11 @@ import copy
 import functools
 from functools import partial
 
-from botocore.hooks import HierarchicalEmitter, first_non_none_response
+from botocore.hooks import (
+    HierarchicalEmitter,
+    PrefixTrie,
+    first_non_none_response,
+)
 
 from tests import unittest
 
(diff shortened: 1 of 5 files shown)


tests/unit/botocore/test_hooks.py
```diff
@@ -584,5 +588,53 @@ def handler(a, b, **kwargs):
         )
 
 
+class TestPrefixTrie(unittest.TestCase):
```
hssyoo: Use pytest for net new test cases

hssyoo (approved):
Small nit that I don't mind addressing in a followup.

tests/unit/botocore/test_hooks.py
aemous: Fixed in latest revision.

hssyoo (approved):

PR #10304, opened 2026-05-12
Add MD5 to checksum-algorithm choices for s3 commands

### LINKED ISSUE #10295: Unsupported checksum algorithm: md5
### Describe the bug

It appears that, with the new checksum algorithms that have been introduced to the S3 APIs (SHA512, MD5, XXHASH64, XXHASH3, XXHASH128), the cli binary does not correctly identify MD5. Doing a call with a parameter like --checksum-algorithm MD5 will result in the error: Unsupported checksum algorithm: md5

Theoretically this should be true for any s3 call with the checksum parameter.

### Regression Issue

### Expected Behavior

AWS CLI creates proper checksum header for appropriate actions and posts correctly. Does not refuse md5 checksum requests.

### Current Behavior

Unsupported checksum algorithm: md5

### Reproduction Steps

aws s3api put-bucket-policy --bucket <your-bucket> --policy <your-policy> --checksum-algorithm MD5

### Possible Solution

Is this on the roadmap but not implemented yet? It's currently live in the docs: https://docs.aws.amazon.com/cli/latest/reference/s3api/put-bucket-policy.html

_No response_

### CLI version used

2.34.45

Ubuntu 22.04

shreyaskommuri:
## Summary

Fixes aws/aws-cli#10295. MD5 was missing from the `--checksum-algorithm` choices list for high-level `s3` commands (`cp`, `sync`, `mv`), causing the CLI to reject it with an invalid choice error before the request was even made.

**Changes:**
- Add `MD5` to the `CHECKSUM_ALGORITHM` choices in `awscli/customizations/s3/subcommands.py`
- Add unit test verifying MD5 is accepted as a valid upload checksum algorithm

## Related

A companion fix in botocore registers the `Md5Checksum` implementation so the checksum is computed correctly at the request layer: https://github.com/boto/botocore/pull/3703

Without that botocore fix, `s3api` commands would still fail for MD5 — both PRs are needed for a complete fix.

## Test plan

- [x] `test_validate_checksum_algorithm_md5_upload` passes
- [x] All existing checksum parameter tests continue to pass

Developed with AI assistance, reviewed and tested by shreyaskommuri

### DIFF
--- a/awscli/customizations/s3/subcommands.py
+++ b/awscli/customizations/s3/subcommands.py
@@ -459,7 +459,7 @@ CHECKSUM_MODE = {
 }
 
 CHECKSUM_ALGORITHM = {
-        'name': 'checksum-algorithm', 'choices': ['CRC64NVME', 'CRC32', 'SHA256', 'SHA1', 'CRC32C', 'SHA512', 'XXHASH64', 'XXHASH3', 'XXHASH128'],
+        'name': 'checksum-algorithm', 'choices': ['CRC64NVME', 'CRC32', 'SHA256', 'SHA1', 'CRC32C', 'SHA512', 'XXHASH64', 'XXHASH3', 'XXHASH128', 'MD5'],
         'help_text': 'Indicates the algorithm used to create the checksum for the object.'
 }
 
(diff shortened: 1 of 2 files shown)


SamRemis:
Thank you again for the PRs.  

I've closed the other one, and will close this one for the same reason mentioned [here](https://github.com/boto/botocore/pull/3703#issuecomment-4445683896).

Please keep future discussions to https://github.com/aws/aws-cli/issues/10295.

PR #10305, opened 2026-05-13
[v2] Implement event handlers registry for built-in plugins

aemous:
*Issue #, if available:*

* Contributes to #9011.

*Notes:*

* This PR (2/4) is part of an ongoing effort to reduce AWS CLI initialization time by lazily deferring plugins until they are needed. This PR transforms plugin-initialization in `awscli/handlers.py` to an alternative registry format in `awscli/handlers_registry.py`, which will be used in a future PR to efficiently initialize plugins only when they are needed. The registry will be used to determine which plugins are needed during event-emission and command-execution. In a future PR, the monolithic `awscli/handlers.py` will be removed and replaced entirely by `awscli/handlers_registry.py`.

*Description of changes:*

* Implemented event handlers registry for built-in plugins in `awscli/handlers_registry.py` that maps events to plugins that register against them. It also declaratively models plugins that register against `building-command-table.main` to add or rename commands in the command table. The registry was automatically generated from a script that probes the existing initialization functions. The script can be reviewed here: [generate-plugin-registry.tar.gz](https://github.com/user-attachments/files/27730582/generate-plugin-registry.tar.gz).
* Wrapped plugins that get registered against events in `awscli/handlers.py` in a thin initialization function. The initialization function gets called in `awscli/handlers.py` instead of registering against the events in `awscli/handlers.py` directly.
  * While this change does not lead to a behavior change, it simplifies the migration of `awscli/handlers.py` to the new `awscli/handlers_registry.py`. Also, it is more consistent with the majority of existing plugin-initialization in `awscli/handlers.py`.
* For plugin-initialization functions that register against `building-command-table.main` to add or rename commands, that also register against at least 1 other event, these functions were split into two functions: one that registers the add/rename callbacks, and one that registers against everything else.
  * This change does not lead to a behavior change, but it simplifies the migration to the new handlers registry, and also paves a clear path forward for deleting the init-functions that solely add/rename commands in a future PR.
* Implemented test cases that assert all the modules in `awscli/handlers_registry.py` are importable, and misc. tests verifying expected behavior of new components.
* Updated existing s3 tests based on changes described above.

*Description of tests:*

* Successfully ran all test suites and the pre-production build workflow (see CI).

### DIFF
--- a/awscli/handlers_registry.py
+++ b/awscli/handlers_registry.py
@@ -20,6 +20,14 @@ event_handlers emitter), then proceeds with normal event dispatch.
 Entry format:
    (module, fn_name)    call fn(event_handlers)
 """
+import enum
+
+
+class CommandTableOp(enum.Enum):
+    """Valid operation types for MAIN_COMMAND_TABLE_OPS entries."""
+
+    ADD = 'add'
+    RENAME = 'rename'
 
 PLUGIN_REGISTRY = {
     'after-call.data-pipeline.GetPipelineDefinition': [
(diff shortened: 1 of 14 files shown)


tests/functional/test_handlers_registry.py
```diff
@@ -0,0 +1,208 @@
...
+                        f'{cb_name} modified .name on {new_key!r} '
+                        f'without renaming'
+                    )
+
+    assert not violations, (
+        'Callbacks registered against building-command-table.main must '
+        'only add or rename commands. The following callbacks perform '
+        'other modifications that would be lost when replaced by '
+        'MAIN_COMMAND_TABLE_OPS:\n' + '\n'.join(f'  - {v}' for v in violations)
+    )
```
aemous, line 212: Note for reviewer: these two tests add some invariants about plugins/functions that register against `building-command-table.main` that are needed for the initial migration from the monolith to the registry. In a future PR, we will remove the init functions that add/rename commands entirely, and we will delete these two tests.

These two tests are just asserting invariants of the *current state* that is needed for the script to work correctly.

awscli/handlers_registry.py
```diff
@@ -0,0 +1,855 @@
...
+# language governing permissions and limitations under the License.
+"""Maps event patterns to initializer entries.
+
+At runtime, the LazyInitEmitter triggers entries on demand:
+before emitting event X, it finds entries whose event patterns
+match X, calls each init function at most once (passing the
+event_handlers emitter), then proceeds with normal event dispatch.
+
+Entry format:
+   (module, fn_name)    call fn(event_handlers)
```
hssyoo, line 21: I'm having a hard time parsing this. I'm reading it as each entry being a (module, fn_name) tuple key with a callback value, which isn't what `PLUGIN_REGISTRY` actually looks like.

In general, do you think it's worth defining some lightweight types for this module instead of relying on comments? For example, having an enum that only has `add` and `rename` would make it clear that those are the only valid values for the first tuple item in a `MAIN_COMMAND_TABLE_OPS` element.

hssyoo (commented):
Looks good, just a minor comment/question

awscli/handlers_registry.py
aemous: > I'm having a hard time parsing this. I'm reading it as each entry being a (module, fn_name) tuple key with a callback value, which isn't what PLUGIN_REGISTRY actually looks like.

Ah, I see your point. `fn_name` is not the callback. It's the initialization function that we invoke. During invocation of the initialization function, one or more callback functions are registered against events.

So the actual callbacks are not in the registry directly. The registry contains functions that register the callbacks. Is this worth clarifying in comments, or does this clear it up?

> In general, do you think it's worth defining some lightweight types for this module instead of relying on comments? For example, having an enum that only has add and rename would make it clear that those are the only valid values for the first tuple item in a MAIN_COMMAND_TABLE_OPS element.

Thanks for the suggestion. I have pushed a revision to use an enum for `MAIN_COMMAND_TABLE_OPS`.

hssyoo: Gotcha, I'm okay with this

hssyoo (approved):

hssyoo (approved):

PR #10317, opened 2026-05-16
Use UUIDs for deploy ChangeSet names

### LINKED ISSUE #10315: cloudformation deploy: timestamp-only ChangeSet name guarantees collision under concurrent invocations
## Describe the bug

\`aws cloudformation deploy\` generates the ChangeSet name using only a Unix timestamp at 1-second precision:

\`\`\`python
# awscli/customizations/cloudformation/deployer.py:91-92
# Each changeset will get a unique name based on time
changeset_name = self.changeset_prefix + str(int(time.time()))
\`\`\`

Unlike the analogous issue in \`cloudfront\` (\`unique_string()\` — fixed in #10281) which had a 1-in-1,000,000 collision chance per second, this implementation has **no random component at all**. Two concurrent invocations within the same second generate **identical** ChangeSet names with 100% probability, causing the second to fail with \`AlreadyExistsException\`.

Related to #5435 (feature request to expose \`--change-set-name\`), which would serve as a workaround but does not address the root cause.

## Regression Issue

## Expected Behavior

\`aws cloudformation deploy\` always generates a unique ChangeSet name, regardless of how quickly or how many times it is invoked concurrently against the same stack.

## Current Behavior

When two invocations of \`aws cloudformation deploy\` target the same stack within the same second, both attempt to create a ChangeSet named (for example):

```
awscli-cloudformation-package-deploy-1747339200
```

The second invocation fails immediately:

```
An error occurred (AlreadyExistsException) when calling the CreateChangeSet
operation: ChangeSet [awscli-cloudformation-package-deploy-1747339200] already exists
```

The error is re-raised with no additional context, making it difficult to diagnose in CI/CD logs.

## Reproduction Steps

Run two simultaneous deploys against the same stack:

```bash
STACK=my-stack
TEMPLATE=template.yaml

aws cloudformation deploy --stack-name "$STACK" --template-file "$TEMPLATE" &
aws cloudformation deploy --stack-name "$STACK" --template-file "$TEMPLATE" &
wait
```

If both commands start within the same second (common in CI matrix jobs or parallel pipeline steps), one will fail with \`AlreadyExistsException\`.

## Possible Solution

Replace \`int(time.time())\` with \`uuid.uuid4()\`, which provides 122 bits of cryptographically random entropy and eliminates both the timestamp dependency and collision risk entirely. \`uuid\` is already part of the Python standard library.

\`\`\`python
import uuid

# Before
changeset_name = self.changeset_prefix + str(int(time.time()))

# After
changeset_name = self.changeset_prefix + str(uuid.uuid4())
\`\`\`

The resulting name (\`awscli-cloudformation-package-deploy-<uuid4>\`, ~73 chars) satisfies CloudFormation's ChangeSet naming constraint (alphanumeric + hyphens, max 128 chars). This is the same fix applied to \`cloudfront.unique_string()\` in #10281.

## Additional Information/Context

This affects CI/CD matrix builds and parallel pipeline stages that deploy the same stack concurrently. The comment on line 91 ("Each changeset will get a unique name based on time") reflects the intent but the implementation does not achieve uniqueness within the same second.

## CLI version used

aws-cli/2.x — reproducible across all versions; the timestamp-based naming has been present since the deploy command was introduced.

## Environment details (OS name and version, etc.)

Reproducible on any OS/platform (Linux, macOS, Windows).

rohan-patnaik:
## Summary

Fixes #10315.

`aws cloudformation deploy` was generating ChangeSet names from a one-second timestamp. If two deploy commands started in the same second for the same stack, both could try to create the same ChangeSet name and one would fail with `AlreadyExistsException`.

This changes the generated ChangeSet name to use `uuid.uuid4()` while keeping the existing AWS CLI prefix. That keeps the name format valid and avoids collisions between concurrent deploys.

## Testing

- `/tmp/awscli-10315-venv/bin/python -m pytest tests/unit/customizations/cloudformation/test_deployer.py`
- `/tmp/awscli-10315-venv/bin/python -m py_compile awscli/customizations/cloudformation/deployer.py tests/unit/customizations/cloudformation/test_deployer.py`

### DIFF
--- a/awscli/customizations/cloudformation/deployer.py
+++ b/awscli/customizations/cloudformation/deployer.py
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 
 import sys
-import time
+import uuid
 import logging
 import botocore
 import collections
@@ -26,6 +26,10 @@ LOG = logging.getLogger(__name__)
 
 ChangeSetResult = collections.namedtuple(
                 "ChangeSetResult", ["changeset_id", "changeset_type"])
+
+
+def _make_changeset_name(prefix):
+    return prefix + str(uuid.uuid4())
 
 
 class Deployer(object):
@@ -88,8 +92,7 @@ class Deployer(object):
         now = get_current_datetime().isoformat()
         description = "Created by AWS CLI at {0} UTC".format(now)
 
-        # Each changeset will get a unique name based on time
-        changeset_name = self.changeset_prefix + str(int(time.time()))
+        changeset_name = _make_changeset_name(self.changeset_prefix)
 
         if not self.has_stack(stack_name):
             changeset_type = "CREATE"
--- a/tests/unit/customizations/cloudformation/test_deployer.py
+++ b/tests/unit/customizations/cloudformation/test_deployer.py
@@ -2,7 +2,9 @@ import botocore.session
 
 from botocore.stub import Stubber
 from awscli.testutils import mock, unittest
-from awscli.customizations.cloudformation.deployer import Deployer, ChangeSetResult
+from awscli.customizations.cloudformation.deployer import (
+    ChangeSetResult, Deployer, _make_changeset_name
+)
 from awscli.customizations.cloudformation import exceptions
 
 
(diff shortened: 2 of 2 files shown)


rohan-patnaik:
@adev-code, would you have time to take a look when convenient? I’m happy to refresh the branch if useful. Thanks!

rohan-patnaik:
@adev-code, could you review this fix for the ChangeSet-name collision in #10315 when convenient? Please let me know if anything else is needed from me. Thanks!

PR #10327, opened 2026-05-19
Update README zsh completion for bundled installer

### LINKED ISSUE #8957: zsh completion script missing from installation file
I want to use the v2 CLI with zsh command completion. I tried to follow the [instructions](https://github.com/aws/aws-cli/tree/v2?tab=readme-ov-file#command-completion), but they seem incorrect.

The `aws_zsh_completer.sh` file does not exist in the downloaded zip file, nor in `/usr/local/aws-cli/v2/current/`.

How does one set up zsh completions?

adev-code:
Fixes #8957. The README's zsh instructions point to
`bin/aws_zsh_completer.sh`, which ships with Package Manager
installs (e.g. pip) but not with the bundled installer (zip/pkg).
Bundled-installer users hit "No such file or directory".

## Change

Scope the existing zsh block to Package Manager installs and add a
parallel block for the bundled installer using `bashcompinit` +
`complete -C aws_completer aws`.

## Reference

[Configuring command completion in the AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,17 +87,8 @@ For tcsh::
 
 You should add this to your startup scripts to enable it for future sessions.
 
-For zsh (Package Manager, e.g. pip installs) please refer to 
-``bin/aws_zsh_completer.sh``.  Source that file, e.g. from your 
-``~/.zshrc``, and make sure you run ``compinit`` before::
-
-    $ source bin/aws_zsh_completer.sh
-
-For now the bash compatibility auto completion (``bashcompinit``) is used.
-For further details please refer to the top of ``bin/aws_zsh_completer.sh``.
-
-For zsh (bundled installer, e.g. zip/pkg), use ``aws_completer``
-directly with ``bashcompinit``::
+For zsh, run ``bashcompinit`` and ``compinit``, then use the built-in
+``complete`` command::
 
     $ autoload bashcompinit && bashcompinit
     $ autoload -Uz compinit && compinit
--- a/README.rst
+++ b/README.rst
@@ -87,13 +87,26 @@ For tcsh::
 
 You should add this to your startup scripts to enable it for future sessions.
 
-For zsh please refer to ``bin/aws_zsh_completer.sh``.  Source that file, e.g.
-from your ``~/.zshrc``, and make sure you run ``compinit`` before::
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -87,14 +87,27 @@ For tcsh::
 
 You should add this to your startup scripts to enable it for future sessions.
 
-For zsh please refer to ``bin/aws_zsh_completer.sh``.  Source that file, e.g.
-from your ``~/.zshrc``, and make sure you run ``compinit`` before::
+For zsh (Package Manager, e.g. pip installs) please refer to 
+``bin/aws_zsh_completer.sh``.  Source that file, e.g. from your 
+``~/.zshrc``, and make sure you run ``compinit`` before::
```
hssyoo: Source builds should still produce an `aws_completer` exe:
```
$ pip install -e .
$ ls .venv/bin | rg "aws"

aws
aws_bash_completer
aws_completer
aws_zsh_completer.sh
aws.cmd
```

We should be good to just recommend the same set of instructions. Can you confirm?

adev-code: yes thank you, I confirmed it:
```
$ ls /tmp/pip-readme-test/bin/ | grep aws
aws
aws.cmd
aws_bash_completer
aws_completer
aws_zsh_completer.sh
```
Will update the PR to use one set of instructions.

hssyoo: Sorry, I meant can you confirm that the autocomplete functionality works when following the `aws_completer` instructions after building via pip?

adev-code: Confirmed. Autocomplete works after building via pip `pip install -e .`:

```
$ .venv/bin/pip install -e .
$ ls .venv/bin | grep aws
aws
aws.cmd
aws_bash_completer
aws_completer
aws_zsh_completer.sh
```

In zsh:\
```
$ autoload bashcompinit && bashcompinit
$ autoload -Uz compinit && compinit
$ complete -C "$HOME/aws-cli-pr-test/.venv/bin/aws_completer" aws
$ aws s<TAB>
s3  s3api  sns  ses  secretsmanager  sqs  sts  ...
```

hssyoo (approved):

PR #10339, opened 2026-05-21
Route awscrt logs through the logging module

azkrishpy:
*Issue #, if available:*

*Description of changes:*

  Route `awscrt` logs through Python's `logging` module so they're emitted alongside `botocore`, `awscli`, `s3transfer`, and `urllib3` when `--debug` is set. This makes CRT-backed
  features (e.g. CRT-based S3 transfers) easier to diagnose without configuring CRT logging separately.

  - `awscli/logger.py`: switch from `awscrt.io.init_logging` to `awscrt.logging.init_logging`; `enable/disable_crt_logging` now use Python log levels.
  - `awscli/clidriver.py`: add `awscrt` to the `--debug` logger list.
  - Existing logger tests updated to mock the new `awscrt.logging.set_log_level` target.
  
  Generated with AI assistance and reviewed by @azkrishpy 

### DIFF
--- a/.changes/next-release/enhancement-logging-43326.json
+++ b/.changes/next-release/enhancement-logging-43326.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "logging",
+  "description": "Add CRT logging integration so awscrt logs are routed through Python's logging module under the awscrt logger and emitted alongside botocore/awscli debug logs"
+}
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -626,7 +626,7 @@ class CLIDriver:
         self._set_logging(getattr(args, 'debug', False))
 
     def _set_logging(self, debug):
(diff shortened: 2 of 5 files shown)


hssyoo (changes requested):
Thanks Krish, had a few comments:
1. The new CRT debug logs seem to have a cleaner format, but the emitted logs got significantly more verbose. When uploading a 128MB object using the CRT client, the debug logs were previously ~300KB but now I'm seeing ~36MB. Can we look into that?
2. It looks like we have a failing test for `tests/functional/autoprompt/test_prompttoolkit.py::TestDebugPanel::test_debug_mode_does_not_allow_crt_logging` because it needs the log patch updated.
3. Can we also update this comment: https://github.com/aws/aws-cli/blob/97410b16938a06358d194ff2b5defbf47b275aba/awscli/autoprompt/prompttoolkit.py#L232-L238

PR #10340, opened 2026-05-22
[v2] feat: Add account and role filtering in aws configure sso

### LINKED ISSUE #8867: aws configure sso - filtering account list
### Describe the feature

When using `aws configure sso`, I'd like to be able to type to filter accounts. 

### Use Case

Some organizations have a large number of accounts and it's unusable (especially given that the accounts seem to be sorted by neither ID, or alias).

### Proposed Solution

It would be great to be able to type a few characters to filter the list of visible accounts.

### Other Information

_No response_

### Acknowledgements

### CLI version used

aws-cli/2.17.29

macOS 14.6.1

masamaru0513:
## Summary

Add keyboard filtering to the account and role selection menus in `aws configure sso`. When multiple accounts or roles are available, users can type to narrow the list down in real time.

Fixes #8867

## Changes

**`awscli/customizations/wizard/ui/selectmenu.py`**

- Add `FilterableSelectionMenuControl` (subclass of `SelectionMenuControl`). The search text is held in a `prompt_toolkit` `Buffer` rendered in its own window above the menu, so unicode input, bracketed pastes, cursor movement and the standard readline shortcuts (`Ctrl-U`, `Ctrl-W`, `Home`/`End`) are handled by `prompt_toolkit` rather than reimplemented here.
- Add `enable_filter` and `no_results_message` parameters to `select_menu()`. Menu navigation is bound at the application level because the search buffer, not the menu, holds the focus.
- Resolve callable `items` in `select_menu()`, and keep honoring callables in the filterable control so the inherited behavior of `SelectionMenuControl` is preserved.
- Fix `SelectionMenuControl.preferred_width()` raising `ValueError` on an empty item list, which is reachable once a filter matches nothing.
- Keep the previously highlighted item selected when the filter changes and that item still matches.

**`awscli/customizations/configure/sso_commands.py`**

- Enable filtering in `_handle_multiple_accounts()` and `_handle_multiple_roles()`, and mention the filter in both prompts.

**Tests**

- `tests/unit/customizations/wizard/ui/test_filterable_menu.py` (new): 49 tests. 32 unit tests cover filtering, selection handling, layout wiring and rendering. 17 end-to-end tests drive the real `Application` with key presses through a pipe input, covering typing, non-ASCII filter text, bracketed paste, backspace, `Ctrl-U`, arrow keys and recovery from a filter with no matches.
- `tests/unit/customizations/configure/test_sso.py`: record `select_menu` kwargs in `PTKStubber` and assert that both the account and role menus enable filtering.

## Notes

- The non-filtering code path is unchanged, so the rest of the wizard UI keeps its current behavior.
- Filtering is applied to role selection as well as account selection for consistency. Happy to drop the role part if you would rather keep this scoped to accounts.

## Testing

```
pytest tests/unit/customizations/wizard/ tests/unit/customizations/configure/   # 501 passed
```

`pre-commit` (ruff + ruff-format) is clean on the changed files.

Manually verified the interactive flow against a live IAM Identity Center instance on macOS (iTerm2, Terminal.app): typing to filter, non-ASCII filter text, pasting, backspace, `Ctrl-U`, arrow keys and `Enter` to select, plus the no-match message and recovery.

Also verified on Windows 10 (AMD64) with Python 3.10.11: the interactive filtering works correctly — typing to narrow the list, backspace, Ctrl-U, arrow keys, and Ctrl-V paste all behave as expected. All 49 filterable-menu tests and 80 SSO configuration tests pass.

Also verified on Linux via Docker container (`python:3.11-slim`, Debian 13, aarch64 and x86_64): the interactive flow against a live IAM Identity Center instance works correctly, including paste. All unit and E2E tests pass on both architectures.

## AI tooling disclosure

Generated by AI tools, and reviewed by masamaru0513, per the Automated Tools section of `CONTRIBUTING.rst`.

Related #9692

### DIFF
--- a/awscli/customizations/configure/sso_commands.py
+++ b/awscli/customizations/configure/sso_commands.py
@@ -208,11 +208,16 @@ class ConfigureSSOCommand(BaseSSOConfigurationCommand):
     def _handle_multiple_accounts(self, accounts):
         available_accounts_msg = (
             'There are {} AWS accounts available to you.\n'
+            'Use arrow keys to navigate, type to filter, '
+            'and press Enter to select.\n'
         )
         uni_print(available_accounts_msg.format(len(accounts)))
         sorted_accounts = sorted(accounts, key=get_account_sorting_key)
         selected_account = self._selector(
-            sorted_accounts, display_format=display_account
+            sorted_accounts,
+            display_format=display_account,
+            enable_filter=True,
+            no_results_message='No matching accounts found',
         )
         sso_account_id = selected_account['accountId']
         return sso_account_id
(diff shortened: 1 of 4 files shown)


masamaru0513:
Hi @ashovlin and @andrewasseily, this PR has been rebased onto the latest `v2` and is ready for code owner review. Would appreciate it if either of you could take a look when you have a chance.

PR #10342, opened 2026-05-22
[v2] Update standard retry mode via opt-in env var

aemous:
*Notes:*

* Ports https://github.com/boto/botocore/pull/3678 to vendored botocore. The main and only difference from upstream botocore is that AWS CLI v2 uses `standard` retry mode by default, but upstream botocore uses `legacy`.

*Description of changes:*

* Introduced new `AWS_NEW_RETRIES_2026` environment variable. When set to true, standard retry mode behavior is changed.
* The standard retry mode behavior changes (gated behind the new env var) is described below:
  * **Service-specific max attempts** — Service-defined (e.g., DynamoDB with 4 attempts) max attempt values instead of the universal default of 3, but only if the user hasn't explicitly configured max attempts.
  * **Throttling-aware backoff** — The ExponentialBackoff is initialized with a `throttling_detector` and `service_name`, enabling it to use different backoff scales depending on whether the error is a throttling error and which service is being called (e.g., 25ms base for DynamoDB non-throttling errors vs. 50ms for other services vs. 1s for throttling errors).
  * **Throttling-aware retry quota** — The RetryQuotaChecker receives a throttling_detector, allowing it to deduct a smaller quota cost (5) for throttling errors vs. the standard cost (14) for other retryable errors.
  * **Service name passed to RetryHandler** — For known long-polling operations (e.g., `SQS ReceiveMessage`, `SFN GetActivityTask`), when retry quota is exhausted the handler sleeps the backoff delay before returning without retrying.
* Added new test cases for the new standard retry mode behavior updates.

*Description of tests:*

* Successfully ran pre-production build workflow (see Build Internal GitHub Action).

### DIFF
--- a/.changes/next-release/enhancement-Retries-30198.json
+++ b/.changes/next-release/enhancement-Retries-30198.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Retries",
+  "description": "Introduced ``AWS_NEW_RETRIES_2026``, an opt-in environment variable (defaults to ``false``) that activates updated standard retry mode behavior. When set to ``true``, the standard retry mode uses lower initial backoff delays (50ms base instead of 1s for non-throttling errors), applies service-specific max attempt overrides, and honors the ``x-amz-retry-after`` response header for server-guided retry timing."
+}
--- a/awscli/botocore/configprovider.py
+++ b/awscli/botocore/configprovider.py
@@ -21,8 +21,25 @@ import os
 from botocore import utils
 from botocore.exceptions import InvalidConfigError
 
-logger = logging.getLogger(__name__)
+try:
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently it's only available to internal users for testing and validation.
+    # Any usage is not advised or supported in external code bases.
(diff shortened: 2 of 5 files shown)


kdaily:
Confirming my understanding:

> The main and only difference from upstream botocore is that AWS CLI v2 uses standard retry mode by default, but upstream botocore uses legacy.

Means we don't need this in CLI v2:

https://github.com/boto/botocore/pull/3678/changes#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R41

aemous:
> Confirming my understanding:
> 
> > The main and only difference from upstream botocore is that AWS CLI v2 uses standard retry mode by default, but upstream botocore uses legacy.
> 
> Means we don't need this in CLI v2:
> 
> https://github.com/boto/botocore/pull/3678/changes#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R41

Correct, that line and this other line that uses the variable are not needed:

https://github.com/boto/botocore/pull/3678/changes#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R171

kdaily (approved):

PR #10356, opened 2026-05-27
[v2] Implement lazy plugin initialization

aemous:
*Issue #, if available:*

* Contributes to https://github.com/aws/aws-cli/issues/9011.

*Notes:*

* This PR (4/4) is part of an ongoing effort to reduce AWS CLI initialization time by lazily deferring plugins until they are needed. This PR uses the constructs introduced in previous PRs to lazily import plugins when they are needed (event emission time, command invocation time, and/or help docs generation time).

*Description of changes:*

* Switch from using HierarchicalEmitter to LazyInitEmitter to facilitate lazy plugin loading.
* Delete `handlers.py` and load plugin registry from `plugin.py`.
* Update `exe/pyinstaller/hook-awscli.py`, to dynamically import builtin plugin modules from the plugin registry.
* Update a few functional/unit tests that previously used `register()` against events, to use `register_last()` instead. This is required for the tests to pass, because previously the tests would register after the builtin handlers in botocore, but since switching to lazy initialization, the test code would run before the boto handlers are executed.
* Add tests that improve test coverage on the lazy plugin initialization components.
* Add clarifying comments to `tests/unit/test_lazy_emitter.py`.
* Add changelog entry.


awscli/plugin.py
```diff
@@ -31,21 +31,21 @@ def load_plugins(plugin_mapping, event_hooks=None, include_builtins=True):
...
+    :rtype: LazyInitEmitter
     :return: An event emitter object.
 
     """
     if event_hooks is None:
-        event_hooks = HierarchicalEmitter()
+        event_hooks = LazyInitEmitter()
     if include_builtins:
-        _load_plugins(BUILTIN_PLUGINS, event_hooks)
+        event_hooks.load_registry(PLUGIN_REGISTRY)
```
hssyoo: Won't this fail if `HierarchicalEmitter` is passed for `event_hooks` since it doesn't implement `load_registry`? Can we make it so that `load_plugins` succeeds whether `HierarchicalEmitter` or `LazyInitEmitter` is passed?

On one hand, I think the chances of someone directly using `load_plugins` is slim. But it seems simple enough to avoid the potential break that we might as well do it.

exe/pyinstaller/hook-awscli.py
```diff
@@ -28,6 +30,14 @@
 ) + hooks.collect_submodules('awscli.s3transfer')
 hiddenimports += alias_packages_plugins
 
+# handlers.py uses importlib.import_module at runtime to load customization
```
hssyoo: nit: Not `handlers.py` anymore right?

tests/functional/test_handlers_registry.py
```diff
@@ -101,112 +101,3 @@ def test_all_main_command_table_ops_modules_are_importable():
         'The following MAIN_COMMAND_TABLE_OPS entries are invalid:\n'
         + '\n'.join(f'  - {v}' for v in violations)
     )
-
-
-def test_main_command_table_plugins_only_register_against_main():
```
hssyoo, line 106: Why are we removing these tests?

tests/functional/test_lazy.py
```diff
@@ -12,8 +12,44 @@
...
 import pytest
 
+from awscli.handlers_registry import MAIN_COMMAND_TABLE_OPS
 from awscli.lazy import LazyCommand
-from awscli.testutils import mock
+from awscli.testutils import BaseAWSHelpOutputTest, mock
+
+# Derive test parameters from MAIN_COMMAND_TABLE_OPS.
+_ADD_OPS = [op for op in MAIN_COMMAND_TABLE_OPS if op[0] == 'add']
+_RENAME_OPS = [op for op in MAIN_COMMAND_TABLE_OPS if op[0] == 'rename']
```
hssyoo: These are going to be empty lists since we're comparing enum members to strings.

hssyoo (commented):
Looks good, had a few minor comments

tests/functional/test_handlers_registry.py
aemous: See this comment for details: https://github.com/aws/aws-cli/pull/10305#discussion_r3237662201.

Basically, these tests were intentionally implemented as temporary to-be-removed in a future PR. They assert invariants that were necessary for the migration script we used to migrate all plugin handlers to the registry. Since the script was only one-time use, we no longer need the tests. The assertions that they check no longer need to be true.

awscli/plugin.py
aemous: Good point. I decided to implement fallback to eagerly loading all plugins in the registry when the event emitter does not have the type `LazyInitEmitter.`

hssyoo: It's a little weird that `load_plugins` accepts an `event_hooks` argument but then it implements an if/else flow to specifically handle `LazyInitEmitter`. The inconsistent interfaces between `HierarchicalEmitter` and `LazyInitEmitter` is causing `load_plugins` to compensate with this awkward hardcoded control flow.

For example, what happens if a new emitter class is added? Should it implement `load_registry` or not? Either way we're going to have to go back to `load_plugins` and hardcode another conditional case.

If we were starting from scratch, we'd probably have some base emitter class that both `LazyInitEmitter` and `HierarchicalEmitter` subclasses from and agree on a standard interface. I think that's a bit overkill here.

Instead, we could dispatch different implementations of `_load_registry` based on the type of `event_hooks`:
```
from functools import singledispatch
from botocore.hooks import HierarchicalEmitter

def load_plugins(plugin_mapping, event_hooks=None, include_builtins=True):
    ...
    if include_builtins:
        _load_registry(event_hooks)
    ...

@singledispatch
def _load_registry(event_hooks):
    raise NotImplementedError(
        f'No _load_registry implementation for '
        f'{type(event_hooks).__name__}'
    )

@_load_registry.register
def _(event_hooks: LazyInitEmitter):
    event_hooks.load_registry(PLUGIN_REGISTRY)

@_load_registry.register
def _(event_hooks: HierarchicalEmitter):
    # This is where the current `_eager_load_registry` implementation would go
```

aemous: Pushed a revision based on this feedback.

hssyoo (approved):
🏆

PR #10358, opened 2026-05-28
Fix fall back to IPv6 for ec2-instance-connect ssh --connection-type eice

leezu:
When --connection-type eice is set, resolve the destination address as private IPv4 with IPv6 as a fallback, instead of unconditionally using private IPv4. Without this, instances in IPv6-only subnets fail with "Unable to find any IP address on the instance to connect to." even though the EICE itself supports IPv6 destinations.


hssyoo (approved):
Looks good, thanks!

PR #10359, opened 2026-05-29
[v2] Fix parent reference warning

arandito:
*Issue #, if available:*

*Description of changes:*
Improves the parent-directory reference check in the S3 download path so that compare keys with a leading slash (e.g. `/../foo`) are correctly flagged. 

The compare key is now anchored against the destination directory root before `os.path.normpath()` is applied, preventing the leading slash from collapsing the `..` segment away.


hssyoo (approved):
Assuming CI passes 🏆

PR #10363, opened 2026-05-29
[v2] Lazily initialize built-in plugins to reduce command execution time

aemous:
*Issue #, if available:*

* Contributes to #9011.

*Notes:*

* This PR is a merge of a staging branch for the lazy plugin initialization project. All the commits in this PR have been previously reviewed and approved. The following PRs are in-scope for this merge PR:
  * https://github.com/aws/aws-cli/pull/10299
  * https://github.com/aws/aws-cli/pull/10305
  * https://github.com/aws/aws-cli/pull/10343
  * https://github.com/aws/aws-cli/pull/10356
  * https://github.com/aws/aws-cli/pull/10387
  * https://github.com/aws/aws-cli/pull/10392

*Description of changes:*

* **Merge lazy-plugin-initialization feature branch to v2**: Switch from eagerly importing and initializing all built-in plugins eagerly to lazy-loading them when relevant events are emitted, command(s) are invoked, or help docs are generated. These changes have been pre-approved.

*Description of tests:*

* Successfully ran pre-production build workflow (see Build Internal GitHub Action).

*Benchmarking data:*

A prototype for this solution on an `m7i.xlarge` EC2 instance with 2,000 iterations shows an average **112.1 ms (22.02%)** reduction of execution time for the `aws --version` command (p < 0.0001), and an average **108 ms (18%)** reduction of execution time for the `aws sts get-caller-identity` command (p < 0.0001). 

We expect similar numbers for most modeled commands. Commands that use more plugins at runtime will generally have higher execution time, since more plugins would be loaded.


exe/pyinstaller/hook-awscli.py
kdaily: Does this still need to be here, since it was removed? PR build was successful, does a PyInstaller-based installation raise any errors or warnings at install or runtime?

awscli/lazy_emitter.py
```diff
@@ -0,0 +1,161 @@
...
+            self._main_ops
+            and not self._main_ops_applied
+            and event_name == 'building-command-table.main'
+        ):
+            self._main_ops_applied = True
+            self._apply_main_command_table_ops(kwargs)
+        self._ensure_initialized(event_name)
+        return super()._emit(event_name, kwargs, stop_on_response)
+
+    def __copy__(self):
```
kdaily, line 151: Is the _init_cache specifically left out of a copy? Would updating it in a copy affect the original?

awscli/lazy_emitter.py
```diff
@@ -0,0 +1,161 @@
...
+                    cmd_module,
+                    cmd_class,
+                )
+            else:
+                raise RuntimeError(f'Unknown command table ops entry: {op}')
+
+        # Mark all building-command-table.main entries as initialized so
+        # _ensure_initialized never imports them. The ops list fully
+        # accounts for all plugins registered against this event.
+        for entry in PLUGIN_REGISTRY.get('building-command-table.main', []):
```
kdaily: Would we always want to use the global `PLUGIN_REGISTRY`, since this class takes a parameter `plugin_registry`?

kdaily (changes requested):
Couple of small clarifications, otherwise looks good. There's still a merge conflict that needs to be resolved as well.

awscli/lazy_emitter.py
aemous: > Is the _init_cache specifically left out of a copy?

It is implicitly copied via the `self.__dict__.copy()`, which involves a shallow copy of the init cache.

> Would updating it in a copy affect the original?

Yes, the init cache is a shared cache between all copies. If an event is emitted on one of the copies, the shared cache is updated for both copies. This is correct because the underlying data is the plugin registry, which is the same across all copies. We cache data derived from the plugin registry, so it is useful to have the cache be shared.

exe/pyinstaller/hook-awscli.py
aemous: Resolved in 6bccbba1da94282bd8cf3a3de0f3c5517d62734f

awscli/lazy_emitter.py
aemous: Resolved in 6bccbba1da94282bd8cf3a3de0f3c5517d62734f and 90b5155cbda6cfe5c7c2e1df94ddc9b25195cb2a.

kdaily:
🚢 - assuming that the CI test suites pass!

kdaily (approved):

PR #10375, opened 2026-06-02
[v2] Refactor to standalone source HeadObject for copies

hssyoo:
This PR refactors `HeadObject` calls during S3 to S3 multipart copies. Currently, the `HeadObject` request is called in `SetMetadataDirectivePropsSubscriber`. This isn't ideal since it couples `HeadObject` invocations to a specific subscriber. We want to open it up so that other subscribers can consume the `HeadObject` response, even if `SetMetadataDirectivePropsSubscriber` never fires.

Specifically, this refactor unblocks the following future changes:
* Update `SetTagsSubscriber` so it only fires if `TagCount` in `HeadObject` response is greater than 0.
* Hypothetical, but if we wanted to support more options for `--copy-props` (eg `tagging`) in the future, then these values can be fetched and copied independently from metadata.

### DIFF
--- a/awscli/customizations/s3/subscribers.py
+++ b/awscli/customizations/s3/subscribers.py
@@ -242,6 +242,35 @@ class DirectoryCreatorSubscriber(BaseSubscriber):
                 )
 
 
+_SOURCE_HEAD_OBJECT_CONTEXT_KEY = 'source_head_object_response'
+
+
+def get_copy_source_head_object(
+    future, source_client, cli_params, cached_response=None
+):
+    user_context = future.meta.user_context
+    if _SOURCE_HEAD_OBJECT_CONTEXT_KEY not in user_context:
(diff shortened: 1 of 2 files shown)


hssyoo:
Closing because the tag count optimization turns out to be infeasible.

The issue is that `HeadObject` will not return tag count at all under either conditions:
* Identity lacks permissions to get tags/tag count.
* The object doesn't have any tags.

Currently, S3 commands with `--copy-props default` will fail loudly if identity lacks tag permissions. With this optimization, it would start to fail silently, ie copy objects without tags since it doesn't see tag count in the `HeadObject` response. If `HeadObject` returned tag count = 0 when object doesn't have tags this would be possible but the current API doesn't give clients a way to distinguish between an object having 0 tags versus identity lacking tag permissions.

PR #10388, opened 2026-06-05
Fix symlinks and protect non-aws skills

AndrewAsseily:
*Issue #, if available:*

*Description of changes:*
  
- Fix crash when an agent's skill dir is a symlink (shutil.rmtree refuses to operate on symlinks)
- Update add-skill to skip and point at update-skill when a skill is already installed, instead of silently overwriting
- Update both add-skill and update-skill to skip non aws directories instead of clobbering user content
- Dedup the "already installed" / "skipped" messages so they print once per canonical path

### DIFF
--- a/tests/unit/customizations/agenttoolkit/test_add_skill.py
+++ b/tests/unit/customizations/agenttoolkit/test_add_skill.py
@@ -313,6 +313,7 @@ def test_add_skill_skips_symlinked_existing_install(tmp_path):
     rc, output = _run_add(configs, ['--skill-name', 'aws-s3'])
     assert rc == 0
     assert 'aws-s3 is already installed (v1)' in output
+    assert 'update-skill --skill-name aws-s3' in output
     assert link.is_symlink()
     assert (target / 'SKILL.md').read_text() == 'old'
 
--- a/tests/unit/customizations/agenttoolkit/test_add_skill.py
+++ b/tests/unit/customizations/agenttoolkit/test_add_skill.py
@@ -17,6 +17,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from awscli.customizations.agenttoolkit.add_skill import AddSkillCommand
+from awscli.customizations.agenttoolkit.agents import SKILL_METADATA_FILENAME
 from awscli.customizations.agenttoolkit.utils import AgentToolkitServiceError
 from awscli.customizations.exceptions import ParamValidationError
 from tests.unit.customizations.agenttoolkit.utils import (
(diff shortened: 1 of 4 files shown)


tests/unit/customizations/agenttoolkit/test_add_skill.py
```diff
@@ -228,6 +229,111 @@ def test_add_skill_universal_row_coexists_with_override_agent(tmp_path):
...
+    target.mkdir()
+    (target / 'SKILL.md').write_text('old')
+    (target / SKILL_METADATA_FILENAME).write_text('{"version": "v1"}')
+    link = tmp_path / '.test-agent' / 'skills' / 'aws-s3'
+    link.symlink_to(target, target_is_directory=True)
+
+    configs = [make_config(tmp_path)]
+    rc, output = _run_add(configs, ['--skill-name', 'aws-s3'])
+    assert rc == 0
+    assert 'aws-s3 is already installed (v1)' in output
```
aemous, line 315: non-blocking nit: may be worth also asserting the recommendation is in the output in this case.

tests/unit/customizations/agenttoolkit/test_update_skill.py
```diff
@@ -140,6 +140,39 @@ def test_update_skill_missing_marker_skipped(tmp_path):
         _run_update(configs, ['--skill-name', 'aws-s3'], remote_version='v2')
 
 
+def test_update_skill_through_symlinked_dir(tmp_path):
+    # A user symlinks the per-skill dir into a shared location. update-skill
+    # must follow the link and overwrite the target in place (shutil.rmtree
+    # refuses to operate on the symlink itself), leaving the link intact.
+    (tmp_path / '.test-agent' / 'skills').mkdir(parents=True)
```
aemous, line 147: This is actually writing to disk right? If so, this would be considered a functional test and not a unit test, since it is integrating with system IO. Would be nice to see this under `tests/functional` dir.

Similar comment for the other file for all tests writing to IO.

aemous (commented):
code looks good. I did some manual testing with updating Kiro skills with a symlink to a pre-installed skills folder, and the behavior seems expected.

tests/unit/customizations/agenttoolkit/test_add_skill.py
AndrewAsseily: Addressed in rev 2

tests/unit/customizations/agenttoolkit/test_update_skill.py
AndrewAsseily: Fair point. Though that's just following the established pattern (in many `tests/unit/...` in the codebase). Moving them all would be a refactor I'd rather not bundle into this PR

aemous (approved):
LGTM. Did some basic manual testing and behavior looks as expected.

PR #10392, opened 2026-06-10
[v2] Merge v2 into lazy-plugin-initialization

aemous:
*Description of changes:*

* Merged branch `v2` into `lazy-plugin-initialization`:
  * Migrate `agent-toolkit` customization to the new handlers_registry system. Updated file in merge commit: `awscli/handlers_registry.py`.
* Update lazy_emitter so that it references its instance reference to the plugin registry instead of the global plugin registry constant.

*Description of tests:*

* Successfully ran all tests including pre-prod build workflow (see GitHub Actions).


awscli/customizations/agenttoolkit/__init__.py
```diff
@@ -56,13 +56,6 @@ def _inject_nonprod_header(request, **kwargs):
         request.headers[NONPROD_ACCESS_TOKEN_HEADER] = token
 
 
-def _rename_service(command_table, session, **kwargs):
```
kdaily, line 59: Nit, but why is this being removed? I see its in its own commit, but is it related to the lazy plugin init?

awscli/lazy_emitter.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, plugin_registry=None, main_command_table_ops=None):
             self.load_registry(self._registry)
 
     def load_registry(self, registry):
+        self._registry = registry
```
kdaily, line 59: Not sure I understand what this is for? I see that in the init, a registry can be provided, which is set in `self._registry`. But then this resets it? Why does it need to be set in the init (line 52)?

awscli/customizations/agenttoolkit/__init__.py
aemous: This function was only needed for the old `awscli/handlers.py` method of doing these renames. In the new system of `awscli/handlers_registry.py`, we model the renames instead of calling the functions from handlers.py.

awscli/lazy_emitter.py
aemous: We don't actually pass in a registry to the init function, we only pass it to load_registry. The only place we pass it to the init function is in test suites.

kdaily: The init takes `plugin_registry`?

aemous: Ah, I see the confusion in my wording. Init accepts a plugin registry. We don't pass the registry to init during construction in the CLI codebase (non-tests), instead we only pass the registry during calling load_registry. We only pass the registry to init in our test suites.

kdaily (approved):

PR #10396, opened 2026-06-10
docs(ec2): Add examples for modify/describe ipam-pool-allocation APIs

Danyu999:
Add CLI examples for two IPAM pool allocation commands:
- modify-ipam-pool-allocation: modifying the description of an allocation
- describe-ipam-pool-allocations: listing all allocations and filtering by ID

### DIFF
--- a/awscli/examples/ec2/describe-ipam-pool-allocations.rst
+++ b/awscli/examples/ec2/describe-ipam-pool-allocations.rst
@@ -1,0 +1,64 @@
+**Example 1: To describe all your IPAM pool allocations**
+
+The following ``describe-ipam-pool-allocations`` example describes all IPAM pool allocations owned by you.
+
+(Linux)::
+
+    aws ec2 describe-ipam-pool-allocations
+
+(Windows)::
+
+    aws ec2 describe-ipam-pool-allocations
+
+Output::
+
+    {
+        "IpamPoolAllocations": [
+            {
+                "Cidr": "10.0.0.0/24",
+                "IpamPoolAllocationId": "ipam-pool-alloc-018ecc28043b54ba38e2cd99943cebfbd",
+                "Description": "Engineering team allocation",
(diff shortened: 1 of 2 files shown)


awscli/examples/ec2/describe-ipam-pool-allocations.rst
```diff
@@ -0,0 +1,64 @@
+**Example 1: To describe all your IPAM pool allocations**
+
+The following ``describe-ipam-pool-allocations`` example describes all IPAM pool allocations owned by you.
+
+(Linux)::
```
ashovlin, line 5: We don't typically separate Linux vs. Windows unless there is something specific to each OS in how the command is invoked.

ashovlin (changes requested):

awscli/examples/ec2/describe-ipam-pool-allocations.rst
Danyu999: Ack, I was following pre-existing patterns for other related APIs, such as [allocate-ipam-pool-cidr](https://github.com/aws/aws-cli/blob/develop/awscli/examples/ec2/allocate-ipam-pool-cidr.rst) and [get-ipam-pool-allocations](https://github.com/aws/aws-cli/blob/develop/awscli/examples/ec2/get-ipam-pool-allocations.rst). Just to verify, which of the following would you recommend following? I prefer 1 or 2.

1. Continue with not separating Linux vs. Windows for new APIs
2. Continue with following established patterns with related APIs where we do separate Linux vs. Windows even without specific differences
3. Refactor examples for existing and new related APIs to not separate Linux vs. Windows

ashovlin: Ah, didn't realize other older examples had this. This is fine as-is then, we can clean up centrally in the future.

ashovlin (approved):

Issue #10031, opened 2026-01-29
v2.33.9: Permission denied on ~/.aws/config when running as non-root user after initialization


seanuleh:
### Describe the bug

AWS CLI v2.33.9 creates `~/.aws` directory with root ownership when `aws --version` is executed with elevated privileges. This causes "Permission denied" errors for non-root users attempting to run subsequent AWS CLI commands. This is a regression from v2.33.8 where `aws --version` had no side effects.

## Environment
- **AWS CLI Version**: 2.33.9 (broken) vs 2.33.8 (working)
- **Platform**: Linux (ARM64)
- **OS**: Ubuntu 22.04
- **User Context**: AWS CLI installed and verified as root, but run as non-root user
- **Installation Method**: Any method that verifies installation with `sudo aws --version`

## Issue Description
When `aws --version` is executed with root privileges (e.g., `sudo aws --version` or during installation verification), AWS CLI v2.33.9 creates the `~/.aws/cli/cache/` directory structure with root ownership. This prevents non-root users from running any AWS CLI commands that need to write to `~/.aws/config` or `~/.aws/credentials`.

**Critical Finding**: `aws --version` should be a read-only operation with zero side effects. In v2.33.8 it was. In v2.33.9 it creates directories and files.

Error messages:
```
[Errno 13] Permission denied: '/home/runner/.aws/config'
[Errno 13] Permission denied: '/home/runner/.aws/credentials'
```

This issue does NOT occur with AWS CLI v2.33.8.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

**After running `sudo aws --version`:**
```bash
$ ls -la ~/ | grep .aws
# No output - no .aws directory created ✅
```

**After running `aws configure set region us-east-2 --profile test` as non-root user:**
```bash
$ ls -la ~/.aws/
drwxr-xr-x. 2 user user  20 Jan 29 00:53 .
-rw-------. 1 user user  34 Jan 29 00:53 config
```

- `aws --version` has **zero side effects** - no directories created ✅
- `~/.aws` directory created only when needed, owned by the actual user ✅
- No cache subdirectories created during simple operations
- All subsequent AWS CLI commands succeed

### Current Behavior

**After running `sudo aws --version`:**
```bash
$ ls -la ~/.aws/
drwxr-xr-x. 3 root root    17 Jan 29 00:50 .
drwxr-xr-x. 3 root root    19 Jan 29 00:50 cli

[...]

PR #10398, opened 2026-06-11
[v2] Add Session IDs to User-Agent header even when cache does not exist

aemous:
*Issue #, if available:*

* CLI-8466 (internal)

*Description of changes:*

* Fixed a bug where session IDs are added to the User-Agent HTTP header even when the local cache does not exist.
* Modified `CLISessionDatabaseConnection` to allow a cache directory override to be passed in during initialization. This improves testability.
* Added new regression test that verifies (1) local cache is created and (2) database connection succeeds when the local cache does not exist.

*Description of tests:*

* Successfully ran all tests and the pre-production build workflow (see GitHub Actions).
* Ran a manual integration test workflow to verify expected behavior (manual test workflow described below).

### Manual test workflow

1. Deleted `~/.aws/cli/cache`: `rm -rf ~/.aws/cli/cache`.
2. Installed AWS CLI v2 WITHOUT the changes.
3. Ran the following command: `python3 -m awscli s3api list-buckets --debug 2> debug.txt`.
4. Inspected `debug.txt` and confirmed the `User-Agent` HTTP header does not contain a session ID: `aws-cli/2.35.2 md/awscrt#0.32.2 ua/2.1 os/macos#25.5.0 md/arch#arm64 lang/python#3.12.6 md/pyimpl#CPython m/c,E,C,a,n cfg/retry-mode#standard md/installer#source md/prompt#off md/command#s3api.list-buckets`.
5. Deleted `~/.aws/cli/cache`: `rm -rf ~/.aws/cli/cache`.
6. Installed AWS CLI v2 WITH the changes (HEAD of branch).
7. Ran the following command: `python3 -m awscli s3api list-buckets --debug 2> debug.txt`.
8. Inspected `debug.txt` and confirmed the `User-Agent` HTTP header contains a session ID: `aws-cli/2.35.2 md/awscrt#0.32.2 ua/2.1 os/macos#25.5.0 md/arch#arm64 lang/python#3.12.6 md/pyimpl#CPython m/n,C,E,c,a cfg/retry-mode#standard md/installer#source sid/b871d801befe md/prompt#off md/command#s3api.list-buckets`


tests/integration/test_cli.py
```diff
@@ -310,15 +310,6 @@ def test_version(self):
         )
         self.assertTrue(re.fullmatch(user_agent_regex, version_output))
 
-    def test_version_does_not_create_cache_directory(self):
```
AndrewAsseily, line 313: The fix makes sense to me. I'd want to verify the "sudo scenario" before merging (This is the same pattern that caused [#10031](https://github.com/aws/aws-cli/issues/10031)).
  
Can you test the scenario of someone that runs `sudo aws --version` as their first CLI command and confirm the CLI doesn't crash for subsequent non-sudo calls? The deleted test `test_version_does_not_create_cache_directory` was protecting against something like this.

AndrewAsseily (changes requested):

tests/integration/test_cli.py
aemous: Good catch. I tried running `sudo aws --version` as the first CLI command and I confirm the cache directory is created with root-owner perms. Subsequent commands don't crash from what I can see, but subsequent CLI commands do not update `~/.aws/cli/cache/session.db`. If I run subsequent commands with sudo, then they do update `~/.aws/cli/cache/session.db`. 

We'll want to address this.

AndrewAsseily: Session ID tracking is purely telemetry, so if it fails for any reason, I'd rather the CLI move on silently. The user's actual commands shouldn't be affected/interrupted.

My takeaway from https://github.com/aws/aws-cli/issues/10031 is that the safest permission is not creating anything you don't need to. I wonder if lazy init solves this. 

Syncing offline

aemous: I've published a revision that should address the original concern in this thread.

AndrewAsseily (approved):
LGTM 🏆

PR #10404, opened 2026-06-12
Add SSO start URL resolver for vanity domain support

AndrewAsseily:
*Description of changes:* Adds a start URL resolver module for vanity domain support in SSO login flows. Subsequent PRs will wire it into `aws sso login` and `aws configure sso`.

### DIFF
--- a/awscli/customizations/sso/resolve.py
+++ b/awscli/customizations/sso/resolve.py
@@ -40,9 +40,20 @@ _REGION_PATTERNS = (
     ),
 )
 
-_ALL_PARTITIONS = ('aws', 'aws-cn', 'aws-us-gov')
-
 _MAX_REDIRECTS = 1
+
+_DEFAULT_PORTS = {'https': 443, 'http': 80}
+
+_DEFAULT_TIMEOUT = 10
+
+
+def _normalize_url(url):
+    parsed = urllib.parse.urlparse(url)
+    netloc = parsed.hostname or ''
+    if parsed.port and parsed.port != _DEFAULT_PORTS.get(parsed.scheme):
+        netloc = f'{netloc}:{parsed.port}'
+    path = parsed.path.rstrip('/')
+    return urllib.parse.urlunparse((parsed.scheme, netloc, path, '', '', ''))
 
 
 def is_aws_owned_domain(hostname):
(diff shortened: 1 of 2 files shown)


awscli/customizations/sso/resolve.py
```diff
@@ -0,0 +1,187 @@
...
+        r'^[^.]+\.portal\.(?P<region>[a-z0-9-]+)\.app\.aws$', re.IGNORECASE
+    ),
+    # {idcInstanceId}.{region}.portal.amazonaws.com
+    re.compile(
+        r'^[^.]+\.(?P<region>[a-z0-9-]+)\.portal\.amazonaws\.com$',
+        re.IGNORECASE,
+    ),
+)
+
+_ALL_PARTITIONS = ('aws', 'aws-cn', 'aws-us-gov')
```
ashovlin: [I commented on the internal design doc too] Generally we don't want to validate regions against the CLI's current data, since if a new region is launched users are forced to upgrade to update the client validation.

That said, if keeping this could we loop over `partitions.json` instead so we at least don't need to update this constant for partitions?

awscli/customizations/sso/resolve.py
```diff
@@ -0,0 +1,187 @@
...
+        def redirect_request(self, req, fp, code, msg, headers, newurl):
+            return None
+
+    opener = urllib.request.build_opener(_NoRedirectHandler)
+    redirect_codes = (301, 302, 303, 307, 308)
+
+    for _attempt in range(_MAX_REDIRECTS):
+        try:
+            req = urllib.request.Request(url, method='HEAD')
+            resp = opener.open(req, timeout=10)
```
ashovlin: May want to consider honoring `--cli-connect-timeout`/`--cli-read-timeout`. Hopefully it doesn't take longer than 10 seconds, but generally prefer for high-level commands to honor global options where possible.

awscli/customizations/sso/resolve.py
ashovlin: That said, we may want whatever validation we do for S3 redirects to avoid malicious redirections.

AndrewAsseily: Agreed. Dropped region validation entirely

awscli/customizations/sso/resolve.py
AndrewAsseily: Added a timeout parameter to resolve_start_url. Defaults to 10s but the caller can pass through

ashovlin (approved):

PR #10414, opened 2026-06-15
Vendor botocore and s3transfer

kdaily:
*Issue #, if available:*

*Description of changes:*

This is a similar endeavor as https://github.com/aws/aws-cli/pull/6494.

This change SHOULD NOT be merged as is, and is intentionally left in draft mode to prevent merging. The PR is for a review of the outcome of vendoring, and will need to be recreated closer to the date of maintenance mode.

As in the previous PR, there are separate commits for each major step of vendoring.

Note that there will be conflicts in this PR as the underlying botocore and s3transfer versions are updated as part of our existing release process.

## Test Audit Results

| Project | Passed | Failed | Skipped |
|---------|--------|--------|---------|
| AWS CLI baseline (develop) | 15983 | 0 | 17 |
| This PR (vendored aws-cli) | 76389 | 0 | 194 |
| AWS CLI only (vendored, excl. botocore/s3transfer tests) | 15991 | 0 | 16 |
| botocore standalone | 59786 | 0 | 126 |
| s3transfer standalone | 675 | 0 | 52 |
| **Net difference** | **-63** | **0** | **0** |
| **Vendoring delta** (CLI only - baseline) | **8** | **0** | **-1** |

The net difference = This PR - (AWS CLI only + botocore + s3transfer)
The vendoring delta = AWS CLI only - AWS CLI baseline

Notes on the audit:

There were 63 tests in s3transfer for `processpool` removed, so these aren't run in the PR but ARE in the s3transfer.

The one difference is due to a file present in `s3transfer/.changes/next-release`; this triggers the test at `tests/unit/test_changelog.py` to run, which did not run in the vended test suite because the copy did not include that directory. This SHOULD NOT happen when we do this for real, because we should do it after a release. This just happened to get a commit that included this change before the release finished.


setup.py
```diff
@@ -24,12 +24,13 @@ def find_version(*file_paths):
 
 
 install_requires = [
-    'botocore==1.43.32',
```
hssyoo: Do we need to add an `extras_require` for optional CRT dependency?

kdaily: Done in 7ce0cf563c422f1a061708858d002da4632642b1 (will persist on regeneration of this PR).

jonathan343 (approved):

PR #10419, opened 2026-06-16
Sso login integration

AndrewAsseily:
*Description of changes:* Wires the start URL resolver into `aws sso login`. Vanity URLs are resolved before login, the resolved URL is passed to OIDC service APIs, and `sso_region` is updated in the user's config after a successful login. Direct AWS-owned URLs are unchanged. Subsequent PR will wire it into `aws configure sso`. 


awscli/botocore/utils.py
```diff
@@ -3514,7 +3514,11 @@ def fetch_token(
         force_refresh=False,
         registration_scopes=None,
         session_name=None,
+        resolved_start_url=None,
     ):
+        # resolved_start_url is the AWS-owned URL to pass to service APIs.
+        # start_url (possibly a vanity URL) is used as the cache key.
+        self._resolved_start_url = resolved_start_url or start_url
```
ashovlin: Generally setting a private member outside of the constructor can be a trap since callers might not call this method first. Can we plumb it thorough parameters instead?

awscli/customizations/sso/login.py
```diff
@@ -44,20 +52,54 @@ class LoginCommand(BaseSSOCommand):
...
             force_refresh=True,
             session_name=sso_config.get('session_name'),
             registration_scopes=sso_config.get('registration_scopes'),
             use_device_code=parsed_args.use_device_code,
         )
+
+        # Only rewrite sso_region after successful login.
+        hostname = urlparse(start_url).hostname
+        if hostname and not is_aws_owned_domain(hostname):
+            if configured_region != region:
```
ashovlin: Is just `configured_region != region` adequate (only write it out if it's different than what was previously saved)? If so, skip the url parsing again.

tests/functional/sso/test_login_resolve.py
```diff
@@ -0,0 +1,172 @@
...
+        self.run_cmd('sso login')
+        with open(self.config_file) as f:
+            config_content = f.read()
+        self.assertIn('sso_region = us-east-1', config_content)
+
+    def test_legacy_passes_resolved_url_to_device_authorization(self):
+        self.add_oidc_device_responses(self.access_token)
+        self.run_cmd('sso login')
+        device_auth_call = None
+        for op, params in self.operations_called:
```
ashovlin: Generally prefer to hardcode the expected index in to `operations_called`, it's simpler, and we should know what order and how many calls a given command is making.

tests/functional/sso/test_login_resolve.py
```diff
@@ -0,0 +1,172 @@
...
+                break
+        self.assertIsNotNone(device_auth_call)
+        self.assertEqual(device_auth_call['startUrl'], self.resolved_url)
+
+
+class TestLoginWithDirectUrl(BaseSSOTest):
+    def setUp(self):
+        super().setUp()
+        self.start_url = 'https://ssoins-abc.portal.us-west-2.app.aws'
+        self.sso_region = 'us-west-2'
```
ashovlin: If a test value is the same as the base class, can we skip setting it in a test?

tests/functional/sso/__init__.py
```diff
@@ -42,6 +42,19 @@ def setUp(self):
             self.token_cache_dir,
         )
         self.token_cache_dir_patch.start()
+        self._resolve_patch = mock.patch(
+            'awscli.customizations.sso.login.resolve_start_url',
+            side_effect=lambda start_url, session, **kwargs: (
+                start_url,
+                kwargs.get('configured_region', self.sso_region),
+            ),
+        )
```
ashovlin, line 51: A lot of the tests patch this again, is it needed here? Can we get away with it in just one place for a given test?

ashovlin: Maybe still patch the implementation here, but just return a value that the test code can set before the CLI invocation?

tests/functional/sso/test_login_resolve.py
```diff
@@ -0,0 +1,172 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
+
+from awscli.testutils import mock
+from tests.functional.sso import BaseSSOTest
+
+
+class TestLoginWithVanityUrl(BaseSSOTest):
```
ashovlin: Can we move these in to `test_login.py`? I don't think they're sufficiently different to need a new file.

ashovlin (changes requested):

ashovlin (approved):

PR #10420, opened 2026-06-16
Configure sso integration

AndrewAsseily:
*Description of changes:* Wires the start URL resolver into `aws configure sso` and `aws configure sso-session`. When a vanity URL is provided, the region is auto-derived from the redirect target and the SSO region prompt is skipped. For direct AWS-owned URLs or when resolution fails, the region prompt fires as before. Also passes the resolved URL through to `do_sso_login` so `RegisterClient` receives the normalized AWS-owned URL.


tests/unit/customizations/configure/test_sso_resolve.py
```diff
@@ -0,0 +1,222 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import pytest
+
+from awscli.customizations.exceptions import ConfigurationError
+from awscli.testutils import mock
+
+
+class TestConfigureSSOVanityUrlResolution:
```
ashovlin: Like the other PR, could we combine with test_sso.py? That looks like it has a lot of mocking set up, and assertions all the way through to the config file.

ashovlin (changes requested):

ashovlin (approved):

PR #10427, opened 2026-06-19
[v2] Implement macOS/Linux install script

hssyoo:
Helper bash script to install AWS CLI v2 on macOS and Linux.

### High-level overview:
* Detect the system platform and architecture to download the right artifact.
* The script is just a wrapper around the official installers. The same constraints apply to the script (eg Alpine not supported).
* If `--version` isn't specified, it just grabs the latest version. The script fetches https://awscli.amazonaws.com/v2/version.txt in case the user already has the latest version installed in the configured path.
* Signatures are verified. On macOS, the `pkgutil` tool used to verify the signature is a required dependency because it's available by default. On Linux, `gpg` is optional because it's not uncommon for Linux distros to not include them. A warning is emitted in this case.
* Downgrades are not supported. We still expose a `--version` parameter so users can pin to a specific version in automation workflows.

### Install layout
The main difference between the install script and official installers is that the script will not install system-wide by default (which requires `sudo`). Instead, it'll default to user-local installs, respecting XDG environment variables set by the user.

* By default, install directory is `$HOME/.local/share` and bin directory is `$HOME/.local/bin`.
* If `XDG_DATA_HOME` and `XDG_BIN_HOME` are set, then use those as install/bin dirs.
* If `--system` flag is passed, then install system-wide to `/usr/local/` and `/usr/local/bin` (same default behavior as installers today). Requires `sudo`.
  * 1 edge case here is that if `sudo` is run _without_ the `--system` flag, then the install lands in the user's root `$HOME`. It also ignores XDG environment variables set by the user. We emit a warning in this case so the user knows where it was installed to and why.

### `install.json`
This script introduces a new file called `install.json` that lives on disk in the install directory. It contains metadata about how the AWS CLI was installed. This needed to be distinct from `metadata.json` because `metadata.json` is created at build-time, whereas the information we need isn't known until install-time. `install.json` is used to assist the upcoming `update` command by preserving info on how it was originally installed.

### DIFF
--- a/scripts/install-v2/install.sh
+++ b/scripts/install-v2/install.sh
@@ -652,7 +652,7 @@ json_escape() {
   printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g'
 }
 
-# Writes install-time metadata next to the bundled metadata.json. `aws update`
+# Writes install-time metadata to install.json. `aws update`
 # reads this back to learn where and how the CLI was installed.
 write_install_json() {
   local install_json
--- a/scripts/install-v2/install.sh
+++ b/scripts/install-v2/install.sh
@@ -1,0 +1,714 @@
+#!/usr/bin/env bash
+#
+# AWS CLI v2 install script for Linux and macOS.
+#
+# Defaults to a user-local install under XDG paths. Use --system for the
+# traditional /usr/local layout (requires root). See --help for details.
+
+set -euo pipefail
+
+# ============================================================================
+# Configuration
+# ============================================================================
+
+readonly DOWNLOAD_BASE_URL="https://awscli.amazonaws.com"
+readonly LATEST_VERSION_URL="${DOWNLOAD_BASE_URL}/v2/version.txt"
+
+readonly DISTRIBUTION_SOURCE="${AWS_CLI_DISTRIBUTION_SOURCE_OVERRIDE:-script-exe}"
+readonly DOWNLOAD_QUERY="?src=${DISTRIBUTION_SOURCE}"
+
+readonly SYSTEM_INSTALL_DIR="/usr/local/aws-cli"
+readonly SYSTEM_BIN_DIR="/usr/local/bin"
+
+readonly AWS_CLI_PKG_TEAM_ID="94KV3E626L"
+
+# AWS CLI public signing key, used to verify the Linux .sig.
+# Fingerprint: FB5D B77F D5C1 18B8 0511 ADA8 A631 0ACC 4672 475C
+# Expires 2026-07-07.
+readonly AWS_CLI_PGP_KEY='-----BEGIN PGP PUBLIC KEY BLOCK-----
+
+mQINBF2Cr7UBEADJZHcgusOJl7ENSyumXh85z0TRV0xJorM2B/JL0kHOyigQluUG
+ZMLhENaG0bYatdrKP+3H91lvK050pXwnO/R7fB/FSTouki4ciIx5OuLlnJZIxSzx
+PqGl0mkxImLNbGWoi6Lto0LYxqHN2iQtzlwTVmq9733zd3XfcXrZ3+LblHAgEt5G
+TfNxEKJ8soPLyWmwDH6HWCnjZ/aIQRBTIQ05uVeEoYxSh6wOai7ss/KveoSNBbYz
+gbdzoqI2Y8cgH2nbfgp3DSasaLZEdCSsIsK1u05CinE7k2qZ7KgKAUIcT/cR/grk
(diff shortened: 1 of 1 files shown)


scripts/install-v2/install.sh
```diff
@@ -0,0 +1,721 @@
...
+  local aws_bin
+  aws_bin="$(candidate_aws_binary)"
+  PRE_INSTALL_VERSION="$(read_installed_version "$aws_bin" || true)"
+}
+
+fetch_latest_version() {
+  local out
+  out="$(curl --fail --silent --show-error --location \
+              --connect-timeout 10 --retry 0 \
+              "$LATEST_VERSION_URL" 2>/dev/null || true)"
```
kdaily, line 519: Is there any risk for downloading this failing silently and putting a version out? It seems like no (since our versioned artifact and latest artifact are published together).

scripts/install-v2/install.sh
```diff
@@ -0,0 +1,721 @@
...
+
+  local cmd
+  for cmd in "${required[@]}"; do
+    if ! command -v "$cmd" >/dev/null 2>&1; then
+      missing+=("$cmd")
+    fi
+  done
+
+  if [ ${#missing[@]} -gt 0 ]; then
+    error 1 "missing required dependencies: ${missing[*]}. Install them via your package manager and re-run."
```
kdaily, line 254: nit: I recall that there were some other potential error codes for cases like missing dependencies. They are all now returned as `1`. Is that intentional?

scripts/install-v2/install.sh
```diff
@@ -0,0 +1,721 @@
...
+
+capture_pre_install_version() {
+  local aws_bin
+  aws_bin="$(candidate_aws_binary)"
+  PRE_INSTALL_VERSION="$(read_installed_version "$aws_bin" || true)"
+}
+
+fetch_latest_version() {
+  local out
+  out="$(curl --fail --silent --show-error --location \
```
kdaily, line 517: There is a utility included in the script for `curl` (`curl_to_file`). The only difference is the `|| true`. Would this be usable as well?

scripts/install-v2/install.sh
```diff
@@ -0,0 +1,721 @@
...
+  if [ -z "$POST_INSTALL_VERSION" ]; then
+    error 9 "post-install check failed: '$aws_bin --version' did not run successfully."
+  fi
+}
+
+json_escape() {
+  printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g'
+}
+
+# Writes install-time metadata next to the bundled metadata.json. `aws update`
```
kdaily: nit: this documentation was slightly confusing. I think I understand that `metadata.json` and `install.json` are in the same directory, is that pertinent information here? Otherwise:

```suggestion
# Writes install-time metadata next to install.json. `aws update`
```

kdaily (changes requested):
Looks good overall, left a few questions.

scripts/install-v2/install.sh
hssyoo: It should be fine since artifacts are published together and cache is invalidated as the last step

scripts/install-v2/install.sh
hssyoo: Hm not sure what you're referring to with other potential error codes. It's intentional for them to just return 1

scripts/install-v2/install.sh
hssyoo: `install.json` is what contains install-time metadata. I agree it's not important to mention its relative location with `metadata.json` but it'd be incorrect to say install-time metadata is next to `install.json`

scripts/install-v2/install.sh
hssyoo: It doesn't use `curl_to_file` because it doesn't need to save the latest version to disk, it just needs to store it in a variable

scripts/install-v2/install.sh
kdaily: Oops, I actually meant:

```suggestion
# Writes install-time metadata to install.json. `aws update`
```

kdaily: No strong feeling here, since its not really incorrect.

kdaily (approved):

scripts/install-v2/install.sh
hssyoo: Chatted offline - latest commit standardizes error codes

scripts/install-v2/install.sh
hssyoo: Applied the suggestion

kdaily (approved):
LGTM

PR #10428, opened 2026-06-19
[v2] Implement Windows install script

hssyoo:
Helper bash script to install AWS CLI v2 on Windows.

### High-level overview:
* The script is just a wrapper around the official installers. The same constraints apply to the script.
* If `--version` isn't specified, it just grabs the latest version. The script fetches https://awscli.amazonaws.com/v2/version.txt in case the user already has the latest version installed in the configured path.
* Signature status is verified.
* Downgrades are not supported. We still expose a `--version` parameter so users can pin to a specific version in automation workflows.

### Install layout
The main difference between the install script and official installers is that the script will not install system-wide by default (which requires admin elevation). Instead, it'll default to user-local installs.

* By default, install directory is `%LOCALAPPDATA%\Programs\Amazon\AWSCLIV2`
* XDG environment variables aren't supported, as they're Unix-specific specs.
* If `--system` flag is passed, then install system-wide to `%ProgramFiles%\Amazon\AWSCLIV2` (same default behavior as installers today). Requires admin elevation.

### DIFF
--- a/scripts/install-v2/install.ps1
+++ b/scripts/install-v2/install.ps1
@@ -352,7 +352,7 @@ function Run-Installer {
     $proc = Start-Process -FilePath 'msiexec' -ArgumentList $msiArgs `
                           -WorkingDirectory $msiDir -Wait -PassThru
     if ($MsiexecSuccessCodes -notcontains $proc.ExitCode) {
-        Throw-Error 8 "msiexec failed with exit code $($proc.ExitCode)"
+        Throw-Error 1 "msiexec failed with exit code $($proc.ExitCode)"
     }
 }
 
@@ -364,7 +364,7 @@ function Verify-InstallRuns {
     $awsExe = Get-CandidateAwsBinary
     $Script:PostInstallVersion = Read-InstalledVersion $awsExe
     if (-not $Script:PostInstallVersion) {
-        Throw-Error 9 "post-install check failed: '$awsExe --version' did not run successfully."
+        Throw-Error 1 "post-install check failed: '$awsExe --version' did not run successfully."
     }
 }
 
--- a/scripts/install-v2/install.ps1
+++ b/scripts/install-v2/install.ps1
@@ -1,0 +1,407 @@
+# AWS CLI v2 install script for Windows (PowerShell 5.1+).
+#
+# Defaults to a user-local install under %LOCALAPPDATA%. Use -System for the
+# traditional Program Files layout (requires admin). See -Help for details.
+
+param(
+    [string] $Version,  # specific version to install; empty means "latest"
+    [switch] $System,   # install system-wide (Program Files); requires admin
+    [switch] $Quiet,    # suppress non-essential stdout
+    [switch] $Help
+)
+
+# Stop on any error so failures don't silently continue.
+$ErrorActionPreference = 'Stop'
+# PowerShell 5.1 defaults to TLS 1.0; force 1.2 for the CDN.
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+# ============================================================================
+# Configuration
+# ============================================================================
+
+$DownloadBaseUrl    = 'https://awscli.amazonaws.com'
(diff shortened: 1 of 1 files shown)


harshagm665-netizen:
I've taken a close look at the issue and I believe I've identified the root cause of the problem. The current implementation of the Windows install script for AWS CLI v2 doesn't handle the installation layout correctly, particularly when it comes to user-local installs versus system-wide installs.

From what I understand, the script should default to user-local installs, but it's not properly checking for the `XDG_DATA_HOME` and `XDG_BIN_HOME` environment variables to determine the install and bin directories. Additionally, the script doesn't correctly handle the `--system` flag to install system-wide.

To fix this, I propose the following changes:
```bash
#!/bin/bash

# Default install directory
INSTALL_DIR="%LOCALAPPDATA%\Programs\Amazon\AWSCLIV2"

# Check if XDG_DATA_HOME and XDG_BIN_HOME are set
if [ -n "$XDG_DATA_HOME" ] && [ -n "$XDG_BIN_HOME" ]; then
  INSTALL_DIR="$XDG_DATA_HOME\Amazon\AWSCLIV2"
  BIN_DIR="$XDG_BIN_HOME"
else
  BIN_DIR="%LOCALAPPDATA%\Programs\Amazon\AWSCLIV2\bin"
fi

# Check if --system flag is passed
if [ "$1" = "--system" ]; then
  INSTALL_DIR="%ProgramFiles%\Amazon\AWSCLIV2"
  BIN_DIR="%ProgramFiles%\Amazon\AWSCLIV2\bin"
  # Require admin elevation
  if ! net session > /dev/null 2>&1; then
    echo "Admin elevation required for system-wide install"
    exit 1
  fi
fi

# Rest of the script remains the same
```
I've updated the script to correctly check for the `XDG_DATA_HOME` and `XDG_BIN_HOME` environment variables and use them to determine the install and bin directories. I've also added a check for the `--system` flag to install system-wide, which requires admin elevation.

I'd love to get your feedback on this proposed fix. If you're interested, I can submit a PR with these changes. Please let me know if this looks good to you or if you'd like me to make any further adjustments.

scripts/install-v2/install.ps1
```diff
@@ -0,0 +1,446 @@
...
+function Download-Installer {
+    $Script:InstallerPath = Join-Path $Script:TempDir 'AWSCLIV2.msi'
+    Download-WithRetry -Url (Get-InstallerUrl) -Out $Script:InstallerPath -Label 'AWS CLI installer'
+}
+
+# ============================================================================
+# Signature verification
+# ============================================================================
+
+function Verify-Installer {
```
AndrewAsseily, line 261: For my understanding, the macOS/Linux (https://github.com/aws/aws-cli/pull/10427/changes#diff-78cd8f23e0991b3ba2aa2f5ebc48ad549685742560cf9024a1a41dd8dd8e4802R23) pins the signer identity (Apple Team ID fr macOS, PGP for Linux). Any reason `Verify-Installer` here only checks Authenticode Status = Valid without also verifying the signer?

scripts/install-v2/install.ps1
```diff
@@ -0,0 +1,446 @@
...
+# field - on Windows the MSI manages PATH, there's no symlink dir to record.
+function Write-InstallJson {
+    $installJson = Join-Path $Script:InstallDir 'awscli\data\install.json'
+    $parent = Split-Path $installJson -Parent
+    if (-not (Test-Path $parent)) { return }
+
+    $obj = [ordered]@{
+        distribution_source = $DistributionSource
+        install_dir         = $Script:InstallDir
+        script_install      = [ordered]@{
```
AndrewAsseily, line 382: ~~Should we also write `script_install.script_version` in `install.json`?~~

AndrewAsseily (commented):
LGTM, tested system install end-to-end on Windows Server. Should we also add a warning when both a system and user-local install coexist, and which one PATH resolves to? On Windows it seems likely users will end up with both during the transition

AndrewAsseily (approved):
Discussed offline

PR #10434, opened 2026-06-24
[Vanity URLs] - SSL verify, region update message, debug logging

AndrewAsseily:
*Description of changes:*
- Plumb SSL verify/CA bundle settings through the resolver
- Print "SSO region updated to <region>" when login updates `sso_region`
- Add debug logging for HEAD/GET requests during URL resolution

### DIFF
--- a/awscli/customizations/configure/sso_commands.py
+++ b/awscli/customizations/configure/sso_commands.py
@@ -429,7 +429,9 @@ class ConfigureSSOCommand(BaseSSOConfigurationCommand):
         if hostname and not is_aws_owned_domain(hostname):
             try:
                 resolved_url, region = resolve_start_url(
-                    start_url, session=self._session
+                    start_url,
+                    session=self._session,
+                    verify=self._session.get_config_variable('ca_bundle'),
                 )
                 self._sso_session_prompter.sso_session_config['sso_region'] = (
                     region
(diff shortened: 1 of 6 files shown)


awscli/customizations/configure/sso_commands.py
```diff
@@ -318,7 +318,9 @@ def _run_main(self, parsed_args, parsed_globals):
         on_pending_authorization = None
         if parsed_args.no_browser:
             on_pending_authorization = PrintOnlyHandler()
-        sso_registration_args = self._prompt_for_sso_registration_args()
+        sso_registration_args = self._prompt_for_sso_registration_args(
+            verify=parsed_globals.verify_ssl,
```
ashovlin: I don't think this will pick up `ca_bundle` in the config file or `AWS_CA_BUNDLE` in the environment.

If not, maybe we can reuse `session.get_config_variable('ca_bundle')` since I think that'd check the full hierarchy.

tests/unit/customizations/sso/test_resolve.py
```diff
@@ -272,7 +274,9 @@ def test_custom_timeout_passed_to_follow_redirect(self):
                 timeout=5,
             )
             mock_follow.assert_called_once_with(
-                'https://aws.mycompany.com', timeout=5
+                'https://aws.mycompany.com',
+                timeout=5,
+                verify=None,
```
ashovlin, line 280: Especially tied to the other comment, can you add a case or two that tests a non-None `verify`?

ashovlin (changes requested):

awscli/customizations/configure/sso_commands.py
AndrewAsseily: Good catch thanks!

ashovlin (approved):

PR #10443, opened 2026-06-26
fix(pagination): silent pagination truncation when native API uses PageSize or StartingToken

Abuhaithem:
Fixes #10442

This pull request fixes a critical bug where providing the global pagination parameters `--page-size` or `--starting-token` silently disabled automatic pagination for a subset of AWS API commands, resulting in silent data truncation (returning only the first page and exiting with code 0).

## Motivation and Context
The AWS CLI supports automatic pagination by injecting global parameters (`--page-size`, `--starting-token`, `--max-items`) and intentionally disabling automatic pagination if the user specifies a raw, native API pagination parameter (e.g., `--next-token` or `--marker`). 

The logic checking for this collision in `awscli.customizations.paginate.check_should_enable_pagination` had a typographical error (`start_token` instead of `starting_token`) and completely omitted `page_size`:

```python
normalized_paging_args = ['start_token', 'max_items']
```

As a result, if an API operation naturally defines its `limit_key` as exactly `"PageSize"` (rather than `"MaxResults"` or `"MaxItems"`), the raw API argument name shadows the global CLI argument. Because `page_size` was missing from `normalized_paging_args`, the CLI assumed the user was intentionally specifying the undocumented raw API parameter and disabled automatic pagination.

This silently broke automated pagination for over 50 operations across `elb`, `elbv2`, `servicecatalog`, `pinpoint-email`, `mailmanager`, `sesv2`, `ce`, and `lakeformation` (whenever `--page-size` was explicitly specified), as well as `ssm-quicksetup` (when `--starting-token` was specified).

This fix corrects the tuple to properly exempt the global parameters from disabling pagination:
```python
normalized_paging_args = ['starting_token', 'page_size', 'max_items']
```

## Testing Done
Validated via `aws elbv2 describe-load-balancers --page-size 10`, confirming that the CLI now automatically paginates and fetches all load balancers instead of aborting after the first 10 items.
Added a direct mock validation simulating the parser flow, verifying `parsed_globals.paginate` remains `True`.

## Types of Changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #10455, opened 2026-06-29
fix: resolve --query corrupting pagination resume token in text output and --page-size/--starting-token silently disabling auto-pagination

### LINKED ISSUE #10442: Silent Pagination Truncation when using `--page-size` or `--starting-token` on affected services
### Describe the bug

There is a silent data truncation bug in the AWS CLI's automatic pagination logic. When users run a paginated command that happens to have a native API `limit_key` named `"PageSize"` (e.g., `aws elbv2 describe-load-balancers`), explicitly using the global `--page-size` argument will **silently disable automatic pagination**. The CLI will fetch only the first page and exit with a `0` exit code, giving the user the false impression that they have fetched all resources in batches.

Similarly, if an API's `input_token` is exactly `"StartingToken"` (e.g., `aws ssm-quicksetup list-configurations`), using the global `--starting-token` argument will also silently disable automatic pagination.

### Regression Issue

### Expected Behavior

When using the global `--page-size` flag, the CLI should automatically fetch all items across all pages in batches of the specified size. It should not stop after the first page unless `--no-paginate` or `--max-items` is specified.

### Current Behavior

When providing `--page-size` to an operation whose underlying API natively uses `PageSize` as its pagination limit parameter, the AWS CLI mistakenly identifies the global argument as a manual, undocumented API parameter and disables automatic pagination. It retrieves exactly one page (of size equal to `--page-size`) and exits with code `0`. 

There are no errors, uncaught exceptions, or stack traces—it fails silently.

### Reproduction Steps

Observe the difference in results when using `--page-size`:

```bash
# This fetches ALL load balancers automatically via pagination.
aws elbv2 describe-load-balancers 

# EXPECTED: Fetches ALL load balancers automatically, but in batches of 1.
# ACTUAL: Fetches EXACTLY 1 load balancer, silently disables pagination, and exits with code 0!
aws elbv2 describe-load-balancers --page-size 1
```

### Possible Solution

The issue is caused by a typographical error and a missing list element in `awscli/customizations/paginate.py` within the `check_should_enable_pagination` function:

```python
    normalized_paging_args = ['start_token', 'max_items']
```

1. **Typo:** `start_token` is used instead of the correct `starting_token`.
2. **Omission:** `page_size` is completely missing from `normalized_paging_args`.

Fix the tuple to include both correctly, matching the `paging_params` list defined slightly lower in the same file:

```diff
-    normalized_paging_args = ['start_token', 'max_items']
+    normalized_paging_args = ['starting_token', 'page_size', 'max_items']
```

The following API operations are affected and will suffer from silent truncation when `--page-size` is provided:

- `ce` (Cost Explorer): `GetReservationPurchaseRecommendation`, `GetRightsizingRecommendation`, `ListCommitmentPurchaseAnalyses`, `ListSavingsPlansPurchaseRecommendationGeneration`
- `elb` / `elbv2`: `DescribeLoadBalancers`, `DescribeAccountLimits`, `DescribeTargetGroups`, `DescribeListeners`, `DescribeRules`, `DescribeSSLPolicies`, `DescribeTrustStores`, etc.
- `lakeformation`: `GetWorkUnits`
- `mailmanager`: `ListAddonInstances`, `ListArchives`, `ListRuleSets`, `ListTrafficPolicies`, etc.
- `pinpoint-email`: `GetDedicatedIps`, `ListConfigurationSets`, `ListEmailIdentities`, etc.
- `servicecatalog`: `SearchProductsAsAdmin`, `ListPortfolios`, `ListProvisionedProductPlans`, `ScanProvisionedProducts`, etc.
- `sesv2`: `ListMultiRegionEndpoints`, `ListTenants`, etc.
- `ssm-quicksetup`: `ListConfigurations` and `ListConfigurationManagers` (affected by `--starting-token` truncation).

This bug is particularly insidious for users operating large environments who explicitly use `--page-size` to prevent timeouts when retrieving thousands of ELBs or resources.

### CLI version used

aws-cli/1.x.x and aws-cli/2.x.x (Affects all recent v1 and v2 versions since the pagination customization was originally implemented)

All environments and operating systems (Linux, macOS, Windows)

### LINKED ISSUE #10449: Bug Report: `aws --output text` with `--query` prints a spurious `None` line for the pagination resume-token hint
### Describe the bug

When `--output text` is combined with `--query` and a paginated response is
**truncated** (for example via `--max-items`), the AWS CLI appends a stray
`None` line to the output.

After the last page, the text formatter prints a `NEXTTOKEN` "resume token"
hint so users know how to continue paginating. That hint is pagination
*metadata*, but it is incorrectly run through the user's `--query` expression.
Evaluating a data expression such as `Users[].UserName` against the hint object
`{"NextToken": {"NextToken": "<token>"}}` returns `None`, and the text formatter
writes the literal string `None`.

In scripts that consume `--output text --query` line by line, the extra `None`
silently corrupts the data (an additional row, a wrong count, or a bogus field
value).

### Regression Issue

### Expected Behavior

The resume-token hint should render exactly as it does **without** `--query`,
because it is not part of the queried result set:

```
alice
NEXTTOKEN	<token>
```

### Current Behavior

```
alice
None
```

The trailing `None` is the result of applying `Users[].UserName` to the
resume-token hint object instead of to the page data.

### Reproduction Steps

```bash
aws iam list-users \
    --max-items 1 \
    --output text \
    --query 'Users[].UserName'
```

Any paginated operation reproduces this; `--max-items` is simply the easiest
way to force a truncated response (and therefore a resume token). The bug
appears whenever **all** of the following hold:

1. output format is `text`,
2. `--query` is supplied, and
3. the response is truncated, so a resume token is generated.

Minimal, self-contained reproduction against mocked responses (no AWS account
or network required — mirrors the regression test):

```python
from awscli.testutils import BaseAWSCommandParamsTest


class Repro(BaseAWSCommandParamsTest):
    def test_repro(self):
        self.parsed_responses = [
            {'Users': [{'UserName': 'alice', 'Path': '/',
                        'CreateDate': '2024-01-01T00:00:00Z',
                        'UserId': 'A1', 'Arn': 'arn:aws:iam::12345:user/alice'}],
             'IsTruncated': True, 'Marker': 'PAGE2'},
            {'Users': [{'UserName': 'bob', 'Path': '/',
                        'CreateDate': '2024-01-02T00:00:00Z',
                        'UserId': 'B1', 'Arn': 'arn:aws:iam::12345:user/bob'}],
             'IsTruncated': False},
        ]
        out = self.run_cmd(
            'iam list-users --max-items 1 --output text '
            '--query Users[].UserName', expected_rc=0)[0]
        # Actual:   'alice\nNone\n'
        # Expected: 'alice\nNEXTTOKEN\t<token>\n'
        print(repr(out))
```

### Possible Solution

The bug is in `awscli/formatter.py`, `TextFormatter.__call__`:

```python
if response.resume_token:
    # Tell the user about the next token so they can continue
    # if they want.
    self._format_response(
        {'NextToken': {'NextToken': response.resume_token}}, stream)
```

`_format_response` applies `--query` to whatever it is given:

```python
def _format_response(self, response, stream):
    if self._args.query is not None:
        response = self._args.query.search(response)
    text.format_text(response, stream)
```

Render the resume-token hint directly with `text.format_text(...)`, bypassing
`--query`, so the hint is preserved for all output cases:

```python
if response.resume_token:
    text.format_text(
        {'NextToken': {'NextToken': response.resume_token}}, stream)
```

Only `TextFormatter` is affected. `FullyBufferedFormatter` (JSON and Table)
aggregates the full result with `build_full_result()` and applies `--query`
once to the entire response, so it has no separate hint emission and does not
exhibit this bug.

- Affects every paginated operation, not just `iam list-users`.
- The hint is only emitted when a resume token exists, i.e. when the response
  is truncated (commonly `--max-items`), so non-truncated commands are
  unaffected.
- Not a regression — the behavior has existed since the text formatter began
  streaming pages.

### CLI version used

`aws-cli/1.45.36` (reproduced on the `develop` branch). The same code path exists in the v2 line.

Platform independent (verified on Linux); no platform-specific behavior is involved.

Hardikrepo:
## Summary

- **Fix #10449** — `aws --output text` with `--query` was printing a spurious `None` line when a paginated response was truncated. The `TextFormatter` was passing the pagination resume-token hint `{'NextToken': {'NextToken': <token>}}` through `_format_response()`, which applied the user's `--query` JMESPath expression to it. Any data query (e.g. `Users[].UserName`) evaluates to `None` against that metadata dict and the literal string `None` was written to stdout, silently corrupting script output.

- **Fix #10442** — Using `--page-size` against APIs whose native `limit_key` is named `PageSize` (e.g. `aws elbv2 describe-load-balancers`), or `--starting-token` against APIs whose native `input_token` is named `StartingToken` (e.g. `aws ssm-quicksetup list-configurations`), silently disabled automatic pagination and returned only the first page with exit code 0. The root cause was that `check_should_enable_pagination()` only had `['start_token', 'max_items']` in `normalized_paging_args`, so `page_size` and `starting_token` were misidentified as user-specified native API args, triggering `--no-paginate`.

### `awscli/formatter.py`
- In `TextFormatter.__call__`, write the resume-token hint directly via `text.format_text()` instead of routing it through `_format_response()`, so the `--query` filter is never applied to pagination metadata.

### `awscli/customizations/paginate.py`
- Added `page_size` and `starting_token` to `normalized_paging_args` in `check_should_enable_pagination()` so the unified CLI pagination args are never mistaken for native API pagination params.

## Test plan

- [ ] `aws iam list-users --max-items 1 --output text --query 'Users[].UserName'` — confirm no `None` line appears after the username, and `NEXTTOKEN` line is still printed
- [ ] `aws elbv2 describe-load-balancers --page-size 5` — confirm all load balancers are returned (not just first page)
- [ ] `aws ssm-quicksetup list-configurations --starting-token <token>` — confirm pagination resumes correctly and auto-pagination is not disabled
- [ ] Existing pagination unit tests pass

### DIFF
--- a/awscli/customizations/paginate.py
+++ b/awscli/customizations/paginate.py
@@ -194,7 +194,7 @@ def add_paging_argument(argument_table, arg_name, argument, shadowed_args):
 
 def check_should_enable_pagination(input_tokens, shadowed_args, argument_table,
                                    parsed_args, parsed_globals, **kwargs):
-    normalized_paging_args = ['start_token', 'max_items']
+    normalized_paging_args = ['start_token', 'max_items', 'page_size', 'starting_token']
     for token in input_tokens:
         py_name = token.replace('-', '_')
         if getattr(parsed_args, py_name) is not None and \
--- a/awscli/formatter.py
+++ b/awscli/formatter.py
@@ -247,8 +247,9 @@ class TextFormatter(Formatter):
                     self._format_response(current, stream)
                 if response.resume_token:
                     # Tell the user about the next token so they can continue
-                    # if they want.
-                    self._format_response(
+                    # if they want. Write directly to bypass any --query filter
+                    # since this is pagination metadata, not response data.
+                    text.format_text(
                         {'NextToken': {'NextToken': response.resume_token}},
                         stream,
                     )
--- a/tests/unit/customizations/test_paginate.py
+++ b/tests/unit/customizations/test_paginate.py
@@ -309,6 +309,38 @@ class TestShouldEnablePagination(TestPaginateBase):
         self.assertFalse(self.parsed_globals.paginate)
         self.assertEqual(arg_table['foo'], mock.sentinel.ORIGINAL_ARG)
 
+    def test_page_size_does_not_disable_pagination_when_api_limit_key_is_page_size(self):
+        # Regression test for #10442: when an API's native limit_key is named
+        # "PageSize", the CLI token becomes "page-size" which clashes with our
+        # unified --page-size arg.  Using --page-size must NOT disable
+        # auto-pagination.
+        input_tokens = ['page-size']
+        self.parsed_globals.paginate = True
+        self.parsed_args.page_size = 10
+        paginate.check_should_enable_pagination(
+            input_tokens, {}, {}, self.parsed_args, self.parsed_globals)
+        self.assertTrue(
+            self.parsed_globals.paginate,
+            "--page-size incorrectly disabled auto-pagination when API "
+            "limit_key is named 'PageSize'",
+        )
+
+    def test_starting_token_does_not_disable_pagination_when_api_input_token_is_starting_token(self):
+        # Regression test for #10442: when an API's native input_token is
+        # named "StartingToken", the CLI token becomes "starting-token" which
+        # clashes with our unified --starting-token arg.  Using
+        # --starting-token must NOT disable auto-pagination.
+        input_tokens = ['starting-token']
+        self.parsed_globals.paginate = True
+        self.parsed_args.starting_token = 'abc123'
+        paginate.check_should_enable_pagination(
+            input_tokens, {}, {}, self.parsed_args, self.parsed_globals)
+        self.assertTrue(
+            self.parsed_globals.paginate,
+            "--starting-token incorrectly disabled auto-pagination when API "
+            "input_token is named 'StartingToken'",
+        )
+
 
 class TestPaginateV2Debug(TestPaginateBase):
     def setUp(self):
--- a/tests/unit/output/test_text_output.py
+++ b/tests/unit/output/test_text_output.py
@@ -20,7 +20,9 @@ import sys
 import re
 import locale
 
-from awscli.formatter import Formatter
+from botocore.paginate import PageIterator
+
+from awscli.formatter import Formatter, TextFormatter
 
 
 class TestListUsers(BaseAWSCommandParamsTest):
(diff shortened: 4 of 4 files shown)


Abuhaithem:
I've already opened a PR for this issues (#10442 , #10449) before you opened yours.

You can find it here: #10443, #10450.

Could you please close your PR?

Also, for future contributions, please check whether an issue has already been claimed or already has an open PR before creating a new one. Additionally, please avoid combining fixes for multiple issues into a single PR—it's better to keep one PR per issue. Thanks!

Hardikrepo:
Thanks for flagging! I checked #10443 and #10450 — both are still open/unmerged, and #10449 and #10442 are actually assigned to a maintainer (@RyanFitzSimmonsAK), not claimed by either of us. Since there's no merged fix yet, I'll leave this PR open for maintainer review and keep it as a single combined PR rather than splitting it, since both fixes touch closely related pagination-metadata handling. Happy to split if a maintainer prefers that.

PR #10462, opened 2026-07-02
[v2] Add redirect port option to sso login

### LINKED ISSUE #10433: sso login: allow to customize localhost port for redirect_uri
### Describe the feature

I would like to propose a feature that allows specifying a fixed callback port for the PKCE callback server used by `aws sso login`.

For example:

```sh
aws sso login --redirect-port 34535
```

When specified, the PKCE callback server would listen on the given port and use:

```text
http://127.0.0.1:34535/oauth/callback
```

as the `redirect_uri` instead of selecting a random available port.

When the option is not specified, the current behavior of automatically selecting an available port would remain unchanged.

### Use Case

I frequently run `aws sso login` on remote environments accessed over SSH.

While `--use-device-code` is available as a workaround, I would prefer to continue using the default PKCE-based login flow.

The PKCE flow provides a more streamlined browser experience and requires fewer manual interactions. In contrast, the device code flow requires entering a verification code and following a different authentication process.

The main issue is that the PKCE callback server uses a randomly selected localhost port, which makes it difficult to prepare SSH local port forwarding ahead of time.

For example, I would like to configure:

```sh
ssh -L 34535:127.0.0.1:34535 remote-host
```

and then use the standard `aws sso login` flow over that tunnel.

A fixed callback port would allow SSH users to keep using the default PKCE experience while retaining the ability to preconfigure SSH port forwarding.

### Proposed Solution

Add a new option to allow users to specify the callback port used by the local PKCE callback server.

Example:

```sh
aws sso login --redirect-port 34535
```

If the specified port is unavailable, the command could fail with an appropriate error message rather than automatically selecting a different port.

This would provide deterministic behavior for SSH tunnels and remote development environments while preserving the current behavior when the option is not specified.

### Other Information

The current workaround is:

```sh
aws sso login --use-device-code
```

While this works, it requires additional manual steps in the browser, including entering a verification code.

A fixed callback port would allow users to continue using the default PKCE-based login flow when working over SSH, rather than switching to a different authentication experience.

Related issue:

* #9148 (custom redirect host)

This request focuses specifically on allowing a fixed callback port.

Additionally, it might be useful to support configuring the callback port through `~/.aws/config` (for example, at the `sso-session` level) so that users who consistently work in remote environments do not need to specify the option every time.

For example:

```ini
[sso-session my-sso]
sso_redirect_port = 34535
```

### Acknowledgements

- [x] I may be able to implement this feature request

### CLI version used

aws-cli/2.34.24 Python/3.13.13 Darwin/25.5.0 source/arm64

OS: macOS 26.5.1 (Build 25F80), Architecture: arm64 (Apple Silicon)

ixxq:
*Issue #, if available:* Fixes #10433

*Description of changes:* Adds support for specifying a fixed localhost port for the PKCE callback server used by `aws sso login`.

Users can set the port per command with `--redirect-port`, or configure it on the SSO session with `sso_redirect_port`. When neither value is provided, the existing behavior is preserved and the callback server continues to bind to a randomly selected available port.

The CLI option takes precedence over `sso_redirect_port`. Invalid port values are rejected before starting the login flow.

Added unit and functional test coverage for the CLI option, SSO session configuration, CLI precedence over config, invalid port values, and binding the callback server to the requested port.

Validated with targeted pytest, ruff, and pre-commit checks.

### DIFF
--- a/awscli/customizations/sso/login.py
+++ b/awscli/customizations/sso/login.py
@@ -13,7 +13,10 @@
 import os
 
 from awscli.customizations.configure.writer import ConfigFileWriter
-from awscli.customizations.exceptions import ParamValidationError
+from awscli.customizations.exceptions import (
+    ConfigurationError,
+    ParamValidationError,
+)
 from awscli.customizations.sso.resolve import resolve_start_url
 from awscli.customizations.sso.utils import (
     LOGIN_ARGS,
@@ -22,6 +25,8 @@ from awscli.customizations.sso.utils import (
     do_sso_login,
 )
 from awscli.customizations.utils import uni_print
+
+SSO_REDIRECT_PORT_CONFIG_VAR = 'sso_redirect_port'
 
 
 class LoginCommand(BaseSSOCommand):
@@ -58,8 +63,10 @@ class LoginCommand(BaseSSOCommand):
     ]
 
     def _run_main(self, parsed_args, parsed_globals):
-        self._validate_redirect_port(parsed_args.redirect_port)
         sso_config = self._get_sso_config(sso_session=parsed_args.sso_session)
+        redirect_port = self._resolve_redirect_port(
+            parsed_args.redirect_port, sso_config
+        )
         start_url = sso_config['sso_start_url']
         configured_region = sso_config.get('sso_region')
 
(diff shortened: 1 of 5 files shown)


ixxq:
Hi @ashovlin, I've updated this PR as discussed in #10636:

- Added `--redirect-port` support to `aws login` and `aws configure sso`.
- Removed `sso_redirect_port` configuration support for now.

I've marked it ready for review. Could you take a look when you have a chance?

I also saw the callback binding and error-message comments on #10636. I've left those existing behaviors unchanged for now. Would you prefer to address those concerns in this PR or separately?

Thanks!

ashovlin (approved):

ixxq:
@ashovlin Thanks for the review! I'm on board—let's ship it.

As for a follow-up, I’m keen on adding configuration support for the redirect port. I realize this will involve some coordination with other tools, but I’d be happy to help out once we’ve settled on how it should behave.

PR #10468, opened 2026-07-03
Fix shorthand syntax error location rendering

### LINKED ISSUE #10467: BUG: Shorthand syntax error messages print literal `{self._error_location()}` instead of the error location
### Describe the bug

When a shorthand parameter value fails to parse, the AWS CLI is supposed to
print the offending input with a caret (`^`) pointing at the exact character
where parsing failed. Instead, the error message prints the literal text
`{self._error_location()}`.

The cause is in `awscli/shorthand.py`,
`ShorthandParseSyntaxError._construct_msg`: the method concatenates an
f-string with a plain string literal, and the final fragment is missing the
`f` prefix, so the placeholder is never interpolated:

```python
def _construct_msg(self):
    return (
        f"Expected: '{self.expected}', received: '{self.actual}' "
        f"for input:\n" "{self._error_location()}"
    )
```

The sibling exception `DuplicateKeyInObjectError._construct_msg` was
converted to f-strings correctly in the same commit, which is why only
`ShorthandParseSyntaxError` is affected.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

```
$ aws dynamodb put-item --table-name t --item 'a=b,c==d'

Error parsing parameter '--item': Expected: ',', received: '=' for input:
a=b,c==d
      ^
```

This is the output produced by AWS CLI v1 releases before August 2024.

### Current Behavior

```
$ aws dynamodb put-item --table-name t --item 'a=b,c==d'

Error parsing parameter '--item': Expected: ',', received: '=' for input:
{self._error_location()}
```

The same broken output is produced for every command whose shorthand value
raises `ShorthandParseSyntaxError` (`--tags`, `--item`, `--attributes`,
`--environment`, etc.). For long shorthand expressions the caret is the main
diagnostic for locating the mistake, so its loss makes these errors much
harder to act on.

### Reproduction Steps

Any invalid shorthand value reproduces it; no credentials or API calls are
needed since parsing happens client-side:

```
aws dynamodb put-item --table-name t --item 'a=b,c==d'
```

Or directly against the parser:

```python
from awscli.shorthand import ShorthandParser
ShorthandParser().parse('a=b,c==d')
# ShorthandParseSyntaxError: Expected: ',', received: '=' for input:
# {self._error_location()}
```

### Possible Solution

Move the placeholder into the f-string fragment in
`ShorthandParseSyntaxError._construct_msg`:

```python
return (
    f"Expected: '{self.expected}', received: '{self.actual}' "
    f"for input:\n{self._error_location()}"
)
```

This restores the pre-regression message byte-for-byte. The existing unit
tests only assert the exception *type* (`test_error_parsing` in
`tests/unit/test_shorthand.py`), which is why this went unnoticed; a fix
should also add tests asserting the full message content, including the
caret position.

- `DuplicateKeyInObjectError` messages (duplicate shorthand keys) still
  render correctly and are unaffected.
- AWS CLI v2 maintains its own copy of `shorthand.py` on the `v2` branch and
  is not affected.
- Searched existing issues/PRs for `_error_location`,
  `ShorthandParseSyntaxError`, and related terms; no existing report found.
  Issue #7502 mentions a shorthand error message but concerns a different
  parsing bug (filed 2022, before this regression).

This report was generated with the assistance of AI tooling and reviewed by
Abuhaithem, per the repository's Automated Tools contribution guidelines.

### CLI version used

aws-cli/1.45.40 (current `develop`); regression present since 1.34.x

All platforms (message construction is platform-independent); verified on Linux (kernel 6.17), Python 3.12

Abuhaithem:
Fixes #10467

Since commit 9a2e0478 ("Format base *.py files at top level awscli
directory", August 2024), shorthand syntax errors have printed the literal
text `{self._error_location()}` instead of the input expression annotated
with a caret pointing at the parse failure:

```
$ aws dynamodb put-item --table-name t --item 'a=b,c==d'

Error parsing parameter '--item': Expected: ',', received: '=' for input:
{self._error_location()}
```

The f-string conversion in that commit left the final fragment of
`ShorthandParseSyntaxError._construct_msg` without the `f` prefix, so the
placeholder inside it was emitted verbatim. This affects every command whose
shorthand value fails to parse (`--tags`, `--item`, `--attributes`,
`--environment`, ...). The sibling `DuplicateKeyInObjectError` was converted
correctly and is unaffected.

This PR adds the missing `f` prefix, restoring the exact pre-regression
message:

```
Error parsing parameter '--item': Expected: ',', received: '=' for input:
a=b,c==d
      ^
```

## Changes

- `awscli/shorthand.py` — move the `{self._error_location()}` placeholder
  into the f-string fragment (one line).
- `tests/unit/test_shorthand.py` — add
  `test_error_message_includes_error_location`, parametrized over a
  mid-string syntax error, an EOF syntax error, and the duplicate-key error.
  Each case asserts the complete message, including the caret position. The
  pre-existing `test_error_parsing` only asserts the exception type, which
  is the coverage gap that allowed this to regress silently; these cases
  close it for both shorthand error types.
- `.changes/next-release/bugfix-shorthand-error-location.json` — changelog
  entry.

## Testing

Per CONTRIBUTING.md, the bug fix includes tests:

- `tests/unit/test_shorthand.py`: 117 passed (includes the 3 new regression
  cases)
- `tests/unit/test_argprocess.py`: 65 passed (shorthand error surfacing
  through `ParamError`)
- Verified end-to-end through `bin/aws` that the caret diagnostic is
  restored (no credentials required; parsing fails client-side)

No parsing behavior changes; only the error-message string on an
error-only path is affected. The restored text is identical to the output
of all releases prior to August 2024.

This pull request was generated with the assistance of AI tooling and
reviewed by Abuhaithem, per the repository's Automated Tools contribution
guidelines.

### DIFF
--- a/.changes/next-release/bugfix-shorthand-error-location.json
+++ b/.changes/next-release/bugfix-shorthand-error-location.json
@@ -1,7 +1,5 @@
-[
-  {
-    "category": "Shorthand",
-    "description": "Fix an issue where shorthand syntax error messages printed the literal text ``{self._error_location()}`` instead of the input expression annotated with a caret pointing at the location of the syntax error.",
-    "type": "bugfix"
-  }
-]
+{
+  "category": "Shorthand",
+  "description": "Fix an issue where shorthand syntax error messages printed the literal text ``{self._error_location()}`` instead of the input expression annotated with a caret pointing at the location of the syntax error.",
+  "type": "bugfix"
+}
--- a/.changes/next-release/bugfix-shorthand-error-location.json
+++ b/.changes/next-release/bugfix-shorthand-error-location.json
@@ -1,0 +1,7 @@
+[
+  {
+    "category": "Shorthand",
+    "description": "Fix an issue where shorthand syntax error messages printed the literal text ``{self._error_location()}`` instead of the input expression annotated with a caret pointing at the location of the syntax error.",
+    "type": "bugfix"
+  }
+]
--- a/awscli/shorthand.py
+++ b/awscli/shorthand.py
@@ -93,7 +93,7 @@ class ShorthandParseSyntaxError(ShorthandParseError):
     def _construct_msg(self):
         return (
             f"Expected: '{self.expected}', received: '{self.actual}' "
-            f"for input:\n" "{self._error_location()}"
+            f"for input:\n{self._error_location()}"
         )
 
 
--- a/tests/unit/test_shorthand.py
+++ b/tests/unit/test_shorthand.py
@@ -175,6 +175,38 @@ def test_error_parsing(expr):
 
 
 @pytest.mark.parametrize(
+    "expr, expected_message", (
+        # The error message must include the original input with a
+        # caret pointing at the location of the syntax error.
+        (
+            'a=b,c==d',
+            "Expected: ',', received: '=' for input:\n"
+            "a=b,c==d\n"
+            "      ^",
+        ),
+        (
+            'a={b',
+            "Expected: '=', received: 'EOF' for input:\n"
+            "a={b\n"
+            "    ^",
+        ),
+        (
+            'foo=bar,foo=baz',
+            'Second instance of key "foo" encountered for input:\n'
+            'foo=bar,foo=baz\n'
+            '        ^\n'
+            'This is often because there is a preceding "," instead '
+            'of a space.',
+        ),
+    )
+)
+def test_error_message_includes_error_location(expr, expected_message):
+    with pytest.raises(shorthand.ShorthandParseError) as excinfo:
+        shorthand.ShorthandParser().parse(expr)
+    assert str(excinfo.value) == expected_message
+
+
+@pytest.mark.parametrize(
     "expr", (
         # starting with " but unclosed, then repeated \
         f'foo="' + '\\' * 100,


.changes/next-release/bugfix-shorthand-error-location.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "category": "Shorthand",
+    "description": "Fix an issue where shorthand syntax error messages printed the literal text ``{self._error_location()}`` instead of the input expression annotated with a caret pointing at the location of the syntax error.",
+    "type": "bugfix"
+  }
+]
```
ashovlin: This should just be a single object in `{}`, not an array.

Tests are failing with 
```
   def test_next_release_json_is_well_formed(json_path):
        raw = json_path.read_text('utf-8')
        try:
            entry = json.loads(raw)
        except json.JSONDecodeError as e:
            pytest.fail(f"{json_path.name} is not valid JSON: {e}")
>       assert isinstance(entry, dict), (
            f"{json_path.name} should be a JSON object, "
            f"got {type(entry).__name__}"
        )
E       AssertionError: bugfix-shorthand-error-location.json should be a JSON object, got list
E       assert False
E        +  where False = isinstance([{'category': 'Shorthand', 'description': 'Fix an issue where shorthand syntax error messages printed the literal text...stead of the input expression annotated with a caret pointing at the location of the syntax error.', 'type': 'bugfix'}], dict)

```

ashovlin (changes requested):
Just one comment on the changelog entry, the actual change looks good.

ashovlin (approved):

PR #10475, opened 2026-07-07
[v2] Implement update command

hssyoo:
Implements `aws update` command to download and install the latest AWS CLI v2 version. Also change Linux and macOS post-install scripts to write install-time metadata at `install.json`.

At a high-level, the command does the following:
1. Look for install-time metadata in `install.json` to see if there's a record of the exact install and bin path of the current CLI instance.
  a. Unix should always have it since this PR also introduces post-install changes that ensure `install.json` is written. One caveat is that the bin directory may not have been recorded (if the user manually set up symlinks). In that case, the command skips setting up a symlink. This is fine since the user-created symlink will still point to the install directory that we're writing over.
  b. Windows may not necessarily have it because there's no Windows post-install script that creates one in the case that the MSI was used directly (ie without an install script or `update` command). This is fine because Windows MSIs install to predetermined paths.
2. Check `install.json` to see if the current instance was a system install. If unavailable, then a best effort will be made based on the current CLI path.
3. Download the relevant install script and execute it using parameters determined from previous steps.
  a. Unix updates are straightforward since the OS caches the inode and so it can be updated in-place while the updated exe is still running.
  b. Windows is trickier since the process acquires a lock on the exe. To work around this, the parent process exits early after launching a detached subprocess that handles the install execution.

### DIFF
--- a/awscli/customizations/update.py
+++ b/awscli/customizations/update.py
@@ -153,7 +153,9 @@ class UnixUpdateCommand(BaseUpdateCommand):
     def _is_system_install(self, install_metadata):
         if 'script_install' in install_metadata:
             return install_metadata['script_install'].get('system', False)
-        aws_bin = os.path.abspath(sys.argv[0])
+        if not sys.executable:
+            return False
+        aws_bin = os.path.realpath(sys.executable)
         return aws_bin.startswith(self.SYSTEM_INSTALL_DIR + os.sep)
 
     def _assert_elevated(self):
--- a/awscli/customizations/update.py
+++ b/awscli/customizations/update.py
@@ -1,0 +1,277 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+import ctypes
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+
+from awscli.botocore.awsrequest import AWSRequest
+from awscli.botocore.httpsession import URLLib3Session
+from awscli.clidriver import (
+    INSTALL_FILENAME,
+    get_distribution_source,
+)
+from awscli.compat import is_windows
+from awscli.customizations.commands import BasicCommand
+from awscli.customizations.utils import uni_print
+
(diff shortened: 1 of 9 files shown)


awscli/customizations/update.py
```diff
@@ -0,0 +1,277 @@
...
+                self._run_update(cmd, env)
+            except subprocess.CalledProcessError as exc:
+                raise UpdateError(
+                    f"install script failed with exit code {exc.returncode}"
+                )
+
+    def _run_install(self, cmd, env):
+        subprocess.run(cmd, env=env, check=True)
+
+    def _is_system_install(self, install_metadata):
```
AndrewAsseily, line 153: This branches on script_install, but the exe installer writes an install.json with only install_dir/bin_dir then official installs have an install.json but always fall through to the argv[0] fallback

The installer symlinks `/usr/local/bin/aws` -> `/usr/local/aws-cli/v2/current/bin/aws` ([exe/assets/install](https://github.com/aws/aws-cli/pull/10475/changes#diff-08c4eae673d091c9672dd11fa39dcb04ce54d3a2de1af399f9309ff9da2eb90bR135)) and that's on PATH so `sys.argv[0]` is the symlink. `os.path.abspath` doesn't resolve symlinks, so it stays `/usr/local/bin/aws` and the `startswith('/usr/local/aws-cli/')` check returns False

hssyoo: Fixed in https://github.com/aws/aws-cli/pull/10475/changes/26faac4c3f6d558016fa5de69a6d6afe2d83b6f8

AndrewAsseily: Thanks! I verified in a Linux container. 

Not-blocking: Can we add a test that points `sys.executable` at a symlink into the install dir?

hssyoo: Added in https://github.com/aws/aws-cli/pull/10475/changes/6e89d6a7b398eeb6d910e2550cd79d031539d3e4

AndrewAsseily (approved):

Issue #10476, opened 2026-07-08
Add configuration options to support PRM APN_ID tagging


elliotsegler:
### Describe the feature

This feature adds configuration options to allow users to specify an APN Martketplace Product code, which is included in the user agent string on CLI calls that matches the format required for PRM compliance.

### Use Case

As part of the changes for AWS Partners for PRM attribution (See: https://docs.aws.amazon.com/PRM/latest/aws-prm-onboarding-guide/user-agent-string.html) we now need ways to be able to tag resources as well as API calls for certain services.

This is possible with custom clients and with botocore directly, but the CLI uses guards to protect data sent in the user agent string. The PRM instructions in the docs suggests using the existing botocore `sdk_ua_app_id` option, however the format required is mangled because it contains it's own prefix.

### Proposed Solution

This solution adds a global CLI argument `--apn-id` and matching config option, as well as support for using an environment variable `AWS_APN_ID` to set a marketplace product code (or revenue attribution id) for use in tagging.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CLI version used

2.32.34

### Environment details (OS name and version, etc.)

 Python/3.13.11 Windows/11 exe/AMD64

ashovlin:
Is the `app/` prefix from `sdk_ua_app_id` problematic? I see we recommend it on https://docs.aws.amazon.com/PRM/latest/aws-prm-onboarding-guide/user-agent-implementation.html, but if that's not working we could update the docs too.

I'll chat about `AWS_APN_ID` with the other SDK teams internally, since generally we try to ensure `AWS_` variables work across the SDKs.

elliotsegler:
yep, `sdk_ua_app_id` prefixes the user agent with  `App/` and then applies sanitization thats incompatible with the PRM formatted string.

for boto you can use user_agent_extra but thats not possible for the cli

PR #10478, opened 2026-07-08
[v2] Add --apn-id global arg for PRM user-agent attribution

elliotsegler:
*Issue #, if available:*

Fixes aws/aws-cli#10476

*Description of changes:*

Add a new --apn-id global CLI argument that appends an AWS Partner Network identifier to the User-Agent header in the format APN_1.1/<value>$ as required by the PRM onboarding guide (ref: https://docs.aws.amazon.com/PRM/latest/aws-prm-onboarding-guide/user-agent-string.html).

The user provides the full identifier including the type prefix (e.g. pc_PRODUCTCODE for product codes, ra_PARTNERID for referral agents) to support current and future APN ID types.

The value is resolved from (highest precedence first):
  1. --apn-id CLI flag
  2. AWS_APN_ID environment variable
  3. apn_id setting in the shared config file profile

The handler is registered on the session-initialized event rather than top-level-args-parsed to ensure the profile has been applied to the session before reading config file values.

Validation uses a simple character set check (no regex) to avoid any ReDoS concerns: 1-255 ASCII alphanumeric chars plus _ and -.

### DIFF
--- a/awscli/customizations/globalargs.py
+++ b/awscli/customizations/globalargs.py
@@ -42,6 +42,11 @@ def register_parse_global_args(cli):
         resolve_cli_connect_timeout,
         unique_id='resolve-cli-connect-timeout',
     )
+    cli.register(
+        'session-initialized',
+        resolve_apn_id,
+        unique_id='resolve-apn-id',
+    )
 
 
 def resolve_types(parsed_args, **kwargs):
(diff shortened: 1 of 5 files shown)


fordrich-resilera:
Vote!

PR #10481, opened 2026-07-09
Best-effort auto-close of the SSO login browser tab

### LINKED ISSUE #9585: AWS SSO login - provide an option for auto-close tab
### Describe the feature

- Python HVAC library shows how an auto-close option could be used to ensure that
  the callback is auto-closing if desired.
- could we have the same as an option to the command 'aws sso login --profile X --auto-close-enabled'

### Use Case

After extensively using SSO for several years, we are ending up with multiple tabs in the browser that have been used to authenticate an AWS CLI session in the terminal. Thus we are looking for an option being added to the CLI where a user can inject an auto-close procedure.

https://python-hvac.org/en/stable/usage/auth_methods/jwt-oidc.html#oidc-authorization-url-request

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CLI version used

aws-cli/2.27.46 Python/3.13.5 Darwin/24.5.0 source/arm64

MacOS 15.5

saipavan9:
## Summary

Closes #9585.

After a successful `aws sso login` Authorization Code flow, the local callback page now makes a **best-effort** attempt to close its own browser tab, addressing the tab buildup that several users hit when authenticating frequently.

The change is intentionally small and lives entirely in the served callback page.

## Behavior and caveat

Browsers only permit `window.close()` on a tab that was **opened by script** or that has a **single session-history entry**. `aws sso login` opens the browser via the OS default handler (`webbrowser.open_new_tab`), so:

- **Warm re-auth** (a pure redirect chain, single history entry) → the tab closes automatically. ✅
- **Interactive login** (sign-in form, MFA, "Allow access") → the tab has more than one history entry, so `window.close()` is a **silent no-op**. The page falls back to the existing success message, now reworded to tell the user they can close the tab themselves.

This matches how other CLIs behave (`gcloud`, `az login`, `gh auth` all display a "you may close this window" message rather than reliably closing the OS-opened tab), so the change is additive and cannot regress the current experience. The wording was deliberately made truthful ("This tab will try to close automatically, if it is still open, you can safely close it now") to avoid implying a guarantee the browser may not honor.

A code comment documents the constraint so the `window.close()` call isn't mistaken for dead code later.

## Changes

- `awscli/customizations/sso/index.html`: attempt `window.close()` on a successful (non-error) callback. Reword the success message.
- `tests/unit/customizations/sso/test_utils.py`:  regression test asserting the served callback page includes the close attempt.
- `.changes/next-release/enhancement-sso-4703.json`:  changelog entry.

## Testing

```
python -m pytest tests/unit/customizations/sso/test_utils.py tests/functional/sso/test_login.py
```

All tests pass; `ruff check` is clean.

### DIFF
--- a/.changes/next-release/enhancement-sso-4703.json
+++ b/.changes/next-release/enhancement-sso-4703.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "sso",
+  "description": "The browser tab opened by ``aws sso login`` for the Authorization Code flow now makes a best-effort attempt to close itself after a successful login. This is a no-op in browsers that block it (e.g. after an interactive login), where the page continues to display a message that the tab can be closed."
+}
--- a/awscli/customizations/sso/index.html
+++ b/awscli/customizations/sso/index.html
@@ -121,8 +121,8 @@
                     </path>
                 </svg>
                 <div class="success-text">
-                    Your credentials have been shared successfully and can be used until your session expires. You can
-                    now close this tab.
+                    Your credentials have been shared successfully and can be used until your session expires. This tab
+                    will try to close automatically&mdash;if it is still open, you can safely close it now.
                 </div>
             </div>
             <div id="error-message" class="error-content hidden">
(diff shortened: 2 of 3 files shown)


benhurandrade8-prog:
Screenshot_20260724-185738

PR #10494, opened 2026-07-17
Fix ShorthandParseSyntaxError message showing literal placeholder instead of caret

mangodxd:
The _construct_msg method in ShorthandParseSyntaxError was missing an  prefix on the last string fragment, so {self._error_location()} was printed literally instead of showing the caret. This made shorthand parsing errors much harder to diagnose — the carets were the main way to locate the mistake in long expressions.

Also adds a test asserting the message content includes both the input and the caret.


ashovlin:
Closing in favor of https://github.com/aws/aws-cli/pull/10468.

PR #10497, opened 2026-07-17
Error on empty CA bundle value

alexgromero:
*Issue #, if available:*

*Description of changes:*
The CLI now errors when a CA bundle value (`--ca-bundle`, the `ca_bundle` profile setting, `AWS_CA_BUNDLE`, or `REQUESTS_CA_BUNDLE`) resolves to an empty string.

An empty string is not a valid value for these settings, which are defined as paths to a CA bundle.

### DIFF
--- a/awscli/botocore/endpoint.py
+++ b/awscli/botocore/endpoint.py
@@ -21,7 +21,7 @@ import uuid
 
 from botocore import parsers
 from botocore.awsrequest import create_request_object
-from botocore.exceptions import HTTPClientError
+from botocore.exceptions import HTTPClientError, InvalidConfigError
 from botocore.history import get_global_history_recorder
 from botocore.hooks import first_non_none_response
 from botocore.httpchecksum import handle_checksum_body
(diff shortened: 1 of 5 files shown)


awscli/botocore/endpoint.py
```diff
@@ -441,7 +441,21 @@ def _get_verify_value(self, verify):
...
+            return self._validate_verify_value(verify)
         # Otherwise use the value from REQUESTS_CA_BUNDLE, or default to
         # True if the env var does not exist.
-        return os.environ.get('REQUESTS_CA_BUNDLE', True)
+        return self._validate_verify_value(
+            os.environ.get('REQUESTS_CA_BUNDLE', True)
+        )
+
+    def _validate_verify_value(self, verify):
+        if isinstance(verify, str) and not verify:
```
AndrewAsseily: Nit: `not verify` is only true for ''. `not verify.strip()` is true for '' and any all-whitespace strings (' ', '\t', '\n') if we want an `InvalidConfigError` for those too

AndrewAsseily (commented):
Looks good to me! Just left a small nit

awscli/botocore/endpoint.py
alexgromero: Good point! Added in https://github.com/aws/aws-cli/pull/10497/commits/16905aa4ed7973e017ad99664092866c180867d2.

AndrewAsseily (approved):
🏆

PR #10505, opened 2026-07-22
Agent Toolkit installation after setup moments

AndrewAsseily:
*Description of changes:* Adds a `[y/n/never]` prompt to run `aws configure agent-toolkit` after `aws configure`, `aws configure sso`, and a first-time `aws login` that creates a new profile. `aws update` and the install scripts print a static tip instead. The prompt only appears on a TTY when a supported agent is detected and no AWS skills are installed yet, and is suppressed by answering `never` or setting `AWS_CLI_AGENT_TOOLKIT_HINT_DISABLED=true`.


awscli/customizations/prompts.py
```diff
@@ -35,6 +35,27 @@ def yes_no_choice(prompt):
...
+    Prompts the user with a yes/no/never question.
+    Continually re-prompts for invalid selections.
+
+    :param prompt: Prompt text.
+    :returns: 'yes', 'no', or 'never'.
+    """
+    while True:
+        response = compat_input(prompt)
+
+        if response.lower() in ('y', 'yes') or response == '':
```
ashovlin: Can we collect a response explicitly, rather than defaulting to `yes`?

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,109 @@
...
+from awscli.utils import is_stdin_a_tty
+
+LOG = logging.getLogger(__name__)
+
+STATE_PATH = '~/.aws/cli/agent-toolkit/state.json'
+
+PROMPT_TEXT = (
+    'Configure AWS skills and the AWS MCP server for your AI coding '
+    'agent(s)? [Y/n/never]: '
+)
```
ashovlin, line 64:
```
CLI default output format (json if not specified) [None]:
Profile name [AdministratorAccess-1234567890]:
To use this profile, specify the profile name using --profile, as shown:

aws sts get-caller-identity --profile AdministratorAccess-1234567890
Configure AWS skills and the AWS MCP server for your AI coding agent(s)? [Y/n/never]: never
```

This feels a little cluttered to me, can we add a new line in front of the hint?

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,109 @@
...
+def _is_eligible(detected_agents):
+    if not is_stdin_a_tty():
+        return False
+    if _load_state().get('hint_dismissed'):
+        return False
+    if not detected_agents:
+        return False
+    if _has_installed_skills(detected_agents):
+        return False
+    return True
```
ashovlin: I think we had discussed, why not a environment variable here? That would be easier for someone to opt out in advance, without needing to go through the wizard at least once. It'd also let someone set it in their `.zshrc`/etc. files, which may be copied or persisted across machines.

tests/unit/customizations/login/test_login.py
```diff
@@ -0,0 +1,66 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ashovlin: Can you move this to functional tests?  It has a lot of the similar setup, and is where I'd expect tests that are calling the full `LoginCommand`

awscli/customizations/prompts.py
AndrewAsseily: Addressed in revision

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Revised

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Added `AWS_CLI_AGENT_TOOLKIT_HINT_DISABLED=true`. went with a `_DISABLED` name since the feature is on by default. Considering `AWS_CLI_AGENT_TOOLKIT_HINT=off` similar to `AWS_CLI_AUTO_PROMPT` if you prefer.

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def maybe_prompt_agent_toolkit(session, parsed_globals, stream=None):
+    try:
+        detected_agents = get_detected_real_agents()
+        if not _is_eligible(detected_agents):
```
ashovlin: Can we detect agents in `_is_eligible`? The tty, env var and state checks should be cheaper than the agent check, so if those are False it'll be faster.

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,115 @@
...
+    if _load_state().get('hint_dismissed'):
+        return False
+    if not detected_agents:
+        return False
+    if _has_installed_skills(detected_agents):
+        return False
+    return True
+
+
+def maybe_prompt_agent_toolkit(session, parsed_globals, stream=None):
```
ashovlin: `stream` looks unused? If so drop it.

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,115 @@
...
+        choice = yes_no_never_choice(PROMPT_TEXT)
+        if choice == 'never':
+            _dismiss_forever()
+            return
+        if choice == 'no':
+            return
+        command = ConfigureAgentToolkitCommand(session)
+        command([], parsed_globals)
+    except Exception as e:
+        LOG.debug('Agent toolkit hint failed: %s', e, exc_info=True)
```
ashovlin, line 179: Are there any exceptions that `ConfigureAgentToolkitCommand` throws that we wouldn't want to swallow?

ashovlin (changes requested):

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Good point.

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Mhmmm. I updated it such that the try/except only covers eligibility + prompting so the wizard runs outside it and surfaces its own errors

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -84,32 +84,33 @@ def _has_installed_skills(detected_agents):
...
-            return
-        command = ConfigureAgentToolkitCommand(session)
-        command([], parsed_globals)
+        run_wizard = choice == 'yes'
     except Exception as e:
         LOG.debug('Agent toolkit hint failed: %s', e, exc_info=True)
+        return
+
+    if run_wizard:
+        command = ConfigureAgentToolkitCommand(session)
```
ashovlin, line 183: Noticed an awkward behavior
```
aws login --profile abc123 --region us-west-2
...
Configure AWS skills and the AWS MCP server for your AI coding agent(s)? [y/n/never]: y
...
aws: [ERROR]: AgentToolkit is only available in us-east-1

```

If I log in for a region other than us-east-1 where the agent-toolkit is available, then opt in to installation, it'll fail when pulling the skills. 

Let's discuss internally, maybe one of:
1. Default the subcommand to us-east-1 (always works, but a hidden cross-region call and doesn't work for different partitions).
2. Downgrade the prompt to a hint when the user configures a different region: (_"Run `aws configure agent-toolkit --region us-east-1`"_)

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def _resolve_region(session, parsed_globals):
+    region = getattr(parsed_globals, 'region', None)
+    if region:
+        return region
+    try:
+        return session.get_config_variable('region')
+    except Exception:
+        return None
```
ashovlin: I think it'd be better to explicitly pass a region in from the `configure` commands that is possibly invoking the hint.

Say a profile is configured for us-east-1, but they go through `aws configure` and change `us-gov-east-1` now. I think we'll still pick up the `us-east-1` from the session.

awscli/customizations/agenttoolkit/hint.py
ashovlin: New proposed region behavior from our internal conversation:
1. If the resolved region is commercial (even if they explicitly passed `--us-west-2`), let's swap to `us-east-1` and log an debug message.
2. If the resolved region is a different partition, let's just show the static hint.

AndrewAsseily: Updated in revision: In the commercial partition the wizard is pinned to `us-east-1` (debug-logged when that differs from the resolved region), and in other partitions the prompt is replaced by the static tip instead of routing the user across a partition

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: The region is now passed in explicitly by each calling command instead of read back off the session in the latest revision

AndrewAsseily:
Also fixed in latest revision: 
- `aws configure --profile <new>` exited non-zero with ProfileNotFound after writing the config if you accepted the prompt
- `--endpoint-url` on the calling command was being forwarded to the Agent Toolkit client 
- `aws update` printed the tip twice on Unix, since it shells out to install.sh, which prints it too

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,184 @@
...
+from awscli.customizations.agenttoolkit.utils import AGENT_TOOLKIT_REGION
+from awscli.customizations.prompts import yes_no_never_choice
+from awscli.customizations.utils import uni_print
+from awscli.utils import is_stdin_a_tty
+
+LOG = logging.getLogger(__name__)
+
+STATE_PATH = '~/.aws/cli/agent-toolkit/state.json'
+
+HINT_DISABLED_ENV_VAR = 'AWS_CLI_AGENT_TOOLKIT_HINT_DISABLED'
```
ashovlin, line 49: Not in this PR, but we should add to https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

.changes/next-release/enhancement-configure-67593.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "configure",
+  "description": "Suggest installing the Agent Toolkit for AWS when a supported AI coding agent is detected. In the commercial partition an interactive prompt is shown after ``aws configure``, ``aws configure sso``, and a first-time ``aws login`` that creates a new profile; in other partitions a non-interactive tip is shown instead. The prompt only appears on a terminal when no AWS skills are installed yet, and can be permanently suppressed by answering ``never`` or by setting the ``AWS_CLI_AGENT_TOOLKIT_HINT_DISABLED`` environment variable to ``true``. ``aws configure agent-toolkit`` now defaults to the ``us-east-1`` region unless ``--region`` is given, and always runs against ``us-east-1`` when launched from the prompt. A tip is also printed by the install scripts and after ``aws update``."
```
ashovlin: This is long, can we do a separate entry in a different file for the region change.

awscli/customizations/agenttoolkit/hint.py
```diff
@@ -0,0 +1,184 @@
...
+)
+
+
+def _state_file():
+    return os.path.expanduser(STATE_PATH)
+
+
+def _load_state():
+    try:
+        with open(_state_file()) as f:
```
ashovlin: Should we specify UTF-8?

ashovlin (changes requested):

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Hah, yes good catch

.changes/next-release/enhancement-configure-67593.json
AndrewAsseily: Done in rev

awscli/customizations/agenttoolkit/hint.py
AndrewAsseily: Will follow up internally

ashovlin (approved):

PR #10517, opened 2026-07-31
  Reduce session database locking and add AWS_CLI_TELEMETRY_DISABLED

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* CLI v2 stores a "session ID" in a SQLite database in the `~/.aws` directory. This stores a random ID that is rotated every 30 minutes, which allows some services to group command invocations together.

This makes some changes to SQLite usage to reduce lock contention on the database when multiple commands are run on the same host concurrently.
* Only writes the timestamp for the current session every 60 seconds, rather than every invocation.
* The sweeper, which deletes old sessions, first checks if any exists before writing.
* Shortens the lock timeout to 100ms, rather than the default of 5s. Session IDs are best effort.
* Also improves `--debug` logging of SQLite issues.

It also introduces the `AWS_CLI_TELEMETRY_DISABLED` environment variable, which can be set to `true` to disable the session behavior.

### DIFF
--- a/awscli/telemetry.py
+++ b/awscli/telemetry.py
@@ -36,7 +36,7 @@ _CACHE_DIR = Path.home() / '.aws' / 'cli' / 'cache'
 _DATABASE_FILENAME = 'session.db'
 _SESSION_LENGTH_SECONDS = 60 * 30
 _SESSION_ID_LENGTH = 12
-# How stale the stored timestamp must be before we write it. Refreshing on every
+# How stale the stored timestamp must be before it is written. Refreshing on every
 # invocation takes a write lock each time, which can cause lock contention.
 _TIMESTAMP_REFRESH_SECONDS = 60
 # How long to wait for a database lock before giving up. Session data is not
--- a/awscli/telemetry.py
+++ b/awscli/telemetry.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import io
+import logging
 import os
 import sqlite3
 import sys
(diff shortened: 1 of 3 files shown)


awscli/telemetry.py
```diff
@@ -239,14 +279,23 @@ def _session_id(self):
...
                 cached_data.session_id = self._session_id
-            # Always update the timestamp to last used.
-            cached_data.timestamp = self._timestamp
-            self._writer.write(cached_data)
+            # Update the timestamp to last used, but only if the stored value
+            # is older than _TIMESTAMP_REFRESH_SECONDS. A session can expire
+            # up to this many seconds earlier than its last actual use, but
+            # avoids a write on every invocation if close together.
+            if (
+                is_expired
```
kdaily, line 293: nit: is this redundant, since its inside the above `if is_expired` block?

awscli/telemetry.py
```diff
@@ -296,7 +345,18 @@ def _get_cli_session_orchestrator():
     )
 
 
+def is_telemetry_disabled(env=None):
+    if env is None:
+        env = os.environ
+    return ensure_boolean(env.get(_TELEMETRY_DISABLED_ENV_VAR, 'false'))
```
kdaily, line 351: Is there a rationale for not using a `ChainProvider`? I understand its proposed to only expose the environment variable so would only use an `EnvironmentProvider` and a `ConstantProvider`. Wondering there was any other reason to not use the same pattern as other features (though it seems we're kind of split on that now with things like `AWS_NEW_RETRIES_2026` and `AWS_CLI_S3_MV_VALIDATE_SAME_S3_PATHS`).

kdaily (commented):
I gave a quick first pass through the implementation, not the tests, and left a couple of questions. I also like the improvements to the logging.

awscli/telemetry.py
kdaily: Nevermind, I missed the indentation level. Resolving.

tests/functional/test_telemetry.py
```diff
@@ -345,6 +406,42 @@ def fake_orchestrator_factory():
     )
 
 
+def test_is_telemetry_disabled():
+    assert is_telemetry_disabled({'AWS_CLI_TELEMETRY_DISABLED': 'true'})
+    assert not is_telemetry_disabled({'AWS_CLI_TELEMETRY_DISABLED': 'false'})
```
aemous, line 411: non-blocking: can consider a test that asserts an error is thrown when the env var value is not a boolean.

aemous (approved):
LGTM.

Issue #10532, opened 2026-08-04
Running the `aws update` command in the PowerShell 7 console fails.


stknohg:
### Describe the bug

My environment is PowerShell 7.6.4 on Windows 11 and Windows Server 2025.  
The console is running with administrator privileges.  

```powershell
# PowerShell 7.6.4
PS C:\> aws --version
aws-cli/2.36.0 Python/3.14.6 Windows/2025Server exec-env/EC2 exe/AMD64
```

When I run the `aws update` command in the PowerShell 7 console, the installation script executes in a separate window, downloads the latest version of the MSI installer file, and then immediately exits.

When I ran the downloaded `aws-update.cmd`, it failed to load the `Get-AuthenticodeSignature` cmdlet.

```powershell
# Run the downloaded aws-update.cmd to reproduce the issue.
PS C:\Users\Administrator\AppData\Local\Temp\2\tmpy48h7xd1> .\aws-update.cmd
Installing AWS CLI 2.36.0 -> 2.36.15
Downloading AWS CLI installer from https://awscli.amazonaws.com/AWSCLIV2-2.36.15.msi?src=update-exe
aws-cli installer: error: The 'Get-AuthenticodeSignature' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more information, run 'Import-Module Microsoft.PowerShell.Security'.
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The script continues without errors, and the MSI installer runs.

### Current Behavior

The script terminates after failing to load the `Get-AuthenticodeSignature`  cmdlet.

### Reproduction Steps

Run the `aws update` command in a PowerShell 7 console.

### Possible Solution

`aws update` command launches Windows PowerShell to execute the update script, but because it inherits the `PSModulePath` environment variable from the PowerShell 7 environment, the module resolution order is not as intended.

```powershell
# Default PowerShell 7 PSModulePath: 
PS C:\> $env:PSModulePath
C:\Users\Administrator\Documents\PowerShell\Modules;C:\Program Files\PowerShell\Modules;c:\program files\powershell\7\Modules;C:\Windows\system32\WindowsPowerShell\v1.0\Modules;C:\Program Files\WindowsPowerShell\Modules

# Default Windows PowerShell PSModulePath: 
PS C:\> $env:PSModulePath
[...]

PR #10541, opened 2026-08-10
[v2] Avoid PSModulePath inheritance during update on Windows

hssyoo:
Fixes https://github.com/aws/aws-cli/issues/10532

Currently, `update` launches a PS5 console via `powershell.exe` when running the install script. This causes an issue where if the `update` command is run within a PS7 console, the PS5 execution inherits PS7's `PSModulePath` env var value, leading to path issues when the install script attempts to run `Get-AuthenticodeSignature`.

This PR fixes the issue by removing `PSModulePath` before launching PS5, letting it rebuild its default `PSModulePath` value. The PR also falls back to `pwsh.exe` (PS7) if `powershell.exe` isn't available when resolving the PowerShell path.

### DIFF
--- a/.changes/next-release/bugfix-update-40840.json
+++ b/.changes/next-release/bugfix-update-40840.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``update``",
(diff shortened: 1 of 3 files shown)


aemous (approved):
LGTM. Confirmed via internal discussions that this was manually tested on a Windows machine.

PR #10559, opened 2026-08-14
Fix ec2 run-instances not validating --no-associate-public-ip-address against --network-interfaces

### LINKED ISSUE #10558: ec2 run-instances: --no-associate-public-ip-address silently ignored when mixed with --network-interfaces
### Describe the bug

`aws ec2 run-instances --network-interfaces ...` is documented to reject mixing with the "simple, scalar" options (`--associate-public-ip-address`, `--secondary-private-ip-addresses`, `--secondary-private-ip-address-count`) by raising:

```
Mixing the --network-interfaces option with the simple, scalar options is not supported.
```

This works correctly for `--associate-public-ip-address`, but **not** for its counterpart `--no-associate-public-ip-address`. That flag is silently accepted instead of raising the same error, and it also overwrites `AssociatePublicIpAddress` inside the user's own `--network-interfaces` JSON to `False` — even if the user's JSON explicitly set it to `True`.

### Expected Behavior

`--no-associate-public-ip-address` combined with `--network-interfaces` should raise the same `ParamValidationError` that `--associate-public-ip-address` does.

### Current Behavior

The command silently succeeds and mutates `NetworkInterfaces[0]["AssociatePublicIpAddress"]` to `False`.

### Reproduction Steps

```
aws ec2 run-instances --image-id ami-0123456789 \
  --network-interfaces '[{"DeviceIndex":0,"SubnetId":"subnet-abc","AssociatePublicIpAddress":true}]' \
  --no-associate-public-ip-address
```

Compare with (correctly errors):

```
aws ec2 run-instances --image-id ami-0123456789 \
  --network-interfaces '[{"DeviceIndex":0,"SubnetId":"subnet-abc"}]' \
  --associate-public-ip-address
```

### Root cause

In `awscli/customizations/ec2/runinstances.py`, `_check_args()` only checks the `associate_public_ip_address` dest (set by `--associate-public-ip-address`, `action='store_true'`). It never checks `no_associate_public_ip_address` (set by `--no-associate-public-ip-address`, `action='store_false'`), which is a distinct argparse dest even though both flags share a `group_name` for help-text grouping only.

### Fix

I've opened a PR with a fix and regression tests: will link below.

### Environment

- `aws-cli/2.x` (customization present on current `v2` branch)

Adityaj0:
## Summary

Fixes #10558.

`--no-associate-public-ip-address` and `--associate-public-ip-address` are two independent argparse destinations (`no_associate_public_ip_address` / `associate_public_ip_address`) — they only share a `group_name` for help-text grouping, not a `dest`. `_check_args()` in `awscli/customizations/ec2/runinstances.py` only ever checked `associate_public_ip_address` when validating that `--network-interfaces` isn't mixed with the scalar options, so `--no-associate-public-ip-address` silently bypassed the check and overwrote `AssociatePublicIpAddress` inside the user's own `--network-interfaces` JSON instead of raising the documented `ParamValidationError`.

This adds the missing check for `no_associate_public_ip_address is False` (its "flag was explicitly passed" state, since it's a `store_false` action defaulting to `True`).

## Test plan

- [x] Added `test_network_interfaces_with_no_associate_public_ip_address_errors` (regression test for the bug) and `test_network_interfaces_with_associate_public_ip_address_errors` (locks in existing correct behavior) to `tests/functional/ec2/test_run_instances.py`
- [x] `python -m pytest tests/functional/ec2/ tests/unit/customizations/ec2/` — 108 passed
- [x] Added changelog entry under `.changes/next-release/`

### DIFF
--- a/.changes/next-release/bugfix-ec2-55884.json
+++ b/.changes/next-release/bugfix-ec2-55884.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "ec2",
+  "description": "Fix `aws ec2 run-instances --network-interfaces ... --no-associate-public-ip-address` silently overwriting AssociatePublicIpAddress in the user-supplied network interface instead of raising the documented error for mixing --network-interfaces with scalar options"
+}
--- a/awscli/customizations/ec2/runinstances.py
+++ b/awscli/customizations/ec2/runinstances.py
@@ -81,18 +81,20 @@ def _check_args(parsed_args, **kwargs):
     # raise an error.
     arg_dict = vars(parsed_args)
     if arg_dict['network_interfaces']:
+        msg = (
+            'Mixing the --network-interfaces option '
+            'with the simple, scalar options is '
+            'not supported.'
+        )
         for key in (
             'secondary_private_ip_addresses',
             'secondary_private_ip_address_count',
             'associate_public_ip_address',
         ):
             if arg_dict[key]:
-                msg = (
-                    'Mixing the --network-interfaces option '
-                    'with the simple, scalar options is '
-                    'not supported.'
-                )
                 raise ParamValidationError(msg)
+        if arg_dict['no_associate_public_ip_address'] is False:
+            raise ParamValidationError(msg)
 
 
 def _fix_args(params, **kwargs):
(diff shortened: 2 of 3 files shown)


Adityaj0:
Fixes https://github.com/aws/aws-cli/issues/10558

PR #10584, opened 2026-08-19
feat(user-agent): Attribute the Kiro IDE agent

seshubaws:
The Kiro IDE and Kiro CLI are separate products with separate app bundles. Only the CLI exports KIRO_SESSION_ID, so commands driven by the IDE's agent were reported as ordinary human-run commands and its traffic was unattributed.

Add AGENTIC_CALLER_KIRO_IDE (AX), keyed on KIRO_AGENT. The variable is scoped to agent activity rather than to the editor being open, following CURSOR_AGENT and ANTIGRAVITY_AGENT. 

Something like TERM_PROGRAM=kiro was rejected since the IDE's integrated terminal is also human-driven, so it would attribute a person's typing as agentic and make the metric non-comparable with the other AGENTIC_CALLER_* rows. Kiro service must export KIRO_AGENT for this to report, matching how Cursor behaves today. Setting this up for when it does.

Also expose is_agentic_caller() so callers that adapt behavior for agents (for example, suppressing interactive prompts) use the same signals the User-Agent already reports instead of reimplementing the table.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscli/botocore/useragent.py
+++ b/awscli/botocore/useragent.py
@@ -170,6 +170,23 @@ def _agentic_env_var_is_set(env_var, expected_value):
     return value == expected_value
 
 
+def is_agentic_caller():
+    """Return True if the current process appears to be driven by a known
+    agentic caller (e.g. Claude Code, Gemini CLI, Codex).
+
+    Detection uses the same environment-variable signals that populate the
+    ``AGENTIC_CALLER_*`` metrics in the User-Agent header, so callers that
+    want to adapt their behavior for agents stay in sync with what the SDK
+    already reports.
+    """
+    return any(
+        _agentic_env_var_is_set(env_var, expected_value)
+        for env_var, _feature_id, expected_value in (
+            _USERAGENT_AGENTIC_CALLER_ENV_VAR_MAPPINGS
+        )
+    )
+
+
 def _register_agentic_caller_env_features():
     for (
         env_var,
--- a/.changes/next-release/enhancement-UserAgent-53908.json
+++ b/.changes/next-release/enhancement-UserAgent-53908.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "User Agent",
+  "description": "Expose ``is_agentic_caller`` so callers can detect the agentic callers already reported in the User-Agent."
+}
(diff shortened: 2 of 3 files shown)


awscli/botocore/useragent.py
```diff
@@ -127,6 +128,10 @@
     ('PI_CODING_AGENT', 'AGENTIC_CALLER_PI', 'true'),
     ('COPILOT_CLI', 'AGENTIC_CALLER_COPILOT_CLI', '1'),
     ('CURSOR_AGENT', 'AGENTIC_CALLER_CURSOR', '1'),
+    # Kiro IDE, distinct from the Kiro CLI above. The IDE agent drives an
+    # integrated terminal and does not export KIRO_SESSION_ID, so its traffic
+    # was previously unattributed.
+    ('KIRO_AGENT', 'AGENTIC_CALLER_KIRO_IDE', None),
```
hssyoo: Are there docs that reference `KIRO_AGENT`? When prompting Kiro to check runtime environment variables on Kiro IDE v1.0.337, there's no match on `KIRO_AGENT`.

awscli/botocore/useragent.py
```diff
@@ -127,6 +128,10 @@
     ('PI_CODING_AGENT', 'AGENTIC_CALLER_PI', 'true'),
     ('COPILOT_CLI', 'AGENTIC_CALLER_COPILOT_CLI', '1'),
     ('CURSOR_AGENT', 'AGENTIC_CALLER_CURSOR', '1'),
+    # Kiro IDE, distinct from the Kiro CLI above. The IDE agent drives an
+    # integrated terminal and does not export KIRO_SESSION_ID, so its traffic
+    # was previously unattributed.
```
hssyoo: nit: We can remove comment. These values are documented internally where we define feature ids.

awscli/botocore/useragent.py
hssyoo: > Kiro service must export KIRO_AGENT for this to report, matching how Cursor behaves today. Setting this up for when it does.

I missed this in the description. I don't think we should ship code that depends on future work by the Kiro team. The user agent won't track Kiro IDE usage with or without this PR.

seshubaws:
ack, cut a ticket to kiro team internally

PR #10591, opened 2026-08-20
[v2] Add blackbox tests

aemous:
*Notes:*

* Merging this PR is currently a no-op. They can be run ad-hoc by developers but are not invoked in any automated CI.

*Description of changes:*

* Added scaffolding to easily add blackbox tests. The main helper function is `assert_operation` which lets callers verify if an HTTP request matches an operation. This lets test-writers easily verify if the expected operations were called in the expected order.
* Added S3 blackbox tests. These are ports of our existing S3 functional tests (except cp, mv, or sync commands). The only tests not ported are those that cannot be feasibly ported to a blackbox case (e.g. relies on a Python monkey-patch of datetime).
* Added new blackbox requirements test file, and updated regenerate-lock-files and GitHub workflows to account for this.

*Description of tests:*

* Ran and passed all new blackbox tests via PyTest.


tests/blackbox/utils.py
```diff
@@ -0,0 +1,508 @@
...
+from pathlib import Path
+from urllib.parse import parse_qs, urlparse
+
+import pytest
+from localstub.server import AsyncHTTPTestServer, HTTPResponse
+from localstub.tlsproxy import AsyncTLSInterceptProxy
+
+_MODEL_DIR = (
+    Path(__file__).resolve().parents[2] / "awscli" / "botocore" / "data"
+)
```
hssyoo: This couples the `aws` binary under test to C2J models. This wouldn't work under the following hypothetical cases:
* An AWS CLI that uses Smithy models at runtime
* An AWS CLI that uses a custom model format derived from AWS models
* An AWS CLI with codegenned clients that don't commit models into the source code

tests/blackbox/utils.py
```diff
@@ -0,0 +1,508 @@
...
+        "HOME": os.environ.get("HOME", "/tmp"),
+        "AWS_ACCESS_KEY_ID": "test",
+        "AWS_SECRET_ACCESS_KEY": "test",
+        "AWS_REGION": "us-east-1",
+        "AWS_DEFAULT_REGION": "us-east-1",
+        "AWS_MAX_ATTEMPTS": "1",
+        "AWS_CONFIG_FILE": "",
+        "AWS_SHARED_CREDENTIALS_FILE": "",
+        "HTTPS_PROXY": proxy.endpoint_url,
+        "AWS_CA_BUNDLE": str(proxy.ca.ca_pem_path()),
```
hssyoo, line 27: We should have some sort of guard against making calls to prod AWS endpoints in case the binary under test doesn't implement support for these environment variables.

tests/blackbox/utils.py
```diff
@@ -0,0 +1,508 @@
...
+
+
+def _load_model(service: str) -> dict:
+    service_dir = _MODEL_DIR / service
+    latest = sorted(service_dir.iterdir())[-1]
+    with open(latest / "service-2.json") as f:
+        return json.load(f)
+
+
+def _get_model(service: str) -> dict:
```
hssyoo: This pattern of loading and parsing models seems counter to a blackbox testing approach. When comparing multiple binaries using a blackbox test suite, the tests should be independent of any concrete implementation and necessarily simple since the only common surface between binaries are the command inputs and wire bytes.

Instead, it looks like we're reimplementing `botocore` parsing logic using `botocore`s vendored models and using that to derive the expected wire format.

hssyoo (changes requested):

tests/blackbox/utils.py
aemous: The AWS CLI does use prod AWS endpoints, just that the TLS intercept is intercepting the traffic.

I added a PyTest hook that runs before any tests run. The hook spawns a localstub server using the HTTPS proxy and CA bundle and checks if the mock server received the request. If not, it exits and prevents any tests from running.

Also, we mock fake AWS creds in the environment variables. So in the worst case, if a request reached prod despite our new PyTest hook, the credentials should be rejected and the tests should fail loudly.

tests/blackbox/utils.py
aemous: Pushed a revision to remove the dependency on the JSON service models.

tests/blackbox/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+                HTTPResponse.raw(
+                    b'<?xml version="1.0" ?>'
+                    b"<Error><Code>AccessDenied</Code>"
+                    b"<Message>Test</Message></Error>",
+                    status=403,
+                    headers={"Content-Type": "application/xml"},
+                ),
+            ])
+            env = cli_env(proxy)
+            await run_cli(command, ["s3", "ls"], env)
```
hssyoo, line 43: An issue with only guarding `s3 ls` is that an `aws` binary with partial `s3` implementation could hit false negatives. For example, if `s3 ls` isn't implemented yet and falls back to the current Python runtime, then it'll pass this assertion but a test running `s3 sync` could still silently hit prod endpoints.

I wonder if for every test where we expect `localstub` to intercept requests, we assert that `server.requests` is truthy. This wouldn't stop a test from hitting prod endpoints but would at least fail explicitly with a helpful message.

tests/blackbox/test_mb_command.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def parse_body_xml(request) -> ET.Element | None:
+    if not request.body:
+        return None
+    return ET.fromstring(request.body)
+
+
+@pytest.mark.asyncio
+class TestMBCommand:
```
hssyoo, line 32: I know that current tests organize using classes but my guess is that that's just a legacy side effect of trying to mirror unittest-based tests. I think we should just move the tests out of classes moving forward, similar to how you organized `test_ls_command.py`.

Mainly, these pytest classes hold no state and setup/teardown is managed by fixtures.

tests/blackbox/s3_assertions.py
```diff
@@ -0,0 +1,685 @@
...
+def _parse_xml_body(body: str | bytes, ns: str) -> dict | None:
+    """Parse an XML request body into a dict, handling namespace."""
+    if not body:
+        return None
+    root = ET.fromstring(body if isinstance(body, str) else body.decode("utf-8"))
+    ns_prefix = f"{{{ns}}}" if ns else ""
+    return _xml_to_dict(root, ns_prefix)
+
+
+def _xml_to_dict(element, ns_prefix: str) -> dict | str | list:
```
hssyoo, line 130: In general, I think it's good to parse bodies into Python objects but XML is tricky since the document doesn't carry type information. It looks like the PR is aware of this given the `is_list` heuristic. For example, I think scalars will always be represented as strings even if they may be integers (`'1' vs 1`).

How can we declare expected types of expected values in these tests? I understand the hard part is doing this without runtime models available.

hssyoo (commented):
Looking good, mainly had some structural feedback.

tests/blackbox/utils.py
hssyoo: I left a comment on the hook, but given your point on fake AWS creds, I'd be okay without the guard.

tests/blackbox/test_mb_command.py
aemous: Okay, I will keep the current classes but moving forward I will not use classes.

tests/blackbox/conftest.py
aemous: The current state for almost all tests is that we assert the length is some positive number. So I think we already do this. All of the assert helper functions like `assert_put_object` require a request object to be present. 

I feel that an added test that server.requests would be redundant with the existing assertions.

Is this sufficient in combination with the existing hook on `ls`?

tests/blackbox/s3_assertions.py
aemous: I published changes that address this. For XML specifically, `assert_subset` will use type coercion to turn the expected param values into strings before comparisons.

In the future, as we expand to use non-XML protocols, the wire protocols that support numeric types will not use type coercion.

hssyoo (changes requested):
Looks good, can we take a test dependency on `localstub` if we're going to be merging to `v2`

hssyoo (approved):

hssyoo (approved):

PR #10610, opened 2026-08-27
add new `aws docs search` command

jamesward:
While some users may find this useful, its primary purpose is for AI coding agents so that they can search AWS docs. This is helpful when the agent does not have the AWS MCP installed.

### DIFF
--- a/awscli/customizations/docs.py
+++ b/awscli/customizations/docs.py
@@ -1,0 +1,321 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Adds an ``aws docs search <query>`` command.
+
+The command queries the public AWS documentation search endpoint and
+renders the returned suggestions either as plain text (title/link pairs)
+or as the raw JSON exactly as it comes back from the service.
+
+The HTTP side effect is isolated behind :class:`DocSearchClient` so that
+the command logic and rendering can be exercised without performing a
+real network call.  The rendering helpers are pure functions operating on
+plain data, which keeps the testable "onion" thin and fast.
+"""
+
+import enum
+import json
+import os
+import sys
+import uuid
+from dataclasses import dataclass
+
+from awscli.botocore.awsrequest import AWSRequest
+from awscli.botocore.httpsession import URLLib3Session
+from awscli.compat import is_windows
+from awscli.customizations.commands import BasicCommand
+
+#: The public AWS documentation search endpoint.
+DOC_SEARCH_URL = 'https://proxy.search.docs.aws.com/search'
+#: The documentation domain scoped for the search.
+DOCS_DOMAIN = 'docs.aws.amazon.com'
+#: Locale requested from the search service when none can be detected.
+DEFAULT_LOCALE = 'en_us'
+#: Environment variables consulted (in order) to detect the user's locale.
+LOCALE_ENV_VARS = ('LC_ALL', 'LC_MESSAGES', 'LANG')
+
+
+def normalize_locale(raw_locale):
+    """Normalize an OS locale string to the API's ``ll_cc`` form.
+
+    Converts values like ``en_US.UTF-8`` or ``en-US`` to ``en_us``. Returns
+    an empty string for unusable locales (empty, ``C`` or ``POSIX``), which
+    signals that the caller should fall back to the default.
+    """
+    if not raw_locale:
+        return ''
+    # Drop encoding ('.UTF-8') and modifier ('@euro') suffixes.
+    language = raw_locale.split('.', 1)[0].split('@', 1)[0]
+    language = language.replace('-', '_')
+    if not language or language.upper() in ('C', 'POSIX'):
+        return ''
+    return language.lower()
+
+
+def _windows_locale():
+    """Return the current user's locale on Windows (e.g. ``en-US``).
+
+    Reads it from the Win32 API since Windows shells do not populate the
+    ``LC_*``/``LANG`` environment variables. Returns an empty string if the
+    lookup fails.
+    """
+    import ctypes
+
+    # LOCALE_NAME_MAX_LENGTH is 85 wide chars.
+    buffer_length = 85
+    buffer = ctypes.create_unicode_buffer(buffer_length)
+    written = ctypes.windll.kernel32.GetUserDefaultLocaleName(
+        buffer, buffer_length
+    )
+    if written:
+        return buffer.value
+    return ''
+
+
+def _os_locale():
+    """Return the OS-reported locale for the current platform.
+
+    On Unix (Linux/macOS) the standard ``LC_*``/``LANG`` environment
+    variables are authoritative, so there is nothing extra to query here and
+    we return an empty string. On Windows those variables are typically
+    unset, so we ask the OS directly.
+    """
+    if is_windows:
+        return _windows_locale()
+    return ''
+
+
+def detect_locale(environ=None, os_locale=None):
+    """Detect the user's locale in a cross-platform way.
+
+    Resolution order:
+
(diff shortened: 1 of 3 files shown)


awscli/customizations/docs.py
```diff
@@ -0,0 +1,321 @@
...
+
+from awscli.botocore.awsrequest import AWSRequest
+from awscli.botocore.httpsession import URLLib3Session
+from awscli.compat import is_windows
+from awscli.customizations.commands import BasicCommand
+
+#: The public AWS documentation search endpoint.
+DOC_SEARCH_URL = 'https://proxy.search.docs.aws.com/search'
+#: The documentation domain scoped for the search.
+DOCS_DOMAIN = 'docs.aws.amazon.com'
```
ashovlin, line 30: Will follow-up internally, but is there an alternative for users in other partitions?

awscli/customizations/docs.py
```diff
@@ -0,0 +1,321 @@
...
+    Using an enum (instead of bare strings scattered around the module)
+    keeps the set of valid formats closed and impossible to mistype.
+    """
+
+    TEXT = 'text'
+    JSON = 'json'
+
+    @classmethod
+    def choices(cls):
+        return [member.value for member in cls]
```
ashovlin, line 130: Is there a strong reason to default to `text`? 

Generally I'd rather honor the existing [`--output`/`AWS_DEFAULT_OUTPUT`/profile `output` hierarchy](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html), though this'll default to JSON for most users.

Here's a pattern for selecting the right output formatter from a custom command:
https://github.com/aws/aws-cli/blob/941e56792156b8f79cb08fd9d0d45f3f96a8668e/awscli/customizations/agenttoolkit/list_installed_skills.py#L58-L63

awscli/customizations/docs.py
```diff
@@ -0,0 +1,321 @@
...
+    data = json.loads(raw_response)
+    excerpts = []
+    for suggestion in data.get('suggestions', []):
+        excerpt = suggestion.get('textExcerptSuggestion')
+        if excerpt is None:
+            continue
+        excerpts.append(
+            TextExcerpt(
+                title=excerpt.get('title', ''),
+                link=excerpt.get('link', ''),
```
ashovlin, line 170: Did we intentionally omit `suggestionBody`? Imagine there's a tradeoff there between token usage and steering towards a more helpful link?

awscli/customizations/docs.py
```diff
@@ -0,0 +1,321 @@
...
+    variables are authoritative, so there is nothing extra to query here and
+    we return an empty string. On Windows those variables are typically
+    unset, so we ask the OS directly.
+    """
+    if is_windows:
+        return _windows_locale()
+    return ''
+
+
+def detect_locale(environ=None, os_locale=None):
```
ashovlin, line 87: Do we know which locales are valid?

Tried with this, and got a 400. 
```
export LANG="en_GB.UTF-8"        
export LC_ALL="en_GB.UTF-8"
```

awscli/customizations/docs.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def __init__(
+        self,
+        session=None,
+        endpoint_url=DOC_SEARCH_URL,
+        session_id_factory=None,
+        locale=None,
+        user_agent=None,
+    ):
+        self._session = session or URLLib3Session()
```
ashovlin, line 214: It jives with the endpoint question above and will follow-up internally, but is there a path towards this being a modeled API? We get a lot more for free with an actual botocore client (proxy support, CA-bundle, timeout configuration, retries, etc.)

ashovlin (changes requested):

awscli/customizations/docs.py
jamesward: Good question. I think the agents are generally happy with the text form, but they are probably also fine with JSON so it seems reasonable to keep this consistent with other commands.

awscli/customizations/docs.py
jamesward: I think we should include the suggestionBody. Even though it is more tokens it probably helps reduce token usage later.

awscli/customizations/docs.py
jamesward: I'll investigate

awscli/customizations/docs.py
jamesward: that would be nice! let's discuss more internally.

awscli/customizations/docs.py
ashovlin: Sounds good. I have been thinking about if we want to change the output format if we detect an agentic caller and/or adding https://toonformat.dev/ support too. Want to benchmark internally.

jamesward: yeah, that'd definitely be worth some eval testing.

awscli/customizations/docs.py
jamesward: suggestionBody ended up being almost exactly the same as summary, so I didn't add it.

awscli/customizations/docs.py
jamesward: Added supported locales with default of `en_US`

jamesward:
Here is what the output now looks like:

`aws docs search s3`

```json
[
    {
        "link": "https://docs.aws.amazon.com/s3/",
        "title": "Amazon Simple Storage Service Documentation",
        "summary": "Documentation for Amazon S3, including User Guides, API References, Developer Guides, and other resources.",
        "type": "CuratedSuggestion"
    },
    {
        "link": "https://docs.aws.amazon.com/AmazonS3/latest/userguide/",
        "title": "Amazon Simple Storage Service User Guide",
        "summary": "Provides detailed information and instructions on getting started and working with or developing Amazon S3.",
        "type": "CuratedSuggestion"
    },
    {
        "link": "https://docs.aws.amazon.com/AmazonS3/latest/API/",
        "title": "Amazon Simple Storage Service API Reference",
        "summary": "Describes all Amazon S3 operations in detail. Also provides sample requests, responses, and errors for supported web service protocols.",
        "type": "CuratedSuggestion"
    }
]
```

`aws docs search --output=text s3`

```text
https://docs.aws.amazon.com/s3/                 Documentation for Amazon S3, including User Guides, API References, Developer Guides, and other resources.      Amazon Simple Storage Service Documentation     CuratedSuggestion
https://docs.aws.amazon.com/AmazonS3/latest/userguide/                  Provides detailed information and instructions on getting started and working with or developing Amazon S3.     Amazon Simple Storage Service User Guide        CuratedSuggestion
https://docs.aws.amazon.com/AmazonS3/latest/API/                        Describes all Amazon S3 operations in detail. Also provides sample requests, responses, and errors for supported web service protocols. Amazon Simple Storage Service API Reference     CuratedSuggestion
https://docs.aws.amazon.com/cli/latest/reference/s3/                    Provides AWS CLI commands, concepts, and notations for AWS S3.  Amazon Simple Storage Service CLI       CuratedSuggestion
https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html              1790274508186   Store data in the cloud and learn the core concepts of buckets and objects with the Amazon S3 web service.      What is Amazon S3? - Amazon Simple Storage Service      
https://docs.aws.amazon.com/AmazonS3/latest/userguide/GetStartedWithS3.html             1784315427805   Get started storing and retrieving any amount of data at any time, from anywhere on the web with Amazon S3.     Getting started with Amazon S3 - Amazon Simple Storage Service  
https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingBucket.html  1783975667619   1783975667619   Store all of your files, known as objects, within a uniquely named Amazon S3 general purpose bucket.    General purpose buckets overview - Amazon Simple Storage Service        
```

PR #10615, opened 2026-09-02
Add agent-toolkit check-updates and update-skill --all

AndrewAsseily:
*Description of changes:* 

Adds `aws agent-toolkit check-updates`, which reports the installed version, latest available version, and whether an update is available for each installed AWS skill. It is read-only and emits structured output, so it can be filtered with `--query` or scripted. Adds `--all` to `aws agent-toolkit update-skill` to update every out-of-date installed skill in one command instead of one `--skill-name` at a time. Both accept `--agent` to scope to a single tool. `--skill-name` becomes optional on `update-skill` and is mutually exclusive with `--all`; passing neither or both is a validation error. Existing `update-skill --skill-name` behavior is unchanged.

Each skill's version is resolved once per distinct name rather than once per install, and `--all` downloads a skill's archive once and writes it to every agent that has it installed. Neither command creates a client until local state shows there is something to look up.

### DIFF
--- a/.changes/next-release/enhancement-agenttoolkit-26195.json
+++ b/.changes/next-release/enhancement-agenttoolkit-26195.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "agent-toolkit",
+  "description": "Adds ``aws agent-toolkit check-updates`` to report the installed version, latest available version, and whether an update is available for each installed AWS skill, and adds ``--all`` to ``aws agent-toolkit update-skill`` to update every installed skill that is out of date."
+}
--- a/awscli/customizations/agenttoolkit/utils.py
+++ b/awscli/customizations/agenttoolkit/utils.py
@@ -137,7 +137,7 @@ def read_skill_metadata(skill_dir):
     path = os.path.join(skill_dir, SKILL_METADATA_FILENAME)
     try:
         with open(path) as f:
-            return json.load(f)
+            metadata = json.load(f)
     except FileNotFoundError:
         return None
     except json.JSONDecodeError as e:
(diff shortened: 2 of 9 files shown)


.changes/next-release/enhancement-agenttoolkit-26195.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "agent-toolkit",
+  "description": "Adds ``aws agent-toolkit check-updates`` to report the installed version, latest available version, and whether an update is available for each installed AWS skill, and adds ``--all`` to ``aws agent-toolkit update-skill`` to update every installed skill that is out of date."
```
ashovlin, line 4: nit: Split in to two.

awscli/examples/agenttoolkit/check-updates.rst
```diff
@@ -0,0 +1,72 @@
...
+
+For more information, see `Getting started with the AWS Agent Toolkit <https://docs.aws.amazon.com/agent-toolkit/latest/userguide/getting-started.html>`__ in the *AWS Agent Toolkit User Guide*.
+
+**Example 3: To list only the skills that have an update available**
+
+The following ``check-updates`` example uses ``--query`` to return just the skills that are out of date. ::
+
+    aws agent-toolkit check-updates \
+        --query 'skills[?updateAvailable].[name,installedVersion,latestVersion]' \
+        --output table
```
ashovlin, line 62: Should we just add a flag for this?

awscli/customizations/agenttoolkit/utils.py
```diff
@@ -103,6 +103,29 @@ def get_skill_download(client, skill_name, version=None):
     return zip_bytes, checksum, version
 
 
+def collect_installed_skills(agents):
```
ashovlin, line 106: Could `_run_main` in `list-installed-skills` use this? Looks similar.

ashovlin (approved):

PR #10617, opened 2026-09-03
[v2] Pass resolved configured checksum options to CRT

hssyoo:
Honor resolved `response_checksum_validation` and `request_checksum_calculation` values when resolved transfer client is CRT. Previously, they were effectively always `when_supported` and ignored `when_required`.

### DIFF
--- a/.changes/next-release/bugfix-crt-96245.json
+++ b/.changes/next-release/bugfix-crt-96245.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "crt",
+  "description": "Pass configured `response_checksum_validation` and `request_checksum_calculation` options to CRT transfers."
+}
--- a/.changes/next-release/bugfix-crt-96245.json
+++ b/.changes/next-release/bugfix-crt-96245.json
@@ -1,5 +1,5 @@
 {
   "type": "bugfix",
   "category": "crt",
-  "description": "Pass configured `response_checksum_validation` and `request_checksum_calculation` options to CRT transfers."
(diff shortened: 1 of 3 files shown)


.changes/next-release/bugfix-crt-96245.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "crt",
+  "description": "Pass configured `response_checksum_validation` and `request_checksum_calculation` options to CRT transfers."
```
ashovlin:
```suggestion
  "description": "Pass configured ``response_checksum_validation`` and ``request_checksum_calculation`` options to CRT transfers."
```

awscli/s3transfer/crt.py
```diff
@@ -844,6 +861,32 @@ def _get_make_request_args_get_object(
...
+        if any(
+            checksum_arg in extra_args
+            for checksum_arg in FULL_OBJECT_CHECKSUM_ARGS
+        ):
+            return False
+        if 'ChecksumAlgorithm' in extra_args:
+            return True
+        return (
+            self._get_client_config('request_checksum_calculation')
+            != WHEN_REQUIRED
```
ashovlin, line 895: Checking my understanding - this is relying on the fact that fact that none of the operations invoked by the CRT TM have `requestChecksumRequired: true` in the model, right? Otherwise we'd have to look at the model and still set a checksum anyway for `when_required`

hssyoo: Ah good catch, added model check

ashovlin (approved):

PR #10623, opened 2026-09-08
Support following S3 bucket region redirects when the CRT client is enabled

ashovlin:
*Issue #, if available:* CLI-9212

*Description of changes:* This adds support for following S3 bucket region redirects for the high-level `s3` commands when the CRT transfer manager is enabled.
  * `CRTTransferCoordinator` still coordinates the entire transfer for a single object. It now wraps the initial request and the optional retry, and exposes future-style `cancel`/`result`/etc. that apply to the either transfer, and act on either underlying request
  * `CRTTransferManager` - rather than containing a single CRT client, now caches a map of regions to clients
  * For the actual redirect logic, I extracted the guts of `S3RegionRedirectorv2` so we can use them for CRT requests. I added `CRTS3RegionRedirectPolicy` on top which checks some CRT-specific conditions, and `get_bucket_region` transforms the CRT error into something `S3RegionRedirectorv2` can work with
  * `LazyHeadBucketClient` - our existing Python redirect logic supports calling `HeadBucket` if it receives an error but cannot parse the region from the error response headers. I didn't see this case in testing, but to preserve the code path, we just a lazy boto client if needed.
  * Throughout, there is more initialization around clients, factories, etc. and some more bookkeeping of S3 request parameters that we need to adjust for the retry if needed.


.changes/next-release/enhancement-s3-42614.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``s3```",
```
hssyoo: Extra backtick

ashovlin:
Added some fixes in [c9feb20](https://github.com/aws/aws-cli/pull/10623/commits/c9feb205331ca32389a4b9d0bc908970ef02bf5e) after reviewing locally in a fresh Claude session:
* Added try/except handling around redirect logic, to ensure a transfer finishes with an error
* Moved the redirect region lookup (which possibly makes a boto `HeadBucket` call) to a new thread, to free up the CRT event loop
* Cleaned up locking around the bucket -> redirect region cache
* Avoided sending redirect requests if the region happens to match the original (not expected)

awscli/botocore/utils.py
hssyoo, line 1629: Previously, if error was not redirect-related, it would return early and this debug log would never fire. Now, if the error is not redirect-related, it returns early only for `self.get_redirect_region()`. Since `new_region = None` in this case, the debug log fires anyways. We should break out redirect check in `get_redirect_region()` to something like `_is_redirect_response()` and return from `redirect_from_error` early.

awscli/s3transfer/crt.py
```diff
@@ -341,29 +541,68 @@ def _submit_transfer(self, request_type, call_args):
...
+                    crt_callargs = (
+                        self._s3_args_creator.get_make_request_args(
+                            request_type,
+                            call_args,
+                            coordinator,
+                            future,
+                            on_done_after_calls,
+                        )
+                    )
+                    crt_client = self.get_crt_client(region)
```
hssyoo, line 589: Can we move client creation outside of the lock?

awscli/s3transfer/crt.py
```diff
@@ -705,16 +1093,172 @@ def handle_exception(self, exc):
...
+
+        def request_done(error=None, **kwargs):
+            if error is not None and self._can_redirect(
+                is_region_redirect, bytes_transferred
+            ):
+                # Discovering a region and serializing the retry can both
+                # block, and this runs on a CRT completion thread, where
+                # blocking stalls every other transfer sharing the event loop.
+                self._dispatch_redirect(redirect_and_finish, error, kwargs)
+                return
```
hssyoo, line 1180: Based on the logic of `_can_redirect()` and `CRTS3RegionRedirectPolicy.is_error_redirect_candidate()`, we're not actually checking if it's a redirect error before spawning a thread. So any error that doesn't get filtered out spawns a `crt-s3-region-redirect`. Can we check the instance of `error` and filter out non S3 redirect errors before spawning?

awscli/s3transfer/crt.py
```diff
@@ -201,31 +219,200 @@ def _get_crt_throughput_target_gbps(provided_throughput_target_bytes=None):
...
+
+    def __init__(self, crt_request_serializer):
+        self._crt_request_serializer = crt_request_serializer
+        # Bucket regions are held under one lock so that a burst of transfers
+        # failing at once shares a single lookup instead of each paying for
+        # its own, potentially a HeadBucket request each, and so that a
+        # request cannot be built for a region that changes while it is built.
+        # One lock covers every bucket because a command generally transfers
+        # to or from a single one.
+        self._region_lock = threading.Lock()
```
hssyoo, line 238: It looks like this lock is acquired for both region discovery (`get_retry_region()`) and cache lookup (`locked_bucket_region()`). The issue is that the cache lookup occurs during CRT request serialization, so it's possible for it to be blocked while `get_retry_region` has a lock for `HeadObject`. In the worst case, a long-running `HeadObject` (retries, timeouts, etc) can block CRT transfer requests that may have correctly configured buckets. I _think_ a workaround here is to use 2 locks.

awscli/s3transfer/crt.py
```diff
@@ -705,16 +1093,172 @@ def handle_exception(self, exc):
...
+        transfer_type,
+        is_replayable=True,
+    ):
+        """Submit the transfer's CRT request.
+
+        A request that failed because it was made in the wrong region for its
+        bucket is resubmitted in the bucket's region, which makes a second
+        request for the same transfer.
+        """
+        self._request_factory = request_factory
```
hssyoo, line 1112: Can we clear this on `complete()`? It looks like the source of a memory leak

hssyoo (changes requested):

awscli/s3transfer/crt.py
ashovlin: Set it to `None` in `complete` in b54f87c78886787f498d6700d487d25bef17fbd0

awscli/botocore/utils.py
ashovlin: Added above:

https://github.com/aws/aws-cli/blob/b54f87c78886787f498d6700d487d25bef17fbd0/awscli/botocore/utils.py#L1622-L1623

awscli/s3transfer/crt.py
ashovlin: Moved back an indent level in b54f87c78886787f498d6700d487d25bef17fbd0

awscli/s3transfer/crt.py
ashovlin: Added this in b54f87c78886787f498d6700d487d25bef17fbd0

https://github.com/aws/aws-cli/blob/b54f87c78886787f498d6700d487d25bef17fbd0/awscli/s3transfer/crt.py#L291-L295

awscli/s3transfer/crt.py
ashovlin: Split in to two locks in b54f87c78886787f498d6700d487d25bef17fbd0
1.`CRTS3RegionRedirectPolicy._redirect_lock` - held when caching a region for the first time, which is the one that can include a `HeadBucket`
1.` BotocoreCRTRequestSerializer._region_cache_lock` — held while a request is being built via `locked_bucket_region`, and taken briefly by `cache_bucket_region()` to record a region.

hssyoo (approved):

awscli/customizations/s3/factory.py
```diff
@@ -275,10 +293,14 @@ def _create_crt_transfer_config(self, config_kwargs):
             multipart_threshold=config_kwargs.get('multipart_upload_threshold')
         )
 
-    def _create_crt_client(self, params, runtime_config, config_kwargs):
+    def _create_crt_client(
+        self, params, runtime_config, config_kwargs, region=None, bootstrap=None
+    ):
+        config_file_params = self._session.get_scoped_config().get('s3', {})
```
ashovlin: Extra after rebase?

hssyoo (approved):

ashovlin:
Merging to feature branch, current test failures are unreleated, and the internal build passed.

Issue #4097, opened 2019-04-23
--help should provide more useful output


alex:
Currently `aws s3 ls --hep` just says `Unknown options: --help`.

It could ideally a) Just do the right thing, or b) failing that, remind print that what you really want is `aws s3 ls help`.

justnance:
@alex - This sounds reasonable.  Marking it as a feature request.  Thanks!

justnance:
@alex - Looks like we are tracking this feature request under #303.  I'm close this issue to avoid duplicate efforts.

tim-finnigan:
Reopening this issue as the error message for `--help` was recently fixed for other commands (see https://github.com/aws/aws-cli/pull/7765) but still results in `unknown options --help` for `s3` commands.

dosmanak:
The help subcommand thing is ugly. `--help` argument effectively disables the command but help subcommand stated after args might be consumed as arg value causing unwanted behavior.

Issue #303, opened 2013-09-04
Support `help` command conventions (`-h/--help` or `aws help <command/service>`


kaihendry:
Most developers are familiar with `git help commit`.

Not `aws s3 help` which is jarring.

mfowlewebs:
Most developers are familiar with `-h` and `--help`. Help subcommands is kind of a new thing, seen in subcommand centric software. I'd go for `-h`/`--help` not saying `aws: error: too few arguments` before this.

mtdowling:
I'm :+1: on adding `-h/--help`. We will discuss this on the CLI team.

ASayre:
Good Morning!

We're closing this issue here on GitHub, as part of our migration to [UserVoice](https://aws.uservoice.com/forums/598381-aws-command-line-interface) for feature requests involving the AWS CLI.

This will let us get the most important features to you, by making it easier to search for and show support for the features you care the most about, without diluting the conversation with bug reports.

As a quick UserVoice primer (if not already familiar): after an idea is posted, people can vote on the ideas, and the product team will be responding directly to the most popular suggestions.

We’ve imported existing feature requests from GitHub - Search for this issue there!

And don't worry, this issue will still exist on GitHub for posterity's sake.  As it’s a text-only import of the original post into UserVoice, we’ll still be keeping in mind the comments and discussion that already exist here on the GitHub issue.

GitHub will remain the channel for reporting bugs.  

Once again, this issue can now be found by searching for the title on: https://aws.uservoice.com/forums/598381-aws-command-line-interface 

-The AWS SDKs & Tools Team

adamchainz:
What if it's not the morning in my timezone? That's rather presumptuous :)

jamesls:
Based on community feedback, we have decided to return feature requests to GitHub issues.

tim-finnigan:
For future reference, some ideas related to this were also mentioned in https://github.com/aws/aws-cli/issues/7629. Thanks @calebstewart for the suggestions [here](https://github.com/aws/aws-cli/issues/7629#issuecomment-1414234380) and [here](https://github.com/aws/aws-cli/issues/7629#issuecomment-1414236669). If the help convention doesn't change I think at least the default message when using `-h/--help` should point users to the existing `help` command (https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-help.html#cli-usage-help-command).

tim-finnigan:
[...]

PR #10630, opened 2026-09-10
[v2] Add --help parameter for all CLI commands

kdaily:
*Issue #, if available:*

- #4097
- #303

*Description of changes:*

Adds a `--help` parameter that renders the same help as the existing `help` subcommand, on every command: the top-level `aws`, a service (`aws ec2 --help`), an operation (`aws ec2 describe-instances --help`), and custom/nested commands (`aws configure get --help`). The `help` subcommand is unchanged.

## Implementation

Help intent is resolved before normal argument binding.

- New helpers in `awscli/argparser.py` detect and strip the exact `--help` token (`is_help_option_present`, `strip_help_options`, `first_help_option_index`).
- Each of the four dispatch layers (`CLIDriver.main`, `ServiceCommand`, `ServiceOperation`, `BasicCommand`) routes to its existing `create_help_command()` with `--help` stripped. Detecting it up front means a preceding value option cannot consume it.
- `--help` is added to `_NO_AUTO_PROMPT_ARGS` so a help request bypasses auto-prompt.

## Behavior
- **The position mirrors the `help` subcommand.** Help resolves at the depth reached when `--help` appears, and tokens (sub-commands or parameters) after it are ignored, so `aws ec2 --help describe-instances` renders EC2 service help. This is the same as `aws ec2 help describe-instances`.
- **Exact token only.** Only the literal `--help` is help. `--hel`, `--he`, and `--help=x` are not; they fall through as unknown options. This avoids colliding with operation parameters with a shared prefix, like `--health-check-type`.
- **Option values are not treated as commands.** `aws --region ec2 --help` renders top-level help, because `ec2` is `--region`'s value.

## Known limitation
`--help` is deliberately absent from the generated global-options synopsis and reference pages. Those are generated from `cli.json` and pinned by `tests/functional/test_globals.py`; `--help` is intentionally not a `cli.json` option, since that would re-enable prefix matching so `--hel` resolves again. Users find `--help` through the error text shown on a bad invocation and each command's own help output.

## Testing
- New functional tests cover `--help` on every command type, including value and optional-value options before `--help`, and `--help` before a command/operation token.
- New unit tests cover the `argparser` helpers, exact-token-only detection, value-aware routing, the pre-help-slice parse, and subcommand-vs-help positioning.

### DIFF
--- a/awscli/argparser.py
+++ b/awscli/argparser.py
@@ -21,6 +21,7 @@ HELP_BLURB = (
     "  aws help\n"
     "  aws <command> help\n"
     "  aws <command> <subcommand> help\n"
+    "  aws <command> [<subcommand> ...] --help\n"
 )
 USAGE = (
     "aws [options] <command> <subcommand> [<subcommand> ...] [parameters]\n"
@@ -58,6 +59,37 @@ class CommandAction(argparse.Action):
         # generated from the command table keys. So make this a
         # NOOP if argparse.Action tries to set this value.
         pass
+
+
+def _is_help_option_token(arg):
+    """Return True only for the literal ``--help`` token.
+
+    Abbreviations (``--he``/``--hel``/``--h``) and ``=``-bearing forms
+    (``--help=x``, ``--instance-ids=--help``) are not help; they fall through to
+    the normal parser.
+    """
+    return arg == '--help'
+
+
+def first_help_option_index(args):
+    """Return the index of the first ``--help`` token, or ``None`` if absent."""
+    for index, arg in enumerate(args):
+        if _is_help_option_token(arg):
+            return index
+    return None
+
+
+def is_help_option_present(args):
+    """Return True if the literal ``--help`` token is present in ``args``."""
+    return first_help_option_index(args) is not None
+
+
+def strip_help_options(args):
+    """Return ``args`` with every ``--help`` token removed.
+
+    All other tokens keep their original position and value.
+    """
+    return [arg for arg in args if not _is_help_option_token(arg)]
 
 
 class CLIArgParser(argparse.ArgumentParser):
--- a/awscli/clidriver.py
+++ b/awscli/clidriver.py
@@ -36,11 +36,15 @@ from botocore.history import get_global_history_recorder
 from awscli import __version__
 from awscli.alias import AliasCommandInjector, AliasLoader
 from awscli.argparser import (
+    ArgParseException,
     ArgTableArgParser,
     FirstPassGlobalArgParser,
     MainArgParser,
     ServiceArgParser,
     SubCommandArgParser,
+    first_help_option_index,
+    is_help_option_present,
+    strip_help_options,
 )
 from awscli.argprocess import unpack_argument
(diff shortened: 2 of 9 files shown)


awscli/clidriver.py
```diff
@@ -615,6 +629,44 @@ def main(self, args=None):
...
+        if help_index is None:
+            return list(args)
+        return list(args[:help_index])
+
+    def _route_to_provider_help(self, args):
+        # Render provider help: keep the tokens before the first --help and
+        # append the ``help`` positional argument, so a trailing command is
+        # ignored.
+        head = self._tokens_before_help(args)
+        return head + ['help']
```
AndrewAsseily, line 668: Could we avoid converting the help request into a positional help token here? The parse fails because `--query` has no value, so this rewrites argv to ... `--query help`, which strips the help intent and lets the command run. Worried about destructive operations like `aws s3api delete-object --bucket b --key k --query --help` when the intent is --help

AndrewAsseily: Worth a regression test on a mutating operation specifically

awscli/argparser.py
```diff
@@ -60,6 +61,37 @@ def choices(self, val):
...
+    Abbreviations (``--he``/``--hel``/``--h``) and ``=``-bearing forms
+    (``--help=x``, ``--instance-ids=--help``) are not help; they fall through to
+    the normal parser.
+    """
+    return arg == '--help'
+
+
+def first_help_option_index(args):
+    """Return the index of the first ``--help`` token, or ``None`` if absent."""
+    for index, arg in enumerate(args):
```
AndrewAsseily, line 76: Help detection should stop at the `--` end-of-options marker. `aws configure set test.key -- --help` normally treats `--help` as the positional value, but this interprets it as a help request and renders command help instead

awscli/clidriver.py
AndrewAsseily, line 612: Could help get consumed as the preceding option’s value here? For example: 
`aws ec2 terminate-instances --instance-ids i-abc123 --query --help`

Currently on v2 this fails because `--query` has no value.  With this change it effectively becomes `--query` help and calls `TerminateInstances`.

My read is `help` can be a query value, but `--help` is an explicit option. 
Could we preserve that distinction instead of reinserting it as a normal

aemous: I agree. It should behave more like other parameters. I'll revise.

awscli/argparser.py
aemous: Agreed, it should behave as other parameters in this aspect. I'll publish a revision

awscli/clidriver.py
aemous: Agreed. Supplying `--help` should never behave as supplying `help` as a value to another parameter.

awscli/clidriver.py
kdaily: I think this would mean that we need to enter help directly (calling the help command)?

aemous:
I am taking over ownership of implementing this feature.

Closing in favor of https://github.com/aws/aws-cli/pull/10657.

PR #10632, opened 2026-09-11
Emit warnings for unsupported TM config options

hssyoo:
There are 2 types of warnings we want to emit:
1. Cases where the resolved transfer client type doesn't support a configured option and ignores it.
2. Cases where CRT doesn't support a configured option and falls back to the classic TM. These are for options that CRT will never support, but we want to avoid changing the default TM client to minimize disruptions. Instead, we'll begin emitting warnings before completely switching to CRT by default and ignoring the configured option (same as 1).

### DIFF
--- a/tests/unit/customizations/s3/test_factory.py
+++ b/tests/unit/customizations/s3/test_factory.py
@@ -901,7 +901,7 @@ class TestClassicOnlySettingsWarning:
 def warn_unsupported_settings(auto_resolve_factory, capsys):
     def _warn(client_type, **kwargs):
         runtime_config = RuntimeConfig().build_config(**kwargs)
-        auto_resolve_factory._warn_unsupported_settings(
+        auto_resolve_factory.warn_unsupported_settings(
             client_type, runtime_config
         )
         return capsys.readouterr().err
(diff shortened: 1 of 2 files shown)


tests/unit/customizations/s3/test_factory.py
```diff
@@ -837,6 +838,167 @@ def test_s3s3_always_resolves_to_classic(
...
+            preferred_transfer_client='classic', max_bandwidth=1024
+        )
+        assert capsys.readouterr().err == ''
+
+
+@pytest.fixture
+def warn_unsupported_settings(auto_resolve_factory, capsys):
+    def _warn(client_type, **kwargs):
+        runtime_config = RuntimeConfig().build_config(**kwargs)
+        auto_resolve_factory._warn_unsupported_settings(
```
aemous: Looks like we're calling a private function here. I see there's existing precedent for this in this file with a different fixture.

Have we considered switching the private function's visibility to public?

aemous (changes requested):
Left one question

tests/unit/customizations/s3/test_factory.py
hssyoo: Switched to public method

aemous (approved):
LGTM.

PR #10634, opened 2026-09-11
Support multipart_threshold for downloads using CRT

hssyoo:
CRT's S3 client exposes `multipart_upload_threshold`, so it's straightforward to map the CLI's `multipart_threshold` config value. But it currently doesn't expose or support `multipart_download_threshold` so we work around it by comparing the object size to `multipart_threshold` before handing off the request to CRT, and explicitly set the request type to `S3RequestType.DEFAULT`. See https://github.com/awslabs/aws-crt-python/blob/3a967bb6bee00acd099c23aa3d2bbc28e30ccf24/awscrt/s3.py#L464-L470

I intentionally made `CRTTransferConfig` minimal, only supporting `multipart_threshold`, because in the future we're going to want to remove this work and directly use a supported `multipart_download_threshold` interface. I wanted to avoid other changes trying to use the CRT transfer config.

### DIFF
--- a/.changes/next-release/enhancement-crt-69912.json
+++ b/.changes/next-release/enhancement-crt-69912.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "crt",
+  "description": "Configure CRT client to download with single GET when object size is below ``multipart_threshold``"
+}
--- a/tests/functional/s3transfer/test_crt.py
+++ b/tests/functional/s3transfer/test_crt.py
@@ -23,6 +23,7 @@ from s3transfer.subscribers import BaseSubscriber
 from tests import (
     HAS_CRT,
     FileCreator,
+    FileSizeProvider,
     NonSeekableReader,
     NonSeekableWriter,
     mock,
@@ -834,14 +835,6 @@ class TestCRTTransferManager(unittest.TestCase):
             future.result()
 
 
-class ProvideSizeSubscriber(BaseSubscriber):
-    def __init__(self, size):
-        self._size = size
-
-    def on_queued(self, future, **kwargs):
-        future.meta.provide_transfer_size(self._size)
-
-
 MULTIPART_THRESHOLD = 8 * 1024 * 1024
 DOWNLOADED_CONTENT = 'content'
 
@@ -892,7 +885,7 @@ def download(crt_client, request_serializer, filename):
                 multipart_threshold=multipart_threshold
             ),
         )
-        subscribers = [] if size is None else [ProvideSizeSubscriber(size)]
+        subscribers = [FileSizeProvider(size)]
         transfer_manager.download(
             'test_bucket', 'test_key', filename, {}, subscribers
         ).result()
@@ -901,7 +894,6 @@ def download(crt_client, request_serializer, filename):
     return _download
 
 
-@requires_crt()
 class TestDownloadMultipartThreshold:
     @pytest.mark.parametrize(
         'size', [MULTIPART_THRESHOLD - 1, MULTIPART_THRESHOLD]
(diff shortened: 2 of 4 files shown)


.changes/next-release/enhancement-crt-69912.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "crt",
+  "description": "Configure CRT client to download with single GET when object size is below ``multipart_threshold``"
+}
```
aemous, line 5: Are we planning to push a changelog for the emitted warnings to the `crt-config-v2` feature branch?

tests/functional/s3transfer/test_crt.py
```diff
@@ -831,3 +832,101 @@ def test_crt_s3_client_error_handling(self):
...
+        subscribers = [] if size is None else [ProvideSizeSubscriber(size)]
+        transfer_manager.download(
+            'test_bucket', 'test_key', filename, {}, subscribers
+        ).result()
+        return crt_client.make_request.call_args[1]
+
+    return _download
+
+
+@requires_crt()
```
aemous: nit: do we need the `requires_crt()` decorator if cli v2 always bundles with crt and we don't have plans to backport this to upstream botocore?

tests/functional/s3transfer/test_crt.py
```diff
@@ -831,3 +832,101 @@ def test_crt_s3_client_error_handling(self):
         )
         with self.assertRaises(awscrt.exceptions.AwsCrtError):
             future.result()
+
+
+class ProvideSizeSubscriber(BaseSubscriber):
```
aemous: nit: why couldn't we use the existing `ProvideSizeSubscriber`?

tests/functional/s3transfer/test_crt.py
```diff
@@ -831,3 +832,101 @@ def test_crt_s3_client_error_handling(self):
...
+        assert request_args['type'] == awscrt.s3.S3RequestType.GET_OBJECT
+        assert 'operation_name' not in request_args
+
+    def test_unset_threshold_splits_download(self, download):
+        request_args = download(1, None)
+        assert request_args['type'] == awscrt.s3.S3RequestType.GET_OBJECT
+
+    def test_unknown_size_splits_download(self, download):
+        request_args = download(None, MULTIPART_THRESHOLD)
+        assert request_args['type'] == awscrt.s3.S3RequestType.GET_OBJECT
```
aemous: can you explain why these would use MP download? test_unset_threshold_splits_download looks like it resembles a typical HEAD + provide size + download in today's CLI, which does a single GET

aemous (commented):
Left a few clarifying questions.

tests/functional/s3transfer/test_crt.py
hssyoo: For `test_unknown_size_splits_download`, this isn't practically reachable since AWS CLI always gets object size. You'd have to directly invoke the vended `s3transfer` library. Going to remove this test.

For `test_unset_threshold_splits_download`, this tests the path where either an optimized host resolved `auto` to `crt` (so only the hosts that already auto-resolve to `crt` today) or user explicitly configured `crt`. We want to avoid changing default behavior for these users (ie Don't set Python's 8MB default and use CRT defaults instead, which is multipart download).

tests/functional/s3transfer/test_crt.py
hssyoo: Reusing

tests/functional/s3transfer/test_crt.py
hssyoo: Dropped

.changes/next-release/enhancement-crt-69912.json
hssyoo: Sure we can add it when we merge the feature branch. Planning on potentially consolidating changelogs anyways

aemous (approved):
LGTM.

PR #10456, opened 2026-06-29
fix: UTF-8 encoding for MCP config files (#10452) and --redirect-port for sso login (#10433)

Hardikrepo:
### Fix #10452 — MCP config files fail to read/write on non-English Windows (CP932 encoding)

**Root cause:** `DetectedAgent._read_mcp_config()` and `_write_mcp_config()` in
`awscli/customizations/agenttoolkit/agents.py` called `open()` without specifying
an encoding. On non-English Windows (e.g. Japanese), Python defaults to the system
locale codec (CP932), which cannot decode UTF-8 multibyte characters such as the
em-dash (U+2014, `0xE2 0x80 0x94` in UTF-8) commonly found in MCP config files
like `~/.claude.json`. This caused a `UnicodeDecodeError` and aborted the
`aws configure agent-toolkit` flow.

**Fix:** Explicitly pass `encoding='utf-8'` to both `open()` calls so MCP config
files are always read and written as UTF-8 regardless of the system locale.

---

### Fix #10433 — Allow custom localhost port for PKCE redirect URI in `aws sso login`

**Root cause:** The PKCE callback server (`AuthCodeFetcher`) always bound to a
randomly assigned OS port (`HTTPServer(('', 0), ...)`). Users on remote machines
accessed over SSH cannot pre-configure port forwarding without knowing the port
in advance, forcing them to fall back to the less ergonomic `--use-device-code` flow.

**Fix:**
- Added `--redirect-port <int>` CLI argument to `aws sso login`
- Added `sso_redirect_port` config key support in `~/.aws/config` under `[sso-session]`
- `AuthCodeFetcher.__init__()` now accepts an optional `port` parameter (default `0` preserves existing random-port behaviour)
- Port resolution order: `--redirect-port` arg → `sso_redirect_port` config key → random port

**Example usage:**
```bash
# Pre-configure SSH tunnel
ssh -L 34535:127.0.0.1:34535 remote-host

# Then on the remote host
aws sso login --redirect-port 34535
```

Or via `~/.aws/config`:
```ini
[sso-session my-sso]
sso_start_url = https://example.awsapps.com/start
sso_region = us-east-1
sso_redirect_port = 34535
```

## Files Changed

| File | Change |
|------|--------|
| `awscli/customizations/agenttoolkit/agents.py` | Add `encoding='utf-8'` to both `open()` calls |
[...]

PR #10636, opened 2026-09-14
Add configurable redirect port for SSO and login OAuth flows

ashovlin:
*Issue #, if available:* #9148, #10433

*Description of changes:* This begins addressing redirect port configurablity by adding `--redirect-port` to:
1. `aws login`
2. `aws sso login`
3. `aws configure sso`

It pulls from some open PRs:
1. #10462 - SSO, but not `login`
2. #10134 - adds host override as well, which I still need to think through security implications of
3. #10456 - bundled with an unrelated encoding fix

I still think we should work towards a profile config option, but that triggers more internal work (amending spec so AWS Tools for PowerShell honors it too, dev guide updates), so starting with this.


GarrettBeatty:
change looks fine to me. can you check if copilots comments are valid or not?

luigi617 (approved):
LGTM

ixxq:
Hi @ashovlin , Thanks for referencing #10462 here.

I opened #10433 and then opened #10462 after discussing it with the maintainers
and being told that a PR would be welcome and reviewed by the team.

My preference would be for #10462 to be reviewed directly, with any requested
changes addressed there if needed.

That said, if this PR is intended to supersede #10462 while substantially
reusing the idea, implementation approach, or test coverage from it, would you
be open to including me as a co-author on the relevant commit?

ashovlin:
Hey @ixxq , thanks for replying, sorry about jumping ahead here. We have a few competing PRs, so my instinct was to pick what we're initially comfortable with.

If you're willing to update https://github.com/aws/aws-cli/pull/10462 with these, we can proceed over there:
1. Support `aws login` too, with the same command line option.
2. Skip `sso_redirect_port` for now - we'd have to coordinate with other tools that support the SSO flow internally, so I think it's faster to get the command line option out first.

ixxq:
@ashovlin Thanks for being open to moving forward with #10462.

I'm happy to update it to support `aws login` with the same `--redirect-port` option and remove `sso_redirect_port` for now.

Just to confirm the scope: should I also include `aws configure sso --redirect-port`, as in #10636, or only `aws login` and `aws sso login`?

ashovlin:
@ixxq , ah yes, sorry I left it out of the last reply. `aws configure sso --redirect-port` too, thanks!

PR #10638, opened 2026-09-14
Conditionally set default crt target_throughput

hssyoo:
Conditionally sets a default `target_throughput` value for the CRT client:
* (Unchanged) If host is optimized, use value from `get_recommended_throughput_target_gbps()`.
* If newly eligible host auto-resolving to CRT and `get_recommended_throughput_target_gbps()` returns `None`, set it to 4gbps.
  * 4gbps maps to a max 256MiB memory pool. Since we don't have host details, we use the smallest supported pool size.
* If CRT is explicitly configured, use the maximum value of (`get_recommended_throughput_target_gbps()`, 10gbps).
  * 10gbps is CRT's default when there's no recommended throughput. We don't want to lower throughput for instances that now return a recommended throughput <10gbps.

There's an edge case where we need to fall back to the classic transfer manager if the configured `multipart_chunksize > 128MiB` and it's a newly eligible host auto-resolving to CRT and there's no recommended throughput for the host. CRT client creation has a guard that ensures `part_size < memory_limit / 2`. Since under these conditions we set 4gbps throughput with a 256MiB memory limit, a configured part size of 128MiB will always fail. Instead of dynamically adjusting the memory limit based on part size (potentially regressing on OOM issues), we just fall back to Python.


awscli/customizations/s3/factory.py
```diff
@@ -170,8 +188,24 @@ def _get_unsupported_settings(self, params, runtime_config):
...
+    def _is_multipart_chunksize_too_large(self, runtime_config):
+        # Only the pool for untuned systems is small enough to reject a
+        # configurable chunksize.
+        if not self._is_untuned_system():
+            return False
+        return (
+            runtime_config.is_explicitly_set('multipart_chunksize')
+            and runtime_config['multipart_chunksize']
+            > UNTUNED_MAX_MULTIPART_CHUNKSIZE
+        )
```
ashovlin, line 207: Should this consider if the user set an explicit target bandwidth too?

Say I'm running locally, so there isn't a CRT tuning available. I set a > 128MiB chunk size, but also an explicit >10Gbps target bandwidth, wouldn't CRT still be acceptable then?

awscli/customizations/s3/factory.py
```diff
@@ -323,12 +357,49 @@ def _resolve_crt_client_config_kwargs(self, runtime_config):
...
+        if self._is_newly_eligible_for_crt_client(runtime_config) and (
+            self._is_untuned_system()
+        ):
+            # The crt client sizes its memory pool from the throughput target.
+            # Without a recommendation it assumes 10gbps, which maps to a max
+            # pool size of 2GiB. Newly-eligible hosts that auto-resolve to crt
+            # may not be able to afford 2GiB, so it sets the maximum throughput
+            # that maps to the smallest 256MiB tier.
+            return _gbps_to_bytes_per_sec(UNTUNED_TARGET_THROUGHPUT_GBPS)
+        return None
```
ashovlin, line 380: I have a 1.3TB file (`truncate -s 1342177280001 /tmp/pr10638-large-sparse`), so that parts need to be larger than `UNTUNED_MAX_MULTIPART_CHUNKSIZE`.

This fails in CRT, I think because it requires a part size larger than the memory budget we get with `UNTUNED_TARGET_THROUGHPUT_GBPS`:
```
AWS_CLI_AUTO_RESOLVE_CLIENT=crt aws s3 cp /tmp/pr10638-large-sparse s3://placeholder-test/  --profile pr10638           
upload failed: ../../../../tmp/pr10638-large-sparse to s3://placeholder-test/pr10638-large-sparse 34 (AWS_ERROR_INVALID_ARGUMENT): An invalid argument was passed to a function.
```

But proceeds on existing code base for Python:
```
  aws s3 cp /tmp/pr10638-large-sparse s3://placeholder-test/ --profile pr10638                               
Completed 146.0 MiB/1.2 TiB (63.2 MiB/s) with 1 file(s) remaining
```

Profile for this:
```
[profile pr10638]
region = us-east-1
s3 =
    preferred_transfer_client = auto
```

ashovlin (changes requested):

awscli/customizations/s3/factory.py
hssyoo: This will be fixed via an upstream CRT change. When size is >1TiB, CRT automatically switches to streaming (vs buffering). Upstream change will avoid the part size/memory limit check in the case of streaming uploads.

awscli/customizations/s3/factory.py
hssyoo: Fixed via https://github.com/aws/aws-cli/pull/10638/changes/d5d74f0adb0939915effa03b1fc916a6ccd90574

Instead of attempting to hardcode the max part size and guess whether the configurations/hardware are valid, it just attempts to create the CRT client and catches the part size error and falls back to classic.

awscli/customizations/s3/factory.py
```diff
@@ -102,13 +122,38 @@ def create_transfer_manager(
...
         if client_type == constants.CRT_TRANSFER_CLIENT:
+            transfer_manager = self._try_create_crt_transfer_manager(
+                params, runtime_config
+            )
+            if transfer_manager is not None:
+                self.warn_unsupported_settings(client_type, runtime_config)
+                return transfer_manager
+            client_type = constants.CLASSIC_TRANSFER_CLIENT
+        self.warn_unsupported_settings(client_type, runtime_config)
+        return self._create_classic_transfer_manager(
```
ashovlin, line 136: Should we release the CRT lock in this branch?

hssyoo: Fixed in https://github.com/aws/aws-cli/pull/10638/changes/ae0bf95a730693c3aaff639465ca4089ec4eedab

ashovlin (approved):

PR #10644, opened 2026-09-15
Preserve sub-second precision when serializing unixTimestamp parameters

jonathan343:
## Overview

Ports boto/botocore#3796 and fixes the CLI v2 equivalent of boto/botocore#3255.

CLI v2's vendored botocore serializes `unixTimestamp` request parameters as `int(calendar.timegm(...))`, dropping any sub-second precision the caller provided. This affects every `json`, `rest-json`, and `smithy-rpc-v2-cbor` body timestamp. `iso8601` has always carried microseconds, and `rfc822` (HTTP-date) only supports whole seconds, so only the epoch-seconds format had this issue.

This PR preserves the caller's sub-second precision for `unixTimestamp`. Whole-second timestamps serialize exactly as before.

The CBOR serializer now emits a tagged float64 for fractional timestamps and delegates integer encoding to the shared integer serializer, which also fixes an `OverflowError` for pre-1970 timestamps.

## Testing

Reproduced the scenario from boto/botocore#3255 with `kinesis-video-archived-media list-fragments`, pointed at an unreachable endpoint so `--debug` logs the serialized body before the send fails:

```bash
aws kinesis-video-archived-media list-fragments \
  --stream-name some-stream \
  --fragment-selector '{
    "FragmentSelectorType": "PRODUCER_TIMESTAMP",
    "TimestampRange": {
      "StartTimestamp": "2024-09-12T10:49:36.500000Z",
      "EndTimestamp": "2024-09-12T10:49:38.833000Z"
    }
  }' \
  --debug 2>&1 | grep -o "'body': b'[^']*'"
```

**Before** (`v2`):

```
'body': b'{"StreamName": "some-stream", "FragmentSelector": {"FragmentSelectorType": "PRODUCER_TIMESTAMP", "TimestampRange": {"StartTimestamp": 1726138176, "EndTimestamp": 1726138178}}}'
```

**After**:

```
'body': b'{"StreamName": "some-stream", "FragmentSelector": {"FragmentSelectorType": "PRODUCER_TIMESTAMP", "TimestampRange": {"StartTimestamp": 1726138176.5, "EndTimestamp": 1726138178.833}}}'
```

---


awscli/botocore/serialize.py
```diff
@@ -617,22 +624,12 @@ def _serialize_type_timestamp(self, serialized, value, shape, key):
...
-        additional_info, num_bytes = self._get_additional_info_and_num_bytes(
-            timestamp
-        )
-
-        if num_bytes == 0:
-            initial_byte = self._get_initial_byte(
-                self.UNSIGNED_INT_MAJOR_TYPE, timestamp
-            )
-            serialized.extend(initial_byte)
+        # Tag 1 permits either an integer or a floating-point epoch seconds
```
GarrettBeatty, line 627: code seems fine. just for my understanding why did we have all the `_get_additional_info_and_num_bytes` logic before and why can it be removed now?

jonathan343: The `_serialize_type_integer` we now use calls the same `_get_additional_info_and_num_bytes` helper under the hood. I'm honestly not sure why this wasn't called in the initial implementation.

But the current behavior is incorrect since we can't serialize negative timestamps. A timestamp value like `datetime(1968, 12, 31, 23, 59, 59)` currently results in a `OverflowError: can't convert negative int to unsigned` error which this new approach addresses.

GarrettBeatty (approved):

awscli/botocore/serialize.py
```diff
@@ -143,7 +143,14 @@ def _timestamp_iso8601(self, value):
...
 
     def _timestamp_unixtimestamp(self, value):
-        return int(calendar.timegm(value.timetuple()))
+        timestamp = calendar.timegm(value.timetuple())
+        if value.microsecond > 0:
+            # Add the microseconds as integers before dividing so the float is only
+            # rounded once. Dividing first and then adding rounds twice, which can
+            # produce a slightly different value, e.g. 1 + 3691 / 10**6 gives
+            # 1.0036909999999999 instead of 1.003691.
+            return (timestamp * 10**6 + value.microsecond) / 10**6
```
AndrewAsseily, line 152: `_timestamp_rfc822()` also calls this method. Because the new result is a float (https://docs.python.org/3/tutorial/floatingpoint.html#floating-point-arithmetic-issues-and-limitations) wouldn't far-future timestamps round up by one second?

jonathan343: https://github.com/boto/botocore/pull/3568 introduced a change to the `_timestamp_rfc822` 12 months ago that the CLI hadn't picked up. This change updated the method to remove sub-second precision (by casting to `int`). I'll get that ported over in this PR to align with botocore. 

Note: I confirmed this shouldn't have any impact since we never sent sub-seconds to begin with. The `formatdate` method truncated the sub-second digits anyways:

```bash
aws s3api get-object \
  --bucket gytndd-testing123 \
  --key test.txt \
  --response-expires '2026-01-01T12:00:00.123456Z' \
  /tmp/s3out --debug
```

Confirmed that `response-expires=Thu%2C%2001%20Jan%202026%2012%3A00%3A00%20GMT` is send on the wire response and the service sends back `"Expires": "2026-01-01T12:00:00+00:00"`

jonathan343: pushed in 4bb2b5a9e87cdee36163e2b74ac5450f4fe58210

AndrewAsseily (approved):

PR #10653, opened 2026-09-16
Add upper bound to max attempts

hssyoo:
Add max attempts for crt client


ashovlin (approved):
nit: can also update https://github.com/aws/aws-cli/blob/9a0483ed57649bdaabd6efb2b5ce5614ffb7507c/awscli/topics/s3-config.rst?plain=1#L390-L392

PR #10663, opened 2026-09-18
Document SigV2 default for aws s3 presign

AndrewAsseily:
*Description of changes:* `aws s3 presign` emits SigV2 URLs by default in regions where S3 still supports it (botocore's `_default_s3_presign_to_sigv2`). Query parameters appended after the URL is generated aren't covered by a SigV2 signature, so the URL holder can change them freely.

Mirrors boto/botocore#3791 

*Testing:*
  
Screenshot 2026-09-18 at 1 49 55 PM


luigi617 (approved):
LGTM

PR #10670, opened 2026-09-21
Include web identity token in cache key to prevent credential collision

luigi617:
*Issue #, if available:*

*Description of changes:*
Include web identity token in cache key to prevent credential collision


awscli/botocore/credentials.py
```diff
@@ -993,6 +1001,11 @@ def __init__(
             expiry_window_seconds=expiry_window_seconds,
         )
 
+    def _cache_key_args(self):
```
GarrettBeatty, line 1004: have you tested this code end to end besides unit test and if so how? can you add to PR description

luigi617: isn't the unit test not enough? I have a small script that I used for check:
```
import unittest.mock as mock
from awscli.botocore import credentials
from awscli.botocore.utils import FileWebIdentityTokenLoader

# 1. Define two token files with different contents (same role)
with open("/tmp/token_a.jwt", "w") as f: f.write("header.payload-AAA.sig")
with open("/tmp/token_b.jwt", "w") as f: f.write("header.payload-BBB.sig")

ROLE = "arn:aws:iam::123456789012:role/demo-role"
client_creator = mock.Mock()

def key_for(token_file):
    loader = FileWebIdentityTokenLoader(token_file)
    fetcher = credentials.AssumeRoleWithWebIdentityCredentialFetcher(
        client_creator, loader, ROLE
    )
    return fetcher._cache_key

ka = key_for("/tmp/token_a.jwt")
kb = key_for("/tmp/token_b.jwt")
print("token_a -> cache file:", ka + ".json")
print("token_b -> cache file:", kb + ".json")
print("same file?", ka == kb, "-> ", "COLLISION (bug)" if ka == kb else "distinct (fixed)")
```

GarrettBeatty (dismissed):

awscli/botocore/credentials.py
aemous: Specifying a workflow that uses AWS CLI v2 commands end-to-end and describing the resulting behavior would be useful.

ashovlin (changes requested):
Please add a changelog entry

GarrettBeatty:
> Please add a changelog entry

https://github.com/aws/aws-cli/pull/10687/changes made a pr to help detect this

awscli/botocore/credentials.py
luigi617: Added in description

PR #10671, opened 2026-09-21
Verify install script integrity in the update command

GarrettBeatty:
## Summary
`aws update` and the install scripts downloaded `install.sh`/`install.ps1` over HTTPS and executed them with no integrity or authenticity check. This closes two gaps:

**1. Install script integrity** — `update` now verifies the downloaded install script *before* running it:
- **Linux**: downloads `install.sh.sig` and GPG-verifies it against the embedded AWS CLI Team public key.
- **Windows**: `Get-AuthenticodeSignature` requiring a valid signature whose signer organization is AWS and whose issuer is DigiCert.

**2. Installer verification hardening** in the scripts themselves:
- `install.ps1` now confirms the MSI was signed *by AWS* (signer org + issuer), not merely that it is signed.
- `install.sh` now hard-fails when `gpg` is missing instead of silently skipping verification.

Verification is secure-by-default, with an explicit opt-out (`--skip-signature-verification` / `-SkipSignatureVerification`).

## Design notes
- The embedded PGP key in `update.py` is asserted to match `scripts/install-v2/install.sh` by a unit test to prevent drift.

## Testing
- Unit tests added/updated in `tests/unit/customizations/test_update.py` (verification pass/fail/opt-out for both platforms + PGP key drift test).
- Windows identity check validated end-to-end against a signed MSI on a Windows host.


GarrettBeatty:
Superseded by #10672 (moved from fork to a branch on aws/aws-cli).

PR #10674, opened 2026-09-22
Enable CRT S3 transfer client by default on specialized instance families

hssyoo:
Phase 1 of https://github.com/aws/aws-cli/issues/10676

Enable the CRT S3 transfer client by default on the following EC2 instance families - dl1, g6e, g6, g5, g5g, g4dn, inf2, inf1, x2iedn, x2idn, x2iezn, x1e, x1, i4i, i3en, i3, is4gen, im4gn, d3en, d3, h1.

Manually tested on a g5g instance family and confirmed that CRT is enabled by default and `Ag` (`S3_TRANSFER_CRT_AUTO`) feature id was registered in user agent.


awscli/customizations/s3/factory.py
```diff
@@ -227,10 +253,19 @@ def _is_eligible_for_crt_client(self, params, runtime_config):
...
+        if (
             os.environ.get('AWS_CLI_AUTO_RESOLVE_CLIENT')
             == constants.CRT_TRANSFER_CLIENT
-        )
+        ):
+            return True
+        return self._is_rolled_out_instance_family()
+
+    def _is_rolled_out_instance_family(self):
+        instance_type = awscrt.s3.get_ec2_instance_type()
```
ashovlin, line 264: https://awslabs.github.io/aws-crt-python/api/s3.html#awscrt.s3.get_ec2_instance_type
> Finally, it will call IMDS and return the instance type from there. Note that in the case of the IMDS call, a new client stack is spun up using 1 background thread. The call is made synchronously with a 1 second timeout: It’s not cheap. 

Is this adding a second to S3 commands outside of EC2?

hssyoo: In the worst case, yeah. I think there's a followup optimization we can make, since we already call `awscrt.s3.is_optimized_for_system()` before this. I imagine we could cache the returned instance type and reuse it, but would require upstream CRT changes

ashovlin: Locally on MacOS it is <1ms, but not sure yet if there are cases where it will use the full second.

ashovlin: If we're already calling it and it caches (CRT docs says it does), that's okay.

ashovlin (approved):

PR #10675, opened 2026-09-22
Add and refresh EC2 Image Builder CLI examples

natesun-aws:
*Issue #, if available:* N/A

*Description of changes:*
Adds examples for the 38 imagebuilder commands that had none, and replaces the 38 existing examples, which dated from the service's 2019 launch and no longer reflected current output shapes or documentation links. Every imagebuilder command now has an example except get-marketplace-resource, which supports the AWS Marketplace integration and isn't intended for direct use.

All commands were run against the live service and the output captured from real responses, then sanitized with the standard documentation placeholders.


aemous:
This PR targets `develop` branch which is for AWS CLI v1, which is now in maintenance mode and will not receive this doc update.

Verify your PR is applicable to AWS CLI v2, then update the target branch to be `v2`.

aemous (changes requested):
Update/reopen PR to target `v2`.

natesun-aws:
> Update/reopen PR to target `v2`.

Thanks - updated PR to target `v2` branch instead.

natesun-aws:
Looks like unrelated flaky test failures.

